PR #76, opened 2020-11-12
Remove Singularity image path

drpatelh:
Given that pulling Singularity images directly isn't supported via `https` by Nextflow yet (see [here](https://github.com/nextflow-io/nextflow/issues/1791#issuecomment-726106010)) it makes sense to remove these paths from module files until this feature is added. Nextflow would need to recognise that the containers need to be downloaded and add them to a cache as is the normal behaviour for Singularity images converted from Docker.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -17,20 +17,20 @@ jobs:
       - name: Run Markdownlint
         run: markdownlint ${GITHUB_WORKSPACE} -c ${GITHUB_WORKSPACE}/.markdownlint.yml
 
-  EditorConfig:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
+  # EditorConfig:
+  #   runs-on: ubuntu-latest
+  #   steps:
+  #     - uses: actions/checkout@v2
 
-      - uses: actions/setup-node@v1
-        with:
-          node-version: "10"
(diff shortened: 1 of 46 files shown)


grst:
why not just use the http url then? 

For me both protocols work: 

```bash
singularity pull http://depot.galaxyproject.org/singularity/fastqc:0.11.9--0
singularity pull https://depot.galaxyproject.org/singularity/fastqc:0.11.9--0    
```

drpatelh:
Yes, but the containers aren't cached properly as they are for example when you pull and convert Docker images. There also seem to be other more sinister things going on possibly at the kernel level as highlighted by @apeltzer in point (1) in [this comment](https://github.com/nextflow-io/nextflow/issues/1791#issue-741516473).

I think we should wait for this to be tested and implemented properly on the Nextflow side before we start introducing it into workflows otherwise the pipelines will break way more often that anyone has the time to keep up 😅

grst:
I see... let's merge this then!

grst (approved):

drpatelh:
Thanks @grst !

drpatelh:
I have also commented out the `EClint` for now because I don't quite know what to do with it and it may be easier to just deal with that all at once. Also, if we are likely to have traffic coming here after the rnaseq release be good if all of the tests were passing!

Issue #83, opened 2020-11-24
Run a test on all modules which have been modified in that push / PR, all in parallel in separate jobs


Emiller88:
I was wondering if using the bits @ewels posted on slack, we could solve both of these issues and introduce a "test autodiscovery" from a single github action, that spawns "pytest workflow" for each changed folder using the "test matrix" strategy. Like that the pytest-workflow test-dir could be set to each module directory, and the  `tests` folder contained in each module. 

IMO, that would clean up quite a few redundancies.

_Originally posted by @grst in https://github.com/nf-core/modules/issues/80#issuecomment-733105099_

Without separate workflow files for each module.

Emiller88:
This may or may not be worth it. See https://github.com/nf-core/modules/runs/1450514231?check_suite_focus=true

Mainly because there won't be a workflow for each individual software module anymore. It'll just be one big workflow `software` that will trigger whenever those files are changed, but it may get broken when there's a change to multiple software packages.

grst:
There's also another limitation: The matrix strategy only supports up to 256 jobs, while there can be unlimited workflow files. 

(It could be debated, it it really makes sense to run everything in a separate job, since most of the CI time is probably spent setting up the environment to then run a test for 2 seconds) 

If we stick with the *one-workflow-per-module* approach, we could maybe create a `pytest-workflow` [github action](https://docs.github.com/en/free-pro-team@latest/actions/creating-actions/creating-a-docker-container-action) to reduce redundant code. 

If we go for *one-workflow-per-repo*, then the way this is done in the docker action seems like the way to go. In fact, I don't see why this wouldn't work with multiple changed modules. What is your concern here?

Emiller88:
Ooo, I'll definitely make the github action at least. We could also have a nextflow install action if we wanted too. I think that's the best way to cut down on the boilerplate.

For *one-workflow-per-repo*, sorry if I wasn't clear. With one-workflow-per-module, we get
![image](https://user-images.githubusercontent.com/20095261/100234601-384f1600-2f23-11eb-9c49-1076adf71c84.png)

With one workflow it'll just say `Test software` and you'll have to sort through all the recent tests, and they won't have a descriptive name like `fastqc`.

grst:
ok, I see your point. With one-workflow-per-repo, this would have to be replaced with some nice logging to the test console. 
But yes -- it's not as clear as with one-workflow-per-module.

Emiller88:
[...]

PR #85, opened 2020-11-25
Refactor all tests to use pytest-workflow

Emiller88:
A continuation of #80


Emiller88:
I think this is ready to review. Few things:

- This doesn't cover #83.

drpatelh:
> I need input on the nextflow test workflow names, currently just doing test_software_command. I'm thinking we could do without the software portion, and then maybe follow the command with some better descriptions for bad inputs etc.

I think `software_command` is fine no? You mean in the `-entry` section of the `command` right? Be good to keep both just in case different tools have the same command. Or am I missing something else?

> I need input on a standard test name format for the basic tests ex: - name: Run fastqc paired-end test workflow or Run fastqc paired-end since it's always a test workflow.

Be great to keep this as generic as possible. But I guess some customisation will be required because different modules will require different tests. If `fastqc` is mentioned somewhere else in the name of the Action then just "Test paired-end" would be fine?

Emiller88:
> I think `software_command` is fine no? You mean in the `-entry` section of the `command` right? Be good to keep both just in case different tools have the same command. Or am I missing something else?

I mean `workflow software_command {` When the test gets ran it prefixes the `- run:` title before it so the `test_software_command` may not have to be so verbose, but I think I like the verbose now that I've slept on it. 

> Be great to keep this as generic as possible. But I guess some customisation will be required because different modules will require different tests. If `fastqc` is mentioned somewhere else in the name of the Action then just "Test paired-end" would be fine?

Yeah, exactly I plan on added a few tests that I stumbled upon writing them, like feeding two reads in and specifying that it's single end is true also, that should throw an error. Just to cover possible regressions. So the run section might say `Run fastqc single-end but pass it two reads` and it's sibling `Run fastqc paired-end but pass it one read`

Emiller88:
@drpatelh @ewels I'm not sure why the CI isn't triggering, can either of you try to kick it off manually for me? I think it may have to do with the subdirectory under workflows not being something they support. I tried to move the fastqc file out to see if it would trigger but it didn't work :upside_down_face:

grst:
Try to quote the paths that contain a wildcard: 

>  If you start a pattern with *, [, or !, you must enclose the pattern in quotes.

(https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet)

It also seems that subdirectories in `.github/workflows` are not supported: 
 * https://github.community/t/feature-request-allow-workflow-configuration-in-sub-folders/16776
 * https://stackoverflow.com/questions/62288443/where-should-you-put-the-git-hub-workflow-directory-for-actions-in-a-full-stack

grst (approved):

Emiller88:
See I thought the same thing when I saw that, but those are talking about separate `.github` dirs not just subdirectories in the`.github/workflows` directory. 

I also don't see an issue with the wildcards because they end in it, which is fine according to the docs.

Emiller88:
Alright I think that solved the CI issues. Sad we can't have more organization. :(

I still need to make issues for modules that haven't been tested and possible e2e tests. But other than that these should be ready to merge.

Created #88 to follow up on what versions(and if multiple) should be supported for CI.

drpatelh:
Thanks @Emiller88! Are the tests supposed to be failing here?

Emiller88:
They are not! It's not giving me a reason that they failed it just had an "internal error" I kicked them off again and lets see if they work this time around.

drpatelh (approved):
Thanks alot @Emiller88 ! This is definitely a good start 😎 🎉

PR #87, opened 2020-11-30
Use STAR to build initial index for RSEM

drpatelh:


software/rsem/preparereference/main.nf
```diff
@@ -24,15 +24,41 @@ process RSEM_PREPAREREFERENCE {
...
+            rsem/genome
 
-    rsem-calculate-expression --version | sed -e "s/Current version: RSEM v//g" > ${software}.version.txt
-    """
+        rsem-calculate-expression --version | sed -e "s/Current version: RSEM v//g" > ${software}.version.txt
+        """
+    } else {
+        """
+        mkdir rsem
+        rsem-prepare-reference \\
```
grst, line 54: Would it make sense to clean up the redundancies? 
E.g. by assigning the entire STAR command to a variable and include it into the script via `${star_command}`.

drpatelh: We could do but I have been consistent so far in all of the modules not to have massive lines containing commands 😅 And `if / else` statement takes up more space but is more readable I think.

grst: Actually, this is what I meant (untested) -- it still can make use of nice multi line strings. 

I still buy the point with consistency. 


```nextflow
  script:                                                                     
    def software = getSoftwareName(task.process)                                
    def args     = options.args.tokenize()                                      
    def star_cmd = ""                                                           
    if (args.contains('--star')) {                                              
        args.removeIf { it.contains('--star') }                                 
        def memory = task.memory ? "--limitGenomeGenerateRAM ${task.memory.toBytes() - 100000000}" : ''
        star_cmd = """                                                          
        STAR \\                                                                 
            --runMode genomeGenerate \\                                         
            --genomeDir rsem/ \\                                                
            --genomeFastaFiles $fasta \\                                        
            --sjdbGTFfile $gtf \\                                               
            --runThreadN $task.cpus \\                                          
            $memory \\                                                          
            $options.args2                                                      
        """                                                                     
    }                                                                           
                                                                                
    """                                                                         
    mkdir rsem                                                                  
    ${star_cmd}                                                                 
    rsem-prepare-reference \\                                                   
        --gtf $gtf \\                                                           
        --num-threads $task.cpus \\                                             
        ${args.join(' ')} \\                                                    
        $fasta \\                                                               
        rsem/genome                                                             
                                                                                
    rsem-calculate-expression --version | sed -e "s/Current version: RSEM v//g" > ${software}.version.txt
    """                                                                         
```

drpatelh: Yup, agreed that is much nicer than having a single-long line but I feel it still breaks up the logic abit when you need to try and figure out where commands are coming from. In this instance, its not that bad because you only have `star_cmd` but in others there may be more.

grst (approved):
ok, let's merge this, then!

PR #91, opened 2020-12-09
Add CODEOWNERS

### LINKED ISSUE #86: Add codeowners
To get pinged and have actual owners who keep up with the software that the modules are taking advantage of.

Emiller88:
Closes #86


Emiller88:
Alright this is ready to review. I've gone through all of the modules and added anyone in the `meta` file to the codeowners. If they didn't have a specific author, I just marked them as TODO and maintainers can PR, or comment here if they want them. I have a few lines I'd like opinions on.

```
*       @drpatelh 
.github/workflows @Emiller88 
tests/ @Emiller88 
```

I think @drpatelh should be the default owner on all of them for now. I'd also like to help maintain the tests and the CI for this repo, I may pick up some orphan modules also. 

I think this will make the review process for changes quicker in the future, I think the notifications on this repo might get noisy in the future.

If anyone else wants to toss their name into anything feel free to drop a comment.

grst:
Is this still a thing? How would we keep this file up-to-date?

maxulysse:
I like this idea
Is there a way to scrape the meta.yml files to update this codeowner file via a github action for example?

Emiller88:
I think this would be a big boost for keeping things maintained.

I think we could probably get away with a script to write the original one, then update it from there?

https://github.com/marketplace/actions/github-codeowners-validator

This looks like it'd keep the manual additions in check. I think there's going to be too many one off rules to maintain it automatically.

Emiller88:
This might get us started though https://github.com/marketplace/actions/author-to-codeowners

grst:
When we have a Github Action updating codeowners, are we better off than having a github action that reads the meta.yml and directly assigns reviewers? 

@drpatelh additionally had the idea to randomly assign reviewers to PRs, which would be particularly useful for new modules. 
 * [slack discussion](https://nfcore.slack.com/archives/CJRH30T6V/p1631092043069300)
 * https://github.com/marketplace/actions/reviewer-lottery

Emiller88:
I think this needs to be autogenerated at this point with `nf-core tools`

matthdsm:
Hi there!

Cheers
the nf-core maintainers

Emiller88:
Hot take: What if we move this to tools? Since I think it's never going to be maintainable without tools.

matthdsm:
I have no stakes in this, just the part of the cleaning crew 🤷🏻 
@maxulysse , @drpatelh ?

maxulysse:
Love @Emiller88 's idea

matthdsm:
> Hot take: What if we move this to tools? Since I think it's never going to be maintainable without tools.

Hi @Emiller88, if you want to move this, can you do so?

muffato:
Supersed by #4161

PR #92, opened 2020-12-09
Update Salmon modules

drpatelh:


drpatelh:
Need these modules for the rnaseq pipeline so will merge. Sorry....tested and working there. 🎄 is getting closer by the day and we need to get another release out ASAP!

PR #110, opened 2020-12-17
Add SEACR module containing a single CALLPEAK process

chris-cheshire:
I have added the primary peak caller for CUT&RUN/TAG. 

I was unsure on how to do the parameterization when all parameters are required. The software needs `["norm" | "non"] ["relaxed" | "stringent"]` as two fields; therefore, the `opt.args` current requires a certain format for the process to run without error.

I am sure there is lots wrong with this PR, I just wanted to get it out there to get some feedback on the correct way to do these modules!

I also have not included at `.yml` file yet as some modules don't seem to have them and I remember a conversation about ditching them some time ago.


software/seacr/callpeak/main.nf
```diff
@@ -0,0 +1,41 @@
...
+
+process SEACR_CALLPEAK {
+    tag "$meta.id"
+    label 'process_medium'
+    publishDir "${params.outdir}",
+        mode: params.publish_dir_mode,
+        saveAs: { filename -> saveFiles(filename:filename, options:params.options, publish_dir:getSoftwareName(task.process), publish_id:meta.id) }
+
+    conda (params.enable_conda ? "bioconda::seacr=1.3" : null)
+
```
drpatelh, line 15:
```suggestion
```

software/seacr/callpeak/main.nf
```diff
@@ -0,0 +1,41 @@
...
+    if (workflow.containerEngine == 'singularity' && !params.singularity_pull_docker_container) {
+        container "https://depot.galaxyproject.org/singularity/mulled-v2-03bfeb32fe80910c231f630d4262b83677c8c0f4:5bb5ed4307a8187a7f34730b00431de93688fa59-0"
```
drpatelh, line 17: Why do we need a multi-tool container here? If we need it for other packages then can we list those in the `conda` scope too?

software/seacr/callpeak/main.nf
```diff
@@ -0,0 +1,41 @@
...
+    
+    output:
+    tuple val(meta), path("*.bed"), emit: bed
+    path "*.version.txt"          , emit: version
+
+    script:
+    def software = getSoftwareName(task.process)
+    def prefix   = options.suffix ? "${meta.id}${options.suffix}" : "${meta.id}"
+    """
+     SEACR_1.3.sh \\
```
drpatelh, line 33:
```suggestion
    SEACR_1.3.sh \\
```

software/seacr/callpeak/main.nf
```diff
@@ -0,0 +1,41 @@
...
+        $bedgraph \\
+        $ctrlbedgraph \\
+        $options.args \\
+        $prefix
+
+    echo 1.3 > ${software}.version.txt
```
drpatelh, line 39: You can't get the version number from the command-line? If not, can we add the version like [here](https://github.com/nf-core/modules/blob/0305d873f4ee9329338863033e5d2df57c95bae9/software/homer/annotatepeaks/main.nf#L7)? Easier not to miss when the tool version is updated.

drpatelh (changes requested):
Looks amazing @chris-cheshire 😍 Few comments / adjustments. Still need to add docs in `meta.yml` - sorry couldn't comment there because it was an empty file.

And there appears to be a Submodule introduced in the repo?

image

software/seacr/callpeak/main.nf
chris-cheshire: Yes its because it requires R and bedtools, added to conda as requested

software/seacr/callpeak/main.nf
chris-cheshire: No I am afraid not (very annoying) - changed as requested

chris-cheshire:
> Looks amazing @chris-cheshire 😍 Few comments / adjustments. Still need to add docs in `meta.yml` - sorry couldn't comment there because it was an empty file.
> 
> And there appears to be a Submodule introduced in the repo?
> 
> image

I don't know why this is here - its not a commit I made and I have only been merging from nf-core/master?!

drpatelh:
Maybe its easier to just re-create this PR @chris-cheshire without the `test-datasets` submodule?

chris-cheshire:
> Maybe its easier to just re-create this PR @chris-cheshire without the `test-datasets` submodule?

So there should be no test-datasets sub-module in this repo?

drpatelh:
Yup, exactly. If you are able to find a way to remove it that would be another solution. I don't I'm afaird.

drpatelh (approved):

drpatelh:
Thanks @chris-cheshire !

PR #111, opened 2020-12-22
Add testing for Phantompeakqualtools 

sruthipsuresh:
I've updated the testing (workflow/main.nf/test.yml) for the Phantompeakqualtools module to follow the pytest-workflow format - this addresses #98.
Please let me know if there is anything which needs to be fixed!

### DIFF
--- a/.github/workflows/phantompeakqualtools.yml
+++ b/.github/workflows/phantompeakqualtools.yml
@@ -1,0 +1,40 @@
+name: phantompeakqualtools
+on:
+  push:
+    paths:
+      - software/phantompeakqualtools/**
+      - .github/workflows/phantompeakqualtools.yml
+      - tests/software/phantompeakqualtools/**
+  pull_request:
+    paths:
+      - software/phantompeakqualtools/**
+      - .github/workflows/phantompeakqualtools.yml
+      - tests/software/phantompeakqualtools/**
+
+jobs:
+  ci_test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        nxf_version: [20.07.1, ""]
+    env:
+      NXF_ANSI_LOG: false
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install Nextflow
+        env:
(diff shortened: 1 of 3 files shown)


KevinMenden:
Couldn't spot any errors, and since the test passes as well, this looks pretty good!

Only thing I was wondering is why you removed the md5 sums for pdf and rdata? Are those files not reproducible?

KevinMenden (commented):
Couldn't spot any errors, and since the test passes as well, this looks pretty good!

Only thing I was wondering is why you removed the md5 sums for pdf and rdata? Are those files not reproducible?

Emiller88:
@KevinMenden The md5 sums change for pdfs and Rdata, my guess is because they may have timestamp in them. We may try to use the contains option in the future to get the important parts out of it.

KevinMenden:
I see, thanks for the explanation @Emiller88  :)

KevinMenden (approved):
Looks good to me and all checks pass - so I say go ahead an merge :) 

👍

Emiller88:
@KevinMenden Thanks! Our thought process on these to just get the tests started, and if anyone finds that the output from the tests isn't what's expected then they can open an issue.

.github/workflows/phantompeakqualtools.yml
```diff
@@ -0,0 +1,40 @@
...
+      - software/phantompeakqualtools/**
+      - .github/workflows/phantompeakqualtools.yml
+      - tests/software/phantompeakqualtools/**
+
+jobs:
+  ci_test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        nxf_version: [20.07.1, ""]
```
drpatelh, line 19:
```suggestion
        nxf_version: [20.11.0-edge]
```

drpatelh (approved):
Thanks guys!!

.github/workflows/phantompeakqualtools.yml
Emiller88: Good catch thanks!

Emiller88 (approved):

PR #112, opened 2021-01-14
Preseq testing

### LINKED ISSUE #99: Add tests for preseq

sruthipsuresh:
Added testing for preseq, both paired and single end. I also added a -bam argument to the original module to ensure that the input is run as a bam file by the preseq tool.
Closes #99 
@Emiller88


software/preseq/lcextrap/main.nf
```diff
@@ -37,7 +37,7 @@ process PRESEQ_LCEXTRAP {
         $options.args \\
         $paired_end \\
         -output ${prefix}.ccurve.txt \\
-        $bam
+        -bam $bam
```
drpatelh, line 40:
```suggestion
        $bam
```

drpatelh: We should keep this as it was before. The `-bam` parameter is actually coming from [here](https://github.com/nf-core/rnaseq/blob/3643a94411b65f42bce5357c5015603099556ad9/conf/modules.config#L118) and, hopefully this will allows developers to trick the module into accepting other sorts of input files without updating the module.

drpatelh (approved):
LGTM! Minor change.

Emiller88 (approved):

PR #119, opened 2021-01-22
ci: Upload logs on failure

Emiller88:
fix #119

PR #120, opened 2021-01-24
Universal CI job

### LINKED ISSUE #83: Run a test on all modules which have been modified in that push / PR, all in parallel in separate jobs

Emiller88:
Fixes #83 #119 

Things on my wishlist still:
- [x] cache the pip stuff
- [x] ~~cache nextflow binary~~ #136 
- [x] ~~Move the composite action repo to nf-core~~ Just gonna spell it all out.
- [ ] Fix the naming of jobs and steps (I just kinda threw them together so thoughts on that would be appreciated.)


Emiller88:
Debating pulling it all out the composite action for a little more control over the caching

ewels:
_Pasted [from Slack](https://nfcore.slack.com/archives/CJRH30T6V/p1611512318074500):_

Beautiful! :tada: I was imagining some kind of path filter magic where we got the tag name based on the directories that changed, but the yaml file is maybe a little more elegant. _Slightly_ more work on the setup but gives more control and likely less likely to break :+1:

It might be nice to add a check that all modules are mentioned in the yaml file..? I’m obsessed about linting everything though, so feel free to ignore me :joy:

What does the composite thing do?

Personally I think I’d rather just keep all of the steps in the one workflow file rather than having the separate action.. Less hunting around for the code to figure out what’s going on then. And it’s only 4 steps so it’s not going to bloat the workflow too much. Also this is only ever going to be in modules and not synced to pipelines so I’m not sure that there is an advantage in having one point of editing for all pipelines etc.

Sorry, just thinking out loud. Cool to see how you’ve built the action! Not sure that I’ve seen one like this before which isn’t all JavaScript and stuff.

Emiller88:
> It might be nice to add a check that all modules are mentioned in the yaml file..? I’m obsessed about linting everything though, so feel free to ignore me joy

I think that's definitely possible, and a good way to handle it. I love the automated dream, but I think it's more likely to break. Having something automated to make sure we're not missing

> What does the composite thing do?

It's basically just the simplest way to package up some steps. Takes your inputs, and spits out your outputs. 

> Personally I think I’d rather just keep all of the steps in the one workflow file rather than having the separate action.. Less hunting around for the code to figure out what’s going on then. And it’s only 4 steps so it’s not going to bloat the workflow too much. Also this is only ever going to be in modules and not synced to pipelines so I’m not sure that there is an advantage in having one point of editing for all pipelines etc.

That's kinda what I thought, but again I'd like to add pytest-workflow to all the nf-core stuff and making adjustments to the composite step would make it easier to role out.

The real problem with it is that you're not able to use `uses` in them, so you can't setup python the official way, etc.

drpatelh (approved):
Let's do this!!

PR #121, opened 2021-01-25
Adding test for STAR

KevinMenden:
Adding tests and `meta.yml` files for `star align` and `star genomegenerate` processes #103 
Some things to note:

### test data
I've added new test data, because no GTF was available and I couldn't find a GTF that works with the available genomes and reads. The data I added is unfortunately a bit on the bigger side (~10MB all in all I think). Am happy to use different datasets if you can point me to some.
Unfortunately, with the current set of data, STAR also doesn't align anything. I don't think this makes the tests any less valid but it would of course be nice to have a few aligned reads.

### star process
As discussed in Slack, I adapted the `star_align` process to emit BAM files by default. I had to do a few more adjustments:

To add BAM output as default but keep this still as an option, I've added the following code:
`def out_sam_type = (options.args.contains('--outSAMtype')) ? '' : '--outSAMtype BAM Unsorted'`
This variable is then just added to the STAR call.

Next, to capture only unsorted/sorted BAMs as the main `bam` output, I've changed the emit line to:
`tuple val(meta), path('*d.out.bam')        , emit: bam`
This captures `Aligned.out.bam` and `Aligned.sortedByCoord.out.bam`  but not `toTranscriptome.out.bam`

Finally, it is possible to get 3 different BAM files as output from STAR ... to capture this, I've added:
`def mv_unsorted_bam = (options.args.contains('--outSAMtype BAM Unsorted SortedByCoordinate')) ? "mv ${prefix}.Aligned.out.bam ${prefix}.Aligned.unsort.out.bam" : ''`

This command is just added to the actual script. So in the likely-never-to-happen-case that a user wants sorted and unsorted BAM as output, the unsorted BAM is renamed and only the sorted BAM file is emitted to the `bam` channel. This avoids having the process emitting two files to this channel, which would certainly cause problems.

This way, we don't have to add any extra input parameter to the process, which I think is better for usability.

Oh and I had to remove the STAR log files from the test files output because the include dates.

### DIFF
--- a/software/star/align/main.nf
+++ b/software/star/align/main.nf
@@ -12,13 +12,12 @@ process STAR_ALIGN {
         saveAs: { filename -> saveFiles(filename:filename, options:params.options, publish_dir:getSoftwareName(task.process), publish_id:meta.id) }
 
     // Note: 2.7X indices incompatible with AWS iGenomes.
-    conda (params.enable_conda ? "bioconda::star=2.6.1d" : null)
+    conda (params.enable_conda ? 'bioconda::star=2.6.1d' : null)
     if (workflow.containerEngine == 'singularity' && !params.singularity_pull_docker_container) {
-        container "https://depot.galaxyproject.org/singularity/star:2.6.1d--0"
+        container 'https://depot.galaxyproject.org/singularity/star:2.6.1d--0'
     } else {
-        container "quay.io/biocontainers/star:2.6.1d--0"
+        container 'quay.io/biocontainers/star:2.6.1d--0'
     }
-
 
     input:
     tuple val(meta), path(reads)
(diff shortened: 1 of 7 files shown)


software/star/align/main.nf
```diff
@@ -12,46 +12,51 @@ process STAR_ALIGN {
...
     path  index
     path  gtf
 
     output:
-    tuple val(meta), path("*Aligned.out.bam") , emit: bam
-    tuple val(meta), path("*Log.final.out")   , emit: log_final
-    tuple val(meta), path("*Log.out")         , emit: log_out
-    tuple val(meta), path("*Log.progress.out"), emit: log_progress
-    path  "*.version.txt"                     , emit: version
+    tuple val(meta), path('*d.out.bam')        , emit: bam
```
drpatelh, line 28:
```suggestion
    tuple val(meta), path('*d.out.bam')       , emit: bam
```

software/star/align/main.nf
```diff
@@ -12,46 +12,51 @@ process STAR_ALIGN {
...
+    tuple val(meta), path('*Log.progress.out'), emit: log_progress
+    path  '*.version.txt'                     , emit: version
 
-    tuple val(meta), path("*sortedByCoord.out.bam")  , optional:true, emit: bam_sorted
-    tuple val(meta), path("*toTranscriptome.out.bam"), optional:true, emit: bam_transcript
-    tuple val(meta), path("*fastq.gz")               , optional:true, emit: fastq
-    tuple val(meta), path("*.tab")                   , optional:true, emit: tab
+    tuple val(meta), path('*sortedByCoord.out.bam')  , optional:true, emit: bam_sorted
+    tuple val(meta), path('*toTranscriptome.out.bam'), optional:true, emit: bam_transcript
+    tuple val(meta), path('*Aligned.unsort.out.bam'), optional:true, emit: bam_unsorted
```
drpatelh, line 36:
```suggestion
    tuple val(meta), path('*Aligned.unsort.out.bam') , optional:true, emit: bam_unsorted
```

software/star/align/main.nf
```diff
@@ -12,46 +12,51 @@ process STAR_ALIGN {
...
         --genomeDir $index \\
         --readFilesIn $reads  \\
         --runThreadN $task.cpus \\
         --outFileNamePrefix $prefix. \\
+        $out_sam_type \\
         $ignore_gtf \\
         $seq_center \\
         $options.args
 
+    ${mv_unsorted_bam}
```
drpatelh, line 58:
```suggestion
    $mv_unsorted_bam
```

software/star/align/meta.yml
```diff
@@ -0,0 +1,77 @@
...
+      pattern: "*Log.out"
+  - log_progress:
+      type: file
+      description: STAR log progress file
+      pattern: "*Log.progress.out"
+  - version:
+      type: file
+      description: File containing software version
+      pattern: "*.{version.txt}"
+
```
drpatelh, line 74: Need to add the optional outputs here too.

drpatelh (changes requested):
This looks great @KevinMenden . Thank you! Minor OCD corrections.

software/star/align/meta.yml
KevinMenden: Alright :)

drpatelh (approved):

PR #123, opened 2021-01-29
Added cutadapt module

KevinMenden:
This PR adds the cutadapt module, modified from https://github.com/drpatelh/nf-core-viralrecon/blob/dsl2/modules/local/cutadapt.nf

### cutadapt module
I have adapted the module to be more flexible, i.e. I have removed 
`def paired   = meta.single_end ? "-a file:adapters.sub.fa"   : "-a file:adapters.sub.fa -A file:adapters.sub.fa"`

and thus the possibility of handing over a channel with adapters to the process. Cutadapt is highly flexible and can do all kinds of different trimming options. The code above, for instance, only allows to trim the same 3'-adapters from both mates. So in this version, the user has to specify the adapter files in the `modules.config`

I think this is reasonable as the adapters shouldn't change for different samples in the same pipeline run and it goes along the nf-core/modules guidelines to have only input and output specified.

### cutadapt test
I have added tests - however no md5 sums. They change for the log files (speed information etc.) but surprisingly also for the fastq.gz output files. I have no idea why, I tried running on a single core but that wasn't the issue. Happy to hear some ideas on that one :-)

### DIFF
--- a/.github/workflows/cutadapt.yml
+++ b/.github/workflows/cutadapt.yml
@@ -1,0 +1,40 @@
+name: cutadapt
+on:
+  push:
+    paths:
+      - software/cutadapt/**
+      - .github/workflows/cutadapt.yml
+      - tests/software/cutadapt/**
+  pull_request:
+    paths:
+      - software/cutadapt/**
+      - .github/workflows/cutadapt.yml
+      - tests/software/cutadapt/**
+
+jobs:
+  ci_test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        nxf_version: [20.11.0-edge]
+    env:
+      NXF_ANSI_LOG: false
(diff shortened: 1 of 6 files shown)


drpatelh (approved):
LGTM! If you are happy with the module then feel free to merge :)

As mentioned on Slack, we won't be able to use that module without having an explicit input for the adapters fasta. We can't provide it as `options.args` because the file itself still needs to be staged by Nextflow and the support for optional inputs isn't quite there yet as far as I know. There is support for optional outputs.

KevinMenden:
Optional input is certainly climbing up my nextflow wishlist ...

Thanks for the review! As discussed I will let think about this one a bit and might make some changes to better accommodate all cutadapt usecases .... 🤔

PR #130, opened 2021-02-01
Adding a QUAST module

KevinMenden:
This PR adds a QUAST module.

To accommodate the two usecases of running quast with and without a reference genome, I ended up creating two modules, or "tools" for QUAST. I tried out a few things but as of now there's really no good way of dealing with optional inputs in DSL2. Until that's possibly, I think this is an adequate solution.

Let me know if you find any bugs or anything I should change :)


drpatelh:
Thanks @KevinMenden . Ok, I think I have a solution that allows us to keep this repo clean from duplicating modules to deal with optional inputs. It's not ideal but possible the best solution that allows us developers to use their own parameter names e.g. instead of `params.gff`, `params.fasta` etc.


```
    input:
    path consensus
    path fasta
    path gff
    val use_fasta
    val use_gff
```

If we add two more boolean variables to the `input` then these can be evaluated in the `script` section to stage the appropriate file. This means we don't have to use `params.gff` anymore and keeps the module as clean as possible in terms of parameter names. Also, we get to keep a single instance of the module which I think is best given that this issue is really Nextflow specific. If/when optional inputs are supported we update the module. Simples.

What do you think?

KevinMenden:
Okay yes, it's the best we can do right now I reckon :/ 
I agree this is nicer than having two different 'modules' for quast.

The only thing I'm a bit worried is how the `fasta` and `gtf` paths are evaluated, we probably need to pass some dummy values to them that look look like paths (basically just `""`). I will test around a bit.

Thanks @drpatelh ! Will ping you when I've changed it.

drpatelh:
Cool. Yup, that is exactly what I do if `params.gff` hasn't been provided. See [here](https://github.com/drpatelh/nf-core-viralrecon/blob/c4747d1691977e02ae90766ad9257d34ae1546fe/subworkflows/local/prepare_genome.nf#L49). This ensures that a file is always staged regardless of whether it needs to be used.

KevinMenden:
@drpatelh okay everything refactored now :)

software/quast/main.nf
```diff
@@ -0,0 +1,49 @@
...
+    } else {
+        container 'quay.io/biocontainers/quast:5.0.2--py37pl526hb5aa323_2'
+    }
+
+    input:
+    path consensus
+    path fasta
+    val use_fasta
+    path gff
+    val use_gff
```
drpatelh, line 26:
```suggestion
    path gff
    val use_fasta
    val use_gff
```

software/quast/main.nf
```diff
@@ -0,0 +1,49 @@
...
+    script:
+    def software  = getSoftwareName(task.process)
+    prefix        = options.suffix ?: software
+    def features  = use_gff ? "--features $gff" : ''
+    def reference = use_fasta ? "-r $fasta" : ''
+    """
+    quast.py \\
+        --output-dir $prefix \\
+        ${reference} \\
+        ${features} \\
```
drpatelh, line 42:
```suggestion
        $reference \\
        $features \\
```

software/quast/main.nf
```diff
@@ -0,0 +1,49 @@
+// Import generic module functions
+include { initOptions; saveFiles; getSoftwareName } from './functions'
+
+params.options = [:]
+def options    = initOptions(params.options)
+
+process QUAST {
+    tag "$fasta"
```
drpatelh, line 8:
```suggestion
```

software/quast/meta.yml
```diff
@@ -0,0 +1,68 @@
...
+        The genome assembly to be evaluated. Has to contain at least a non-empty string dummy value.
+  - use_fasta:
+      type: boolean
+      description: Whether to use a fasta reference
+  - gff:
+      type: file
+      description: The genome GFF file. Has to contain at least a non-empty string dummy value.
+  - use_gff:
+      type: boolean
+      description: Whether to use a gff reference
```
drpatelh, line 52:
```suggestion
      description: Whether to use the provided gff reference annotation file
```

software/quast/meta.yml
```diff
@@ -0,0 +1,68 @@
...
+      type: file
+      description: |
+        Fasta file containing the assembly of interest
+  - fasta:
+      type: file
+      description: |
+        The genome assembly to be evaluated. Has to contain at least a non-empty string dummy value.
+  - use_fasta:
+      type: boolean
+      description: Whether to use a fasta reference
```
drpatelh, line 46:
```suggestion
      description: Whether to use the provided fasta reference genome file
```

tests/software/quast/main.nf
```diff
@@ -0,0 +1,25 @@
...
+include { QUAST }   from '../../../software/quast/main.nf'   addParams(options: [:])
+
+workflow test_quast_ref {
+    consensus = file("${launchDir}/tests/data/fasta/sarscov2/GCA_011545545.1_ASM1154554v1_cds_from_genomic.fna", checkIfExists: true)
+    gff = file("${launchDir}/tests/data/gff/sarscov2/GCA_011545545.1_ASM1154554v1_genomic.gtf", checkIfExists: true)
+    fasta = file("${launchDir}/tests/data/fasta/sarscov2/GCA_011545545.1_ASM1154554v1_genomic.fna", checkIfExists: true)
+    def use_fasta = true
+    def use_gtf = true
+
+    QUAST( consensus, fasta, use_fasta, gff, use_gtf)
```
drpatelh, line 14:
```suggestion
    QUAST( consensus, fasta, gff, use_fasta, use_gtf )
```

tests/software/quast/main.nf
```diff
@@ -0,0 +1,25 @@
...
+}
+
+workflow test_quast_noref {
+    consensus = file("${launchDir}/tests/data/fasta/sarscov2/GCA_011545545.1_ASM1154554v1_cds_from_genomic.fna", checkIfExists: true)
+    gff = file('gff_dummy')
+    fasta = file('fasta_dummy')
+    def use_fasta = false
+    def use_gtf = false
+
+    QUAST( consensus, fasta, use_fasta, gff, use_gtf)
```
drpatelh, line 24:
```suggestion
    QUAST( consensus, fasta, gff, use_fasta, use_gtf )
```

drpatelh (changes requested):
Amazing! Thanks @KevinMenden . Few minor changes. I find it more intuitive grouping by variable type than by name.

tests/software/quast/main.nf
KevinMenden: you know I'll never get those spaces right :)

KevinMenden:
Alright thanks for the review!

drpatelh (approved):

PR #133, opened 2021-02-02
Update bwa/index container

MaxUlysse:
Use container from quay.io.

This `biocontainers/bwa:v0.7.17_cv1` is not working for me.

### DIFF
--- a/software/bwa/index/main.nf
+++ b/software/bwa/index/main.nf
@@ -15,7 +15,7 @@ process BWA_INDEX {
     if (workflow.containerEngine == 'singularity' && !params.singularity_pull_docker_container) {
         container "https://depot.galaxyproject.org/singularity/bwa:0.7.17--hed695b0_7"
     } else {
-        container "biocontainers/bwa:v0.7.17_cv1"
+        container "quay.io/biocontainers/bwa:0.7.17--hed695b0_7"
     }
 
     input:


KevinMenden (approved):
Make sense 👍

drpatelh:
This module needs to be updated anyway along with the `bwa_mem` one too. It would be better to create and move directories around for indices instead of using globs. I can update in a separate PR but the idea would be make and create a directory with all of the index files like [here](https://github.com/nf-core/modules/blob/4301afb7711bb10851753a6b0fcec9a6373769dd/software/bowtie2/build/main.nf#L31) and then stage them in the `bwa_mem` process like [this](https://github.com/nf-core/modules/blob/4301afb7711bb10851753a6b0fcec9a6373769dd/software/bowtie2/align/main.nf#L37). This also means that pipelines are able to accept tar archives of the indexing directory if required.

MaxUlysse:
The `bwa_mem` has a mulled container, but the conda version for bwa in the conda declaration is the same as specified in this one.

MaxUlysse:
@drpatelh will you be ok, if I don't update the `bwa_mem`, as I'm not sure what to update there? or if there is even something to update...

drpatelh:
Ok. Let me get a PR together now and pull you in to review 👍🏽

PR #135, opened 2021-02-02
Update indices syntax for BWA index and mem modules

drpatelh:


software/bwa/mem/main.nf
```diff
@@ -32,13 +31,15 @@ process BWA_MEM {
     def prefix     = options.suffix ? "${meta.id}${options.suffix}" : "${meta.id}"
     def read_group = meta.read_group ? "-R ${meta.read_group}" : ""
     """
+    INDEX=`find -L ./ -name "*.amb" | sed 's/.amb//'`
```
MaxUlysse, line 34: I still don't think this is the right way to do it...
And I think you're missing some indices as well

drpatelh: This is just getting the genome prefix automatically. You don't need to list all of the files in the index for that. Just one.

MaxUlysse: why don't you keep the fasta as an input?

drpatelh: This also means you don't need to stage the `fasta` file in the module too and pipelines can accept tar archives / untarred directores containing the indices. You don't even need to know what all of those indices file extensions are as long as the directory contains the index. Much cleaner.

MaxUlysse: yes, but if you don't have the fasta as an input, how will it work?
Also don't you need the fasta index as well?
or do you include that in the index path?

software/bwa/mem/main.nf
```diff
@@ -32,13 +31,15 @@ process BWA_MEM {
...
+
     bwa mem \\
         $options.args \\
         $read_group \\
         -t $task.cpus \\
-        $fasta \\
+        \$INDEX \\
         $reads \\
-        | samtools view $options.args2 -@ $task.cpus -bS -o ${prefix}.bam -
+        | samtools view $options.args2 -@ $task.cpus -bhS -o ${prefix}.bam -
```
MaxUlysse, line 42: I think it should be
`samtools $options.args2 --threads $task.cpus -o ${prefix}.bam -`

the `view -bhS` should be part of $options.args2.

software/bwa/mem/main.nf
drpatelh: I think the easiest option is to try that command in a directory containing a BWA index and see what it returns 😉

software/bwa/mem/main.nf
drpatelh:
```
  -b       output BAM
  -h       include header in SAM output
  -S       ignored (input format is auto-detected)
      --input-fmt-option OPT[=VAL]
               Specify a single input file format option in the form
               of OPTION or OPTION=VALUE
```

Those options are related to the input/output files so should be in the module. What this doesn't cater for are things like CRAM format but it's going to be another level of complexity adding that in here. Maybe save that for another day 😅

drpatelh:
Ping @MaxUlysse.

MaxUlysse (approved):
ok for now

drpatelh:
Thanks!

PR #142, opened 2021-02-03
Split tests

drpatelh:
This PR splits out all of the unit test we currently have for nf-core/modules such that each sub-tool for a particular software e.g. `samtools flagstat` / `samtools idxstats` now have their own tests. This is beneficial for a number of reasons including more fine-grained control, CI and reporting for only the relevant tools, especially when creating PRs to update them.


Emiller88 (approved):
Looks good I'll fix the failing tests in a follow up PR

PR #143, opened 2021-02-03
Get Singularity CI tests working

Emiller88:
Should fix issues with singularity builds #145 and "fixes" the `bowtie*_build` tags not running.

### DIFF
--- a/tests/software/bowtie/align/test.yml
+++ b/tests/software/bowtie/align/test.yml
@@ -42,6 +42,6 @@
     - path: output/bowtie/bowtie/NC_010473.rev.2.ebwt
       md5sum: f3c398bba5158f4039334a932d79c051
     - path: output/bowtie/test.out
-      md5sum: a23e9a2a76e949aeb3693bcfae41a615
+      md5sum: a81cb18024616415a6cec3108a36fccd
     - path: output/bowtie/test.bam
       should_exist: true
--- a/tests/software/bowtie/align/test.yml
+++ b/tests/software/bowtie/align/test.yml
@@ -44,4 +44,3 @@
     - path: output/bowtie/test.out
       md5sum: a81cb18024616415a6cec3108a36fccd
     - path: output/bowtie/test.bam
-      should_exist: true
--- a/tests/software/bowtie/build/test.yml
+++ b/tests/software/bowtie/build/test.yml
@@ -1,8 +1,8 @@
-- name: bowtie build
+- name: bowtie build test
   command: nextflow run ./tests/software/bowtie/build -entry test_bowtie_build -c tests/config/nextflow.config
   tags:
     - bowtie
-    - bowtie_buil
+    - bowtie_build
   files:
     - path: output/bowtie/bowtie/NC_010473.1.ebwt
       md5sum: 90f0b7aa5bbaeaaa999839ac13ad203c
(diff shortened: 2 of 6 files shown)


tests/software/bowtie/align/test.yml
```diff
@@ -42,6 +42,6 @@
     - path: output/bowtie/bowtie/NC_010473.rev.2.ebwt
       md5sum: f3c398bba5158f4039334a932d79c051
     - path: output/bowtie/test.out
-      md5sum: a23e9a2a76e949aeb3693bcfae41a615
+      md5sum: a81cb18024616415a6cec3108a36fccd
```
KevinMenden, line 45: Any idea why this changed?

tests/software/bowtie/build/test.yml
```diff
@@ -1,4 +1,4 @@
-- name: bowtie build
+- name: bowtie build test
```
KevinMenden, line 1: I thought we want to keep this minimal (without obvious 'test')

tests/software/bowtie2/build/test.yml
```diff
@@ -1,4 +1,4 @@
-- name: bowtie2 build
+- name: bowtie2 build test
```
KevinMenden, line 1: same as above.

or is there some weird test-reason why we need this?

KevinMenden (commented):
Looks good just a few questions basically.

The singularity version necessary I guess?

tests/software/bowtie2/build/test.yml
drpatelh, line 1:
```suggestion
- name: bowtie2 build
```

tests/software/bowtie/build/test.yml
drpatelh, line 1:
```suggestion
- name: bowtie build
```

drpatelh: Good spot 👍🏽

tests/software/bowtie/align/test.yml
```diff
@@ -42,6 +42,6 @@
...
     - path: output/bowtie/test.bam
       should_exist: true
```
drpatelh, line 47:
```suggestion
```

tests/software/bowtie/align/test.yml
drpatelh:
```
# reads processed: 10000
# reads with at least one alignment: 2 (0.02%)
# reads that failed to align: 9998 (99.98%)
Reported 2 alignments
```

This is the latest version which is only showing 2 alignments...are you able to run `pytest` via the old tests so we can see the difference?

drpatelh (approved):
LGTM! I think we should create a separate issue for why we had to change the `md5sum` for `bowtie align` and come back to that so we can get this in to start figuring out how to deal with Conda!

tests/software/bowtie/align/test.yml
drpatelh: The old version generated the output below which is why the tests are failing:

```
# reads processed: 10000
# reads with at least one alignment: 1 (0.01%)
# reads that failed to align: 9999 (99.99%)
Reported 1 paired-end alignments
```

I guess this highlights the importance of using seed values where possible in `main.nf` for the tests to always generate the same alignments and not by random. At least, this is what I think is happening here.

drpatelh:
💥

PR #144, opened 2021-02-03
Start from scratch

drpatelh:


drpatelh:
Tweeks found here added in https://github.com/nf-core/modules/pull/143

PR #148, opened 2021-02-04
ci: Fix artifact path

Emiller88:
Had to change tmp dir to ~ for singularity but didn't update where the log upload was looking for logs


KevinMenden (approved):
Ah perfect :)

PR #79, opened 2020-11-13
Bring bedtools up to speed

Emiller88:
Hello all! @sruthipsuresh is an undergraduate in our lab, and she's been helping
me rewrite a GRO-Seq pipeline using nf-core. We've been waiting for the DSL2
template to get wrapped up, so to get familar with the nf-core modules we've
been working on writing some modules that we know we'll need. A few things we need some opinions on definately:

- Review slop params and how best to handle those(in the "nf-core" way)
- How bed to handle the logic of which params in `bedtools intersect` between 2
  or 3 files. Should the user just input their string in `options.args`?
- `bedtools merge` requires that the files be sorted ahead of time. Should we
  sort before merge? Should we just check that it's sorted somehow? Just specify
  in the docs maybe?
    - Maybe some end-to-end testing then? To ensure for example the sort process
      can be fed into the merge process during any future refactoring or version
      bumps.

There's a few cleanups left to do but I think this is ready for feedback!
----

#

PR #151, opened 2021-02-07
Follow up fixes for bedtools modules added in #79

drpatelh:
Couldn't request you for review @sruthipsuresh but I have just sent you an invite to become a member of the nf-core Github org which should fix it.

The only thing I am uncertain of really is that these modules don't necessarily have to be used with respect to a particular sample i.e. all of the input channels have `meta` information coming in as well as the input BED / BAM files which won't always be the required use case. However, I supposed it's easier to include dummy sample information than to reconfigure channels to add it beforehand....so let's go with this and expect more follow-up PRs / questions from the community!

Note: I have also activate the CI Conda tests and they are passing! 🎉 Hopefully, this will help us isolate and figure out why some tools are failing.

Follow up to https://github.com/nf-core/modules/pull/79


sruthipsuresh:
Thank you so much! I accepted the invite. I wasn't sure of that either - I followed the meta.id format used in the Samtools module to figure out what the use case could be. I can change that to better follow the multiqc model format/deprecated bedtools to not take the meta.id if you prefer?
Thank you again for activating and fixing the Conda test! I really appreciate it!

drpatelh:
No worries. I think it's ok as it is for now. Developers can come and update the modules via a PR if need be. Be great if you or @Emiller88 can review the changes here so I can merge the PR :)

sruthipsuresh (approved):
It looks good to me! Is there a reason why complement specifically takes .suffix as a parameter?

sruthipsuresh:
Sounds good! I went ahead and reviewed the changes!

drpatelh:
Great spot! Thank you :)

.github/workflows/pytest-workflow.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       matrix:
         nxf_version: ['20.11.0-edge']
         tags: ['${{ fromJson(needs.changes.outputs.modules) }}']
-        profile: ['docker', 'singularity'] ## 'conda'
+        profile: ['docker', 'singularity', 'conda']
```
Emiller88, line 28: Is this going to break everything else once it's in master though?

Emiller88 (approved):
LGTM just make sure that enabling conda is intended!

.github/workflows/pytest-workflow.yml
drpatelh: Shouldn't do. Some tests may fail for incoming PRs but we can just approve and merge them anyway. I am hoping this will twist our arm and help us find a permanent fix.

PR #152, opened 2021-02-08
Update template module and add more docs for testing and finding containers

drpatelh:


README.md
```diff
@@ -119,21 +120,21 @@ for examples.
...
+│   └── TOOL
+│       └── SUBTOOL
+│           ├── functions.nf    ## Utility functions imported in main module script
+│           ├── main.nf         ## Main module script
+│           └── meta.yml        ## Documentation for module, input, output, params, author
+├── tests
+│   └── TOOL
+│       └── SUBTOOL
+│           ├── main.nf         ## Minimal workflow to test module
+│           └── test.yml        ## Pytest-workflow test file
```
KevinMenden, line 137: Currently we have `tests/software/tool` and also the `data` and `config` folders in the `tests`directory.
Are you planning to switch to this structure then? With `data` and `config` staying under the `tests` folder?

Just asking because of the linter :) although that will be quick to adapt to these changes.

README.md
```diff
@@ -213,51 +231,75 @@ using a combination of `bwa` and `samtools` to output a BAM file instead of a SA
...
+
+        ```console
+        cd /path/to/git/clone/of/nf-core/modules/
+        PROFILE=conda pytest --tag tests/software/bowtie/build --symlink --wt 2 --keep-workflow-wd
+        ```
 
 ### Documentation
 
-- A module MUST be documented in the [`meta.yml`](software/fastqc/meta.yml) file. It MUST document `params`, `input` and `output`. `input` and `output` MUST be a nested list.
+- A module MUST be documented in the [`meta.yml`](software/TOOL/SUBTOOL/meta.yml) file. It MUST document `params`, `input` and `output`. `input` and `output` MUST be a nested list.
```
KevinMenden, line 296: `name` is optional? (again asking for my friend the linter)

software/TOOL/SUBTOOL/meta.yml
```diff
@@ -1,25 +1,24 @@
...
-            You get information about adapter contamination and other
-            overrepresented sequences.
-        homepage: https://www.bioinformatics.babraham.ac.uk/projects/fastqc/
-        documentation: https://www.bioinformatics.babraham.ac.uk/projects/fastqc/Help/
+            SAMtools is a set of utilities for interacting with and post-processing
+            short DNA sequence read alignments in the SAM, BAM and CRAM formats, written by Heng Li.
+            These files are generated as output by short read aligners like BWA.
+        homepage: http://www.htslib.org/
+        documentation: hhttp://www.htslib.org/doc/samtools.html
+        doi: 10.1093/bioinformatics/btp352
```
KevinMenden, line 21: Personally I think we can omit the `homepage` because most tools don't have any, and if the do, it's the documentation anyways (as is the case here, too). I think `documentation` and `doi` are really helpful though.

KevinMenden (approved):
Looks good :) 

One thing I'm still wondering is whether it should at least be noted that it would be nice to have multi-tool containers emit all software versions, just in case that could be helpful some day.
E.g. main tool would still be just `version` and then the helper tools are `tool_version` or something like this. Not super important but it also wouldn't hurt to add this to these containers.
Just my two cents though :)

README.md
drpatelh: Yup. `data` and `config` will still stay under `tests` 👍🏽

README.md
drpatelh: This bit still needs a bit of work once we know what and how the `meta.yml` will look. Hence https://github.com/nf-core/modules/pull/152/commits/c92fe8701a0c32d36754fca3dd5fe88b8c37d523 😆

software/TOOL/SUBTOOL/meta.yml
drpatelh: With documentation I always go with the attitude that the more the better. I think we can just be lenient as to how we `lint` for it instead?

drpatelh:
Thanks @KevinMenden !

software/TOOL/SUBTOOL/meta.yml
KevinMenden: Yeah that's a good point :)

Emiller88 (commented):
Just a thought. Looks good to me otherwise!

README.md
```diff
@@ -21,7 +22,7 @@ A repository for hosting [Nextflow DSL2](https://www.nextflow.io/docs/latest/dsl
 - [Adding a new module file](#adding-a-new-module-file)
     - [Module template](#module-template)
     - [Guidelines](#guidelines)
-    - [Testing](#testing)
+    - [CI tests](#ci-tests)
```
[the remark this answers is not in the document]
drpatelh: They are both no? But the primary goal for adding the tests is CI which I explained briefly in the first para. There is a separate section for local tests now.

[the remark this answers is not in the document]
drpatelh: But you may be right. We could leave it as being generic.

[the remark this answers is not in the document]
Emiller88: My main goal of all of this was just to be able to test locally and quickly :laughing: . The CI stuff was a nice bonus. This has always been a white whale of mine coming from more general software development that testing workflows was so slow and the commands were painful. Which makes the feedback loop slow and thus development slow.

README.md
Emiller88: This path isn't right as it is right now, It should be `tests/software/TOOL/SUBTOOL/`

README.md
```diff
@@ -213,51 +231,75 @@ using a combination of `bwa` and `samtools` to output a BAM file instead of a SA
...
+        ```console
+        cd /path/to/git/clone/of/nf-core/modules/
+        TMPDIR=~ PROFILE=singularity pytest --tag tests/software/bowtie/build --symlink --wt 2 --keep-workflow-wd
+        ```
+
+    - Typical command with Conda:
+
+        ```console
+        cd /path/to/git/clone/of/nf-core/modules/
+        PROFILE=conda pytest --tag tests/software/bowtie/build --symlink --wt 2 --keep-workflow-wd
```
Emiller88, line 291: These tags aren't right. They may work but I think they're not the best direction to put people on. We should be able to get by with just `--tag bowtie_build`

Emiller88: Just confirmed these tags don't work

drpatelh: Ok. Let me try.

drpatelh: `--tag bowtie_build` doesn't work for me because that one is specifically broken as we know! Sorry, it was late and I was speed writing docs! However, I do know why I put the path down. It's because you can also run the tests without a `--tag` but with the full path to the folder containing the `test.yml`

e.g.
```console
TMPDIR=~ PROFILE=singularity pytest tests/software/bowtie/build/ --symlink --wt 2 --keep-workflow-wd
```

This test still works even with the issues we were having with getting `bowtie build`.

PR #155, opened 2021-02-09
ci: Add number of modules

Emiller88:


Emiller88:
Closing in favor of just adding it to the nf-core site.

PR #161, opened 2021-02-09
Add bedtools getfasta module

JoseEspinosa:
Adding the `bedtools-getfasta` module as discussed [here](https://github.com/nf-core/viralrecon/issues/149)

PR #163, opened 2021-02-09
Add conda build id

JoseEspinosa:
Some modules did not have the Conda build id as spotted by @drpatelh on https://github.com/nf-core/modules/pull/161


Emiller88 (approved):
LGTM!

drpatelh:
Beat me to it! Thanks @JoseEspinosa ! Another thing we need to look out for are `output` channels that are emitted with `*.txt`. I noticed this [here](https://github.com/nf-core/modules/blob/6765db81bebad465acaf54a28f289ce6f1c55ac6/software/bcftools/stats/main.nf#L24-L25) when I added this module to viralrecon. The `txt` channel will contain the version file too. Part of me is thinking we rename the version file to `.version` but that might be too much hassle 🤷🏽

PR #165, opened 2021-02-09
Bump Pangolin version and minor amendments

drpatelh:


drpatelh:
This is what the file looks like with Singularity. Can't seem to get Conda working locally to figure out why there is a discrepancy in the md5sum, and I also don't have the will power anymore. Will have a look tomorrow.

```console
$ more test.pangolin.csv
taxon,lineage,probability,pangoLEARN_version,status,note
MT192765.1_Severe_acute_respiratory_syndrome_coronavirus_2_isolate_SARS-CoV-2/human/USA/PC00101P/2020__complete_genome,B.1,1.0,2021-02-05,passed_qc,
```

KevinMenden:
Okay, I ran it locally with conda and could reproduce the md5 sum. Checked with Docker as well. Here are the outputs:

Conda:
```
taxon,lineage,probability,pangoLEARN_version,status,note
MT192765.1_Severe_acute_respiratory_syndrome_coronavirus_2_isolate_SARS-CoV-2/human/USA/PC00101P/2020__complete_genome,B.1,1.0,2021-02-06,passed_qc,
```

Docker:
```
taxon,lineage,probability,pangoLEARN_version,status,note
MT192765.1_Severe_acute_respiratory_syndrome_coronavirus_2_isolate_SARS-CoV-2/human/USA/PC00101P/2020__complete_genome,B.1,1.0,2021-02-05,passed_qc,
```

So the pangoLEARN_version is different by a day 🤦 
I have the feeling (hope) that this might be the same version with just a different time stamp. But that's not good - it's actually a quite important thing to check. Will try to dig a bit into the different pangolin implementations to spot the source ...

KevinMenden:
Okay so `pangoLearn` is part of the conda recipe for `pangolin`.  The problem is that the version is not pinned:
`pangolearn >=2021.02.01` 

Hence while the pangolin version stays the same, if the pangoLearn version get's an update, installing `pangolin` via conda will pull the latest `pangoLearn` version, while the containers keep the old version.

The obvious way to solve this would be to pin the pangoLearn version in the pangolin recipe, but I'm assuming the developers want to not link them for a reason (i.e. so you can update the database without having to reinstall pangolin).

drpatelh:
Ah, great spot! Right, so we have no  choice here but to pin the `pangoLearn` version too. Our emphasis here is reproducibility and we need these tests to work across Conda/Docker/Singularity. This was actually a good test for that. If we want to use a more updated version of `pangoLearn` then the recipe should be pushed to Bioconda periodically as a different build for the same version of `pangolin`? That would be the best way to deal with it...or users can provide their own container if they wish.

What do you think?

Pinging @andersgs here because he recently updated the `pangolin` recipe on Bioconda changing it use the latest version of `pangoLearn` by default.

KevinMenden:
Yes totally agree with you, the best option would be to pin the pangoLearn version. If the recipe is already updated then it should be fine 🤔 But at the least in the current version it's not pinned.

Seems like those conda CI tests arrived just in time!

drpatelh:
Done for now @KevinMenden 🙂

KevinMenden (approved):
👍

PR #177, opened 2021-02-16
Add gatk/createsequencedictionary

MaxUlysse:

PR #178, opened 2021-02-16
Add cpus, memory and time requirements to limit resource consumption

MaxUlysse:
I think it's a good practice to have that especially to limit resources for our CI tests here with GitHub actions
cf #177

### DIFF
--- a/tests/config/nextflow.config
+++ b/tests/config/nextflow.config
@@ -5,6 +5,12 @@ params {
     singularity_pull_docker_container = false
 }
 
+process {
+  cpus = 2
+  memory = 6.GB
+  time = 48.h
+}
+
 if ("$PROFILE" == "singularity") {
     singularity.enabled = true
     singularity.autoMounts = true


Emiller88 (approved):
Looks good to me! Can't tell if the tests will pass or not.

MaxUlysse:
I'll ping @drpatelh and @KevinMenden to have their review too then

KevinMenden:
Agree makes sense but I'll leave that the decision to the master of modules himself 😁

drpatelh (approved):
Sure. Makes sense!

Emiller88:
I take it back, I think these may need to override the process labels. It's at least not working on my pipelines local modules.

MaxUlysse:
@Emiller88 What's your issue?

Emiller88:
> @Emiller88 What's your issue?

Ah I think I figured out what my problem was. I'm working on testing nf-core pipelines with pytest-workflow, and there's not any process label resources defined in modules, but in an nf-core pipeline it just takes the default label resources and tries to run them with too much request memory.

Anyways don't mind me!

PR #134, opened 2021-02-02
Bcftools viralrecon

JoseEspinosa:
Adding `bcftools-filter`, `bcftools-stats`, `bcftools-bgzip`, `bcftools-tabix`, `bcftools-consensus` and `bcftools-isec` as listed [here](https://github.com/nf-core/viralrecon/issues/149).

PR #179, opened 2021-02-16
Replace bcftools/tabix by tabix/tabix

MaxUlysse:


MaxUlysse:
I'm assuming here that the input is a bgzip vcf file.
I'm wondering if we should consider that input could be just a vcf too, and in this case, we would need to add a bgzip command to compress it.
In any case, the vcf file need to be sorted for indexing (whether it is gzip or not).

MaxUlysse:
And actually, just realized that we can also tabix index bed files, so will need to include that in current version too

MaxUlysse:
ok, so I made HTSLIB_TABIX more modular, works for bed, gff and vcf files.
All files need to be bgziped before, so added a bgziped bed and gff.
(I assumed the A.bed.gz was gunziped and not bgziped, so prefered to add a B.bed.gz).
I'll add a HTSLIB_BGZIP process later that can work with this module to start from regular bed, gff and vcf files.
I did not include the tabix indexing of sam as I assume no one is using that anymore.

MaxUlysse:
Another point of interest that could be good to address.
I'm not using meta here, mainly because even if I do tabix index all my vcf files within Sarek (that would have such information), I also use that to index some reference files that don't have it.
So I'm wondering on how to asses that matter.

MaxUlysse:
cf https://github.com/nf-core/modules/pull/134

MaxUlysse:
Following the conversation on Slack:

- Rename HTSLIB/TABIX into TABIX/TABIX

- Merge BCFTOOLS/TABIX with it

- use the TABIX container

tests/software/tabix/tabix/main.nf
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env nextflow
+
+nextflow.enable.dsl = 2
+
+include { TABIX_TABIX as TABIX_BED } from '../../../../software/tabix/tabix/main.nf' addParams( options: ['args': '-p bed'] )
+include { TABIX_TABIX as TABIX_GFF } from '../../../../software/tabix/tabix/main.nf' addParams( options: ['args': '-p gff'] )
+include { TABIX_TABIX as TABIX_VCF } from '../../../../software/tabix/tabix/main.nf' addParams( options: ['args': '-p vcf'] )
```
drpatelh, line 7: Love these different tests ❤️

.github/filters.yml
```diff
@@ -136,6 +132,10 @@ gunzip:
   - software/gunzip/**
   - tests/software/gunzip/**
 
+htslib_tabix:
+  - software/htslib/tabix/**
+  - tests/software/htslib/tabix/**
```
drpatelh, line 137:
```suggestion
tabix_tabix:
  - software/tabix/tabix/**
  - tests/software/tabix/tabix/**
```

drpatelh (approved):

tests/software/tabix/tabix/main.nf
MaxUlysse: Choose just not to include the sam possibility, as it's more common to use bam and their bai index anyway

PR #238, opened 2021-02-22
Splitcigar

KevinMenden:
Adding the `gatk4 SplitNCigarReads` module.

Also added minimal `fastq` and `bam` files (from the fastq files) for Sarscov2. Wasn't able to find matching BAM files otherwise ...


MaxUlysse (approved):
:fire:

PR #239, opened 2021-02-22
add bismark/align + tests

phue:
## PR checklist

- [x] Add a resource `label`
    - [x] `PROFILE=docker pytest --tag <MODULE> --symlink --wt 2 --keep-workflow-wd`
    - [x] `PROFILE=singularity pytest --tag <MODULE> --symlink --wt 2 --keep-workflow-wd`
    - [ ] `PROFILE=conda pytest --tag <MODULE> --symlink --wt 2 --keep-workflow-wd`


phue:
Conda failure is somewhat expected.

The other failures are becauses of checksum mismatches. I assume it's because the `*report.txt` contains timestamps, indicating how long the alignment took. That's why the checksum keeps changing, can't do much about it except for removing it.

software/bismark/align/meta.yml
```diff
@@ -0,0 +1,75 @@
...
+    - meta:
+        type: map
+        description: |
+            Groovy Map containing sample information
+            e.g. [ id:'test', single_end:false ]
+    - bam:
+        type: file
+        description: Output BAM file containing read alignments
+        pattern: "*.{bam}"
+    - report:
```
drpatelh, line 66: Add entry for `unmapped`

software/bismark/align/main.nf
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    input:
+    tuple val(meta), path(reads)
+    path index
+
+    output:
+    tuple val(meta), path("*bam")       , emit: bam
+    tuple val(meta), path("*report.txt"), emit: report
+    tuple val(meta), path("*fq.gz")     , optional:true, emit: unmapped
```
drpatelh, line 28: So you don't need to add any customisation here for  single- or paired-end reads e.g. [here](https://github.com/nf-core/modules/blob/4703a604431b48524a0e043a107eb76eb4674717/software/bowtie2/align/main.nf#L35)?

software/bismark/align/main.nf
```diff
@@ -0,0 +1,65 @@
...
+        // Check that we have enough memory, assuming 13GB memory per instance (typical for mouse alignment)
+        try {
+            def tmem = (task.memory as nextflow.util.MemoryUnit).toBytes()
+            def mcore = (tmem / mem_per_multicore) as int
+            ccore = Math.min(ccore, mcore)
+        } catch (all) {
+            log.warn "Not able to define bismark align multicore based on available memory"
+        }
+
+    }
```
drpatelh, line 49: This stuff still scares me 🙈

drpatelh (approved):
Couple of minor comments @phue 👍🏽

software/bismark/align/main.nf
phue: Took this from the current implementation in nf-core/methylseq
This is a good default for mouse data I suppose, but maybe it would be better to do this sort of resource fine-tuning elsewhere?
@ewels what do you think?

software/bismark/align/main.nf
phue: `bismark` handles this. 
Excerpt from the [docs](https://github.com/FelixKrueger/Bismark/blob/60653293acc5ad1943865661fd17140322910aaa/Docs/README.md#output):

```
Written reads will appear as they did in the input, without any translation of quality values that may have taken place within Bowtie or Bismark. Paired-end reads will be written to two parallel files with _1 and _2 inserted in their filenames, i.e. unmapped_reads_1.fq.gz and unmapped_reads_2.fq.gz
```

software/bismark/align/main.nf
phue: I removed the whole thing, it's probably not a good idea to just use `task.cpus` here, because it somewhat depends on available memory.
So I suggest we leave that to the pipeline code and pass it via `options.args`

software/bismark/align/main.nf
```diff
@@ -0,0 +1,44 @@
...
+def options    = initOptions(params.options)
+
+process BISMARK_ALIGN {
+    tag "$meta.id"
+    label 'process_high'
+    publishDir "${params.outdir}",
+        mode: params.publish_dir_mode,
+        saveAs: { filename -> saveFiles(filename:filename, options:params.options, publish_dir:getSoftwareName(task.process), publish_id:meta.id) }
+
+    conda (params.enable_conda ? "bioconda::bismark==0.23.0" : null)
```
phue, line 14:
```suggestion
    conda (params.enable_conda ? "bioconda::bismark=0.23.0" : null)
```

phue:
Merged in https://github.com/nf-core/modules/pull/297

software/bismark/align/main.nf
ewels: I am many (many) months too late, but I'm not sure that it is a good idea to remove this. Without it we basically can't use multi-threading for bismark alignment which drastically slows execution.

I don't think that we can (easily) set this via a config file as it has to be dynamically calculated based on the resources supplied. I wrote this code with a lot of trial and error and discussion with Felix about what Bismark is doing under the hood. It is not simple, but the code is well tested.

Although the defaults are set for mouse, I think that this is acceptable as most people will be running on Human or Mouse, or something smaller. If not then they can customise the values used to optimise.

By removing the code we just flatten to single core execution unless the user goes out of their way to customise the config to set this, which I suspect basically no-one will ever do.

PR #240, opened 2021-02-22
Added GATK4 Revertsam

KevinMenden:
Adding the `gatk4 revertsam` module.

https://github.com/nf-core/modules/issues/233


MaxUlysse (approved):
:fire:

PR #241, opened 2021-02-22
Attempt to fix conda CI

grst:


.github/workflows/pytest-workflow.yml
```diff
@@ -80,7 +80,7 @@ jobs:
 
       # Test the module
       - name: Run pytest-workflow
-        run: TMPDIR=~ PROFILE=${{ matrix.profile }} pytest --tag ${{ matrix.tags }} --symlink --wt 2 --kwdof
+        run: TMPDIR=~ PROFILE=${{ matrix.profile }} pytest --tag ${{ matrix.tags }} --symlink --wt 1 --kwdof
```
grst, line 83: need to set `wt=1` for conda to avoid race condition on the conda cache.

phue:
Great! This should totally fix the issue.
The current failure is a bit puzzling but I guess it's because the packages installed in the container are build-pinned and the one installed via conda is not, so the binaries are probably different.

Issue #242, opened 2021-02-23
Shrinking test data to a minimal, coherent set


KevinMenden:
For the vast majority of tools, we need similar test data, like fastq, bams etc.
Often is required to use different files from the same organism. Currently, for many test datasets, it is not clear how they have been generated (e.g. which organism, fasta, gtf ...) which limits their usage.

I would therefore propose to try and limit the test data to one single organism, and to remove files which are named like 'test_r1.fastq' because nobody knows what they are anymore.

Emiller88:
Just going to repost this from slack because I didn't know there was an issue.

I would vote we do one of two options(both use [bfg](https://rtyley.github.io/bfg-repo-cleaner/), and just because keeping data files in git repos is a personal pet peeve):

1. Use bfg to remove all the history of the data files(or change them to git-lfs) then delete them and use the file('https://github.com/smithlabcode/preseq/raw/master/data/SRR1003759_5M_subset.mr', checkIfExists: true format for all the tests. Then we can pull whatever that module calls for from the nf-core test-data, from the repos of the software(which has been really handy). You won't be able to get started working on a plane though.

2. Run bfg and change all the data files to git-lfs, then change them to a single organism.

As @grst pointed out
>(just be aware that using these tools will change all commit ids)

KevinMenden:
Okay, as discussed on Slack (ping @drpatelh ), we want to make everything properly organized. Just a few things to decide:

Structure:
* generic
* genomics
    * organism
        * fasta
            * test.fasta
 
With the above structure we would not have any naming of the organism in the actual file anymore (`test.fasta`) because that would be taken care of by the parent directory. Advantage (?) is that all files of the different organisms are named the same. Disadvantage is that we loose any potentially helpful information in the name, e.g. the exact genome version. This would therefore require careful documentation in the README.md.
Personally I prefer having some kind of meaningful file names over complete standardization between organisms.

I will furthermore try to remove one of the two sarscov2 sequences and associated files. 

The generic folder is not properly organized currently because there are still some files we have to get rid of. If we end up needing the E.coli data after all, we'll move that into `genomics` as well. 

Okay so maybe we write down any other thoughts about this here as well so it is easier to track :)

drpatelh:
My bad, I am a lazy issue poster at the best of times due to barrages of Slack messages 😅 

Ok so this is what I originally posted on Slack:
[...]

PR #249, opened 2021-02-24
Converge test data usage

KevinMenden:
As discussed in slack and mentioned here https://github.com/nf-core/modules/issues/242, I have restructured the test data a bit.

There are now two folders, `generic` and `genomics`. The former contains any random stuff or things that we want to remove later and doesn't fit a category, while genomics contains test data for genomics experiments, sorted by organism.

The size of the test data did now shrink from about 70MB to 17 MB. There are still some unnecessary files in generic, mostly for `STAR` and `methyldackel`, which need to be removed or exchanged, or at least use only one set of data for E.coli.

The README will also need to be formatted a bit better, and we should add some guidelines on how to add tests.


.github/filters.yml
```diff
@@ -66,7 +66,7 @@ blast_blastn:
   - software/blast/blastn/**
   - tests/software/blast/blastn/**
 
-bowtie:
+bowtie_build:
```
Emiller88, line 69: Does this work now?

KevinMenden: Ah no ... sadly it didn't. Tried around a bit and forgot about this change. But actually I think the naming makes more sense this way.
The only way around the bowtie thing is to call the tool different, so not `build`, maybe `build-index` or something

KevinMenden:
Okay there are only 10 tests failing now, which are either conda-related or `bowtie build`tests - so that's as good as it gets.

The data structure should be good for now as well. I would really like to merge this into modules as fast as possible, because all the new modules still use the old structure and I will have to fix all of them too.
We can take care of the stuff in generic one by one afterwards, because that's usually only used by specific tools. For new modules we should really try to use the `genomics` folder wherever possible.
I will make an issue for all the data that we still need to fix.

Also let me know what else we should have in the README or just edit it :) 

ping @drpatelh

phue:
Do we actually need the `dummy_file.txt`?  

The alternative would be to just use `file("dummy_file.txt")` to create an empty file where needed

KevinMenden:
Yes that should work as long as we don't check whether this file exists. If it works I'm all for it!

phue:
It is a bit strange that the `bwameth` tests don't run. They should fail because the methylated reads are gone

phue:
@KevinMenden 
I generated (highly) artificial bisulfite reads for SARS-CoV-2 using [Sherman](https://github.com/FelixKrueger/Sherman) 
We can build indices and bamfiles from that, so we can get rid of some E.coli data.

Shall I push it here or in a separate PR afterwards?

KevinMenden:
> It is a bit strange that the bwameth tests don't run. They should fail because the bisulfite reads are gone

Might be because I didn't change anything in that code, then no tests are run I think.

> I generated (highly) artificial bisulfite reads for SARS-CoV-2 using Sherman
We can build indices and bamfiles from that, so we can get rid of some E.coli data.
Shall I push it here or in a separate PR afterwards?

Awesome, thanks!! 
However you want, both works 👍

.github/filters.yml
Emiller88: It was bowtie_index before, a git blame shows @drpatelh changed it to the proper name, so maybe he can weigh in.

KevinMenden:
@drpatelh would be great if you could have a look at this :) 
no need to look at all the tests, i Just exchanged the file paths.

I would like to get this done before adding any other module.

drpatelh:
Looks great @KevinMenden ! Minor things. Can we rename `genomic` -> `genome` in the names of the tests files? This leaves the options open for using `transcriptome` where required too so great you have factored that in. 

Still not happy about the `test_1.2.fastq.gz`, `test_1.fastq.gz` file naming convention when you have multiple test files of the same type, they aren't listed together this way when sorting by name...maybe we go for `test2_1.fastq.gz`, `test2_2.fastq.gz` instead and rename the files in the other folders accordingly too? e.g. `bed/test.2.bed` -> `bed/test2.bed`

`bed/test.genome.sizes` isn't strictly speaking a BED file so maybe we have a `genomics/sarscov2/general/` folder for these sorts of files unless you can think of a better option?

KevinMenden:
> Can we rename genomic -> genome in the names of the tests files? This leaves the options open for using transcriptome where required too so great you have factored that in.

I took the naming convention that is used by I think NCBI or where I downloaded this, there the genome files have `genomic` in the file name.

Okay but that's the naming convention we agreed on (are rather you suggested 😜 ) and if the only reason to change it again is to make the files float together in the folder, well ... please consider that I would have to change this for every test file where there are copies, and then go through every single module because I don't know by heart which one uses which files, change the name, potentially run the tests again and adapt the md5-sums because they sometimes depend on the file name, which I often don't know beforehand. 
It would probably be a bit quicker this time because I don't have to change the whole file path but still, I will need at a few hours just to change the name again.

Basically what I'm trying to say is - I would rather not change it *again*. 😁 
Sorry for being a bit stubborn here but I already changed every single test twice, which is incredibly boring work and I am a bit terrified from doing it again. 

I'm down for moving the `test.genome.sizes` though. 😎

drpatelh:
Ok. Don't worry, I understand. I have had to change everything a bunch of times as we have gone along too and it's.....

Let's leave `genomic` as it is then, and if and when I have the energy I will see what I can do about renaming the test files seeing as it was my suggestion and I obviously hadn't thought it through at that point 🙂 

Thanks!

drpatelh:
PS: If you want to find files recursively in a directory containing some sort of content you can use this nugget. It has saved me alot of time and helps when renaming / removing / changing things:

```console
find ./ -type f -exec grep -H 'test_2.2.fastq.gz' {} \;
```

KevinMenden:
Ah thanks, that's super useful indeed!!
I was just thinking that I could probably just write a simple python script that reads through all the file names and changes all the files with exact match to a certain name. That would probably work pretty easily .... 
It would still break a bunch of tests but we could see this here and then only repair those. So this way it wouldn't be that painful actually ...
Might have a look at it, maybe I'll do it after all. Or with the next PR - there will still be some cleaning up to do but at least new modules can start using the new tests data structure once this one is merged.

KevinMenden:
Turns out I was *a bit* paranoid - forgot that this only affects a few files, because I don't have to change the normal `test.bed` files etc. Anyway wrote a little python script for it.

KevinMenden:
By the way a transcriptome is already included --> `test_cds_from_genomic.fasta` 
That's the NCBI way of naming it.

I could change that to `test_transcriptome.fasta` and `test_genome.fasta` using the little python helper. It will probably cause some trouble with the `test.yml` files and indices but shouldn't be too bad if the name-changing is already done. So let me know what you think.
I will start compiling the VCF files for sarscov2 and also look into the remaining 'generic' files today - but this will be a separate PR.

drpatelh:
Ah, that wasn't obvious to me at the time although it is now that you have pointed it out. Yes, renaming would be great please 😎 It is how we have have named those sorts of files in nf-core/rnaseq and how they are named on AWS iGenomes too.

tests/data/README.md
```diff
@@ -0,0 +1,49 @@
...
+    * gtf
+        * 'test_genomic.gtf': GTF for MT192765.1 genome
+        * 'test_genomic.gff3': GFF for MT192765.1 genome
+        * 'test_genomic.gff3.gz': bgzipped-version
+    * paf
+        * 'test_cds_from_genomic.paf': PAF file for MT192765.1  genome
+
+### generic
+
+* 'a.gff3.gz': bgzipped gff3 file currently necessary for TABIX test
```
MaxUlysse, line 40: we can gzip any gff file for this tabix test now that we have the tabix_gzip module as well

KevinMenden: yes good point I forgot about that. Think it just has to be bgzip and then it works.

KevinMenden: swapped it

KevinMenden:
Shall we merge @drpatelh ? (need an approval)

drpatelh:
Yup! Let's do it! Thanks @KevinMenden :)

drpatelh (approved):

PR #253, opened 2021-02-24
Bedtools output

Emiller88:
There may be more in the bedtools modules like this, and I'll add them to the PR once I try them, but @drpatelh let me know if you'd like the naming style to be different.


KevinMenden:
I have basically already addressed all those things in https://github.com/nf-core/modules/pull/249 
So that PR would overwrite this one

Emiller88:
Solid I'll close this one, just wanted to keep people from making the same mistake I did!

PR #261, opened 2021-03-08
Exchange VCF data by sarscov2 data

KevinMenden:
This PR exchanges the old VCF data in `tests/data/generic/vcf` by sarscov2 VCF data in `tests/data/genomics/sarscov2/vcf` 
The old data is now removed. The new data was generated using `bcftools` and the commands `mpileup`, `call` and `filter`  from the sarscov2 bam files and genome (see also `README.md`). 

File paths were exchanged for tests of:
* `bcftools`
* `tabix`
* `gatk4 mergevcfs` 

Minor bug fix: I exchanged the " quotes in `bcftools consensus` by ' quotes because it didn't work for me otherwise.


MaxUlysse (approved):
LGTM

PR #270, opened 2021-03-12
Remove def from module options definition in main

### LINKED ISSUE #268: [FEATURE]: Additional module `options` should trigger process to be rerun
As reported by @d4straub on #modules Slack channel

_I ran into a problem when re-writing nf-core/ampliseq in DSL2:
When I run the pipeline and then resume with a changed parameter, cached processes are loaded and processes are not newly started. This is true for parameter that are given to the modules by 'addParams'. When using channels (e.g. "meta"), than it works as expected. I also heard of that problem in Harshil's bytesize talk. If I remember correctly, all "sample specific" parameters, e.g. single_end, should be in "meta", but I am talking about global parameters that affect all samples and that might be changed relatively frequently to optimize specific processes. For example using a different adapter/primer to remove with the cutadapt module. For example, I'd want that fastqc is not re-processed, but cutadapt and all dependent processes are re-started when using resume. This gets more relevant for later processes in a pipeline of course.
There are some solutions that I can think of that all seem not perfect to me:
(1) Add all parameters in the beginning of the pipeline into the "meta" of the input channel
-> problem: all processes will be re-processed when one parameter is changed, e.g. fastqc on raw data even if the changed parameter is not changing anything in that process but later in the pipeline
(2) Add parameters relevant to a specific step just before that specific step into "meta"
-> advantage: only that specific step is re-processed (and all dependent), problem: joining ("join()") channels with different "meta" doesn't work out, therefore the parameter needs to be removed again after the process, making a lot of mapping (".map{}") necessary
(3) Add parameters relevant to a specific step as separate input, such as:_
```
input:
tuple val(meta), path(data)
val(additional meta)
val(nochnmeta)
-> problem: doesnt work with nf-core modules, advantage: works beautifully with local modules, keeps additional code at a minimum, meta not touched and therefore "join" does not break.
I'd be happy to hear about a better solution that I didnt think of. Until now, I favour (3) because ampliseq uses software/modules that are amplicon specific and therefore most likely not relevant for other pipeline anyway.
``` 
From @drpatelh:

_Overall, however it was a much cleaner approach to provide all of these options via a single options map via addParams but now as you pointed out the caching ability when changing tool-specific parameters doesn't work as expected.
I suspect something is not quite working in terms of how these tool-specific options are being passed to the module itself. That would be the best fix because we wouldn't need to explicitly have a val options input for every single module.
If we can pinpoint something in NF that is causing this or maybe the custom functions script I have written then we can try and fix it._

JoseEspinosa:
The  definition of `options` map locally in the `main.nf` of the module process was preventing the resuming of the processes when additional arguments were provided to the module. By removing the `def` the behavior results as expected. Closes #268 (see for a more detailed explanation of the original problem)

## PR checklist


drpatelh (approved):
🆒

PR #274, opened 2021-03-15
add bismark/methylation_extractor

phue:
## PR checklist

- [x] Add a resource `label`


phue:
Today I learned that gzip does not produce deterministic results because it stores timestamps in the header. That's why the md5sums will keep changing.

Even if we omit the  `--gzip` flag from the process, the `bedGraph` and `coverage` outputs will be compressed by default, so that is not really an option.
Maybe there is a way to fake a timestamp in Github actions?
Otherwise I would just remove the md5sum checks for the compressed files, unless someone has an idea?

software/bismark/methylation_extractor/main.nf
```diff
@@ -0,0 +1,48 @@
...
+    } else {
+        container "quay.io/biocontainers/bismark:0.23.0--0"
+    }
+
+    input:
+    tuple val(meta), path(bam)
+    path index
+
+    output:
+    tuple val(meta), path("*.bedGraph.gz")         , emit: bedGraph
```
drpatelh, line 26:
```suggestion
    tuple val(meta), path("*.bedGraph.gz")         , emit: bedgraph
```

software/bismark/methylation_extractor/main.nf
```diff
@@ -0,0 +1,48 @@
...
+
+    script:
+    def seqtype  = meta.single_end ? '-s' : '-p'
+    def software = getSoftwareName(task.process)
+    """
+    bismark_methylation_extractor \\
+        --bedGraph \\
+        --counts \\
+        --gzip \\
+        --report \\
```
drpatelh, line 41: Should these be coming in via `options.args` instead?

software/bismark/methylation_extractor/main.nf
```diff
@@ -0,0 +1,48 @@
+// Import generic module functions
+include { initOptions; saveFiles; getSoftwareName } from './functions'
+
+params.options = [:]
+def options    = initOptions(params.options)
```
drpatelh, line 5:
```suggestion
options        = initOptions(params.options)
```

software/bismark/methylation_extractor/meta.yml
```diff
@@ -0,0 +1,87 @@
...
+        type: dir
+        description: Bismark genome index directory
+        pattern: "BismarkIndex"
+output:
+    - meta:
+        type: map
+        description: |
+            Groovy Map containing sample information
+            e.g. [ id:'test', single_end:false ]
+    - bedGraph:
```
drpatelh, line 62:
```suggestion
    - bedgraph:
```

drpatelh (changes requested):
Looks great @phue ! Couple of changes.

software/bismark/methylation_extractor/main.nf
```diff
@@ -0,0 +1,48 @@
...
+
+process BISMARK_METHYLATION_EXTRACTOR {
+    tag "$meta.id"
+    label 'process_high'
+    publishDir "${params.outdir}",
+        mode: params.publish_dir_mode,
+        saveAs: { filename -> saveFiles(filename:filename, options:params.options, publish_dir:getSoftwareName(task.process), publish_id:meta.id) }
+
+    conda (params.enable_conda ? "bioconda::bismark==0.23.0" : null)
```
drpatelh, line 14:
```suggestion
    conda (params.enable_conda ? "bioconda::bismark=0.23.0" : null)
```

drpatelh:
> Today I learned that gzip does not produce deterministic results because it stores timestamps in the header. That's why the md5sums will keep changing.

No way?!  Does that go for any zipped file? Assuming you have ruled out this isn't down to the core file changing?

phue:
> > Today I learned that gzip does not produce deterministic results because it stores timestamps in the header. That's why the md5sums will keep changing.
> 
> No way?! Does that go for any zipped file? Assuming you have ruled out this isn't down to the core file changing?

When I extract the files, checksums match, so gzip is the culprit here

software/bismark/methylation_extractor/main.nf
phue: I *think* these are sensible defaults. By default, the tool produces output per strand so the `--bedGraph` is useful to consolidate this into one file. Also the output can get really massive so it is sensible to put `--gzip`
And `--report` is something that `bismark_summary` and ultimately `MultiQC` needs

FelixKrueger: Just to clarify, `--report` is on by default, but doesn't hurt to specify explicitly. 

The same goes for `--counts`: the coverage files does have the counts, the `bedGraph` file doesn't, and both files are produced for each input if `--bedGraph` is given.

drpatelh: Cool beans!

drpatelh (approved):
Gracias!

PR #278, opened 2021-03-18
Check for minimum NF version

drpatelh:
This exits nicely if you are running any module tests with `nf-core modules create-test-yml` and are running with an older version of Nextflow.


ewels (approved):
LGTM!

PR #317, opened 2021-03-22
ci: Add 21.03.0 version to matrix

Emiller88:


drpatelh:
I suspect it's best to stick with the latest edge version but no harm to try I guess.

drpatelh (approved):

Issue #299, opened 2021-03-22
Add read-groups to test data bam files


KevinMenden:
At least some of the BAM files in the sarscvo2 test data should have read-groups, as some tools need that information.

yocra3:
Solved with PR #324

PR #324, opened 2021-03-23
Add readgroup to bam files

yocra3:
This PR solves the bug described in #299 as discussed during the nf-core hackathon. I have added readgroup information to all bam files. Then, I have reindexed the files that already had a bai file.  Finally, I have added a recalibration table for testing `applybsqr` and `gatherbqsrreports`. All these changes have been applied to the README.


KevinMenden (approved):
Yep that looks good to me! It's probably going to break a few tests but that's just life 🤷

Issue #262, opened 2021-03-08
BWA_MEM and BWAMEM2_MEM pass too many indexs


edmundmiller:
## Description of the bug

When using these modules with aws-igenomes the mem fails, because the below lines pick up multiple indexes

```
INDEX=`find -L ./ -name "*.amb" | sed 's/.amb//'`
```

This line was introduced by @drpatelh 

Some more background
```s3://ngi-igenomes/igenomes/Staphylococcus_aureus_NCTC_8325/NCBI/2006-02-13/Sequence/BWAIndex/genome.fa
s3://ngi-igenomes/igenomes/Staphylococcus_aureus_NCTC_8325/NCBI/2006-02-13/Sequence/BWAIndex/genome.fa.amb
s3://ngi-igenomes/igenomes/Staphylococcus_aureus_NCTC_8325/NCBI/2006-02-13/Sequence/BWAIndex/genome.fa.ann
s3://ngi-igenomes/igenomes/Staphylococcus_aureus_NCTC_8325/NCBI/2006-02-13/Sequence/BWAIndex/genome.fa.bwt
s3://ngi-igenomes/igenomes/Staphylococcus_aureus_NCTC_8325/NCBI/2006-02-13/Sequence/BWAIndex/genome.fa.pac
s3://ngi-igenomes/igenomes/Staphylococcus_aureus_NCTC_8325/NCBI/2006-02-13/Sequence/BWAIndex/genome.fa.sa
s3://ngi-igenomes/igenomes/Staphylococcus_aureus_NCTC_8325/NCBI/2006-02-13/Sequence/BWAIndex/version0.5.x/genome.fa
s3://ngi-igenomes/igenomes/Staphylococcus_aureus_NCTC_8325/NCBI/2006-02-13/Sequence/BWAIndex/version0.5.x/genome.fa.amb
s3://ngi-igenomes/igenomes/Staphylococcus_aureus_NCTC_8325/NCBI/2006-02-13/Sequence/BWAIndex/version0.5.x/genome.fa.ann
s3://ngi-igenomes/igenomes/Staphylococcus_aureus_NCTC_8325/NCBI/2006-02-13/Sequence/BWAIndex/version0.5.x/genome.fa.bwt
s3://ngi-igenomes/igenomes/Staphylococcus_aureus_NCTC_8325/NCBI/2006-02-13/Sequence/BWAIndex/version0.5.x/genome.fa.pac
[...]

PR #326, opened 2021-03-23
Fix bwa index failing to put outputs in correct dir

aunderwo:
## PR checklist
* The argument `-p bwa/${${fasta.baseName}` doesn't make the indicies in the `bwa` dir unless it comes before the fasta sequence
* If these are made in a directory called index then the publishDir will become `bwa/index`

### DIFF
--- a/software/bwa/index/main.nf
+++ b/software/bwa/index/main.nf
@@ -22,14 +22,14 @@ process BWA_INDEX {
     path fasta
 
     output:
-    path "bwa"          , emit: index
+    path "index"          , emit: index
     path "*.version.txt", emit: version
 
     script:
     def software = getSoftwareName(task.process)
     """
-    mkdir bwa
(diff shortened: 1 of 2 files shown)


MaxUlysse:
we have this issue which is related: https://github.com/nf-core/modules/issues/262
In which I suggested a solution that needs to be experimented further:
https://github.com/MaxUlysse/nf-core_sarek/commit/35956ab3be9e09fea2142f7428a03ce8b4d30198

drpatelh:
Ah, sorry, I think you may have misunderstood my waffle on Slack @aunderwo We should change the `publish_dir` line [here](https://github.com/nf-core/modules/blob/9115c12f8861b587bea7b49cc6cab8d1501d051c/software/bwa/index/main.nf#L12). Will rustle up a PR for `bwa` and the other aligners and pull you in to review.

PR #334, opened 2021-03-23
add vcftools module

### LINKED ISSUE #214: new module: vcftools
I think it would be good to have module for [vcftools](https://bioconda.github.io/recipes/vcftools/README.html)

Mark-S-Hill:
## PR checklist

Closes #214 

New PR for vcftools module after cleaning up conflicts. 

- [x] Add a resource `label`

### DIFF
--- a/software/vcftools/main.nf
+++ b/software/vcftools/main.nf
@@ -99,31 +99,31 @@ process VCFTOOLS {
     def args     = options.args.tokenize()
 
     def bed_arg  = (options.args.contains('--bed')) ? "--bed ${bed}" :
-                   (options.args.contains('--exclude-bed')) ? "--exclude-bed ${bed}" :
-                   (options.args.contains('--hapcount')) ? "--hapcount ${bed}" : ''
+                       (options.args.contains('--exclude-bed')) ? "--exclude-bed ${bed}" :
+                       (options.args.contains('--hapcount')) ? "--hapcount ${bed}" : ''
     args.removeIf { it.contains('--bed') }
     args.removeIf { it.contains('--exclude-bed') }
     args.removeIf { it.contains('--hapcount') }
 
     def diff_variant_arg = (options.args.contains('--diff')) ? "--diff ${diff_variant_file}" :
-                           (options.args.contains('--gzdiff')) ? "--gzdiff ${diff_variant_file}" :
-                           (options.args.contains('--diff-bcf')) ? "--diff-bcf ${diff_variant_file}" : ''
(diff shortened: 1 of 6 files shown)


software/vcftools/main.nf
```diff
@@ -0,0 +1,129 @@
...
+    def software = getSoftwareName(task.process)
+    def prefix   = options.suffix ? "${meta.id}${options.suffix}" : "${meta.id}"
+    def args     = options.args.tokenize()
+
+    def bed_arg  = (options.args.contains('--bed')) ? "--bed ${bed}" :
+                   (options.args.contains('--exclude-bed')) ? "--exclude-bed ${bed}" :
+                   (options.args.contains('--hapcount')) ? "--hapcount ${bed}" : ''
+    args.removeIf { it.contains('--bed') }
+    args.removeIf { it.contains('--exclude-bed') }
+    args.removeIf { it.contains('--hapcount') }
```
grst, line 106: There's no way of using multiple of these options at the same time?

software/vcftools/main.nf
```diff
@@ -0,0 +1,129 @@
...
+    // Other optional input files can be utilised in a similar way to below but we do not exhaustively itterate through all
+    // possible options. Instead we leave that to the user.
+    tuple val(meta), path(variant_file)
+    path(bed)
+    path(diff_variant_file)
+
+    output:
+    path("*.version.txt"), emit: version
+
+    tuple val(meta), path("*.vcf"), optional:true, emit: vcf
```
grst, line 33: Wow, that's an impressive list, great job!

software/vcftools/main.nf
Mark-S-Hill: That's a good point. Using --bed and --exclude-bed at the same time wouldn't really make sense (include and exclude sites) but i guess using the hapcount together with the earlier options could be a real scenario, although i'm not sure how common. One solution would be to add an additional optional input for this and separate it from this block. Alternatively this could be done in separate steps in the workflow context as is (maybe this is better for the moment?). I am unsure what's best. We already can't cover all use cases as they are just too broad - so we are relying on users being able to make small tweaks to best match their needs. But happy to look in to adding the aforementioned extra input channel to handle scenarios where someone wants to work with more than one bed file with a single command if you think that would be better?

rivera10:
Looks good @Mark-S-Hill. Only some lint problems.

drpatelh:
This one is a beast! Converted to Draft so I can have a look at this properly.

drpatelh (approved):
Thanks @Mark-S-Hill ! Will merge and make any changes in a follow up PR so I can re-organise all of the test data.

Issue #363, opened 2021-03-24
Fix failing tests after data restructuring


KevinMenden:
This is a list of all module tests that we need to get the tests passing for again: Please assign yourself by editing this issue:

- [x] bowtie2/align (assigned @JoseEspinosa )
- [x] bowtie2/build_test (assigned @JoseEspinosa )
- [x] bwameth/align (assigned @JoseEspinosa )
- [x] bwameth/index (assigned @JoseEspinosa )
- [x] multiqc (assigned @JoseEspinosa )
- [x] gffread (assigned @KevinMenden )
- [x] sequenzautils/gcwiggle (assigned @KevinMenden )
- [x] homer/annotatepeaks (assigned @KevinMenden )
- [ ] bandage/image (assigned @KevinMenden )
- [x] seqwish/induce (assigned @KevinMenden )
- [x] salmon/index (assigned to @BatoolMM )
- [x] salmon/quant (assigned to @BatoolMM)
- [x] trimgalore (@grst)
- [x] picard/markduplicates  (@grst)
- [x] picard/mergesamfiles  (@grst)
- [x] picard/collectmultiplemetrics  (@grst)
- [x] picard/collectwgsmetrics  (@grst)
- [x] minia
- [x] pangolin (@rpetit3)
- [x] minimap2/align (@yuukiiwa)
- [x] vcftools (@KevinMenden )
- [x] bismark/report  (@KevinMenden )
- [x] bismark/align  (@KevinMenden )
- [x] bismark/genomepreparation  (@KevinMenden )
- [x] bismark/summary  (@KevinMenden )
- [x] bismark/deduplicate  (@KevinMenden )
- [x] bismark/methylationextractor  (@KevinMenden )
- [x] untar (@JoseEspinosa )
- [x] cat/fastq (@JoseEspinosa )
- [x] blast/blastn (@rpetit3)
- [x] blast/makeblastdb  (@rpetit3)
- [x] kraken2/run (@JoseEspinosa )
- [x] bedtools/genomecov (✔️ @drpatelh)
- [x] bedtools/slop (✔️ @drpatelh)
- [x] bedtools/merge (✔️ @drpatelh)
- [x] bedtools/getfasta (✔️ @drpatelh)
- [x] bedtools/sort (✔️ @drpatelh)
- [x] bedtools/complement (✔️ @drpatelh)
- [x] bedtools/intersect (✔️ @drpatelh)
- [x] bedtools/maskfasta (✔️ @drpatelh)
[...]

PR #379, opened 2021-03-24
Fix some module test after adapting test config approach

JoseEspinosa:
Updates path to new test files for as assigned on #363 

`bwameth/index`
`bwameth/align`
`bowtie2/align`
`bowtie2/build_test` aka as the black sheep of the tests... 
`multiqc`


drpatelh (approved):
Thanks @JoseEspinosa !

PR #388, opened 2021-03-25
ci: Upload work directory on failure

### LINKED ISSUE #374: [FEATURE] Include nextflow work directory in uploaded artifacts
I think it would be quite helpful for debugging to include the nextflow work directory into the artifacts 
generated for failed tests. 

@Emiller88, could you look into that at some point?

Emiller88:
Closes #374


drpatelh (approved):
Thanks @Emiller88 ! Yes, I am slightly concerned that these file sizes might get too big too. Be so cool if we could just post the `.nextflow.log` contents and other relevant files on the issue to save having to dig around the artifacts. I mean even a `tree` of the `work` directory would be fine because the failure info should be in `.nextflow.log` anyway. Not sure having the files in the `work/` directory helps too much.

PR #389, opened 2021-03-25
Lint modules ci

Emiller88:
This allows for linting the CI. I changed all the filters, so they'll be the same as how the `nf-core modules lint -t`likes them.

### DIFF
--- a/.github/workflows/pytest-workflow.yml
+++ b/.github/workflows/pytest-workflow.yml
@@ -40,6 +40,20 @@ jobs:
           key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
           restore-keys: |
             ${{ runner.os }}-pip-
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.x"
+      - name: Install dependencies
+        run: python -m pip install --upgrade pip pytest-workflow
+        # FIXME: Remove this when nf-core modules lint stabilizes
+      - name: Install nf-core tools development version
+        run: python -m pip install --upgrade --force-reinstall git+https://github.com/nf-core/tools.git@dev
+
+      - name: Lint ${{ matrix.tags }}
+        run: nf-core modules lint . -t ${{ matrix.tags }}
+
       - uses: actions/cache@v2
         with:
           path: /usr/local/bin/nextflow
@@ -54,12 +68,6 @@ jobs:
           wget -qO- get.nextflow.io | bash
           sudo mv nextflow /usr/local/bin/
 
-      - name: Set up Python
-        uses: actions/setup-python@v2
-        with:
-          python-version: "3.x"
-      - name: Install dependencies
-        run: python -m pip install --upgrade pip pytest-workflow
 
       - name: Set up Singularity
         if: matrix.profile == 'singularity'
--- a/tests/config/pytest_software.yml
+++ b/tests/config/pytest_software.yml
@@ -475,7 +475,6 @@ subread/featurecounts:
   - tests/software/subread/featurecounts/**
 
 tabix/bgzip:
->>>>>>> 4709a3e (ci: _ => /)
   - software/tabix/bgzip/**
   - tests/software/tabix/bgzip/**
 
--- a/tests/config/pytest_software.yml
+++ b/tests/config/pytest_software.yml
@@ -6,85 +6,85 @@ allelecounter:
   - software/allelecounter/**
   - tests/software/allelecounter/**
 
-bandage_image:
+bandage/image:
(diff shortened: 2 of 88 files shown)


.github/workflows/pytest-workflow.yml
```diff
@@ -40,6 +49,20 @@ jobs:
...
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.x"
+      - name: Install dependencies
+        run: python -m pip install --upgrade pip pytest-workflow
+        # FIXME: Remove this when nf-core modules lint stabilizes
+      - name: Install nf-core tools development version
+        run: python -m pip install --upgrade --force-reinstall git+https://github.com/nf-core/tools.git@dev
```
drpatelh, line 61: Nice!

tests/software/bandage/image/test.yml
```diff
@@ -2,7 +2,7 @@
   command: nextflow run ./tests/software/bandage/image -entry test_bandage_image -c tests/config/nextflow.config
   tags:
     - bandage
-    - bandage_image
+    - bandage/image
```
drpatelh, line 5: So what was the reasoning behind changing this for all modules? Not all modules have been changed because we have at least 110 of those puppies now.

.github/workflows/pytest-workflow.yml
```diff
@@ -8,18 +8,27 @@ jobs:
...
+          cmd: |
+            output=$(echo ${{ steps.filter.outputs.changes }} | yq e 'length' -)
+            echo "::set-output name=magic_number::$output"
+
   test:
     runs-on: ubuntu-20.04
     name: ${{ matrix.tags }} ${{ matrix.profile }} ${{ matrix.nxf_version }}
     needs: changes
-    if: needs.changes.outputs.modules != '[]'
+    if: needs.changes.outputs.modules != '[]' && needs.changes.outputs.num_modules <= 256
```
drpatelh, line 31: This won't run, period if we have more than 256 tests?

tests/software/bandage/image/test.yml
Emiller88: Basically anything that a `_` was supposed to represent a `/. That way the `modules lint`can get fed the proper tag. and I had to change them in all the tags because then the filter wouldn't trigger the pytest workflow.

.github/workflows/pytest-workflow.yml
Emiller88: That's the idea! For my next trick I'm going to have a job that only runs if the changes are greater than 256, and it's just going to run everything for now.

tests/software/bandage/image/test.yml
drpatelh: Gotcha! 🍬

.github/workflows/pytest-workflow.yml
drpatelh: Cool.

drpatelh:
Still some failures in the checks?

Emiller88:
Yeah trying to find the magic combo to run the all job if it fails. I'll ping you all when I figure it out.

.github/workflows/pytest-workflow.yml
KevinMenden: Nice :ok_hand:

KevinMenden:
Cool, so everything as usual when < 256, otherwise run it all in one go, right? Good solution :+1: 
As far as I can understand it looks good to me. I guess the reason for the `Test all modules` thing being skipped is that the change is now less than 256.

Emiller88:
@KevinMenden No I think the issue is because [it's hitting 256](https://github.com/nf-core/modules/actions/runs/687717730).

The reason it's still hitting 256 is because I'm using yq's default action, and that's not setting the output correctly so [it's blank](https://github.com/nf-core/modules/runs/2196425290?check_suite_focus=true#step:5:8). I think I'm going to just have to install it like we do nextflow

Emiller88:
Removed the scope creep and made a separate branch to fix the 256 issue at a later date.

tests/config/pytest_software.yml
```diff
@@ -454,47 +454,48 @@ spades:
...
   - tests/software/stringtie/**
 
-subread_featurecounts:
+subread/featurecounts:
   - software/subread/featurecounts/**
   - tests/software/subread/featurecounts/**
 
-tabix_bgzip:
+tabix/bgzip:
+>>>>>>> 4709a3e (ci: _ => /)
```
drpatelh, line 478:
```suggestion
```

drpatelh (approved):
Thanks @Emiller88 !

PR #397, opened 2021-03-31
Salmon alevin

KevinMenden:
Adding the `salmon alevin` module.

The tests work locally with some test data, that is too big to upload here. So I'll make this a draft PR until that issue is solved.
Because I'm not uploading the test data, all tests will fail.

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`


FloWuenne:
Looks good to me. I couldn't run the tests locally since the larger test data isn't included though! 

One thing to maybe consider is, similar to your comment on kallistobustools/ref, maybe move the `protocol` variable from being an input val to being a parameter in options? Since it's not an input file I guess based on the guidelines it technically shouldn't be an input channel?

KevinMenden:
Yeah that is something I thought about, not sure what's the best option 🤔 
The main reason I put it here is because this is not an optional argument, but something that always has to be provided. I could instead default to some protocol and then check if a new protocol is defines in `options.args` and then change it there .... might be the better option indeed. Downside would be that you can't define the protocol on the command line this way.

KevinMenden:
Alright I thought I can now finally merge this but alevin doesn't work with the 10X genomics data ... presumably because it's only 100 reads and that somehow leads to a segmentation fault? No idea ... it works when I use the full dataset. But both kallisto and STAR are happy to eat the test dataset .... :angry:

KevinMenden:
Closing this now and will create new later.

PR #405, opened 2021-04-06
Added module kb/ref

FloWuenne:
## PR checklist

Closes #358  

This PR adds the module for kb/ref, which is required to create index files for the single-cell processing version of kallisto (kallisto|bustools or kb). The existing reference data, unfortunately, does not work with this at the moment, due to missing entries in the gtf. Test files for this module have currently been added to: `tests/data/delete_me/kb`.

This version can run 3 of the 4 index workflows (standard, lamanno, nucleus). The only workflow not implemented is `--workflow kite`, as this only uses one input file (feature.tsv), that is different from the 2 input files (fasta,gtf) that are required to be supplied for all other options and I did not know how to achieve this yet.

- [x] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [x] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [x] If necessary, include test data in your PR.
- [x] Remove all TODO statements.
- [x] Emit the `<SOFTWARE>.version.txt` file.
- [x] Follow the naming conventions.
- [x] Follow the parameters requirements.
- [x] Follow the input/output options guidelines.
- [ ] Add a resource `label`
- [x] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
    - [x] `PROFILE=docker pytest --tag <MODULE> --symlink --keep-workflow-wd`
    - [ ] `PROFILE=singularity pytest --tag <MODULE> --symlink --keep-workflow-wd`
    - [ ] `PROFILE=conda pytest --tag <MODULE> --symlink --keep-workflow-wd`

PR #408, opened 2021-04-07
Added kallistobustools/ref

### LINKED ISSUE #358: new module: kb/ref
I think it would be good to have a module for [kb/ref](https://www.kallistobus.tools/kb_usage/kb_ref/)

kb-ref is slightly different from kallisto/index (#357), as it can produce specific references for scRNA-seq, that kallisto/index can't.

- [X] This module does not exist yet with the [`nf-core modules list`](https://github.com/nf-core/tools#list-modules) command
- [X] There is no [open pull request](https://github.com/nf-core/modules/pulls) for this module
- [X] There is no [open issue](https://github.com/nf-core/modules/issues) for this module

FloWuenne:
## PR checklist

Closes #358  

New PR instead of kb/ref (#405). Renamed module to `kallistobustools/ref` based on suggestion by @drpatelh.

This PR adds the module for kb/ref, which is required to create index files for the single-cell processing version of kallisto (kallisto|bustools or kb). The existing reference data, unfortunately, does not work with this at the moment, due to missing entries in the gtf. Test files for this module have currently been added to: `tests/data/delete_me/kb`.

This version can run 3 of the 4 index workflows (standard, lamanno, nucleus). The only workflow not implemented is `--workflow kite`, as this only uses one input file (feature.tsv), that is different from the 2 input files (fasta,gtf) that are required to be supplied for all other options and I did not know how to achieve this yet.

- [ ] Add a resource `label`


KevinMenden (approved):
Nice, looks good to me!

The only thing I'm wondering is about the `workflow` variable stored in the meta map :thinking: 
Is this stored in there because this can be different from sample to sample in a pipeline? Otherwise it might be easier to read if that's an additional input variable that you pass to the tool.

software/kallistobustools/ref/main.nf
```diff
@@ -0,0 +1,69 @@
...
+    path "*cdna.fa"                           , optional:false  ,   emit: cdna
+    path "*intron.fa"                         , optional:true   ,   emit: intron
+    path "*cdna_t2c.txt"                      , optional:true   ,   emit: cdna_t2c
+    path "*intron_t2c.txt"                    , optional:true   ,   emit: intron_t2c
+    path "*.version.txt"                      , optional:false  ,   emit: version
+
+    script:
+    def software = getSoftwareName(task.process)
+    def prefix   = options.suffix ? "${meta.id}${options.suffix}" : "${meta.id}"
+    if(meta.workflow == "standard"){
```
FloWuenne, line 37: Thanks again also for this review @KevinMenden ! I agree with you about the workflow variable. In the kallistobustools/count module, I put that into options. My problem here was that I didn't know how else to solve the if else statement that determines which output files get returned depending on the workflow... If anyone knows how to retain this if but move workflow from meta to options, please submit a change for this!

FloWuenne:
@KevinMenden I moved the workflow from meta to options.args and then defined a variable called workflow. This works but I don't know if this is the proper way to go? No other arguments need to be passed to options.args at the moment, so that works. I didn't know how to add a named value to the options that are directly called workflow...

KevinMenden:
HI @FloWuenne , thanks! The `options.args` is reserved for command line arguments that are passed directly to the tool, so it cannot be used like this. The way to do this here would be to add an additional input channel (e.g. `val workflow`) that is either true or false, and then use that to evaluate the if/else loop.
I ran into a similar issue with the `quast` module, maybe that helps:
https://github.com/nf-core/modules/blob/master/software/quast/main.nf

FloWuenne:
Thanks, @KevinMenden ! I can implement as you suggested, I am just wondering, the --workflow is also not optional and should be set with every run. Is that in the proper scope for input variables then? It's also a bit confusing since it goes a bit against the guidelines of only having files be input and output variables no?

software/kallistobustools/ref/main.nf
```diff
@@ -0,0 +1,68 @@
...
+    conda (params.enable_conda ? "bioconda::kb-python=0.25.1" : null)
+    if (workflow.containerEngine == 'singularity' && !params.singularity_pull_docker_container) {
+        container "https://depot.galaxyproject.org/singularity/kb-python:0.25.1--py_0"
+    } else {
+        container "quay.io/biocontainers/kb-python:0.25.1--py_0"
+    }
+
+    input:
+    tuple   val(meta), path(fasta)
+    path    gtf
```
drpatelh, line 23:
```suggestion
    tuple val(meta), path(fasta)
    path  gtf
```

software/kallistobustools/ref/main.nf
```diff
@@ -0,0 +1,68 @@
...
+
+    conda (params.enable_conda ? "bioconda::kb-python=0.25.1" : null)
+    if (workflow.containerEngine == 'singularity' && !params.singularity_pull_docker_container) {
+        container "https://depot.galaxyproject.org/singularity/kb-python:0.25.1--py_0"
+    } else {
+        container "quay.io/biocontainers/kb-python:0.25.1--py_0"
+    }
+
+    input:
+    tuple   val(meta), path(fasta)
```
drpatelh, line 22: Do we need `meta` here? i.e. will this index be created for all samples in which case we don't need to define `meta`

software/kallistobustools/ref/main.nf
```diff
@@ -0,0 +1,68 @@
...
+    path    gtf
+
+    output:
+    tuple val(meta), path("*_kb_ref_out.idx") , optional:false  ,   emit: kb_ref_idx
+    path "*t2g.txt"                           , optional:false  ,   emit: t2g
+    path "*cdna.fa"                           , optional:false  ,   emit: cdna
+    path "*intron.fa"                         , optional:true   ,   emit: intron
+    path "*cdna_t2c.txt"                      , optional:true   ,   emit: cdna_t2c
+    path "*intron_t2c.txt"                    , optional:true   ,   emit: intron_t2c
+    path "*.version.txt"                      , optional:false  ,   emit: version
```
drpatelh, line 32:
```suggestion
    path "*_kb_ref_out.idx", emit: index
    path "*t2g.txt"        , emit: t2g
    path "*cdna.fa"        , emit: cdna
    path "*.version.txt"   , emit: version
    path "*intron.fa"      , optional:true, emit: intron
    path "*cdna_t2c.txt"   , optional:true, emit: cdna_t2c
    path "*intron_t2c.txt" , optional:true, emit: intron_t2c
```

software/kallistobustools/ref/main.nf
```diff
@@ -0,0 +1,68 @@
...
+
+    script:
+    def software = getSoftwareName(task.process)
+    def prefix   = options.suffix ? "${meta.id}${options.suffix}" : "${meta.id}"
+    def workflow = "${options.args}"
+    if(workflow == "standard"){
```
drpatelh, line 38:
```suggestion
    if (workflow == "standard") {
```

drpatelh: You can only have one `options.args` for this tool? and that option is the name of the `workflow`? Doesn't seem right?

software/kallistobustools/ref/main.nf
```diff
@@ -0,0 +1,68 @@
...
+        """
+        kb \\
+        ref \\
+        -i ${prefix}_kb_ref_out.idx \\
+        -g t2g.txt \\
+        -f1 cdna.fa \\
+        --workflow ${workflow} \\
+        $fasta \\
+        $gtf
```
drpatelh, line 47:
```suggestion
        kb \\
            ref \\
            -i ${prefix}_kb_ref_out.idx \\
            -g t2g.txt \\
            -f1 cdna.fa \\
            --workflow standard \\
            $fasta \\
            $gtf
```

software/kallistobustools/ref/main.nf
```diff
@@ -0,0 +1,68 @@
...
+    path "*t2g.txt"                           , optional:false  ,   emit: t2g
+    path "*cdna.fa"                           , optional:false  ,   emit: cdna
+    path "*intron.fa"                         , optional:true   ,   emit: intron
+    path "*cdna_t2c.txt"                      , optional:true   ,   emit: cdna_t2c
+    path "*intron_t2c.txt"                    , optional:true   ,   emit: intron_t2c
+    path "*.version.txt"                      , optional:false  ,   emit: version
+
+    script:
+    def software = getSoftwareName(task.process)
+    def prefix   = options.suffix ? "${meta.id}${options.suffix}" : "${meta.id}"
```
drpatelh, line 36:
```suggestion
```

drpatelh: Delete this line if not using `meta.map`

software/kallistobustools/ref/main.nf
```diff
@@ -0,0 +1,68 @@
...
+        ref \\
+        -i ${prefix}_kb_ref_out.idx \\
+        -g t2g.txt \\
+        -f1 cdna.fa \\
+        -f2 intron.fa \\
+        -c1 cdna_t2c.txt \\
+        -c2 intron_t2c.txt \\
+        --workflow ${workflow} \\
+        $fasta \\
+        $gtf
```
drpatelh, line 63:
```suggestion
        kb \\
            ref \\
            -i ${prefix}_kb_ref_out.idx \\
            -g t2g.txt \\
            -f1 cdna.fa \\
            -f2 intron.fa \\
            -c1 cdna_t2c.txt \\
            -c2 intron_t2c.txt \\
            --workflow $workflow \\
            $fasta \\
            $gtf
```

software/kallistobustools/ref/meta.yml
```diff
@@ -0,0 +1,62 @@
...
+      tool_dev_url: https://github.com/pachterlab/kb_python
+      doi: "https://doi.org/10.22002/D1.1876"
+      licence: MIT License
+
+input:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. [ id:'test', --workflow ]
```
drpatelh, line 20:
```suggestion
```

drpatelh: ?

software/kallistobustools/ref/meta.yml
```diff
@@ -0,0 +1,62 @@
...
+  - fasta:
+      type: file
+      description: Genomic DNA fasta file
+      pattern: "*.{fasta,fasta.gz}"
+
```
drpatelh, line 25: Add entry for `gtf` as `input` too.

drpatelh (changes requested):
First pass changes @FloWuenne

KevinMenden:
Ah right oversaw that this is an index, then you can omit the `meta` completely of course ...

Regarding the other issue, I think all necessary commands should be set somehow in the module (so that you don't depend on anything in `options.args`, because that will usually come from the `modules.config` file, which is not part of `nf-core/modules`)
So the `--workflow` variable must depend on some input features, right? Can you parse them with some groovy logic from the input? Otherwise I think it would be okay to pass them to the module as input. Alternative would be to have them in the `modules.config`.... @drpatelh what do you think?

software/kallistobustools/ref/main.nf
FloWuenne: No you are right, since this is an index I guess meta is not needed! It's good to know for me for any future modules. Might also be useful to mention this somewhere in the docs for new module developers that they can omit meta when creating the module... (Maybe it's written somewhere and I missed it!)

FloWuenne:
Ok I have incorporated @drpatelh changes to remove the `meta` parameter and I have added the gtf channel to the yaml. Now it remains to move the --workflow argument to move. I believe the -x (technology) flag, should probably be treated the same as workflow. Right now I called it in options, but this probably needs to be changed. @drpatelh @KevinMenden , what do you think? These 2 parameters: --worfklow and -x (defining the sequencing technology used), which should be called with every run of ref and count, where should they go in these two modules?

software/kallistobustools/ref/main.nf
```diff
@@ -0,0 +1,67 @@
+// Import generic module functions
+include { initOptions; saveFiles; getSoftwareName } from './functions'
+
+params.options = [:]
+options        = initOptions(params.options)
+
+process KALLISTOBUSTOOLS_REF {
+    tag "$meta.id"
```
drpatelh, line 8:
```suggestion
    tag "$fasta"
```

software/kallistobustools/ref/main.nf
```diff
@@ -0,0 +1,67 @@
...
+    label 'process_medium'
+    publishDir "${params.outdir}",
+        mode: params.publish_dir_mode,
+        saveAs: { filename -> saveFiles(filename:filename, options:params.options, publish_dir:getSoftwareName(task.process), publish_id:meta.id) }
```
drpatelh, line 12:
```suggestion
        saveAs: { filename -> saveFiles(filename:filename, options:params.options, publish_dir:getSoftwareName(task.process), publish_id:'') }
```

tests/software/kallistobustools/ref/main.nf
```diff
@@ -0,0 +1,40 @@
...
+nextflow.enable.dsl = 2
+
+include { KALLISTOBUSTOOLS_REF } from '../../../../software/kallistobustools/ref/main.nf' addParams( options: [args:"standard"] )
+include { KALLISTOBUSTOOLS_REF as KALLISTOBUSTOOLS_REF_LAMANNO } from '../../../../software/kallistobustools/ref/main.nf' addParams( options: [args:"lamanno"] )
+include { KALLISTOBUSTOOLS_REF as KALLISTOBUSTOOLS_REF_NUCLEUS } from '../../../../software/kallistobustools/ref/main.nf' addParams( options: [args:"nucleus"] )
+
+workflow test_kallistobustools_ref_standard {
+
+    input = [ [id:'test.standard'], // meta map
+        [file("${launchDir}/tests/data/delete_me/kallistobustools/GRCm39.chr19_100k.fa.gz", checkIfExists: true)]
```
drpatelh, line 12: We need to remember to actually delete all of this custom test data. I am wondering whether its better to just add it to another repo somewhere so we don't bloat the git history of nf-core/modules 🤔 We should maybe try and get that human data PR in as soon as (Ping @FriederikeHanssen)

tests/software/kallistobustools/ref/main.nf
```diff
@@ -0,0 +1,40 @@
...
+}
+
+workflow test_kallistobustools_ref_lamanno {
+
+    input = [ [id:'test.lamanno'], // meta map
+        [file("${launchDir}/tests/data/delete_me/kallistobustools/GRCm39.chr19_100k.fa.gz", checkIfExists: true)]
+        ]
+
+    gtf      = file("${launchDir}/tests/data/delete_me/kallistobustools/gencode.VM26.chr19_10k.gtf.gz", checkIfExists: true) 
+
```
drpatelh, line 27:
```suggestion
    input = [ [id:'test.lamanno'], // meta map
              [file("${launchDir}/tests/data/delete_me/kallistobustools/GRCm39.chr19_100k.fa.gz", checkIfExists: true)]
            ]
    gtf   = file("${launchDir}/tests/data/delete_me/kallistobustools/gencode.VM26.chr19_10k.gtf.gz", checkIfExists: true) 
```

tests/software/kallistobustools/ref/main.nf
```diff
@@ -0,0 +1,40 @@
...
+    KALLISTOBUSTOOLS_REF_LAMANNO( input, gtf)
+}
+
+workflow test_kallistobustools_ref_nucleus {
+
+    input = [ [id:'test.nucleus'], // meta map
+        [file("${launchDir}/tests/data/delete_me/kallistobustools/GRCm39.chr19_100k.fa.gz", checkIfExists: true)]
+        ]
+
+    gtf      = file("${launchDir}/tests/data/delete_me/kallistobustools/gencode.VM26.chr19_10k.gtf.gz", checkIfExists: true)
```
drpatelh, line 37:
```suggestion
    input = [ [id:'test.nucleus'], // meta map
              [file("${launchDir}/tests/data/delete_me/kallistobustools/GRCm39.chr19_100k.fa.gz", checkIfExists: true)]
            ]
    gtf   = file("${launchDir}/tests/data/delete_me/kallistobustools/gencode.VM26.chr19_10k.gtf.gz", checkIfExists: true) 
```

tests/software/kallistobustools/ref/main.nf
FloWuenne: I have actually added some of this test-data to nf-core test-datasets:
https://github.com/nf-core/test-datasets/pull/231

I just didn't know how to correctly use data that is part of the test-datasets repo in these tests without having them inside the modules repo?!

drpatelh:
Thanks @FloWuenne ! If `--workflow and -x` are mandatory parameters they should be added as additional inputs to the module.

FloWuenne:
Ok I made the workflow parameter an input value that needs to be supplied. I removed all remnants of the meta value and finally, I fixed the tests... (hopefully). 

The only thing missing from this module at the moment is the --workflow kite option, because this one requires only 1 input file (a table file with barcode sequences and IDs, instead of the fasta and the gtf). I didn't know how to implement an optional input file, that only gets used with 1 parameter (worfklow == kite) and then allows to have the other input channels (fasta and gtf) empty...

tests/software/kallistobustools/ref/main.nf
KevinMenden: Ah good point ... we could either make a branch in nf-core/test-datasets for temporary test data or make a whole repo that we could monitor and later delete if necessary 🤔 
@FloWuenne if you have a look at the `test.config` profiles (in the conf folder) of any published pipeline, it shows you how you can just link to a file somewhere on github.

KevinMenden:
Yeah these cases aren't pretty and there is no "nice" solution to it. The current workaround would be to define that input for all cases, and just ignore it if you don't need it. So in the case where you don't need a gtf or fasta file, you would supply a dummy file and just ignore it in the module. So this would be really just a channel pointing to any useless file (e.g. `dummy.txt`). Don't remember if this also works if the file doesn't actually exist 🤔

FloWuenne:
Given that the --workflow kite option is only for one specific case (Feature barcoding like CITE-seq or CRISPR-seq), might be ok to just add this at a later point, when we actually need it in the scRNA-seq pipeline?!
The standard scRNA-seq indexes we need can be produced with this version!

tests/config/pytest_software.yml
```diff
@@ -285,6 +285,10 @@ kallisto_index:
   - software/kallisto/index/**
   - tests/software/kallisto/index/**
 
+kallistobustools_ref:
```
drpatelh, line 288:
```suggestion
kallistobustools/ref:
```

tests/software/kallistobustools/ref/test.yml
```diff
@@ -0,0 +1,41 @@
+- name: kallistobustools ref test_kallistobustools_ref_standard
+  command: nextflow run tests/software/kallistobustools/ref -entry test_kallistobustools_ref_standard -c tests/config/nextflow.config
+  tags:
+    - kallistobustools
+    - kallistobustools_ref_standard
+    - kallistobustools_ref
```
drpatelh, line 6:
```suggestion
    - kallistobustools/ref
```

drpatelh (approved):
Thanks @FloWuenne ! Looks awesome. Would have been nice to have pulled in the test data from a remote repo using "raw" Github links like [here](https://github.com/nf-core/viralrecon/blob/4556fce27b54952eb065adc9c4b7511aaaa7d402/conf/test.config#L20). NF will understand them and stage them as required for the tests. Maybe one for next time. But be great if you can delete the data eventually when the proper human data is added.

tests/software/kallistobustools/ref/test.yml
```diff
@@ -0,0 +1,41 @@
...
+      md5sum: c0ffe9554cb6fdcc7cbb4a1285ca8f29
+    - path: output/kallistobustools/t2g.txt
+      md5sum: 3089d303d9cd1a39a27969eca2dcaba7
+
+- name: kallistobustools ref test_kallistobustools_ref_nucleus
+  command: nextflow run tests/software/kallistobustools/ref -entry test_kallistobustools_ref_nucleus -c tests/config/nextflow.config
+  tags:
+    - kallistobustools_ref_nucleus
+    - kallistobustools
+    - kallistobustools_ref
```
drpatelh, line 34:
```suggestion
    - kallistobustools/ref
```

tests/software/kallistobustools/ref/test.yml
```diff
@@ -0,0 +1,41 @@
...
+      md5sum: c0ffe9554cb6fdcc7cbb4a1285ca8f29
+    - path: output/kallistobustools/t2g.txt
+      md5sum: 3089d303d9cd1a39a27969eca2dcaba7
+
+- name: kallistobustools ref test_kallistobustools_ref_lamanno
+  command: nextflow run tests/software/kallistobustools/ref -entry test_kallistobustools_ref_lamanno -c tests/config/nextflow.config
+  tags:
+    - kallistobustools
+    - kallistobustools_ref_lamanno
+    - kallistobustools_ref
```
drpatelh, line 20:
```suggestion
    - kallistobustools/ref
```

PR #428, opened 2021-04-10
Add default publish_dir for index processes

drpatelh:
This was missed on the auto-update script I wrote yesterday but I noticed it when testing after re-installing all  nf-core/modules for the viralrecon pipeline [here](https://github.com/nf-core/viralrecon/pull/172).


drpatelh:
Fairly uncontroversial changes. Would like to install these modules into pipelines so I can get on with something else. Will self-merge.

PR #432, opened 2021-04-12
Move test data to nf-core/test-datasets

KevinMenden:
This PR switches the path to the test data to https://github.com/nf-core/test-datasets/tree/modules/data 
as a first step to stop hosting it on `nf-core/modules`

If all tests work, we can remove the data on `nf-core/modules`, but I first want to check whether the tests work.

We will also have to
* update the modules documentation for adding test data
* update the README (and copy it over to `nf-core/test-datasets`)
* draft some documentation about how to handle imperfect test data (i.e. temporary data only used for releasing a module)


KevinMenden:
MultiQC test works, Fastqc worked offline, too. So I would assume it should work for all tests because the data is exactly the same.

drpatelh (approved):
LGTM!!

Issue #430, opened 2021-04-12
Finish unfinished modules


KevinMenden:
There are several modules that have been added very early to `nf-core/modules` and have been a bit under the radar since, leading to some of them lacking tests, `meta.yml` files etc. It would be nice to finish them so we can have a presentable linting result for the main modules repo :-) 

I would suggest to approach this as usual, meaning I will make a checkbox-list here and everybody that works one any of those modules should edit this issue to indicate which module is currently being worked on or finished.

Here goes the list:

- [ ] deeptools/computematrix                                                                                            
- [ ] deeptools/plotfingerprint                                                                                     
- [ ] deeptools/plotheatmap                                                                   
- [ ] deeptools/plotprofile
- [ ]  hisat2/build
- [ ] macs2/callpeak
- [ ] phantompeakqualtools
- [ ] qualimap/rnaseq
- [ ] rsem/calculateexpression
- [ ] rsem/preparereference
- [ ] rseqc/bamstat
- [ ] rseqc/inferexperiment
- [ ] rseqc/innerdistance
- [ ] rseqc/junctionannotation
- [ ] rseqc/junctionsaturation
- [ ] rseqc/readdistribution
- [ ] rseqc/readduplication
- [ ] sortmerna
- [ ] umitools/dedup
- [ ] umitools/extract

KevinMenden:
For rseqc:
https://github.com/nf-core/modules/pull/439

Need to wait for the human data to fix `junctionannotation` and `junctionsaturation`

JoseEspinosa:
For deeptools:
https://github.com/nf-core/modules/pull/442

klkeys:
for `umitools`: #860

PR #439, opened 2021-04-13
Finish rseqc modules

KevinMenden:
This finishes most of the `rseqc` subtools (see https://github.com/nf-core/modules/issues/430)

I could not finish `junctionannotation` and `junctionsaturation` yet because they don't work with viral data (makes sense, they need introns).

Also I don't know why I keep having these old commits in my PRs 🤔  maybe I'll delete my repo and fork again after this .... doesn't really matter that much since it gets squashed anyway but still.


drpatelh (approved):
🤩

PR #440, opened 2021-04-13
remove deprecated test tags

KevinMenden:
Remove all the deprecated test tags.

### DIFF
--- a/software/gatk4/applybqsr/main.nf
+++ b/software/gatk4/applybqsr/main.nf
@@ -19,7 +19,7 @@ process GATK4_APPLYBQSR {
     }
 
     input:
-    tuple val(meta), path(bam), path(bqsr_table)
+    tuple val(meta), path(bam), path(bai), path(bqsr_table)
     path fasta
     path fastaidx
     path dict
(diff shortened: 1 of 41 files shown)


JoseEspinosa (approved):
LGTM

KevinMenden:
Seems like I broke some tests though 🤔

KevinMenden:
thanks for the review @JoseEspinosa !

JoseEspinosa:
> thanks for the review @JoseEspinosa !

Was just about to highlight it... Seems that relaunching of in principle working modules always come with surprises :cry:

KevinMenden:
Damnit 😠  well I'll try to fix those then ....

KevinMenden:
Okay I think I fixed most tests.
`star/align` and `kallistobustools/ref` are still failing because they are missing the test data ... I've put them on my list, but @FloWuenne is working on that anyway and already has some test data I think, we just have to switch the paths at some point. Might be something for another PR.
STAR is basically waiting for human data as well.

KevinMenden:
ah I forgot `yara/mapper`

KevinMenden:
okay now unicycler is failing too, no idea what happened there ... :/ aaah I thought I just change the tags! 🤦

PR #499, opened 2021-05-17
Fix number of cpus for modules with piped tools

### LINKED ISSUE #498: Fix number of cpus for modules with piped tools
As reported by @lindenb on [this](https://twitter.com/yokofakun/status/1392929991822921733?s=20) tweet, some piped commands should share the number of available CPUs instead of taking each of them all the CPUs. As proposed by @ewels  on the modules slack channel, the easiest solution is to assign the CPUs to each piped command with something such as `floor(task.cpus /2)`.

Affected modules:

[ ] `bowtie/align`
[ ] `bowtie2/align`
[ ] `bwa/mem`
[ ] `bwamem2/mem`
[ ] `bwameth/align`

JoseEspinosa:
Split the number of CPUs in modules with two tools in a piped command and each of them with an explicit declaration of the CPUs to be used. See #498 

## PR checklist

Closes #498

### DIFF
--- a/software/bowtie/align/main.nf
+++ b/software/bowtie/align/main.nf
@@ -29,6 +29,7 @@ process BOWTIE_ALIGN {
     tuple val(meta), path('*fastq.gz'), optional:true, emit: fastq
 
     script:
+    def split_cpus = Math.floor(task.cpus/2)
     def software  = getSoftwareName(task.process)
     def prefix    = options.suffix ? "${meta.id}${options.suffix}" : "${meta.id}"
     def unaligned = params.save_unaligned ? "--un ${prefix}.unmapped.fastq" : ''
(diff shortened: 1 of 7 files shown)


ewels (approved):
LGTM 👌 Just need to fix the tests..

maxulysse (approved):
<3

JoseEspinosa:
> LGTM 👌 Just need to fix the tests..

These are the infamous `conda` tests... I can not also resolve the environment locally, so I guess that as usual, since `docker` and `singularity` are passing we can ignore it... 😓

ramprasadn:
Sorry for resurrecting this thread almost four months after it was closed, but I wonder if splitting by 2 affects the performance of the aligner a bit. 

Looks like when we have a piped command, its almost always an aligner paired with samtools view, and samtools view is not a resource intensive process. In the case, the aligner should be getting more resources instead of samtools, right? What do you guys think?

PR #513, opened 2021-05-25
Metaphlan3 run

MGordon09:
## PR checklist

Closes #XXX 

- [x] Add a resource `label`


MGordon09:
Hi @drpatelh, hope you are keeping well. I have created a module for metaphlan3 and would greatly appreciate your input and feedback (when you have time of course, there is no rush  :)  )  

This module has given me a lot more trouble than the previous one! My biggest concern has been how to choose the best approach for making the metaphlan marker gene database available (the indexed files are  nearly 2GB in size), so I didn't think it was appropriate to include in the PR. (I have left a link in comments in the  `main.nf` script where this is available and will remove after )

I have also included a small folder with test data for the module. I have tried to use what is already available  in `test_data.config`.

The module is running  when I test on the HPC and I would really appreciate your advice on how best to make this available to the nf-core community. 

Thank you again for your guidance, patience and all your hard work!

Martin

software/metaphlan3/run/meta.yml
```diff
@@ -0,0 +1,57 @@
...
+      description: List of input FastQ files of size 1 and 2 for single-end and paired-end data, respectively.
+  - sam:
+      type: file
+      description: Bowtie2-produced SAM files containing read alignments to the MetaPHlAn marker database
+  - fasta:
+      type: file
+      description: Input files in FASTA format
+  - bowtie2out:
+      type: file
+      description: Intermediate Bowtie2 output produced from mapping the metagenome against the MetaPHlAn marker database ( not compatible with `bowtie2out` files generated with MetaPhlAn versions below 3 )
```
KevinMenden, line 35: I think it would be better to list all the input options under on entry called `input`, as that's how you called it in the `main.nf` script - this way it is clearer what is meant. Then you could add a pattern to indicated the different input options and just mention that different file inputs are possible in the description.

software/metaphlan3/run/meta.yml
```diff
@@ -0,0 +1,57 @@
...
+
+output:
```
KevinMenden, line 37: this is missing the (optional) bt2out output

tests/software/metaphlan3/run/main.nf
```diff
@@ -0,0 +1,59 @@
...
+include { SAMTOOLS_VIEW  } from '../../../../software/samtools/view/main.nf'  addParams( options: ['suffix': '.sam'] ) //metaphlan can't take bam input
+include { METAPHLAN3_RUN } from '../../../../software/metaphlan3/run/main.nf' addParams( options: [ 'args':'--index mpa_v30_CHOCOPhlAn_201901 --add_viruses --bt2_ps very-sensitive-local' ] ) //database version (latest, if not supplied attempts database autodownload and run fails)
+
+workflow test_metaphlan3_single_end {
+
+    input = [ [ id:'test', single_end:true ], // meta map
+              [ file(params.test_data['sarscov2']['illumina']['test_1_fastq_gz'], checkIfExists: true) ]
+            ]
+
+    db    = channel.fromPath('https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/delete_me/metaphlan_database.tar.gz', type: 'dir', checkIfExists: true)
```
KevinMenden, line 15: Fine for now, but since this looks like a good test dataset, we can actually add this to the `genomics` folder I'd say. Maybe something to discuss on slack :)

software/metaphlan3/run/main.nf
```diff
@@ -0,0 +1,51 @@
...
+    tuple val(meta), path('*.bowtie2out.txt'), optional:true, emit: bt2out
+    path "*.version.txt"                     ,                emit: version
+
+    script:
+    def software    = getSoftwareName(task.process)
+    def prefix      = options.suffix ? "${meta.id}${options.suffix}" : "${meta.id}"
+    //contains.() method to handle sam/bam and uncompressed files.. Any better suggestions would be greatly appreciated!!
+    def input_type  = ("$input".endsWith(".fastq.gz")) ? "--input_type fastq" :  ("$input".contains(".fasta")) ? "--input_type fasta" : ("$input".endsWith(".bowtie2out.txt")) ? "--input_type bowtie2out" : "--input_type sam" // .contains() to accomodate fasta non-compressed input also?
+    def input_data  = ("$input_type".contains("fastq")) && !meta.single_end ? "${input[0]},${input[1]}" : "$input"
+    def bowtie2_out = "$input_type" == "--input_type bowtie2out" || "$input_type" == "--input_type sam" ? '' : "--bowtie2out ${prefix}.bowtie2out.txt" // no intermediate alignment files produced for sam & bowtie2out input
```
KevinMenden, line 37: It would also be possible to add the `input_type` as a parameter to the input, but I have to say that I like the way you currently have more, as it makes it easier to use. So fine from my side 👍

KevinMenden (commented):
Hi @MGordon09 , this looks really good to me, just a few suggestions.

And one more general thing - do you plan to add any other metaphlan3 related modules? Because if not, there is no need to create the subtool `metaphlan3/run` but it would make more sense to just use `metaphlan3` as a module here. I'm asking because you're not really calling a `run` command in `main.nf` but just plainly run `metaphlan3`.

software/metaphlan3/run/meta.yml
MGordon09: Thanks for the spot, have updated

software/metaphlan3/run/meta.yml
MGordon09: Sounds good, have updated

software/metaphlan3/run/main.nf
MGordon09: thanks, sounds good!

tests/software/metaphlan3/run/main.nf
MGordon09: I think the metaphlan marker database I created may be better suited as a draft at the moment as it needs a little work.. It seems to be 'working' fine (input successfully aligned and output is generated) but it struggles to fully classify the test data (tool authors suggested this could be related to low coverage of input data when I asked about this..) I think it is ok at the moment for testing but I hope to improve it (great shout; I'm sure the nf-core folks will have some good ideas on building custom metaphlan DB!)

MGordon09:
Hi @KevinMenden 

Thank you for the detailed feedback and suggestions, I have implemented suggested changes and addressed comments above.

Regarding metaphlan3 vs metaphlan3/run: I was considering creating a module to download and index the metaphlan marker gene database which I was going to call `metaphlan3/build` (I have not had the chance to build this yet, and wasn't sure if it was sensible to build a module like this or allow the user to acquire the database by their own means (it's quite large so perhaps leaving it to user to source makes more sense). I looked at naming convention of kracken2 module for some direction here, but am happy to change this to simply metaphlan3 if suits more?)

Thanks for all your helpful feedback!

KevinMenden:
Hi @MGordon09,

that sounds like a reasonable subtool/submodule for metaphlan3. Would it also be using metaphlan3 for the download, or would that be just a `curl` command or something similar?
You're right, I didn't know about the `kraken2/run` module ... which doesn't have any submodules though.
I think I will start the discussion on slack to get some more feedback from the others :)

MGordon09:
Hi @KevinMenden 

Metaphlan3 can autodownload and index the database, but I previously had some issues with the metaphlan3 conda env.. I found using `wget` and `bowtie2-build`  more straightforward,  so intended to go this route if works for you guys?

MGordon09:
Just caught up on the discussion on slack :) I should add that I don't think there is a separate metaphlan commandto exclusively download the database (it tries to autodownload if db path is not supplied), so it is no problem to stick with the standard naming convention ( `metaphlan3`).

KevinMenden:
Okay great, in that case if you don't mind it would be nice to go with the standard naming :) 

Alternatively you could to a `metaphlan3/metaphlan3` module if you want to add the download module later.

PR #567, opened 2021-07-06
Rename software/ directory to modules/ to re-organise module structure

drpatelh:
* See discussion [here](https://github.com/nf-core/tools/issues/1159)
* Renamed `pytest_software.yml` -> `pytest_modules.yml`


grst (approved):
I suggest we just merge this as

- there's no way to meaningfully review 1.4k files
- the changes have been performed automatically by @drpatelh's script
- The stuff that wasn't changed, but should have doesn't show up in the diff anyway. 

What's wrong with the linting, though?

KevinMenden:
Yup I agree with @grst that we should just merge it. I tried looking at it a bit (what I saw looked good), but GitHub can't really handle that kind of a PR.

The linting is failing because the linter is confused where all these files are now. We'll have to adapt every single modules command in nf-core/tools after merging this, and then we just lint the repo manually to make sure everything is in place. If not, we should be able to find it this way, too.

PR #583, opened 2021-07-16
Add --paired flag to umitools dedup and bump to 1.1.2

drpatelh:
Fix module as a result of https://github.com/nf-core/rnaseq/issues/668


grst (approved):
Awesome!

PR #615, opened 2021-07-23
Specify more guidelines on input channels

jfy133:
This improves documentation on what and how input channels should be specified


README.md
```diff
@@ -18,17 +18,27 @@ A repository for hosting [Nextflow DSL2](https://www.nextflow.io/docs/latest/dsl
...
-    - [Test data](#test-data)
-    - [Running tests manually](#running-tests-manually)
-    - [Uploading to `nf-core/modules`](#uploading-to-nf-coremodules)
-    - [Guidelines](#guidelines)
-- [Terminology](#terminology)
-- [Help](#help)
-- [Citation](#citation)
+
+- [!nf-core/modules](#img-srcdocsimagesnfcore-modules_logopng-altnf-coremodules)
+    - [Table of contents](#table-of-contents)
```
drpatelh, line 23: Why do we need these? The top one looks broken?

README.md
```diff
@@ -432,6 +442,9 @@ using a combination of `bwa` and `samtools` to output a BAM file instead of a SA
 #### Module parameters
 
 - A module file SHOULD only define input and output files as command-line parameters to be executed within the process.
+    - All possible input files MUST have a defined input channel declaration (see [Input/output options](#input-output-options) for dealing with optional input files)
+    - Directly associated auxiliary files to an input file CAN be defined within the same input channel alongside the main input channel  (e.g. [BAM and BAI](https://github.com/nf-core/modules/blob/e937c7950af70930d1f34bb961403d9d2aa81c7d/modules/samtools/flagstat/main.nf#L22)).
+    - Other generic auxiliary files used across different input files (e.g. common reference sequences) CAN be defined using a dedicated input channel (e.g. [reference files](https://github.com/nf-core/modules/blob/3cabc95d0ed8a5a4e07b8f9b1d1f7ff9a70f61e1/modules/bwa/mem/main.nf#L21-L23))
```
drpatelh, line 447: Nice!

drpatelh: Think this should go in the `Input/output options` section though as it relates to handling those features of the module.

drpatelh: `Module parameters` are for things like `options.args` / command-line parameters used to run the tool

grst, line 447: I'm being a bit nitpicky here, but RFC2119 doesn't define `CAN`. Unless we decide to ditch it entirely :shrug: 

```suggestion
    - Directly associated auxiliary files to an input file MAY be defined within the same input channel alongside the main input channel  (e.g. [BAM and BAI](https://github.com/nf-core/modules/blob/e937c7950af70930d1f34bb961403d9d2aa81c7d/modules/samtools/flagstat/main.nf#L22)).
    - Other generic auxiliary files used across different input files (e.g. common reference sequences) MAY be defined using a dedicated input channel (e.g. [reference files](https://github.com/nf-core/modules/blob/3cabc95d0ed8a5a4e07b8f9b1d1f7ff9a70f61e1/modules/bwa/mem/main.nf#L21-L23))
```

README.md
jfy133: I have absolutely no where that has come from!? Maybe an autoformatting thing in VSCode :facepalm:

README.md
```diff
@@ -429,6 +429,15 @@ using a combination of `bwa` and `samtools` to output a BAM file instead of a SA
...
 
+#### Input/output options
+
+- All possible input files MUST have a defined input channel declaration for dealing with optional input files
+    - Directly associated auxiliary files to an input file MAY be defined within the same input channel alongside the main input channel  (e.g. [BAM and BAI](https://github.com/nf-core/modules/blob/e937c7950af70930d1f34bb961403d9d2aa81c7d/modules/samtools/flagstat/main.nf#L22)).
+    - Other generic auxiliary files used across different input files (e.g. common reference sequences) MAY be defined using a dedicated input channel (e.g. [reference files](https://github.com/nf-core/modules/blob/3cabc95d0ed8a5a4e07b8f9b1d1f7ff9a70f61e1/modules/bwa/mem/main.nf#L21-L23))
+
+- Named file extensions MUST be emitted for ALL output channels e.g. `path "*.txt", emit: txt`.
+
+- Optional inputs are not currently supported by Nextflow. However, passing an empty list (`[]`) instead of a file as a module parameter can be used to work around this issue.
```
jfy133, line 440:
```suggestion
- Optional inputs are not currently supported by Nextflow. However, passing an empty list (`[]`) instead of a file as a module parameter can be used to work around this issue.

```

jfy133:
@drpatelh I've moved the new lines to Input/Output, but I've moved that entire section above. To me it makes more sense to refer to input/output channels _before_ module parameters as I personally (at least) always define these before parameters when writing a module, and also I think it makes sense to already say where they are coming from before we get to module parameters

@grst good catch, had the wrong word in my head. Updated to reflect the actually used RFC terminology.

JoseEspinosa (approved):
Very nice @jfy133, although to me params on top of input/output section was more logic since it follows the order of appearance in the script. Anyway, this a subjective question so either way is OK

PR #617, opened 2021-07-25
Notebook modules

grst:
Attempt to implement generic modules for rendering Rmarkdown or jupyter notebooks into HTML reports. 
I'm deviating from the one module - one PR rule here, because I feel the two should be discussed together. 

This is still WIP and there are a few things to be discussed - I will raise the questions as a code review once I feel ready ;) 

## PR checklist

- [x] Add a resource `label`

### DIFF
--- a/modules/jupyternotebook/main.nf
+++ b/modules/jupyternotebook/main.nf
@@ -1,0 +1,79 @@
+// Import generic module functions
+include { initOptions; saveFiles; getSoftwareName } from './functions'
+
+// TODO nf-core: If in doubt look at other nf-core/modules to see how we are doing things! :)
+//               https://github.com/nf-core/modules/tree/master/software
+//               You can also ask for help via your pull request or on the #modules channel on the nf-core Slack workspace:
+//               https://nf-co.re/join
+
+// TODO nf-core: A module file SHOULD only define input and output files as command-line parameters.
+//               All other parameters MUST be provided as a string i.e. "options.args"
+//               where "params.options" is a Groovy Map that MUST be provided via the addParams section of the including workflow.
+//               Any parameters that need to be evaluated in the context of a particular sample
+//               e.g. single-end/paired-end data MUST also be defined and evaluated appropriately.
+// TODO nf-core: Software that can be piped together SHOULD be added to separate module files
+//               unless there is a run-time, storage advantage in implementing in this way
+//               e.g. it's ok to have a single module for bwa to output BAM instead of SAM:
+//                 bwa mem | samtools view -B -T ref.fasta
+// TODO nf-core: Optional inputs are not currently supported by Nextflow. However, using an empty 
+//               list (`[]`) instead of a file can be used to work around this issue.
+
+params.options = [:]
+options        = initOptions(params.options)
+
+process JUPYTERNOTEBOOK {
+    tag "$meta.id"
+    label 'process_low'
+    publishDir "${params.outdir}",
+        mode: params.publish_dir_mode,
+        saveAs: { filename -> saveFiles(filename:filename, options:params.options, publish_dir:getSoftwareName(task.process), meta:meta, publish_by_meta:['id']) }
+
+    // TODO nf-core: List required Conda package(s).
+    //               Software MUST be pinned to channel (i.e. "bioconda"), version (i.e. "1.10").
+    //               For Conda, the build (i.e. "h9402c20_2") must be EXCLUDED to support installation on different operating systems.
+    // TODO nf-core: See section in main README for further information regarding finding and adding container addresses to the section below.
+    conda (params.enable_conda ? "YOUR-TOOL-HERE" : null)
+    if (workflow.containerEngine == 'singularity' && !params.singularity_pull_docker_container) {
+        container "https://depot.galaxyproject.org/singularity/YOUR-TOOL-HERE"
+    } else {
+        container "quay.io/biocontainers/YOUR-TOOL-HERE"
+    }
+
+    input:
+    // TODO nf-core: Where applicable all sample-specific information e.g. "id", "single_end", "read_group"
+    //               MUST be provided as an input via a Groovy Map called "meta".
+    //               This information may not be required in some instances e.g. indexing reference genome files:
+    //               https://github.com/nf-core/modules/blob/master/software/bwa/index/main.nf
+    // TODO nf-core: Where applicable please provide/convert compressed files as input/output
+    //               e.g. "*.fastq.gz" and NOT "*.fastq", "*.bam" and NOT "*.sam" etc.
+    tuple val(meta), path(bam)
+
+    output:
+    // TODO nf-core: Named file extensions MUST be emitted for ALL output channels
+    tuple val(meta), path("*.bam"), emit: bam
+    // TODO nf-core: List additional required output channels/values here
+    path "*.version.txt"          , emit: version
+
+    script:
+    def software = getSoftwareName(task.process)
+    def prefix   = options.suffix ? "${meta.id}${options.suffix}" : "${meta.id}"
+    // TODO nf-core: Where possible, a command MUST be provided to obtain the version number of the software e.g. 1.10
+    //               If the software is unable to output a version number on the command-line then it can be manually specified
+    //               e.g. https://github.com/nf-core/modules/blob/master/software/homer/annotatepeaks/main.nf
+    // TODO nf-core: It MUST be possible to pass additional parameters to the tool as a command-line string via the "$options.args" variable
+    // TODO nf-core: If the tool supports multi-threading then you MUST provide the appropriate parameter
+    //               using the Nextflow "task" variable e.g. "--threads $task.cpus"
+    // TODO nf-core: Please replace the example samtools command below with your module's command
+    // TODO nf-core: Please indent the command appropriately (4 spaces!!) to help with readability ;)
+    """
+    samtools \\
+        sort \\
+        $options.args \\
+        -@ $task.cpus \\
+        -o ${prefix}.bam \\
+        -T $prefix \\
+        $bam
+
+    echo \$(samtools --version 2>&1) | sed 's/^.*samtools //; s/Using.*\$//' > ${software}.version.txt
+    """
+}
--- a/modules/jupyternotebook/main.nf
+++ b/modules/jupyternotebook/main.nf
@@ -1,23 +1,6 @@
 // Import generic module functions
 include { initOptions; saveFiles; getSoftwareName } from './functions'
 
-// TODO nf-core: If in doubt look at other nf-core/modules to see how we are doing things! :)
-//               https://github.com/nf-core/modules/tree/master/software
-//               You can also ask for help via your pull request or on the #modules channel on the nf-core Slack workspace:
-//               https://nf-co.re/join
-
-// TODO nf-core: A module file SHOULD only define input and output files as command-line parameters.
-//               All other parameters MUST be provided as a string i.e. "options.args"
-//               where "params.options" is a Groovy Map that MUST be provided via the addParams section of the including workflow.
-//               Any parameters that need to be evaluated in the context of a particular sample
-//               e.g. single-end/paired-end data MUST also be defined and evaluated appropriately.
-// TODO nf-core: Software that can be piped together SHOULD be added to separate module files
-//               unless there is a run-time, storage advantage in implementing in this way
-//               e.g. it's ok to have a single module for bwa to output BAM instead of SAM:
-//                 bwa mem | samtools view -B -T ref.fasta
-// TODO nf-core: Optional inputs are not currently supported by Nextflow. However, using an empty 
-//               list (`[]`) instead of a file can be used to work around this issue.
-
 params.options = [:]
 options        = initOptions(params.options)
 
@@ -32,7 +15,7 @@ process JUPYTERNOTEBOOK {
     //               Software MUST be pinned to channel (i.e. "bioconda"), version (i.e. "1.10").
     //               For Conda, the build (i.e. "h9402c20_2") must be EXCLUDED to support installation on different operating systems.
     // TODO nf-core: See section in main README for further information regarding finding and adding container addresses to the section below.
-    conda (params.enable_conda ? "YOUR-TOOL-HERE" : null)
+    conda (params.enable_conda ? "ipykernel jupytext nbconvert papermill" : null)
     if (workflow.containerEngine == 'singularity' && !params.singularity_pull_docker_container) {
         container "https://depot.galaxyproject.org/singularity/YOUR-TOOL-HERE"
     } else {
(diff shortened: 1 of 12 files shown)


modules/jupyternotebook/main.nf
```diff
@@ -0,0 +1,87 @@
...
+    tag "$meta.id"
+    label 'process_low'
+    publishDir "${params.outdir}",
+        mode: params.publish_dir_mode,
+        saveAs: { filename -> saveFiles(filename:filename, options:params.options, publish_dir:getSoftwareName(task.process), meta:meta, publish_by_meta:['id']) }
+
+    //NB: You likely want to override this with a container containing all required
+    //dependencies for you analysis. The container at least needs to contain the
+    //yaml and rmarkdown R packages.
+    //TODO: what container to use as default image?
```
grst, line 21: `TODO: what container to use as default image?`

For most notebooks, users will have to override the conda/container directive anyway with their own environment that covers all the specific dependencies needed. The question for me is, if the default image should be a minimal stack that runs the test notebooks, or if we should use a fully-fletched data-science image (pandas, numpy, scipy etc.) that covers many notebooks out-of-the box.

modules/jupyternotebook/parametrize.nf
```diff
@@ -0,0 +1 @@
+../rmarkdown/parametrize.nf
```
grst, line 1: TODO: replace with the actual file... just keeping the symlink for development such that the files don't get out-of-sync.

modules/rmarkdown/parametrize.nf
```diff
@@ -0,0 +1,23 @@
+@Grab(group='org.yaml', module='snakeyaml', version='1.28')
```
grst, line 1: I use the external dependency snakeyaml for dumping the notebook parameters. 
Grab automatically fetches the dependency, but this obviously won't work on offline clusters. 

Possible solutions
 * Bundle the `jar` with the module somehow
 * Re-implement a simplistic yaml dumper in groovy 
 * make `nf-core download` fetch `@Grab` jars automatically

modules/jupyternotebook/meta.yml
```diff
@@ -0,0 +1,47 @@
...
+      description: File containing software version
+      pattern: "*.{version.txt}"
+  ## TODO nf-core: Delete / customise this example output
+  - bam:
+      type: file
+      description: Sorted BAM/CRAM/SAM file
+      pattern: "*.{bam,cram,sam}"
+
+authors:
+  - "@grst"
```
grst, line 47: Please refer to the `rmarkdown` variant of `meta.yml` for now. They are highly redundant, and let's discuss it in a single place first.

modules/jupyternotebook/main.nf
```diff
@@ -0,0 +1,87 @@
...
+    # dump parameters to yaml
+    ${params_cmd}
+
+    # Convert notebook to ipynb using jupytext, execute using papermill, convert using nbconvert
+    jupytext --to notebook --output - --set-kernel - ${notebook}  \\
+        | ${render_cmd} \\
+        | jupyter nbconvert --stdin --to html --output ${notebook.baseName}.html
+
+    # TODO how to output versions of multiple tools?
+    echo \$(jupytext --version) > ${software}.version.txt
```
grst, line 85: The nbconvert and papermill versions would also be relevant. How do we handle modules that use multiple pieces of software?

tests/modules/jupyternotebook/main.nf
```diff
@@ -0,0 +1,13 @@
...
+
+include { JUPYTERNOTEBOOK } from '../../../modules/jupyternotebook/main.nf' addParams( options: [:] )
+
+workflow test_jupyternotebook {
+    
+    input = [ [ id:'test', single_end:false ], // meta map
+              file(params.test_data['sarscov2']['illumina']['test_paired_end_bam'], checkIfExists: true) ]
+
+    JUPYTERNOTEBOOK ( input )
+}
```
grst, line 13: For testing, I would add an example rmarkdown and jupyter notebook to the test dataset repository?

grst (commented):
I added a few open questions as a code review - would be great if I could get some feedback!

grst:
Asking additional review from @jianhong, who showed interest in such a module on slack.

modules/jupyternotebook/main.nf
mahesh-panchal: I could have sworn any of the processes that piped also had a line for printing each tool version, but I can't find a single example. 
The bwa module for example only prints the bwa version and not the samtools version. The output declaration catches all version.txt files. 

Looking at the template, it seems it doesn't handle this:
The lines https://github.com/nf-core/tools/blob/b48bce26b7da46cb71e4a8f78b68d9ceac579ca6/nf_core/pipeline-template/workflows/pipeline.nf#L94-L95 seem to assume a single file coming out of the `first` operator rather than a list of files (although flattening should take care of that). I'm also not seeing how different versions of a tools are supported here. The `map` after `groupTuple` only takes the first entry.

modules/rmarkdown/main.nf
```diff
@@ -0,0 +1,91 @@
...
+
+    # Set parallelism for BLAS/MKL etc. to avoid over-booking of resources
+    export MKL_NUM_THREADS="${task.cpus}"
+    export OPENBLAS_NUM_THREADS="${task.cpus}"
+    export OMP_NUM_THREADS="${task.cpus}"
+
+    # dump parameters to yaml
+    ${params_cmd}
+
+    # work around  https://github.com/rstudio/rmarkdown/issues/1508
```
mahesh-panchal, line 80: Why not have the `output_dir='.'` in the `render` function instead?

grst: I think `'.'` wouldn't work, it would resolve to the directory the symbolic link points to. [Others have mentioned](https://www.biostars.org/p/9473324/#9473334) `output_dir=getwd()` but there are some concerns that it could make problems in some cases so I thought the current version would be safest.

grst:
I added tests now, see the sister PR https://github.com/nf-core/test-datasets/pull/297 in test-datasets.

modules/jupyternotebook/main.nf
```diff
@@ -0,0 +1,98 @@
...
+    # Convert notebook to ipynb using jupytext, execute using papermill, convert using nbconvert
+    jupytext --to notebook --output - --set-kernel - ${notebook}  \\
+        | ${render_cmd} \\
+        | jupyter nbconvert --stdin --to html --output ${prefix}.html
+
+    # TODO how to output versions of multiple tools?
+    echo ${software},jupytext,\$(jupytext --version) > versions.csv
+    echo ${software},ipykernel,\$(python -c "import ipykernel; print(ipykernel.__version__)") >> versions.csv
+    echo ${software},nbconvert,\$(jupyter nbconvert --version) >> versions.csv
+    echo ${software},papermill,\$(papermill --version | cut -f1 -d' ') >> versions.csv
```
mahesh-panchal, line 86:
```suggestion
    cat <<-END_VERSIONS > versions.csv 
    ${software},jupytext,\$(jupytext --version)
    ${software},ipykernel,\$(python -c "import ipykernel; print(ipykernel.__version__)")
    ${software},nbconvert,\$(jupyter nbconvert --version)
    ${software},papermill,\$(papermill --version | cut -f1 -d' ')
    END_VERSIONS
```
I think using `cat` here is less prone to errors than using multiple `echo`'s and redirects

modules/jupyternotebook/main.nf
grst: I decided to go for a minimal image and requested the corresponding biocontainers.

modules/jupyternotebook/main.nf
JoseEspinosa: I agree, but I will argue that the version of @grst is more newbie-friendly.

mahesh-panchal: I think using a heredoc is still beginner-friendly, as long as it's explained in the comment text of the template module. I agree that a beginner would be more familiar with `echo` and likely never heard of a heredoc, but I do not think that should stop us from using it if it results in code which is easier to read and debug. I think a mistyped `>` which should be a `>>` for example is easy to overlook. A beginner would still need to figure out the version number wrangling, which I think can be potentially more complex than understanding a heredoc allows for formatting text directly as input.

JoseEspinosa: Fair enough 👍

modules/jupyternotebook/main.nf
jianhong: I will prefer a function similar to R::sessionInfo() to export all the imported modules and their versions.

modules/rmarkdown/main.nf
jianhong: Should it be better to make a hard copy from the hyperlink, then render and rename the output back?

modules/rmarkdown/main.nf
```diff
@@ -0,0 +1,91 @@
...
+
+    output:
+    tuple val(meta), path("*.html"), emit: report
+    path("artifacts/*"), emit: artifacts, optional: true
+    path "*.version.txt", emit: version
+
+    script:
+    def software = getSoftwareName(task.process)
+    def prefix = options.suffix ? "${meta.id}${options.suffix}" : "${meta.id}"
+
```
jianhong, line 42: check the rmd file contain sessionInfo section or not. if not add seesionInfo() at the end of rmd file.

grst: While I like the idea of storing `sessionInfo`, I'm not a big fan of modifying the notebook. But maybe we can separately dump the `sessionInfo` into a log file.

modules/rmarkdown/main.nf
grst: I want to preserve the original filename of the Rmd, because if you omit the `title:` section in the YAML frontmatter, rmarkdown will use the filename as title.

modules/jupyternotebook/main.nf
```diff
@@ -0,0 +1,96 @@
...
+            nb_params["meta"] = meta
+        }
+        nb_params += parameters
+        params_cmd = dump_params_yml(nb_params)
+        render_cmd = "papermill -f .params.yml"
+    } else {
+        render_cmd = "papermill"
+    }
+
+    params_cmd +
```
jfy133, line 66: Hanging command/variable definition?

modules/jupyternotebook/main.nf
jfy133: I haven't heard of heredocs, but I think the above looks relatively straightforward to work out

I would recommend adding some documentation in the nf-core/modules guidelines, and pasting this as an example.

modules/jupyternotebook/parametrize.nf
jfy133: I know it's already consistent in what's already imported in the modules, but shouldn't it be `paramet**e**rize.nf` (or even better `parameterise.bf` 💂 ;) )

modules/rmarkdown/main.nf
```diff
@@ -0,0 +1,92 @@
+// Import generic module functions
+include { initOptions; saveFiles; getSoftwareName } from './functions'
+include { dump_params_yml } from "./parametrize"
+
+params.options = [:]
+options        = initOptions(params.options)
+params.parametrize = true
+params.implicit_params = true
+params.meta_params = true
```
jfy133, line 9:
```suggestion
params.options         = [:]
options                = initOptions(params.options)
params.parametrize     = true
params.implicit_params = true
params.meta_params     = true
```

inb4 @drpatelh ;)

modules/rmarkdown/main.nf
```diff
@@ -0,0 +1,92 @@
...
+
+process RMARKDOWN {
```
jfy133, line 11:
```suggestion
process RMARKDOWNNOTEBOOK {
```

RMarkdown is a separate thing in principle (could have a module rendering a plain Rmarkdown document to a HTML), and would be more consistent with JUPYTERNOTEBOOK

modules/rmarkdown/main.nf
jfy133: sessionInfo might be overkill just for the rendering the notebook itself.

https://github.com/nf-core/modules#general

```
If the software is unable to output a version number on the command-line then a variable called VERSION can be manually specified to create this file e.g. homer/annotatepeaks module.
```
Might be sufficient? Or a some custom Rscript command to load the  two packages and then run `sessionInfo()` and 
then some bash cleanup?

Half completed example:

```
R -e "library('rmarkdown')" -e "library('yaml')" -e 'sessionInfo()' | grep -e 'R version' -e 'other' -A 1
```

modules/rmarkdown/main.nf
```diff
@@ -0,0 +1,92 @@
...
+        params_cmd = dump_params_yml(nb_params)
+        render_cmd = """
+            params = yaml::read_yaml('.params.yml')
+            rmarkdown::render('${prefix}.Rmd', params=params, envir=new.env())
+        """
+    } else {
+        render_cmd = "rmarkdown::render('${prefix}.Rmd')"
+    }
+
+    params_cmd +
```
jfy133, line 69: Another half completed variable definition?

modules/rmarkdown/meta.yml
```diff
@@ -0,0 +1,69 @@
+name: rmarkdown
+description: Render an rmarkdown notebook. Supports parametrization.
+keywords:
+  - sort
```
jfy133, line 4:
```suggestion
  - R
  - notebook
  - reports
```

modules/rmarkdown/meta.yml
```diff
@@ -0,0 +1,69 @@
...
+      description: HTML report generated from Rmarkdown
+      pattern: "*.html"
+  - version:
+      type: file
+      description: File containing software version
+      pattern: "*.{version.txt}"
+  - bam:
+      type: file
+      description: Sorted BAM/CRAM/SAM file
+      pattern: "*.{bam,cram,sam}"
```
jfy133, line 66:
```suggestion
```

jfy133 (commented):
Papercut comments

modules/rmarkdown/main.nf
grst: You're opening two flame-wars :rofl: 

* About parametrize/parameterize I don't care much, but apparently [both is ok](https://english.stackexchange.com/questions/79847/parametrise-or-parameterise-a-curve). I am happy to take votes :P
* I quite dislike aligned `=` because the create a lot of noise in diffs. But for the sake of consistency, I'll adopt it with gritted teeth. Here's an example why I think it's bad: https://github.com/nf-core/modules/pull/649/files#diff-5a76e5e2d27c76053ced9fdba0a3e5045bafc08de5014b8df3cda826f38730a7L42-**R47**. In that PR, just one line has been added, but the diff shows 6 modified lines.

modules/jupyternotebook/main.nf
grst: Actually not, this concatenates params_cmd with the """ script section. 

I did it that way because of different indentation levels between `params_cmd` and the script section. Ideally I find a way that's clearer.

modules/jupyternotebook/main.nf
grst: Here's the related PR by @mahesh-panchal: https://github.com/nf-core/tools/pull/1247

modules/rmarkdown/main.nf
grst: Could even think about moving the modules to `custom`?

modules/rmarkdown/main.nf
jfy133: Yeaaaah I know. I also personally don't like it but given 90% of Harshil's review of my PRs were those I thought better to soften the blow ;)

modules/rmarkdown/main.nf
grst: Ok, I added a sessioninfo log as separate output. 
Into the version.yml/versions.txt, I will only include the `rmarkdown` and `yaml` packages which are directly needed for rendering the notebook.

modules/jupyternotebook/main.nf
grst: see other discussion thread and nf-core/tools#1247

PR #618, opened 2021-07-25
Add variantbam

bjohnnyd:
Add variantbam for alignment filtering and downsampling.

- [x] Add a resource `label`

### DIFF
--- a/modules/variantbam/main.nf
+++ b/modules/variantbam/main.nf
@@ -33,9 +33,9 @@ process VARIANTBAM {
     // TODO nf-core: See section in main README for further information regarding finding and adding container addresses to the section below.
     conda (params.enable_conda ? "bioconda::variantbam=1.4.4a" : null)
     if (workflow.containerEngine == 'singularity' && !params.singularity_pull_docker_container) {
-        container "https://depot.galaxyproject.org/singularity/https://depot.galaxyproject.org/singularity/variantbam:1.4.4a--h7d7f7ad_5"
+        container "https://depot.galaxyproject.org/singularity/variantbam:1.4.4a--h7d7f7ad_5"
     } else {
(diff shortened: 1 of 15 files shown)


tests/modules/variantbam/test.yml
```diff
@@ -0,0 +1,7 @@
+- name: variantbam test_variantbam
+  command: nextflow run tests/modules/variantbam -entry test_variantbam -c tests/config/nextflow.config
+  tags:
+    - variantbam
+  files:
+    - path: output/variantbam/test.bam
+      md5sum: fc08f065475d60b3b06ee32920564d4b
```
jfy133, line 7: Missing the `qcreport`? Or does the test not generate one (as I  see you've specified it as optional)?

jfy133 (approved):
Looks pretty straight forward to me, although getting second eyes might be worth it as I'm still a newbie with modules.

tests/modules/variantbam/test.yml
bjohnnyd: Good catch, yes definitely should add test for that and implement it.

drpatelh (approved):
LGTM too! Feel free to merge after adding the extra test.

modules/variantbam/main.nf
```diff
@@ -25,7 +25,6 @@ process VARIANTBAM {
 
     output:
     tuple val(meta), path("*.bam")         , emit: bam
-    tuple val(meta), path("*.qcreport.txt"), optional: true, emit: qcreport
```
bjohnnyd: I have decided that it might be best to not actually add this, the reason being is that I feel it would be rarely used for this but it requires an additional input to always be passed given the current module setup (optional `val` as an `input`).  Given that there is `samtools/stats` which even has a its own `multiqc module` I feel like it might be best to not include this optional input as it is catering to very rare usage at the expense of an additional input.

PR #624, opened 2021-07-26
module: bwa/aln

### LINKED ISSUE #589: new module: BWA/ALN
I think it would be good to have a module for [BWA/ALN](https://bioconda.github.io/recipes/bwa/README.html)

jfy133:
## PR checklist

Adds BWA aln module.

I ensure to also export the reads alongside the `aln` produced `.sai` files as these are required for the `sai` to `sam` conversion (coming in next PE), and it is safer to ensure they are linked together (i.e., a pipeline developer don't need to add their own `.combine()` function), as the `.sai` files aren't used for anything else (AFAIK).

Closes #589 

- [x] Add a resource `label`


modules/bwa/aln/main.nf
```diff
@@ -0,0 +1,67 @@
...
+            \$INDEX \\
+            ${reads}
+
+        echo \$(bwa 2>&1) | sed 's/^.*Version: //; s/Contact:.*\$//' > ${software}.version.txt
+        """
+    } else {
+        """
+        INDEX=`find -L ./ -name "*.amb" | sed 's/.amb//'`
+
+        bwa aln \\
```
drpatelh, line 50: Hmmm...this is probably going to be quite inefficient running these in serial for paired-end data but I don't have a good alternative suggestion. Unless the input channels are tweaked to provide the `process` one fastq at a time 🤔 and we remove all of the logic here that checks if the sample is SE/PE because the module would just take a channel containing fastq files. What do you think? The workflow run-times would be better for it.

tests/modules/bwa/aln/main.nf
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env nextflow
+
+nextflow.enable.dsl = 2
+
+include { BWA_INDEX } from '../../../../modules/bwa/index/main.nf' addParams( options: [:] )
+include { BWA_ALN } from '../../../../modules/bwa/aln/main.nf' addParams( options: [:] )
```
drpatelh, line 6:
```suggestion
include { BWA_INDEX } from '../../../../modules/bwa/index/main.nf' addParams( options: [:] )
include { BWA_ALN   } from '../../../../modules/bwa/aln/main.nf'   addParams( options: [:] )
```

modules/bwa/aln/main.nf
jfy133: I get your point, and I actually also considered that originally. However I ended up going against that for the following reason:

1) Puts more effort/complication for a pipeline developer to write custom channel logic to syncronised them as input to SAMSE/SAMPE (although I guess a subworkflow could be made for this)
2) Very few people use `aln` nowadays anyway because there are better methods for longer reads. `aln` is mostly just used in the realm of palaeogenomics I believe, which has very short reads, and we are used to it being slow anyway.
3) It also would break the 'logic' that is expected by users, that the associated files come together, as nextflow couldn't label the associated output files properly, without breaking typical TSV input structure (i.e. which is F and R, how would the input channels know which is which for labelling without putting each FASTQ on a new line - which is then also difficult for downstream processing)

At least for EAGER there will be loads more runtime optimisation due to other modules being even more atomised, so I think having  one module (using old software!) still being a bit slower isn't so bad.

drpatelh: Fair enough 👍🏽

drpatelh (approved):

PR #637, opened 2021-07-28
Added a new module star fusion

praveenraj2018:
Added STAR-fusion module.


praveenraj2018:
@drpatelh @maxulysse - could you please review and approve this PR?

modules/star/fusion/main.nf
```diff
@@ -0,0 +1,73 @@
...
+        echo \$(STAR-Fusion --version 2>&1) | grep -i 'version' | sed 's/STAR-Fusion version: //' > ${software}.version.txt
+        """
+    }
+    else if (genome_resource == "not_found") {
+
+        """
+        wget https://data.broadinstitute.org/Trinity/CTAT_RESOURCE_LIB/__genome_libs_StarFv1.10/GRCh38_gencode_v37_CTAT_lib_Mar012021.plug-n-play.tar.gz .
+        mkdir -p $genome_resource_lib
+        tar -zxf GRCh38_gencode_v37_CTAT_lib_Mar012021.plug-n-play.tar.gz --strip-components=2 -C $genome_resource_lib
+        rm GRCh38_gencode_v37_CTAT_lib_Mar012021.plug-n-play.tar.gz
```
drpatelh, line 53:
```suggestion
```

drpatelh: All of this custom genome stuff should be done upstream in another module. This won't work for offline pipelines because you are using a `wget` and it will only work for `GRCh38` so is restrictive for people wanting to use other genomes.

praveenraj2018: Ok. You mean an additional module that only does the genome reference download and use it upstream to fusion module? We can also use `params.genome` to decide what to download? Do we have a example / reference module in our repo that does something similar? I want to know the save.reference, etc.

grst:
Hi @praveenraj2018, 

since we updated the modules to use a `versions.yml` file instead of a `*.version.txt` file, you'll need to update the module to the meet the new guidelines: https://nf-co.re/developers/adding_modules#general

Essentially, you'll need to 
 * use the latest version of `functions.nf` and 
 * make a `versions.yml` file as described in the link above. 

Feel free to contact us on slack if anything is unclear.

JoseEspinosa:
Hi @praveenraj2018, thank you for this PR! We are merging as many modules as possible now due to and impending restructuring of the entire repo. This will mean you will need to update the module to reflect these changes before it can be merged in the future. It appears like this module isn't ready to be merged so if applicable, we are converting it to draft and adding the WIP label. If this isn't the case please let us know and we will try to get the module in before the changes. Thanks again!

matthdsm:
Hi there!

Cheers
the nf-core maintainers

PR #638, opened 2021-07-28
Update Nextclade to output all files by default

drpatelh:
This PR will fix https://github.com/nf-core/viralrecon/issues/226. I have thought about this a little and it is going to be too much hassle adding logic to check for multiple output formats in the module / pipeline. We can simply just publish whichever files with `publish_files` in `modules.config` and leave it up to the user to amend that via a `custom.config` if required. The files will be tiny and the `work/` directory will most likely be deleted after the run anyway.


FriederikeHanssen (approved):
LGTM

PR #641, opened 2021-07-29
Add bcftools/concat module.

abhi18av:
## PR checklist

- [x] Add a resource `label`

### DIFF
--- a/modules/bcftools/concat/main.nf
+++ b/modules/bcftools/concat/main.nf
@@ -32,7 +32,7 @@ process BCFTOOLS_CONCAT {
     bcftools concat \\
         --output ${prefix}.vcf.gz \\
         $options.args \\
-        *.vcf.gz
+        ${vcfs}
 
     echo \$(bcftools --version 2>&1) | sed 's/^.*bcftools //; s/ .*\$//' > ${software}.version.txt
     """
--- a/modules/bcftools/concat/main.nf
+++ b/modules/bcftools/concat/main.nf
@@ -1,0 +1,39 @@
+// Import generic module functions
+include { initOptions; saveFiles; getSoftwareName } from './functions'
+
+params.options = [:]
+options        = initOptions(params.options)
+
+process BCFTOOLS_CONCAT {
+    tag "$meta.id"
+    label 'process_medium'
+    publishDir "${params.outdir}",
+        mode: params.publish_dir_mode,
+        saveAs: { filename -> saveFiles(filename:filename, options:params.options, publish_dir:getSoftwareName(task.process), meta:meta, publish_by_meta:['id']) }
+
+    conda (params.enable_conda ? "bioconda::bcftools=1.11" : null)
+    if (workflow.containerEngine == 'singularity' && !params.singularity_pull_docker_container) {
+        container "https://depot.galaxyproject.org/singularity/bcftools:1.11--h7c999a4_0"
+    } else {
+        container "quay.io/biocontainers/bcftools:1.11--h7c999a4_0"
+    }
+
+    input:
+    tuple val(meta), path(vcfs), path(tbis)
+
+    output:
+    tuple val(meta), path("*.gz"), emit: vcf
+    path  "*.version.txt"        , emit: version
+
+    script:
+    def software = getSoftwareName(task.process)
+    prefix       = options.suffix ? "${meta.id}${options.suffix}" : "${meta.id}"
+    """
+    bcftools concat -Oz \\
+        --output ${prefix}.vcf.gz \\
+        $options.args \\
+        *.vcf.gz
+
+    echo \$(bcftools --version 2>&1) | sed 's/^.*bcftools //; s/ .*\$//' > ${software}.version.txt
+    """
+}
--- a/modules/bcftools/concat/main.nf
+++ b/modules/bcftools/concat/main.nf
@@ -19,7 +19,7 @@ process BCFTOOLS_CONCAT {
     }
 
     input:
-    tuple val(meta), path(vcfs), path(tbis)
+    tuple val(meta), path(vcfs)
 
     output:
     tuple val(meta), path("*.gz"), emit: vcf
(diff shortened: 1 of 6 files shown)


modules/bcftools/concat/main.nf
```diff
@@ -0,0 +1,39 @@
...
+    path  "*.version.txt"        , emit: version
+
+    script:
+    def software = getSoftwareName(task.process)
+    prefix       = options.suffix ? "${meta.id}${options.suffix}" : "${meta.id}"
+    """
+    bcftools concat \\
+        --output ${prefix}.vcf.gz \\
+        $options.args \\
+        *.vcf.gz
```
jfy133, line 35:
```suggestion
        ${vcfs}
```

For safety you could so something like `${vcfs}.join('  ')` or something, just a suggestion though. 

I guess you can't guarantee that a user supplies ungzipped VCFs, so safer to supply just whatever the user inputs

jfy133 (commented):
I know not finished (`meta.yaml` needs to be updated), but couple of comments

modules/bcftools/concat/main.nf
abhi18av: Thanks for the pointer on this @jfy133  - makes sense 👍

abhi18av: I'll be adding a few more modules later today before I open up this PR for a more general review - shall I ping you then?

jfy133: Sure! Although not sure if I can review today. Just make sure each module is a separate PR as per guidelines!

modules/bcftools/concat/meta.yml
```diff
@@ -0,0 +1,41 @@
+name: bcftools_concat
+description: Concatenate VCF files
+keywords:
+    - variant calling
+    - merge
```
abhi18av, line 5:
```suggestion
    - concat
```

modules/bcftools/concat/main.nf
```diff
@@ -0,0 +1,39 @@
...
+    output:
+    tuple val(meta), path("*.gz"), emit: vcf
+    path  "*.version.txt"        , emit: version
+
+    script:
+    def software = getSoftwareName(task.process)
+    prefix       = options.suffix ? "${meta.id}${options.suffix}" : "${meta.id}"
+    """
+    bcftools concat \\
+        --output ${prefix}.vcf.gz \\
```
jfy133, line 33: Just to check, as bcftools isn't clear: does bcftools pick up the extension automatically and compress accordingly?

I see there is an `--output-type` flag which may be requried... (not used it myself ;) )

modules/bcftools/concat/main.nf
```diff
@@ -0,0 +1,39 @@
...
+        $options.args \\
```
jfy133, line 34: `--threads` is offered by bcftools for compression, so I would recommend adding this as per guidelines: http://samtools.github.io/bcftools/bcftools.html#common_options `--threads`)

jfy133 (changes requested):
Most important is adding `--threads`

modules/bcftools/concat/main.nf
abhi18av: I was also in doubt regarding this one, I referred to other modules in `bcftools` but it seems that none was using this option and I imagined that there is some other global config that is intended to control that (apart from `process_medium` etc) . 

I'd be happy to add `--threads` here and update others as well - what do you think?

modules/bcftools/concat/meta.yml
```diff
@@ -0,0 +1,41 @@
+name: bcftools_concat
+description: Concatenate VCF files
+keywords:
+    - variant calling
+    - concat
+    - VCF
```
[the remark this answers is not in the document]
abhi18av: Ah, whoops. Good catch!

modules/bcftools/concat/main.nf
jfy133: I think it would be worth it given the guidelines also says that all output should be compressed by default anyway (and is the only case when `--threads` can be applied according to the docs)!

modules/bcftools/concat/main.nf
abhi18av: Good idea, though I think `bcftools` does pick up the extension.

The `--output-type` could always be passed through the `options.args`, no?

modules/bcftools/concat/main.nf
abhi18av: Yup, done now :)

modules/bcftools/concat/main.nf
jfy133: If it does so automatically then yes, fine to leave it to `options.args`. just wanted to double check 👍

jfy133 (approved):
One minor query otherwise lgtm

modules/bcftools/concat/meta.yml
```diff
@@ -0,0 +1,42 @@
...
+        documentation: http://www.htslib.org/doc/bcftools.html
+        doi: 10.1093/bioinformatics/btp352
+input:
+    - meta:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. [ id:'test', single_end:false ]
+    - vcfs:
+        type: files
```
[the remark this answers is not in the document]
abhi18av: Hmm, in this case I've followed the example of [bcftools/merge](https://github.com/nf-core/modules/blob/28b023e6f4d0d2745406d9dc6e38006882804e67/modules/bcftools/merge/main.nf#L22)

```nextflow
// main.nf
    input:
    tuple val(meta), path(vcfs), path(tbis)

```



And it's corresponding [meta.yml](https://github.com/nf-core/modules/blob/28b023e6f4d0d2745406d9dc6e38006882804e67/modules/bcftools/merge/meta.yml#L21) file mentions `vcf` only as `files`, instead of being a part of a map.

```nextflow
input:
    - meta:
        type: map
        description: |
    - vcfs:
        type: files
```

Or have I missed the point here? 😅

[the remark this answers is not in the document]
jfy133: OK fair point. Go with the precendent ;).

I would argue the object going in is a `map` rather than a physical `file` object, but I don't think the `meta.yml` specs are standardised yet so safer to keep it consistent within the package.

abhi18av:
@jfy133 , thanks for your time for reviewing this PR - I really appreciate it 😊 

Could you please merge it? The option doesn't show up for me.

PR #649, opened 2021-08-02
Added PL tag in read group information of STAR/Align

praveenraj2018:
Added Platform (PL) tag in the RG line of STAR/Align command.
This is necessary for the downstream analysis especially for GATK to process the bam files.


maxulysse (approved):
LGTM

PR #665, opened 2021-08-12
Update fastqc to produce multi-version versions.yml

grst:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`

Exemplary PR to discuss changes of https://github.com/nf-core/tools/pull/1247 
to nf-core/modules. 

Once this PR is reviewed, we'll need to update all modules (can be automated
mostly).


README.md
```diff
@@ -411,10 +411,22 @@ using a combination of `bwa` and `samtools` to output a BAM file instead of a SA
     - `*.fastq.gz` and NOT `*.fastq`
     - `*.bam` and NOT `*.sam`
 
-- Where applicable, each module command MUST emit a file `<SOFTWARE>.version.txt` containing a single line with the software's version in the format `<VERSION_NUMBER>` or `0.7.17` e.g.
+- Where applicable, each module command MUST emit a file `versions.yml` emitting the version number for each tool executed by the module, e.g.
 
     ```bash
-    echo \$(bwa 2>&1) | sed 's/^.*Version: //; s/Contact:.*\$//' > ${software}.version.txt
```
drpatelh, line 417: Maybe we need some docs about why we are using HERE DOCUMENTS here?

modules/fastqc/functions.nf
```diff
@@ -3,10 +3,10 @@
 //
 
 //
-// Extract name of software tool from process name using $task.process
+// Extract name of module from process name using $task.process
 //
-def getSoftwareName(task_process) {
```
drpatelh, line 8: We will still need `getSoftwareName` too because there is an additional `tokenize` to get the software name? I think it's much cleaner to have functions for this sort of thing rather than a bunch of operations in the module itself?

modules/fastqc/main.nf
```diff
@@ -9,7 +9,7 @@ process FASTQC {
     label 'process_medium'
     publishDir "${params.outdir}",
         mode: params.publish_dir_mode,
-        saveAs: { filename -> saveFiles(filename:filename, options:params.options, publish_dir:getSoftwareName(task.process), meta:meta, publish_by_meta:['id']) }
+        saveAs: { filename -> saveFiles(filename:filename, options:params.options, publish_dir:getProcessName(task.process).toLowerCase(), meta:meta, publish_by_meta:['id']) }
```
drpatelh, line 12: `getProcessName(task.process).toLowerCase()` here will give you the full process name in lowecase e.g. `bwa_mem` and not `bwa`. See comment above.

drpatelh (commented):
Looks awesome @grst @mahesh-panchal ! Couple of minor comments.

drpatelh:
Will will also need to update the linting code and other sections of the `nf-core modules` command that was built around using the old code. Ping @ErikDanielsson. It would be really nice if we can somehow sanity check the versions dumped to this YAML file, especially whilst reviewing PRs to make sure the correct commands are used. Not entirely sure how to do this though 🤔 @Emiller88 would it be possible to upload the output files from `pytest workflow` as a comment on the PR? I know we have discussed this in the past and things get tricky whilst using all 3 of Singularity / Docker / Conda but thought I would ask anyway :)

modules/fastqc/functions.nf
mahesh-panchal: Reverted changes for this in the other pull. 
Better to keep the function.

grst:
>  It would be really nice if we can somehow sanity check the versions dumped to this YAML file

yes, checking the yaml for validity would be great! Also, it should not contain `END_VERSIONS` (one space at the wrong position will lead to the delimiter ending up within the yaml file)

README.md
grst: Do you mean something like this? 

> We chose a [HEREDOC](https://tldp.org/LDP/abs/html/here-docs.html) over piping into the versions file line-by-line as we believe the latter makes it easy to accidentally overwrite the file. Moreover, the exit status of the sub-shells evaluated in within the HEREDOC is ignored, ensuring that a tool's version command does not erroneously terminate the module.

modules/fastqc/functions.nf
```diff
@@ -37,7 +44,9 @@ def getPathFromList(path_list) {
 // Function to save/publish module results
 //
 def saveFiles(Map args) {
-    if (!args.filename.endsWith('.version.txt')) {
+    // TODO better way to detect that we are in a test and want to emit `versions.yml`
+    // or a least use a dedicated, unique environment variable.
+    if (!args.filename.equals('versions.yml') || System.getenv("PROFILE")) {
```
grst, line 49: I added a custom pytest-workflow test to ensure `versions.yml` is valid. 
The problem is that `versions.yml` needs to be published to be accessible from the test. By default, we exclude `verions.yml` from `publishDir`. This is my current workaround, maybe someone has a better idea...

grst:
> Will also need to update the linting code and other sections of the `nf-core modules` command that was built around using the old code. Ping @ErikDanielsson. It would be really nice if we can somehow sanity check the versions dumped to this YAML file, especially whilst reviewing PRs to make sure the correct commands are used. Not entirely sure how to do this though thinking @Emiller88 would it be possible to upload the output files from `pytest workflow` as a comment on the PR? I know we have discussed this in the past and things get tricky whilst using all 3 of Singularity / Docker / Conda but thought I would ask anyway :)

I had a quick look at the commenting feature, and it seems like it's difficult when the PR comes from a fork (which applies to basically all of them). Which of these issues do we need to address now? For the others we could create a separate issue and follow up later.

modules/fastqc/functions.nf
grst: @nf-core/modules-team, any suggestions how to solve this cleanly?

mahesh-panchal:
I'm not understanding something here. What's the purpose of `tests/test_versions_yml.py` ?

grst:
> I'm not understanding something here. What's the purpose of `tests/test_versions_yml.py` ?

It parses the `versions.yml` created by the module and checks that it's valid YAML.

mahesh-panchal:
I'm not asking the right question then.. This is a test to be run by what? I'm not familiar with all the mechanics here of testing.

Ah, wait this is meant to be run by the pytest workflow thing? And the `versions.yml` should be part of the output tested in `test.yml` as part of a standard modules test?

grst:
> Ah, wait this is meant to be run by the pytest workflow thing?

Exactely. This is what is called a [custom test](https://pytest-workflow.readthedocs.io/en/stable/#writing-custom-tests) in the pytest-workflow documentation. 

> And the versions.yml should be part of the output tested in test.yml as part of a standard modules test?

So far, we don't test for `version.txt` at all (because it is not a published file, so there's nothing to test). Here I propose to publish the file iff. we know we are running on the CI. 

Testing the checksum alone would not be helpful, as the checksum would be automatically generated by the tools command which doesn't help finding out if it actually is a valid `versions.yml`.

mahesh-panchal:
Perhaps an alternative is to check an exclude list rather than explicitly refer to a file ( if file in params.modules.excludeFileList ). The exclude list could then be a default parameter in the DSL2 template, but left empty in the test workflows. The only complication is then for people who decide to use nf-core modules in their own workflow need to understand this is a thing too.

README.md
drpatelh: This will need to be moved to the website now since we have moved the docs but looks great 👍🏽

modules/fastqc/functions.nf
drpatelh: Could we maybe export a custom variable just for testing with nf-core/modules? or check if `pytest` is available as a command? 🤔

tests/test_versions_yml.py
```diff
@@ -0,0 +1,26 @@
+from pathlib import Path
+import pytest
+import yaml
```
drpatelh, line 3: Looking forward to seeing how this works! 😍

README.md
grst: resolved in https://github.com/nf-core/nf-co.re/pull/877

modules/fastqc/functions.nf
grst: we decided to go for a variable, see discussion on slack: 
https://nfcore.slack.com/archives/CJRH30T6V/p1632300389380200

tests/test_versions_yml.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+@pytest.mark.workflow(*_get_workflow_names())
+def test_ensure_valid_version_yml(workflow_dir):
+    workflow_dir = Path(workflow_dir)
+    software_name = workflow_dir.name.split("_")[0].lower()
+    versions_yml = (workflow_dir / f"output/{software_name}/versions.yml").read_text()
+    assert "END_VERSIONS" not in versions_yml
+    # Raises an exception if yaml is not valid
+    yaml.safe_load(versions_yml)
```
grst, line 26: TODO: add check that the yaml contains at least one entry and the version number starts with a number.

grst:
Now, let's hope the CI works with the new #tools version

tests/test_versions_yml.py
grst: done

drpatelh (approved):
🤩

PR #686, opened 2021-09-03
Fix bug in gffread

### LINKED ISSUE #684: [BUG] gffread fails when the input file has .gtf extension
# nf-core/module bug report

## Check Documentation

I have checked the following places for your error:

- [X ] [nf-core website: troubleshooting](https://nf-co.re/usage/troubleshooting)
- [ X] [nf-core/module documentation](https://github.com/nf-core/modules/blob/master/README.md)

## Description of the bug

The module nf-core/modules/gffread will fail if the input file has .gtf extension. The module
is expecting the input file to have a gff (or other) extension and write the output to the same
file name but with .gtf extension. Therefore if the input has .gtf extension the module will
break the file and make it empty. 

## Steps to reproduce

Steps to reproduce the behaviour:

Invoke the module with a file that has a .gtf extension. 

Solution: add a string to the output file name like 'out':
`gffread $gff $options.args -o ${gff.baseName}.out.gtf`

jfnavarro:
Closes #684  https://github.com/nf-core/modules/issues/684

The module will fail if the input has .gtf as extension.


jfy133 (approved):

modules/gffread/meta.yml
```diff
@@ -14,26 +14,16 @@ tools:
...
-  - meta:
-      type: map
-      description: |
-        Groovy Map containing sample information
-        e.g. [ id:'test', single_end:false ]
   - gtf:
       type: file
       description: GTF file resulting from the conversion of the GFF input file
-      pattern: "*.gtf"
+      pattern: "*.{out.gtf}"
```
drpatelh, line 26:
```suggestion
      pattern: "*.{gtf}"
```

drpatelh (changes requested):
Thanks @jfnavarro ! The best way to fix this would be to use the `prefix` variable so developers can pick whatever suffix they want to use as opposed to using `out` as hard-coded. 

You will also need to change the line below:
https://github.com/nf-core/modules/blob/561f16fe7430eefce5b668afccde7c20790dda0f/tests/modules/gffread/main.nf#L5

to

```
include { GFFREAD } from '../../../modules/gffread/main.nf' addParams( options: [suffix: '.out'] )
```

and it should work 🤞🏽

drpatelh (approved):

PR #688, opened 2021-09-03
new module: pbccs

sguizard:
The ccs program generate the CCS sequences from PacBio Isoseq raw reads. 

## PR checklist

- [X] If necessary, include test data in your PR. (test-dataset repo)
- [X] Emit the `<SOFTWARE>.version.txt` file.
- [X] Add a resource `label`
    - [X] `PROFILE=docker pytest --tag <MODULE> --symlink --keep-workflow-wd`
    - [X] `PROFILE=singularity pytest --tag <MODULE> --symlink --keep-workflow-wd`
    - [X] `PROFILE=conda pytest --tag <MODULE> --symlink --keep-workflow-wd`


sguizard:
@nf-core/modules-team Hi, could you, please, review my code?

modules/pbccs/main.nf
```diff
@@ -0,0 +1,49 @@
...
+
+    conda (params.enable_conda ? "bioconda::pbccs=6.0.0" : null)
+    if (workflow.containerEngine == 'singularity' && !params.singularity_pull_docker_container) {
+        container "https://depot.galaxyproject.org/singularity/pbccs:6.0.0--h9ee0642_2"
+    } else {
+        container "quay.io/biocontainers/pbccs:6.0.0--h9ee0642_2"
+    }
+
+    input:
+    tuple val(meta), path(bam), path(pbi), val(chunk_num), val(chunk_on)
```
jfy133, line 22: I'm not familiar with the tool nor the data, so sorry for the naïve question: but is the `chunk_{num,on}` a specific bit of metadata of the given BAM file? Or are they options? If the former, where is the `chunk` value defined - an upstream module?

`pbi` is OK to be with `bam` as I assume that's an [index file directly related to the BAM](https://github.com/nf-core/modules#inputoutput-options)?

modules/pbccs/main.nf
```diff
@@ -0,0 +1,49 @@
...
+
+    output:
+    tuple val(meta), path("*.ccs.bam"), path("*.ccs.bam.pbi"), emit: bam
+    tuple path("*.ccs_report.txt"), path("*.ccs_report.json"), path ("*.zmw_metrics.json.gz"), emit: reports
```
jfy133, line 26: These should all be in separate channels (each file can be output with the `meta` tuple, so a pipeline dev can associate them together themselves if required)

modules/pbccs/main.nf
sguizard: The `chunk_{num,on}` are options of `ccs`.
They have a parallelization purpose. 
Instead of splitting the bam file and run `ccs` on each part, you run `ccs` X times on the same bam file but you define the part to be processed.
`chunk_on` defines the number of parts of the bam file to process, and `chunk_num` select which part of the file is processed. 

In the pipeline I'm preparing, there is a `chunk` parameter. It creates a channel with all couples of `chunk_num` / `chunk_on` (1/100, 2/100, 3/100 ...) and then combine it to the `meta + bam + pbi` channel. 

I'm aware that program options should be passed through `addParams`at the module import. But because `chunk_num` is variable, I can't set it at module import.

The `pbi` file is the `bam` PacBio Index. `ccs` needs the pair of files for running.

modules/pbccs/main.nf
sguizard: Understood, I update that.

modules/pbccs/main.nf
jfy133: Ok. I so I guess then `chunk_new` would be dynamic, and would be associated with a specific BAM file.

However would `chunk_on` be typically the same across all BAM files? So in a workflow could you theoretically specify that _all_ BAM files would be split to the same number of chunks?

I ask as AFAIK (please correct me if I'm wrong @drpatelh ) we try to keep things as generic as possible and not make input channels for modules structured to a way specific to a certain pipeline (if that makes sense). Therefore we _prefer_ having `one input = one channel`

sguizard: Yes.
The idea is, for each (bam + pbi) input, ccs is run `chunk` times. 
So, if there's two bam files (samp1.bam + samp1.bam.pbi) and `chunk == 3`, the input channel will be: 
[samp1.bam, samp1.bam.pbi, 1, 3], 
[samp1.bam, samp1.bam.pbi, 2, 3], 
[samp1.bam, samp1.bam.pbi, 3, 3], 
[samp2.bam, samp2.bam.pbi, 1, 3], 
[samp2.bam, samp2.bam.pbi, 2, 3], 
[samp2.bam, samp2.bam.pbi, 3, 3]
Indeed, the `chunk_on` parameter is the same and could be defined at module import in addParams. 
I can the code update for that. 

I understand the point about `one input = one channel`. 
I can update the code to follow that.

sguizard: @jfy133 After code reviews, the `chunk_on` parameter cannot be passed at import with addParams. 
Indeed, the arg of the option `--chunk`  must be formatted like this: '`chunk_number/number_of_chunks`'. So the complete argument is variable. 
The best I can do is to reduce the number of arguments and merged `chunk_num` and `chunk_on` into one input (`chunk`)

modules/pbccs/main.nf
```diff
@@ -0,0 +1,55 @@
...
+    conda (params.enable_conda ? "bioconda::pbccs=6.0.0" : null)
+    if (workflow.containerEngine == 'singularity' && !params.singularity_pull_docker_container) {
+        container "https://depot.galaxyproject.org/singularity/pbccs:6.0.0--h9ee0642_2"
+    } else {
+        container "quay.io/biocontainers/pbccs:6.0.0--h9ee0642_2"
+    }
+
+    input:
+    tuple val(meta), path(bam)
+    path pbi
```
jfy133, line 23: Sorry to contradict myself now, but I've seen a lot of modules having BAM + BAI associated with each other, so if you would rather have the pbi index with the BAM together I think this is OK.

modules/pbccs/meta.yml
```diff
@@ -0,0 +1,53 @@
...
+      type: file
+      description: Pacbio BAM Index
+      pattern: "*.pbi"
+  - chunk_num:
+      -type: integer
+      -description: BAM part to process
+  - chunk_on:
+      -type: integer
+      -description: Total number of bam parts to process
+  - rq:
```
jfy133, line 34: This doesn't have a input channel associated with it? Either add in `main.nf` or remove here

jfy133 (changes requested):
Thing we are pretty much there. A couple of comments but then good to go

modules/pbccs/main.nf
jfy133: Ah ok, don't worry - I meant precisely that to have the chunks options as input channels, not as addParams. You have done what I meant anyway :)

modules/pbccs/main.nf
sguizard: No problem. If it's allowed/tolerate, I prefer keep them together.

jfy133 (approved):

PR #722, opened 2021-09-20
CAT CAT

erikrikarddaniel:
This is the `CAT_CAT` module discussed on Slack.

It concatenates gzipped or uncompressed files. Whether input files are gzipped or not is decided based on the suffix of the first file in the list of input files. The `options.args` option is honoured; if it's not set, the suffix will be `.out`.

The `args2` option is used to implement filters and need to be preceded with a `|`. This is not documented anywhere except here.

## PR checklist

- [x] This comment contains a description of changes (with reason).
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [x] If necessary, include test data in your PR.
- [x] Remove all TODO statements.
- [x] Emit the `<SOFTWARE>.version.txt` file.
- [x] Follow the naming conventions.
- [x] Follow the parameters requirements.
- [x] Follow the input/output options guidelines.
- [x] Add a resource `label`
- [x] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
    - [x] `PROFILE=docker pytest --tag <MODULE> --symlink --keep-workflow-wd`
    - [ ] `PROFILE=singularity pytest --tag <MODULE> --symlink --keep-workflow-wd`
    - [x] `PROFILE=conda pytest --tag <MODULE> --symlink --keep-workflow-wd`

PR #723, opened 2021-09-20
Refactor cat/cat module

drpatelh:
Follow up PR to https://github.com/nf-core/modules/pull/722

### DIFF
--- a/modules/cat/cat/main.nf
+++ b/modules/cat/cat/main.nf
@@ -29,12 +29,12 @@ process CAT_CAT {
     def file_list = files_in.collect { it.toString() }
     if (file_list.size > 1) {
 
-        // | input    | output   | command1 | command2 |
-        // |----------|----------|-------- -|----------|
-        // | zipped   | zipped   | cat      |          |
-        // | unzipped | unzipped | cat      |          |
-        // | zipped   | unzipped | zcat     |          |
-        // | unzipped | zipped   | cat      | pigz     |
+        // | input     | output     | command1 | command2 |
+        // |-----------|------------|----------|----------|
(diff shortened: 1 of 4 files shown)


modules/cat/cat/main.nf
```diff
@@ -18,22 +18,36 @@ process CAT_CAT {
...
+    def file_list = files_in.collect { it.toString() }
+    if (file_list.size > 1) {
+
+        // | input    | output   | command1 | command2 |
+        // |----------|----------|-------- -|----------|
+        // | zipped   | zipped   | cat      |          |
+        // | unzipped | unzipped | cat      |          |
+        // | zipped   | unzipped | zcat     |          |
+        // | unzipped | zipped   | cat      | pigz     |
+
```
heuermh, line 38: This looks reasonable, suggest to replace `zipped` with `gzipped` or `compressed` and `unzipped` with `ungzipped` or `uncompressed` here and everywhere else

erikrikarddaniel: Since `.gz` is what's supported (and not `.zip` or any other) I think `gzipped` and `ungzipped` are most transparent. `compressed` indicates to me any compression and `zipped` the `.zip`.

erikrikarddaniel (approved):
Nice generalisation! :-)

modules/cat/cat/main.nf
drpatelh: Makes sense. Fixed https://github.com/nf-core/modules/pull/723/commits/614853fa11486ca3e2367f9fe548b25881500da4

heuermh:
Thank you, @drpatelh @erikrikarddaniel!

PR #724, opened 2021-09-21
Allow one single end or two paired end files in cat/fastq

heuermh:
While reviewing another issue, I noticed that attempting to cat one single end or two paired end files in `cat/fastq` would result in error, e.g.

```
$ PROFILE=docker nextflow run ./tests/modules/cat/fastq \
  -entry test_cat_fastq_paired_end_single_pair_files \
  -c tests/config/nextflow.config

N E X T F L O W  ~  version 21.04.1
Launching `./tests/modules/cat/fastq/main.nf` [chaotic_williams] - revision: a5ff39ccbb
executor >  local (1)
[8f/7b37c5] process > test_cat_fastq_paired_end_single_pair_files:CAT_FASTQ (test) [100%] 1 of 1, failed: 1 ✘
Error executing process > 'test_cat_fastq_paired_end_single_pair_files:CAT_FASTQ (test)'

Caused by:
  Process `test_cat_fastq_paired_end_single_pair_files:CAT_FASTQ (test)` terminated with an error exit status (127)

Command executed:

  null

Command exit status:
  127

Command output:
  (empty)

Command error:
  .command.sh: line 2: null: command not found
```

This pull request adds tests for those cases and fixes the errors.  I'm not sure if this helps though, perhaps this module was never intended to handle those cases.  There is still an issue if `input` were entirely empty, not sure what to do in that case.

### DIFF
--- a/modules/cat/fastq/main.nf
+++ b/modules/cat/fastq/main.nf
@@ -28,13 +28,13 @@ process CAT_FASTQ {
     def prefix   = options.suffix ? "${meta.id}${options.suffix}" : "${meta.id}"
     def readList = reads.collect{ it.toString() }
     if (meta.single_end) {
-        if (readList.size > 1) {
+        if (readList.size >= 1) {
             """
             cat ${readList.sort().join(' ')} > ${prefix}.merged.fastq.gz
             """
         }
     } else {
-        if (readList.size > 2) {
+        if (readList.size >= 2) {
             def read1 = []
             def read2 = []
             readList.eachWithIndex{ v, ix -> ( ix & 1 ? read2 : read1 ) << v }
--- a/tests/modules/cat/fastq/main.nf
+++ b/tests/modules/cat/fastq/main.nf
@@ -4,7 +4,15 @@ nextflow.enable.dsl = 2
 
 include { CAT_FASTQ } from '../../../../modules/cat/fastq/main.nf' addParams( options: [:] )
 
-workflow test_cat_fastq_single_end {
+workflow test_cat_fastq_single_end_single_file {
+    input = [ [ id:'test', single_end:true ], // meta map
+              [ file(params.test_data['sarscov2']['illumina']['test_1_fastq_gz'], checkIfExists: true) ]
+            ]
+
+    CAT_FASTQ ( input )
+}
+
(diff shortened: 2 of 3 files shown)


drpatelh:
Good spot @heuermh ! I intentionally wrote the module this way because it is better to filter the channels for those that have 1 set of single/paired-end files like in the logic [here](https://github.com/nf-core/rnaseq/blob/8094c42add6dcdf69ce54dfdec957789c37ae903/workflows/rnaseq.nf#L258-L264) and then pass any channels with `multiple` FastQ files to this module like [here](https://github.com/nf-core/rnaseq/blob/8094c42add6dcdf69ce54dfdec957789c37ae903/workflows/rnaseq.nf#L271). This means were aren't essentially generating a duplicate FastQ file unecessarily via `cat` for those samples that don't need one.

Maybe the middle ground here is to add a comment in the module referencing this comment or the code above?

heuermh:
> Good spot @heuermh ! I intentionally wrote the module this way because it is better to filter the channels for those that have 1 set of single/paired-end files like in the logic [here](https://github.com/nf-core/rnaseq/blob/8094c42add6dcdf69ce54dfdec957789c37ae903/workflows/rnaseq.nf#L258-L264) and then pass any channels with `multiple` FastQ files to this module like [here](https://github.com/nf-core/rnaseq/blob/8094c42add6dcdf69ce54dfdec957789c37ae903/workflows/rnaseq.nf#L271). This means we aren't essentially generating a duplicate FastQ file unecessarily via `cat` for those samples that don't need one.

Thanks!  Figured this was the case.

> Maybe the middle ground here is to add a comment in the module referencing this comment or the code above?

Comments in the module would be ok, or have we established any pattern for emitting warnings or throwing errors in nf-core modules when preconditions are not met?

drpatelh:
A warning would be nice too 👍🏽 We can use `log.info` like here maybe linking to a comment too?

https://github.com/nf-core/modules/blob/77a2895785587219ee2b9d0ea16fa0d71ad14f3a/modules/picard/markduplicates/main.nf#L35

heuermh:
Closing this as WontFix, will create a new PR with comment(s) and `log.info` later after multi-version changes go through.

PR #736, opened 2021-09-22
Add tests and yml file for macs2/callpeak

JoseEspinosa:
This PR add test and metadata for the macs2/callpeak module

As listed on #430


drpatelh:
Added in https://github.com/nf-core/modules/pull/1038

PR #743, opened 2021-09-25
Run CI only on pull-requests and on the master-branch.

grst:
Currently the CI runs twice unless the PR comes from a forked repository, see for instance #739

image

The suggested change makes the CI only run on pull requests, and if the changes actually get merged into the master branch.

### DIFF
--- a/.github/workflows/code-linting.yml
+++ b/.github/workflows/code-linting.yml
@@ -1,5 +1,10 @@
 name: Code Linting
-on: [push, pull_request]
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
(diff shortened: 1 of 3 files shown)


Emiller88 (approved):
Beautiful! This has been bugging me for a while in this repo too!

Can we update this in the main nf-core template if this is also an issue there?

grst:
This becomes trickier if there's also a dev branch. If we add both
branches, then PRs from dev to master would run twice.

Maybe
```
pull_request: [ dev ]
push: [ dev, master ]
```

would work?

Emiller88:
I think that's the right idea!

Issue #745, opened 2021-09-27
Fix `versions.yml` in all modules


grst:
## COMPLETED

- [x] 	abacas (assigned @drpatelh)
- [x] 	agrvate (assigned @drpatelh)
- [x] 	bandage/image (assigned @drpatelh)
- [x] 	bismark/align (assigned @drpatelh)
- [x] 	bismark/deduplicate (assigned @drpatelh)
- [x] 	bismark/genomepreparation (assigned @drpatelh)
- [x] 	bismark/methylationextractor (assigned @drpatelh)
- [x] 	bismark/report (assigned @drpatelh)
- [x] 	bismark/summary (assigned @drpatelh)
- [x] 	bowtie/align (assigned @drpatelh)
- [x] 	bowtie/build (assigned @drpatelh)
- [x] 	bowtie2/align (assigned @drpatelh)
- [x] 	bowtie2/build (assigned @drpatelh)
- [x] 	bwa/aln (assigned @drpatelh)
- [x] 	bwa/index (assigned @drpatelh)
- [x] 	bwa/mem (assigned @drpatelh)
- [x] 	bwa/sampe (assigned @drpatelh)
- [x] 	bwa/samse (assigned @drpatelh)
- [x] 	bwamem2/index (assigned @drpatelh)
- [x] 	bwamem2/mem (assigned @drpatelh)
- [x] 	bwameth/align (assigned @drpatelh)
- [x] 	bwameth/index (assigned @drpatelh)
- [x] 	cat/fastq (assigned @drpatelh)
- [x] 	chromap/chromap  (assigned @rpetit3)
- [x] 	delly/call  (assigned @rpetit3)
- [x] 	ensemblvep  (assigned @rpetit3)
- [x] 	expansionhunter  (assigned @rpetit3)
- [x] 	fgbio/callmolecularconsensusreads  (assigned @rpetit3)
- [x] 	fgbio/sortbam  (assigned @rpetit3)
- [x] 	flash (assigned @drpatelh)
- [x] 	gatk4/applybqsr (assigned @drpatelh)
- [x] 	gatk4/baserecalibrator (assigned @drpatelh)
- [x] 	gatk4/bedtointervallist (assigned @drpatelh)
- [x] 	gatk4/createsequencedictionary (assigned @drpatelh)
- [x] 	gatk4/fastqtosam (assigned @drpatelh)
- [x] 	gatk4/getpileupsummaries (assigned @drpatelh)
[...]

PR #750, opened 2021-09-27
Fix more version commands

drpatelh:
As listed in https://github.com/nf-core/modules/issues/745


drpatelh:
The following modules are failing because the `END_VERSIONS` tag is still being found in `versions.yml`. No idea why though...

- `chromap/chromap`
- `minia`

mahesh-panchal:
Chromap is missing the samtools entry from the compression commands. Are you also including gzip?

drpatelh:
We could do. No harm. Will need to do another round of all multi-tool modules adding the relevant software versions.

PR #749, opened 2021-09-27
Fix version commands

drpatelh:
As listed in https://github.com/nf-core/modules/issues/745

PR #753, opened 2021-09-28
Module new version reporting fixes

jfy133:
Fixing stuff as per #749


drpatelh (approved):
Thanks @jfy133

tests/modules/picard/markduplicates/test.yml
```diff
@@ -12,4 +12,4 @@
   tags:
     - picard
     - picard/markduplicates
-  exit_code: 1
+
```
drpatelh, line 15: Add a path and an md5sum here if possible?

jfy133: Good catch

PR #754, opened 2021-09-28
Fix version commands: round 3

drpatelh:
As listed in https://github.com/nf-core/modules/issues/745

### DIFF
--- a/modules/bbmap/align/main.nf
+++ b/modules/bbmap/align/main.nf
@@ -57,6 +57,8 @@ process BBMAP_ALIGN {
     cat <<-END_VERSIONS > versions.yml
     ${getProcessName(task.process)}:
         ${getSoftwareName(task.process)}: \$(bbversion.sh)
+        samtools: \$(echo \$(samtools --version 2>&1) | sed 's/^.*samtools //; s/Using.*\$//')
+        pigz: \$( pigz --version 2>&1 | sed 's/pigz //g' )
     END_VERSIONS
     """
 }
--- a/modules/sequenzautils/bam2seqz/main.nf
+++ b/modules/sequenzautils/bam2seqz/main.nf
@@ -24,8 +24,8 @@ process SEQUENZAUTILS_BAM2SEQZ {
     path wigfile
 
     output:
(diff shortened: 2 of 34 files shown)


modules/bbmap/align/main.nf
```diff
@@ -57,6 +57,8 @@ process BBMAP_ALIGN {
     cat <<-END_VERSIONS > versions.yml
     ${getProcessName(task.process)}:
         ${getSoftwareName(task.process)}: \$(bbversion.sh)
+        samtools: \$(echo \$(samtools --version 2>&1) | sed 's/^.*samtools //; s/Using.*\$//')
+        pigz: \$( pigz --version 2>&1 | sed 's/pigz //g' )
```
jfy133, line 61: Missing echo for pigz?

modules/sequenzautils/bam2seqz/main.nf
```diff
@@ -38,11 +38,11 @@ process SEQUENZAUTILS_BAM2SEQZ {
         -t $tumourbam \\
         --fasta $fasta \\
         -gc $wigfile \\
-        -o ${prefix}.seqz.gz
+        -o ${prefix}.gz
 
     cat <<-END_VERSIONS > versions.yml
     ${getProcessName(task.process)}:
-        ${getSoftwareName(task.process)}: \$(sequenzautils --version 2>&1 | sed 's/^.*sequenzautils //; s/Using.*\$//')
+        ${getSoftwareName(task.process)}: \$(echo \$(sequenza-utils 2>&1) | sed 's/^.*is version //; s/ .*\$//')
```
jfy133, line 45: Missing `--version`? Or not required anymore?

jfy133 (approved):
I don't see anyhting wrong

modules/sequenzautils/bam2seqz/main.nf
jfy133: So why was it there first time :thinking: thats worrying :grimacing:. I wonder if it is possible to have pytest also spit out the contents of the `.yaml` so you can quickly check a version regex still works... (not just fail, but that it's actually formatting correctly)

PR #756, opened 2021-09-28
Fix yml ver

JoseEspinosa:
## PR checklist

As listed in #745

### DIFF
--- a/modules/shovill/main.nf
+++ b/modules/shovill/main.nf
@@ -43,8 +43,8 @@ process SHOVILL {
         --force
 
     cat <<-END_VERSIONS > versions.yml
-    ${getProcessName(task.process)}:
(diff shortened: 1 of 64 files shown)


modules/shovill/main.nf
```diff
@@ -43,7 +43,7 @@ process SHOVILL {
         --force
 
     cat <<-END_VERSIONS > versions.yml
-    ${getProcessName(task.process)}:
+    ${getProcessName(task.process)}
```
grst, line 46: the colon should be there?

grst (approved):
kallistobustools and shovill are still complaining. But we can also fix them in a follow-up PR. Other than that :boom:

modules/shovill/main.nf
JoseEspinosa: Yes, just commited it, I merged it on the browser and deleted it by mistake

PR #760, opened 2021-09-29
remove cpu restrictions for piped commands

ramprasadn:
In case of modules with piped commands, currently, we are splitting the number of cpus assigned for that process equally between the two softwares. This is good idea in general, but when we have a combination of aligner and samtools (for converting the resulting output to bam), I think it is very inefficient to assign half the number of cpus to samtools when the rate limiting step here is the alignment.

These are the affected modules.
-  bowtie/align
-  bowtie2/align
-  bwa/mem
-  bwamem2/mem
-  bwameth/align

Based on the comments and discussion in the nf-core slack channel [here](https://nfcore.slack.com/archives/CJRH30T6V/p1632822976224200), [here](https://nfcore.slack.com/archives/CJRH30T6V/p1632822291214200), and [here](https://nfcore.slack.com/archives/CJRH30T6V/p1632822851221500), I am getting rid of these restrictions in this PR. The rationale being that these manual restrictions hurt most users, and almost all schedulers in use today should be able to handle resource allocation efficiently on their own. 

Should anyone face any issues due to changes in this PR there are other workarounds (like, taskset) that they can adopt to fix those issues. 

## PR checklist

- [ ] Add a resource `label`

### DIFF
--- a/modules/bowtie/align/main.nf
+++ b/modules/bowtie/align/main.nf
@@ -29,7 +29,6 @@ process BOWTIE_ALIGN {
     tuple val(meta), path('*fastq.gz'), optional:true, emit: fastq
 
     script:
-    def split_cpus = Math.floor(task.cpus/2)
     def software  = getSoftwareName(task.process)
     def prefix    = options.suffix ? "${meta.id}${options.suffix}" : "${meta.id}"
     def unaligned = params.save_unaligned ? "--un ${prefix}.unmapped.fastq" : ''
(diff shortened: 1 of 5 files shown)


drpatelh (commented):
Thanks @ramprasadn ! LGTM 🎉  Let's wait for @pontus to confirm .

grst (approved):
This solution I can live with even on a "broken scheduler" ;)

pontus (approved):
Looks good. I'm a bit unsure if it's best to give a number of threads like this or let the tools figure things out, but this is better than what we have and I don't think there's a real noticable difference between the two "good cases" anyway.

ramprasadn:
Excellent! In that case, I will merge this PR to master.

PR #368, opened 2021-03-24
Add scrapesoftwareversions module

grst:
## PR checklist

Closes #294  

Implement a draft for a module with a custom script (`get_software_versions.py`)

- [x] This comment contains a description of changes (with reason).
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [x] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [x] If necessary, include test data in your PR.
- [ ] Remove all TODO statements.
- ~~[ ] Emit the `<SOFTWARE>.version.txt` file.~~
- [x] Follow the naming conventions.
- [x] Follow the parameters requirements.
- [x] Follow the input/output options guidelines.
- [x] Add a resource `label`
- [x] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
    - [ ] `PROFILE=docker pytest --tag <MODULE> --symlink --keep-workflow-wd`
    - [x] `PROFILE=singularity pytest --tag <MODULE> --symlink --keep-workflow-wd`
    - [ ] `PROFILE=conda pytest --tag <MODULE> --symlink --keep-workflow-wd`

PR #761, opened 2021-09-29
Add custom/dumpsoftwareversions modules for nf-core pipeline template

drpatelh:
* Updated version of (and closes) https://github.com/nf-core/modules/pull/368
* Since the Python code is self-contained within the module we don't have to worry about providing a separate executable for the module
* We can now install this module directly into the pipeline template
* Can't check for md5sums in the tests because the NF version outputted to the YAML will be subjective on where you run the tests


maxulysse (approved):
LGTM

PR #780, opened 2021-10-01
Remove def software lines and emit versions channel as plural

drpatelh:

PR #781, opened 2021-10-01
Address Jose's comments in #780

drpatelh:
Ping @JoseEspinosa . Addressed these in a separate PR so we don't re-run all of the tests.


JoseEspinosa (approved):
Perfect!

PR #788, opened 2021-10-04
Fix bug with SortMeRna not working on a single db file

drpatelh:


JoseEspinosa (approved):
LGTM! But shouldn't this be implemented at the module level and then reinstall the module with the fixed version?

grst (approved):

drpatelh:
> LGTM! But shouldn't this be implemented at the module level and then reinstall the module with the fixed version?

Yup! We are on nf-core/modules 😆 

I also pushed a cheeky commit to fix a bug in the `bbmap/bbsplit` module that generates the error below if `meta.id` is `null` e.g. when we just want to create an index.:

```
java.lang.NullPointerException: Cannot get property 'id' on null object
        at org.codehaus.groovy.runtime.NullObject.getProperty(NullObject.java:60)
        at org.codehaus.groovy.runtime.InvokerHelper.getProperty(InvokerHelper.java:194)
```

JoseEspinosa:
ups... I thought we were in the rnaseq repo :upside_down_face: 
Sorry!

PR #807, opened 2021-10-09
Add a new module for fasterq-dump

Midnighter:
Add a module for sra-tools' fasterq-dump which will be part of a sub-workflow that is later added to the fetchngs pipeline.

## PR checklist

- [x] This comment contains a description of changes (with reason).
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [x] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [x] If necessary, include test data in your PR.
- [x] Remove all TODO statements.
- [x] Emit the `versions.yml` file.
- [x] Follow the naming conventions.
- [x] Follow the parameters requirements.
- [x] Follow the input/output options guidelines.
- [x] Add a resource `label`
- [x] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
    - [x] `PROFILE=docker pytest --tag <MODULE> --symlink --keep-workflow-wd`
    - [ ] `PROFILE=singularity pytest --tag <MODULE> --symlink --keep-workflow-wd`
    - [ ] `PROFILE=conda pytest --tag <MODULE> --symlink --keep-workflow-wd`

## Open Design Decisions

- [x] Should a multi-tool container be used to include pigz/crabz to compress the FASTQ output?

PR #815, opened 2021-10-12
ci: Duplicate test jobs for subworkflows

edmundmiller:
Issue from #807


Emiller88:
image

I put these all in one file, so we get a cool visualization in one place. I'd also like to cut down on the repeated parts in the future with an elegant solution like we had before.

PR #933, opened 2021-10-27
Fix subworkflows seperate from modules

Emiller88:


drpatelh (approved):
Noice!! 😎

tests/subworkflows/nf-core/align_bowtie2/test.yml
```diff
@@ -5,14 +5,14 @@
...
     # Modules
-    - bowtie2
-    - bowtie2/align
-    - samtools
-    - samtools/index
-    - samtools/sort
-    - samtools/stats
-    - samtools/idxstats
-    - samtools/flagstat
+    # - bowtie2
```
heuermh, line 8: Would rather these lines are deleted rather than commented out

.github/workflows/nf-core-linting.yml
```diff
@@ -60,7 +60,7 @@ jobs:
 
         # FIXME: Remove this when nf-core modules lint stabilizes and install stable release
       - name: Install nf-core tools development version
-        run: python -m pip install --upgrade --force-reinstall git+https://github.com/nf-core/tools.git@dev
+        run: python -m pip install --upgrade --force-reinstall git+https://github.com/nf-core/tools.git@subworkflow_hacks
```
heuermh, line 63: Not "_hacks" if it works, right? 😉

Emiller88: Lol well it was ignoring them because we need to update the standards to match, but there's already subworkflows flowing in, so we need to punt on that. https://github.com/nf-core/tools/pull/1319

I'm trying to fix the tests first, and then I'll be able to test for all of these new things.

tests/subworkflows/nf-core/align_bowtie2/test.yml
Emiller88: Not opposed, but it would be more to go track these down later. But I'm not as scared now, I think we have a lot of hands to mobilize and can add all of these in quickly.

PR #987, opened 2021-10-29
Busco

### LINKED ISSUE #569: new module: BUSCO
Hello, would there be interest for a [BUSCO](https://busco.ezlab.org/) module ?  I will prepare one privately in any case, and would be pleased to contribute it here.

Test data would need to be submitted to nf-core/test-datasets. I have a branch ready in a fork (https://github.com/oist/plessy_nf-core_test-datasets/commit/7eeb3bb37369e3f74ab9f9f09a8da362e4699105).

priyanka-surana:
Closes #569  

Worked on this module with @charles-plessy

This BUSCO module clears all manual testing. Fails with pytest, as it doesn't copy relevant output files outside of nextflow work directory.
`$ PROFILE=singularity nextflow run tests/modules/busco -entry test_busco -c tests/config/nextflow.config`

Currently emits the `full_table.tsv` and `short_summary.txt` files outside the nextflow work directory. Let us know if we should make it more generic and output the entire `run_*` folder.


modules/busco/main.nf
```diff
@@ -0,0 +1,48 @@
...
+        container "quay.io/biocontainers/busco:5.2.2--pyhdfd78af_0"
+    }
+
+    input:
+    tuple val(meta), path(fasta)
+    path(augustus_config)
+
+    output:
+    tuple val(meta), path("${meta.id}/run_*/full_table.tsv"), emit: tsv
+    tuple val(meta), path("${meta.id}/run_*/short_summary.txt"), emit: txt
```
d4straub, line 27: please use `prefix` instead:
```suggestion
    tuple val(meta), path("${prefix}/run_*/full_table.tsv"), emit: tsv
    tuple val(meta), path("${prefix}/run_*/short_summary.txt"), emit: txt
```
Also, output should be detailed, each relevant file in one channel, also see [here](https://nf-co.re/developers/adding_modules#inputoutput-options) (it probably doesn't need to be compressed, as long as the ouput is quite small)

modules/busco/main.nf
```diff
@@ -0,0 +1,48 @@
...
+    conda (params.enable_conda ? "bioconda::busco=5.2.2" : null)
+    if (workflow.containerEngine == 'singularity' && !params.singularity_pull_docker_container) {
+        container "https://depot.galaxyproject.org/singularity/busco:5.2.2--pyhdfd78af_0"
+    } else {
+        container "quay.io/biocontainers/busco:5.2.2--pyhdfd78af_0"
+    }
+
+    input:
+    tuple val(meta), path(fasta)
+    path(augustus_config)
```
d4straub, line 23: I am missing handling of a database such as in the [nf-core/mag local module](https://github.com/nf-core/mag/blob/d4f00bc2d178c1c8dc25b576c9e77e70511d5d21/modules/local/busco.nf#L23-L24)

modules/busco/meta.yml
```diff
@@ -0,0 +1,52 @@
...
+input:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. [ id:'test', single_end:false ]
+  - fasta:
+      type: file
+      description: Nucleic or amino acid sequence file in FASTA format
+      pattern: "*.{fasta}"
```
d4straub, line 26: Are you sure it cannot handle compressed files? It might be worth to add a conditional unzipping to allow zipped input! Because modules should output zipped `fasta` (and therefore allow compressed input), according to the module guidelines.
Also, does it require the fasta extension? If not, remove it or give more choice (`fa,fa.gz`,etc).

modules/busco/meta.yml
```diff
@@ -0,0 +1,52 @@
...
+      description: File containing software versions
+      pattern: "versions.yml"
+  - tsv:
+      type: file
+      description: Full summary table
+      pattern: "*.{tsv}"
+  - txt:
+      type: file
+      description: Short summary text
+      pattern: "*.{txt}"
```
d4straub, line 48: As far as I know BUSCO has several modes, one of which allows for several output summaries, you can add optional output such as [here](https://github.com/nf-core/mag/blob/d4f00bc2d178c1c8dc25b576c9e77e70511d5d21/modules/local/busco.nf#L27-L28) to the main.nf

d4straub (commented):
Hi, looks quite good! But I do have some comments.

priyanka-surana:
I tried the prefix suggestion but it started failing the tests, so I changed it back.

modules/busco/meta.yml
priyanka-surana: I tested with `--auto-lineage` and it already works. The output folder contains files from all the different databases:
```
$ tree output/busco
output/busco
└── test
    ├── run_bacteria_odb10
    │   ├── full_table.tsv
    │   └── short_summary.txt
    └── run_bacteroidales_odb10
        ├── full_table.tsv
        └── short_summary.txt

3 directories, 4 files
```

Is this what you were thinking or something more?

modules/busco/main.nf
priyanka-surana: So in this module, we give database as one of the arguments rather than an input. In tests, it works with both local database and downloaded database. This makes it more generic, rather than absolutely needing a database.

charles-plessy:
@d4straub I also tried to substitute `meta.id` by `prefix` as you suggested and I also ran into error like Priyanka.  The error message suggests that `prefix` evaluates to `null` when searching for the result files in the `output:` section:

    Missing output file(s) `null/run_*/full_table.tsv` expected by process `test_busco:BUSCO (test)`

Interestingly, `prefix` does evaluate the same as `meta.id` in the `script:` section.  I can not find where the error is...

charles-plessy (changes requested):
@priyanka-surana For pytest to succeed, the paths and MD5 sum in the `test.yml` file have to be correct as in the patch below. (I am less sure if the two first changes of the patch are important, but this is what `create-test-yml` has output.

@d4straub The 'Edit' option is grayed and unavailable to me; do I need to be added to some kind of whitelist ?

```
diff --git a/tests/modules/busco/test.yml b/tests/modules/busco/test.yml
index aecccc45..a1363738 100644
--- a/tests/modules/busco/test.yml
+++ b/tests/modules/busco/test.yml
@@ -1,9 +1,9 @@
-- name: busco
-  command: nextflow run ./tests/modules/busco -entry test_busco -c tests/config/nextflow.config
+- name: busco test_busco
+  command: nextflow run tests/modules/busco -entry test_busco -c tests/config/nextflow.config
   tags:
     - busco
   files:
-    - path: work/*/*/test/run_bacteria_odb10/full_table.tsv
-      md5sum: e02aff5ce681637afaec99dbb845b1dc
-    - path: work/*/*/test/short_summary*.txt
+    - path: output/busco/test/run_bacteroidales_odb10/full_table.tsv
+      md5sum: 8d7b401d875ecd9291b01bf4485bf080
+    - path: output/busco/test/run_bacteroidales_odb10/short_summary.txt
       contains: ['Complete BUSCOs (C)']
```

modules/busco/meta.yml
charles-plessy: I just verified that BUSCO crashes on compressed FASTA input.  Following recent discussions on Slack, transparent decompression can be achieved with the following patch.  I chose `__UNCOMPRESSED_FASTA_FILE__` as output name to make it extremely unlikely that it clashes with an existing file.

```
index 6a1a5644..8f3b1249 100644
--- a/modules/busco/main.nf
+++ b/modules/busco/main.nf
@@ -33,11 +33,13 @@ process BUSCO {
     # Copy the image's AUGUSTUS config directory if it was not provided to the module
     [ ! -e augustus_config ] && cp -a /usr/local/config augustus_config
     AUGUSTUS_CONFIG_PATH=augustus_config \\
+    # Ensure the input is uncompressed
+    gzip -cdf $fasta > __UNCOMPRESSED_FASTA_FILE__
     busco \\
         $options.args \\
         --augustus \\
         --cpu $task.cpus \\
-        --in  $fasta \\
+        --in __UNCOMPRESSED_FASTA_FILE__ \\
         --out $meta.id
```

I verified that it works with the following change to the the test file.

```
diff --git a/tests/modules/busco/main.nf b/tests/modules/busco/main.nf
index bf03cf10..1cc8fbdb 100644
--- a/tests/modules/busco/main.nf
+++ b/tests/modules/busco/main.nf
@@ -8,8 +8,7 @@ include { BUSCO } from '../../../modules/busco/main.nf' addParams( options: [arg
 workflow test_busco {
     
        compressed_genome_file = file(params.test_data['bacteroides_fragilis']['genome']['genome_fna_gz'], checkIfExists: true)
-       GUNZIP ( compressed_genome_file )
-       input = GUNZIP.out.gunzip.map { row -> [ [ id:'test' ], row] }
+       input = [ [ id:'test' ], compressed_genome_file ]
```

I also opened an [issue upstream to request transparent decompression of input files](https://gitlab.com/ezlab/busco/-/issues/504).

modules/busco/main.nf
charles-plessy: I agree with @d4straub that it is important to support offline use. I just submitted a [pull request for a minimised lineage dataset that can be used to test the module in offline mode](https://github.com/nf-core/test-datasets/pull/392). The offline lineage datasets can be passed to the module via an extra channel, with the possibility to keep it empty like for `path(augustus_config)`.  I was using something like below in some local modules.

```
if (lineage) options.args += "--offline --lineage_dataset $lineage"
```

Shall we follow this approach, using `--download_path` instead?

d4straub: > So in this module, we give database as one of the arguments rather than an input. In tests, it works with both local database and downloaded database. This makes it more generic, rather than absolutely needing a database.

Where? Which arguments? Do you mean `option.args`? If yes, than thats possible but this is typically solved by a separate input channel. E.g.:
```suggestion
    input:
    tuple val(meta), path(fasta)
    path(augustus_config)
    path(database_path)
```
and using than in the `script:` block
`    def database = database_path ? "--download_path $database_path" : ""`
And than use 
```
    busco \\
        $options.args \\
        --augustus \\
        $database \\
        --cpu $task.cpus \\
        --in  $fasta \\
        --out $meta.id
```

> Shall we follow this approach, using --download_path instead?

I think so, yes, because it is more flexible.

modules/busco/meta.yml
d4straub: Hm I'd rather thought rather to add in the `script` block about something like 
```
if [[ $file != *.gz ]]; then
    gzip -n $fasta > $fasta.gz
fi
```
(code not tested! I am pretty sure there are examples in existing modules!)

modules/busco/meta.yml
d4straub: Each of that files in your example (4 in total) should have their own output channel. That makes the output more predictable, therefore easier to use.

d4straub:
> @d4straub The 'Edit' option is grayed and unavailable to me; do I need to be added to some kind of whitelist ?

It seems to me you are not part of the nf-core group at github, so please habe a look at https://nf-co.re/join#github

priyanka-surana:
@charles-plessy I have included all of our discussed changes. Please take a look and merge if possible. The tests worked for me.

charles-plessy (changes requested):
Hi @priyanka-surana as you may have noticed there were major changed in the structure of the modules.  I tried to apply them to `busco` using `samtools/sort` as a template.  Can you check it works ?

priyanka-surana:
hi @charles-plessy this is not working anymore. both the tests failed with the error: The parameter "mode (--mode)" was not provided. I am going to work on this later this week and get back to you. I might just start from a new template to keep things clean.

charles-plessy:
Hi @priyanka-surana I figured out that the structure of the tests also changed.  Options are now passed in a `nextflow.config` file.  Following the last commit I made, we need `tests/modules/busco/nextflow.config` to contain:

```
process {

    withName: BUSCO_BACTE {
        ext.args = '--mode genome --lineage_dataset bacteroidales_odb10'
    }

    withName: BUSCO_CHR22 {
        ext.args = '--mode genome --offline'
    }

}
```

I did not find how to create a new file in this PR using the review system.  Can you create it ?

With this, `pytest` succeeds on my side after merging on the master branch (I did not test if it was strictly required) and using the nf-core tools as of December 6th.

priyanka-surana:
@charles-plessy I have added the `nextflow.config` file and testing without pytest works. Pytest gives syntax errors, I posted on [nf-core help channel](https://nfcore.slack.com/archives/CE6SDBX2A/p1638873033132200), will have to wait till it resolves.

priyanka-surana:
@charles-plessy pytest checks successful.

modules/busco/main.nf
grst: all input files, including databases, should be passed as an input channel - only like that can nextflow take care of mounting files into containers, upload them to AWS worker nodes etc.

modules/busco/main.nf
```diff
@@ -0,0 +1,41 @@
...
+    # Ensure the input is uncompressed
+    gzip -cdf $fasta > __UNCOMPRESSED_FASTA_FILE__
+    # Copy the image's AUGUSTUS config directory if it was not provided to the module
+    [ ! -e augustus_config ] && cp -a /usr/local/config augustus_config
+    AUGUSTUS_CONFIG_PATH=augustus_config \\
+    busco \\
+        $args \\
+        --augustus \\
+        --cpu $task.cpus \\
+        --in __UNCOMPRESSED_FASTA_FILE__ \\
```
grst, line 33: If it works this is preferrable, as the uncompressed file never gets written to disk. 

```suggestion
        --in <(gzip -cdf $fasta) \\
```

If the tool reads the input file multiple times, this will fail and there's no better way than what you are already doing.

modules/busco/main.nf
```diff
@@ -0,0 +1,41 @@
...
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    if (lineage) args += " --lineage_dataset $lineage"
+    """
+    # Ensure the input is uncompressed
+    gzip -cdf $fasta > __UNCOMPRESSED_FASTA_FILE__
+    # Copy the image's AUGUSTUS config directory if it was not provided to the module
+    [ ! -e augustus_config ] && cp -a /usr/local/config augustus_config
```
grst, line 27: What's this /usr/local/config` directory? Is it part of the container? What happens if the tool is running  with conda? 
Maybe it's safest to just mandate providing the config directory as input channel.

tests/config/test_data.config
```diff
@@ -114,8 +114,8 @@ params {
                 genome_bed_gz_tbi                              = "${test_data_dir}/genomics/homo_sapiens/genome/genome.bed.gz.tbi"
                 transcriptome_fasta                            = "${test_data_dir}/genomics/homo_sapiens/genome/transcriptome.fasta"
                 genome2_fasta                                  = "${test_data_dir}/genomics/homo_sapiens/genome/genome2.fasta"
-		        genome_chain_gz                                = "${test_data_dir}/genomics/homo_sapiens/genome/genome.chain.gz"
-
+		genome_chain_gz                                = "${test_data_dir}/genomics/homo_sapiens/genome/genome.chain.gz"
+                chr22_odb10_tar_gz                             = "${test_data_dir}/genomics/homo_sapiens/genome/BUSCO/chr22_odb10.tar.gz"
```
grst, line 118: please fix the indentation

grst (changes requested):
Thanks for adding this module! 

I added a few comments! Most importantly, please add an entry to the `pytest_modules.yml` file, currently no tests are run.

PR #1015, opened 2021-11-02
Merge freebayes subtools

maxulysse:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`


maxulysse:
I figured it was a good idea to go towards input_1 and input_2 as we do in the test data, but I'm ok for any other suggestion

modules/freebayes/freebayes/main.nf
```diff
@@ -4,7 +4,7 @@ include { initOptions; saveFiles; getProcessName; getSoftwareName } from './func
 params.options = [:]
 options        = initOptions(params.options)
 
-process FREEBAYES_GERMLINE {
+process FREEBAYES_FREEBAYES {
```
FriederikeHanssen, line 7: If there is only a single freebayes module, can we use just `FREEBAYES`?

modules/freebayes/freebayes/main.nf
```diff
@@ -19,7 +19,7 @@ process FREEBAYES_GERMLINE {
     }
 
     input:
-    tuple val(meta), path(input), path(input_index)
+    tuple val(meta), path(input_1), path(input_1_index), path(input_2), path(input_2_index)
```
FriederikeHanssen, line 22: My understanding now was that Freebayes for non-somatic calling can also be run on many individuals. Should it be possible to just provide a single input which can either be a list of bam/cram files or a single one? I _think_ the somatic calling is just determined by the passed `options.args`

FriederikeHanssen: We sort of have to `flatten`the input in sarek then, which is a bit different then the other tools we use. but I think that that is a sarek problem, not a module one

maxulysse: Not sure how I understand how it was handled before, so not sure about what needs to be done there

FriederikeHanssen: let's see what @maxibor and @jfy133 say then :)

jfy133 (approved):
LGTM, thanks for this <3

modules/freebayes/freebayes/main.nf
jfy133: Normally bacterial stuff is really small, at least, so I would guess it doesn't really make a difference if you run it each sample separately or all-in-one? Do to multiple individuals.

To supply multiple BAM files you would have to spceify with `-L` and give a txt file with a list of BAM files which complicates things...

FriederikeHanssen: so I think then basically we can have only one input, if it is tumor no -L, if it is bacteria with -L and jsut have the txt file as additional optional input. The meta.yml proabbly needs some info on proper usage then or have self-explanatory tests. This would be good anyways to make sure somatic calling and microbes work :D

FriederikeHanssen: I think we should leave to the pipeline devs to decide on how to use the module and not pre-determine it here, but leave it more flexible.

PR #1030, opened 2021-11-04
CNVkit: Make targets file optional when running in WGS mode

### LINKED ISSUE #1029: CNVkit: Make targets file optional when running in WGS mode
## Is your feature request related to a problem? Please describe

CNVkit with --method wgs does not need a regions.bed file but it is currently required by the module because the targets options is required with --method hybrid and --method amplicon

## Describe the solution you'd like

Make targets an optional input parameter when options.args contains `--method wgs` (or `-m wgs`)

fbdtemme:
## PR checklist

Closes #1029

This PR makes targetfile an optional input parameter when options.args contains --method wgs (or -m wgs)
It also adds passing of $task.cpus.

- [x] Add a resource `label`


tests/modules/cnvkit/main.nf
```diff
@@ -2,18 +2,29 @@
...
             ]
     fasta   = file(params.test_data['sarscov2']['genome']['genome_fasta'], checkIfExists: true)
     targets = file(params.test_data['sarscov2']['genome']['baits_bed'], checkIfExists: true)
 
     CNVKIT ( input, fasta, targets )
 }
+
+workflow test_cnvkit_wgs {
+    input = [ [ id:'test'], // meta map
+              [ file(params.test_data['homo_sapiens']['illumina']['test2_paired_end_sorted_bam'], checkIfExists: true) ],
```
FriederikeHanssen, line 24: I guess this module will also be part of the sarek copy number calling sub workflow? In that case it needs to also function with cram files. Could you add a test for that?

modules/cnvkit/main.nf
```diff
@@ -31,13 +31,23 @@ process CNVKIT {
...
     script:
+    def target_args = ""
+
+    if (options.args.contains("--method wgs") || options.args.contains("-m wgs")) {
+        target_args = targetfile ? "--targets $targetfile" : ""
+    }
+    else {
+        target_args = "--targets $targetfile"
+    }
+
```
FriederikeHanssen, line 42: wouldn't 

```
target_args = targetfile ? "--targets $targetfile" : ""
```

be enough without the if-clause?

FriederikeHanssen (changes requested):

tests/modules/cnvkit/main.nf
lassefolkersen: Ok, [done](https://github.com/fbdtemme/modules/blob/07fa77109dba21ed57119677332e2cac418e65f1/tests/modules/cnvkit/main.nf#L33-L41).

modules/cnvkit/main.nf
lassefolkersen: The way I read the code, it seems that making the --target argument optional when running as `--method wgs`(which makes sense according to the documentation and it's the [name of the commit](https://github.com/fbdtemme/modules/commit/384994353e63cccfc0672d86e4b1d50ef1b32f5d)). There's also [test for it](https://github.com/fbdtemme/modules/blob/07fa77109dba21ed57119677332e2cac418e65f1/tests/modules/cnvkit/main.nf#L6). If you give that test a targets argument (other than `[]`) it fails. My guess is that it actually should be blank in [line 37](https://github.com/fbdtemme/modules/blob/07fa77109dba21ed57119677332e2cac418e65f1/modules/cnvkit/main.nf#L37), like return a blank line no matter what - because [it's not compatible](https://cnvkit.readthedocs.io/en/stable/nonhybrid.html#whole-genome-sequencing-wgs) to have `wgs` and `targets` together. But it's a guess - I'm actually a bit confused what the intention is here.

lassefolkersen:
I've read this through in detail, including the documentation, and run the the test locally. All that looks fine. I've also included [one more test for receiving cram files](https://github.com/fbdtemme/modules/blob/6ed15d19d2038066c163ef91805fb2fdb89b5def/tests/modules/cnvkit/main.nf#L33-L41) and that works fine now (thanks [to this update](https://github.com/etal/cnvkit/pull/495)). I couldn't really figure out the intention for the optional targets file in wgs mode - seems to be set up to handle how arguments differ when running WGS-data. But I'm unsure of it's importance - I mean there are several other modes of running CNVkit that won't be possible in this setup, e.g. it's impossible to run germline-only runs because [tumor+normal is a hard requirement](https://github.com/fbdtemme/modules/blob/6ed15d19d2038066c163ef91805fb2fdb89b5def/modules/cnvkit/main.nf#L46-L47) right now. But maybe out of scope to fix?

FriederikeHanssen:
ok looking through the docu I think it should be `CNVKIT_BATCH` within cnvkit/batch/ folder structure. They are several subcommands as I understand it similarly to samtools for example

lassefolkersen:
> Is there a way to make the module inclusive to also run germline only? I mean for strelka for example we settled for having one in the germline and one in a somatic subfloder. this would also be an option

I guess the problem is that the [input line ](https://github.com/fbdtemme/modules/blob/cnvkit-optional-targetfile/modules/cnvkit/main.nf#L22)would have to be split and have $normalbam be made optional. I think that's not a good solution either. Breaks too much in the tumor/normal workflow, which after all is the main focus of cnvkit (germline is, according to their [documentation](https://cnvkit.readthedocs.io/en/stable/germline.html), a sort of add-on with caveats). 

I think this could argue for having a `cnvkit/<method>` structure already, as you also suggest. Don't call it `cnvkit/batch` though, since `batch` is virtually all the entryways into cnvkit. Maybe `tumor`, `tumor-normal`, and `normal-only`, that's a common use case. What do you think @fbdtemme ?

modules/cnvkit/main.nf
fbdtemme: As I read the documentation it is possible to use a --target argument in wgs mode to limit the analysis to specific regions of interest, but this is an optional parameter. For the hybrid mode the targets is a required input parameter. The intention was to make it possible to pass [] as target when running in wgs mode while still allowing the parameter if required.

fbdtemme:
I personally prefer to keep the input line as as, but make the `normalbam` optional.
```
input:
tuple val(meta), path(tumourbam), path(normalbam)

script:
normal_options = normalbam ? "--normal $normalbam" : ""

cnvkit.py \\
      batch \\
      $mode_args \\
      $tumourbam \\
      $normal_options \\
```

Then we can run CNVkit in tumor only mode with `CNVKIT(meta, tumor_samples, [])`

lassefolkersen:
Sounds good with me @fbdtemme, if you put that in, then I can make a test for it, if you would like?

fbdtemme:
I am happy to do so, but I cannot do it today. I can probably look at it tomorrow. If you want to go ahead quicker and implement it you are welcome to.

lassefolkersen:
Ok, sure - I had it it all set up for testing anyway, so I'll take a go at it during today

lassefolkersen:
Hmm.. not so easy unfortunately. Giving the optional normal setup you suggested @fbdtemme starts ok, but then complains:
```
Option -n/--normal must be given to build a new reference if -r/--reference is not used
```
It's gonna be a several if statements in the module-main.nf code (or skipping a test) if we do it as-is. Still tinkering a little with this, but input very welcome.

modules/cnvkit/main.nf
```diff
@@ -40,13 +40,16 @@ process CNVKIT {
         target_args = "--targets $targetfile"
     }
 
+    normal_args = normalbam ? "--normal $normalbam" : ""
+
+    fasta_args = fasta ? "--fasta $fasta" : ""
     """
     cnvkit.py \\
         batch \\
         $tumourbam \\
```
maxulysse, line 49: I do prefer the tumor over tumour spelling
But it's mainly me being picky

lassefolkersen: Will do once I fix this current failure 😄 🇺🇸🇺🇸🇺🇸🇺🇸🇺🇸🇺🇸

modules/cnvkit/main.nf
```diff
@@ -19,7 +19,7 @@ process CNVKIT {
     }
 
     input:
-    tuple val(meta), path(tumourbam), path(normalbam)
+    tuple val(meta), path(tumorbam), path(normalbam)
```
maxulysse, line 22: <3

lassefolkersen:
I had some more time with this now. Per yesterdays message, it's not trivial to have all functionalities in one sub-module. But can be done, just needs to set up some more. The most recent commit should work, with one caveat - the `cnn`, `cnr`and `cns` outputs need to be made `optional: true`. Please advice @fbdtemme @FriederikeHanssen @maxulysse 

Should we? Or split up into two?

Here's a working test of running with no normal (i.e. for a germline or tumoronly). But it'll fail if I include it now because those three outputs are not optional (+I didn't figure out how to insert the [reference.cnn](https://github.com/nf-core/test-datasets/blob/modules/data/generic/cnn/reference.cnn) into the workflow, but that's a detail).
```
include { CNVKIT as CNVKIT_TUMORONLY } from '../../../modules/cnvkit/main.nf' addParams( options: [ 'args': '--reference reference.cnn --method wgs' ] )

workflow test_cnvkit_tumoronly {
    input = [ [ id:'test'], // meta map
              [ file(params.test_data['homo_sapiens']['illumina']['test2_paired_end_sorted_bam'], checkIfExists: true) ],
              [ ]
            ]
    fasta   = file(params.test_data['homo_sapiens']['genome']['genome_fasta'], checkIfExists: true)
    reference   = file(params.test_data['generic']['cnn']['reference'], checkIfExists: true)

    CNVKIT_TUMORONLY ( input, [], [] )
}
```

FriederikeHanssen:
> > Is there a way to make the module inclusive to also run germline only? I mean for strelka for example we settled for having one in the germline and one in a somatic subfloder. this would also be an option
> 
> I guess the problem is that the [input line ](https://github.com/fbdtemme/modules/blob/cnvkit-optional-targetfile/modules/cnvkit/main.nf#L22)would have to be split and have $normalbam be made optional. I think that's not a good solution either. Breaks too much in the tumor/normal workflow, which after all is the main focus of cnvkit (germline is, according to their [documentation](https://cnvkit.readthedocs.io/en/stable/germline.html), a sort of add-on with caveats).
> 
> I think this could argue for having a `cnvkit/<method>` structure already, as you also suggest. Don't call it `cnvkit/batch` though, since `batch` is virtually all the entryways into cnvkit. Maybe `tumor`, `tumor-normal`, and `normal-only`, that's a common use case. What do you think @fbdtemme ?

Hey, I am still thinking about whether it should be called cnvkit_batch or not and from this in the docu I would say, yes: 
Screenshot 2021-11-17 at 14 34 09
https://cnvkit.readthedocs.io/en/stable/pipeline.html. Even if it is rarely used and we might not even have the modules it would stay consistent with other modules like samtools.

lassefolkersen:
> Hey, I am still thinking about whether it should be called cnvkit_batch or not and from this in the docu I would say, yes: Screenshot 2021-11-17 at 14 34 09 https://cnvkit.readthedocs.io/en/stable/pipeline.html. Even if it is rarely used and we might not even have the modules it would stay consistent with other modules like samtools.

Ok - sure, that `batch` part is fine, of course. It's just - for it to be able to _do_ all the things that `batch` can do (tumor-normal, tumor-only, germline), it'll either A) have to make some outputs optional (three of them) or B) split up, such that there is then batch/cnvkit and - say - batchnonormal/cnvkit or something like that. I'm unsure which is best.

FriederikeHanssen:
I don't see a problem with having optional outputs tbh. We have this for several modules. see ehre for example: https://github.com/nf-core/modules/blob/master/modules/ncbigenomedownload/main.nf

lassefolkersen:
> I don't see a problem with having optional outputs tbh. We have this for several modules. see ehre for example: https://github.com/nf-core/modules/blob/master/modules/ncbigenomedownload/main.nf

Ok cool - I'll move it to cnvkit/batch and make those three outputs optional then. Then make some sort of test for that as well 👍

lassefolkersen:
OK - I think it's at the finish line now. It's:
* running as `cnvkit/batch` to resemble e.g. `samtools/sort` and make (easy) room for future expansions
* can handle both tumor-normal setups, as well as germline and tumor-only setup. This includes one [extra test](https://github.com/fbdtemme/modules/blob/8f02f053a72fef462244ff8236a7805c642b0ecf/tests/modules/cnvkit/batch/main.nf#L9-L18), plus required [test-input](https://github.com/nf-core/test-datasets/blob/de06de5a9b4b778cff579437f516f4f64b66a789/data/generic/cnn/reference.cnn) for it. It 'cost' one more [input](https://github.com/fbdtemme/modules/blob/8f02f053a72fef462244ff8236a7805c642b0ecf/modules/cnvkit/batch/main.nf#L25), that however is made [optional](https://github.com/fbdtemme/modules/blob/8f02f053a72fef462244ff8236a7805c642b0ecf/modules/cnvkit/batch/main.nf#L39).
* It's [testing and succeeding](https://github.com/fbdtemme/modules/blob/8f02f053a72fef462244ff8236a7805c642b0ecf/tests/modules/cnvkit/batch/main.nf#L45-L53) with cram-input.
* It's speaking 'murican (tumor-not-tumour) @maxulysse  😄 
* Linting fails, but I think that's a [general thing](https://app.slack.com/client/TE6CZUZPH/CJRH30T6V/thread/CJRH30T6V-1637166578.225900) right now.

Have a look when you have time @FriederikeHanssen

fbdtemme:
Thanks for taking over from me @lassefolkersen.

modules/cnvkit/batch/main.nf
```diff
@@ -19,25 +19,42 @@ process CNVKIT {
     }
 
     input:
-    tuple val(meta), path(tumourbam), path(normalbam)
+    tuple val(meta), path(tumorbam), path(normalbam)
```
FriederikeHanssen, line 22: sorry for being picky, but since it also takes cram file, how about renaming to `tumor_input` etc?

modules/cnvkit/batch/main.nf
```diff
@@ -19,25 +19,42 @@ process CNVKIT {
...
-    tuple val(meta), path("*.cnr"), emit: cnr
-    tuple val(meta), path("*.cns"), emit: cns
+    tuple val(meta), path("*.cnn"), emit: cnn, optional: true
+    tuple val(meta), path("*.cnr"), emit: cnr, optional: true
+    tuple val(meta), path("*.cns"), emit: cns, optional: true
     path "versions.yml"           , emit: versions
 
     script:
+    normal_args = normalbam ? "--normal $normalbam" : ""
+
```
FriederikeHanssen, line 36:
```suggestion
```

modules/cnvkit/batch/main.nf
```diff
@@ -19,25 +19,42 @@ process CNVKIT {
...
+    fasta_args = fasta ? "--fasta $fasta" : ""
+
```
FriederikeHanssen, line 38:
```suggestion
```

modules/cnvkit/batch/main.nf
```diff
@@ -19,25 +19,42 @@ process CNVKIT {
...
+    reference_args = reference ? "--reference $reference" : ""
+
```
FriederikeHanssen, line 40:
```suggestion
```

tests/config/test_data.config
```diff
@@ -263,6 +263,9 @@ params {
             'txt' {
                 hello = "${test_data_dir}/generic/txt/hello.txt"
             }
+            'cnn' {
+                 reference = "${test_data_dir}/generic/cnn/reference.cnn"
```
FriederikeHanssen, line 267: is there a corresponding PR on test-data? Can review that one as well then

tests/modules/cnvkit/batch/main.nf
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env nextflow
+
+nextflow.enable.dsl = 2
+
+include { CNVKIT as CNVKIT_TUMORONLY } from '../../../../modules/cnvkit/batch/main.nf' addParams( options: [ 'args': '--reference reference.cnn --method wgs' ] )
+include { CNVKIT }               from '../../../../modules/cnvkit/batch/main.nf' addParams( options: [ 'args': '--output-reference reference.cnn' ] )
```
FriederikeHanssen, line 6: can you align this on `from` and `addParams`

tests/modules/cnvkit/batch/main.nf
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env nextflow
+
+nextflow.enable.dsl = 2
+
+include { CNVKIT as CNVKIT_TUMORONLY } from '../../../../modules/cnvkit/batch/main.nf' addParams( options: [ 'args': '--reference reference.cnn --method wgs' ] )
```
FriederikeHanssen, line 5: Since the process is in cnvkit/batch it should also be called this: `CNVKIT_BATCH` in the `main.nf` and imported here accordingly

tests/modules/cnvkit/batch/test.yml
```diff
@@ -0,0 +1,101 @@
+- name: cnvkit batch test_cnvkit_tumoronly
+  command: nextflow run tests/modules/cnvkit/batch -entry test_cnvkit_tumoronly -c tests/config/nextflow.config
+  tags:
+    - cnvkit
+    - cnvkit/batch
+  files:
+    - path: output/cnvkit/reference.antitarget-tmp.bed
+      md5sum: d41d8cd98f00b204e9800998ecf8427e
```
FriederikeHanssen, line 8: This is the checksum for emtpy files, if I am not wrong. maybe try with the human reads?

tests/modules/cnvkit/batch/test.yml
```diff
@@ -0,0 +1,101 @@
...
+      md5sum: 26d25ff2d6c45b6d92169b3559c6acdb
+
+- name: cnvkit batch test_cnvkit_hybrid
+  command: nextflow run tests/modules/cnvkit/batch -entry test_cnvkit_hybrid -c tests/config/nextflow.config
+  tags:
+    - cnvkit
+    - cnvkit/batch
+  files:
+    - path: output/cnvkit/baits.antitarget.bed
+      md5sum: d41d8cd98f00b204e9800998ecf8427e
```
FriederikeHanssen, line 19: this is also emtpy md5sum, which sort of makes sense if the same bam files are compared with each other

tests/modules/cnvkit/batch/test.yml
```diff
@@ -0,0 +1,101 @@
...
+      md5sum: aa8a018b1d4d1e688c9f9f6ae01bf4d7
+
+- name: cnvkit batch test_cnvkit_wgs
+  command: nextflow run tests/modules/cnvkit/batch -entry test_cnvkit_wgs -c tests/config/nextflow.config
+  tags:
+    - cnvkit
+    - cnvkit/batch
+  files:
+    - path: output/cnvkit/genome.antitarget.bed
+      md5sum: d41d8cd98f00b204e9800998ecf8427e
```
FriederikeHanssen, line 48: also empty target bed. If it is expected then please ignore my statements. in that case maybe a comment line would help, although it would not get generated with the create-test-yml

FriederikeHanssen (changes requested):
Very nice. Just some minor things, then it is ready to go 🚀

rpetit3:
Heads up, the name in the `meta.yml` will need to be changed to `cnvkit_batch`

tests/modules/cnvkit/batch/test.yml
lassefolkersen: That's fine - it's good to memorize that `d41d8cd98f00b204e9800998ecf8427e`md5sum I guess. It seems with the small datasets we are running in test it's common that the antitarget.bed will be empty. I tried to re-run with a bigger `tumorbam` and then it's not empty anymore. I also tried to run with `'args': '--output-reference reference.cnn --antitargets /home/lasse/test.bed'` i.e. a local antitargets.bed arguments, (just used the one from [sars2](https://github.com/nf-core/test-datasets/blob/de06de5a9b4b778cff579437f516f4f64b66a789/data/genomics/sarscov2/genome/bed/test.bed)). Then the empty file problem disappears, because it doesn't output antitargets anymore. As expected.  So I don't see a good way around it, other than large input test data. Which is not good either.

tests/config/test_data.config
lassefolkersen: No, actually I just pushed it to the test-data repo. Doesn't seem to need code-review. That's why the unit tests are working already.

tests/modules/cnvkit/batch/main.nf
lassefolkersen: 👍

modules/cnvkit/batch/main.nf
lassefolkersen: No worries - it's a good idea. I put tumor and normal. That also better aligns with the input switch names (tumor and normal), Similarly target_file is now targets to match with the `CNVkit` switch name

lassefolkersen:
Yes - ok, I've looked at all the remaining things. I just hit 'resolve' for each of them, @FriederikeHanssen - but do re-open if we need to re-visit. I think that empty file output for `antitargets` is unavoidable though, at least with such small data sets.

drpatelh (approved):
Thanks @lassefolkersen @fbdtemme ! 💪🏽

PR #1036, opened 2021-11-04
Bismark: Add back multi-core usage

ewels:
Adding back the original code to automatically calculate the required flags for parallelisation with `bismark align` and `bismark methylationextractor`

This code is currently in the nf-core/methylation DSL1 pipeline but was removed when the DSL2 module was ported by @phue because the code scared @drpatelh: https://github.com/nf-core/modules/pull/239#discussion_r743230938 (rightly so, I should add).

As I say in a comment there, the code is scary and horrible but in my opinion it is a lesser of evils to keep it. The performance benefits from having it are huge (depending on resources, can be 4-10X faster run time) and the code has been in the production pipeline for a long time so is hopefully pretty stable.

So as I try to push the methylseq towards its DSL2 future, I'm adding it back. I expect a lively discussion 😁 

## PR checklist

- [ ] Add a resource `label`


ewels:
ps. I'm still a DSL2 n00b so have a feeling I'm doing something wrong with parameters here. If someone could set me straight that would be magic 👍🏻

ewels:
ps. To put some numbers to my argument - using multicore can speed up execution time by about 30%. On a process that can often take around 8 days or so to run, that is pretty significant.

JoseEspinosa (changes requested):
I let for you the discussion about the cpus :smile: 
About the `params`, we decided to add them as explicit input channels, see slack discussion [here](https://nfcore.slack.com/archives/CJRH30T6V/p1635410693219300). As an example you can take a look at [this](https://github.com/nf-core/modules/pull/981/files) PR where I refactored `genrich` module to avoid having parameters.

ewels:
Awesome, thanks @JoseEspinosa 👍🏻  So those are mainly for mandatory flags I guess?

Trying to dig into / catch up on all of this DSL2 stuff. Looking at it with fresh eyes today and trying to piece bits together I think I'm starting to understand the argument for keeping the logic for the parallelisation in the pipeline. Basically, we could have the same code there and get that to add the resulting flags to `$options.args`, right?

I'm coming around to this argument now. Some of the code has logic in it which relates to how the _pipeline_ is run, so I can see how that makes sense. Then that keeps the module as simple as possible and generic. No multi-threading for other users of the module, but they could add that themselves from the pipeline.

If @phue / @drpatelh etc think that this is the better approach then I'm happy. Main thing is that I don't want to lose the multithreading from the pipeline.

modules/bismark/methylationextractor/main.nf
```diff
@@ -31,6 +31,13 @@ process BISMARK_METHYLATIONEXTRACTOR {
     path "versions.yml"                            , emit: versions
 
     script:
+    // Assign sensible numbers for multicore and buffer_size based on bismark docs
+    def ccore = task.cpus ? ((task.cpus as int) / 3) as int : 1
+    def multicore = (ccore > 1) ? "--multicore ${ccore}" : ""
+    // Only set buffer_size when there are more than 6.GB of memory available
+    def mbuffer = task.memory ? (task.memory as nextflow.util.MemoryUnit) - 2.GB : (4.GB).toBytes()
+    def buffer = (mbuffer.compareTo(4.GB) == 1) ? "--buffer_size ${mbuffer.toGiga()}G" : ""
```
phue, line 39:
```suggestion
    def buffer = (task.memory.giga >= 6) ? "--buffer_size ${task.memory.giga - 2}G" : ""
```

Do we really need to check if `task.memory` is TRUEish? Are there cases where it can be `null`?
Unless I'm missing something obvious, this code could be simplified further.

ewels: Yeah it can be `null`. Didn't use to have this until someone came along with a cluster that doesn't allow job submissions asking for memory (only cpus).

phue: I see..
This should work though (safe navigation):

```suggestion
    def buffer = (task.memory?.giga > 6) ? "--buffer_size ${task.memory.giga - 2}G" : ""
```

modules/bismark/methylationextractor/main.nf
```diff
@@ -31,6 +31,13 @@ process BISMARK_METHYLATIONEXTRACTOR {
     path "versions.yml"                            , emit: versions
 
     script:
+    // Assign sensible numbers for multicore and buffer_size based on bismark docs
+    def ccore = task.cpus ? ((task.cpus as int) / 3) as int : 1
+    def multicore = (ccore > 1) ? "--multicore ${ccore}" : ""
```
phue, line 36:
```suggestion
    def multicore = (task.cpus >= 6) ? "--multicore ${(task.cpus / 3) as int}" : ""
```

modules/bismark/methylationextractor/main.nf
ewels: Nope, some configs [explicitly overwrite to give `null`](https://github.com/nf-core/configs/blob/0866682e9b38acf8cc56b087ed21eb2ccb841cfb/conf/hebbe.config#L16-L17) as it's the only way to get the pipeline to run with their job scheduler.

But yeah, your suggestion is more succinct 🙂

modules/bismark/methylationextractor/main.nf
ewels: Same as memory, this will break if cpus is `null` right? Not that I've ever actually seen that happen in the wild.

phue: Even if it happens, `(null >= 6)` evaluates to `false`

modules/bismark/align/main.nf
```diff
@@ -29,14 +29,46 @@ process BISMARK_ALIGN {
...
 
     script:
+    // Try to assign sensible bismark memory units according to what the task was given
+    def ccore = 1
+    def cpu_per_multicore = 3
+    def mem_per_multicore = (13.GB).toBytes()
+
+    if( task.cpus ){
+        // Numbers based on recommendation by Felix for a typical mouse genome
+        if( params.single_cell || params.zymo || params.non_directional ){
```
maxulysse, line 39: not sure about the params usage here

modules/bismark/align/main.nf
```diff
@@ -29,14 +29,46 @@ process BISMARK_ALIGN {
...
+        // Check that we have enough memory, assuming 13GB memory per instance (typical for mouse alignment)
+        try {
+            def tmem = (task.memory as nextflow.util.MemoryUnit).toBytes()
+            def mcore = (tmem / mem_per_multicore) as int
+            ccore = Math.min(ccore, mcore)
+        } catch (all) {
+            log.warn "Not able to define bismark align multicore based on available memory"
+        }
+    }
+    def multicore  = (ccore > 1) ? "--multicore ${ccore}" : ''
```
grst, line 61: Why not just `--multicore ${task.cpus}` and leave it to the pipeline developer to assign appropriate memory resources?

ewels: Because the Bismark flags don't correspond with actual usage. So if you want to use 8 cores then you need to specify a different number on the command line.

The calculation to reach this number is (mostly) specific to Bismark and not pipeline / sample specific. So if we can handle that complexity in the module it standardises in and means the pipeline developer / user only needs to think about cpus used (as is normal).

In theory.

mahesh-panchal:
Is there any reason not to move that to the `modules.config` of the workflow and perhaps include a comment in the module itself to pointing to the example of how to optimize parameters here?

mahesh-panchal:
Perhaps the meta.yml for modules could have a special notes field that has misc information like optimization notes, and this gets printed to screen when `nf-core modules install/update` is used.

ewels:
Mostly because of [this code](https://github.com/nf-core/modules/pull/1036/files#diff-8bfa33cf609bea1129b46592ce3dbfdeb8726879f23e827bbadcc8d9654e07d1R42-R51):

```nextflow
        // How many multicore splits can we afford with the cpus we have?
        ccore = ((task.cpus as int) / cpu_per_multicore) as int
        // Check that we have enough memory, assuming 13GB memory per instance (typical for mouse alignment)
        try {
            def tmem = (task.memory as nextflow.util.MemoryUnit).toBytes()
            def mcore = (tmem / mem_per_multicore) as int
            ccore = Math.min(ccore, mcore)
        } catch (all) {
            log.warn "Not able to define bismark align multicore based on available memory"
        }
```

which I didn't really want to have to duplicate at pipeline level. It's more specific to bismark itself and could be reused anywhere that the module is being reused.

It would be easy enough to have defaults for `mem_per_multicore` and `cpu_per_multicore` which could be overwritten at module level according to the parameters being used for bismark, but it just seemed a bit hacky to move a big code block out into the pipeline when it would be the same every time it is used.

Is there any reason not to use `ext.mem_per_multicore` and `ext.cpu_per_multicore`?

maxulysse:
I kinda like the `ext.*_per_multicore` idea

mahesh-panchal:
> Is there any reason not to use `ext.mem_per_multicore` and `ext.cpu_per_multicore`?

My opinion is that we should limit the creation/use of `ext` fields. They should only be used when they apply to a wide proportion of the modules. I would argue that the creation of process specific fields is bad for standardization. 

The way I would then suggest solving this is to check the `ext.args` for the presence of `--multicore` and if it's not there, set it using the code above. The parts set by parameter should still be in the pipeline modules.config.

mahesh-panchal:
I'm not understanding something then. Why is this part defined as it is?
```
    def ccore = 1
    def cpu_per_multicore = 3
    def mem_per_multicore = (13.GB).toBytes()
```
I was assuming that this was based off the default resource settings given by the label, but there's something else that governs this ( and how should a user know this )?

ewels:
It's because setting `--multicore` for Bismark sets how many cpus should be used for the aligner it wraps, and not Bismark itself. Bismark runs multiple alignments, so that's why for every `--multicore` 3 cpus are needed. Unless `(params.single_cell || params.zymo || params.non_directional )` is set, in which case even more alignments are being done and you need even more cpus per `--multicore` (and more memory). These numbers have been empirically determined through a lot of discussion and trial and error 🥵 

Explaining this has been a great rubber ducking exercise though! 🦆 (sorry). Going back to the [original DSL1 code](https://github.com/nf-core/methylseq/blob/03972a686bedeb2920803cd575f4d671e9135af0/main.nf#L503) I see that these options are all setting the `--non_directional` flag in the Bismark command:

```nextflow
non_directional = params.single_cell || params.zymo || params.non_directional ? "--non_directional" : ''
```

So I could presumably try to detect this in the `task.args` and this would be enough to inform the logic for determining the resources to set.

I could also check `task.args` for `--multicore` and not set automatically if it is being set in the command.

Doing this should allow the "sensible defaults" logic and also allow end users to overwrite. Bingo! ✨

ewels:
ok, it's not perfect as you can't adjust the logic for the _calculation_ from the pipeline. But you can just do that calculation locally and set `--multicore` yourself from the pipeline if it's a problem so I think it's a good compromise 👍🏻

ewels:
ok, hopefully ready for re-review and merge now 🙏🏻 

Thanks for the review + feedback everyone! Let's hope that it still actually works 😅

modules/nf-core/bismark/align/main.nf
```diff
@@ -22,14 +22,42 @@ process BISMARK_ALIGN {
 
     script:
     def args = task.ext.args ?: ''
-    def prefix = task.ext.prefix ?: "${meta.id}"
```
mahesh-panchal, line 25: `bismark` has a `--prefix` flag so this should be kept.

modules/nf-core/bismark/methylationextractor/main.nf
```diff
@@ -24,16 +24,25 @@ process BISMARK_METHYLATIONEXTRACTOR {
 
     script:
     def args = task.ext.args ?: ''
```
mahesh-panchal, line 26: Should there be a way to prefix the output files here?

mahesh-panchal:
> rubber ducking exercise

I've never heard of this before: https://en.wikipedia.org/wiki/Rubber_duck_debugging

modules/nf-core/bismark/methylationextractor/main.nf
```diff
@@ -24,16 +24,25 @@ process BISMARK_METHYLATIONEXTRACTOR {
...
+    // Assign sensible numbers for multicore and buffer_size based on bismark docs
+    if(!args.contains('--multicore') && task.cpus >= 6){
+        args += "--multicore ${(task.cpus / 3) as int}"
```
ewels, line 29:
```suggestion
        args += " --multicore ${(task.cpus / 3) as int}"
```

modules/nf-core/bismark/methylationextractor/main.nf
```diff
@@ -24,16 +24,25 @@ process BISMARK_METHYLATIONEXTRACTOR {
...
+    }
+    // Only set buffer_size when there are more than 6.GB of memory available
+    if(!args.contains('--buffer_size') && task.memory?.giga > 6){
+        args += "--buffer_size ${task.memory.giga - 2}G"
```
ewels, line 33:
```suggestion
        args += " --buffer_size ${task.memory.giga - 2}G"
```

modules/nf-core/bismark/methylationextractor/main.nf
ewels: I don't really understand what you mean.. Bismark already sets the output files with a prefix based on the input filenames..

modules/nf-core/bismark/align/main.nf
ewels: Never noticed it before, will add back 👍🏻

modules/nf-core/bismark/methylationextractor/main.nf
ewels: Sorry, read your comments in the wrong answer. I'll check..

ewels: Can't see an option for this in Bismark (see [source](https://github.com/FelixKrueger/Bismark/blob/430df15af5b52686487d4eeb6a97d1b2390bbf96/bismark_methylation_extractor#L959-L994)).

There is `--output_dir`, should I use the prefix for that if set?

mahesh-panchal: That would mess with the output capture since it doesn't recurse down. 

I guess the question on whether this should have a prefix depends on if the workflow will process groups of samples and if the files are passed along for further batch processing, will this then create a filename collision? If this is a possibility, then the files might need to be renamed based on the prefix. If the tool should not be used like this, then just leave it out.

ewels: I don't think it'll be a problem - extraction always happens after deduplication so on complete samples. As such, I've not had to worry about filename collisions in the past.

mahesh-panchal (approved):

ewels:
> 

Yeah!! That's why we did them as swag in the latest hackathon 🦆 
https://twitter.com/nf_core/status/1585941877571796992?s=20&t=XIKmIzMibKuILmRQdUzuyA
FgJlHHHXwAEdn8h

Emiller88 (approved):
LGTM! :cowboy_hat_face:

mahesh-panchal:
> > 
> 
> Yeah!! That's why we did them as swag in the latest hackathon 🦆 https://twitter.com/nf_core/status/1585941877571796992?s=20&t=XIKmIzMibKuILmRQdUzuyA FgJlHHHXwAEdn8h

Dang. I want some of those for my kids, and the apron for myself. Ah well.

PR #1077, opened 2021-11-16
feat: Add main.nf

Emiller88:
Allows for launching of subworkflows


mashehu (approved):

rpetit3:
Heads up this breaks `nf-core modules create` since it checks for a `main.nf` file

https://github.com/nf-core/tools/blob/dev/nf_core/modules/create.py#L284-L288

PR #1078, opened 2021-11-16
Launching one-off subworkflows

Emiller88:


Emiller88:
Anyone have any ideas how we can pass the input channels in?

```log
Workflow `ALIGN_BOWTIE2` declares 2 input channels but 0 were specified
```

mahesh-panchal:
> Anyone have any ideas how we can pass the input channels in?
> 
> ```log
> Workflow `ALIGN_BOWTIE2` declares 2 input channels but 0 were specified
> ```

I've tried several variations and so far nothing works. I came to the conclusion it wasn't possible to pass anything to `take`, except by using another script.

mahesh-panchal:
Can this work though?
Consider if this snippet was at the bottom of `main.nf`:
```
workflow {
    
    ALIGN_BOWTIE2( 
        params.reads // channel: [ val(meta), [ reads ] ]
        params.index // channel: /path/to/bowtie2/index/
    )
}
```
How would one tell Nextflow the string path is actually a `file` path?

Emiller88:
I have a feeling we need something like https://www.nextflow.io/docs/edge/dsl2.html#module-templates
And then we can call a template that handles the file input params, or maybe something that just takes a samplesheet and outputs a meta map like channel, and then we pass in the subworkflow name in the template.

mahesh-panchal:
Doesn't that mean the process has to take care of the file staging Nextflow usually does?
Otherwise what's the difference to writing another anonymous workflow which calls the named workflow?

Emiller88:
That was all that I was imagining the template would be is an anonymous workflow, that might hook it up to a samplesheet module or something. Your snippet might work with `fromfilepairs` and maybe some magic to turn the name into a `meta.id`

maxulysse:
Shall we close this one, as I'm assuming it's no longer relevant.
@Emiller88 @mahesh-panchal

mahesh-panchal:
Fine by me.

Emiller88:
I think it's even more relevant now that we've actually added subworkflows

matthdsm:
Hi there!

Cheers
the nf-core maintainers

Emiller88:
Closing in favor of https://github.com/nf-wrap

PR #1095, opened 2021-11-24
ci: Skip subworkflows

Emiller88:
Punting for now on these. We can still accept PRs to start collecting them, but they'll all need to be cleaned up eventually with the new syntax.


drpatelh:
Thanks! That's ok. I fixed it for now on my fork where we are making the mass changes. By just brute force commenting stuff out. Problem is that even when testing locally the `tags` will be found by `pytest` and still run the subworkflow tests.

grst:
so we still need this?

Emiller88:
Nope I think we're good!

PR #1110, opened 2021-12-02
Change syntax from task.ext.suffix to tast.ext.prefix in all modules

drpatelh:
x-ref: https://github.com/nf-core/tools/pull/1347


tests/modules/dshbio/splitgff3/nextflow.config
```diff
@@ -3,7 +3,7 @@ process {
     publishDir = { "${params.outdir}/${task.process.tokenize(':')[-1].tokenize('_')[0].toLowerCase()}" }
 
     withName: DSHBIO_SPLITGFF3 {
-        ext.suffix = '.'
+        ext.prefix = { "${meta.id}." }
```
jfy133, line 6: Maybe this should be kept just as `.`? I gues it's specifying an ourput directory?

tests/modules/dshbio/splitbed/nextflow.config
```diff
@@ -3,7 +3,7 @@ process {
     publishDir = { "${params.outdir}/${task.process.tokenize(':')[-1].tokenize('_')[0].toLowerCase()}" }
 
     withName: DSHBIO_SPLITBED {
-        ext.suffix = '.'
+        ext.prefix = { "${meta.id}." }
```
jfy133, line 6: Maybe this should be kept just as `.`? I gues it's specifying an ourput directory?

jfy133 (approved):
Only two queries, where I think output prefix is trying specify the output directory rather than with `$prefix` itself

tests/modules/dshbio/splitgff3/nextflow.config
drpatelh: Nope. So it's actually the prefix for any output files but the `.` is required here to get for example to get `test.1.bed.gz` instead of `test1.bed.gz`. 

https://github.com/nf-core/modules/blob/1f8f86b79361fb9999b56680105f439df81d718d/tests/modules/dshbio/splitbed/test.yml#L7

tests/modules/dshbio/splitbed/nextflow.config
drpatelh: See above.

tests/modules/dshbio/splitgff3/nextflow.config
jfy133: Ah understood :+1:

PR #1115, opened 2021-12-06
Add bcl2fastq2 to cellranger dockerfile

RHReynolds:
Co-authored-by: Edmund Miller <edmund.a.miller@gmail.com>
Co-authored-by: Gisela Gabernet <gisela.gabernet@gmail.com>

* build(cellranger): add bcl2fastq2 to dockerfile and update cellranger version
* refactor(cellranger): update module containers to reflect updated docker image
* fix(cellranger): fix output path for mkfastq module
* docs(cellranger): document how to update bcl2fastq
* chore: add bcl2fastq2 file to .gitignore

## PR checklist

    - [X] `PROFILE=docker pytest --tag <MODULE> --symlink --keep-workflow-wd`
    - [X] `PROFILE=singularity pytest --tag <MODULE> --symlink --keep-workflow-wd`


Emiller88 (changes requested):
Awesome work! Just need to have @ggabernet push the new image, and we will probably need to find a new test-dataset since the one cellranger provides is pretty big. Actually we could probably get away with `--tiles` if we can pass it through mkfastq :thinking:

RHReynolds:
That's a really good idea re. using `--tiles`! According to the cellranger help page, `cellranger mkfastq` is able to accept `bcl2fastq` arguments. And as the mkfastq module is set up to take external arguments, it should be possible :)

Emiller88:
Missed referencing this in the commit message https://github.com/brwnj/bcl2fastq/blob/73af6199eae55bf12c481d4bf367899d15a930c8/bcl2fastq.py#L186

ggabernet:
@RHReynolds and @Emiller88 I've added some edits to the `Dockerfile` and `Readme`, as discussed on slack it would be great to have a cellranger container with and without the bcl2fastq. So I've build and pushing now both containers to the nf-core Dockerhub already

RHReynolds:
@ggabernet, thanks so much for building the containers. Just made one small change, which was to update the container used in mkfastq.

RHReynolds:
I updated some of the md5sums in the tests, which had changed following the cellranger version update. I wonder if this is something we should mention in the README? 

@Emiller88 - the mkfastq test using `--tiles 1105` didn't work for me. According to the bcl2fastq error this is because the tile is not found. We can correct the error by using `--tiles 1101`, but looking at the files available in `cellranger-tiny-bcl-1.2.0/Data/Intensities/L001/` it looks like 10x are only providing one tile anyway (`s_1_1101.clocs`). In other words, even with the tile workaround, the test dataset will remain the same size. So we're back to finding a new test dataset for mkfastq 🤔

Emiller88:
> I updated some of the md5sums in the tests, which had changed following the cellranger version update. I wonder if this is something we should mention in the README?

Nah, that's expected, probably just something like the versions being printed out. 

>  it looks like 10x are only providing one tile anyway (`s_1_1101.clocs`). In other words, even with the tile workaround, the test dataset will remain the same size. So we're back to finding a new test dataset for mkfastq thinking

That's unfortunate. I'll think on in, I think we can write the tests and just leave them commented out with a `FIXME` in the pytest_workflow.

ggabernet (commented):
Looks good to me so far! Just regarding the tests not passing for the mkfastq module, would it make sense to use a similar strategy as for `cellranger count` and use a `stub` run until there is a better solution?

RHReynolds:
> Looks good to me so far! Just regarding the tests not passing for the mkfastq module, would it make sense to use a similar strategy as for `cellranger count` and use a `stub` run until there is a better solution?

Had to look up what a `stub` run is -- but nice! Yes, absolutely makes sense to use a `stub` run. Looking at the `cellranger count` example, I can see there is a `stub` block in `modules/cellranger/count/main.nf`, but the `/tests/modules/cellranger/count/test.yml` does not include `-stub-run` flag in the `command`. Is this because a solution was found? I just want to check that in the mkfastq situation we would want to include `-stub-run` in the mkfastq test.yml?

ggabernet:
> > Looks good to me so far! Just regarding the tests not passing for the mkfastq module, would it make sense to use a similar strategy as for `cellranger count` and use a `stub` run until there is a better solution?
> 
> Had to look up what a `stub` run is -- but nice! Yes, absolutely makes sense to use a `stub` run. Looking at the `cellranger count` example, I can see there is a `stub` block in `modules/cellranger/count/main.nf`, but the `/tests/modules/cellranger/count/test.yml` does not include `-stub-run` flag in the `command`. Is this because a solution was found? I just want to check that in the mkfastq situation we would want to include `-stub-run` in the mkfastq test.yml?

yes, exactly, in that case we would include the `-stub-run` flag in the test command.

RHReynolds:
I've added in what I think would be a suitable stub run, but let me know if there's something missing :)

Emiller88 (approved):
Awesome work!

ggabernet (approved):
Looks great, thanks a lot for your contribuition @RHReynolds

RHReynolds:
> Looks great, thanks a lot for your contribuition @RHReynolds

You're very welcome! And thank you and @Emiller88 for all the patience answering my questions! Great learning curve for me :)

PR #1117, opened 2021-12-06
Replace remaining task.ext.suffix with task.ext.prefix

mahesh-panchal:
Fixes remaining modules that still had `task.ext.suffix` to `task.ext.prefix`.

### DIFF
--- a/modules/gffread/main.nf
+++ b/modules/gffread/main.nf
@@ -15,8 +15,8 @@ process GFFREAD {
     path "versions.yml" , emit: versions
 
     script:
-    def args = task.ext.args ?: ''
-    def prefix = task.ext.suffix ? "${gff.baseName}${task.ext.suffix}" : "${gff.baseName}"
+    def args   = task.ext.args   ?: ''
+    def prefix = task.ext.prefix ?: "${gff.baseName}"
     """
     gffread \\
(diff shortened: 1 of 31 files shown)


modules/gffread/main.nf
```diff
@@ -15,8 +15,8 @@ process GFFREAD {
     path "versions.yml" , emit: versions
 
     script:
-    def args = task.ext.args ?: ''
-    def prefix = task.ext.suffix ? "${gff.baseName}${task.ext.suffix}" : "${gff.baseName}"
+    def args   = task.ext.args   ?: ''
+    def prefix = task.ext.prefix ?: "${gff.baseName}"
```
jfy133, line 19:
```suggestion
    def prefix = task.ext.prefix ?: "${gff.baseName}"
```

Is def needed here? Not used in the other modules it looks like

jfy133 (approved):

modules/gffread/main.nf
mahesh-panchal: In all the other modules (that have already been updated) `prefix` is localized to the script context. It's only unlocalized when used in the output.

jfy133: Ok !

PR #1125, opened 2021-12-07
Add new boolean sort_bam as input to be able to chose between sort and view in bwamem and bwamem2mem

maxulysse:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`

### DIFF
--- a/modules/bwa/mem/main.nf
+++ b/modules/bwa/mem/main.nf
@@ -27,7 +27,7 @@ process BWA_MEM {
     bwa mem \\
         $args \\
         $read_group \\
-        -t $task.cpus \\
+        -t ${split_cpus} \\
         \$INDEX \\
         $reads \\
         | samtools $args2 --threads ${split_cpus} -o ${prefix}.bam -
--- a/modules/bwa/mem/main.nf
+++ b/modules/bwa/mem/main.nf
@@ -21,7 +21,6 @@ process BWA_MEM {
     def args2 = task.ext.args2 ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
     def read_group = meta.read_group ? "-R ${meta.read_group}" : ""
-    def split_cpus = Math.floor(task.cpus/2)
     def samtools_command = sort_bam ? 'sort' : 'view'
     """
     INDEX=`find -L ./ -name "*.amb" | sed 's/.amb//'`
@@ -29,10 +28,10 @@ process BWA_MEM {
     bwa mem \\
         $args \\
         $read_group \\
-        -t ${split_cpus} \\
+        -t $task.cpus \\
         \$INDEX \\
         $reads \\
-        | samtools $samtools_command $args2 --threads ${split_cpus} -o ${prefix}.bam -
+        | samtools $samtools_command $args2 --threads $task.cpus -o ${prefix}.bam -
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
--- a/modules/bwa/mem/main.nf
+++ b/modules/bwa/mem/main.nf
@@ -20,6 +20,7 @@ process BWA_MEM {
     def args2 = task.ext.args2 ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
     def read_group = meta.read_group ? "-R ${meta.read_group}" : ""
+    def split_cpus = Math.floor(task.cpus/2)
     """
     INDEX=`find -L ./ -name "*.amb" | sed 's/.amb//'`
 
@@ -29,7 +30,7 @@ process BWA_MEM {
         -t $task.cpus \\
         \$INDEX \\
         $reads \\
-        | samtools view $args2 -@ $task.cpus -bhS -o ${prefix}.bam -
+        | samtools $args2 --threads ${split_cpus} -o ${prefix}.bam -
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
--- a/modules/bwa/mem/main.nf
+++ b/modules/bwa/mem/main.nf
@@ -10,6 +10,7 @@ process BWA_MEM {
     input:
     tuple val(meta), path(reads)
     path  index
+    val   sort_bam
 
     output:
     tuple val(meta), path("*.bam"), emit: bam
@@ -21,6 +22,7 @@ process BWA_MEM {
     def prefix = task.ext.prefix ?: "${meta.id}"
     def read_group = meta.read_group ? "-R ${meta.read_group}" : ""
     def split_cpus = Math.floor(task.cpus/2)
+    def samtools_command = sort_bam ? 'sort' : 'view'
     """
     INDEX=`find -L ./ -name "*.amb" | sed 's/.amb//'`
 
@@ -30,7 +32,7 @@ process BWA_MEM {
         -t ${split_cpus} \\
         \$INDEX \\
         $reads \\
-        | samtools $args2 --threads ${split_cpus} -o ${prefix}.bam -
+        | samtools $samtools_command $args2 --threads ${split_cpus} -o ${prefix}.bam -
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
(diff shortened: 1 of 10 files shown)


modules/bwa/mem/main.nf
```diff
@@ -29,7 +30,7 @@ process BWA_MEM {
         -t $task.cpus \\
         \$INDEX \\
         $reads \\
-        | samtools view $args2 -@ $task.cpus -bhS -o ${prefix}.bam -
+        | samtools $args2 --threads ${split_cpus} -o ${prefix}.bam -
```
mahesh-panchal, line 33: What's the need here to remove `view`? I don't understand the context. What's the difference between the #raredisease context and the #sarek context?

maxulysse: I remove view because in sarek we do need sorted bams, I assume they need the same in raredisease

modules/bwa/mem/main.nf
```diff
@@ -20,6 +20,7 @@ process BWA_MEM {
     def args2 = task.ext.args2 ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
     def read_group = meta.read_group ? "-R ${meta.read_group}" : ""
+    def split_cpus = Math.floor(task.cpus/2)
```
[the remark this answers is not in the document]
maxulysse: we add issues with task.cpus because both bwa mem and samtools are requesting the same amount of cpus, so you end up eating more cpus than requested, we had a big discussion about it somewhere, and we ended up doing that in sarek

modules/bwa/mem/main.nf
mahesh-panchal: I'm lost. Isn't that what the `SAMTOOLS_SORT` subworkflow is for?

maxulysse: Yes, but I can do that here in the same process if I change this part.
Data footprint if greatly reduced this way, and it's also more time/cpu efficient

mahesh-panchal: Is there any reason not to have `sort` there instead of `view`?

modules/bwa/mem/main.nf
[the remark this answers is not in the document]
FriederikeHanssen: hm here is some related discussion https://nfcore.slack.com/archives/CJRH30T6V/p1632822976224200
almost seems like it might be good to overthink this line as far as i understand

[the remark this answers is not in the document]
FriederikeHanssen: in that case the documentation for multitool modules would need to be updated to

modules/bwa/mem/main.nf
maxulysse: @drpatelh ?

Emiller88:
Can we rope dragmap in on this? I'm pretty sure bowtie does something similar as well.

mahesh-panchal:
> Can we rope dragmap in on this? I'm pretty sure bowtie does something similar as well.

It's probably a good idea to change any occurrence of `samtools view` to `samtools sort` in a module.

maxulysse:
Updated with new logic after conversation with @drpatelh.
cc @mahesh-panchal @FriederikeHanssen @Emiller88

modules/bwa/mem/meta.yml
```diff
@@ -32,6 +32,10 @@ input:
         type: file
         description: BWA genome index files
         pattern: "Directory containing BWA index *.{amb,ann,bwt,pac,sa}"
+    - sort_bam:
+        type: boolean
+        description: use samtools sort or samtools view
```
mahesh-panchal, line 37:
```suggestion
        description: use samtools sort (true) or samtools view (false)
```

modules/bwa/mem/main.nf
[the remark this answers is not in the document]
drpatelh: Ping @pontus for a once over.

tests/modules/bwa/mem/nextflow.config
```diff
@@ -2,4 +2,8 @@ process {
 
     publishDir = { "${params.outdir}/${task.process.tokenize(':')[-1].tokenize('_')[0].toLowerCase()}" }
 
+    withName: BWA_MEM {
+        ext.args2 = { sort_bam ? "" : "-bh" }
```
mahesh-panchal, line 6: Why is this not the default for both?

maxulysse: samtools view and samtools sort don't have the same options

FriederikeHanssen (commented):
Looks good to me. Just the split cpus thing is still a bit unclear. the point made in the slack discussion, that just splitting by 2 is a bit arbitrary and not transparent to the user when assigning ressources to a process struck a chord with me. but I am no expert in this matter and maybe someone else could weigh in. Maybe @pontus?

tests/modules/bwa/mem/nextflow.config
maxulysse: output is bam by default for example

maxulysse: (and decided by the output name)

mahesh-panchal: Ah, right, and `sort` automatically autodetects the format from the file ending right?

modules/bwa/mem/main.nf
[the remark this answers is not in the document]
pontus: Without actually having checked how they use these threads, the usage that makes most sense to me is to let the system handle things (i.e. `-@ task.cpus`).

There has (as noted by Friederike) been discussions about broken systems where the batch system kills jobs that uses too much CPU rather than restricting how much they can use to start with.

I believe those are very few, and it makes more sense to me to address in other ways if it happens (how would depend on what's going on).

It's also _theoretically_ possible that job performance degrades by telling it to use more threads (e.g. because of CPU cache exhaustion leading to stalls, limited memory bandwidth leading to stalls),  but a sensible basic approach is 
 to not limit tools if not needed and troubleshoot cases where that's visibly suboptimal. Modern operating systems are very good at handling thousands of threads in a good way.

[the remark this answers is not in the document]
maxulysse: ok, then I remove all that, and we'll figure it out later on if we need that or not

drpatelh (approved):
Nice!

FriederikeHanssen:
I think I missed the discussion onslack on the resolution of split_cpus. Basically It should not be handeled inside the module anymore at all?

mahesh-panchal:
> I think I missed the discussion onslack on the resolution of split_cpus. Basically It should not be handeled inside the module anymore at all?

The discussion was here in one of the resolved comments. But basically it was to let the OS handle CPU allocation, and when something goes wrong handle it on a case by case basis.

pontus:
I can do a short recap for the benefit of those not in slack also.

I'll generally exclude things that can appear like magic, like caches, memory bandwidth and NUMAs, but to show how things can be, it's possible to have e.g. a single threaded tool that runs in a tight loop that uses memory fully within the CPU cache to an extent that nothing else fits. Adding a similar second thread for another data area on a separate core will lead to a lot of cache evictions and stalls, probably leading to (much) worse performance than provided by the single thread.

I'll also assume it's not exceptionally bad software (e.g. doing non-blocking IO with busy retry).

With that said, the best is almost always to let things handle themselves and trust in the operating system. Assuming a system with a producer and a consumer where each consumer core can consume what a producer core can produce, a 50/50 split makes sense and the system will see to that automatically.

If the producer needs the times the CPU resources to produce a "work unit" that consumer needs to consume it, a 50/50 split doesn't make sense - we'll end up with half the cores used for the producer, but  then only roughly 1/6th of the cores used for the  consumer, essentially wasting 1/3 of the resources. If  the system is allowed to handle it, one will have full usage and a 75/25 split.

That also doesn't go into the fact that typically it will take some work for a producer to produce something that can be consumed (where letting the system handle it will let the producer use all resources wheras a fixed split limits to the resources specified by the split), and obviously, the reverse case when the producer has finished and only the consumer remains.

PR #1150, opened 2021-12-09
Adding new module smoothxg

### LINKED ISSUE #1133: new module: smoothxg
I think it would be good to have a module for [smoothxg](https://bioconda.github.io/recipes/smoothxg/README.html)

https://github.com/nf-core/pangenome/issues/70

heuermh:
Fixes #1133 

## PR checklist

Closes #XXX 

- [x] Add a resource `label`


heuermh:
`process smoothxg` in nf-core/pangenome provides a lot more functionality than this minimal module does; optional consensus graph and MAF outputs, running more than one iteration, and much configuration.  We'll need to determine how much of that goes in a shared module and how much is specific to the nf-core/pangenome workflow.  Although to be honest that would be the only target workflow at this point.

subwaystation:
As the pangenome workflow is the only target workflow, I would vote to design this module for it. 
I would suggest to also add optional outputs for the consensus graphs and the MAF file. See https://github.com/nf-core/pangenome/blob/676b05103c4d3fc6cb448b1d74eb3520389aa495/main.nf#L171.

Anyhow, this should wait for https://github.com/nf-core/pangenome/blob/676b05103c4d3fc6cb448b1d74eb3520389aa495/main.nf#L171.

JoseEspinosa:
Hi @heuermh, thank you for this PR! We are merging as many modules as possible now due to and impending restructuring of the entire repo. This will mean you will need to update the module to reflect these changes before it can be merged in the future. It appears like this module isn't ready to be merged so if applicable, we are converting it to draft and adding the WIP label. If this isn't the case please let us know and we will try to get the module in before the changes. Thanks again!

modules/smoothxg/main.nf
```diff
@@ -0,0 +1,35 @@
+def VERSION = '0.6.1' // Version information not provided by tool on CLI
```
subwaystation, line 1: Please update to v0.6.5. https://anaconda.org/bioconda/smoothxg. This should also come with a parameter which prints out the version of `smoothxg`.

modules/smoothxg/main.nf
```diff
@@ -0,0 +1,35 @@
...
+
+    output:
+    tuple val(meta), path("*.gfa"), emit: gfa
+    path "versions.yml"           , emit: versions
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    smoothxg \\
```
subwaystation, line 24: I am voting for a more complex script, which ensures we can run several iterations of `smoothxg`: https://github.com/nf-core/pangenome/blob/571d649719db87ffc7769eb0b3f526df99cf1e7a/main.nf#L269-L343. 
It might be possible to port some Bash magic to Groovy, but I doubt it will be much.

tests/modules/smoothxg/main.nf
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env nextflow
+
+nextflow.enable.dsl = 2
+
+include { SMOOTHXG } from '../../../modules/smoothxg/main.nf'
+
+workflow test_smoothxg {
```
subwaystation, line 7: Please add another workflow test using the pangenomics test data: https://github.com/nf-core/test-datasets/blob/modules/data/pangenomics/homo_sapiens/pangenome.seqwish.gfa. Thanks!

modules/smoothxg/main.nf
heuermh: Cool, will do!

modules/smoothxg/main.nf
heuermh: I seem to recall a new Nextflow feature that might help with this, will look into it

heuermh: https://github.com/nextflow-io/patterns/blob/master/docs/feedback-loop.adoc

modules/smoothxg/main.nf
heuermh: Updated, will still need to add the bit to ask smoothxg to provide its own version

subwaystation: Ironically, it's written to stdout:

```
smoothxg --version 2>&1 | cut -f 1 -d '-' | cut -f 2 -d 'v'
```

Will do the trick.

subwaystation: https://github.com/pangenome/smoothxg/issues/171

modules/smoothxg/main.nf
subwaystation: It was also discussed in https://github.com/nextflow-io/nextflow/discussions/2521#discussioncomment-2374495.

So the idea would be to code the whole thing in a subworkflow. Which means we leave this module as it is?

subwaystation: What is still unclear to me is if the output of the previous iteration really is the input of the current iteration using the recursion feature.

subwaystation:
Recurse does not work, because we also emit the optional MAF output which is not declared as input for the module.

So we would have to recurse n-1 times. Then we have to invoke the module one last time setting the MAF output. Would that be possible?

subwaystation:
https://github.com/pangenome/smoothxg/releases/tag/v0.6.7 new version incoming!

subwaystation:
> 

I don't know if there are any other solutions @heuermh? Edge cases are always a huge pain in Nextflow.
I would vote to just stay in `bash` for this particular module.

subwaystation:
Alright @heuermh. There are 3 ways forward:

1. We ignore the nf-core rules and just blindly copy the bash script from PGGB.
2. We write 2 modules. The first one is ignoring the nf-core modules and does not output a versions.yml. The second one works as the smoothxg module works now. 
3. As suggested by @ggabernet modify the tool smoothxg itself so that the iterations are possible when invoking it from the command line. I talked to @AndreaGuarracino and @ekg and this is a possible way forward. If we are lucky, we save I/O.

I would pick number 3, but it will take some coding and testing from my side. Stay tuned!

heuermh:
By option 1 you mean the [for loop in bash from earlier](https://github.com/nf-core/pangenome/blob/571d649719db87ffc7769eb0b3f526df99cf1e7a/main.nf#L269-L343)?  That could work.  I tried option 2 a few different ways without success.  I would also prefer option 3.

subwaystation:
Yes, I meant the bash loop from earlier. Hopefully option 3 will fly in next week. Else we can just push forward with option 1 and correct later.

AndreaGuarracino:
If option 1 is already doable and faster (no changes and new tests in `smoothxg`), I would go in that direction ,and improve this aspect later. I would like to achieve a complete nf-core porting of `pggb` asap, in-sync with its github/docker flavors (bioconda is almost there too, but not at full 100% power yet).

heuermh:
Trying something new, I've moved the existing module to `SMOOTHXG_ONCE` and created a new one `SMOOTHXG_ITERATE` with a for loop in bash.

It works for me on the test data with `iterations = 2` but fails at `iterations = 3`.  Are there more arguments necessary to have iterations work with smoothxg?

Also still todo would be figuring out how to specify `iterations` and to flesh out the test assertions.  I think now the test files all clobber one another.

heuermh:
Ah sweet!  Nice work.  We can roll back my offending commits when it is ready.

subwaystation:
Alright, @AndreaGuarracino tested and I also did some small scale tests in https://github.com/subwaystation/pangenome/pull/26.
Now we need to pray for a new `smoothxg` release ;)

subwaystation:
https://github.com/pangenome/smoothxg/pull/174#issuecomment-1311608084 stay tuned!

heuermh:
Reverted the split commit and updated to the latest container image.

subwaystation:
@heuermh The new release is there! https://anaconda.org/bioconda/smoothxg.
Thanks @AndreaGuarracino

subwaystation:
I think this is good to go now. 
Pinging @heuermh @ggabernet @FriederikeHanssen to get one review. Thanks!

subwaystation (approved):
LGTM.

subwaystation:
Just discovered I can do this on my own. Hehehe.

heuermh:
Thank you, @subwaystation!

PR #1161, opened 2021-12-12
add: custom/dumprunparameters

jfy133:
## PR checklist

This adds an idea I've had rolling my head for a while. Sometimes when someone is presenting MultiQC results in a presentation or meeting they maybe asked about specifically what particular pipeline parameter was used (e.g. map quality filter).

This custom modules pulls all resolved `params` from the run, and formats them, and creates a MultiQC compatible TSV file that can be displayed in the run report.

Additional functionality:

1. Converts any parameter consisting of arrays/lists to just `...truncated list...` because I couldn't find a way to nicely parse these when they are nested
2. Allows exclusion of certain parameters that a developer feels is not relevant for displaying in a MultiQC report, e.g. iGenomes references or the contents of `test_data.config`.
3. Pure groovy + bash, no container needed!

We could also theoretically add this to the template as well (as with dump software versions), but for the time being I would suggest people optionally try out it out as I'm not sure what the variety of parameters different pipelines have.

image

- [x] Add a resource `label`


modules/custom/dumprunparams/main.nf
```diff
@@ -0,0 +1,45 @@
...
+    to_exclude = exclude instanceof List && exclude.size() > 0 ? exclude.join('|') : ''
+    exclude_cmd = exclude ? "| sed -E '/${to_exclude}/d'" : ''
+    println exclude
+    println exclude_cmd
+    """
+    echo "${run_params}" > run_params.txt
+    table_header=\$( sed 's/,/\t/g' <( echo "Parameter,col1" ) )
+
+    ## Remove outer brackets, remove nested arrays, split each param new line, delete lines containing params to ignore, convert params to cli flags
+    run_params=\$( sed 's#^\\[##g;s#\\]\$##g' run_params.txt | sed 's/\\[.*]/\\.\\.\\.truncated_list\\.\\.\\./g' | sed 's#, #\\n#g;s#:#\t#g' ${exclude_cmd} | sed 's#^#--#g' )
```
grst, line 24: :exploding_head: 

I guess this works, but this looks like a nightmare to debug if anything goes wrong. 
Since `params` is already a groovy map, couldn't you just dump it to yaml directly in groovy using an `exec` section? 

Take a look at: https://github.com/nf-core/modules/blob/edbbbbf42097c0696305dd2fd6d0e1070a7dbb07/modules/jupyternotebook/parametrize.nf#L22-L32

grst (commented):
I love the idea, I'm just wondering if there is a better way to get this into a multiqc-readable format than this sed magic.

modules/custom/dumprunparams/main.nf
jfy133: Definitely up for removing the whole `sed` mess! Will try it out!

jfy133: Ok, getting halfway there but it _will_ need some form of `sed` mess to format it in the way MQC wants it... but hopefully less :+1:

ewels: You have it as a structured variable in Groovy here, so my inclination would be to keep the entire process as pure groovy without a `script` block. Make a deep copy, restructure as needed and then write out a YAML / JSON file. MultiQC can handle these just as well as a tsv / csv.

ewels: eg. Flatten the nested params: https://stackoverflow.com/a/51288381 and stick in a new object under a `data` key. 

Then add the MultiQC config keys and you're away..

jfy133: I was leaning towards that so :+1: 

the only problem is the nested params will be extremely messy to display - for example in the tests (and I suspect also igenomes) the reference genomes wiht all the sub directories would be completely incomprhensiable. But maybe that's OK given I'm trying to have the `exclude` functionality.

ewels: Yeah exactly - we exclude those when rendering `--help` for the same reason. Might even exclude anything nested, not sure now 🤔

jfy133:
I played around this today, and I couldn't get my head around how to display nested lists, nor how to get a MultiQC compatible YAML file (given we don't have a `sample_a`, `sample_b` row-like system, as expected for most MQC plots).

So what I've reverted to is a simple custom MQC HTML section, that just prints all (but explicitly excluded) params in a YAML-like formatting code block.

It's simple but it works, and given it would be the person who _ran_ the pipeline who would be checking parameters, they likely would be semi-familiar with such a structure (e.g. it being similarish to pipeline `nextflow.config`s).

The only one outstanding issue is that all subsequent params after the first one is indented for some reason. This is derived from @grst 's parameterise.yml `indent_code_block` function, but I couldn't work out why this was happening.

Examples:

1. With nested parameters

    image


2. With nested parameter `test_data` excluded
 
     
image

grst:
I like new overview, as it can also display nested parameters! 
Will check what's going on with the indentation!

modules/custom/dumprunparams/main.nf
```diff
@@ -0,0 +1,32 @@
...
+    output:
+    path "params_mqc.html", emit: mqc_html
+    path "versions.yml"  , emit: versions
+
+    script:
+    run_params = params
+
+    to_exclude = run_params.subMap(exclude)
+    run_params_cleaned = exclude ? params.minus(to_exclude) : params
+    run_params_formatted = run_params_cleaned.each{ it.getValue() instanceof LinkedHashMap | it.getValue() instanceof List | it.getValue() instanceof nextflow.config.ConfigMap ? run_params[it.getKey()] = '<truncated_list>' : null }
```
grst, line 20: is this still needed as it is now possible to display nested parameters?

modules/custom/dumprunparams/main.nf
```diff
@@ -0,0 +1,32 @@
...
+
+    yml_cmd = dump_params_yml(run_params_formatted)
+
+    """
+    ${yml_cmd}
+    echo ""
```
grst, line 26: what's the purpose of the echo statement?

modules/custom/dumprunparams/main.nf
```diff
@@ -0,0 +1,32 @@
...
+    run_params = params
+
+    to_exclude = run_params.subMap(exclude)
+    run_params_cleaned = exclude ? params.minus(to_exclude) : params
+    run_params_formatted = run_params_cleaned.each{ it.getValue() instanceof LinkedHashMap | it.getValue() instanceof List | it.getValue() instanceof nextflow.config.ConfigMap ? run_params[it.getKey()] = '<truncated_list>' : null }
+
+    yml_cmd = dump_params_yml(run_params_formatted)
+
+    """
+    ${yml_cmd}
```
grst, line 25:
```suggestion
    ${indent_code_block(yml_cmd, 4)}
```

I hope this fixes the indentation problem.
For explanation: 
 * multiline strings, per default, include the indentation exactely as written. 
 * if it doesn't match the destination code block, the indentation will not be consistent
 * nextflow applies one final `stripIndent` to the multiline string in the script section.

modules/custom/dumprunparams/parametrize.nf
```diff
@@ -0,0 +1,49 @@
...
+    return """\
+    cat <<"END_PARAMS_SECTION" > ./params_mqc.html
+    <!--
+    id: 'pipeline-run-params'
+    section_name: 'Pipeline Run Parameters'
+    description: 'Resolved parameters for ${workflow.manifest.name} (v${workflow.manifest.version}) pipeline run: ${workflow.runName}. Note: All possible parameters are listed, but not necessarily used - some will only be utilised if a given module has been explicitly activated.'
+    -->
+    <pre><code>${indent_code_block(yaml_str, 8)}</code></pre>
+    END_PARAMS_SECTION
+    """
```
grst, line 48:
```suggestion
    <pre><code>
    ${indent_code_block(yaml_str, 4)}
    </code></pre>
    END_PARAMS_SECTION
    """
```

not sure if this change is needed, too, to fix the indentation.

modules/custom/dumprunparams/main.nf
jfy133: Good point, it didn't work anyway ;)

modules/custom/dumprunparams/main.nf
jfy133: Oops, that was a debugigng thing (I was struggling with HEREDOC)

jfy133:
Bingo! thanks @grst !

image

grst (approved):
:rocket:

modules/custom/dumprunparams/meta.yml
```diff
@@ -0,0 +1,26 @@
...
+
+input:
+  - exclude:
+      type: list
+      description: (Optional) groovy list of params to exclude from MultiQC report. Must be in format of list!
+
+output:
+  - mqc_csv:
+      type: file
+      description: MultiQC custom content YML file containing run parameters.
```
jfy133, line 22:
```suggestion
      description: MultiQC custom content HMTL file containing run parameters.
```

modules/custom/dumprunparams/meta.yml
```diff
@@ -0,0 +1,26 @@
...
+      documentation: https://github.com/nf-core/tools
+      licence: ['MIT']
+
+input:
+  - exclude:
+      type: list
+      description: (Optional) groovy list of params to exclude from MultiQC report. Must be in format of list!
+
+output:
+  - mqc_csv:
```
jfy133, line 20:
```suggestion
  - mqc_html:
```

jfy133:
Closing in preference of https://github.com/nf-core/tools/pull/1389

PR #1162, opened 2021-12-13
fix: exclude conda/singularity images from pytest workflow output

jfy133:
Currently adding entire `work` directory of CI tests to artifacts includes entire singularity images or ocnda environments which take a long time to upload and subsequently download when trying debug modules.

This excludes those specific folders to make the whole process faster, but also reduce our disk footprint


grst (approved):
Lgtm

PR #1163, opened 2021-12-13
Fix transcriptome staging issues on DNAnexus for rsem/prepareference

drpatelh:
Closes https://github.com/nf-core/rnaseq/issues/727

### DIFF
--- a/modules/rsem/preparereference/main.nf
+++ b/modules/rsem/preparereference/main.nf
@@ -12,9 +12,9 @@ process RSEM_PREPAREREFERENCE {
     path gtf
 
     output:
-    path "rsem"                , emit: index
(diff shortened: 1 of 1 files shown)


modules/rsem/preparereference/main.nf
```diff
@@ -40,6 +40,8 @@ process RSEM_PREPAREREFERENCE {
             $fasta \\
             rsem/genome
 
+        cp rsem/genome.transcripts.fa .
```
mahesh-panchal, line 43: Doesn't this collide with the input?

mahesh-panchal: Never mind. I apparently cannot tell the difference between the words `input` and `output`.

mahesh-panchal (approved):

PR #1166, opened 2021-12-13
Revert PR #1147

drpatelh:
Reverts https://github.com/nf-core/modules/pull/1147 to keep with `emit` a single output file per channel as per the nf-core/modules guidelines.


maxulysse:
Not sure I'm for this solution still

drpatelh:
Ok. But I think we should keep the "standard" solution until we have discussed things through.

maxulysse (approved):
OK with following the guidelines until the guidelines are changed accordingly

PR #1188, opened 2021-12-22
Uppdate `Lima` module: remove old suffix option

sguizard:
## PR checklist

This commit remove the old suffix option and check if input and output filenames are not the same.


FriederikeHanssen (approved):
lgtm

PR #1191, opened 2021-12-23
Fix modules warned or failed with nf-core modules lint

drpatelh:
Fixed some of these whilst developing https://github.com/nf-core/tools/pull/1374

* Now properly allow for `input:` and `output:` not being specified in module
* Allow for containers from other biocontainers resource
* Allow for `stageAs` syntax as defined here: https://github.com/nf-core/modules/blob/cde237e7cec07798e5754b72aeca44efe89fc6db/modules/cat/fastq/main.nf#L11
* Fix mismatching container versions and replace `%3A` notation with `:`
* Add missing modules to `pytest_modules.yml`
* Fix modules using `version` instead of `versions` in `emit` statement

Most of the remaining lint warnings are for modules that needs a version bump and those that generate empty files.


drpatelh:
No idea why `pirate` is failing I'm afraid. But nothing to do with the changes on this PR.

grst (approved):
LGTM

PR #1198, opened 2022-01-07
Bump software versions for viralrecon modules

drpatelh:
* Remove custom params.save_unaligned from bowtie2_align
* Unify samtools modules and error if input and output names are the same


erikrikarddaniel (approved):
LGTM.

drpatelh:
Just the conda tests failing for `picard` and `pangolin` now so should be good to go.

PR #1218, opened 2022-01-19
update manta to enable jointcalling

ramprasadn:
## PR checklist

Update manta to enable jointcalling for germline variants

Closes #XXX 

- [ ] Add a resource `label`


modules/manta/germline/main.nf
```diff
@@ -8,11 +8,12 @@ process MANTA_GERMLINE {
...
     input:
-    tuple val(meta), path(input), path(input_index)
+    tuple val(meta), path(input)
+    path index
     path fasta
-    path fai
-    path target_bed
-    path target_bed_tbi
+    path fasta_fai
+    tuple val(bed_meta), path(target_bed), path(target_bed_tbi)
```
FriederikeHanssen, line 15: Is it possible to run multiple regions per patient? In that case target_bed should go with the input file. What is the content of the bed_meta, I haven't seen this yet in the other modules?

modules/manta/germline/main.nf
```diff
@@ -8,11 +8,12 @@ process MANTA_GERMLINE {
         'quay.io/biocontainers/manta:1.6.0--h9ee0642_1' }"
 
     input:
-    tuple val(meta), path(input), path(input_index)
+    tuple val(meta), path(input)
```
FriederikeHanssen, line 11: I wuld propose to leave the index with the input. We have been going back and forth on this multiple times. This just avoids needless merging and splitting of channels

modules/manta/germline/main.nf
ramprasadn: target_bed is not defined for each patient rather it is defined for all the patients in a case. 

My rationale behind having a bed_meta is that the target_bed file need not have the same information in its meta as the alignment files. For instance, the id here could be something other than the case/sample id. So defining it as bed_meta avoids those conflicts.

ramprasadn: On second thought, bed_meta isn't really used in the module so I could just get rid of it, and the user can pass a tuple containing the path to target_bed and target_bed_tbi. What do you think?

modules/manta/germline/main.nf
ramprasadn: Sure! I will fix that 👍🏻

modules/manta/germline/main.nf
FriederikeHanssen: Yes I think this is sensible.

FriederikeHanssen: Do you know if it is possible to parallelize manta about regions. I.e. for each patient run manta in aprallel per lets say chromosome?

modules/manta/germline/main.nf
```diff
@@ -29,7 +30,7 @@ process MANTA_GERMLINE {
     def options_manta = target_bed ? "--exome --callRegions $target_bed" : ""
```
FriederikeHanssen, line 30: IMO the `--exome` should be removed herre and put into the `modules.config` in the pipeline itself. These two variable can be used independently

modules/manta/germline/main.nf
```diff
@@ -29,7 +30,7 @@ process MANTA_GERMLINE {
...
     """
     configManta.py \
-        --bam $input \
+        --bam ${input.join(' --bam ')} \
         --reference $fasta \
         $options_manta \
```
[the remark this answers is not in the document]
ramprasadn: Good catch! Sure!

modules/manta/germline/main.nf
ramprasadn: As far as I know, the option to do that is not baked into manta so I think that should be handled upstream.

FriederikeHanssen: hm yes, ok. Just to be sure, my thought was something similar to what Haplotypecaller does, were you can ran all regions parallely and join them afterwards. (The module itself wouldn't do any parallelization):

chr1 -> run_manta(patient1,chr1) -> results_pat1_chr1
chr1 -> run_manta(patient2,chr1)-> results_pat2_chr1
chr2 -> run_manat(patient1,chr2)-> results_pat1_chr2
chr2 -> run_manat(patient2,chr2)-> results_pat2_chr2

join(results_pat1_chr1, results_pat1_chr2)
join(results_pat2_chr1, results_pat2_chr2)

ramprasadn: ah, I haven't tried it, but I don't see a reason for why it would not work because the information doesn't traverse chromosomes.

tests/modules/manta/germline/main.nf
```diff
@@ -7,29 +7,46 @@ include { MANTA_GERMLINE } from '../../../../modules/manta/germline/main.nf'
 workflow test_manta_germline {
     input = [
         [ id:'test'], // meta map
-        file(params.test_data['homo_sapiens']['illumina']['test_paired_end_recalibrated_sorted_cram'], checkIfExists: true),
-        file(params.test_data['homo_sapiens']['illumina']['test_paired_end_recalibrated_sorted_cram_crai'], checkIfExists: true)
+        [ file(params.test_data['homo_sapiens']['illumina']['test_paired_end_sorted_cram'], checkIfExists: true)],
```
FriederikeHanssen, line 10: what is the reason that you switched the test files?

ramprasadn: Ah, because there was mismatch between the fasta header and the reference sequence in the cram. The genome fasta sequence is from chr22, and cram alignments are from chr21. Hence, the switch.

modules/manta/germline/main.nf
```diff
@@ -26,13 +26,14 @@ process MANTA_GERMLINE {
     script:
     def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
-    def options_manta = target_bed ? "--exome --callRegions $target_bed" : ""
+    def options_manta = target_bed ? "--callRegions $target_bed" : ""
     """
     configManta.py \
-        --bam $input \
+        --bam ${input.join(' --bam ')} \
```
maxulysse, line 32: don't we get 2 `--bam` in a row if we keep the first `--bam` here?

modules/manta/germline/main.nf
```diff
@@ -26,13 +26,14 @@ process MANTA_GERMLINE {
     script:
     def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
-    def options_manta = target_bed ? "--exome --callRegions $target_bed" : ""
+    def options_manta = target_bed ? "--callRegions $target_bed" : ""
```
maxulysse, line 29: Why removing the --exome?
Or this is now something that should be done with the `ext.args`?

modules/manta/germline/main.nf
```diff
@@ -54,4 +55,20 @@ process MANTA_GERMLINE {
...
+    touch ${prefix}.candidate_sv.vcf.gz
+    touch ${prefix}.candidate_sv.vcf.gz.tbi
+    touch ${prefix}.diploid_sv.vcf.gz
+    touch ${prefix}.diploid_sv.vcf.gz.tbi
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        manta: \$( configManta.py --version )
+    END_VERSIONS
+    """
```
maxulysse, line 73: <3

modules/manta/germline/main.nf
FriederikeHanssen: This allows us to use callRegions but for example using interval files for WGS. --exome then via ext.args if the data is actual some type of panel

modules/manta/germline/main.nf
ramprasadn: We get additional `--bam` parameters in the command only when we have two or more bam files in the list. You can see that first of three tests for this module only uses a single cram file and it works :)

maxulysse (approved):

PR #1220, opened 2022-01-20
CI: print out pytest logs on failure

grst:
## PR checklist

If nextflow fails within `pytest` it is relatively hard to figure out what's going on. At the moment it is necessary to download the artifacts and retrieve the log files from there. This PR prints out the logfiles to the github actions log in case of failure, which I find more convenient. 

If something goes wrong within a work directory, it will still be necessary to obtain the artifact. 

---
- [ ] Add a resource `label`

### DIFF
--- a/.github/workflows/pytest-workflow.yml
+++ b/.github/workflows/pytest-workflow.yml
@@ -93,6 +93,16 @@ jobs:
         # only use one thread for pytest-workflow to avoid race condition on conda cache.
         run: TMPDIR=~ PROFILE=${{ matrix.profile }} pytest --tag ${{ matrix.tags }} --symlink --kwdof
 
+      - name: Output log on failure
+        if: failure()
+        run: |
+          echo "======> log.out <======="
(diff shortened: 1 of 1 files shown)


Emiller88 (approved):
Looks great to me! Maybe we could use `bat` though for some prettier printing?

grst:
hmm, I guess we need to add another step to install that first? 

btw, as I just happened to notice: Is there a reason we are running the "Setup python" action twice, once before and once after the cache got restored?

Emiller88:
> hmm, I guess we need to add another step to install that first?
Yeah, this can definately get merged and we can address it later. Or it looks like it's in the ubuntu repo so it should be a `sudo apt install bat` away.
> btw, as I just happened to notice: Is there a reason we are running the "Setup python" action twice, once before and once after the cache got restored?
https://github.com/nf-core/modules/pull/1225 I'll save you the `git blame`, it was me. :upside_down_face:

PR #1238, opened 2022-01-27
Update mergebam syntax and sample specific unmapped channel

### LINKED ISSUE #1237: new syntax for mergeBamAlignment + move unmapped channel to sample specific [UPDATE]
### Update mergebamalignment syntax:
While bugfixing for a subworkflow, I got the following warning for gatk4/mergeBamAlignment:
"NOTE: Picard's command line syntax is changing.

For more information, please see:
https://github.com/broadinstitute/picard/wiki/Command-Line-Syntax-Transition-For-Users-(Pre-Transition)

The command line looks like this in the new syntax:

MergeBamAlignment -ALIGNED test.bam -UNMAPPED test.paired_end.umi_converted.bam -R genome.fasta -O test.bam "

The old syntax still works, but so does the new syntax, as such I think it would be a good idea to update to the new syntax now before the old syntax is retired.

### Make unmapped input sample specific:
MergeBamAlignment is meant to add information about the reads, which is lost when the ubam is converted to fastq, back into the aligned bam file. So I think it makes sense to move the unmapped input channel to the sample-specific tuple, as this would make it easier to pass the unmapped and mapped bam files into the tool as a matched pair.

I have put together a PR updating the syntax and moving the unmapped channel to the sample-specific tuple.

GCJMackenzie:
## PR checklist

Closes #1237 

This PR updates the gatk4 MergeBamAlignment module to use the new syntax being introduced for the tool. It also moves the unmapped input channel to the sample-specific tuple, as I think it makes it easier to pass in the mapped and unmapped bam files as a matched pair.


tests/modules/gatk4/mergebamalignment/main.nf
```diff
@@ -6,11 +6,11 @@ include { GATK4_MERGEBAMALIGNMENT } from '../../../../modules/gatk4/mergebamalig
 
 workflow test_gatk4_mergebamalignment {
     input    = [ [ id:'test' ], // meta map
-                 file(params.test_data['sarscov2']['illumina']['test_single_end_bam'], checkIfExists: true)
+                 file(params.test_data['sarscov2']['illumina']['test_single_end_bam'], checkIfExists: true),
```
FriederikeHanssen, line 9: As per usual my question is: does this work with cram files too?

GCJMackenzie: From what I have seen in both the GATK and picard tool documentations, I don't think it works with cram files as neither of them mention cram. Is there an example of an unaligned bam (from what I understand you don't get unaligned crams)  and an aligned cram we could use to double check on the test datasets repo?

FriederikeHanssen: I have seen some discussions around https://www.biostars.org/p/389788/ https://davetang.org/muse/2022/01/20/storing-fastq-as-unaligned-cram/ but not seen it out in the wild. For me it is fine to leave it as is

FriederikeHanssen (approved):

PR #1242, opened 2022-01-28
Fix for version file output check when performing an expected error test

chris-cheshire:
The `exit_code` test feature of pytest that allows for checking expected errors cannot currently be used. This is because the versions file auto-test does not distinguish between workflows that have no expected output.

I have put in a small check to test if the workflow has an error code setting in the yaml and will ignore this workflow if this is the case.


grst (approved):
LGTM! Thanks @chris-cheshire!

grst:
One small thing: can you maybe add a comment in the code, why this is necessary? I feel I wouldn't understand in two months from now why the test for `exit_code` is there...

chris-cheshire:
Yeah thats a good idea, done

PR #1252, opened 2022-02-01
Add repository_type to .nf-core.yml

ewels:
Add config `repository_type`, to be used by nf-core/tools after https://github.com/nf-core/tools/pull/1391 is merged.


grst (approved):
LGTM!

PR #1264, opened 2022-02-08
fix incorrect yaml formatting.

mashehu:


charles-plessy (approved):
This PR indeed fixes incorrect yaml formatting.

PR #1279, opened 2022-02-09
Fix formatting in yaml files, add yamllint config

mashehu:
for some reason the yamllint seems to only have started throwing errors recently. I added a config to be less picky about line lengths and document headers.
We have quite a few incorrect yaml files now (~157). Mainly incorrect indentation (which can mess up the rendering on the modules page). Would be great if other's could help with this effort.


.github/workflows/code-linting.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - uses: actions/setup-node@v1
+      - uses: actions/setup-node@v2
```
ewels, line 15: You still need to remove the `with: node-version:` lines to automatically use a recent version of npm.

Once you do that the markdownlint should kick in again.

drpatelh (approved):
🚀 Thanks alot @mashehu !

PR #1281, opened 2022-02-10
Bump node-version to 16

JoseEspinosa:
Fixes markdown linting, which now requires at least node version 12.
As done in `tools` [here](https://github.com/nf-core/tools/pull/1398) version 16 is bumped.


JoseEspinosa:
Just realized that is already addressed in #1279, so I close it

PR #1308, opened 2022-02-16
include intervals_index

nickhsmith:
In order to run the GATK `GenotypeGVCFs` tool with an interval file (such as a .bed file) you need to have the corresponding index file. Otherwise you get the following error message
```
A USER ERROR has occurred: An index is required but was not found for file genome.bed.gz: work/6c/541fce2dc670597f62a571c72288c6/genome.bed.gz. Support for unindexed block-compressed files has been temporarily disabled. Try running IndexFeatureFile on the input.
```

including the index file fixes this problem


maxulysse (approved):

maxulysse:
Could you add the information regarding the index to the yaml file?

Issue #1344, opened 2022-02-23
[BUG] inconsistant input structure for manta germline/somatic/tumor_only


nickhsmith:
The input for somatic and tumor only manta are 5 channels, but for germline it's 4. where the last 2 channels are combined. This seems silly. I will work on fixing this if someone else doesn't
```
    input:
    tuple val(meta), path(input_normal), path(input_index_normal), path(input_tumor), path(input_index_tumor)
    path fasta
    path fai
    path target_bed
    path target_bed_tbi
```
vs 

```
    input:
    tuple val(meta), path(input), path(index)
    path fasta
    path fasta_fai
    tuple path(target_bed), path(target_bed_tbi)
```

FriederikeHanssen:
ok we agreed with @ramprasadn and @maxulysse on a structure for these modules, they should be consistent now. Please re-open if you are still unhappy with the state of the modules

PR #1349, opened 2022-02-24
Nickhsmith patch 1

nickhsmith:
Match the input stylings of somatic and tumor_only to the germline as mentioned in #1344 
   
    input:
    tuple val(meta), path(input), path(index)
    path fasta
    path fasta_fai
    tuple path(target_bed), path(target_bed_tbi)


maxulysse:
Can you update tests as well?

PR #1356, opened 2022-02-27
Fix: Leehom meta output channel descriptions

jfy133:
Copy and paste error from R1 into R2 in `meta.yml`


FriederikeHanssen (approved):
lgtm

PR #1360, opened 2022-02-28
Add stranger

### LINKED ISSUE #1248: new module: stranger
I think it would be good to have a module for [stranger](https://anaconda.org/bioconda/stranger).

ljmesi:
## PR checklist

Adding stranger which annotatates output vcf files from ExpansionHunter with pathologic implications of the repeat sizes. 

Closes #1248 

- [X] Add a resource `label`

### DIFF
--- a/modules/stranger/main.nf
+++ b/modules/stranger/main.nf
@@ -14,6 +14,9 @@ process STRANGER {
     tuple val(meta), path("*.gz"), emit: vcf
     path "versions.yml"          , emit: versions
 
+    when:
+    task.ext.when == null || task.ext.when
+
     script:
     def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
--- a/modules/stranger/main.nf
+++ b/modules/stranger/main.nf
@@ -1,0 +1,30 @@
+process STRANGER {
+    tag "$meta.id"
+    label 'process_low'
+
+    conda (params.enable_conda ? "bioconda::stranger=0.8.1" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/stranger:0.8.1--pyh5e36f6f_0':
+        'quay.io/biocontainers/stranger:0.8.1--pyh5e36f6f_0' }"
+
+    input:
+    tuple val(meta), path(vcf)
+
+    output:
+    tuple val(meta), path("*.gz"), emit: vcf
+    path "versions.yml"          , emit: versions
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    stranger \\
+        $args \\
+        $vcf | gzip --no-name > ${prefix}.vcf.gz
+
+    cat <<-END_VERSIONS > versions.yml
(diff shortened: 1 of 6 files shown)


modules/stranger/main.nf
```diff
@@ -0,0 +1,33 @@
...
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    stranger \\
+        $args \\
+        $vcf | gzip --no-name > ${prefix}.vcf.gz
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        stranger: \$(stranger --version 2>&1)
```
mahesh-panchal, line 30:
```suggestion
        stranger: \$( stranger --version )
```

mahesh-panchal (approved):
LGTM

modules/stranger/main.nf
```diff
@@ -0,0 +1,33 @@
...
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    stranger \\
+        $args \\
+        $vcf | gzip --no-name > ${prefix}.vcf.gz
```
jemten, line 26: I would try to use bgzip for this step. I don't think gzip will produce a blocked gzip file so you won't be able to index this for quick access. Might be that bgzip isn't part of the container

jemten (commented):
Nice work @ljmesi 
Jus a question about using gzip instead of bgzip

modules/stranger/main.nf
mahesh-panchal: `bzip2` is part of the container, but I think most modules assume `gzip`ped input. This may be something to bring up generally in the #modules channel since I don't think we have guidelines for this.

mahesh-panchal: I didn't know that. We should definitely add this to guidelines in some way: https://www.uppmax.uu.se/support/faq/resources-faq/which-compression-format-should-i-use-for-ngs-related-files/

ljmesi: Thank you for your feedback! Should I switch the command in line 26 to `bzip2` before merging? Is the file name suffix then `vcf.bz2`?

mahesh-panchal: I would say yes, do it. The consideration is time vs usage benefits tradeoff. `bzip2` is a slower algorithm, which means it might cost more to run in a cloud environment, but perhaps this is negligible ( I'm going to assume it is until someone complains ).
And yes to the suffix.
Feel free to request a re-review.

jemten: I have usually just used `bgzip` from the htslib package (http://www.htslib.org/doc/bgzip.html) and then used the `.vcf.gz` suffix. Not that experienced with bzip2 (same thing?).

mahesh-panchal: Oh, it's a different tool altogether. `bgzip` is not in the container and would require making a mulled container. I think in that case, stick with `gzip`. Having to update a container regularly while simple, is unnecessary extra work.

mahesh-panchal: It seems I completely misread the name.

ljmesi: Ok. I will merge the pull request then.

PR #1377, opened 2022-03-09
Add centrifuge module

sofstam:
Add centrifuge module 

## PR checklist

Closes [#833 ](https://github.com/nf-core/modules/issues/833)

- [x] Add a resource `label`


modules/centrifuge/centrifuge/main.nf
```diff
@@ -0,0 +1,39 @@
+process CENTRIFUGE_CENTRIFUGE {
```
sateeshperi, line 1: are there more than one centrifuge tools that you have `CENTRIFUGE_CENTRIFUGE` as the process name rather than `CENTRIFUGE` ?

sofstam: It is only one tool, CENTRIFUGE. I followed the same format as [kraken2](https://github.com/nf-core/modules/tree/master/modules/kraken2/kraken2) module.

sofstam: I used the following command: `nf-core modules create centrifuge/centrifuge` and the process was named that way by default, I did not change it manually.

sateeshperi: ah yours is a single tools so you just need `nf-core modules create centrifuge`

sateeshperi: `nf-core modules create <tool>/<sub-tool>` is when a tool has multiple subtools

sateeshperi: indeed! that needs fixing....will work on it....thanks for bringing this up

sofstam: The PR is now updated.

modules/centrifuge/main.nf
```diff
@@ -0,0 +1,38 @@
...
+    def paired = meta.single_end ? "-U ${reads}" :  "-1 ${reads[0]} -2 ${reads[1]}"
+    """
+    tar -xf $db
+    centrifuge \\
+        -x ${db.toString().replace(".tar.gz","")} \\
+        -p $task.cpus \\
+        $paired \\
+        --report-file ${prefix}.centrifuge.report.txt \\
+        $args
+    cat <<-END_VERSIONS > versions.yml
```
jfy133, line 33:
```suggestion

    cat <<-END_VERSIONS > versions.yml
```

modules/centrifuge/main.nf
```diff
@@ -0,0 +1,38 @@
...
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def paired = meta.single_end ? "-U ${reads}" :  "-1 ${reads[0]} -2 ${reads[1]}"
+    """
+    tar -xf $db
+    centrifuge \\
+        -x ${db.toString().replace(".tar.gz","")} \\
+        -p $task.cpus \\
+        $paired \\
+        --report-file ${prefix}.centrifuge.report.txt \\
```
jfy133, line 31: I think you're missing the `-S` output from somewhere here?

From the docs:

```
  centrifuge [options]* -x <cf-idx> {-1 <m1> -2 <m2> | -U <r> | --sample-sheet <s> } [-S <filename>] [--report-file <report>]
  <filename>      File for classification output (default: stdout)
```

modules/centrifuge/main.nf
```diff
@@ -0,0 +1,38 @@
...
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/centrifuge:1.0.4_beta--h9a82719_6' :
+        'quay.io/biocontainers/centrifuge:1.0.4_beta--h9a82719_6' }"
+
+    input:
+    tuple val(meta), path(reads)
+    path db
+
+    output:
+    tuple val(meta), path('*report.txt')  , emit: report
```
jfy133, line 15: This tool can also output SAM files, you will need to account for this here, e.g. with a new line

```
tuple val(meta), path('*.sam') , optional: true emit: sam
```
Based on
```
  --out-fmt <str>       define output format, either 'tab' or 'sam' (tab)
```

jfy133 (changes requested):
Missing a few output options

Also there are the following outputs that might be important to consider that a user may optionally want, so you need to make sure that this accounted for:

```
  --un <path>           write unpaired reads that didn't align to <path>
  --al <path>           write unpaired reads that aligned at least once to <path>
  --un-conc <path>      write pairs that didn't align concordantly to <path>
  --al-conc <path>      write pairs that aligned concordantly at least once to <path>
  (Note: for --un, --al, --un-conc, or --al-conc, add '-gz' to the option name, e.g.
  --un-gz <path>, to gzip compress output, or add '-bz2' to bzip2 compress output.)
```

modules/centrifuge/meta.yml
```diff
@@ -0,0 +1,46 @@
...
+keywords:
+  - classify
+  - metagenomics
+  - fastq
+  - db
+tools:
+  - centrifuge:
+      description: Centrifuge is a classifier for metagenomic sequences.
+      homepage: https://ccb.jhu.edu/software/centrifuge/
+      documentation: https://github.com/DaehwanKimLab/centrifuge
```
jfy133, line 12:
```suggestion
      documentation: https://ccb.jhu.edu/software/centrifuge/manual.shtml
```

modules/centrifuge/meta.yml
```diff
@@ -0,0 +1,46 @@
...
+        Groovy Map containing sample information
+        e.g. [ id:'test', single_end:false ]
+  - reads:
+      type: file
+      description: |
+        List of input FastQ files of size 1 and 2 for single-end and paired-end data,
+        respectively.
+  - db:
+      type: directory
+      description: Centrifuge database
```
jfy133, line 28: Given you strip off a bunch of hard-coded suffixes, and requires `tar.gz` files (it seems) this should at least be documented here.

sofstam: Thank you for the suggestions, I am not quite sure what you would expect here, something like:

```
- db:
      type: directory
      description: Centrifuge database in tar.gz format
```

or:

```
- db:
      type: directory
      description: Centrifuge database
      pattern: "*.{tar.gz}"
```

jfy133: Both would be good :D

modules/centrifuge/main.nf
sofstam: I was thinking if those should be optional? Taking into account how centrifuge is implemented in mag.
 ```
 --un <path>           write unpaired reads that didn't align to <path>
 --al <path>           write unpaired reads that aligned at least once to <path>
 --un-conc <path>      write pairs that didn't align concordantly to <path>
 --al-conc <path>      write pairs that aligned concordantly at least once to <path>
 (Note: for --un, --al, --un-conc, or --al-conc, add '-gz' to the option name, e.g.
 --un-gz <path>, to gzip compress output, or add '-bz2' to bzip2 compress output.)
```

**Comment by sofstam** [comment]
I updated the PR.

File: modules/centrifuge/main.nf [hunk 7]
```diff
@@ -0,0 +1,57 @@
...
+    path db
+    val save_unaligned
+    val sam_format
+
+    output:
+    tuple val(meta), path('*report.txt')  , emit: report
+    tuple val(meta), path('*results.txt') , emit: results
+    tuple val(meta), path('*kreport.txt') , emit: kreport
+    tuple val(meta), path('*.sam')        , optional: true, emit: sam
+    tuple val(meta), path('*.fastq.gz')   , optional:true, emit: fastq
```
jfy133, line 21:
```suggestion
    tuple val(meta), path('*.fastq.gz')   , optional: true, emit: fastq
```

modules/centrifuge/main.nf
jfy133: What about `-al`? Also I see there is the `--met-file` option... I'm not entirely sure what that means here but it might be useful if it containts logging information?

modules/centrifuge/meta.yml
```diff
@@ -0,0 +1,57 @@
...
```
jfy133, line 28:
```suggestion
      description: Centrifuge database in .tar.gz format
```

modules/centrifuge/main.nf
```diff
@@ -0,0 +1,57 @@
...
+    }
+    def sam_output = sam_format ? "--out-fmt 'sam'" : ''
+    """
+    tar -xf $db
+    centrifuge \\
+        -x $db_name \\
+        -p $task.cpus \\
+        $paired \\
+        --report-file ${prefix}.centrifuge.report.txt \\
+        -S ${prefix}.centrifuge.results.txt \\
```
jfy133, line 46: You shouldn't specify anything the output file name other than `${prefix}.<extension>`, as this depends on the pipeline user.

In this case as you have multiple `txt` files, you could instead go `${prefix}.report` and `${prefix}.results` etc. (make sure to update the output block).

Last point here: https://nf-co.re/developers/adding_modules#naming-conventions

jfy133 (changes requested):
Getting there! Good implementation of `--un-gz`

modules/centrifuge/main.nf
sofstam: Thanks for the further clarification.

modules/centrifuge/main.nf
sofstam: I attach an example of how `--met-file` option looks like:
[met.txt](https://github.com/nf-core/modules/files/8243538/met.txt)

jfy133: Ok `--met-file` looks like developer debugging things, I think we can skip that :)

sofstam: Regarding the `-al` and `-un`, would that name convention work? Or would you prefer skipping `unmapped`, `mapped` in the filename?

```
def unaligned = ''
def aligned = ''
if (meta.single_end) {
     unaligned = save_unaligned ? "--un-gz ${prefix}.unmapped.fastq.gz" : ''
     aligned = save_aligned ? "--al-gz ${prefix}.mapped.fastq.gz" : ''
} else {
      unaligned = save_unaligned ? "--un-conc-gz ${prefix}.unmapped.fastq.gz" : ''
      aligned = save_aligned ? "--al-conc-gz ${prefix}.mapped.fastq.gz" : ''
}
```

jfy133: In this case we would indeed have to specify. I think it's is fine in this csae :+1:

sofstam:
The PR is now updated based on your comments :)

modules/centrifuge/meta.yml
```diff
@@ -0,0 +1,58 @@
...
+        Groovy Map containing sample information
+        e.g. [ id:'test', single_end:false ]
+  - reads:
+      type: file
+      description: |
+        List of input FastQ files of size 1 and 2 for single-end and paired-end data,
+        respectively.
+  - db:
+      type: directory
+      description: Centrifuge database in .tar.gz format
```
jfy133, line 28: Missing the new input channels (the values for aligned/unaligned etc.)

modules/centrifuge/main.nf
```diff
@@ -0,0 +1,62 @@
...
+    val save_unaligned
+    val save_aligned
+    val sam_format
+
+    output:
+    tuple val(meta), path('*report.txt')  , emit: report
+    tuple val(meta), path('*results.txt') , emit: results
+    tuple val(meta), path('*kreport.txt') , emit: kreport
+    tuple val(meta), path('*.sam')        , optional: true, emit: sam
+    tuple val(meta), path('*.fastq.gz')   , optional: true, emit: fastq
```
jfy133, line 22: I would rather export these as two separate channels - one for unmapped, one for mapped. I've had problems in other modules where it's been complicated to separate out them when I only wanted one.

modules/centrifuge/meta.yml
jfy133, line 28:
```suggestion
      description: Centrifuge database in .tar.gz format
      pattern: "*.tar.gz"
```

jfy133 (approved):
Almost there! But otherwise LGTM once last couple of issues addressed

modules/centrifuge/main.nf
sofstam: Sounds good! Would the following work?

```
tuple val(meta), path('*mapped.fastq.gz')       , optional: true, emit: fastq
tuple val(meta), path('*unmapped.fastq.gz')   , optional: true, emit: fastq
```

jfy133: No, would need separate emits: `emit: fastq_mapped` and `mit: fastq_unmapped` for example

sofstam: I updated the PR with the separate input channels and emits.

jfy133:
@sofstam just need to update the meta.yml and good to go I think

sofstam:
I updated the `meta.yml`. I can merge it if everything is ok.

jfy133:
@sofstam sorry I forgot one more thing :sweat_smile: could you add one or two more tests for when you include the `true/false` in the save_aligned stuff. Just to make sure the logic is working :)

sofstam:
@jfy133 That is a good point 😄 The tests have passed for both single-end and paired-end reads but the `test.yml` has only been updated for the `single-end`. I guess this means that something is not fully correct

sateeshperi:
hi @sofstam looks like this module is almost ready for merge, however I am seeing that this branch is out-of-date with the base branch and I guess the force-pushes might have broken something. Would you mind submitting a new PR just to be sure we don't break anything else going forward?

sofstam:
Thank you @sateeshperi for pointing this out.

PR #1393, opened 2022-03-15
fix controlfreek's meta.yml: fix indent error, avoiding html tag confusion

mashehu:
Should be soon™ rendered as:
Screenshot 2022-03-15 at 11 39 15


FriederikeHanssen (approved):
thank you for fixing this. wonder why it passed linting

PR #1395, opened 2022-03-16
round the < to ( to make markdown work for meta.yml

lassefolkersen:
## PR checklist

Closes #XXX 
This is a minor change to make sure the markdown in the `meta.yml` renders correctly. Thanks for noticing @mashehu

### DIFF
--- a/modules/ascat/meta.yml
+++ b/modules/ascat/meta.yml
@@ -22,16 +22,16 @@ input:
           [
             "gender": "XX",
             "genomeVersion": "hg19"
-            "purity": <optional>,
-            "ploidy": <optional>,
-            "gc_files": <optional>,
-            "minCounts": <optional>,
-            "chrom_names": <optional>,
-            "min_base_qual": <optional>,
-            "min_map_qual": <optional>,
-            "ref_fasta": <optional>,
-            "skip_allele_counting_tumour": <optional>,
-            "skip_allele_counting_normal": <optional>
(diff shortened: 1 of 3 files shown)


modules/ascat/meta.yml
```diff
@@ -22,16 +22,16 @@ input:
...
-            "gc_files": <optional>,
-            "minCounts": <optional>,
-            "chrom_names": <optional>,
-            "min_base_qual": <optional>,
-            "min_map_qual": <optional>,
-            "ref_fasta": <optional>,
-            "skip_allele_counting_tumour": <optional>,
-            "skip_allele_counting_normal": <optional>
+            "purity": (optional),
+            "ploidy": (optional>,
```
mashehu, line 26:
```suggestion
            "ploidy": (optional),
```

mashehu (commented):
There is also an <optional> in line 18.

modules/ascat/meta.yml
lassefolkersen: Ok, thanks I've corrected

mashehu (approved):

lassefolkersen:
Ok, here's a new commit with _all_ the < converted to (, thanks @mashehu - also had to do some other changes because the linting started failing for empty files (it didn't before); which in turn needed an update in the stub-running documentation (being [pushed](https://github.com/nf-core/nf-co.re/pull/1079) right now).

PR #1397, opened 2022-03-16
Fix DeepARG depenency issue with singularity containers

jfy133:
There is a 'root' read-only file in an ancient conda depency of deeparg (as in the recipe for the dependency is from 2015 or something), which isn't compatible with singulatiy. This adds a containerOptions to allow singularity to run 'as root' for this specific tool. This currently won't work on AWS Batch, but this normally uses docker anyway for our tests, and will be supported in the next NXF release.

## PR checklist

- [ ] Add a resource `label`


modules/deeparg/downloaddata/main.nf
```diff
@@ -7,6 +7,7 @@ process DEEPARG_DOWNLOADDATA {
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
         'https://depot.galaxyproject.org/singularity/deeparg:1.0.2--pyhdfd78af_1' :
         'quay.io/biocontainers/deeparg:1.0.2--pyhdfd78af_1' }"
+    containerOptions { "${workflow.containerEngine}" == 'singularity' ? '--fakeroot' : '' }
```
Midnighter, line 10: Using a similar pattern to how the container image is set, I think it should be

```suggestion
    containerOptions "${workflow.containerEngine == 'singularity' ? '--fakeroot' : ''}"
```

modules/deeparg/predict/main.nf
```diff
@@ -8,6 +8,7 @@ process DEEPARG_PREDICT {
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
         'https://depot.galaxyproject.org/singularity//deeparg:1.0.2--pyhdfd78af_1' :
         'quay.io/biocontainers/deeparg:1.0.2--pyhdfd78af_1' }"
+    containerOptions { "${workflow.containerEngine}" == 'singularity' ? '--fakeroot' : '' }
```
Midnighter, line 11: As above

```suggestion
    containerOptions "${workflow.containerEngine == 'singularity' ? '--fakeroot' : ''}"
```

Midnighter (commented):
I think somebody else should comment on whether setting `--fakeroot` is okay to use.

Emiller88:
I think that's really clever! It doesn't affect any other modules, and the end user won't even notice it. 

Maybe you could print a warning in it just in case it's a security concern?

jfy133:
One point raised by @grst on the funscan issue is that this flag might not be available on all systems... so maybe we should put a `WARN` why that may be however on the flipside, it might be too late at this point about the security perspective? 

So I think it might be better to leave these to pipeline devs? I'll add a comment in the module code at least.

modules/deeparg/downloaddata/main.nf
```diff
@@ -7,6 +7,7 @@ process DEEPARG_DOWNLOADDATA {
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
         'https://depot.galaxyproject.org/singularity/deeparg:1.0.2--pyhdfd78af_1' :
         'quay.io/biocontainers/deeparg:1.0.2--pyhdfd78af_1' }"
+    containerOptions "${workflow.containerEngine == 'singularity' ? '--fakeroot' : ''}"
```
jfy133, line 10:
```suggestion
    /*
    We have to force singularity to run with --fakeroot to allow reading of a problematic file with borked read-write permissions in an upstream dependency (theanos). 
    This flag may not be available on all systems and may be considered a security problem. so please document and /or warn for this in your pipeline! 
    */
    containerOptions "${workflow.containerEngine == 'singularity' ? '--fakeroot' : ''}"
```

modules/deeparg/predict/main.nf
```diff
@@ -8,6 +8,7 @@ process DEEPARG_PREDICT {
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
         'https://depot.galaxyproject.org/singularity//deeparg:1.0.2--pyhdfd78af_1' :
         'quay.io/biocontainers/deeparg:1.0.2--pyhdfd78af_1' }"
+    containerOptions "${workflow.containerEngine == 'singularity' ? '--fakeroot' : ''}"
```
jfy133, line 11:
```suggestion
    /*
    We have to force singularity to run with --fakeroot to allow reading of a problematic file with borked read-write permissions in an upstream dependency (theanos). 
    This flag may not be available on all systems and may be considered a security problem. so please document and /or warn for this in your pipeline! 
    */
    containerOptions "${workflow.containerEngine == 'singularity' ? '--fakeroot' : ''}"
```

grst (approved):
Don't see any better way

PR #1404, opened 2022-03-17
[COOLER] run through multiple resolutions

nservant:
Hi

A small changes to allow to run the `cload` and `dump` cooler functions across multiple resolutions
Thanks
Nicolas


nservant:
Note that I also proposed to change the channel to have always the `resolution` at the end to avoid the usage of `map` to change the order

nservant:
Sorry @jianhong, I missed your message ! I'll check the test files !

nservant:
@jianhong, I just updated the tests. However, I'm not very familiar with modules and did not yet find a way to test my changes locally ... do you know how to do it ?

jianhong:
> @jianhong, I just updated the tests. However, I'm not very familiar with modules and did not yet find a way to test my changes locally ... do you know how to do it ?

I will do that. But I need some time to finish it.

Jianhong.

muffato:
Hi @nservant 

Can you explain what you mean by "allow to run the cload and dump cooler functions across multiple resolutions" ? I don't see how that's not possible with the existing version.

muffato:
I took a look at the tests, and there is an issue in `cooler/dump`. The module appends the resolution to the name of the input file, cf https://github.com/nf-core/modules/blob/ec22f9a9079bd93ff8ec152349ead5f684edeb9f/modules/cooler/dump/main.nf#L29 , which makes the path invalid and cooler complain that the file doesn't exist. I suggest removing the `resolution` parameter from `cooler/dump` altogether because it's not used at all.
But then, in order to be able to chain both modules without a `map`, `cooler/cload` shouldn't output `cool_bin` at all either. Just meta and the cool file. How would that work for you ?

muffato:
Hi. I actually now think it'd be easier to address these comments in #1937 since the latter contains these changes. I'll close this PR and reraise my comments over there.

PR #1437, opened 2022-03-24
Add samtools to homer/maketagdirectory to use bam files

### LINKED ISSUE #1363: Homer maketagdirectory uses samtools
http://homer.ucsd.edu/homer/introduction/install.html

This might be something that should be fixed in the bioconda recipe now that I'm making the issue. Right now, the nf-core homer modules only accepts bed files. The snakemake wrapper adds samtools as a dependency https://snakemake-wrappers.readthedocs.io/en/stable/wrappers/homer/makeTagDirectory.html#software-dependencies

Emiller88:
Closes #1363

### DIFF
--- a/modules/homer/maketagdirectory/main.nf
+++ b/modules/homer/maketagdirectory/main.nf
@@ -26,9 +26,9 @@ process HOMER_MAKETAGDIRECTORY {
     """
     makeTagDirectory \\
         tag_dir \\
+        -genome $fasta \\
         $args \\
-        -genome $fasta
-        $bam \\
+        $bam
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
--- a/modules/homer/maketagdirectory/main.nf
+++ b/modules/homer/maketagdirectory/main.nf
@@ -4,7 +4,7 @@ process HOMER_MAKETAGDIRECTORY {
     tag "$meta.id"
     label 'process_medium'
 
-    conda (params.enable_conda ? "homer=4.11,samtools=1.11,r-base=4.0.2,bioconductor-deseq2=1.30.0,bioconductor-edger=3.32.0,perl=5.26.2" : null)
+    conda (params.enable_conda ? "bioconda::homer=4.11 bioconda::samtools=1.11 conda-forge::r-base=4.0.2 bioconda::bioconductor-deseq2=1.30.0 bioconda::bioconductor-edger=3.32.0 anaconda::perl=5.26.2" : null)
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
         'https://depot.galaxyproject.org/singularity/mulled-v2-29293b111ffe5b4c1d1e14c711264aaed6b97b4a:594338b771cacf1623bd27772b5e12825f8835f2-0' :
         'quay.io/biocontainers/mulled-v2-29293b111ffe5b4c1d1e14c711264aaed6b97b4a:594338b771cacf1623bd27772b5e12825f8835f2-0' }"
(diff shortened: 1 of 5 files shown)


modules/homer/maketagdirectory/main.nf
```diff
@@ -27,12 +27,13 @@ process HOMER_MAKETAGDIRECTORY {
...
         tag_dir \\
         $args \\
-        $bed \\
         -genome $fasta
+        $bam \\
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
         homer: $VERSION
+        samtools: \$(echo \$(samtools --version 2>&1) | sed 's/^.*samtools //; s/Using.*\$//')
```
FriederikeHanssen, line 36: `r-base=4.0.2,bioconductor-deseq2=1.30.0,bioconductor-edger=3.32.0,perl=5.26.2` are missing

modules/homer/maketagdirectory/meta.yml
```diff
@@ -3,23 +3,34 @@ description: Create a tag directory with the HOMER suite
...
       licence: ["GPL-3.0-or-later"]
+  - samtools:
+      description: |
+        SAMtools is a set of utilities for interacting with and post-processing
+        short DNA sequence read alignments in the SAM, BAM and CRAM formats, written by Heng Li.
+        These files are generated as output by short read aligners like BWA.
+      homepage: http://www.htslib.org/
+      documentation: hhttp://www.htslib.org/doc/samtools.html
+      doi: 10.1093/bioinformatics/btp352
+      licence: ["MIT"]
```
FriederikeHanssen, line 23: I am not quite sure where to draw the line, but deseq2 has a doi and publication https://bioconductor.org/packages/release/bioc/html/DESeq2.html, so I would say it woudn't hurt to add it here

modules/homer/maketagdirectory/meta.yml
```diff
@@ -3,23 +3,34 @@ description: Create a tag directory with the HOMER suite
...
 input:
   - meta:
       type: map
       description: |
         Groovy Map containing sample information
         e.g. [ id:'test', single_end:false ]
-  - bed:
+  - bam:
```
FriederikeHanssen, line 30: runs on both `bed` and `bam`, right. What about renaming to `input`?

Emiller88: I just saw you did that in #1445. I like input, but I also like when we use `reads` for raw reads in a fastq. I guess I'm also afraid of everything being named `input` :see_no_evil:

modules/homer/maketagdirectory/meta.yml
Emiller88: Good catch!

modules/homer/maketagdirectory/meta.yml
FriederikeHanssen: fair enough :D, I think I started this originally because all the variantcallers said `meta,bam` but we want to use `cram`. In some places i think `input_aligned` is used, which I also like, gives a bit more description. Is there an overlap between `bam`/`bed` content wise?

Emiller88: `aligned_reads`? `regions` in bed/bam case? I like `input_aligned` too.

I think sticking with bam here for now is fine until we get everyone's input on some naming conventions. I had to make this PR because using a `bed` while possible runs into memory issues and then homer silently crashes on some samples. So while it's possible, I don't want to encourage people to use it.

FriederikeHanssen: ok sounds reasonable 👍

Emiller88: https://github.com/nf-core/modules/issues/1450

Summarized here so we can follow up!

FriederikeHanssen (approved):
🚀

PR #1440, opened 2022-03-24
Split preseq lcextrap and ccurve

Emiller88:
Split them into two separate processes, and use correct name in lcextrap.

### DIFF
--- a/modules/preseq/ccurve/main.nf
+++ b/modules/preseq/ccurve/main.nf
@@ -1,0 +1,40 @@
+process PRESEQ_CCURVE {
+    tag "$meta.id"
+    label 'process_medium'
+    label 'error_ignore'
+
+    conda (params.enable_conda ? "bioconda::preseq=3.1.2" : null)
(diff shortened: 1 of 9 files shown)


modules/preseq/ccurve/main.nf
```diff
@@ -0,0 +1,40 @@
+process PRESEQ_CCURVE {
+    tag "$meta.id"
+    label 'process_medium'
+    label 'error_ignore'
```
FriederikeHanssen, line 4: haven't seen this yet, where does it come from?

Emiller88: The `error_ignore`? 

https://github.com/nf-core/rnaseq/blob/5aadd93370c836e12b777057eb2298fc7bc24fc9/conf/base.config#L43

Useful for qc checks that may fail, but you don't want to exit the whole workflow. I don't know why @drpatelh included it originally.

FriederikeHanssen: ah ok, never saw it :)

FriederikeHanssen (approved):

PR #1474, opened 2022-03-31
Add task.ext.args to phantompeakqualtools and finish the module

JoseEspinosa:
As explained [here](https://github.com/nf-core/chipseq/issues/220), when the data set is too big the `phantompeakqualtools` throws an exception. Adding the option fixes the problem `--max-ppsize=500000`. On this PR, `task.ext.args` to the  `phantompeakqualtools` command and also, all the missing bits of the module were added (tests + meta.yml file)

## PR checklist


drpatelh (approved):
So you went with `args` and not `args2` too? It can be changed later if required.

JoseEspinosa:
Yep thought it was going to be more simpler since there are no other arguments passed to the tool

PR #1485, opened 2022-04-04
Module/bclconvert

### LINKED ISSUE #1484: new module: bcl-convert
I think it would be good to have a module for [bcl-convert](https://support.illumina.com/sequencing/sequencing_software/bcl-convert/downloads.html)

matthdsm:
Add module for Illumina bcl-convert

## TODO
- [x] Fix stub for testing
- [x] Build docker image en push to docker hub. Needs to be done by someone with admin access to the `nfcore` organisation on docker hub @nf-core/core 

## PR checklist

Closes #1484 

- [x] Add a resource `label`


matthdsm:
Attempt at creating a module for `bcl-convert` as discussed on [slack](https://nfcore.slack.com/archives/CE6SDEDAA/p1649057339733939)

matthdsm:
I left in some TODO's for stuff I'm not sure about.
I've also added a stub for testing, but perhaps we're able to add some `PhiX` test data somewhere (if someone has a small run lying about)

matthdsm:
[RPM needed for docker build](https://github.com/nf-core/modules/files/8418508/bcl-convert-3.9.3-2.el7.x86_64.rpm.gz).
Please note the `.gz` extension is only to make github upload possible and should be removed

matthdsm:
Hi @ewels 

I fixed the stubs, most checks pass, except the conda one, because there's no recipe for `bcl-convert`.
Care to review? Or do I bother someone else with this?

Thanks
M

matthdsm:
There still seems to be something awry with the docker image.
versions.yml
```yaml
"test_bclconvert:BCLCONVERT":
    bclconvert: echo /tmp/tmp1jtsj6hl/f7/e63e3b6e66ca12a3d79b9b5d0ac3ac/.command.sh: line 31: bcl-convert: command not found
```

Which is strange because:
```bash
$ docker run --rm nfcore/bclconvert which bcl-convert
/usr/local/bin/bcl-convert
```

matthdsm:
docker image `nfcore/bclconvert:3.9.3` seems to be broken.
Could someone from @nf-core/core rebuild and push it.

Other than that, the module should be good to go.

matthdsm:
Awaiting new docker image, other than that it's ready

modules/bclconvert/main.nf
```diff
@@ -0,0 +1,84 @@
...
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+
+    """
+    bcl-convert \
+        $args \\
+        --output-directory \$PWD \
```
FriederikeHanssen, line 30: I would set ${prefix} here

modules/bclconvert/main.nf
```diff
@@ -0,0 +1,84 @@
...
+        --bcl-input-directory ${run_dir} \\
+        --sample-sheet ${samplesheet} \\
+        --bcl-num-parallel-tiles $task.cpus \\
+        --bcl-num-conversion-threads $task.cpus \\
+        --bcl-num-compression-threads $task.cpus \\
+        --bcl-num-decompression-threads $task.cpus
+
+    mkdir InterOp
```
FriederikeHanssen, line 38: I don't quite understand qhy you need to move the files here, since you don't `output` the directory

FriederikeHanssen (commented):
Looks good to me overal.. I am not experienced with creating docker containers, so i would someone else look over this

matthdsm:
Anyone know of a way to disable the conda check so this can pass the tests?

matthdsm:
@jfy133, if it's not too much of a hassle, could you please check out this one too? If this is merged I can get going on the demultiplex workflow

modules/bclconvert/meta.yml
```diff
@@ -0,0 +1,45 @@
+name: "bclconvert"
+description: Demultiplex Illumina BCL files
+keywords:
+  - demultiplex
```
jfy133, line 4:
```suggestion
  - demultiplex
  - illumina
  - fastq
```

modules/bclconvert/meta.yml
```diff
@@ -0,0 +1,45 @@
...
+      licence: "NA"
+
+input:
+  - samplesheet:
+      type: file
+      description: "Input samplesheet"
+      pattern: "*.{csv}"
+  - run_dir:
+      type: directory
+      description: "Input run directory"
```
jfy133, line 21:
```suggestion
      description: "Input run directory"
```

containing what? `bcl` files I would assume?

jfy133 (approved):
This is overall out of my comfort zone as there is a lot of extra custom stuff, but I think as a first version this is fine.

It might be good to get more eyes on this e.g. from @nf-core/core (I see Rike has already reviewed it)

modules/bclconvert/meta.yml
matthdsm: yep, bcl data and RunInfo.xml in the directory structure as the sequencer has written it

modules/bclconvert/main.nf
```diff
@@ -0,0 +1,84 @@
...
+    """
+    bcl-convert \
+        $args \\
+        --output-directory $PWD \\
+        --bcl-input-directory ${run_dir} \\
+        --sample-sheet ${samplesheet} \\
+        --bcl-num-parallel-tiles $task.cpus \\
+        --bcl-num-conversion-threads $task.cpus \\
+        --bcl-num-compression-threads $task.cpus \\
+        --bcl-num-decompression-threads $task.cpus
```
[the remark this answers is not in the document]
matthdsm: fixed!

matthdsm:
Without any further comments I'm going to go ahead and merge this, so I can proceed with downstream work.

modules/bclconvert/main.nf
```diff
@@ -0,0 +1,84 @@
+process BCLCONVERT {
+    tag 'demultiplexing'
```
Emiller88, line 2: This might want to be 
```suggestion
    tag '$samplesheet'
```
To support multiple runs. Otherwise, they'll all be tagged demultiplexing.

modules/bclconvert/main.nf
```diff
@@ -0,0 +1,84 @@
...
+    """
+    bcl-convert \
+        $args \\
+        --output-directory $PWD \\
+        --bcl-input-directory ${run_dir} \\
+        --sample-sheet ${samplesheet} \\
+        --bcl-num-parallel-tiles ${task.cpus / 4} \\
+        --bcl-num-conversion-threads ${task.cpus / 4} \\
+        --bcl-num-compression-threads ${task.cpus / 4} \\
+        --bcl-num-decompression-threads ${task.cpus / 4}
```
Emiller88, line 36: Is the reason for dividing the cores up documented somewhere?

Emiller88 (changes requested):
Two pieces of feedback

modules/bclconvert/main.nf
Emiller88: Yeah, there's nothing in the commit message https://github.com/nf-core/modules/commit/c426128325e5afa14dca0664ecca911dd76aae0a

modules/bclconvert/main.nf
```diff
@@ -0,0 +1,84 @@
...
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+
+    """
+    bcl-convert \
+        $args \\
+        --output-directory $PWD \\
```
maxulysse, line 30: I think this need to be escaped

modules/bclconvert/main.nf
maxulysse: :confused:

tests/modules/bclconvert/nextflow.config
```diff
@@ -0,0 +1,5 @@
+process {
+
+    publishDir = { "${params.outdir}/${task.process.tokenize(':')[-1].tokenize('_')[0].toLowerCase()}" }
+    
+}
```
ewels, line 5: Linting should be failing here I think.

ewels (approved):
LGTM

tests/modules/bclconvert/main.nf
```diff
@@ -0,0 +1,22 @@
...
+    output:
+    path "SampleSheet.csv"          ,emit: samplesheet
+    path "DDMMYY_SERIAL_FLOWCELL"   ,emit: run_dir
+
+    stub:
+    """
+    mkdir DDMMYY_SERIAL_FLOWCELL
+    echo "SampleSheet" > SampleSheet.csv
+    """
+}
```
maxulysse, line 17: no need to fake the process, we can just fake the input

modules/bclconvert/.gitignore
```diff
@@ -0,0 +1,2 @@
+bcl-convert
+*.rpm
```
maxulysse, line 2: do we really need to keep this `.gitignore` here?

modules/bclconvert/main.nf
jfy133: https://nf-co.re/docs/contributing/modules#resource-requirements point 3

This might be a good one for @Midnighter to add to his Structizer one... there is some discussion **somewhere** on slack where this was originally brought up :sweat_smile:

Midnighter: Not sure it's such a good fit for an ADR. Those are more supposed to answer why certain design decisions were made. Seems well placed as it is in the docs you linked to.

FriederikeHanssen: Resource requirements for multiple tools are documented here: https://nf-co.re/contributing/modules?q=resource#resource-requirements

FriederikeHanssen: Here is at least part of the discussion, if anyone is interested: https://nfcore.slack.com/archives/CJRH30T6V/p1632822087211400

Emiller88: I just meant this specific instance of this one, and why it's divided by 4 and just wanted to see if there were any Illumina docs on this being the correct usage.

modules/bclconvert/.gitignore
Emiller88: Why not? We did it with Cellranger. Just preventing foot guns for our future selves.

modules/bclconvert/main.nf
matthdsm: Hi @Emiller88,

The help message states this
```
--bcl-num-parallel-tiles arg           # of tiles to process in parallel (default 1)
--bcl-num-conversion-threads arg       # of threads for conversion (per tile, default # cpu threads)
--bcl-num-compression-threads arg      # of threads for fastq.gz output compression (per tile, default # cpu threads,
                                       or HW+12)
--bcl-num-decompression-threads arg    # of threads for bcl/cbcl input decompression (per tile, default half # cpu 
                                       threads, or HW+8. Only applies when preloading files)

```

It seems to me, `bcl-convert` is made to take over the whole node it runs on since it's meant to be container native (for usage on the dragen platform). I think it just assumes to use all cores available in to the container.

We could set `${task.cpus}` to `bcl-num-parallel-tiles` and all the rest to 1, but I think that would be detrimental to users on a slower FS. The exact configuration is very dependent on the type of run and the hardware you run it on.

TL;DR, the current config is more 'just to be safe' and not very performance tuned. If it were up to me, I'd revert this and set everything to `${task.cpus}`. We've had no issues so far with such a setting.

As allways, open to suggestions!

modules/bclconvert/.gitignore
matthdsm: I can take it or leave it. I modeled this after Cellranger. Perhaps keep it around, just to be safe.

modules/bclconvert/main.nf
matthdsm: Nice catch! Fixed it

tests/modules/bclconvert/main.nf
matthdsm: How do you mean? 
I'd be happy to be able to run a real world test, but I since `bcl-convert` is very particular about it's inputs and the structure of it, I think it would be hard to support this. Not to mention the size of the data needed. Even a very small MiSeq run is a couple of Gb.

modules/bclconvert/main.nf
FriederikeHanssen: Any reason not to have a possible custom prefix here as well?

matthdsm: The `$PWD` only designates an output directory but has no effect on the names of the data whatsoever.
Personally I don't really see the benefit of introducing another subdir into the mix, especially considering `bcl-convert` can also make its own subdirs, depending on the settings.

It's possible, but I wouldn't do it. We'd also have to modify the customisations after the process to collect the `InterOp` bin files

matthdsm: However, if there is a consensus or guideline as to how this should be handled, I think that should be followed.

Emiller88: Have you tried `.` instead of `$PWD`? That's more common to use, I agree on not having a custom prefix.

modules/bclconvert/main.nf
Emiller88: > the current config is more 'just to be safe' and not very performance tuned

This was the piece I was afraid of. Maybe a quick run with both options? It also looks like you could just set `--bcl-num-parallel-tiles ${task.cpus}` and let `bcl-convert` figure out how it wants to use those threads.

modules/bclconvert/Dockerfile
```diff
@@ -0,0 +1,10 @@
+# Dockerfile to create container with bcl-convert
+# Push to nfcore/bclconvert:<VER>
+
+FROM debian:bullseye-slim
+LABEL authors="Matthias De Smet <matthias.desmet@ugent.be>" \
+    description="Docker image containing bcl-convert"
+# Disclaimer: this container is not provided nor supported by Illumina
+
```
MrMarkW, line 8: I got this error when using this image in my Tower environment. `Command 'ps' required by nextflow to collect task metrics cannot be found`

```suggestion
# 'ps' command is need by some nextflow executions to collect system stats
# Install procps and clean apt cache
RUN apt-get install -y \
    procps \
    && apt-get clean -y && rm -rf /var/lib/apt/lists/*
```

modules/bclconvert/main.nf
matthdsm: Actually, I agree. The true parallelisation comes from the parallel handling of the tiles. We could argue the last 3 args are purely for performance tweaking.

How about we drop the latter 3 args and let users start with the defaults, enabling them to tweak as they go using `ext.args`?

modules/bclconvert/main.nf
matthdsm: I thought of using `.` but personally I think `PWD` is more readable. A single dot is very easily looked over (in my experience), hence my perference for `$PWD`

modules/bclconvert/Dockerfile
matthdsm: Agreed!
@ewels , could you rebuild and push the image again?

matthdsm:
Any more suggestions/remarks?

modules/bclconvert/main.nf
Emiller88: I'm just trying to keep a consistent style, `.` is more common in the `nf-core/modules`. 

After looking at this again, should the `$` be escaped with `\`?

modules/bclconvert/main.nf
Emiller88: Sounds good to me, I'm also happy to leave them/add them back in, I just think we should test it to make sure we're utilizing all of the resources.

matthdsm:
Thanks! Cool to merge? Going once?

PR #1500, opened 2022-04-06
Motus

jianhong:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`

### DIFF
--- a/modules/motus/downloaddb/main.nf
+++ b/modules/motus/downloaddb/main.nf
@@ -1,0 +1,34 @@
+process MOTUS_DOWNLOADDB {
+    label 'process_low'
+    label 'error_ignore'
+
+    conda (params.enable_conda ? "bioconda::motus=3.0.1" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/motus:3.0.1--pyhdfd78af_0' :
+        'quay.io/biocontainers/motus:3.0.1--pyhdfd78af_0' }"
+
+    input:
+    path motus_downloaddb
+
+    output:
+    path "db_mOTU"                                       , emit: db
+    path "versions.yml"                                  , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    """
+    ## must copy file to working directory,
+    ## otherwise the reference_db will be download to other than current directory
+    cp $motus_downloaddb dwdDB.py
+    python dwdDB.py \\
+        -t $task.cpus
+    rm dwdDB.py
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        mOTUs: \$(grep motus db_mOTU/db_mOTU_versions | sed 's/motus\\t//g')
+    END_VERSIONS
+    """
+}
--- a/modules/motus/merge/main.nf
+++ b/modules/motus/merge/main.nf
@@ -1,0 +1,37 @@
+process MOTUS_MERGE {
+    label 'process_low'
+    label 'error_ignore'
+
+    conda (params.enable_conda ? "bioconda::motus=3.0.1" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/motus:3.0.1--pyhdfd78af_0' :
+        'quay.io/biocontainers/motus:3.0.1--pyhdfd78af_0' }"
+
+    input:
+    tuple val(meta), path(motus_outs)
+    path db
+
+    output:
+    tuple val(meta), path("${prefix}_motus.out")         , emit: merged_out
+    path "versions.yml"                                  , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args   = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "merged"
+    """
+    motus merge \\
+        -i "${motus_outs.join(',').trim()}" \\
+        -t $task.cpus \\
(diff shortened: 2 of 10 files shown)


modules/motus/profile/main.nf
```diff
@@ -0,0 +1,51 @@
...
+        -I ${prefix}.bam \\
+        -o ${prefix}_motus.out
+    # for krona
+    motus profile \\
+        -i ${prefix}.bam \\
+        -t $task.cpus \\
+        -n ${prefix} \\
+        -db ${db} \\
+        -C parenthesis \\
+        -o ${prefix}_motus.out.krona
```
jfy133, line 42: We don't want to run the same profiling twice here - the user can specify `-C` themselves in via $args in the original command.

modules/motus/downloaddb/main.nf
```diff
@@ -0,0 +1,34 @@
...
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    """
+    ## must copy file to working directory,
+    ## otherwise the reference_db will be download to other than current directory
+    cp $motus_downloaddb dwdDB.py
+    python dwdDB.py \\
```
jfy133, line 25: I'm not sure where/what this is coming from? The mOTUs tutorial doesn't reference anything to do with this script?

It looks like the mOTUs bioconda recipe comes with the datbaase included?

modules/motus/downloaddb/main.nf
```diff
@@ -0,0 +1,34 @@
+process MOTUS_DOWNLOADDB {
+    label 'process_low'
+    label 'error_ignore'
```
jfy133, line 3: Where is this coming from? A pipeline developer would define the error strategy at pipeline level so should not be included here (applies for the other tools too)

modules/motus/merge/main.nf
```diff
@@ -0,0 +1,37 @@
...
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/motus:3.0.1--pyhdfd78af_0' :
+        'quay.io/biocontainers/motus:3.0.1--pyhdfd78af_0' }"
+
+    input:
+    tuple val(meta), path(motus_outs)
+    path db
+
+    output:
+    tuple val(meta), path("${prefix}_motus.out")         , emit: merged_out
```
jfy133, line 15: You also need to support the `-B` flag which (optionally) generates  BIOm files

modules/motus/merge/main.nf
```diff
@@ -0,0 +1,37 @@
...
+    label 'error_ignore'
+
+    conda (params.enable_conda ? "bioconda::motus=3.0.1" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/motus:3.0.1--pyhdfd78af_0' :
+        'quay.io/biocontainers/motus:3.0.1--pyhdfd78af_0' }"
+
+    input:
+    tuple val(meta), path(motus_outs)
+    path db
```
jfy133, line 12: Again, I'm not sure where this `db` thing is coming from...

modules/motus/merge/main.nf
jfy133: You should also remove `meta` as you're combining multiple samples anyway.

modules/motus/merge/main.nf
```diff
@@ -0,0 +1,37 @@
...
+    script:
+    def args   = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "merged"
+    """
+    motus merge \\
+        -i "${motus_outs.join(',').trim()}" \\
+        -t $task.cpus \\
+        -db ${db} \\
+        $args \\
+        -o ${prefix}_motus.out
```
jfy133, line 30:
```suggestion
        -o ${prefix}.out
```

I would just hardcode this rather than using prefix, as you will always be combining multiple samples. e.g. `motus_merged_profiles.txt` or similar

jfy133 (changes requested):
Hi @jianhong 

The tests aren't running because the created pytest_modules.yml aren't following the scheme of one upper-level entry per 'subtool'.

Did you create the modules with `nf-core modules create`? Because this shouldn't happen if so...

Also it would be preferential to have one PR per module (so one for downloaddb, one for merge, one for profile) as this makes it easier to review and also debug problems - as we can see with the broken tests.

I've provided some other comments in the meantime when you redo the PRs

modules/motus/profile/main.nf
```diff
@@ -0,0 +1,51 @@
...
+    script:
+    def args   = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "${meta.id}"
+    def inputs = meta.single_end ? "-s $reads" : "-f ${reads[0]} -r ${reads[1]}"
+    """
+    motus profile \\
+        $inputs \\
+        -t $task.cpus \\
+        -n ${prefix} \\
+        -db ${db} \\
```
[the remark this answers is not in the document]
jianhong: En, they did not expose the parameter here. But if you go to the source code: https://github.com/motu-tool/mOTUs/blob/master/motus/motus.py#L111-L122
you will see they ask `db_mOTU` installed by `downloadDB` into the script path. However, in docker and singularity setup, it will be difficult to install files into the container. Then I used `-db` parameter to pass the installed local database to the script.

jfy133:
Ugh ok fair enough! Ignore my comments about the database flag then.

However I'm unsure what to do for the database download thing. I think I
would rather not have that as a separate module as it's not an actual
tool... Maybe we should ask in core what to do there

---

jfy133:
Close in favour of https://github.com/nf-core/modules/pull/1616

PR #1504, opened 2022-04-07
Fix Controlfreec: Add stub runs to test single sample input & make conda work with R scripts

### LINKED ISSUE #1455: [BUG] controlfreec fails with conda
Follow up to

https://github.com/nf-core/modules/pull/1452#discussion_r835546725

In the conda tests
```
cat: /usr/local/bin/assess_significance.R: No such file or directory
mv: cannot stat '*.p.value.txt': No such file or directory
```
We should figure out in assess_signifcance can just be called.

FriederikeHanssen:
The test data is not suited to pass as single sample through Controlfreec, because the tool requires a matched normal for wes/panel data. 

I therefore proposing to at least add a stub run test for single sample input

also closes #1455 
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`


tests/modules/controlfreec/freec/test.yml
```diff
@@ -20,4 +20,26 @@
...
+
+- name: controlfreec test_controlfreec_freec
+  command: nextflow run tests/modules/controlfreec/freec -entry test_controlfreec_freec_single -c tests/config/nextflow.config -stub-run
+  tags:
+    - controlfreec
+    - controlfreec/freec
+  files:
+    - path: output/controlfreec/config.txt
+    - path: output/controlfreec/test.mpileup.gz_control.cpn
+      md5sum: d41d8cd98f00b204e9800998ecf8427e
```
maxulysse, line 32: I think we don't need md5sum for empty files

FriederikeHanssen: ok

maxulysse (approved):

FriederikeHanssen:
hm I am stuck. 

I created the controlfreec conda environment on my machine. The scripts `assess_significance.R` and `makeGrpah.R` are both available in my PATH variable. Typing their name in directly also works. As soon as I add `cat` (as mentioned in the documentation), the scripts are not available anymore. I have no idea why, googleing so far has not helped. If anyone has an idea...

FriederikeHanssen:
@nf-core/modules-team anyone know how to get `assess_significance` to work with conda? 
I have tried setting the path to :
`~/miniconda3/envs/*/bin/` and also `~/miniconda3/envs/controlfreec/bin/`, `~/miniconda3/envs/control-freec/bin/` . I don't understand to where nextflow installs this to.

modules/controlfreec/assesssignificance/main.nf
```diff
@@ -21,7 +21,7 @@ process CONTROLFREEC_ASSESSSIGNIFICANCE {
     def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
     """
-    cat /usr/local/bin/assess_significance.R | R --slave --args ${cnvs} ${ratio}
+    cat \$(which assess_significance.R) | R --slave --args ${cnvs} ${ratio}
```
maxulysse, line 24: <3

JoseEspinosa (commented):
LGTM!
Only one thing from the top of my head that will be indeed a more general question to discuss for all modules, adding only the `stub-run` to test the modules could not be a little bit misleading?
My point is that in fact, we are just testing several `touch` commands and that the version is correctly output but the code of the real module (i.e. the `script` section) is not tested in any way.

FriederikeHanssen:
I agree, that is why I ahve been very hesitant to use stub tests anywhere. And also why I added both tests - once with stub and once without. The problem here is that paired test data passes, but the tools also accepts unpaired data. Unfortuantely, the test data we have makes the tool then complain ( would need full size wgs data as far as I know). 

Also I wanted to add some CI tests to sarek, which then could also only test the paired subworkflow but not the unpaired one. While not ideal to just test the file existence it now at least allows me to test that the subworkflow as a whole works without modifying the module code locally.

FriederikeHanssen:
Conda test now fails because of mismatching md5sum. Ignoring this

JoseEspinosa:
OK, fair enough :rocket:

PR #1509, opened 2022-04-08
Tool/snap aligner paired

matthdsm:
## PR checklist
Add aligner (part 1) to snap-aligner

- [x] Add a resource `label`

### DIFF
--- a/modules/snapaligner/paired/main.nf
+++ b/modules/snapaligner/paired/main.nf
@@ -35,7 +35,7 @@ process SNAPALIGNER_PAIRED {
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
-        snapaligner: \$(snap-aligner 2>&1| head -n 1 | sed 's/^.*version //')
+        snapaligner: \$(snap-aligner 2>&1| head -n 1 | sed 's/^.*version //;s/\.\$//')
     END_VERSIONS
     """
 }
--- a/modules/snapaligner/paired/main.nf
+++ b/modules/snapaligner/paired/main.nf
@@ -1,0 +1,31 @@
+process SNAPALIGNER_PAIRED {
+    tag '$meta.id'
+    label 'process_high'
+
+    conda (params.enable_conda ? "bioconda::snap-aligner=2.0.1" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/snap-aligner:2.0.1--hd03093a_1':
+        'quay.io/biocontainers/snap-aligner:2.0.1--hd03093a_1' }"
+
+    input:
+    tuple val(meta), path(bam)
+
+    output:
+    tuple val(meta), path("*.bam"), emit: bam
+    path "versions.yml"           , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    snap-aligner paired <index-dir> <inputFile(s)> [<options>] where <input file(s)> is a list of files to process.
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        snapaligner: \$(snap-aligner 2>&1| head -n 1 | sed 's/^.*version //')
+    END_VERSIONS
+    """
+}
--- a/modules/snapaligner/paired/main.nf
+++ b/modules/snapaligner/paired/main.nf
@@ -8,7 +8,8 @@ process SNAPALIGNER_PAIRED {
         'quay.io/biocontainers/snap-aligner:2.0.1--hd03093a_1' }"
 
     input:
-    tuple val(meta), path(bam)
+    tuple val(meta), path(reads)
+    path index
 
     output:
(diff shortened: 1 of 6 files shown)


sateeshperi (commented):
LGTM! Lets get another review before merge!

modules/snapaligner/paired/main.nf
```diff
@@ -0,0 +1,41 @@
...
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+
+    """
+    mkdir -p index
+    mv $index index/
```
jfy133, line 27: Why not just pass $index as an directory to the module in the first place?

modules/snapaligner/paired/meta.yml
```diff
@@ -0,0 +1,48 @@
...
+input:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. [ id:'test', single_end:false ]
+  - reads:
+      type: file
+      description: List of input fastq files of size 2 for fastq or 1 for bam
+      pattern: "*.{fastq.gz, fastq, bam}"
```
jfy133, line 26:
```suggestion
      pattern: "*.{fastq.gz, .fq.gz, fastq, fq, bam}"
```

jfy133 (approved):
Couple of minor suggestions and one question

modules/snapaligner/paired/main.nf
matthdsm: Because I didn't know the syntax 😇  and the output from `index` are just the files

matthdsm: Suggestions to improve my knowledge and the code quality are welcome!

jfy133: Ok fair enough!

modules/snapaligner/paired/main.nf
```diff
@@ -0,0 +1,41 @@
...
+    snap-aligner paired \\
+        index \\
+        ${reads.join(" ")} \\
+        -o -bam ${prefix}.bam \\
+        -t ${task.cpus} \\
+        $args
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        snapaligner: \$(snap-aligner 2>&1| head -n 1 | sed 's/^.*version //')
```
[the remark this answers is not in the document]
matthdsm: fixed

modules/snapaligner/paired/main.nf
jfy133: No, BWA takes a diretory but uses a `find` command to find the relevant files to get the 'basename' and path for the reference genome.

Fine either way tbh

matthdsm:
Unrelated question. I saw `bwamem2` does something like this 
```
```
to add the RG. Is that something that can be done from the config file `ext.args`?  Or should I add it to command line?

matthdsm:
It doesn't, but for the sake of completeness, you can add a custom RG through that option.
That's something we'd like to do too, hence why I'm asking 😄

matthdsm:
Wouldn't removing it break some stuff in sarek and other pipelines using bwa/bwamem2?

jfy133:
> Wouldn't removing it break some stuff in sarek and other pipelines using bwa/bwamem2?

@maxulysse made the mistake in the first place, so that's his problem ;).

But not necessarily no, when you update a module with `nf-core modules update <tool>` it offers you to display the diff so you can see what has changed. Given each module is installed into the pipeline with a git commit hash as a 'version', it means you can keep the old one if you want ffor as long as you want.

Lucky for him I don't have the time at the moment ;)

maxulysse:
> > Wouldn't removing it break some stuff in sarek and other pipelines using bwa/bwamem2?
> 
> @maxulysse made the mistake in the first place, so that's his problem ;).
> 
> But not necessarily no, when you update a module with `nf-core modules update <tool>` it offers you to display the diff so you can see what has changed. Given each module is installed into the pipeline with a git commit hash as a 'version', it means you can keep the old one if you want ffor as long as you want.
> 
> Lucky for him I don't have the time at the moment ;)

I hate you soooo much <3

jfy133:
<3

ramprasadn:
@jfy133 @maxulysse Have a question. If the option to pass the RGs were moved to config files, would we still be able to assign read groups dynamically? So for instance, raredisease pipeline creates RGs dynamically from the sample sheet. https://github.com/nf-core/raredisease/blob/e22443cfc3770667ceff9978f457079bd3869797/subworkflows/local/check_input.nf#L36 Could this be somehow done in the configs?

jfy133:
Yes, you can just insert into the configs;

e.g.
```
ext.args = { "-R ${meta.readgroups}" }
```

We are actually using this heavily in nf-core/taxprofiler, where actually [our entire argument/parameter](https://github.com/nf-core/taxprofiler/blob/1ac5a129a8ea3a212682e5545a8e657de16559bf/conf/modules.config#L199) set for the tool is passed via a meta element

maxulysse:
Yes of course, you just use ext.args to print out meta.read_group

ramprasadn:
Ah, like that. I misunderstood your point @jfy133. I thought you were talking about removing readgroup information from meta, now it all makes sense 😅

PR #1525, opened 2022-04-13
update to kraken2: breaking change - output channels renamed

lescai:
## PR checklist

Closes #XXX 

- [ ] The existing version of the module did not output a taxonomic classification of *each* input reads, which is needed for the pipeline hgtseq. by missing the command option --output this classification is output on stderr. After conversation on Slack, it was also realised that outputting classified fastq files would be best an optional output. The module has been therefore recoded by adding to additional channels, with boolean values to indicate if fastq classified/unclassified reads should be written and whether classification of each reads should be returned.
- [ ] Tests have been successfully run with singularity
- [ ] No new tools added
- [ ] Test data already online
- [ ] Add a resource `label`

### DIFF
--- a/modules/antismash/antismashlitedownloaddatabases/main.nf
+++ b/modules/antismash/antismashlitedownloaddatabases/main.nf
@@ -1,0 +1,46 @@
+process ANTISMASH_ANTISMASHLITEDOWNLOADDATABASES {
+    label 'process_low'
+
+    conda (params.enable_conda ? "bioconda::antismash-lite=6.0.1" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/antismash-lite:6.0.1--pyhdfd78af_1' :
+        'quay.io/biocontainers/antismash-lite:6.0.1--pyhdfd78af_1' }"
+
+    /*
+    These files are normally downloaded by download-antismash-databases itself, and must be retrieved for input by manually running the command with conda or a standalone installation of antiSMASH. Therefore we do not recommend using this module for production pipelines, but rather require users to specify their own local copy of the antiSMASH database in pipelines. This is solely for use for CI tests of the nf-core/module version of antiSMASH.
+    Reason: Upon execution, the tool checks if certain database files are present within the container and if not, it tries to create them in /usr/local/bin, for which only root user has write permissions. Mounting those database files with this module prevents the tool from trying to create them.
+    */
+
+    containerOptions {
+        workflow.containerEngine == 'singularity' ?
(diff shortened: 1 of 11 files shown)


jfy133 (commented):
I see you've ot finished yet (still TODO strings etc), but overall is looking good

One bit of reedback, I find classified/unclassified/classifiedreads to be quite ambiguous what they are... maybe we can come up with a better naming scheme for both flags ?

What about for input channels:

```
save_output_fastqs
save_reads_report
```

And for variables:

```
classified_reads_fastq
unclassified_reads_fastq
classified_reads_report
```

or something like that? I think it'll be faster for a developer to parse that way.

lescai:
Hi @jfy133 
completely agree. 
I've modified as 
`save_output_fastqs` and
`save_reads_assignment` 
(this because kraken has a default output called "report" so we don't mix the two).

as a consequence the outputs have been renamed to
`classified_reads_fastq`
`unclassified_reads_fastq`
`classified_reads_assignment`
`report`

both docker and singularity tests are passing here and on my HPC, except the usually annoying conda here.

modules/kraken2/kraken2/meta.yml
```diff
@@ -27,25 +27,40 @@ input:
...
   - meta:
       type: map
       description: |
         Groovy Map containing sample information
         e.g. [ id:'test', single_end:false ]
-  - classified:
+  - classified_reads_fastq:
       type: file
       description: |
         Reads classified to belong to any of the taxa
```
jfy133, line 49:
```suggestion
        Reads classified as belonging to any of the taxa
```

modules/kraken2/kraken2/meta.yml
```diff
@@ -27,25 +27,40 @@ input:
...
         on the Kraken2 database.
       pattern: "*{fastq.gz}"
-  - unclassified:
+  - unclassified_reads_fastq:
       type: file
       description: |
         Reads not classified to belong to any of the taxa
```
jfy133, line 55:
```suggestion
        Reads not classified to any of the taxa
```

modules/kraken2/kraken2/meta.yml
```diff
@@ -27,25 +27,40 @@ input:
...
         on the Kraken2 database.
       pattern: "*{fastq.gz}"
-  - txt:
+  - classified_reads_assignment:
+      type: file
+      description: |
+        Kraken2 output file indicating the taxonomic assignment of
+        each input reads
```
jfy133, line 62:
```suggestion
        each input read
```

jfy133 (approved):
A couple of minor phrasing changes in the meta.

Make sure to that the  commit message is very clear when squash/merge that this is a 'breaking' change/major update (or just that you're renaming the output channels)

Issue #1517, opened 2022-04-12
RFC: New submodule bam_postprocessing


matthdsm:
Hi, I'd like to suggest a new submodule or suite of submodules for bam postprocessing. 
This would consist of the common steps taken after alignment.
- sort
- markdup
- optional BQSR

Since there are multiple toolchains that offer this, I would suggest adding a "suite" of subworkflow, so each can be added on demand.

Examples:
- Biobambam suite
```mermaid

flowchart LR
        BIOBAMBAM_BAM([BAM])                --> SPLIT[Split tool TBD]
        SPLIT                               --split by chromosome--> BAMSORMADUP[BamSorMaDup]
        BAMSORMADUP                         --sort/mark duplicates-->  BIOBAMBAM_SORTBAM([Postprocessed BAM])
```

- Elprep suite
```mermaid
flowchart LR
        ELPREP_BAM([BAM])                   --> ELPREP_SPLIT[Elprep split]
        ELPREP_SPLIT                        --split by chromosome--> ELPREP_FILTER[Elprep filter]
        ELPREP_FILTER                       --sort/mark duplicates--> ELPREP_MERGE[Elprep merge]
        ELPREP_FILTER                       --BQSR/variant calling--> ELPREP_MERGE[Elprep merge]
        ELPREP_MERGE                        --> ELPREP_SORTBAM([Postprocessed BAM])
        ELPREP_MERGE                        --> ELPREP_GVCF(["Optional gVCF"])
```

- GATK suite
-- no example

- [x] This module does not exist yet with the [`nf-core modules list`](https://github.com/nf-core/tools#list-modules) command
- [x] There is no [open pull request](https://github.com/nf-core/modules/pulls) for this module
- [x] There is no [open issue](https://github.com/nf-core/modules/issues) for this module
- [x] If I'm planning to work on this module, I added myself to the `Assignees` to facilitate tracking who is working on the module

FriederikeHanssen:
Sarek has some bam_postprocessing subworkflows based on GATk Best Practices that @maxulysse and me have been planning to add to subworkflows at some point. I think they would fit here:
Merge or MarkDup -> QC (Qualimap, EstimateLibraryComplexity,samtools) ->  BQSR (split by GATk recommended intervals; nothing needs be done, the tool takes care of it) -> GatherTables -> ApplyBQSR  -> QC (Qualimap, EstimateLibraryComplexity,samtools)

matthdsm:
[...]

PR #1535, opened 2022-04-19
Update bamtools split

matthdsm:
Bump version for `bamtools split`

Prefix command with `bamtools merge` to enable merging bam chunks before splitting per chromosome, cfr `biobambam/bamsormadup` and `elprep/split`

- [x] Add a resource `label`


matthdsm:
At the moment, this PR intoduces breaking changes to the module (extra input). This can be avoided by omitting the extra list input for `bamtools merge`.
@jfy133, I'd love your opinion on this.

sguizard:
Following the doc, the module should be as atomic as possible. 
@drpatelh Is it possible to have samtools merge and samtools split in a same module? Or should they have their own modules? 

Otherwise, the code Looks good to me. 
I think a second testing the merge and split two or more bams should be added.

matthdsm:
Since both commands are part of `bamtools`, I don't think it's an issue (IMO). `biobambam/bamsormadup` also prepends `bamcat` from the same package to enable merging.

drpatelh:
So why are we using both of these modules exactly? What if I didn't want to merge anything beforehand?

matthdsm:
If you don't want to merge anything you use a single file as input, like you would beforehand. Nothing's changed for this use case.

The purpose here is to be able to merge bam chunks from an aligner (for example) and split them into chunks per chromosome for parallel processing, after which they get merged again.

matthdsm:
This relates to the discussion in #1517
Also, to avoid confusion, @sguizard mentioned `samtools`, but I suppose this is a typo, because all tools used are part of `bamtools`

sguizard:
> This relates to the discussion in #1517 Also, to avoid confusion, @sguizard mentioned `samtools`, but I suppose this is a typo, because all tools used are part of `bamtools`
Oups! corrected!

drpatelh (changes requested):
Yeah, these really belong in separate modules. I could imagine using `bamtools/merge` exclusively without `bamtools/split` which will make more modules usable.

I think we need to be careful how we add/change some of these modules so as to adhere to the guidelines we have written. nf-core/modules aren't necessarily being written to make sub-workflows more efficient - they need to be functional and work in their own right as individual units. Also, each module may need to be used in any sub-workflow and not just the ones linked to in the issue.

Lemme know when you have made the changes and I can review.

matthdsm:
I agree you'd want to keep these separate, but in this case I'd argue there's no loss of functionality, no added dependencies and just a little added complexity, while adding the benefit of being able to use multiple files as input.
This also avoids having to store an extra intermediate bam file (satisfying the "true benefits from piping" requirement).

If you're sternly against it, I'll just close the PR, because there are no meaningful additions then. `bamtools merge` can be added separately afterwards.

matthdsm:
To further argue my case, the modifications even had no influence on the checksums in `test.yml`

drpatelh:
Ok. As a compromise, how about we:

- Add a separate `bamtools/merge` module anyway to cater for it being used stand-alone.
- Add some tests for single and individual BAM files in this module. 
- Instead of hard-coding `bamtools merge` in the `script` section it may make sense to only use it when running on multiple BAM files. Should be able to do this in the pre-amble logic in the `script` section.

matthdsm:
I've added the extra test, but get the following complaint:
```
CRITICAL Error running test workflow (exit code 1)                                                                                                                                            __main__.py:585
         N E X T F L O W  ~  version 21.10.6                                                                                                                                                                 
         Launching `tests/modules/bamtools/split/main.nf`  - revision: 250e892c20                                                                                                                            
         [-        ] process > test_bamtools_split_multiple:BAMTOOLS_SPLIT_MULTIPLE -                                                                                                                        
                                                                                                                                                                                                             
         [-        ] process > test_bamtools_split_multiple:BAMTOOLS_SPLIT_MULTIPLE -                                                                                                                        
         Error executing process > 'test_bamtools_split_multiple:BAMTOOLS_SPLIT_MULTIPLE (test)'                                                                                                             
                                                                                                                                                                                                             
         Caused by:                                                                                                                                                                                          
           Process `test_bamtools_split_multiple:BAMTOOLS_SPLIT_MULTIPLE` input file name collision -- There are multiple input files for each of the following file names:                                  
         test.paired_end.sorted.bam                                                                                                                                                                          
                                                                                                                                                                                                             
                                                                                                                                                                                                             
         Tip: view the complete command output by changing to the process work dir and entering the command `cat .command.out`                                                                               
                                                                                                                                                                                                             
                                                                                                                                                                                                             
                                                                                                                                                                                                             
         [-        ] process > test_bamtools_split_multiple:BAMTOOLS_SPLIT_MULTIPLE -                                                                                                                        
         Error executing process > 'test_bamtools_split_multiple:BAMTOOLS_SPLIT_MULTIPLE (test)'                                                                                                             
                                                                                                                                                                                                             
         Caused by:                                                                                                                                                                                          
           Process `test_bamtools_split_multiple:BAMTOOLS_SPLIT_MULTIPLE` input file name collision -- There are multiple input files for each of the following file names:                                  
         test.paired_end.sorted.bam                                                                                                                                                                          
                                                                                                                                                                                                             
                                                                                                                                                                                                             
         Tip: view the complete command output by changing to the process work dir and entering the command `cat .command.out`           
```

Any idea how I fix this name collision?

matthdsm:
- I add the extra module later
- I added the extra test for multiple files
- I'm not really sure if there's any added benefit to throwing in a bunch of conditionals. The output for a single file is identical with or without the merge.

drpatelh:
Thanks! 

> I'm not really sure if there's any added benefit to throwing in a bunch of conditionals. The output for a single file is identical with or without the merge.

It's pointless streaming file contents via `bamtools/merge` if it isn't required? Might as well reduce I/O and use `bamtools/split` directly.

modules/bamtools/split/main.nf
```diff
@@ -20,10 +20,14 @@ process BAMTOOLS_SPLIT {
     script:
     def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
+    def input_list = bam.collect{"-in $it"}.join(' ')
+    def stub =  !args.contains("-stub") ? "-stub ${prefix}" : ""
+
```
drpatelh, line 25:
```suggestion
```

modules/bamtools/split/main.nf
```diff
@@ -20,10 +20,14 @@ process BAMTOOLS_SPLIT {
...
     """
-    bamtools \\
-        split \\
-        -in $bam \\
+    bamtools merge \\
+        ${input_list} \\
+    | bamtools split \\
+        $stub \\
         $args
```
drpatelh, line 31:
```suggestion
    bamtools \\
        merge \\
        $input_list \\
        | bamtools \\
            split \\
            -stub $prefix \\
            $args
```

modules/bamtools/split/meta.yml
```diff
@@ -23,8 +23,13 @@ input:
         e.g. [ id:'test', single_end:false ]
   - bam:
       type: file
-      description: A BAM file to split
+      description: A list of BAM files to merge and then split
```
drpatelh, line 26:
```suggestion
      description: A list of one or more BAM files to merge and then split
```

modules/bamtools/split/meta.yml
```diff
@@ -23,8 +23,13 @@ input:
...
       pattern: "*.bam"
+  - bam_list:
+      type: file
+      description: |
+        Optional input file containing bam files to merge before splitting,
+        one line per file
```
drpatelh, line 32:
```suggestion
```

tests/modules/bamtools/split/test.yml
```diff
@@ -1,10 +1,23 @@
-- name: bamtools split test_bamtools_split
-  command: nextflow run ./tests/modules/bamtools/split -entry test_bamtools_split -c ./tests/config/nextflow.config -c ./tests/modules/bamtools/split/nextflow.config
+- name: bamtools split test_bamtools_split_single_input
+  command: nextflow run tests/modules/bamtools/split -entry test_bamtools_split_single_input -c tests/config/nextflow.config
```
drpatelh, line 2:
```suggestion
  command: nextflow run ./tests/modules/bamtools/split -entry test_bamtools_split_single_input -c ./tests/config/nextflow.config -c ./tests/modules/bamtools/split/nextflow.config
```

drpatelh: This is the syntax we use in all other modules so we should keep it the same. Helps when programmatically updating all modules at the same time if required.

tests/modules/bamtools/split/test.yml
```diff
@@ -1,10 +1,23 @@
...
-    - path: output/bamtools/test.paired_end.sorted.REF_chr22.bam
+    - path: output/bamtools/test.REF_chr22.bam
       md5sum: b7dc50e0edf9c6bfc2e3b0e6d074dc07
-    - path: output/bamtools/test.paired_end.sorted.REF_unmapped.bam
+    - path: output/bamtools/test.REF_unmapped.bam
       md5sum: e0754bf72c51543b2d745d96537035fb
+    - path: output/bamtools/versions.yml
+
+- name: bamtools split test_bamtools_split_multiple
+  command: nextflow run tests/modules/bamtools/split -entry test_bamtools_split_multiple -c tests/config/nextflow.config
```
drpatelh, line 14:
```suggestion
    command: nextflow run ./tests/modules/bamtools/split -entry test_bamtools_split_multiple -c ./tests/config/nextflow.config -c ./tests/modules/bamtools/split/nextflow.config
```

tests/modules/bamtools/split/test.yml
matthdsm: That's weird, because this file was generated by `nf-core`

matthdsm:
> Might as well reduce I/O and use `bamtools/split` directly.

That is also true of course, although I'm not really sure if a pipe add's all that much IO, since nothing is written out

matthdsm:
@drpatelh, would you care to re-review and approve so this can be merged?

drpatelh (approved):
🚀 Thanks!

PR #1539, opened 2022-04-20
Update samtools view to add input path(index)

Lucpen:
## PR checklist

Closes #XXX 

- [X] I added as input path(index) to be able to subsample according to a chromosomal area.
- [] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [] If necessary, include test data in your PR.
- [X ] Emit the `versions.yml` file.
- [X] Add a resource `label`


jfy133 (changes requested):
Overall looks OK to me, however you need to specify the new requirement in `meta.yml`

FriederikeHanssen (commented):
HI! sorry that it took a few days, since this change will affect virtually all pipelines I was trying to find time to dig through the manual. 
@nf-core/modules-team  This PR will affect some of the existing subworkflows. They are not part of the testing at the moment. Should they all be updated too? I lost track for the consensus on this.  Some of them still use an older syntax so they need an overhaul at some point.

Lucpen:
@jfy133 thanks for noticing! I have added it to the .yml
@FriederikeHanssen thanks for taking the time, you are right, it will break some subworkflows. I guess that the update will have to be done at some point anyway because more people will need to use the samtools view in the same way as me. However, I do see how it will be painful to update everything so let me know you think I should keep it locally.

FriederikeHanssen:
@Lucpen no no I think it is good to add these changes. 😀  And I think @jfy133 in particularhas been advocating to add all possible input files to a module to avoid having to go back in the future like here. Just want to make sure that  everything is fine with the subworkflows

modules/samtools/view/meta.yml
```diff
@@ -25,6 +25,10 @@ input:
       type: file
       description: BAM/CRAM/SAM file
       pattern: "*.{bam,cram,sam}"
+  - index:
+      type: optional file
+      description: BAM.BAI/CRAM.CRAI file
+      pattern: "*.{bam.bai,cram.crai}"
```
jfy133, line 31:
```suggestion
      pattern: "*.{.bai,.crai}"
```

jfy133 (approved):
LGTM

@FriederikeHanssen I'm being the proverbial ostrich with my head in the sand on subworkflows. As they aren't offiical I would say lets try and break as much as possible to find all the cases in which things COULD break ;)

I would assume the subworkflows should have a module-version pinned though, no?

FriederikeHanssen:
> LGTM
> 
> @FriederikeHanssen I'm being the proverbial ostrich with my head in the sand on subworkflows. As they aren't offiical I would say lets try and break as much as possible to find all the cases in which things COULD break ;)
> 
> I would assume the subworkflows should have a module-version pinned though, no?

https://github.com/nf-core/modules/blob/master/subworkflows/nf-core/align_bowtie2/main.nf nope not as far as I know. No way to ping versions yet. I suppose that should go on the feature list of subworkflows

Lucpen:
Thanks for your help!  I will go on with the update then ^^

PR #1542, opened 2022-04-21
update picard/collecthsmetrics

matthdsm:
## PR checklist

This PR brings the output naming in line with CollectWgsMetrics and CollectMultipleMetrics

- [x] Add a resource `label`


modules/picard/collecthsmetrics/main.nf
```diff
@@ -41,7 +41,8 @@ process PICARD_COLLECTHSMETRICS {
         -BAIT_INTERVALS $bait_intervals \\
         -TARGET_INTERVALS $target_intervals \\
         -INPUT $bam \\
-        -OUTPUT ${prefix}_collecthsmetrics.txt
+        -OUTPUT ${prefix}.CollectHsMetrics.coverage_metrics
```
JoseEspinosa, line 44: According to the [conventions](https://nf-co.re/docs/contributing/modules#naming-conventions) I will suggest naming the output just as `${prefix}.hs_metrics.txt`, `${prefix}.metrics.txt` or `${prefix}.coverage.txt` if there isnt another reason to name it differently that I am not aware off, then nothing :smile: 
```suggestion
        -OUTPUT ${prefix}.metrics.txt
```

JoseEspinosa (commented):
Nice @matthdsm :+1: 
I would keep the output file naming a little bit simpler as I suggested in my comment

modules/picard/collecthsmetrics/main.nf
matthdsm: I agree with the suggestion, but the point of this PR was get the output naming scheme the same as collectWgsMetrics and collectMultipleMetrics, which also use the more complex version.

If this isn't desirable, I think we should also modify those 2 modules ( and sweep all other modules that output metrics for that matter)

matthdsm: So I would think of this PR as more an intermediate step, before overhauling the whole bunch

JoseEspinosa: Fair enough! Did not check the other modules :roll_eyes:
Then I will approve this one and as you suggested and at some point, it will be nice to modify all of them! :rocket:

matthdsm:
Thanks @JoseEspinosa!

JoseEspinosa (approved):

PR #1557, opened 2022-04-22
Replace vanilla Linux Biocontainer with Ubuntu

drpatelh:
x-ref: https://github.com/nf-core/rnaseq/issues/764

### DIFF
--- a/modules/untar/main.nf
+++ b/modules/untar/main.nf
@@ -2,10 +2,10 @@ process UNTAR {
     tag "$archive"
     label 'process_low'
 
-    conda (params.enable_conda ? "conda-forge::tar=1.34" : null)
+    conda (params.enable_conda ? "conda-forge::sed=4.7" : null)
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
-        'https://containers.biocontainers.pro/s3/SingImgsRepo/biocontainers/v1.2.0_cv2/biocontainers_v1.2.0_cv2.img' :
-        'biocontainers/biocontainers:v1.2.0_cv2' }"
+        'https://depot.galaxyproject.org/singularity/ubuntu:20.04' :
+        'ubuntu:20.04' }"
 
     input:
     tuple val(meta), path(archive)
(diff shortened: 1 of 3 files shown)


matthdsm:
Is there an option to use a base alpine container for these utilities? That would limit the network traffic needed to pull the images?

drpatelh:
We could do but like with all other modules it makes sense to use a container that has both Docker and Singularity images that we can pull directly, and this particular container does.

Emiller88:
> Is there an option to use a base alpine container for these utilities? That would limit the network traffic needed to pull the images?

You could override the container with your `nextflow.config`. I'd vote we let anyone who cares about that set it in their institutions' config.

Also, 22.04 got released yesterday https://discourse.ubuntu.com/t/jammy-jellyfish-release-notes/24668 just in case you weren't aware that was an option.

drpatelh:
22.04 doesn't have a Singularity container 😏 But no reason we can't update in the future. At this point, I think we just need access to basic Linux utilities so I think it's fine. As you mentioned, you can always override the configuration with your own container definitions.

matthdsm (approved):

modules/untar/main.nf
```diff
@@ -2,10 +2,10 @@ process UNTAR {
     tag "$archive"
     label 'process_low'
 
-    conda (params.enable_conda ? "conda-forge::tar=1.34" : null)
+    conda (params.enable_conda ? "conda-forge::sed=4.7" : null)
```
Midnighter, line 5: That's the only change that I don't get. Isn't it much more explicit to install tar in the untar module rather than sed which happens to include tar?

drpatelh: Good spot! Missed that in the copy/paste. Wondering whether we need to care about this specifically because these tools should all be installed in the base Conda env anyway right? So do we go for standardisation here where we always define `sed` for modules using the `ubuntu` image or do we define the specific tools we are using? The latter could also be a pain in general because in most instances you will want to use multiple utilities like `cat`, `awk`, `sed`  in the same module etc

Midnighter: Fair point. If there's a way to pull in multiple standard tools that are also available in the container images, then that's probably preferable. Would that make available `ps` btw? I think that's needed for resource tracing, right?

PR #1561, opened 2022-04-25
add samtools/bamtocram modules

maxulysse:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`


matthdsm:
Shouldn't this be part of samtools/view? Or is there a specific reason for this?

modules/samtools/bamtocram/main.nf
```diff
@@ -0,0 +1,35 @@
...
+
+    conda (params.enable_conda ? "bioconda::samtools=1.15.1" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/samtools:1.15.1--h1170115_0' :
+        'quay.io/biocontainers/samtools:1.15.1--h1170115_0' }"
+
+    input:
+    tuple val(meta), path(input), path(index)
+    path  fasta
+    path  fai
```
jfy133, line 14: Always wondered why these don't go together... but w/e

modules/samtools/bamtocram/meta.yml
```diff
@@ -0,0 +1,52 @@
...
+output:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. [ id:'test', single_end:false ]
+  - cram_crai:
+      type: file
+      description: filtered/converted CRAM file + index
+      pattern: "*{.cram,.crai}"
```
jfy133, line 45: I thought these should be listed as separate elements (even though they are output as a single tuple?)

modules/samtools/bamtocram/main.nf
```diff
@@ -0,0 +1,35 @@
...
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args  ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    samtools view --threads ${task.cpus-1} --reference ${fasta} -C $args $input > ${prefix}.cram
+    samtools index -@${task.cpus} ${prefix}.cram
```
jfy133, line 28: If the are running sequentuially, just give the firs tcommand the full CPUs?
```suggestion
    samtools view --threads ${task.cpus} --reference ${fasta} -C $args $input > ${prefix}.cram
    samtools index -@${task.cpus} ${prefix}.cram
```

jfy133 (approved):
Overall looks fine to me :+1: couple of random comments

modules/samtools/bamtocram/main.nf
maxulysse: Some tools needs just one, some the two, so more difficult to actually combine and give out, much easier to give out separately

modules/samtools/bamtocram/main.nf
FriederikeHanssen, line 28: From a sarek perspective I like having this here, but from a modules one, I am not so sure if it should be in one module. On the other hand we also have `collate|bam2fastq` in one module 🤔

jfy133: :eyes: I thought maybe in this case of indexing in the same module (After the big bust up a year ago ;) )? But I would also personally keep it separate

maxulysse: I think that if we use it in the sarek preprocessing, it might be used at some point by other pipelines, so I'd go for a nf-core module over a local one in this case

maxulysse:
> @matthdsm
> Shouldn't this be part of samtools/view? Or is there a specific reason for this?

Want actually to view+index the file in just one module instead of two.
Also output extension is different that input, so quite a specific usage, but still large enough I think for it to be here.

modules/samtools/bamtocram/meta.yml
maxulysse: do you want me to make a PR for it?

PR #1568, opened 2022-04-26
Add Trimgalore output for unpaired reads

drpatelh:
x-ref: https://github.com/nf-core/rnaseq/issues/799

### DIFF
--- a/modules/trimgalore/main.nf
+++ b/modules/trimgalore/main.nf
@@ -11,12 +11,13 @@ process TRIMGALORE {
     tuple val(meta), path(reads)
 
     output:
-    tuple val(meta), path("*.fq.gz")    , emit: reads
-    tuple val(meta), path("*report.txt"), emit: log
-    path "versions.yml"                 , emit: versions
+    tuple val(meta), path("${outprefix}"), emit: reads
+    tuple val(meta), path("*report.txt") , emit: log
+    path "versions.yml"                  , emit: versions
 
-    tuple val(meta), path("*.html"), emit: html optional true
-    tuple val(meta), path("*.zip") , emit: zip optional true
+    tuple val(meta), path("*unpaired*.fq.gz"), optional: true, emit: unpaired
+    tuple val(meta), path("*.html")          , optional: true, emit: html
+    tuple val(meta), path("*.zip")           , optional: true, emit: zip
 
     when:
     task.ext.when == null || task.ext.when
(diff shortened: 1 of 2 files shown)


modules/trimgalore/main.nf
```diff
@@ -11,12 +11,13 @@ process TRIMGALORE {
     tuple val(meta), path(reads)
 
     output:
-    tuple val(meta), path("*.fq.gz")    , emit: reads
-    tuple val(meta), path("*report.txt"), emit: log
-    path "versions.yml"                 , emit: versions
+    tuple val(meta), path("${outprefix}"), emit: reads
```
rpetit3, line 14: Thoughts on dropping `outprefix` for? Just curious

```
tuple val(meta), path("*{trimmed,val}*.fq.gz"), emit: reads
```

modules/trimgalore/main.nf
```diff
@@ -52,13 +54,15 @@ process TRIMGALORE {
...
             ${prefix}.fastq.gz
+
         cat <<-END_VERSIONS > versions.yml
         "${task.process}":
             trimgalore: \$(echo \$(trim_galore --version 2>&1) | sed 's/^.*version //; s/Last.*\$//')
             cutadapt: \$(cutadapt --version)
         END_VERSIONS
         """
     } else {
+        outprefix = "*val*.fq.gz"
```
rpetit3, line 65: Does `val` point to something, or does trimgalore output it as `val`? (again just curious to learn more about trim galor)

rpetit3: Sure does! 

```
--basename <PREFERRED_NAME>	Use PREFERRED_NAME as the basename for output files, instead of deriving the filenames from
                        the input files. Single-end data would be called PREFERRED_NAME_trimmed.fq(.gz), or
                        PREFERRED_NAME_val_1.fq(.gz) and PREFERRED_NAME_val_2.fq(.gz) for paired-end data. --basename
                        only works when 1 file (single-end) or 2 files (paired-end) are specified, but not for longer lists.
```

modules/trimgalore/main.nf
drpatelh: I thought `outprefix` was more explicit but this works too 👍🏽 
https://github.com/nf-core/modules/pull/1568/commits/e6cefebf1b3a0d284dec82d30820eb6caef348dc

modules/trimgalore/main.nf
drpatelh: TrimGalore automatically names paired-end output files using `*val_1.fq.gz` and `val_2.fq.gz`.

PR #1571, opened 2022-04-27
add intervals possibilities to gatk4/splitncigarreads

maxulysse:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`

### DIFF
--- a/modules/gatk4/splitncigarreads/main.nf
+++ b/modules/gatk4/splitncigarreads/main.nf
@@ -8,7 +8,7 @@ process GATK4_SPLITNCIGARREADS {
         'quay.io/biocontainers/gatk4:4.2.5.0--hdfd78af_0' }"
 
     input:
-    tuple val(meta), path(bam)
+    tuple val(meta), path(bam), path(bai), path(intervals)
     path  fasta
     path  fai
     path  dict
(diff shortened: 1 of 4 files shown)


modules/gatk4/splitncigarreads/main.nf
```diff
@@ -8,7 +8,7 @@ process GATK4_SPLITNCIGARREADS {
         'quay.io/biocontainers/gatk4:4.2.5.0--hdfd78af_0' }"
 
     input:
-    tuple val(meta), path(bam)
+    tuple val(meta), path(bam), path(bai), path(intervals)
```
JoseEspinosa, line 11: Would not be better to pass `intervals` as a separate input channel?

JoseEspinosa (changes requested):
LGTM! Just made a small suggestion

modules/gatk4/splitncigarreads/main.nf
maxulysse: Could be an idea, but so far, everytime we passed intervals it's in the input channel containing the main input files.
I think it could be interesting to try out `each` at some point with intervals, but that would mean changing a lot of logic in multiples pipelines

JoseEspinosa: I tend to separate inputs since it seems to me that it usually saves doing an operator to join back output from other modules but fair enough :rocket:

JoseEspinosa (approved):

PR #1572, opened 2022-04-27
Bump SAMtools version for custom/getchromsizes

drpatelh:
All SAMtools modules have been updated to 1.15 and this PR does the same for `custom/getchromsizes` so a single container is downloaded when running pipelines (if possible).


sateeshperi (approved):
lgtm

PR #1601, opened 2022-05-02
Update Krona modules to use ext.args

carpanz:
## PR checklist

- [ ] The following changes have been introduced: 
(a) the `ext.args` variable was instantiated but not used in the code, and has therefore been added to the code of both *kronadb* and *ktimporttaxonomy* modules
(b) the tests previously executed were not representative of the tool functionality: these have been modified to use the newly added test data, as well as the database actually imported by *kronadb*
- [ ] nothing new has been introduced in terms of functionality, output channel names or other breaking changes
- [ ] available test data have been used
- [ ] Add a resource `label`

### DIFF
--- a/modules/krona/kronadb/main.nf
+++ b/modules/krona/kronadb/main.nf
@@ -18,7 +18,9 @@ process KRONA_KRONADB {
     script:
     def args = task.ext.args ?: ''
     """
-    ktUpdateTaxonomy.sh taxonomy
+    ktUpdateTaxonomy.sh \\
+        $args \\
+        taxonomy
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
--- a/modules/krona/ktimporttaxonomy/main.nf
+++ b/modules/krona/ktimporttaxonomy/main.nf
@@ -23,7 +23,10 @@ process KRONA_KTIMPORTTAXONOMY {
     script:
     def args = task.ext.args ?: ''
     """
-    ktImportTaxonomy "$report" -tax taxonomy
+    ktImportTaxonomy \\
+       "$report" \\
+       $args \\
+       -tax taxonomy
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
--- a/tests/modules/krona/ktimporttaxonomy/main.nf
+++ b/tests/modules/krona/ktimporttaxonomy/main.nf
@@ -28,6 +28,6 @@ workflow test_krona_ktimporttaxonomy_report {
     ]
     taxonomy = file(params.test_data['generic']['txt']['hello'], checkIfExists: true)
 
-    TDOWNLOAD_DB()
+    DOWNLOAD_DB()
     TAXONOMY_REPORT ( input, DOWNLOAD_DB.out.db )
 }
--- a/tests/modules/krona/ktimporttaxonomy/main.nf
+++ b/tests/modules/krona/ktimporttaxonomy/main.nf
@@ -2,9 +2,11 @@
 
 nextflow.enable.dsl = 2
 
-include { KRONA_KTIMPORTTAXONOMY_READS }  from '../../../../modules/krona/ktimporttaxonomy/main.nf'
+include { KRONA_KTIMPORTTAXONOMY as TAXONOMY_READS  } from '../../../../modules/krona/ktimporttaxonomy/main.nf'
 
-include { KRONA_KTIMPORTTAXONOMY_REPORT } from '../../../../modules/krona/ktimporttaxonomy/main.nf'
+include { KRONA_KTIMPORTTAXONOMY as TAXONOMY_REPORT } from '../../../../modules/krona/ktimporttaxonomy/main.nf'
+
+include { KRONA_KRONADB          as DOWNLOAD_DB     } from '../../../../modules/krona/kronadb/main.nf'
 
 workflow test_krona_ktimporttaxonomy_reads {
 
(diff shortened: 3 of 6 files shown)


carpanz:
this PR is now dependent on https://github.com/nf-core/test-datasets/pull/521
i.e. once test database is merged in repository, I can change the config and update test code on this one

modules/krona/kronadb/main.nf
```diff
@@ -18,7 +18,9 @@ process KRONA_KRONADB {
     script:
     def args = task.ext.args ?: ''
     """
-    ktUpdateTaxonomy.sh taxonomy
+    ktUpdateTaxonomy.sh \\
+        $args \\
+        taxonomy
```
jfy133, line 23:
```suggestion
        taxonomy/
```

modules/krona/ktimporttaxonomy/main.nf
```diff
@@ -23,7 +23,10 @@ process KRONA_KTIMPORTTAXONOMY {
     script:
     def args = task.ext.args ?: ''
     """
-    ktImportTaxonomy "$report" -tax taxonomy
+    ktImportTaxonomy \\
+        "$report" \\
```
jfy133, line 27: I normaly rather put the input files at the end of teh command, but that's a minor quibble.

tests/modules/krona/ktimporttaxonomy/main.nf
```diff
@@ -2,15 +2,27 @@
 
 nextflow.enable.dsl = 2
 
-include { KRONA_KTIMPORTTAXONOMY } from '../../../../modules/krona/ktimporttaxonomy/main.nf'
+include { KRONA_KTIMPORTTAXONOMY as KRONA_TAXONOMY_READS  } from '../../../../modules/krona/ktimporttaxonomy/main.nf'
+include { KRONA_KTIMPORTTAXONOMY as KRONA_TAXONOMY_REPORT } from '../../../../modules/krona/ktimporttaxonomy/main.nf'
```
jfy133, line 6: I would rather imporat as:

`KRONA_KTIMPORTTAXONOMY_READS`
`KRONA_KTIMPORTTAXONOMY_REPORT`

I got very confused what I was looking at in the tests...

jfy133 (changes requested):
Good improvements, but some inconsistencies that should be resolved

General comment: `ktonadb` module should be called `KRONA_KTUPDATETAXONOMY`... I'm not sure how `KRONADB` got through here originally... could you create a **NEW** module and you can copy the code over (deleting will cause downstream problems)

tests/modules/krona/ktimporttaxonomy/main.nf
```diff
@@ -2,15 +2,27 @@
...
-workflow test_krona_ktimporttaxonomy {
+workflow test_krona_ktimporttaxonomy_reads {
 
     input = [
         [ id:'test', single_end:false ], // meta map
-        file(params.test_data['generic']['txt']['hello'], checkIfExists: true)
+        file(params.test_data['sarscov2']['metagenome']['classified_reads_assignment'], checkIfExists: true)
     ]
-    taxonomy = file(params.test_data['generic']['txt']['hello'], checkIfExists: true)
+    taxonomy = file("https://raw.githubusercontent.com/lescai/test-datasets/modules/data/genomics/sarscov2/metagenome/krona_taxonomy.tab")
```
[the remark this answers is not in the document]
carpanz: we will follow the config scheme here and use
`file(params.test_data['sarscov2']['metagenome']['krona_taxonomy'], checkIfExists: true)`
changing the config too

tests/modules/krona/ktimporttaxonomy/main.nf
carpanz: agreed

modules/krona/kronadb/main.nf
carpanz: done together with other commits

lescai (approved):
With the last change of ext.args LGTM!

modules/krona/ktupdatetaxonomy/meta.yml
```diff
@@ -0,0 +1,30 @@
+name: krona_ktupdatetaxonomy
+description: KronaTools Update Taxonomy downloads a taxonomy database
+keywords:
+  - database
+  - taxonomy
+  - krona
```
jfy133, line 6:
```suggestion
  - krona
  -visualisaiton
```

jfy133 (changes requested):
You need to re-include the kronadb tests! We need to keep them as we are retaining the kronadb module even if it's beeing supplanted by the correct ktUpdateTaxonomy module!

Please also check the output declaration of the KtUpdatetaxonomy module... this should be a directory right? Not a the TAB file?

At lesa the help message implies this:

```
   [/custom/dir]  Taxonomy will be built in this directory instead of the
                  directory specified during installation. This custom
                  directory can be referred to with -tax in import scripts.
```

lescai:
Hi @jfy133 thanks for raising these additional points.
I'll chip in on this because it's connected to the test data I've recently submitted.
As we discussed, we thought best not to question how these modules got in in the first place, and therefore we tried to propose non-breaking changes.
The existing `kronadb` module outputs the taxonomy this way: `path 'taxonomy/taxonomy.tab', emit: db`
And the existing `ktImportTaxonomy` receives this channel this way: `path  "taxonomy/taxonomy.tab"`
In order not to break the code we kept it as it is, which perhaps is also convenient for the test data, since we can just upload the .tab file and pass it on.
The command *ktImportTaxonomy* expects to receive a folder path, and to find inside it a `taxonomy.tab` file with the taxonomy: naturally I don't have any preference either way.
Maybe we could just clarify this in the help message / yml file and keep things as they were initially proposed, so we don't break anyone's code?

modules/krona/ktupdatetaxonomy/meta.yml
carpanz: agreed!

jfy133:
> 

OK that is a fair point! Thanks for clarifyihng. Indeed - then lets keep the compatibility.

Then I tihnk the only thing missing was the deleted `kronadb` test files, and otherwise we are GTG

jfy133 (approved):
Thanks very much @carpanz !

PR #1604, opened 2022-05-02
feat: add module for slimfastq

Midnighter:
I know that nf-core generally works with gzipped FASTQ files but I'm going to use slimfastq for its lossless high compression capabilities. Maybe it helps someone else 🤷🏼 

## PR checklist

- [x] Add a resource `label`

### DIFF
--- a/modules/slimfastq/main.nf
+++ b/modules/slimfastq/main.nf
@@ -1,0 +1,52 @@
+def VERSION = '2.04'
+
+process SLIMFASTQ {
+    tag "$meta.id"
+    label 'process_low'
+
+    conda (params.enable_conda ? "bioconda::slimfastq=2.04" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/slimfastq:2.04--h87f3376_2':
+        'quay.io/biocontainers/slimfastq:2.04--h87f3376_2' }"
+
+    input:
+    tuple val(meta), path(fastq)
+
+    output:
+    tuple val(meta), path("*.sfq"), emit: sfq
+    path "versions.yml"           , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    if (meta.single_end) {
+        """
+        gzip -d -c '${fastq}' | slimfastq \\
+            $args \\
+            -f '${prefix}.sfq'
+
+        cat <<-END_VERSIONS > versions.yml
+        "${task.process}":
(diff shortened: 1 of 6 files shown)


tests/modules/slimfastq/test.yml
```diff
@@ -0,0 +1,41 @@
+- name: slimfastq test_slimfastq_single_end
+  command: nextflow run tests/modules/slimfastq -entry test_slimfastq_single_end -c tests/config/nextflow.config
+  tags:
+    - slimfastq
+  files:
+    - path: output/slimfastq/test.sfq
+      md5sum: 6a942eeca6c99ee9a9a0cedab5d246f1
+    - path: output/slimfastq/versions.yml
+      md5sum: f52351f5c9e6259af02745c8eae5c780
```
matthdsm, line 9: Probably best to omit the `versions` md5sum?

modules/slimfastq/main.nf
```diff
@@ -0,0 +1,52 @@
...
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    if (meta.single_end) {
+        """
+        gzip -d -c '${fastq}' | slimfastq \\
+            $args \\
+            -f '${prefix}.sfq'
```
matthdsm, line 29: Wouldn't it be "prettier" to chop of the `fastq.gz` extension and use the `SimpleName` from the original file?
I especially see an issue with the paired data  and the `_1` suffix. For example, at CMGG all data is standardized with an `_R1/_R2` suffix according to Illumina standard. This would break the naming convention for us (and others I imagine)

tests/modules/slimfastq/test.yml
Midnighter: Sure, can do that.

modules/slimfastq/main.nf
Midnighter: `simpleName` is not great, in my opinion, because if your filename contains dots, it will happily chop away all of them. More importantly, I thought it was nf-core convention to use the prefix in output names so, yes, I can strip away extensions but I should rely on prefix as far as I know.

matthdsm: By `SimpleName` I meant stripping of the the `{.fq,.fastq,.fq.gz,.fastq.gz}` extensions, possibly by regex or whatever.
I'm aware of the convention, but I still think it's dangerous and might cause issues down the line for users.

Anyway, I'm not very opinionated atm, since I'm not planning to use it, but I like future proofing.

matthdsm (approved):

Midnighter:
Hmm, I was still working on making those changes you mentioned.

matthdsm:
Wasn't me! I just hit the approve button

Midnighter:
Well, guess it will be part of a future update then 😛

PR #1615, opened 2022-05-05
shasta module added

### LINKED ISSUE #1185: new module: shasta assembler
I think it would be good to have a module for [shasta](https://anaconda.org/bioconda/shasta)

fmalmeida:
## PR checklist

Adding Shasta assembler module

Closes #1185 

- [x] Add a resource `label`
  - [x] `PROFILE=docker pytest --tag shasta --symlink --keep-workflow-wd --git-aware`
  - [ ] `PROFILE=singularity pytest --tag shasta --symlink --keep-workflow-wd --git-aware`
  - [ ] `PROFILE=conda pytest --tag shasta --symlink --keep-workflow-wd --git-aware`


modules/shasta/main.nf
```diff
@@ -0,0 +1,43 @@
...
+
+    # run shasta
+    shasta \\
+        --input reads.fq \\
+        --config $model \\
+        $args \\
+        --threads $task.cpus
+
+    # compress results
+    gzip -c ShastaRun/Assembly.fasta > ShastaRun/Assembly.fasta.gz
```
mahesh-panchal, line 35: It would be helpful to name the assemblies using the prefix

modules/shasta/main.nf
```diff
@@ -0,0 +1,43 @@
...
+    tuple val(meta), path("ShastaRun/")                 , emit: results
+    path "versions.yml"                                 , emit: versions
+
+    script:
+    def args   = task.ext.args   ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def model  = "${meta.model}" ?: 'Nanopore-Oct2021'
+    """
+    # shasta requires uncompressed
+    zcat -f $reads > reads.fq
```
mahesh-panchal, line 25: Please clean up temporary files like this at the end of the process.

tests/modules/shasta/test.yml
```diff
@@ -0,0 +1,7 @@
+- name: shasta test_shasta
+  command: nextflow run ./tests/modules/shasta -entry test_shasta -c ./tests/config/nextflow.config -c ./tests/modules/shasta/nextflow.config
+  tags:
+    - shasta
+  files:
```
mahesh-panchal, line 5: You need to run:
`nf-core modules create-test-yml shasta`

fmalmeida:
Thanks for the review @mahesh-panchal. I'll work on them this weekend 😁

tests/modules/shasta/test.yml
fmalmeida: When a run this one, the module is killed due to lack of memory. It seems that the create-test-yml execution allows less than 6.GB.

What can I add or modify to increase it for this module?

mahesh-panchal: There are files you can use in the `tests/config/test.config` files as test datasets. You should use a small file just to ensure you get output. Doesn't matter if the output makes sense. Just that it runs and makes the output files.

fmalmeida:
Now it has some errors on pytest that I am not familiar.

Maybe they appear because the output with this dataset is empty? Or this shouldn’t be a problem?

modules/shasta/main.nf
```diff
@@ -0,0 +1,46 @@
...
+    # run shasta
+    shasta \\
+        --input reads.fq \\
+        --config $model \\
+        $args \\
+        --threads $task.cpus
+
+    # compress results
+    gzip -c ShastaRun/Assembly.fasta > ShastaRun/${meta.id}_Assembly.fasta.gz
+    gzip -c ShastaRun/Assembly.gfa   > ShastaRun/${meta.id}_Assembly.gfa.gz
```
mahesh-panchal, line 36:
```suggestion
    gzip -c ShastaRun/Assembly.fasta > ${prefix}_Assembly.fasta.gz
    gzip -c ShastaRun/Assembly.gfa   > ${prefix}_Assembly.gfa.gz
```
You should use the configurable `$prefix` for naming output so someone can rename it via a config file. 
For publishing reasons, it would be better to create the compressed files outside the ShastaRun directory. Then you don't get issues with file writing conflicts if the publishing directory properties are changed.

mahesh-panchal:
> Now it has some errors on pytest that I am not familiar.
> 
> Maybe they appear because the output with this dataset is empty? Or this shouldn’t be a problem?

I think we should ask on #modules about the empty files. Typically, empty files are not permitted for tests, but I feel like a valid test dataset may end up being too large. We should ask if we can just remove the checksum for these empty files, unless you're able to find an intermediate test dataset that's small enough to host in the test-datasets repo but large enough to make a valid shasta output.

mahesh-panchal:
It seems I already asked this question: https://nfcore.slack.com/archives/CJRH30T6V/p1652695030983019
As long as it's not too many empty files, just remove the md5sums, and check it works correctly with a larger data set ( which you started with, so you have )

fmalmeida:
Changed dataset and added a parameter in the tests nextflow.config for fasta that avoids filtering test dataset too much, and now it actually produces outputs with contents that do not flag pytest warnings:

https://github.com/nf-core/modules/blob/5fa0cff083ed7c186f66f0beb0f2dad3b436708d/tests/modules/shasta/nextflow.config#L1-L7

https://github.com/nf-core/modules/blob/5fa0cff083ed7c186f66f0beb0f2dad3b436708d/tests/modules/shasta/main.nf#L1-L15

Now everything seems to be working and resolving previous comments, right?

modules/shasta/meta.yml
```diff
@@ -0,0 +1,52 @@
...
+        Groovy Map containing sample information
+        e.g. [ id:'test', single_end:false ]
+  - versions:
+      type: file
+      description: File containing software versions
+      pattern: "versions.yml"
+  - assembly:
+      type: file
+      description: Assembled FASTA file
+      pattern: "ShastaRun/Assembly.fasta.gz"
```
SPPearce, line 41: These patterns now have the prefix.

tests/modules/shasta/test.yml
```diff
@@ -0,0 +1,45 @@
+- name: shasta test_shasta
+  command: nextflow run ./tests/modules/shasta -entry test_shasta -c ./tests/config/nextflow.config  -c ./tests/modules/shasta/nextflow.config
+  tags:
+    - shasta
+  files:
+    - path: output/shasta/ShastaRun/Assembly-BothStrands-NoSequence.gfa
+      contains: "[ # TODO nf-core: file md5sum was variable, please replace this text with a string found in the file instead ]"
```
SPPearce, line 7: These need dealing with, it shows that the test generated different output files when it was run twice.
This might just be that the file contains a timestamp, non-deterministic ordering of results, or rounding errors, or it might mean there is a random seed that needs to be set in the program (at which point a default value for the random seed that can be overridden by a config would be the best option).
If it is just a timestamp or similar, then you can just check for a string as the TODO states. Or drop the test check altogether as you have a lot of files being emitted here.

SPPearce (changes requested):
Please address the TODOs in the test.yml file.

SPPearce:
@fmalmeida , we are going to be making a change to the modules syntax soon, which will involve updating all the modules. Can we try and get this merged before the weekend, so it gets included in that update automatically?

oh, and let's try @nf-core-bot fix linting

tests/modules/shasta/test.yml
fmalmeida: I've checked the flagged files with the TO-DO statement, and s not timestamps or such. What actually varies is the assembly itself between runs of the software.

I saw that the Assembly.fasta is there, correct, as expected, is an assembled genome. But the md5sum checks are different because some nucleotides may have changed. And, this, causes all the others to also change as all of the other ones flagged are relative to the Assembly itself.

So, I guess the most relevant for those files would be to check whether they exist only.

modules/shasta/meta.yml
fmalmeida: changed

fmalmeida:
@nf-core-bot fix linting

fmalmeida:
> @fmalmeida , we are going to be making a change to the modules syntax soon, which will involve updating all the modules. Can we try and get this merged before the weekend, so it gets included in that update automatically?
> 
> oh, and let's try @nf-core-bot fix linting

Hi @SPPearce 

I've tried to address the comments.

tests/modules/shasta/test.yml
SPPearce: Is there an option to set the random seed on the tool via the command line?

fmalmeida: Not that I am aware of. These are all the command line options: https://chanzuckerberg.github.io/shasta/CommandLineOptions.html

And all of them seem to be only defaults that keep the same across executions ... I did not see something like a random seed. :/

SPPearce: Well that is both awkward and strange that it isn't included. So running the tool twice gives different actual output because of the randomness? The code of the tool itself seems to set the seed to 231 in [several places]:(https://github.com/chanzuckerberg/shasta/search?l=C%2B%2B&p=1&q=seed), but I don't understand C++ to actually follow the structure of the code.

SPPearce: Can you ask on the tool github?

fmalmeida: Let’s wait:

https://github.com/chanzuckerberg/shasta/issues/296#issue-1390697538

SPPearce: Thanks, we'll see what they say.

SPPearce: Hmm, ok. Can you make the tests pass. You can check that files contain a string of text if that is possible as the TODO mentions, or you can just require the existence of the file (can't remember the syntax for that off the top of my head).

fmalmeida: That’s how it is at the moment, you can see that the tests are passing because I changed to only check for existence

modules/shasta/meta.yml
```diff
@@ -0,0 +1,52 @@
+name: shasta
+description: The goal of the Shasta long read assembler is to rapidly produce accurate assembled sequence using DNA reads generated by Oxford Nanopore flow cells as input.
+keywords:
+  - nanopore
+  - de-novo
+  - assembly
+  - longread
+tools:
+  - shasta:
+      description: Rapidly produce accurate assembled sequence using as input DNA reads generated by Oxford Nanopore flow cells.
```
SPPearce, line 10: Can we add a note to say that it isn't deterministic and then we're done.

fmalmeida: added

SPPearce:
@nf-core-bot fix linting

PR #1627, opened 2022-05-06
Use a template to add retrying to prefetch

Midnighter:
In the fetchngs pipeline, we are currently using a local prefetch module and a bash script in `bin`. With this change, we can have both installed from modules.

## PR checklist


Midnighter:
@drpatelh I suspect the linting error and warning rather have something to do with the linting tool than the actual module.

drpatelh (approved):
Happy with this if you are! I assume you tested right?

Midnighter:
> The linting error may be because we are using `shell` instead of `script` but would need to test.

Yeah, I figured it would be something like that. And the warning related to versions.yml is probably because it's in a template.

> Happy with this if you are! I assume you tested right?

I ran the existing tests, do you mean something more than that?

PR #1631, opened 2022-05-08
Refactor prefetch

Midnighter:
Change prefetch such that it can accept an NCBI settings file as input. Test both variants properly.

## PR checklist


Midnighter:
Usual linting problems with `shell` block and template.

drpatelh:
> Usual linting problems with shell block and template.

Can you create an issue on tools for this please? We should fix this.

modules/sratools/prefetch/main.nf
```diff
@@ -9,18 +9,17 @@ process SRATOOLS_PREFETCH {
 
     input:
     tuple val(meta), val(id)
+    path(ncbi_settings)
```
drpatelh, line 12:
```suggestion
    path ncbi_settings
```

tests/modules/sratools/prefetch/nextflow.config
```diff
@@ -1,3 +1,6 @@
+params.settings_path = '/tmp/.ncbi'
+params.settings_file = "${params.settings_path}/user-settings.mkfg"
```
drpatelh, line 2: This is getting increasingly complex to follow 😏 Would it be easier to provide `ncbi_settings` as input and if it is empty e.g. `[]` then to create one. If it isn't then to use that?

Midnighter: To me the other configuration is the more complex one. I'm testing two branches of logic,

1. The settings are supplied as a file.
2. The settings are available on the filesystem (or mounted into the container), in a location that can be discovered by the sra-tools.

The test cases use these parameters. So to my mind it makes more sense to set the path in a single location rather than hard code it everywhere.

drpatelh: We can add the file directly as a test-dataset and update [this](https://github.com/nf-core/modules/blob/master/tests/config/test_data.config) file so it truly is set and available from a single location (even for multiple modules)?

Midnighter: Oh yeah, good call. Still need to download it to a sensible location on the local filesystem, though, for testing one branch of the logic. So I don't think I can entirely avoid setting parameters.

tests/modules/sratools/prefetch/nextflow_mount.config
```diff
@@ -0,0 +1,17 @@
...
+
+process {
+
+    publishDir = { "${params.outdir}/${task.process.tokenize(':')[-1].tokenize('_')[0].toLowerCase()}" }
+
+    withName: SRATOOLS_PREFETCH {
+        containerOptions = {
+            (workflow.containerEngine == 'singularity') ?
+            "-B ${params.settings_path}:${params.settings_path}" :
+            "-v ${params.settings_path}:${params.settings_path}"
```
drpatelh, line 14: This makes me a bit nervous 😬

drpatelh: Why would they make it easier to use their tool?!

drpatelh (approved):
Beautiful! 🚀

PR #1680, opened 2022-05-20
Modified cnvkit/batch module: cram2bam conversion

SusiJo:
## PR checklist

This PR improves the runtime of `cnvkit/batch` when `cram` files are used as input. It relies on a new biocontainer that has `samtools` and `cnvkit` in order to convert `cram` to `bam` before running `cnvkit`. I've also added a new test for `cram` input in `tumor_only mode`.

- [ ] Add a resource `label`

### DIFF
--- a/modules/cnvkit/batch/main.nf
+++ b/modules/cnvkit/batch/main.nf
@@ -27,17 +27,17 @@ process CNVKIT_BATCH {
 
     script:
     def args = task.ext.args ?: ''
+
     // execute samtools only when cram files are input, cnvkit runs natively on bam but is prohibitively slow
     // input pair is assumed to have same extension if both exist
     def is_cram = tumor.Extension == "cram" ? true : false
     def tumor_out = is_cram ? tumor.BaseName + ".bam" : "${tumor}"
+
     // do not run samtools on normal samples in tumor_only mode
     def normal_exists = normal ? true: false
     // tumor_only mode does not need fasta & target
     // instead it requires a pre-computed reference.cnn which is built from fasta & target
     def (normal_out, normal_args, fasta_args) = ["", "", ""]
-    def target_args = targets ? "--targets $targets" : ""
-    def reference_args = reference ? "--reference $reference" : ""
 
     if (normal_exists){
         def normal_prefix = normal.BaseName
@@ -45,6 +45,9 @@ process CNVKIT_BATCH {
         normal_args = normal_prefix ? "--normal $normal_out" : ""
         fasta_args = fasta ? "--fasta $fasta" : ""
     }
+
+    def target_args = targets ? "--targets $targets" : ""
+    def reference_args = reference ? "--reference $reference" : ""
 
     """
     if $is_cram; then
--- a/modules/cnvkit/batch/main.nf
+++ b/modules/cnvkit/batch/main.nf
@@ -27,7 +27,7 @@ process CNVKIT_BATCH {
 
     script:
     def args = task.ext.args ?: ''
-    // execute samtools only when cram files are input, cnvkit runs natively on bam but is prohibitively slow 
+    // execute samtools only when cram files are input, cnvkit runs natively on bam but is prohibitively slow
     // input pair is assumed to have same extension if both exist
     def is_cram = tumor.Extension == "cram" ? true : false
     def tumor_out = is_cram ? tumor.BaseName + ".bam" : "${tumor}"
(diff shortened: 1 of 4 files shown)


SusiJo:
The `nf-core linting` is expected to fail, due to empty files. That's also the case for the original module.

SusiJo:
The updated module allows to output scatter and diagram plots when specifying `--scatter --diagram` in `ext.args`.

SusiJo:
Also, the `workflow test_cnvkit_cram` now actually uses `cram` as input instead of `bam`.

modules/cnvkit/batch/main.nf
```diff
@@ -18,28 +18,51 @@ process CNVKIT_BATCH {
...
+    // instead it requires a pre-computed reference which is built from fasta & target
+    def (normal_out, normal_args, fasta_args, target_args) = ["", "", "", ""]
     def reference_args = reference ? "--reference $reference" : ""
 
-    def target_args = ""
+    if (normal_exists){
+        def normal_prefix = normal.BaseName
+        normal_out = is_cram ? "${normal_prefix}" + ".bam" : "${normal}"
+        normal_args = normal_prefix ? "--normal $normal_out" : ""
+        fasta_args = fasta ? "--fasta $fasta" : ""
```
FriederikeHanssen, line 45: Is the fasta only needed for normal?

SusiJo: The `fasta_args` are indeed only needed when normal samples are provided. In the tumor_only mode the `reference.cnn` will be provided and then options to construct a new reference (-f/--fasta) should not be used.

modules/cnvkit/batch/main.nf
```diff
@@ -18,28 +18,51 @@ process CNVKIT_BATCH {
...
     path "versions.yml"           , emit: versions
 
     when:
     task.ext.when == null || task.ext.when
 
     script:
     def args = task.ext.args ?: ''
-    def normal_args = normal ? "--normal $normal" : ""
-    def fasta_args = fasta ? "--fasta $fasta" : ""
+    // execute samtools only when cram files are input
```
FriederikeHanssen, line 30:
```suggestion
    // execute samtools only when cram files are input, cnvkit runs natively on bam but is prohibitively slow 
```

modules/cnvkit/batch/main.nf
```diff
@@ -18,28 +18,51 @@ process CNVKIT_BATCH {
...
+    // input pair is assumed to have same extension if both exist
+    def is_cram = tumor.Extension == "cram" ? true : false
+    def tumor_out = is_cram ? tumor.BaseName + ".bam" : "${tumor}"
+    // do not run samtools on normal samples in tumor_only
```
FriederikeHanssen, line 34:
```suggestion
    // do not run samtools on normal samples in tumor_only mode
```

modules/cnvkit/batch/main.nf
```diff
@@ -18,28 +18,51 @@ process CNVKIT_BATCH {
...
+    def normal_exists = normal ? true: false
+    // tumor_only mode does not need fasta & target
+    // instead it requires a pre-computed reference which is built from fasta & target
```
FriederikeHanssen, line 37:
```suggestion
    // instead it requires a pre-computed reference.cnn which is built from fasta & target
```

modules/cnvkit/batch/main.nf
FriederikeHanssen, line 30: Suggestion for this section: For readability, add some blank lines after groups of parameters that belong together

modules/cnvkit/batch/main.nf
```diff
@@ -18,28 +18,51 @@ process CNVKIT_BATCH {
...
     def reference_args = reference ? "--reference $reference" : ""
 
-    def target_args = ""
+    if (normal_exists){
+        def normal_prefix = normal.BaseName
+        normal_out = is_cram ? "${normal_prefix}" + ".bam" : "${normal}"
+        normal_args = normal_prefix ? "--normal $normal_out" : ""
+        fasta_args = fasta ? "--fasta $fasta" : ""
+    }
+
```
FriederikeHanssen, line 47: I think the code below 
```
    if (args.contains("--method wgs") || args.contains("-m wgs")) {
        target_args = targets ? "--targets $targets" : ""
    }
    else {
        target_args = "--targets $targets"
    }
```
can be set in the modules.config and it could be reduced here to:
`def target = targets ? "--targets $targets" : ""`

FriederikeHanssen (commented):
🚀 nice work! I know that some of the things I mentioned were there before. Could you possibly still tidy them up? Otherwise I can also push to the branch after lunch

modules/cnvkit/batch/main.nf
SusiJo: Sure, i'll remove the `params` section and i'll add a description.

modules/cnvkit/batch/main.nf
SusiJo: Great, thanks! I was also wondering why that section is included at all and why it's not solely in the `ext.args`.

FriederikeHanssen (approved):
nice work

PR #1693, opened 2022-05-24
mosdepth: cram compatiblility

matthdsm:
- Add fasta input for cram compatibility
- fix meta.yaml
- add new tests
-   <b>dropped</b>  `window` input, this can be set through `ext.args` 

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`

### DIFF
--- a/modules/mosdepth/main.nf
+++ b/modules/mosdepth/main.nf
@@ -10,13 +10,13 @@ process MOSDEPTH {
     input:
     tuple val(meta), path(bam), path(bai)
     path  bed
-    val   window_size
+    path  fasta
 
     output:
     tuple val(meta), path('*.global.dist.txt')    , emit: global_txt
     tuple val(meta), path('*.region.dist.txt')    , emit: regions_txt , optional:true
     tuple val(meta), path('*.summary.txt')        , emit: summary_txt
-    tuple val(meta), path('*.per-base.d4')        , emit: d4          , optional:true
+    tuple val(meta), path('*.per-base.d4')        , emit: per_base_d4 , optional:true
     tuple val(meta), path('*.per-base.bed.gz')    , emit: per_base_bed, optional:true
     tuple val(meta), path('*.per-base.bed.gz.csi'), emit: per_base_csi, optional:true
     tuple val(meta), path('*.regions.bed.gz')     , emit: regions_bed , optional:true
@@ -29,36 +29,17 @@ process MOSDEPTH {
     script:
     def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
-    if (window_size) {
-        interval = "--by ${window_size}"
-    } else if ( bed ) {
-        interval = "--by ${bed}"
-    } else {
-        interval = ""
-    }
+    def reference = fasta ? "--fasta ${fasta}" : ""
+    def interval = bed ? "--by ${bed}" : ""
+
     """
     mosdepth \\
+        --threads ${task.cpus} \\
         $interval \\
+        $reference \\
         $args \\
         $prefix \\
         $bam
-    cat <<-END_VERSIONS > versions.yml
-    "${task.process}":
-        mosdepth: \$(mosdepth --version 2>&1 | sed 's/^.*mosdepth //; s/ .*\$//')
-    END_VERSIONS
-    """
-
-    stub:
-    def prefix = task.ext.prefix ?: "${meta.id}"
-    """
-    touch ${prefix}.global.dist.txt
-    touch ${prefix}.region.dist.txt
-    touch ${prefix}.summary.txt
-    touch ${prefix}.per-base.d4
-    touch ${prefix}.per-base.bed.gz
-    touch ${prefix}.per-base.bed.gz.csi
-    touch ${prefix}.regions.bed.gz
-    touch ${prefix}.regions.bed.gz.csi
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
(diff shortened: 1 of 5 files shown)


modules/mosdepth/main.nf
```diff
@@ -29,36 +29,17 @@ process MOSDEPTH {
...
+
     """
     mosdepth \\
+        --threads ${task.cpus} \\
         $interval \\
+        $reference \\
         $args \\
         $prefix \\
         $bam
-    cat <<-END_VERSIONS > versions.yml
```
JoseEspinosa, line 45: What is the reason to remove the stub section?

JoseEspinosa (changes requested):
Nice @matthdsm ! Made a couple of suggestions/questions?

modules/mosdepth/main.nf
matthdsm: There was no reason to keep it, since it's not used in the tests. I wanted to clear some clutter, nothing more

modules/mosdepth/main.nf
```diff
@@ -10,13 +10,13 @@ process MOSDEPTH {
     input:
     tuple val(meta), path(bam), path(bai)
     path  bed
-    val   window_size
```
drpatelh, line 13: Might be a good idea to keep this in. I added it in because according to the docs you can only provide one or the other:

```
-b --by <bed|window>       optional BED file or (integer) window-sizes.
```

The `if/else` logic that has been deleted allows us to sanity check this and use one over the other. Have you tried running the module when providing both because if we assume that `--window` will be provided via `ext.args` then it could break the module if both are provided.

Another solution is to check `ext.args` to see if `--by` has been provided and use one over the other.

modules/mosdepth/main.nf
```diff
@@ -29,36 +29,17 @@ process MOSDEPTH {
...
-    def prefix = task.ext.prefix ?: "${meta.id}"
-    """
-    touch ${prefix}.global.dist.txt
-    touch ${prefix}.region.dist.txt
-    touch ${prefix}.summary.txt
-    touch ${prefix}.per-base.d4
-    touch ${prefix}.per-base.bed.gz
-    touch ${prefix}.per-base.bed.gz.csi
-    touch ${prefix}.regions.bed.gz
-    touch ${prefix}.regions.bed.gz.csi
```
drpatelh, line 43: Need to add the `stub` test back in.

matthdsm: done

modules/mosdepth/main.nf
matthdsm: Ok, but which one gets precedence then?

matthdsm:
@JoseEspinosa , @drpatelh, I added an error when `--by` is defined twice. I also added a test for this case.

modules/mosdepth/main.nf
JoseEspinosa: I see, I don't really know what is the plan here. I thought it could be nice that all modules have a stub section so it will be possible to perform a kind of `dry-run` of any pipeline. But as I said I don't know if this is something I only aiming for or whether this would be useful for others.

matthdsm: I wasn't even aware there were plans for `dry-run` support. If so, it does make sense to have the stub.
A `dry-run` would enable us to quickly prototype workflows

modules/mosdepth/main.nf
```diff
@@ -29,19 +29,21 @@ process MOSDEPTH {
...
-    if (window_size) {
-        interval = "--by ${window_size}"
-    } else if ( bed ) {
-        interval = "--by ${bed}"
-    } else {
-        interval = ""
+    def reference = fasta ? "--fasta ${fasta}" : ""
+    def interval = bed ? "--by ${bed}" : ""
+    if (bed && args.contains("--by")) {
+        exit 1, "--by can only be specified once."
```
drpatelh, line 35:
```suggestion
        exit 1, "'--by' can only be specified once when running mosdepth! Either remove input BED file definition or remove '--by' from 'ext.args' definition"
```

modules/mosdepth/main.nf
drpatelh: The plan is to eventually add a `stub` for all modules but we need someone to do it! It's kinda in limbo because it is only useful if it is added to all modules rather than a subset in the context of a pipeline. Quite a few should have `stub` already.

matthdsm: That's good to know. Perhaps this notion should be added to the guidelines (and enforced by reviewers).

tests/modules/mosdepth/main.nf
```diff
@@ -2,35 +2,60 @@
...
-    MOSDEPTH ( input, bed, [] )
+workflow test_mosdepth_fail {
+    input  = [
+                [ id:'test', single_end:true ],
+                [ file(params.test_data['homo_sapiens']['illumina']['test_paired_end_sorted_bam'], checkIfExists: true) ],
+                [ file(params.test_data['homo_sapiens']['illumina']['test_paired_end_sorted_bam_bai'], checkIfExists: true) ]
+            ]
+    bed  = [ file(params.test_data['homo_sapiens']['genome']['genome_bed'], checkIfExists: true) ]
+
+    MOSDEPTH_FAIL ( input, bed, [] )
```
drpatelh, line 60: Might make sense to add a test for specifying window size and the tests passing too.

matthdsm: 👍🏻

matthdsm:
All requested tests were added!

drpatelh:
Awesome! Thank you 🚀

JoseEspinosa (commented):
Nice done @matthdsm 😄

PR #1703, opened 2022-05-27
Bump container version for STAR to 2.7.10a

drpatelh:
Bumping here to be able to use default nf-core/module in rnaseq and a custom one for iGenomes. See https://github.com/nf-core/rnaseq/issues/834

### DIFF
--- a/modules/star/align/main.nf
+++ b/modules/star/align/main.nf
@@ -2,16 +2,15 @@ process STAR_ALIGN {
     tag "$meta.id"
     label 'process_high'
 
-    // Note: 2.7X indices incompatible with AWS iGenomes.
-    conda (params.enable_conda ? 'bioconda::star=2.7.9a' : null)
+    conda (params.enable_conda ? "bioconda::star=2.7.10a bioconda::samtools=1.15.1 conda-forge::gawk=5.1.0" : null)
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
-        'https://depot.galaxyproject.org/singularity/star:2.7.9a--h9ee0642_0' :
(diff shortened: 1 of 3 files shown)


modules/star/align/main.nf
```diff
@@ -67,6 +66,8 @@ process STAR_ALIGN {
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
         star: \$(STAR --version | sed -e "s/STAR_//g")
+        samtools: \$(echo \$(samtools --version 2>&1) | sed 's/^.*samtools //; s/Using.*\$//')
```
JoseEspinosa, line 69: `samtools` is in the container but not used in the script section, should we report its version? won't it be misleading?`

JoseEspinosa (approved):
LGTM! just make a small question about the version reporting

modules/star/align/main.nf
drpatelh: Well spotted. You are right. I thought it made sense to use the same container as `STAR_GENOMEGENERATE` to save having to download an additional container just for STAR. Don't think there is any harm in reporting the version of SAMtools if it's in the container?

JoseEspinosa:
Seems that md5sums should be updated, though :cry:

modules/star/align/main.nf
JoseEspinosa: My concern is that if a user checks the versions reported by the module in the multiqc report, they would expect that `samtools` has been used in the process and it will be hard to spot that the reason is just that we are using the same container (this I agree that makes sense). Maybe I am over paranoid :laughing:

drpatelh: Yep, but if we use a mulled container here and say someone updates just SAMtools we won't know if we don't report it explicitly.

PR #1704, opened 2022-05-27
Adding QC subworkflows

lescai:
This PR adds Subworkflows dedicated to reads QC and BAM files QC.
Tests have been run using Singularity, and existing test data and are passing.
nf-core linting will be additionally run when pipeline is assembled.
command line will be shared privately, since it contains private folders.


lescai:
apologies :)

PR #1710, opened 2022-05-31
Fix filtlong - makes running without shortreads work correctly, and also ensures all tests produce …

jfy133:
…non-empty fastq.gz files

Noticed when adding to taxprofiler that the FILTLONG module actually did not work when not supplying short reads, as the command was propogated with `-1 null -2 null` and actually produced empty `.fastq.gz` files.

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`

### DIFF
--- a/modules/filtlong/main.nf
+++ b/modules/filtlong/main.nf
@@ -20,7 +20,7 @@ process FILTLONG {
     script:
     def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
-    def short_reads = meta.single_end ? "-1 $shortreads" : "-1 ${shortreads[0]} -2 ${shortreads[1]}"
+    def short_reads = !shortreads ? "" : meta.single_end ? "-1 $shortreads" : "-1 ${shortreads[0]} -2 ${shortreads[1]}"
     """
     filtlong \\
         $short_reads \\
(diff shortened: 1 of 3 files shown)


d4straub (approved):
Good found, thanks!

> Noticed when adding to taxprofiler that the FILTLONG module actually did not work when not supplying short reads, as the command was propogated with -1 null -2 null and actually produced empty .fastq.gz files.

If that is true, why did the module produce before your fix for all tests empty files, also for those _with_ short reads? That seems a little troubling. 

Also, it seems the addition of short read data doesn't influence the output of the tests (everywhere same md5sum), might be just the small test data though.

jfy133:
> Good found, thanks!
> 
> > Noticed when adding to taxprofiler that the FILTLONG module actually did not work when not supplying short reads, as the command was propogated with -1 null -2 null and actually produced empty .fastq.gz files.
> 
> If that is true, why did the module produce before your fix for all tests empty files, also for those _with_ short reads? That seems a little troubling.

I don't know, but might be related to my comment below!

> Also, it seems the addition of short read data doesn't influence the output of the tests (everywhere same md5sum), might be just the small test data though.

Yes I think it could be in this case. The only filter I apply in the new tests is a length filter (<10bp), no quality base filtering so the short reads actually don't 'do' anything in this case.

d4straub:
I see, makes sense!

PR #1775, opened 2022-06-13
Optional output from the VEP-moduel

### LINKED ISSUE #1774: [FEATURE] Optional output from VEP in vcf-, json- or tab-format

As part of the Sarek-project, we would like the VEP-module `ENSEMBLVEP` (`modules/ensemblvep/main.nf`) could be configured to output either a vcf-file, a json-file or a tab-file, c.f. https://github.com/nf-core/sarek/issues/575

The output-file-format is controlled through `task.ext.args` while the output-file-extension is set by `task.ext.args2`. This approach was developed with @maxulysse and @FriederikeHanssen, here: https://github.com/nf-core/sarek/pull/582

_No response_

### Additional context

_No response_

asp8200:
Making the output from the VEP-moduel (ENSEMBLVEP) optionally vcf, json or tab.  Request by user in issue #1774.

Closes #1774 

- [ ] Add a resource `label`

### DIFF
--- a/modules/atlas/splitmerge/main.nf
+++ b/modules/atlas/splitmerge/main.nf
@@ -1,0 +1,36 @@
+process ATLAS_SPLITMERGE {
+    tag "$meta.id"
+    label 'process_low'
+
+    conda (params.enable_conda ? "bioconda::atlas=0.9.9" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/atlas:0.9.9--h082e891_0':
+        'quay.io/biocontainers/atlas:0.9.9--h082e891_0' }"
+
+    input:
+    tuple val(meta), path(bam), path(bai), path(read_group_settings), path(blacklist)
+
+    output:
+    tuple val(meta), path("*_mergedReads.bam"), path("*.txt.gz"), emit: data
+    path "versions.yml", emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def optional = blacklist ? 'blacklist=${blacklist}' : ''
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    atlas \\
+        task=splitMerge bam=${bam} \\
+        readGroupSettings=${read_group_settings}\\
+        $optional \\
+        $args
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        atlas: \$((atlas 2>&1) | grep Atlas | head -n 1 | sed -e 's/^[ \t]*Atlas //')
+    END_VERSIONS
+    """
+}
--- a/modules/atlas/splitmerge/meta.yml
+++ b/modules/atlas/splitmerge/meta.yml
@@ -1,0 +1,70 @@
+name: "atlas_splitmerge"
+description: split single end read groups by length and merge paired end reads
+keywords:
+  - split
+  - merge
+  - bam
+  - read group
+tools:
+  - "atlas":
+      description: "ATLAS, a suite of methods to accurately genotype and estimate genetic diversity"
+      homepage: "https://bitbucket.org/wegmannlab/atlas/wiki/Home"
+      documentation: "https://bitbucket.org/wegmannlab/atlas/wiki/Home"
+      tool_dev_url: "https://bitbucket.org/wegmannlab/atlas"
+      doi: "10.1101/105346"
+      licence: "['GPL v3']"
+
+input:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. [ id:'test', single_end:false ]
+
+  - bam:
+      type: file
+      description: Single input BAM file.
+      pattern: "*.bam"
+  - bai:
+      type: file
+      description: The BAI file for the input BAM file
+      pattern: "*.bai"
+  - read_group_setting:
+      type: file
+      description: |
+        TXT file containing the split and merge settings for
+        each readgroup. Each line consist of one readgroup,
+        single/double identifier and the maximum cycle number
+        of the sequencer. e.g. "RG1 single 100"
+      pattern: "*.txt"
+  - blacklist:
+      type: file
+      description: |
+        blacklist.txt (optional), A txt file with blacklisted read names
+        that should be ignored and just written to file, each on a new line
+      pattern: "*.txt"
+
+output:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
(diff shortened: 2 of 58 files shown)


nvnieuwk:
I don't think working with `args2` is the best way to solve this problem, I would suggest making an extra optional input field that defines the output file type (but maybe wait until @maxulysse gives some feedback)

lassefolkersen:
@nvnieuwk the VEP-arguments for file type are `-vcf`, `-json`, `-tab` - defaulting to vcf. So that is currently passed through the `-args`. So it's not `-args2` that handles the actual type. Just the ending.

nvnieuwk:
> 

Yeah but it does seem a bit redundant to that you have to specify the type and the ending on two different locations. `args2` is also mainly used to add arguments for a second tool (can't really find the docs for this right now, sorry). So I think it would be best to add an extra input field called `output_file_type` or something like that with three options: "vcf", "json", or "tab". You can then use this value to fill in the argument (e.g `--${output_file_type}`) and the file extension as `${prefix}.ann.${output_file_type}`. Also this is just a suggestion from me, I think this would be better though, since you only have to specify it once and the user needs to specify it, instead of figuring out how to do it in a config file. :)

nvnieuwk:
> > > 
> > 
> > 
> > Yeah but it does seem a bit redundant that you have to specify the type and the ending on two different locations. `args2` is also mainly used to add arguments for a second tool (can't really find the docs for this right now, sorry). So I think it would be best to add an extra input field called `output_file_type` or something like that with three options: "vcf", "json", or "tab". You can then use this value to fill in the argument (e.g `--${output_file_type}`) and the file extension as `${prefix}.ann.${output_file_type}`. Also this is just a suggestion from me, I think this would be better though, since you only have to specify it once and the user needs to specify it, instead of figuring out how to do it in a config file. :)
> 
> @nvnieuwk : Thanks for your feedback. I basically started out solving this issue in much the same way as you are proposing. However, @maxulysse and, to some degree @FriederikeHanssen, advised me to opt for the solution presented in this draft-PR. I don't have any strong opinions on this - both solutions could work for what I'm trying to do in Sarek ([nf-core/sarek#582](https://github.com/nf-core/sarek/pull/582)). @maxulysse stated that he was happy with the current solution ([nf-core/sarek#582 (comment)](https://github.com/nf-core/sarek/pull/582#issuecomment-1153829628)). I suggest that we wait for input from @maxulysse or @FriederikeHanssen. Cheers

Allright, that's fine by me then :p.

FriederikeHanssen:
> > > > 
> > > 
> > > 
> > > Yeah but it does seem a bit redundant that you have to specify the type and the ending on two different locations. `args2` is also mainly used to add arguments for a second tool (can't really find the docs for this right now, sorry). So I think it would be best to add an extra input field called `output_file_type` or something like that with three options: "vcf", "json", or "tab". You can then use this value to fill in the argument (e.g `--${output_file_type}`) and the file extension as `${prefix}.ann.${output_file_type}`. Also this is just a suggestion from me, I think this would be better though, since you only have to specify it once and the user needs to specify it, instead of figuring out how to do it in a config file. :)
> > 
> > 
> > @nvnieuwk : Thanks for your feedback. I basically started out solving this issue in much the same way as you are proposing. However, @maxulysse and, to some degree @FriederikeHanssen, advised me to opt for the solution presented in this draft-PR. I don't have any strong opinions on this - both solutions could work for what I'm trying to do in Sarek ([nf-core/sarek#582](https://github.com/nf-core/sarek/pull/582)). @maxulysse stated that he was happy with the current solution ([nf-core/sarek#582 (comment)](https://github.com/nf-core/sarek/pull/582#issuecomment-1153829628)). I suggest that we wait for input from @maxulysse or @FriederikeHanssen. Cheers
> 
> Allright, that's fine by me then :p.

Hi @nvnieuwk ! Good points raised. I think this is one of those modules were either solution has some draw backs. I was proposing a solution that would not pass the type as `input` but through the config, in order for pipeline usage to be more adaptable by the user. For example the pipeline generates vcfs, but someone really wants json they can overwrite this in the `modules.config` (which is more or less the paradigm for all modules, only hand over files & meta via input, everything is set by configs). The other problem is, that `prefix` alone is not enough here. It would work fine for the vcf/json/tab output, but the `html` file then has a weird naming scheme. However, you are right that `args2` is usually used for a second tool. Ideally I would want to see `suffix` there, it is the most descriptive of whats going on IMO, but it was sort of retired a while ago, so I was not sure if we can revive it for this one module

nvnieuwk:
> > > > > 
> > > > 
> > > > 
> > > > Yeah but it does seem a bit redundant that you have to specify the type and the ending on two different locations. `args2` is also mainly used to add arguments for a second tool (can't really find the docs for this right now, sorry). So I think it would be best to add an extra input field called `output_file_type` or something like that with three options: "vcf", "json", or "tab". You can then use this value to fill in the argument (e.g `--${output_file_type}`) and the file extension as `${prefix}.ann.${output_file_type}`. Also this is just a suggestion from me, I think this would be better though, since you only have to specify it once and the user needs to specify it, instead of figuring out how to do it in a config file. :)
> > > 
> > > 
> > > @nvnieuwk : Thanks for your feedback. I basically started out solving this issue in much the same way as you are proposing. However, @maxulysse and, to some degree @FriederikeHanssen, advised me to opt for the solution presented in this draft-PR. I don't have any strong opinions on this - both solutions could work for what I'm trying to do in Sarek ([nf-core/sarek#582](https://github.com/nf-core/sarek/pull/582)). @maxulysse stated that he was happy with the current solution ([nf-core/sarek#582 (comment)](https://github.com/nf-core/sarek/pull/582#issuecomment-1153829628)). I suggest that we wait for input from @maxulysse or @FriederikeHanssen. Cheers
> > 
> > 
> > Allright, that's fine by me then :p.
> 
> Hi @nvnieuwk ! Good points raised. I think this is one of those modules were either solution has some draw backs. I was proposing a solution that would not pass the type as `input` but through the config, in order for pipeline usage to be more adaptable by the user. For example the pipeline generates vcfs, but someone really wants json they can overwrite this in the `modules.config` (which is more or less the paradigm for all modules, only hand over files & meta via input, everything is set by configs). The other problem is, that `prefix` alone is not enough here. It would work fine for the vcf/json/tab output, but the `html` file then has a weird naming scheme. However, you are right that `args2` is usually used for a second tool. Ideally I would want to see `suffix` there, it is the most descriptive of whats going on IMO, but it was sort of retired a while ago, so I was not sure if we can revive it for this one module

Yeah I see what you mean! It's a bit of a tricky situation indeed. `suffix` seems like the best solution to this problem indeed IMO. The solution proposed by @asp8200 is maybe a good middle ground for this problem then. Seems like there is no real golden solution for this.

nvnieuwk (approved):
LGTM!

nvnieuwk (approved):

PR #1788, opened 2022-06-16
samtools/*stats*: use prefix for outputs

matthdsm:
## PR checklist

Closes #XXX 

- [x] Add a resource `label`

### DIFF
--- a/modules/samtools/stats/main.nf
+++ b/modules/samtools/stats/main.nf
@@ -8,7 +8,7 @@ process SAMTOOLS_STATS {
         'quay.io/biocontainers/samtools:1.15.1--h1170115_0' }"
 
     input:
-    tuple val(meta), path(input), path(input_index)
+    tuple val(meta), path(bam), path(bai)
     path fasta
 
     output:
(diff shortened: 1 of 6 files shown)


nvnieuwk (approved):

modules/samtools/stats/main.nf
```diff
@@ -8,7 +8,7 @@ process SAMTOOLS_STATS {
         'quay.io/biocontainers/samtools:1.15.1--h1170115_0' }"
 
     input:
-    tuple val(meta), path(input), path(input_index)
+    tuple val(meta), path(bam), path(bai)
```
maxulysse, line 11: Actually as input can be `cram`, we prefer to keep `input` here instead of `bam`

matthdsm: I was trying to come up with a bit a standard naming scheme here, since it was a bit alle over the place and the nf-core recommendations are to use `bam`.

I do agree something like `reads` would be better suited, but then that'd need to be changed across all modules. (I like uniformity in my code)

maxulysse: I like uniformity too.
I think we more or less have agreed on the notion that reads is good for unmapped reads.
For mapped reads, I'd go for either bam or cram depending on the file type, but when both are possible, we went for a more generic `input`

maxulysse: anyhow, we can continue in #1880

PR #1802, opened 2022-06-22
Proposal: Rename output from fastp module

FriederikeHanssen:
Since FastP can be used for many operations, I find it slightly confusing that in the results folders of the pipeline fastp-modified reads even if not trimmed contain `.trim.` in the file name. Ideally, I would like to set this via the `modules.config`, but seem like `prefix` is taken for some softlinking. 

Instead, I propose to just add the toolname.

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`

### DIFF
--- a/modules/fastp/main.nf
+++ b/modules/fastp/main.nf
@@ -13,7 +13,7 @@ process FASTP {
     val   save_merged
 
     output:
-    tuple val(meta), path('*.trim.fastq.gz')  , optional:true, emit: reads
+    tuple val(meta), path('*.fastp.fastq.gz')       , optional:true, emit: reads
     tuple val(meta), path('*.json')           , emit: json
     tuple val(meta), path('*.html')           , emit: html
     tuple val(meta), path('*.log')            , emit: log
@@ -34,7 +34,7 @@ process FASTP {
         [ ! -f  ${prefix}.fastq.gz ] && ln -s $reads ${prefix}.fastq.gz
         fastp \\
             --in1 ${prefix}.fastq.gz \\
-            --out1 ${prefix}.trim.fastq.gz \\
+            --out1 ${prefix}.fastp.fastq.gz \\
             --thread $task.cpus \\
             --json ${prefix}.fastp.json \\
             --html ${prefix}.fastp.html \\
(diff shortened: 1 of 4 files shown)


matthdsm (approved):

matthdsm:
LGTM, although I think you'll have to regenerate the checksums since the filenames changed

mahesh-panchal (approved):
I think this is a good idea. 
Could you check the alignment of the `, optional: true` part please?

JoseEspinosa (approved):
I am not using the tool myself, so others might have objections, but I find it a very reasonable change. Tests are failing probably due to the file renaming, feel free to merge once fixed. 🚀

modules/fastp/main.nf
```diff
@@ -34,7 +34,7 @@ process FASTP {
         [ ! -f  ${prefix}.fastq.gz ] && ln -s $reads ${prefix}.fastq.gz
```
matthdsm, line 34:
```suggestion
        [ ! -f  ${prefix}.fastq.gz ] && ln -sf $reads ${prefix}.fastq.gz
```

matthdsm: Sorry to hijack your PR, but while you're changing stuff, this could be a nice addition. Currently, fastp jobs fail on retry because of the linking.

FriederikeHanssen: what does the -sf do?

matthdsm: It overwrites the existing symlink. Without the `-f` , `ln` throws an error if it already exists, causing the job to fail

matthdsm: the other `ln` commands for paired mode should also have the `-f` added

FriederikeHanssen: cool, i think there are some more modules with a similar setup, should probably update it everywhere then

matthdsm: That might be advisable yes. Can be done as we encounter them

maxulysse (approved):
<3

maxulysse:
still some issue with linting

FriederikeHanssen:
linting fails because of empty file :( , I'll try with some different reads

FriederikeHanssen:
Alright, linting fails, because there is an empty file (was empty before but I guess the check didn't exist back then). I tried substituting with some other fastq files from the test repo but no luck so far.

Conda fails because of mismatching md5sums as reported in the  GHA log.

maxulysse:
> Alright, linting fails, because there is an empty file (was empty before but I guess the check didn't exist back then). I tried substituting with some other fastq files from the test repo but no luck so far.

I'm sure I saw that issue somewhere...

> Conda fails because of mismatching md5sums as reported in the GHA log.

As usual

PR #1830, opened 2022-06-29
New modules: `ultra/index` and `ultra/align`

sguizard:
Add ultra/index and ulta/align modules. The ultra/pipeline perform both index and align steps. Separate the two steps allow to run indexing once and reuse the index for reads alignment. 

## PR checklist

- [X] Add a resource `label`


modules/ultra/align/main.nf
```diff
@@ -0,0 +1,41 @@
...
+
+    conda (params.enable_conda ? "bioconda::ultra_bioinformatics=0.0.4" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/ultra_bioinformatics:0.0.4.1--pyh5e36f6f_0':
+        'quay.io/biocontainers/ultra_bioinformatics:0.0.4.1--pyh5e36f6f_0' }"
+
+    input:
+    tuple val(meta), path(reads)
+    path genome
+    tuple path(pickle), path(db)
```
mirpedrol, line 13: These variables are not used, shouldn't they be provided as the index path?

modules/ultra/align/main.nf
```diff
@@ -0,0 +1,41 @@
...
+
+    output:
+    tuple val(meta), path("*.sam"), emit: sam
```
mahesh-panchal, line 16: Sam files are usually requested to be compressed. Is there a reason not to compress here?

modules/ultra/index/main.nf
```diff
@@ -0,0 +1,37 @@
...
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/ultra_bioinformatics:0.0.4.1--pyh5e36f6f_0':
+        'quay.io/biocontainers/ultra_bioinformatics:0.0.4.1--pyh5e36f6f_0' }"
+
+    input:
+    path fasta
+    path gtf
+
+    output:
+    tuple path("*.pickle"), path("*.db"), emit: index
```
mahesh-panchal, line 15: Files are normally emitted into separate channels. Must these two always be present together?

tests/modules/ultra/align/main.nf
```diff
@@ -0,0 +1,26 @@
...
+    ]
+
+    genome = file(params.test_data['homo_sapiens']['genome']['genome_fasta'], checkIfExists: true)
+    gtf    = file(params.test_data['homo_sapiens']['genome']['genome_gtf']  , checkIfExists: true)
+
+    GUNZIP ( input )
+    GFFREAD ( gtf )
+    ULTRA_INDEX ( genome, GFFREAD.out.gtf )
+    ULTRA_ALIGN ( GUNZIP.out.gunzip, genome, ULTRA_INDEX.out.index )
+    SAMTOOLS_SORT ( ULTRA_ALIGN.out.sam )
```
mahesh-panchal, line 25: The SAMTOOLS_SORT process ( albeit quick ) is unnecessary.

modules/ultra/align/main.nf
sguizard: uLTRA produce sam files as output, qnd the container do not contains samtools for compressing.

modules/ultra/index/main.nf
sguizard: To be usable, the pickle files and the db files have to be both together for aligning with uLTRA align. I thought it would be the best way to keep them together.

tests/modules/ultra/align/main.nf
sguizard: Agree, I clean this.

modules/ultra/align/main.nf
mahesh-panchal: Usually nf-core requests a mulled container be built in these cases such as for the modules like Chromap , bwa mem2 , etc

mahesh-panchal: https://nf-co.re/developers/modules#software-requirements

modules/ultra/index/main.nf
mahesh-panchal: That's fair. I think it's not a good idea to split files that must be used together.

modules/ultra/align/main.nf
sguizard: Please tells me if I am wrong, mulled containers are interesting if you can pipe the result directly into samtools to avoid intermediary files? uLTRA can't print output on STDOUT. So if I mulled it with samtools, compress the sam file, at the end I will both the sam file and the bam file, which is less space efficient than the sam file alone.

mahesh-panchal: You can always `rm` the sam file. It's not the most ideal situation if you're not using `process.scratch` since your work directory will ultimately need space for both, but by the end of the process you'll only have the bam. 

Depending on the tool, process substitution can also be an option to let you pipe directly to another process. 
```bash
command --outfile >( samtools sort ... - ) ...
```

sguizard: Thanks for the resources, I had a look to the mulled containers and how to apply it to the module. 
For the uLTRA module, the management of the outputs is made with two parameters, the `--prefix` option and the output directory parameter. Therefore, I can't use process substitution to avoid temporary files.
The only option for compressing the sam file into bam will require space for both files. 
If you think that have a bam file as output is a mandatory requirement, I prepared a mulled container and I can update the code.

mahesh-panchal: I'll let the core team have the final decision on that. 
@mirpedrol What's your take?

mirpedrol: I would usually say to compress all files, but if this requires even more memory it doesn't have much sense right?

mahesh-panchal: The issue isn't memory but temporary disk space. While the process is running, it will need space for both the sam and the bam until until `rm *.sam` is called at the end of the process. I don't think it's much of an issue, and it's generally better practice to use `process.scratch`, but not everyone has access to scratch space.

mirpedrol: I see, then converting from sam to bam seems a good option 👍

sguizard: I added `samtools sort` to sort and convert the output `sam` in the same time. 
The module now fully comply to the output requirement.

modules/ultra/index/main.nf
```diff
@@ -0,0 +1,37 @@
...
+    path "versions.yml"                 , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
```
mahesh-panchal, line 22: Need `args2` now with samtools present.

mahesh-panchal (approved):
Aside from `args2` missing for samtools, all good. Thank you.

sguizard:
Thanks for the reviews!

PR #1834, opened 2022-06-30
Bump version of kallistobustools/count

apeltzer:
Just bumps kallistobustools/count to latest version. 

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`


modules/kallistobustools/count/main.nf
```diff
@@ -2,18 +2,20 @@ process KALLISTOBUSTOOLS_COUNT {
...
 
     input:
     tuple val(meta), path(reads)
     path  index
     path  t2g
     path  t1c
     path  t2c
-    val   workflow_mode
+    val   use_t1c
+    val   use_t2c
```
maxulysse, line 17: Can't we specify `-c1 $tc1` and `-c2 $tc2` only if the paths are there? instead of having these values?

apeltzer: Fair point & good idea. That should work. What I'm wondering mostly is why the `nf-core modules create-test-yml` only created YML for the `kallistbustools/ref` part...

modules/kallistobustools/count/main.nf
```diff
@@ -36,12 +38,11 @@ process KALLISTOBUSTOOLS_COUNT {
         -g $t2g \\
         $cdna \\
         $introns \\
-        --workflow $workflow_mode \\
+        --workflow $sc_workflow \\
```
maxulysse, line 41: also, can this `--workflow $sc_worfklow` be passed as args as well?
Sorry, I'm not at all familiar with the tool and params

grst, line 41: Should we add a `-m <memory>` flag. 

according to the docs it supports the giga notation: `4G`

apeltzer: Not sure either

modules/kallistobustools/count/main.nf
grst: @apeltzer @fmalmeida is this the last point missing here?

modules/kallistobustools/count/main.nf
grst: @fmalmeida, I saw you re-added this. 

We were wondering if this arguments can be passed via `ext.args`. But happy to keep `--workflow` if you think it makes more sense!

modules/kallistobustools/count/main.nf
fmalmeida: Almost. I am addressing this. There is this point and that the workflow is not correctly connected thus a YAML for the count module is not being created.

modules/kallistobustools/count/main.nf
grst: Sorry for the noise then 🙂

fmalmeida: But to use as `ext.args` we would need to also change the `REF` module, right?

fmalmeida:
Hi @grst and @apeltzer,

Now I believe the module should be looking good.

* Passed the `sc_wokflow` variable to `ext.args`
* properly connected the workflow to generate the test YML. Was just missing a "placeholder" for when the  t1c/t2c cdna and introns channels were empty
* Made the `use_t1c` and `use_t1c` not required, creating string on whether the input is available or not. When not available, the filename should be an empty string

If possible, I would say for you guys to run it as well, to make sure everything is accordingly 😄

fmalmeida:
@nf-core-bot fix linting

modules/kallistobustools/count/main.nf
```diff
@@ -24,10 +23,10 @@ process KALLISTOBUSTOOLS_COUNT {
     task.ext.when == null || task.ext.when
 
     script:
-    def args = task.ext.args ?: ''
-    def prefix = task.ext.prefix ?: "${meta.id}"
-    def cdna     = t1c ? "-c1 $t1c" : ''
-    def introns  = t2c ? "-c2 $t2c" : ''
+    def args    = task.ext.args ?: ''
+    def prefix  = task.ext.prefix ?: "${meta.id}"
+    def cdna    = ("$t1c" != "") ? "-c1 $t1c" : ''
```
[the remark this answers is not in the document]
grst: This is how this is handled in another module

https://github.com/nf-core/modules/blob/be0f022053af49d6e3dfb2006be127a23b5ddf82/modules/rtgtools/vcfeval/main.nf#L36

Does this not work?

modules/kallistobustools/count/main.nf
apeltzer: Would be ideal if both are working in the same way as  they belong together yeah

modules/kallistobustools/count/main.nf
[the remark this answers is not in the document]
fmalmeida: Yeah, it works. I was putting the expected values in the wrong sides of the ifelse statement 😅

modules/kallistobustools/count/main.nf
fmalmeida: So, that being said. Maybe we should stick with the variable then, because the `REF` module seems to dependent of it and it would be difficult to extract the fact that the standard module is being selecte through `ext.args` to then select the appropriate command line.

Look how this is in `REF`:
https://github.com/nf-core/modules/blob/b33a59e756ede6107c9a457d5a5cddb63573487e/modules/kallistobustools/ref/main.nf#L27-L64

apeltzer:
@nf-core-bot fix linting

modules/kallistobustools/count/main.nf
```diff
@@ -2,18 +2,17 @@ process KALLISTOBUSTOOLS_COUNT {
...
+        'quay.io/biocontainers/kb-python:0.27.2--pyhdfd78af_0' }"
 
     input:
     tuple val(meta), path(reads)
     path  index
     path  t2g
     path  t1c
     path  t2c
-    val   workflow_mode
     val   technology
```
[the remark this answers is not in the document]
grst: it is sample dependent

apeltzer:
@nf-core-bot fix linting

maxulysse (approved):
just need to fix linting

apeltzer:
@nf-core-bot fix linting

modules/kallistobustools/count/main.nf
apeltzer: Agreed, lets keep it like this

fmalmeida: Just to check.
You left `COUNT` as not using the variable and `REF` using? Or you re-added the variable to `COUNT` so they are the same?

PR #1841, opened 2022-06-30
Fix Krona ktimporttaxonomy to allow multiple input files and simplifies database input

jfy133:
This PR makes the following changes:

1. Removes quotes around input variable, as this causes the tool to interpret a list of files as one file name rather than as each file separately
2. Allows custom file names using ${prefix} for `-o`
3. Simpleifies the slightly convoluted database input to just require the user to point directly `.tab` file is (Rather than assuming the name of the directory etc.), infer the directory name based on the path of this file and supply this to the command.

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`


jfy133:
I need a way to infer the directory of the supplied path, in cases when `taxonomy.tab` is supplied (as in the case in the test).

I.e., detect where taxonmy.tab is in the working directory, and infer the path but without the filename

lescai (approved):
Looks good to me, makes the module easier to understand. 
It does require a little magic for the directory path :) but does not change the way channels worked before (one would still have to provide the .tab file) so it shouldn't break anyone's code.
thanks for revising this!

PR #1859, opened 2022-07-07
Bump tool versions for modules required in viralrecon

drpatelh:
- Bumped tool versions for modules required in viralrecon
- Where required I also bumped versions for other sub-commands too e.g. for all bcftools modules.


modules/nextclade/run/main.nf
```diff
@@ -29,17 +34,14 @@ process NEXTCLADE_RUN {
         run \\
         $args \\
         --jobs $task.cpus \\
-        --input-fasta $fasta \\
```
JoseEspinosa, line 32: I see that the `fasta` remains as a declare input I guess now is automatically detected by the tool, right?

JoseEspinosa (approved):
Looks great! Just one small question 🚀

PR #1867, opened 2022-07-12
Add manual version definition for artic modules

drpatelh:
Looks like the version of `artic` was bumped after the latest release:
https://github.com/artic-network/fieldbioinformatics/compare/1.2.2...master

The Biocontainer would have been built relative to the code in the tagged version and not `master` which is why it is showing the incorrect version on the CLI:

```
$ docker run -it quay.io/biocontainers/artic:1.2.2--pyhdfd78af_0 bash
root@9e6ed8b35659:/# artic --version
artic 1.2.1
```

Don't think rebuilding the Bioconda recipe will fix this because the assets will be frozen relative to the 1.2.2 tagged release. The only solution right now is to manually define the version in the module.


JoseEspinosa (approved):
LGTM! 🚀

PR #1872, opened 2022-07-12
diminish memory demand

fmalmeida:
Kallisto was being told to ask for the same amount of memory as it was said for nextflow to limit.

So, when kallisto tried to use it, nextflow was killing it. Thus, diminishing it per 1 GB solved the issue.


apeltzer (approved):

fmalmeida:
@nf-core-bot lint

PR #1908, opened 2022-07-19
update for bcftools merge

nvnieuwk:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`


modules/bcftools/merge/main.nf
```diff
@@ -9,22 +9,39 @@ process BCFTOOLS_MERGE {
...
     def args = task.ext.args   ?: ''
-    prefix   = task.ext.prefix ?: "${meta.id}"
+    def prefix   = task.ext.prefix ?: "${meta.id}"
+
+    def regions = bed ? "--regions-file $bed" : ""
+    def extension = args.contains("--output-type b") || args.contains("-Ob") ? "bcf.gz" :
+                    args.contains("--output-type u") || args.contains("-Ou") ? "bcf" :
+                    args.contains("--output-type z") || args.contains("-Oz") ? "vcf.gz" :
+                    args.contains("--output-type v") || args.contains("-Ov") ? "vcf" :
+                    "vcf.gz"
```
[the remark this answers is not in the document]
nvnieuwk: I'm not really sure what the default is, but I think it's vcf.gz indeed

modules/bcftools/merge/main.nf
```diff
@@ -9,22 +9,39 @@ process BCFTOOLS_MERGE {
...
+
     """
-    bcftools merge -Oz \\
+    bcftools merge \\
+        $regions \\
+        --threads $task.cpus \\
         --output ${prefix}.vcf.gz \\
```
maxulysse, line 41:
```suggestion
        --output ${prefix}.${extension} \\
```
Shouldn't we use the extension?

nvnieuwk: yes oops

modules/bcftools/merge/main.nf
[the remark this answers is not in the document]
maxulysse: That's was is implied by this checks anyway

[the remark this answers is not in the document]
nvnieuwk: Yes true, I'll add it like this

nvnieuwk:
It's more a case of: if it is not possible otherwise, people will have the option to get the uncompressed version, but it's not advised indeed :p

maxulysse (approved):
LGTM

JoseEspinosa:
> It's more a just in case situation here: if it is not possible otherwise, people will have the option to get the uncompressed version, but it's not advised indeed :p

👌  My point was that we could altogether remove the not compressed output and give an error if the argument to get it is present, but might be too drastic option

JoseEspinosa (approved):

nvnieuwk:
> > It's more a just in case situation here: if it is not possible otherwise, people will have the option to get the uncompressed version, but it's not advised indeed :p
> 
> ok_hand My point was that we could altogether remove the not compressed output and give an error if the argument to get it is present, but might be too drastic option

Yes, I understand what you mean! I've had a similar situation in another pipeline where I needed an uncompressed vcf file for a tool. I had to use BGZIP to decompress it first, so an option like this would have saved me an extra process there :) That's a bit why I like to keep all options open

modules/bcftools/merge/main.nf
```diff
@@ -9,22 +9,38 @@ process BCFTOOLS_MERGE {
 
     input:
     tuple val(meta), path(vcfs), path(tbis)
+    path bed
```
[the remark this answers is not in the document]
maxulysse: Good point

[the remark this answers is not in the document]
nvnieuwk: Thanks! I'll change it!

modules/bcftools/merge/main.nf
```diff
@@ -8,23 +8,38 @@ process BCFTOOLS_MERGE {
...
+    path fasta
+    path fasta_fai
 
     output:
-    tuple val(meta), path("*.gz"), emit: vcf
-    path  "versions.yml"         , emit: versions
+    tuple val(meta), path("*.vcf.gz"), optional:true , emit: vcf_gz
+    tuple val(meta), path("*.vcf")   , optional:true , emit: vcf
+    tuple val(meta), path("*.bcf.gz"), optional:true , emit: bcf_gz
+    tuple val(meta), path("*.bcf")   , optional:true , emit: bcf
```
adamrtalbot, line 19: It might be a bit bold, but you could "simplify" this to one output channel:
`    tuple val(meta), path("*{.bcf,.vcf}{,.gz}"), emit: vcf`

But it's a bit trickier to follow!

matthdsm: we could call it `variants` 🤷🏻‍♂️

modules/bcftools/convert/meta.yml
```diff
@@ -53,10 +53,22 @@ output:
       type: file
       description: File containing software versions
       pattern: "versions.yml"
+  - vcf_gz:
```
matthdsm, line 56: Did you mean to edit this file?

modules/bcftools/merge/main.nf
```diff
@@ -8,23 +8,38 @@ process BCFTOOLS_MERGE {
         'quay.io/biocontainers/bcftools:1.15.1--h0ea216a_0' }"
 
     input:
-    tuple val(meta), path(vcfs), path(tbis)
+    tuple val(meta), path(vcfs), path(tbis), path(bed)
```
matthdsm, line 11:
```suggestion
    tuple val(meta), path(vcfs), path(tbis)
    path(bed)
```

matthdsm: You're merging mutliple vcf's here. I you want to limit the merge for _all_  vcf's to a region, it makes no sense to attach the bed to a single sample

modules/bcftools/merge/main.nf
nvnieuwk: Yeah I've been thinking about that too. I could give it a try :p

modules/bcftools/convert/meta.yml
nvnieuwk: yes I forgot to add the output to meta.yml in my last PR so I quickly added it here :)

modules/bcftools/merge/main.nf
nvnieuwk: pinging @adamrtalbot for this. That was my first choice too, but here you give just one bed file for all the vcfs and tbis entered. But it's specific for these samples together and as @adamrtalbot said here: https://github.com/nf-core/modules/pull/1908#discussion_r924597652

modules/bcftools/merge/main.nf
nvnieuwk: It seems to do the job perfectly! Thanks for the suggestion

drpatelh: I second `variants` and the Adam's suggestion. So

```
tuple val(meta), path("*.{bcf,vcf}{,.gz}"), emit: variants
```

nvnieuwk: Okay, awesome! Thanks :)

modules/bcftools/merge/main.nf
adamrtalbot: @matthdsm makes a good point - I was thinking in the cases of one BED file per sample, but given we have >1 VCF here it makes that you would want to merge all VCF files under a specific set of regions. 

So yes, separate it out again. Sorry!

nvnieuwk: Okay, thanks for all the feedback!

PR #1911, opened 2022-07-20
new module: Picard/fastqtosam

matthdsm:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`


modules/picard/fastqtosam/main.nf
```diff
@@ -0,0 +1,45 @@
...
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def avail_mem = 3
+    if (!task.memory) {
+        log.info '[Picard FastqToSam] Available memory not known - defaulting to 3GB. Specify process memory requirements to change this.'
+    } else {
+        avail_mem = task.memory.giga
+    }
+    def input = meta.single_end ? "--FASTQ ${reads}" : "--FASTQ ${reads[0]} --FASTQ2 ${reads[1]}"
+    def sample_name = args.contains("--SAMPLE_NAME") || args.contains("-SM") ? "" : "--SAMPLE_NAME ${prefix}"
```
matthdsm, line 30: This doensn't seem to do anything 😞

matthdsm: The custom samplename testcase just ignores this line it seems

nvnieuwk: Why don't you just add the prefix in the `args` itself? like `ext.args = {"--SAMPLE_NAME $prefix"}` instead of `ext.args = "--SAMPLE_NAME CustomSample"` and adjusting it in the module itself?

matthdsm: `--SAMPLE_NAME` is a mandatory argument, so we need to catch the case where the user did not set an explicit sample name in `ext.args`

nvnieuwk: Allright, looks good to me then. Is it working now?

matthdsm: seems like it 👍🏻

tests/modules/picard/fastqtosam/nextflow.config
```diff
@@ -0,0 +1,7 @@
+process {
+
+    publishDir = { "${params.outdir}/${task.process.tokenize(':')[-1].tokenize('_')[0].toLowerCase()}" }
+    withName: ".*:PICARD_FQTS_CUSTOM_SAMPLENAME" {
```
Midnighter, line 4: I see why you used an import alias now. Could prefix with the workflow name to achieve the same thing, I guess.

modules/picard/fastqtosam/main.nf
```diff
@@ -0,0 +1,45 @@
+process PICARD_FASTQTOSAM {
+    tag "$meta.id"
+    label 'process_medium'
+
+    conda (params.enable_conda ? "bioconda::picard=2.27.4" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/picard:2.27.4--hdfd78af_0' :
+        'quay.io/biocontainers/picard:2.27.4--hdfd78af_0' }"
```
Midnighter, line 8: Have you tried using picard-slim as you suggested on Slack?

modules/picard/fastqtosam/meta.yml
```diff
@@ -0,0 +1,46 @@
+name: "picard_fastqtosam"
+description: Converts a FASTQ file to an unaligned BAM or SAM file.
+keywords:
+  - fastq
+  - unaligned
+  - bam
```
Midnighter, line 6: Add 'sam' as a keyword, too?

modules/picard/fastqtosam/main.nf
```diff
@@ -0,0 +1,45 @@
...
+        -Xmx${avail_mem}g \\
+        FastqToSam \\
+        ${args} \\
+        ${input} \\
+        ${sample_name} \\
+        --OUTPUT ${prefix}.bam
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        picard: \$(picard FastqToSam --version 2>&1 | grep -o 'Version:.*' | cut -f2- -d:)
```
Midnighter, line 42: Could also use `sed` here:

```suggestion
        picard: \$(picard FastqToSam --version 2>&1 | sed 's/^Version://')
```

modules/picard/fastqtosam/main.nf
```diff
@@ -0,0 +1,45 @@
...
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def avail_mem = 3
+    if (!task.memory) {
+        log.info '[Picard FastqToSam] Available memory not known - defaulting to 3GB. Specify process memory requirements to change this.'
```
Midnighter, line 25: This will create a log message for every time this task is run, right? Could be annoying. I would raise the level to warning, though.

modules/picard/fastqtosam/main.nf
```diff
@@ -0,0 +1,45 @@
...
+    } else {
+        avail_mem = task.memory.giga
+    }
```
Midnighter, line 28: I think you can use a ternary to good effect here.

```suggestion
    if (!task.memory) {
        log.warn '[Picard FastqToSam] Available memory not known - defaulting to 3GB. Specify process memory requirements to change this.'
    }
    def avail_mem = task.memory ? task.memory.giga : 3
```

Midnighter (approved):
Looks good! Tests are passing for me, too, so it should work. I have left a couple of suggestions which you can use or not.

modules/picard/fastqtosam/main.nf
matthdsm: I didn't. I was thinking most pipelines will already have the picard image downloaded for one tool or another, and this would cause an extra image to be loaded

matthdsm:
thanks for the review @Midnighter !

PR #1921, opened 2022-07-23
Added support for meta in bwamem2/index

### LINKED ISSUE #1917: bwamem2/index and minimap2/index don't consider meta

### Description of the bug

I've found out that bwamem2/index and minimap2/index don't consider meta in their input and output. Is there a reason why ?
Otherwise, I'm happy to go update the modules.

_No response_

### Relevant files

https://github.com/nf-core/modules/blob/a8b0fce8ce1f44ff3ee776639431a2472f9f308b/modules/bwamem2/index/main.nf#L11
https://github.com/nf-core/modules/blob/a8b0fce8ce1f44ff3ee776639431a2472f9f308b/modules/minimap2/index/main.nf#L10

### System information

_No response_

muffato:
I think bwamem2/index and minimap2/index should consider meta in their input and output. This would make it much easier to plug them in workflows (no need to do `.map { it[1] }`).

## PR checklist

Closes #1917

- [ ] Add a resource `label`

### DIFF
--- a/modules/bwamem2/mem/main.nf
+++ b/modules/bwamem2/mem/main.nf
@@ -13,14 +13,13 @@ process BWAMEM2_MEM {
     val   sort_bam
 
     output:
-    tuple val(merged_meta), path("*.bam"), emit: bam
-    path  "versions.yml"                 , emit: versions
+    tuple val(meta), path("*.bam"), emit: bam
+    path  "versions.yml"          , emit: versions
 
     when:
     task.ext.when == null || task.ext.when
 
     script:
-    merged_meta = meta2 + meta    // Priority given to the reads' meta map
     def args = task.ext.args ?: ''
     def args2 = task.ext.args2 ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
--- a/modules/bwamem2/mem/main.nf
+++ b/modules/bwamem2/mem/main.nf
@@ -9,17 +9,18 @@ process BWAMEM2_MEM {
 
     input:
     tuple val(meta), path(reads)
-    path  index
+    tuple val(meta2), path(index)
     val   sort_bam
 
     output:
-    tuple val(meta), path("*.bam"), emit: bam
-    path  "versions.yml"          , emit: versions
+    tuple val(merged_meta), path("*.bam"), emit: bam
+    path  "versions.yml"                 , emit: versions
 
     when:
     task.ext.when == null || task.ext.when
 
     script:
+    merged_meta = meta2 + meta    // Priority given to the reads' meta map
     def args = task.ext.args ?: ''
     def args2 = task.ext.args2 ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
--- a/modules/bwamem2/mem/meta.yml
+++ b/modules/bwamem2/mem/meta.yml
@@ -21,13 +21,18 @@ input:
   - meta:
       type: map
       description: |
-        Groovy Map containing sample information
+        Groovy Map containing sample information (about the reads)
         e.g. [ id:'test', single_end:false ]
   - reads:
       type: file
       description: |
         List of input FastQ files of size 1 and 2 for single-end and paired-end data,
         respectively.
+  - meta2:
+      type: map
+      description: |
+        Groovy Map containing sample information (about the reference genome)
+        e.g. [ id:'mus_musculus', assembly:'GRCm38' ]
   - index:
       type: file
       description: BWA genome index files
@@ -37,6 +42,11 @@ input:
       description: use samtools sort (true) or samtools view (false)
       pattern: "true or false"
 output:
+  - merged_meta:
+      type: map
+      description: |
+        Groovy Map containing sample information, combined from the reads and index parameters
+        e.g. [ id:'test', single_end:false, assembly:'GRCm38' ]
   - bam:
       type: file
       description: Output BAM file containing read alignments
--- a/modules/bwamem2/mem/meta.yml
+++ b/modules/bwamem2/mem/meta.yml
@@ -21,18 +21,13 @@ input:
   - meta:
       type: map
       description: |
-        Groovy Map containing sample information (about the reads)
+        Groovy Map containing sample information
         e.g. [ id:'test', single_end:false ]
   - reads:
       type: file
       description: |
         List of input FastQ files of size 1 and 2 for single-end and paired-end data,
         respectively.
-  - meta2:
-      type: map
-      description: |
-        Groovy Map containing sample information (about the reference genome)
-        e.g. [ id:'mus_musculus', assembly:'GRCm38' ]
   - index:
       type: file
       description: BWA genome index files
@@ -42,11 +37,11 @@ input:
       description: use samtools sort (true) or samtools view (false)
       pattern: "true or false"
 output:
-  - merged_meta:
+  - meta:
       type: map
       description: |
-        Groovy Map containing sample information, combined from the reads and index parameters
-        e.g. [ id:'test', single_end:false, assembly:'GRCm38' ]
+        Groovy Map containing sample information
+        e.g. [ id:'test', single_end:false ]
   - bam:
       type: file
       description: Output BAM file containing read alignments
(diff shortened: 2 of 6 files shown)


muffato:
Of course I need to update the tests ! 🤦🏼

muffato:
My understanding of meta is that it's a way of propagating additional information (metadata) throughout a workflow. Say the workflow is 1) get a genome, 2) index the genome, 3) align some reads; having to strip out meta before indexing, and then put it back for the alignment (`bwamem2/align` does expect meta) is convoluted and awkward. Why can't `bwamem2/index` use meta in the first place ?

nvnieuwk:
`bwamem2/mem` (which is what I think you mean with `bwamem2/align` does not require meta for the index (https://github.com/nf-core/modules/blob/master/modules/bwamem2/mem/main.nf)). It would be a bit weird to add the meta to this IMO since it's a main file used for all samples.

nvnieuwk:
@jfy133 what is your opinion on this?

jfy133:
> Why would you need to use meta for a mapper index? Isn't it the same for all samples of the same species, so you just have to use one? (Just trying to understand your reasoning :p feel free to enlighten me :) )

What if your pipeline needs to map against multiple species/references!

I'm actually all for this. I'm more and more just saying add `meta` to everything, because it costs 'nothing' to add an empty input for meta.

But I think such a change will require a large update across all aligners, and update to the guidelines. Lets bring it up there

nvnieuwk:
That indeed sounds very reasonable now. I haven't thought of it that way!

jfy133:
> That indeed sounds very reasonable now. I haven't thought of it that way!

Genomicists! :shakesfist: You sound like those 'sarek' people ;) (affectionate teasing pinging @maxulysse @FriederikeHanssen)

muffato:
@nvnieuwk . Ah, right ! In my comment, I indeed got confused between the meta map associated with the reference genome, and the meta map associated with the reads. You're right: `bwamem2/mem` only accepts the latter, not the former.
Base on your comment alone, as a compromise, I would have suggested to make `bwamem2/index` take the meta map of the fasta file, but discard it and output the index alone. My motivation for this pull-request was to be able to chain modules directly without having to edit the inputs/outputs, so the easiest would have been to keep the index free of meta in both `index` and `mem`.
But if the consensus is that the index should keep a meta map, then indeed, it should be added to `mem` too. But `mem` would then take two meta maps. Should it merge them for its output (named `bam`) ?

nvnieuwk:
> 

No problem about the confusion, it happens to us all :p. I think this could open up a very interesting discussion, let's see where it will lead!

jfy133:
> But if the consensus is that the index should keep a meta map, then indeed, it should be added to `mem` too. But `mem` would then take two meta maps. Should it merge them for its output (named `bam`) ?

That is indeed the one tricky thing here... and probably the reason why it wasn't directl implemented before (correct me if I'm wrong @drpatelh )? 

You could have clashes e.g. two `ids` in which you wouldn't necessarily know what to select (should you use `meta.id` or `meta2.id` for example) . On the flipside, the pipeline developer could modify this as necessary between modules, and if you don't want any meta - a blank `[]` is still fine.

drpatelh:
I think everyone is right here for one reason or another! 😅 

The reason it wasn't added inititally is because the `meta` map was reserved for sample-level metadata (and not generic metadata). If your pipeline uses one genome for all samples then the indexing modules don't need any sample-level metadata. However, if your pipeline has been written to accept separate reference genomes per sample e.g. [nanoseq](https://nf-co.re/nanoseq/3.0.0/usage#introduction) then it makes sense to have the reference genome info in the `meta` map too so you can propagate that through the pipeline as required.

Overall, I think we should add it in for all indexing modules because the only downside if it isn't defined is that you need to do a little channel manipulation or provide empty `meta` maps. However, if you do required the usage of one genome per sample then you can't use the modules in their current state.

maxulysse:
Agree, we should add `meta` to all modules.

muffato:
Hi. Here is how it would look like for bwamem2. `mem` takes two meta maps, and merges them, the reads' map overriding the index's in case the same key is used in both.

drpatelh:
Bit confused as to why we need 2 `meta` maps if the `meta` map is always required for sample metadata only.

maxulysse:
> Hi. Here is how it would look like for bwamem2. `mem` takes two meta maps, and merges them, the reads' map overriding the index's in case the same key is used in both.

Same here

maxulysse:
Oooh I see, you want to give out the reference with a meta map as well...
Why would you want to do so?
Can you elaborate a use case for that?

muffato:
> Bit confused as to why we need 2 `meta` maps if the `meta` map is always required for sample metadata only.

> Why would you want to do so?
> Can you elaborate a use case for that?

I thought that's what decided with https://github.com/nf-core/modules/pull/1921#issuecomment-1193947929 and https://github.com/nf-core/modules/pull/1921#issuecomment-1194106070 😕

drpatelh:
Let's keep the current standard notation:

```
input:
tuple val(meta), path(fasta)
```

modules/bwamem2/mem/main.nf
```diff
@@ -9,17 +9,18 @@ process BWAMEM2_MEM {
...
-    tuple val(meta), path("*.bam"), emit: bam
-    path  "versions.yml"          , emit: versions
+    tuple val(merged_meta), path("*.bam"), emit: bam
+    path  "versions.yml"                 , emit: versions
 
     when:
     task.ext.when == null || task.ext.when
 
     script:
+    merged_meta = meta2 + meta    // Priority given to the reads' meta map
```
drpatelh, line 23:
```suggestion
```

modules/bwamem2/mem/main.nf
```diff
@@ -9,17 +9,18 @@ process BWAMEM2_MEM {
...
     tuple val(meta), path(reads)
-    path  index
+    tuple val(meta2), path(index)
     val   sort_bam
 
     output:
-    tuple val(meta), path("*.bam"), emit: bam
-    path  "versions.yml"          , emit: versions
+    tuple val(merged_meta), path("*.bam"), emit: bam
+    path  "versions.yml"                 , emit: versions
```
drpatelh, line 17:
```suggestion
    tuple val(meta), path("*.bam"), emit: bam
    path  "versions.yml"          , emit: versions
```

modules/bwamem2/mem/meta.yml
```diff
@@ -21,13 +21,18 @@ input:
...
   - reads:
       type: file
       description: |
         List of input FastQ files of size 1 and 2 for single-end and paired-end data,
         respectively.
+  - meta2:
+      type: map
+      description: |
+        Groovy Map containing sample information (about the reference genome)
+        e.g. [ id:'mus_musculus', assembly:'GRCm38' ]
```
drpatelh, line 35:
```suggestion
```

modules/bwamem2/mem/meta.yml
```diff
@@ -21,13 +21,18 @@ input:
   - meta:
       type: map
       description: |
-        Groovy Map containing sample information
+        Groovy Map containing sample information (about the reads)
```
drpatelh, line 24:
```suggestion
        Groovy Map containing sample information
```

modules/bwamem2/mem/meta.yml
```diff
@@ -37,6 +42,11 @@ input:
       description: use samtools sort (true) or samtools view (false)
       pattern: "true or false"
 output:
+  - merged_meta:
+      type: map
+      description: |
+        Groovy Map containing sample information, combined from the reads and index parameters
+        e.g. [ id:'test', single_end:false, assembly:'GRCm38' ]
```
drpatelh, line 49:
```suggestion
  - meta:
      type: map
      description: |
        Groovy Map containing sample information
        e.g. [ id:'test', single_end:false ]
```

drpatelh (commented):
🚀

drpatelh (approved):
Thanks @muffato !

muffato:
Thanks all for your patience, and sorry I misunderstood the explanations

PR #1937, opened 2022-07-29
[COOLER] new functions - balance / makebins

nservant:
Update of the COOLER modules

- New `balance` tool
- New `makebins` tool
- I just realized that my PR also includes an old one that I did a few months ago to update the `cload`
https://github.com/nf-core/modules/pull/1404


nservant:
Of note, I removed the `md5sum` from `cload` and `balance` because coolers store creation date in the object. So the `md5sum` are always changing !

```
cooler info 1A2_1A6_vehicle.mm10.mapq_30.100.cool
{
    "bin-size": 100,
    "bin-type": "fixed",
 >>>"creation-date": "2022-06-01T14:16:44.409574",
    "format": "HDF5::Cooler",
    "format-url": "https://github.com/open2c/cooler",
    "format-version": 3,
    "generated-by": "cooler-0.8.11",
    "genome-assembly": "mm10",
    "nbins": 27255386,
    "nchroms": 22,
    "nnz": 294182628,
    "storage-mode": "symmetric-upper",
    "sum": 344516030
}
```

muffato:
Hi @nservant . Let's first do #1404 ? That'll make this PR cleaner

tests/modules/cooler/cload/test.yml
```diff
@@ -1,29 +1,23 @@
...
     - cooler/cload
     - cooler
   files:
     - path: output/cooler/test_pairs.2000000.cool
-    - path: output/cooler/test_pairs.bedpe
-      md5sum: 7f832733fc7853ebb1937b33e4c1e0de
 
 - name: cooler cload test_cooler_cload_tabix
-  command: nextflow run ./tests/modules/cooler/cload -entry test_cooler_cload_tabix -c ./tests/config/nextflow.config -c ./tests/modules/cooler/cload/nextflow.config
+  command: nextflow run ./tests/modules/cooler/cload -entry test_cooler_cload_tabix -c ./tests/config/nextflow.config  -c ./tests/modules/cooler/cload/nextflow.config
```
JoseEspinosa, line 18:
```suggestion
  command: nextflow run ./tests/modules/cooler/cload -entry test_cooler_cload_tabix -c ./tests/config/nextflow.config -c ./tests/modules/cooler/cload/nextflow.config
```

tests/modules/cooler/cload/test.yml
```diff
@@ -1,29 +1,23 @@
...
   tags:
     - cooler/cload
     - cooler
   files:
     - path: output/cooler/test_tabix.2000000.cool
-    - path: output/cooler/test_tabix.bedpe
```
JoseEspinosa, line 28: Same here

tests/modules/cooler/cload/test.yml
```diff
@@ -1,29 +1,23 @@
...
 
 - name: cooler cload test_cooler_cload_pairs
-  command: nextflow run ./tests/modules/cooler/cload -entry test_cooler_cload_pairs -c ./tests/config/nextflow.config -c ./tests/modules/cooler/cload/nextflow.config
+  command: nextflow run ./tests/modules/cooler/cload -entry test_cooler_cload_pairs -c ./tests/config/nextflow.config  -c ./tests/modules/cooler/cload/nextflow.config
   tags:
     - cooler/cload
     - cooler
   files:
     - path: output/cooler/test_pairs.2000000.cool
-    - path: output/cooler/test_pairs.bedpe
```
JoseEspinosa, line 18: Should we keep the check for the presence of the file?

JoseEspinosa (approved):
LGTM! Just dropped two minor comments

JoseEspinosa:
Sorry I missed @muffato comment above, then go ahead as you discussed and finish first the other PR 😅

muffato:
No worries, @JoseEspinosa , I should have added myself as a reviewer. I've got some questions / concerns about cload and dump, which I'll try to address in #1404 . Otherwise happy with the new functions

modules/cooler/balance/meta.yml
```diff
@@ -0,0 +1,44 @@
+name: "cooler_balance"
+description: Run matrix balancing on a cool file
+keywords:
+  - balance
+tools:
+  - "cooler":
+      description: Sparse binary format for genomic interaction matrices
+      homepage: https://cooler.readthedocs.io/en/latest/index.html
```
muffato, line 8:
```suggestion
      homepage: https://open2c.github.io/cooler/
```

modules/cooler/makebins/meta.yml
```diff
@@ -0,0 +1,33 @@
+name: "cooler_makebins"
+description: Generate fixed-width genomic bins
+keywords:
+  - makebins
+tools:
+  - "cooler":
+      description: Sparse binary format for genomic interaction matrices
+      homepage: https://cooler.readthedocs.io/en/latest/index.html
```
muffato, line 8:
```suggestion
      homepage: https://open2c.github.io/cooler/
```

modules/cooler/balance/main.nf
```diff
@@ -0,0 +1,38 @@
...
+    output:
+    tuple val(meta), path("${prefix}.${extension}"), emit: cool
+    path "versions.yml"                            , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "${cool.baseName}_balanced"
```
muffato, line 22: The convention is to default to `${meta.id}`

modules/cooler/makebins/main.nf
```diff
@@ -0,0 +1,33 @@
...
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    """
+    cooler makebins \\
+        $args \\
+        ${chromsizes} \\
+        ${cool_bin} > cooler_bins_${cool_bin}.bed
```
muffato, line 26: The output file needs to be configurable, as per https://nf-co.re/docs/contributing/modules#naming-conventions

modules/cooler/balance/main.nf
```diff
@@ -0,0 +1,38 @@
...
+    suffix = resolution ? "::$resolution" : ""
+    extension = cool.getExtension()
+    """
+    cp ${cool} ${prefix}.${extension}
```
muffato, line 26: If `${cool}` and `${prefix}.${extension}` are the same, it raises an error

modules/cooler/dump/main.nf
```diff
@@ -8,8 +8,7 @@ process COOLER_DUMP {
         'quay.io/biocontainers/cooler:0.8.11--pyh3252c3a_0' }"
 
     input:
-    tuple val(meta), path(cool)
-    val resolution
+    tuple val(meta), path(cool), val(resolution)
```
muffato, line 11: I actually propose to remove the resolution parameter

muffato (changes requested):
`cooler/balance` does *not* understand the `${file}::${resolution}` syntax like `cooler/load` and `cooler/cload`, cf https://cooler.readthedocs.io/en/latest/cli.html#cooler-balance vs https://cooler.readthedocs.io/en/latest/cli.html#cooler-load
In fact, `cooler/dump` doesn't either, and the `resolution` parameter should be removed from that module too.

Assuming you had left the "allow edits from maintainers" checkbox ticked, @nservant , I'm happy to push these changes myself, as I'm quite eager to get this PR merged.

modules/cooler/dump/main.nf
muffato, line 11: This change needs to be reflected in `tests/modules/cooler/merge/main.nf`

ewels:
> Of note, I removed the `md5sum` from `cload` and `balance` because coolers store creation date in the object. So the `md5sum` are always changing !

@nservant note that pytest-workflow has other ways of verifying file contents aside from md5sum: https://pytest-workflow.readthedocs.io/en/stable/#test-options

So you could use `contains` to check for a specific string within the file if you want, which is probably quite a bit more reliable than just checking for the path 👍🏻

muffato:
> > Of note, I removed the `md5sum` from `cload` and `balance` because coolers store creation date in the object. So the `md5sum` are always changing !
> 
> @nservant note that pytest-workflow has other ways of verifying file contents aside from md5sum: https://pytest-workflow.readthedocs.io/en/stable/#test-options
> 
> So you could use `contains` to check for a specific string within the file if you want, which is probably quite a bit more reliable than just checking for the path 👍🏻

Hi @ewels . Can't use `contains` here because the .cool files are binary. Instead, the trick is to convert them to .bedpe with cooler/dump

tests/modules/cooler/cload/test.yml
muffato: I reinstated the check on the .bedpe file, since it's the one that remains identical between runs

muffato (approved):

nservant:
Thanks for all your comments. Anything else I can do ?

muffato:
Hi @nservant . If you can confirm you're happy with my further changes, that'd be great. I hope I didn't lose anything from your original intent !

nservant:
Yes, it looks all good to me. Could you merge ?

PR #1955, opened 2022-08-16
Mpileup intervals

nickhsmith:
It seems that intervals are defined, but not actually passed into the tool. This PR should fix that


nickhsmith:
@FriederikeHanssen please review when you have time.

FriederikeHanssen (approved):

PR #1960, opened 2022-08-18
Hmmbuild

erikrikarddaniel:
## PR checklist

Adds the HMMER `hmmbuild` tool.

- [ x] This comment contains a description of changes (with reason).
- [x ] If you've fixed a bug or added code that should be tested, add tests!
- [ x] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [x ] If necessary, include test data in your PR.
- [x ] Remove all TODO statements.
- [x] Add a resource `label`


tests/config/pytest_modules.yml
```diff
@@ -2235,6 +2235,10 @@ ucsc/bedgraphtobigwig:
   - modules/ucsc/bedgraphtobigwig/**
   - tests/modules/ucsc/bedgraphtobigwig/**
 
+ucsc/bedtobigbed:
+  - modules/ucsc/bedtobigbed/**
+  - tests/modules/ucsc/bedtobigbed/**
+
```
muffato, line 2241: Sorry, this one is my fault. I renamed this module recently, and didn't think of changing its position in tests/config/pytest_modules.yml
Thanks !

modules/hmmer/hmmbuild/main.nf
```diff
@@ -0,0 +1,35 @@
...
+    conda (params.enable_conda ? "bioconda::hmmer=3.3.2" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/hmmer:3.3.2--h87f3376_2':
+        'quay.io/biocontainers/hmmer:3.3.2--h1b792b2_1' }"
+
+    input:
+    tuple val(meta), path(alignment)
+
+    output:
+    tuple val(meta), path("*.hmm"), emit: hmm
```
jfy133, line 14: `.hmm` file should be gzipped: https://github.com/nf-core/modules/blob/11cf45043cf61ab88fa02e2f67d83308ca724fe2/modules/hmmer/hmmalign/main.nf#L28 [^1]

IIRC when making the two other HMM modules, (most) HMMer tools supported gzipped input. You can ignore this message if build is an exception.

[^1]: https://nf-co.re/docs/contributing/modules#general

modules/hmmer/hmmbuild/main.nf
```diff
@@ -0,0 +1,35 @@
...
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+
+    """
+    hmmbuild \\
+        $args \\
+        --cpu $task.cpus \\
+        ${meta.id}.hmm \\
```
jfy133, line 27: I would explicitly add the `  -o <f> : direct summary output to file <f>, not stdout` as such a log may be interesting to some people (for multiQC maybe? I don't know what the contents looks like admittedly)

modules/hmmer/hmmbuild/main.nf
```diff
@@ -0,0 +1,35 @@
...
+    tag '$meta.id'
+    label 'process_low'
+
+    conda (params.enable_conda ? "bioconda::hmmer=3.3.2" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/hmmer:3.3.2--h87f3376_2':
+        'quay.io/biocontainers/hmmer:3.3.2--h1b792b2_1' }"
+
+    input:
+    tuple val(meta), path(alignment)
```
jfy133, line 11: missing optoinal inputs of:

```
  --mx <s>      : substitution score matrix (built-in matrices, with --singlemx)
  --mxfile <f>  : read substitution score matrix from file <f> (with --singlemx)
```

modules/hmmer/hmmbuild/meta.yml
```diff
@@ -0,0 +1,36 @@
+name: "hmmer_hmmbuild"
+description: create an hmm profile from a multiple sequence alignment
+keywords:
+  - search
```
jfy133, line 4:
```suggestion
  - search
  - hidden Markov model
  - HMM
  - hmmer
  - hmmsearch
```

tests/modules/hmmer/hmmbuild/main.nf
```diff
@@ -0,0 +1,15 @@
...
+
+nextflow.enable.dsl = 2
+
+include { HMMER_HMMBUILD } from '../../../../modules/hmmer/hmmbuild/main.nf'
+
+workflow test_hmmer_hmmbuild {
+    
+    input = [
+        [ id: 'PF14720' ],      // meta map
+        file('https://raw.githubusercontent.com/nf-core/test-datasets/phyloplace/testdata/PF14720_seed.alnfaa', checkIfExists: true)
```
jfy133, line 11: Generally we prefer to put a file like this into modules, because theoretically you could restrutur ethe pipeline test data.

Also maybe there already exists a input FASTA MSA you could use (e.g. with SARS-Cov2 dataset?)? The output can be nonsense, as long as the too lruns

jfy133 (changes requested):

modules/hmmer/hmmbuild/main.nf
erikrikarddaniel: I never used them... Can't they be left to `$args`?

modules/hmmer/hmmbuild/main.nf
erikrikarddaniel: Build can take a gzipped file, and autodetects. I have however never seen gzipped hmm files, probably because they're mostly not so large.

modules/hmmer/hmmbuild/main.nf
erikrikarddaniel: Difficult with the file perhaps...

jfy133: Exactly, they won't be staged, so we need to accomodate _all_ possible input files (even if optional)

erikrikarddaniel: There are many other options that are not explicitly covered, but could be left to `$args`. I think single sequence input is a very rare use case. 
Or do you perhaps mean only the `--mxfile <f>` that requires an input file? That would perhaps be difficult to have in `$args`? Should I prepare for input via `meta`?

modules/hmmer/hmmbuild/main.nf
erikrikarddaniel: Done, thanks for the suggestion.

modules/hmmer/hmmbuild/main.nf
jfy133: Oops, sorry yes - I meant just `--mxfile`! No, should not come via meta, should be an explicit input chanel - but if you don't need it can be just filled with an empty list `[]`.

jfy133: I base my request of all 'possible' input files based on what is reported in `--help` of that particular tool, so I would just focus on those.

tests/modules/hmmer/hmmbuild/main.nf
erikrikarddaniel: Right... There are alignments in SARS-Cov2, but they are, AFAIK, very large, so would take unnecessarily long. I would prefer to move this test file to the `modules` branch.

jfy133: Ok then copying over to modules is the best approach :)

modules/hmmer/hmmbuild/main.nf
jfy133: I also generatlly don't see them too, but in other cases you may end up generating hundreds of HMM files (e.g. annotating a genome) so it is preferred -definitely as the tool accepts gzipped files.

modules/hmmer/hmmbuild/main.nf
erikrikarddaniel: Done.

tests/modules/hmmer/hmmbuild/main.nf
erikrikarddaniel: Done. I placed it in `data/delete_me/hmmer` for lack of a better place. These aren't in the config file for tests in the modules repo, right?

modules/hmmer/hmmbuild/main.nf
erikrikarddaniel: Done.

modules/hmmer/hmmbuild/meta.yml
```diff
@@ -0,0 +1,40 @@
...
+      pattern: "*"
+
+output:
+  - versions:
+      type: file
+      description: File containing software versions
+      pattern: "versions.yml"
+  - hmm:
+      type: file
+      description: HMM file
```
jfy133, line 36:
```suggestion
      description: Gzipped HMM file
```

modules/hmmer/hmmbuild/meta.yml
```diff
@@ -0,0 +1,40 @@
...
+input:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. [ id:'test', single_end:false ]
+  - alignment:
+      type: file
+      description: multiple sequence alignment in fasta, clustal, stockholm or phylip format
+      pattern: "*"
```
jfy133, line 27: Missing description of `-mxfile` input

tests/modules/hmmer/hmmbuild/test.yml
```diff
@@ -0,0 +1,10 @@
+- name: hmmer hmmbuild test_hmmer_hmmbuild
+  command: nextflow run ./tests/modules/hmmer/hmmbuild -entry test_hmmer_hmmbuild -c ./tests/config/nextflow.config  -c ./tests/modules/hmmer/hmmbuild/nextflow.config
+  tags:
+    - hmmer
+    - hmmer/hmmbuild
+  files:
+    - path: output/hmmer/PF14720.hmm.gz
```
jfy133, line 7: You can still check for strings in gzipped files with pytest workflow, if you can find a good stirng in there!

jfy133 (changes requested):
Almost there, main thing is missing desctipion of mxfile in meta.yml and you're good to good

modules/hmmer/hmmbuild/meta.yml
erikrikarddaniel: Done.

tests/modules/hmmer/hmmbuild/test.yml
erikrikarddaniel: Sophisticated! Thanks for the tip!

jfy133 (approved):

PR #1980, opened 2022-08-30
Fgbio/fastqtobam

matthdsm:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`

### DIFF
--- a/modules/fgbio/fastqtobam/main.nf
+++ b/modules/fgbio/fastqtobam/main.nf
@@ -22,7 +22,9 @@ process FGBIO_FASTQTOBAM {
     def prefix = task.ext.prefix ?: "${meta.id}"
     def sample_name = args.contains("--sample") ? "" : "--sample ${prefix}"
     def library_name = args.contains("--library") ? "" : "--library ${prefix}"
-
+    def output = prefix.endsWith(".bam") ? ${prefix} :
+                 prefix.endsWith(".cram")? ${prefix} :
+                 "${prefix}.bam"
     """
 
     fgbio \\
@@ -30,7 +32,7 @@ process FGBIO_FASTQTOBAM {
         FastqToBam \\
         ${args} \\
         --input ${reads} \\
-        --output ${prefix}.bam \\
+        --output ${output} \\
         ${sample_name} \\
         ${library_name}
 
--- a/modules/fgbio/fastqtobam/main.nf
+++ b/modules/fgbio/fastqtobam/main.nf
@@ -9,11 +9,10 @@ process FGBIO_FASTQTOBAM {
 
     input:
     tuple val(meta), path(reads)
-    val read_structure
 
     output:
-    tuple val(meta), path("*_umi_converted.bam"), emit: umibam
-    path "versions.yml"                         , emit: versions
+    tuple val(meta), path("*.bam"), emit: bam
+    path "versions.yml"           , emit: versions
 
     when:
     task.ext.when == null || task.ext.when
@@ -21,17 +20,19 @@ process FGBIO_FASTQTOBAM {
     script:
     def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
+    def sample_name = args.contains("--sample") ? "" : "--sample ${prefix}"
+    def library_name = args.contains("--library") ? "" : "--library ${prefix}"
+
     """
 
     fgbio \\
         --tmp-dir=. \\
         FastqToBam \\
-        -i $reads \\
-        -o "${prefix}_umi_converted.bam" \\
-        --read-structures $read_structure \\
-        --sample $meta.id \\
-        --library $meta.id \\
-        $args
+        ${args} \\
+        --input ${reads} \\
+        --output ${prefix}.bam \\
+        ${sample_name} \\
+        ${library_name}
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
--- a/modules/fgbio/fastqtobam/main.nf
+++ b/modules/fgbio/fastqtobam/main.nf
@@ -11,7 +11,7 @@ process FGBIO_FASTQTOBAM {
     tuple val(meta), path(reads)
 
     output:
-    tuple val(meta), path("*.bam") , emit: bam, optional: true
+    tuple val(meta), path("*.bam") , emit: bam , optional: true
     tuple val(meta), path("*.cram"), emit: cram, optional: true
     path "versions.yml"            , emit: versions
 
@@ -23,9 +23,7 @@ process FGBIO_FASTQTOBAM {
     def prefix = task.ext.prefix ?: "${meta.id}"
     def sample_name = args.contains("--sample") ? "" : "--sample ${prefix}"
     def library_name = args.contains("--library") ? "" : "--library ${prefix}"
-    def output =    prefix.endsWith(".bam") ? ${prefix} :
-                    prefix.endsWith(".cram")? ${prefix} :
-                    "${prefix}.bam"
+    def output = prefix =~ /\.(bam|cram)$/ ? prefix : "${prefix}.bam"
     """
 
     fgbio \\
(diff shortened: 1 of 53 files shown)


modules/fgbio/fastqtobam/main.nf
```diff
@@ -9,29 +9,30 @@ process FGBIO_FASTQTOBAM {
...
+    def library_name = args.contains("--library") ? "" : "--library ${prefix}"
+
     """
 
     fgbio \\
         --tmp-dir=. \\
         FastqToBam \\
-        -i $reads \\
-        -o "${prefix}_umi_converted.bam" \\
-        --read-structures $read_structure \\
```
Midnighter, line 31: Is it fair to remove the `--read-structures` option completely? Are you thinking it should be included via `task.ext.args` when needed?

modules/fgbio/fastqtobam/main.nf
```diff
@@ -9,29 +9,30 @@ process FGBIO_FASTQTOBAM {
...
-        --sample $meta.id \\
-        --library $meta.id \\
-        $args
+        ${args} \\
```
Midnighter, line 31: Would it make sense to include async I/O?

> --async-io[[=true|false]]     Use asynchronous I/O where possible, e.g. for SAM and BAM files. [Default: false]

modules/fgbio/fastqtobam/meta.yml
```diff
@@ -1,6 +1,6 @@
 name: fgbio_fastqtobam
 description: |
-  Using the FGBIO tools, converts FASTQ files sequenced with UMIs into BAM files, moving the UMI barcode into the RX field of the BAM file
+  Using the FGBIO tools, converts FASTQ files sequenced into BAM files possibly moving the UMI barcode into the RX field of the BAM file
```
Midnighter, line 3:
```suggestion
  Using the fgbio tools, converts FASTQ files sequenced into BAM files possibly moving the UMI barcode into the RX field of the BAM file
```

Midnighter (commented):
Thanks for the updates to the module. Looks good to me.

One useful enhancement would be to make the output flexible since fgbio can write cram directly.

modules/fgbio/fastqtobam/main.nf
matthdsm: possibly, but I'd leave that option to `ext.args` since it might not be suitable for all filesystems

modules/fgbio/fastqtobam/main.nf
matthdsm: I'd keep it in `ext.args` (like I did in the tests) since it's a non mandatory option, and the tool also supports straight up conversion without UMI parsing

matthdsm:
> One useful enhancement would be to make the output flexible since fgbio can write cram directly.

It should support cram, but I didn't find any options in the [manual](http://fulcrumgenomics.github.io/fgbio/tools/latest/FastqToBam.html) how to enable this feature.

matthdsm:
How would you suggest we add the option? We could add a flag, but that just seems a bit ugly to me

matthdsm:
Personally, I like we've got the option to go straight to cram, but I'd still prefer using samtools for compression since we've got more control over the settings

Midnighter:
> How would you suggest we add the option? We could add a flag, but that just seems a bit ugly to me

Yeah, not entirely sure about that either. Maybe we could expect users to change the `task.ext.prefix` to include `.cram`? And otherwise default to creating `"${prefix}.bam"`?

> Personally, I like we've got the option to go straight to cram, but I'd still prefer using samtools for compression since we've got more control over the settings

What kind of settings are you thinking of? More control is usually nice but submitting less jobs and moving less files around is also nice.

matthdsm:
> And otherwise default to creating "${prefix}.bam"?
That's a possiblity, though it seems a but hacky

> What kind of settings are you thinking of?
Compression factors, block sizes etc. You can control all of those through samtools, but I don't see any options available here. There might be the possibility to tweak some HTSjdk settings, but those aren't documented

modules/fgbio/fastqtobam/main.nf
```diff
@@ -9,29 +9,30 @@ process FGBIO_FASTQTOBAM {
...
+    path "versions.yml"           , emit: versions
 
     when:
     task.ext.when == null || task.ext.when
 
     script:
     def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
+    def sample_name = args.contains("--sample") ? "" : "--sample ${prefix}"
+    def library_name = args.contains("--library") ? "" : "--library ${prefix}"
```
FriederikeHanssen, line 24: hm intuitively I would say they should go to args, but I  have been seeing some modules that have been more lenient on this :D

matthdsm: Both those options are mandatory, so that's why they're defined in the module, with a fallback to `ext.args`

modules/fgbio/fastqtobam/main.nf
```diff
@@ -22,15 +22,17 @@ process FGBIO_FASTQTOBAM {
     def prefix = task.ext.prefix ?: "${meta.id}"
     def sample_name = args.contains("--sample") ? "" : "--sample ${prefix}"
     def library_name = args.contains("--library") ? "" : "--library ${prefix}"
-
+    def output = prefix.endsWith(".bam") ? ${prefix} :
+                 prefix.endsWith(".cram")? ${prefix} :
+                 "${prefix}.bam"
```
Midnighter, line 27: Maybe a bit slimmer with a regex (untested):

```suggestion
    def output = prefix =~ /\.(bam|cram)$/ ? prefix : "${prefix}.bam"
```

matthdsm:
I added some updates, but it seems like the config options aren't getting picked up. Probably something wrong with the selectors.
Suggestions are welcome!

Midnighter:
> I added some updates, but it seems like the config options aren't getting picked up. Probably something wrong with the selectors. Suggestions are welcome!

Can you say more about what you mean exactly? Selectors are usually a workflow configuration thing and not for modules or am I wrong?

matthdsm:
I'm talking about `nextflow.config` in the module tests. I'm trying to add different config options to test for multiple usecases.

matthdsm:
I dropped the checksum for the cram output because the tests don't seem able to get a match.
However:
- I converted 2 separate runs to sam and ran a `diff` without anything coming up
- I manually checksummed 2 separate runs and came up with the samen checksum
So no idea what's going wrong here...

modules/fgbio/fastqtobam/main.nf
```diff
@@ -9,29 +9,31 @@ process FGBIO_FASTQTOBAM {
...
+    path "versions.yml"            , emit: versions
 
     when:
     task.ext.when == null || task.ext.when
 
     script:
     def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
+    def sample_name = args.contains("--sample") ? "" : "--sample ${prefix}"
+    def library_name = args.contains("--library") ? "" : "--library ${prefix}"
```
Midnighter, line 25: I see an issue with these, when the prefix is already named `.bam` or `.cram` then this will also be the sample and library name. I think that should either be stripped or use an optional input to toggle CRAM. What do you think?

matthdsm: I noticed that too, but when you're already adding option to modify the `$prefix` you might as well set `--sample` and `--library` in `ext.args` IMO.

matthdsm: I added a test where both are set explicitly to test for this case

Midnighter: Perfect, I only commented on this because you were _not_ configuring it that way in the tests 😉

matthdsm: yeah, I did use separate tests. I probably should've done it all in one.

Midnighter (approved):
Very good improvements. It's slightly uncomfortable that the cram md5sums don't match or can otherwise be compared but overall the module looks much better.

matthdsm:
Thanks! 
It's kind of weird the checksums don't match, but I did as much as possible to rule out any weird stuff 🤷🏻‍♂️

PR #1985, opened 2022-08-31
Creation of Glimpse Chunk

LouisLeNezet:
# First subtool of the glimpse module
Add sub tool chunk with configuration and needed files for unitest.

## PR checklist

Closes #XXX 

- [X] Add a resource `label`


LouisLeNezet:
Hello,

Does the parameter 'glimpse' in the test:
`nf-core modules lint glimpse`
come from the name of the branch ?
So I would need to do as many branch as subtools I need to PR ?

I think it would be nice to precise those informations in the documentations available.

LouisLeNezet:
I just find why.
The pytest.yml was referring to the whole glimpse folder and not the subtool folder.

modules/glimpse/chunk/main.nf
```diff
@@ -0,0 +1,43 @@
...
+    output:
+    tuple val(meta), path("*.txt"), emit: chunk_chr
+    path "versions.yml"           , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def prefix          = task.ext.prefix ?: "${meta.id}"
+    def thread_command  =                    "--thread $task.cpus"
```
nvnieuwk, line 23: I would suggest just putting `--thread $task.cpus` in the command itself, since you don't really check for anything here and it will be the same everytime you run it.

modules/glimpse/chunk/main.nf
```diff
@@ -0,0 +1,43 @@
...
+
+    def input_command   = reference_panel ? "--input $reference_panel" : ""
```
nvnieuwk, line 25: Is it normal to run this command without the actual input? If not, I would remove this check and also move this to the command itself.

modules/glimpse/chunk/main.nf
```diff
@@ -0,0 +1,43 @@
...
+
+    output:
+    tuple val(meta), path("*.txt"), emit: chunk_chr
+    path "versions.yml"           , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def prefix          = task.ext.prefix ?: "${meta.id}"
```
nvnieuwk, line 22: Don't forget to add `def args = task.ext.args?: ""`

modules/glimpse/chunk/meta.yml
```diff
@@ -0,0 +1,84 @@
...
+        The file could possibly be without GT field (for efficiency reasons a file containing only the positions is recommended).
+      pattern: "*.{vcf,bcf}"
+
+  - region:
+      type: string
+      description: |
+        Target region, usually a full chromosome (e.g. chr20:1000000-2000000 or chr20).
+        For chrX, please treat PAR and non-PAR regions as different choromosome in order to avoid mixing ploidy.
+
+  - window-size:
```
nvnieuwk, line 35: You should only add the inputs of the process here. So everything after window-size shouldn't be in here.

modules/glimpse/chunk/meta.yml
```diff
@@ -0,0 +1,84 @@
...
+      licence: "['MIT']"
+
+input:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. [ id:'test', single_end:false ]
+
+  - input:
```
nvnieuwk, line 22: The input is called `reference_panel` inside the module, so it's better to also use this here instead of `input`:)

tests/modules/glimpse/chunk/main.nf
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env nextflow
+
+nextflow.enable.dsl = 2
+
+include { GLIMPSE_CHUNK } from '../../../../modules/glimpse/chunk/main.nf' addParams( options: [args: ''] )
```
nvnieuwk, line 5: Just curious: what does this `addParams` block do? I've never seen it before :)

tests/modules/glimpse/chunk/nextflow.config
```diff
@@ -0,0 +1,14 @@
+process {
+    withName: glimpse_chunk {
+        params {
```
nvnieuwk, line 3: You don't have to use params in the modules test, I mean it works, but it's just more typing :). You can just add the raw values in the `args` itself

nvnieuwk (changes requested):

modules/glimpse/chunk/meta.yml
LouisLeNezet: Okay. If so do I need to specify somewhere all the different non mandatory parameters available for the software ?

nvnieuwk: No that's not necessary as far as I know

modules/glimpse/chunk/meta.yml
LouisLeNezet: Maybe it will be better to do the opposite ?
For people not to mix-up the different argument, I should maybe use the same name as used in the software ?

nvnieuwk: Yeah that sounds like a good idea :)

tests/modules/glimpse/chunk/main.nf
LouisLeNezet: It's an old part I wrote, that should set the args parameter as an empty string.
But it's no more necessary.

LouisLeNezet:
Hi,
When I run nf-core modules lint glimpse/chunk I get a warning:
`modules/glimpse/chunk/main.nf │ Container versions do not match`
Do you know where does it come from ?
Thanks

modules/glimpse/chunk/main.nf
```diff
@@ -19,20 +19,16 @@ process GLIMPSE_CHUNK {
...
     def VERSION = '1.1.1' // WARN: Version information not provided by tool on CLI. Please update this string when bumping container versions.
 
     """
     GLIMPSE_chunk \\
         $args \\
-        $input_command \\
-        $region_command \\
-        $thread_command \\
+        --input $input \\
+        --region $region \\
```
nvnieuwk, line 30: Is region a mandatory option? If not, you could keep the previous code for the region

nvnieuwk (changes requested):

nvnieuwk:
> Hi, When I run nf-core modules lint glimpse/chunk I get a warning: `modules/glimpse/chunk/main.nf │ Container versions do not match` Do you know where does it come from ? Thanks

Your version of the galaxyproject and quay containers do not match, you can see if it works if both containers are the same version, but it shouldn't matter that much because they both contain the same version of the tool

modules/glimpse/chunk/main.nf
LouisLeNezet: It is a mandatory option.
The thing is that function ask for a specific region in the genome (generally a chromosome: eg 'chr21') and will cut this whole region in small part ready to use for the next function of the software.

nvnieuwk: allright, this is perfect then!

nvnieuwk (approved):
LGTM

modules/glimpse/chunk/main.nf
LouisLeNezet: I don't quite understand where I and when I should add the non-mandatory options.
Do I need to explicitly put them if there is a need for a control ?

nvnieuwk: If a option is not mandatory you need to add a check to see if the options has a value (Nextflow does not natively support optional inputs, so you need to enter `[]` for the inputs that should be empty). The checks are exactly like you did with `region_command` for example. You first check if the `region` is specified, if it is specified you tell it to add this parameter with the input to the command, otherwise you just let it add an empty string that does nothing.

LouisLeNezet: And what is exactly the difference with `args` ?
Because there is plenty of non mandatory arguments that I could add ([here](https://odelaneau.github.io/GLIMPSE/commands.html)).
Do I need to specify them ?

LouisLeNezet:
I'm quite new to the whole git world and I don't really know the whole vocabulary ^^
Does LGTM means: Looks Good To Me or Let's Get This Merged ?

modules/glimpse/chunk/main.nf
nvnieuwk: `args` is mainly used to add arguments that do not need an input file. Although you can do some magic stuff with it, but you won't have to do that very often.

nvnieuwk:
> I'm quite new to the whole git world and I don't really know the whole vocabulary ^^ Does LGTM means: Looks Good To Me or Let's Get This Merged ?

There is not really one good answer for this, both are good. I always read it as Looks Good To Me though :)

LouisLeNezet:
> 

So now all I need to do is to `Squash and merge` ?

nvnieuwk:
> > 
> 
> So now all I need to do is to `Squash and merge` ?

Yep! Your very first contribution to `nf-core`, congratulations!

PR #2009, opened 2022-09-05
Add hmmer/eslreformat

erikrikarddaniel:
## PR checklist

The `esl-reformat` tool distributed with HMMER reformats aligned or not sequence files between different formats.

- [x] Add a resource `label`


erikrikarddaniel:
Did I do something so that two of the tests were skipped?

nvnieuwk:
It seems like `tests/config/pytest_modules.yml` was removed. I think that's the problem

erikrikarddaniel:
> It seems like `tests/config/pytest_modules.yml` was removed. I think that's the problem

Thanks. `prettier tests/config/pytest_modules.yml` reports lots of errors and outputs an empty file. I must be doing something wrong...

nvnieuwk:
> 

You can run `prettier -w .`. This should fix this normally (this command checks for linting errors and automatically fixes those it can fix)

modules/hmmer/eslreformat/main.nf
```diff
@@ -0,0 +1,38 @@
...
+    label 'process_low'
+
+    conda (params.enable_conda ? "bioconda::hmmer=3.3.2" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/hmmer:3.3.2--h1b792b2_1':
+        'quay.io/biocontainers/hmmer:3.3.2--h1b792b2_1' }"
+
+    input:
+    tuple val(meta), path(seqfile)
+    val   format
```
nvnieuwk, line 12: Inputs should only be used for actual file inputs. You can specify the `format` in the `nextflow.config` file using `ext.args` like: https://github.com/nf-core/modules/blob/b5e3d16d80abfc6aa1c86ffde8c1b507bea80f0d/tests/modules/bcftools/convert/nextflow.config#L9-L11

modules/hmmer/eslreformat/meta.yml
```diff
@@ -0,0 +1,44 @@
...
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. [ id:'test' ]
+  - seqfile:
+      type: file
+      description: Sequences, aligned or not, in any supported format
+      pattern: "*"
+  - format:
```
nvnieuwk, line 24: Don't forget to remove format here too :)

nvnieuwk (changes requested):

erikrikarddaniel:
> > 
> 
> You can run `prettier -w .`. This should fix this normally (this command checks for linting errors and automatically fixes those it can fix)

Right. I had a duplicate entry, which somehow messed things up. Deleting that seems to have worked.

modules/hmmer/eslreformat/main.nf
erikrikarddaniel: OK, makes sense. I thought this was only for "real" params with minuses...

modules/hmmer/eslreformat/main.nf
```diff
@@ -0,0 +1,36 @@
...
+    output:
+    tuple val(meta), path("*.sequences.gz"), emit: seqreformated
+    path "versions.yml"                    , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
```
nvnieuwk, line 22: You could also add a check here to see if the user has specified the format in the `args` and if not, you could add a default. A bit like this: https://github.com/nf-core/modules/blob/b5e3d16d80abfc6aa1c86ffde8c1b507bea80f0d/modules/bcftools/convert/main.nf#L31-L35

nvnieuwk (changes requested):

modules/hmmer/eslreformat/main.nf
erikrikarddaniel: That's a little bit tricky as the available formats depends on whether the input file is aligned or not. I'd rather not have to guess that, as there are so many file suffixes used for various sequence formats.

Another problem with this is that the format must come second last, after all the `--` params, before the filename, so the caller needs to get the order right in the config file.

nvnieuwk: Allright, maybe better to leave it out then. But you should maybe specify in `meta.yml` that the user should give a format in `args`

erikrikarddaniel: Yes, I was looking for a place to document that, thanks. As it doesn't fit the headers in `meta.yml`, should one do that as a comment? Do you have an example of how arguments are documented in `meta.yml`?

nvnieuwk: I don't really know of an example, but I think it's okay if you just add it to the top description. Or you could add it to the description of the output file?

erikrikarddaniel: OK, done and pushed, in the hope it didn't get too long!

nvnieuwk: Looks good!

PR #2023, opened 2022-09-07
epang/place

erikrikarddaniel:
## PR checklist

This adds a module for the [EPA-NG](https://github.com/Pbdas/epa-ng) tool. The program has different modes of action controlled by which inputs are given to it. Although this is not exactly subcommands, I thought it would be cleanest to implement this as subcommands, of which this is the first, and the only one I'm planning at the moment. There are hence several file input options that are not supported by this module.

Besides three input file parameters, the `--model` parameter is key, and is likely to be controlled by a parameter or an input table in the calling workflow. I put this into the `meta` map.

- [x] Add a resource `label`


modules/epang/place/main.nf
```diff
@@ -0,0 +1,41 @@
...
+    tag "$meta.id"
+    label 'process_high'
+
+    conda (params.enable_conda ? "bioconda::epa-ng=0.3.8" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/epa-ng:0.3.8--h9a82719_1':
+        'quay.io/biocontainers/epa-ng:0.3.8--h9a82719_1' }"
+
+    input:
+    tuple val(meta), path(queryaln), path(referencealn), path(referencetree)
```
d4straub, line 11: Do I understand correctely that `queryaln` is essentially the sequences to be placed (i.e. samples), and `referencealn` & `referencetree` reference files?
If yes, I'd assume that `referencealn` & `referencetree` are sample independent and therefore I would suggest
```suggestion
    tuple val(meta), path(queryaln)
    path(referencealn)
    path(referencetree)
```

modules/epang/place/main.nf
```diff
@@ -0,0 +1,41 @@
...
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    epa-ng \\
+        $args \\
+        --threads $task.cpus \\
+        --query $queryaln \\
+        --ref-msa $referencealn \\
+        --tree $referencetree \\
+        --model $meta.model
```
d4straub, line 31: Is the model sample specific? If yes, it might be fine to have it in the meta map (or as part of the above input tuple), if no, it might be better added as `val` (or `path` if file, but it seems to me its a string?) under `input`.

tests/modules/epang/place/test.yml
```diff
@@ -0,0 +1,13 @@
+- name: epang_place test_epang_place
+  command: nextflow run ./tests/modules/epang/place -entry test_epang_place -c ./tests/config/nextflow.config  -c ./tests/modules/epang/place/nextflow.config
+  tags:
+    - epang
+    - epang/place
+  files:
+    - path: output/epang/test.epa_info.log
+      contains:
+        - "INFO 3 Sequences done"
+    - path: output/epang/test.epa_result.jplace
```
d4straub, line 10: No checksum? So the output isnt reproducible? (Just making sure)

d4straub (commented):
Looks pretty good I think, but I am uncertain about a few points.

modules/epang/place/main.nf
erikrikarddaniel: Your half correct. In most cases I think the reference alignment and tree will differ between samples. But I changed anyway.

modules/epang/place/main.nf
erikrikarddaniel: It's actually tied to the reference, so there's a possibility that it's not, but I kept it in `meta` anyway for the moment as I was suggested not to have other value parameters in another module.

tests/modules/epang/place/test.yml
erikrikarddaniel: According to the test generator it's not, which  is a bit funny as it's not gzipped. Haven't looked further into this.

modules/epang/meta.yml
```diff
@@ -0,0 +1,67 @@
...
+input:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. [ id:'test', model:'LG' ]
+  - queryaln:
+      type: file
+      description: aligned query sequences in any supported format including phylip and fasta, may be gzipped
+      pattern: "*"
```
jfy133, line 25:
```suggestion
      pattern: "*.{fa,fas,fna,fasta,fa.gz,.fas.gz,fna.gz,fasta.gz,.phylip,.phylip.gz}"
```

modules/epang/meta.yml
```diff
@@ -0,0 +1,67 @@
...
+  - referencealn:
+      type: file
+      description: reference alignment in any supported format including phylip and fasta, may be gzipped
+      pattern: "*"
```
jfy133, line 29:
```suggestion
      pattern: "*.{fa,fas,fna,fasta,fa.gz,.fas.gz,fna.gz,fasta.gz,.phylip,.phylip.gz}"
```

modules/epang/meta.yml
```diff
@@ -0,0 +1,67 @@
...
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. [ id:'test' ]
+  - versions:
+      type: file
+      description: File containing software versions
+      pattern: "versions.yml"
+  - jplace:
```
jfy133, line 57: Missing output directory

modules/epang/main.nf
```diff
@@ -0,0 +1,57 @@
...
+    def args       = task.ext.args   ?: ''
+    def prefix     = task.ext.prefix ?: "${meta.id}"
+    def queryarg   = queryaln        ? "--query $queryaln"       : ""
+    def refalnarg  = referencealn    ? "--ref-msa $referencealn" : ""
+    def reftreearg = referencetree   ? "--tree $referencetree"   : ""
+    def modelarg   = meta.model      ? "--model $meta.model"     : ""
+    def bfastarg   = bfastfile       ? "--bfast $bfastfile"      : ""
+    def splitarg   = splitfile       ? "--split $splitfile"      : ""
+    def binaryarg  = binaryfile      ? "--binary $binaryfile"    : ""
+    """
```
jfy133, line 37:
```suggestion
    if ( binaryfile && ( referencealn || referencetree ) ) error "[EPANG] Cannot supply both binary and reference MSA or reference tree. Check input"
    """
```

Addresses your comment on slack?

modules/epang/main.nf
```diff
@@ -0,0 +1,57 @@
...
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args       = task.ext.args   ?: ''
+    def prefix     = task.ext.prefix ?: "${meta.id}"
+    def queryarg   = queryaln        ? "--query $queryaln"       : ""
+    def refalnarg  = referencealn    ? "--ref-msa $referencealn" : ""
+    def reftreearg = referencetree   ? "--tree $referencetree"   : ""
+    def modelarg   = meta.model      ? "--model $meta.model"     : ""
```
jfy133, line 33: I just remembered, you don't need this at all. You can do this entirely with `ext.args` in `modules.conf`! this way you don't have to 'assume' a `meta` entry at all!

```
    ext.args = { "--model ${meta.model}" }
```

This way you don't have to assume anything and reduce the number of input channels.

We do this in taxprofiler: https://github.com/nf-core/taxprofiler/blob/6f87c681fd455a30594b2fb37f8d42e7f668b067/conf/modules.config#L267

jfy133 (changes requested):
I wouldn't assume the name contents of `meta`. I just remembered you don't need to specify it within the module at all - this can be done with `ext.args!.

modules/epang/meta.yml
erikrikarddaniel: There's `faa` and `faa.gz` too, plus I'm using `alnfaa`/`alnfna` for aligned fasta, I'm sure there are other suffixes. But I suppose this is not validated anywhere, or? If so, one could of course have these patterns, but I don't see the point really if the list is very long.

jfy133: Ok fair enough! Can ignore this one

modules/epang/meta.yml
erikrikarddaniel: Thanks, done!

modules/epang/meta.yml
```diff
@@ -0,0 +1,70 @@
...
+
+output:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. [ id:'test' ]
+  - epang:
+      type: directory
+      description: directory in which EPA-NG was run
```
jfy133, line 55:
```suggestion
      description: directory in which EPA-NG was run containing all output files
```

jfy133 (approved):
Note I didn't check the condition for the  mututally exclusive intpu files, probably good to check first :sweat_smile: 

Otherwise LGTM!

erikrikarddaniel:
> Note I didn't check the condition for the mututally exclusive intpu files, probably good to check first sweat_smile
> 
It looked very promising, but I'll try to add a test!

> Otherwise LGTM!

erikrikarddaniel:
> > Note I didn't check the condition for the mututally exclusive intpu files, probably good to check first sweat_smile
> 
> It looked very promising, but I'll try to add a test!
> 
It turned out to be more work than expected to test the check for incompatible file inputs, since binary format requires nucleotides.

> > Otherwise LGTM!

Thanks!

PR #2036, opened 2022-09-09
Added an output channel for the gzi index

muffato:
When bgzip is given the `-i` option, it creates an index (in gzi format). The index needs to be considered an output so that it can be published. This PR adds an optional output channel to capture the index.

The `-I` option is automatically injected to name the index according to the convention.

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`

### DIFF
--- a/modules/tabix/bgzip/main.nf
+++ b/modules/tabix/bgzip/main.nf
@@ -12,7 +12,7 @@ process TABIX_BGZIP {
 
     output:
     tuple val(meta), path("${prefix}*"), emit: output
-    tuple val(meta), path("index/*gzi"), emit: gzi, optional: true
+    tuple val(meta), path("*gzi")      , emit: gzi, optional: true
     path  "versions.yml"               , emit: versions
 
     when:
@@ -24,9 +24,8 @@ process TABIX_BGZIP {
     in_bgzip = input.toString().endsWith(".gz")
     command1 = in_bgzip ? '-d' : '-c'
     command2 = in_bgzip ? ''   : " > ${prefix}.${input.getExtension()}.gz"
-    gzi_args = args.matches("(^| )-i\\b") ? "-I index/${prefix}.${input.getExtension()}.gz.gzi" : ''
+    gzi_args = args.matches("(^| )-i\\b") ? "-I ${prefix}.${input.getExtension()}.gz.gzi" : ''
     """
-    mkdir -p index
     bgzip $command1 $args $gzi_args -@${task.cpus} $input $command2
 
     cat <<-END_VERSIONS > versions.yml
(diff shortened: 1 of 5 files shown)


modules/tabix/bgzip/main.nf
```diff
@@ -23,8 +24,10 @@ process TABIX_BGZIP {
     in_bgzip = input.toString().endsWith(".gz")
     command1 = in_bgzip ? '-d' : '-c'
     command2 = in_bgzip ? ''   : " > ${prefix}.${input.getExtension()}.gz"
+    gzi_args = args.matches("(^| )-i\\b") ? "-I index/${prefix}.${input.getExtension()}.gz.gzi" : ''
     """
-    bgzip $command1 $args -@${task.cpus} $input $command2
+    mkdir -p index
```
FriederikeHanssen, line 29: why does it need its own subfolder?

muffato: Let me try again, but I think I tried first keeping everything in the same folder, and somehow the index was picked up by the "output" channel, not the "gzi" channel

FriederikeHanssen: ah yes I can see why , the ${prefix}* will pick up everything. there is probably a regex way around it.

muffato: It works ! Thanks for the poke

modules/tabix/bgzip/main.nf
```diff
@@ -23,8 +24,9 @@ process TABIX_BGZIP {
     in_bgzip = input.toString().endsWith(".gz")
     command1 = in_bgzip ? '-d' : '-c'
     command2 = in_bgzip ? ''   : " > ${prefix}.${input.getExtension()}.gz"
+    gzi_args = args.matches("(^| )-i\\b") ? "-I ${prefix}.${input.getExtension()}.gz.gzi" : ''
     """
-    bgzip $command1 $args -@${task.cpus} $input $command2
+    bgzip $command1 $args $gzi_args -@${task.cpus} $input $command2
```
maxulysse, line 29: I'd rather we use $args2 instead of $gzi_args as we do in other modules

FriederikeHanssen: why do we need the second one in the first place? Can you not check the $args string for what you need?

FriederikeHanssen:
@muffato sorry for the review delay :/ I was OOO

modules/tabix/bgzip/main.nf
muffato: Good point @FriederikeHanssen : before adding `-I (...)` I should actually double-check it's not already there. And I should check both the short and long forms of the options.

Therefore, the feature becomes two-fold:

1. Capture the `.gzi` file if indexing has been requested (`-i` or `--index`)
2. Name the `.gzi` file according to the standard prefix unless a specific name has been requested (`-I` or `--index-name`)

I'm not keen on using `$args2` because the convention seems to be 1) it refers to the _second_ command of the script, whereas here it would be used in the _first_ command, and 2) it's always initialised from user input (`task.ext.args2`), whereas here we want to complement the existing arguments from `task.ext.args`.
With the latest version, I modify `$args` directly.

FriederikeHanssen (approved):
🚀 thank you for adding that

muffato:
Thank you very much !

PR #2053, opened 2022-09-13
Add hmmer/unalign

erikrikarddaniel:
## PR checklist

This module uses the HMMER distributed tool `esl-reformat` to "unalign" sequences in an alignment file and outputs them
as a gzipped fasta file.

Reading the source code for `esl-reformat` it seems one could just pass `fasta` to the tool to unalign. However, this fails for me.

- [x] Add a resource `label`

### DIFF
--- a/modules/hmmer/unalign/main.nf
+++ b/modules/hmmer/unalign/main.nf
@@ -1,0 +1,37 @@
+process HMMER_UNALIGN {
+    tag "$meta.id"
+    label 'process_low'
+
+    conda (params.enable_conda ? "bioconda::hmmer=3.3.2" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/hmmer:3.3.2--h1b792b2_1':
+        'quay.io/biocontainers/hmmer:3.3.2--h1b792b2_1' }"
+
+    input:
+    tuple val(meta), path(alignment)
+
+    output:
+    tuple val(meta), path("*.fa.gz"), emit: fasta
+    path "versions.yml"             , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    esl-reformat \\
+        $args \\
+        --gapsym=- \\
+        afa \\
+        $alignment | \\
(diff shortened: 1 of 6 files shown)


erikrikarddaniel:
> Looks good to me, but it seems to me that the module is quite limited in its output by
> 
> ```
>         --gapsym=- \\
>         afa \\
>         $alignment | \\
>         sed '/^>/!s/-//g' | \\
> ```
> 
> It seems to me that it really can only output a fasta file. But the command could do so much more if `--gapsym=- afa` would be kept to task.ext.args & the sed command removed (or if you insist adding a ` def sed_cmd = task.ext.sed_cmd ?: ""`). But I might be wrong. 

The thing is that there is a module based on the same tool (`esl-reformat`) that can reformat the output if someone needs that. Keeping to fasta only here makes it much easier to use the module, I think, because one doesn't have to figure out what to put in the config file for the pipeline.

In the earlier DSL2 module format nf-core used, there was a possibility to add a pipe argument for postprocessing, but I'm unsure of this. Something like: `command $ext.args $input $postprocessing | gzip -c > $output` where `postprocessing` could be defined in a config file to e.g. `| sed 's/NN/MM/g'`. What you say above (`sed_cmd`) sounds like I'm free to add a postprocessing step, or? If so, I could add that to the `hmmer/eslreformat` module and I'm done.

> Do you think such a msa to fasta conversion is a common task?

Not very common, but, as I mention in the PR description, I can foresee that the subworkflow I use in phyloplace can be copied over to Ampliseq. Having all required modules in nf-core would make that easier, I believe.

d4straub:
> The thing is that there is a module based on the same tool (esl-reformat) that can reformat the output if someone needs that. Keeping to fasta only here makes it much easier to use the module, I think, because one doesn't have to figure out what to put in the config file for the pipeline.

True, it would be easier to format it into a fasta with pre-set such as here. That is why I think if its a common task it would be fine to have such a module.

> What you say above (sed_cmd) sounds like I'm free to add a postprocessing step, or? 

Yes, `sed_command` that I proposed could be also `postprocessing`.

> If so, I could add that to the hmmer/eslreformat module and I'm done.

Yes, I think that should work. But please do not take my word for it but test it!

erikrikarddaniel:
@d4straub , I modified the `hmmer/eslreformat` module to allow postprocessing. I'm happy to close this one if you prefer that.

d4straub:
Great! Yes, I prefer the other change.
(Sorry, I am very slow here, WIFI in German trains is close to non-existant,)

PR #2063, opened 2022-09-15
Update MULTIQC for an additional custom config

jfy133:
Adds ability to have one(!) other custom config to the MultiQC module.

This allows nf-core pipelines to have the default pipeline one, and then a user to modify sections by supplying their second config file, the contents of which [overrides anything common](https://multiqc.info/docs/#configuring-multiqc) with the pipeline base config file. 

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`

### DIFF
--- a/modules/multiqc/main.nf
+++ b/modules/multiqc/main.nf
@@ -9,6 +9,7 @@ process MULTIQC {
     input:
     path  multiqc_files, stageAs: "?/*"
     path(multiqc_config)
+    path(multiqc_logo)
 
     output:
     path "*multiqc_report.html", emit: report
--- a/modules/multiqc/main.nf
+++ b/modules/multiqc/main.nf
@@ -9,6 +9,7 @@ process MULTIQC {
     input:
     path  multiqc_files, stageAs: "?/*"
     path(multiqc_config)
+    path(extra_multiqc_config)
     path(multiqc_logo)
 
     output:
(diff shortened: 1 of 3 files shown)


ewels (approved):
Wonder if we should stick to `custom_` instead of `extra_` to match the params names that we have in pipelines?

Otherwise looks good 👍🏻

jfy133:
> Wonder if we should stick to `custom_` instead of `extra_` to match the params names that we have in pipelines?
> 
> Otherwise looks good 👍🏻

I put it as `extra` because if someone was to use the module outside an nf-core pipeline they might not have a 'default'

ggabernet:
I'm late to the party @jfy133 and @ewels , but is there a reason why not mixing the default config + custom config channels does not work, as it was used until now?
I've just experienced a problem with the current implementation, if the user does not want to provide a custom config, the the channel is empty and the process is not executed

ggabernet:
maybe I missed your discussion in Slack, tried to find something about it but didn't

jfy133:
We weren't even sure if just mixing them straight in worked in the first place. According to the MultiQC docs, each file needs to have `-c` or `--config` at the beginning so they get loaded in sequence.

However your problem shouldn't happen :-1: , can you point to the pipeline code with the new implementation? (Note I am about to make a template update how it should work in theorey...)

ggabernet:
ok great, then I will just look at your implementation ;)

ggabernet:
Ah NVM, forgot to add `.ifEmpty([])`, it works for me now! Thanks for the update!

PR #2074, opened 2022-09-15
fix(bases2fastq): Remove stray ls

Emiller88:


FriederikeHanssen (commented):
looks good, but why do the singularity checksums not match?

Emiller88:
IDK, it's using the docker container, so it should be the same. It could be because the RunManifest.json gets rewritten with any overrides and maybe the singularity image uses different line endings?

matthdsm (approved):
LGTM, pending the CI 😛

JoseEspinosa (commented):
It has been reported as one of the failing modules after the restructuring (#2154)
So it would be great if you could update the list once the tests pass :smiley:

PR #2097, opened 2022-09-23
[POC] Get subworkflows working again - bam_stats_samtools

drpatelh:
This has been tested locally with `pytest` and works! @FriederikeHanssen @mribeirodantas witnessed it.

The idea is to use this PR/subworkflow as a template for the functionality we plan to add to nf-core/tools and to get the Github actions working again. The subworkflow specific tests were working but we stripped them out previously and need to add them back in. Ping @Emiller88 . The plan is to configure the subworkflow tests so they are automatically triggered when the upstream modules are changed.

@mirpedrol is working on a solution so the anchors in `pytest_modules.yml` aren't stripped out by nf-core/tools when creating modules.

Let's try to get this working as soon as possible so we can start properly tackling this at the Hackathon in a couple of weeks.


tests/config/pytest_modules.yml
```diff
@@ -2307,6 +2307,13 @@ subread/featurecounts:
   - modules/subread/featurecounts/**
   - tests/modules/subread/featurecounts/**
 
+subworkflows/bam_stats_samtools: &subworkflows_bam_stats_samtools
+  - subworkflows/nf-core/bam_stats_samtools/**
+  - tests/subworkflows/nf-core/bam_stats_samtools/**
+  # - *samtools_stats
```
drpatelh, line 2313: Had to comment this out because Prettier and nf-core linting was failing. Not sure whether they need to be added back in.

mirpedrol:
Fix for maintaining anchors in the yml file in #1836, I still need to figure out how to maintain the indentation so Prettier won't fail.

tests/config/pytest_modules.yml
```diff
@@ -2467,3 +2474,6 @@ yara/index:
 yara/mapper:
   - modules/yara/mapper/**
   - tests/modules/yara/mapper/**
+
+__anchors__:
```
Emiller88, line 2478: Just for my understanding, what's the purpose of the explicit anchor?

tests/config/pytest_modules.yml
Emiller88: There are two options to get the tests to trigger:
1. Add the module anchors to the subworkflows (less things to update and maintain because we only have to update the module triggers)
2. Add the tags to the pytest workflow spec.

I think we stayed away from 2 in the past because it triggers all the subworkflows locally, but personally those are all things that should be tested, and I think with `nf-core modules test` we can just use the specific test name if we want to avoid that behavior locally for most users and then let the CI fail when they've at least got the module behaving. It's a bit more code to maintain and update, but it's very declarative, and less "clever" so I think it'll keep the barrier for understanding low and therefor keep the barrier for contributing low, which should be our goal here.

So I vote 2, less work on the nf-core tools side and then we avoid anchors.

tests/subworkflows/nf-core/bam_stats_samtools/test.yml
```diff
@@ -1,13 +1,14 @@
 - name: bam stats samtools single-end
-  command: nextflow run ./tests/subworkflows/nf-core/bam_stats_samtools -entry test_bam_stats_samtools_single_end -c tests/config/nextflow.config
+  command: nextflow run ./tests/subworkflows/nf-core/bam_stats_samtools -entry test_bam_stats_samtools_single_end -c ./tests/config/nextflow.config -c ./tests/subworkflows/nf-core/bam_stats_samtools/nextflow.config
   tags:
```
Emiller88, line 3:
```suggestion
  tags: &tags
```

We could use an anchor here to avoid copying and pasting the tags over and over, but that's probably going to make linting harder.

Emiller88 (approved):
Looks good to me! I vote we do 2 and avoid anchors, I think some people were against adding all the modules before because it made testing locally difficult, but I think we can fix that with some small nf-core tools tweaks.

tests/config/pytest_modules.yml
mirpedrol: To be able to maintain the anchors in the file when we update it with tools, we need to add the aliases. I think that grouping all of them under the same name can be easier to maintain. 
The only modification that I found while adding the functionality in tools is that we would  need to rename `__anchors__` so it doesn't get mixed with other modules, as we sort them alphabetically, I would suggest `~anchors:` which will make it be last in the Python sorting.

drpatelh:
> I don't think we can test whether the module changes work until we merge this
From @Emiller88 on Slack so we know why this was merged! Let's test it out then!

PR #2104, opened 2022-09-26
Adding module bamtools stats

lescai:
## PR checklist

- [ ] Bamtools exists already among the module tools: this PR adds its "stats" function, which is an easy and very lightweight solution to generate bam statistics, less resource consuming than other tools. It is already compatible with multiQC, and it will be therefore easy to include in other nf-core pipelines
- [ ] followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] Uses standard SARS-CoV-2 test bam
- [ ] Add a resource `label`
- All test carried out, and module works with either Docker / Singularity. Conda CI tests can be quite flaky:

### DIFF
--- a/modules/bamtools/stats/main.nf
+++ b/modules/bamtools/stats/main.nf
@@ -1,6 +1,6 @@
 process BAMTOOLS_STATS {
     tag "$meta.id"
-    label 'process_low'
+    label 'process_single'
 
     conda (params.enable_conda ? "bioconda::bamtools=2.5.1" : null)
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
--- a/modules/bamtools/stats/main.nf
+++ b/modules/bamtools/stats/main.nf
@@ -1,0 +1,35 @@
+process BAMTOOLS_STATS {
+    tag "$meta.id"
+    label 'process_low'
+
+    conda (params.enable_conda ? "bioconda::bamtools=2.5.1" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/bamtools:2.5.1--h9a82719_9' :
+        'quay.io/biocontainers/bamtools:2.5.1--h9a82719_9' }"
+
+    input:
+    tuple val(meta), path(bam)
+
+    output:
+    tuple val(meta), path("*.stats"), emit: stats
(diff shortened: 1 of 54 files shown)


modules/bamtools/stats/main.nf
```diff
@@ -0,0 +1,35 @@
+process BAMTOOLS_STATS {
+    tag "$meta.id"
+    label 'process_low'
```
SPPearce, line 3: This should be process_single now we have that.

modules/bamtools/stats/main.nf
```diff
@@ -0,0 +1,35 @@
...
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+
+    """
+    bamtools \\
+        stats \\
+        -in $bam \\
+        >${prefix}.bam.stats
+
```
SPPearce, line 29: Is there any optional arguments for the tool at all?

lescai: there isn't. there's an alternative input `-list` instead of `-in` if you have multiple bam files listed in a file, but I thought in an nf-core pipeline we would want to distribute the tasks.

modules/bamtools/stats/main.nf
lescai: "should"? I kind of missed that. nf-core/tools version 2.5.1 suggests "process_low" as default

SPPearce: Sorry, I should have been more explicit. The `process_single` tag is a new one that has been recently added, so that isn't present in tools (or the documentation by the looks of it). It only requests one core by default, instead of the two that `process_low` requests.

SPPearce: To be precise, these are the default options for the process tag:
https://github.com/nf-core/tools/blob/master/nf_core/pipeline-template/conf/base.config#L29

SPPearce: If the tool is going to work fine with 6GB of RAM, then single is the better choice for a program that doesn't use more than one core.

lescai: super @SPPearce thanks! the process is single threaded, but wasn't sure about the memory. I've checked the profiles of execution on my cluster (exomes of different species) and it should be fine. just committed the change.

SPPearce (approved):

PR #2126, opened 2022-09-28
Clean up duplicate/deprecated samtools/fastq modules

matthdsm:
Drop modules:
- samtools/bam2fq, deprecated
- samtools/collatefastq, absorbed into main samtools/fastq module

## PR checklist
- [ ] Add a resource `label`


modules/samtools/fastq/main.nf
```diff
@@ -20,14 +23,28 @@ process SAMTOOLS_FASTQ {
...
+    // Interleaved / SE output
+    // -0 ${prefix}_other.fastq.gz | gzip --no-name > ${prefix}_interleaved.fastq.gz
+
+    // PE output
+    // -1 ${prefix}_1.fastq.gz -2 ${prefix}_2.fastq.gz -0 ${prefix}_other.fastq.gz -s ${prefix}_singleton.fastq.gz
+
+    // NOTE: When outputting fastq for downstream alignment, the "-n" flag should be set to avoid editing the read names
     """
-    samtools \\
+    samtools collate \\
```
maxulysse, line 37:
```suggestion
    samtools \\
      collate \\
```

maxulysse:
shouldn't we actually keep collatefastq and remove fastq.
I feel like the new module name should reveal what's inside

matthdsm:
I'm not sure.. I agree with the descriptive name part, but without looking into the sarek code I wouldn't have guessed `collatefastq` actually converts bam to fastq.
Perhaps that's just me

lescai:
To be honest, while I agree with the need to cleanup bam2fq which is no longer a valid command under the current version, [_samtools fastq_](https://www.htslib.org/doc/samtools-fasta.html) and [_samtools collate_](https://www.htslib.org/doc/samtools-collate.html) are 2 different commands under samtools.
I'd find very confusing to put them inside the same module. 

As we say in the guidelines, the module should be as [atomic](https://nf-co.re/docs/contributing/modules#module) as possible, which is what samtools fastq was doing. _collatefastq_ responds to a specific need, which is ok.

If you incorporate _samtools collate_ inside the original _fastq_ module though, you introduce an unnecessary command for those who use the module for their original purpose (i.e. converting a single bam to a fastq): this is why it would be a breaking change, and it needs introducing 2 unnecessary inputs for those who just want to convert a single bam file.
In theory I could still pass a single bam and get the result, but then the first part of the pipe under _collate_ is completely unnecessary.
I am quite against this change I'm afraid.

matthdsm:
I get the atomic module part, but there have been precedents where sticking 2 tools together is allowed when the functions complement each other (especially when they're part of the same package)

lescai:
I'm not entirely sure why you'd say that `samtools collate` is necessary to run `samtools fastq` "correctly".
if the 2 commands are separate functions, they're quite independent aren't they?
that some of us might prefer or need to pipe them together, is quite different from saying that's the correct way to run them. I'm sure there's many different use cases and we'd like to accommodate them as flexibly as we can

matthdsm:
It's not required per se, but it _is_ advised when you want to use the fastq downstream for alignment
https://www.htslib.org/doc/samtools-fasta.html
As I said, it changes nothing for people who already have a "correct" input, but facilitates the use of the module for those who don't.

I agree with the need for being flexible, but IMO all I see are benefits 🤷🏻‍♂️

matthdsm:
I could live with that, but wanted to avoid as much duplication as possible

matthdsm:
@lescai , are there any cases where you could see `collate` having a negative impact on the outputs?

lescai:
In our specific use case, we extract reads from a bam using flags, then we convert the resulting bam with `samtools fastq`. Having those reads grouped by their names _should_ not have a negative impact on the downstream analysis.
I haven't run tests to compare the 2 though. I am not particularly happy to pass inputs that are not really needed for the computation I have to run, but I don't want to create a blocker either if the majority only sees benefits of course.
it seems we're splitting hairs :)

matthdsm:
Oh, if you're worried about the extra inputs, remember they're all optional, so you can just add `[]` for each argument you don't want to bother with

priyanka-surana:
If we are keeping the combined module, then keep the name as `collatefastq` as well. Because it best represents the module.

matthdsm:
So we've got 2 votes for keeping both.
I'll ask for a last round of opinions and then get cracking to fix the PR

matthdsm:
I also feel weird about the name, but I've got no better alternative

nvnieuwk:
I agree with keeping both `collatefastq` and `fastq`. This does remind me a lot about the existence of both `samtools/view` and `samtools/convert` which does almost exactly the same with the exception that `convert` also outputs and index for convenience.

matthdsm:
@nf-core/core, can we get a vote from you guys?

lescai:
works for me, I believe we clarified any doubts we had. thanks for the effort guys!

priyanka-surana:
Yes I agree with `collate`, `fastq` and to be named `collate | fastq` modules

matthdsm:
Closing, started to fix this on a per module basis in #2354

PR #2127, opened 2022-09-28
Update cores specification in TrimGalore module

drpatelh:
Ping @jordeu for review


jordeu:
With this change runs ~50% faster. Still have some spikes of CPU that reach the 12 allocated CPUs, even if at the end the average is ~8 CPUs. So I think it's a minimal change that will benefit all the users that use default allocation values.   

Screenshot from 2022-09-28 15-21-25

jordeu (approved):

ewels (approved):

PR #1562, opened 2022-04-25
Allow to pass arguments to the phantompeakqualtools script itself

JoseEspinosa:
Add `args2` to the command so that it is possible to pass arguments to the `phantompeakqualtools` script itself.

- [x] This comment contains a description of changes (with reason).
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - [x] `PROFILE=docker pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [x] `PROFILE=singularity pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [x] `PROFILE=conda pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`

PR #2129, opened 2022-09-28
Add tags to test subworkflows

Emiller88:
This updates the tags for subworkflows, so the tests get triggered.

### DIFF
--- a/tests/subworkflows/nf-core/annotation/ensemblvep/test.yml
+++ b/tests/subworkflows/nf-core/annotation/ensemblvep/test.yml
@@ -1,7 +1,11 @@
 - name: ensemblvep annotation_ensemblvep
   command: nextflow run ./tests/subworkflows/nf-core/annotation_ensemblvep -entry annotation_ensemblvep -c ./tests/config/nextflow.config -c ./tests/subworkflows/nf-core/annotation_ensemblvep/nextflow.config
   tags:
-    - annotation_ensemblvep
+    - subworkflows
+    - subworkflows/annotation/ensemblvep
+    # Modules
+    - ensemblvep
+    - tabix/bgziptabix
   files:
     - path: output/annotation/test_VEP.ann.vcf.gz
     - path: output/annotation/test_VEP.ann.vcf.gz.tbi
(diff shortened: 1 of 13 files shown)


tests/subworkflows/nf-core/annotation/ensemblvep/test.yml
```diff
@@ -1,7 +1,11 @@
 - name: ensemblvep annotation_ensemblvep
   command: nextflow run ./tests/subworkflows/nf-core/annotation_ensemblvep -entry annotation_ensemblvep -c ./tests/config/nextflow.config -c ./tests/subworkflows/nf-core/annotation_ensemblvep/nextflow.config
   tags:
-    - annotation_ensemblvep
+    - subworkflows
+    - subworkflows/annotation/ensemblvep
```
mirpedrol, line 5: Can subworkflows have subfolders? I didn't take it into account in the new nf-core tools commands 😮

tests/subworkflows/nf-core/annotation/ensemblvep/test.yml
```diff
@@ -1,7 +1,11 @@
 - name: ensemblvep annotation_ensemblvep
   command: nextflow run ./tests/subworkflows/nf-core/annotation_ensemblvep -entry annotation_ensemblvep -c ./tests/config/nextflow.config -c ./tests/subworkflows/nf-core/annotation_ensemblvep/nextflow.config
```
mirpedrol, line 2:
```suggestion
  command: nextflow run ./tests/subworkflows/nf-core/annotation_ensemblvep -entry annotation_ensemblvep -c ./tests/config/nextflow.config
```
These config paths can be removed, they weren't removed from modules to avoid changing too many files (see discussion in [#1562](https://github.com/nf-core/tools/issues/1562)). As we are starting fresh with subworkflows I think it's a good opportunity to avoid them :)

tests/subworkflows/nf-core/annotation/ensemblvep/test.yml
Emiller88: Apparently. Ii makes sense though for `gatk` for example, I want to move those into their own directory. But we can put that off for now and just do something like `annotation_ => annotation/` in the future if we ever get around to it.

tests/subworkflows/nf-core/annotation/ensemblvep/test.yml
Emiller88: Good idea, done!

drpatelh (approved):
LGTM!

PR #2136, opened 2022-09-30
Explicitly name the output files to avoid the gzi index being flown through the `output` channel

muffato:
This is a fix for the issue I've mentioned on Slack (https://nfcore.slack.com/archives/CJRH30T6V/p1664528844937629)

The problem was that the `.gzi` file, when available, was flown through both output channels (`output` and `gzi`). I considered regular expressions as advised in the Slack thread, but it felt like it'd be a tricky regex to write.
Seeing the other discussion that I was referred to (about DeepVariant), I also had the idea of being more explicit about the file names I'm expecting, which is actually straightforward in the case of bgzip, since it simply removes the extension.

At the same time, I've added a few more valid bgzip extensions, cf https://github.com/samtools/htslib/blob/develop/bgzip.c#L82

I've tested locally that the output of the channels is correct. If you've got any idea how to embed those checks in the CI, I'm all open

## PR checklist

- [ ] Add a resource `label`

### DIFF
--- a/modules/tabix/bgzip/main.nf
+++ b/modules/tabix/bgzip/main.nf
@@ -11,9 +11,9 @@ process TABIX_BGZIP {
     tuple val(meta), path(input)
 
     output:
-    tuple val(meta), path("${prefix}*"), emit: output
-    tuple val(meta), path("*gzi")      , emit: gzi, optional: true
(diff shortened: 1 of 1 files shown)


nvnieuwk (approved):
Nice fix!

PR #2156, opened 2022-10-06
Update and fix bam_sort_samtools subworkflow

drpatelh:


subworkflows/nf-core/bam_sort_samtools/main.nf
```diff
@@ -2,23 +2,20 @@
...
-    ch_bam // channel: [ val(meta), [ bam ] ]
+    bam   // channel: [ val(meta), [ bam ] ]
+    fasta // channel: [ fasta ]
 
     main:
 
     ch_versions = Channel.empty()
 
-    SAMTOOLS_SORT ( ch_bam )
+    SAMTOOLS_SORT ( bam )
```
maxulysse, line 18: why not keep `ch_bam`, if you have `ch_bam_bai`?

drpatelh: Yeah, I debated about that to myself before I made this change. You are probably right. I'll update.

maxulysse (approved):

subworkflows/nf-core/bam_sort_samtools/main.nf
matthdsm: I though the consensus was to prefix all channels with `ch_`?  I could swear I read that somewhere...

drpatelh: It was. But in some instances we may be passing a `file` object instead of a channel like `fasta` where this wouldn't apply. In these scenarios, `ch_` may not necessarily apply. Fine to have it in for all things files/channels and not for booleans and other types of objects.

Issue #2155, opened 2022-10-06
new module: deseq2/differential


pinin4fjords:
See https://nfcore.slack.com/archives/CE8SSJV3N/p1664996414210659?thread_ts=1661174347.177469&cid=CE8SSJV3N

Wrapper module for differential analysis with DESeq. I have my own version, and @ggabernet flagged her existing pipeline at https://github.com/qbic-pipelines/rnadeseq. 

Suggest keeping differential analysis distinct from plotting, to allow reuse of the latter and facilitate software dependency management per process.

PR #2167, opened 2022-10-06
Add starting deseq2 differential module

pinin4fjords:
This PR adds a wrapper for DESeq2. Notes:

- Implemented as a template, initially due to need to include a wrapper script and current lack of module `bin` directory (see https://nfcore.slack.com/archives/CJRH30T6V/p1665066826554919)
  - I think this actually works better than use of a module `bin`,  because we don't need an additional library to provide a CLI! So here we can just use the biocontainer for DESeq2 and nothing else, no need to create bespoke containers etc.
- Optional params can be supplied via the usual `task.ext.args` route, and are handled within the template.
- I've defined contrasts with a groovy map, in an analagous way to sample meta in other workflows, comprising:
  - variable - a variable from the supplied sample sheet that can be used to group samples
  - reference - level of contrast variable indicating 'reference' samples.
  - target -  level of contrast variable indicating 'target'/ 'treatment' samples.
  - blocking - comma-separated list of blocking variables to be included in the model.
- A test expression matrix was added to the test data [here.](https://github.com/nf-core/test-datasets/pull/637) 
- Tiny variations between machines at the 12th decimal point (or so) in some test outputs were causing the CI to fail. So I've just added some very limited rounding to ensure the reproducibility, which can be activated specifically in the CI context.

## PR checklist

Tackling https://github.com/nf-core/modules/issues/2155

- [x] Adds module for differential expression analysis with DESeq2
- [x] Add a resource `label`


pinin4fjords:
@nf-core-bot fix-linting

pinin4fjords:
@nf-core-bot fix linting

modules/deseq2/differential/main.nf
```diff
@@ -0,0 +1,46 @@
+params.deseq_gene_id_col = "gene_id"
```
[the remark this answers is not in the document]
pinin4fjords: Ahh, thanks- I knew this wasn't quite right, but I was having trouble seeing how it should be done properly given the template situation, and I didn't see any other usages of module.config in the nf-core modules.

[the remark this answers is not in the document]
pinin4fjords: @SPPearce sorry, just to clarify, by module.config do you mean the pipeline-level [modules.config](https://github.com/nf-core/rnaseq/blob/master/conf/modules.config), or is there something at the module level I can use to set the defaults?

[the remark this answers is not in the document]
SPPearce: The module level yes.
Generally the `ext.args` option is used to access this. If you look in the `tests/modules/` files you can see how it is then used.
The pipelines have a config file in `conf/modules.config` that controls these options for each module.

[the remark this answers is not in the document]
pinin4fjords: @SPPearce Just to double-check:

**What I have now**

```
params.deseq_gene_id_col = "gene_id"
```

Which I can retrieve from within the template like:

```
opt <- list(
    ...
    'gene_id_col' = '$params.deseq_gene_id_col',
...
)
```

I assume that param value could be overridden in calling workflows. 

**What should be happening**

`task.ext.args` is a single string, right? So I'd have to parse the 'gene_id' value out of a longer string like `--deseq_gene_id_col column_A --foo bar --another-option blah`.

Because I'm using a template rather than an existing CLI tool, I then have to assign that variable back to something that can be seen from the template- which I think probably means `params`. 

That seems like a lot of trouble to avoid using params in the first place :-D

nvnieuwk:
Did you use the `nf-core modules create` tool to create this module? The test files are absent in this PR (which is why some of the actions aren't running)

pinin4fjords:
> Did you use the `nf-core modules create` tool to create this module? The test files are absent in this PR (which is why some of the actions aren't running)

I'll get there- just a WIP for visibility and discussion, I'll get things tested before requesting formal review.

nvnieuwk:
Maybe change this to a draft PR then :)

pinin4fjords:
@nf-core-bot fix linting

pinin4fjords:
@nf-core-bot fix linting

pinin4fjords:
@nf-core-bot fix linting

pinin4fjords:
Almost there, just tripping up on reproducibility of results with DESeq I think- puzzled how the same container could be doing something different local and remote. Doing `set.seed()` in R locally does not change the result files, so I don't think that's a valid fix...

pinin4fjords:
huh, now some of the CI md5sums have flipped back to a state they had previously, so this is not entirely random

modules/deseq2/differential/main.nf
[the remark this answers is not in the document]
SPPearce: When you aren't using a template, you'd just have this sort of set up: 
`deseq_de.r $ext.args`
where the `modules.config` will specify e.g.
`ext.args = '--gene_id_col "gene_id" --sample_id_col = "experiment_accession" --test = "Wald" --fit_type = "parametric"'`
and you'd use an argument parsing function inside the R script. 
So then in a pipeline, the modules.config can be customised (and may well refer to external params for them!). 
I haven't used template personally (I've used the bin folder in our local pipeline to do this), I assume that you can't do  e.g. `template 'my_script.R -a foo -b bar'?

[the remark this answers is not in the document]
pinin4fjords: No, I don't think they work that way. Even if they did I'd have to compose the CLI for the script, which would require r-optparse, which would make the software dependency handling harder (right now I can just use the bioconductor-deseq2 biocontainer, which is very convenient). Plus I'm not sure what the purpose would be of contriving that CLI to pass parameters that can be read directly by the template.

Essentially I think the  params/ template usage is justified in situations like this where we're wrapping libraries rather than command line tools, which from poking around in this repo seems quite novel for nf-core (though I can see myself using the pattern repeatedly).

The test pipeline that's in place now illustrates the usage, in that parameter settings can be overridden from the calling workflow and read directly by the template. I really don't see the need to stuff those params into an arg string and parse them back out again for the template (though it makes complete sense in the more conventional case where args are passed through to a command line tool).

SPPearce:
Don't worry too much about rounding differences in the output files, that is a common issue. You can just check for a string inside the file instead, or failing that that the file exists at all.

I think we do need `optparse` here. You can pass the `args` string along by generating a template file that calls the R script, so you have `templates/deseq_de.R` and `templates/deseq_de.sh`, 

with `templates/deseq_de.sh` as:
```
#!/usr/bin/env sh
deseq_de.R --sample_file !{samplesheet} --counts_file !{counts} !{args}
```
This will pass the samplesheet and counts file names correctly based on the inputs, and then any other optional arguments from `task.ext.args`.
Then in the main.nf you have:
```
args = task.ext.args ?: 'blah blah' 
template "deseq_de.sh"
```
Without `optparse`, you could manually sort through the `args` string, but that seems suboptimal to just getting a container with deseq2 and optparse. There is [a string ](https://github.com/antunderwood/multi-package-containers/blob/master/combinations/hash.tsv#L92) suggesting a multi-container exists which would do the job, but I can't figure out how to actually find the resulting name, I get nothing from `mulled-search --destination quay singularity conda  --search deseq2  | grep "mulled"`.

pinin4fjords:
Thanks @SPPearce - appreciate the info, the rationale is clearer to me now, and it makes sense not to have the global parameters. Will update the PR notes to match the below.

In https://github.com/nf-core/modules/pull/2167/commits/4b4adfc44fce37c80e64631be600a7fba05130e4 I take the overrides from `task.ext.args` and parse . Doing a `def args` doesn't seem to work for templates since a new`args` wasn't visible from within the template. Re-assigning task.ext.args worked, but seemed messy, and handling the overrides directly in R seemed neater anyway. All the optional params can now be supplied via `--` options. 

I hope you'll forgive my persistence in being an optparse refusenik. I can see myself making quite a few of these based on different R packages and I'd really like to avoid having to create a new container each time just to include optparse. We don't need most of the functionality anyway, since users won't be calling the script directly, and the basic parsing is pretty trivial to implement.

SPPearce:
I have no qualms about you using optparse or not personally, I definitely agree that it is much easier to use predefined containers! Although I struggle writing anything in base-R anymore without tidyverse available to me nowadays.
And it looks like you have clean enough code for performing that, does it error if you provide an unexpected argument (i.e. a typo)?
Are you expecting the contrast_meta to be constructed from rows of a csv file?
I'll take a proper look tomorrow and make suggestions.

pinin4fjords:
> does it error if you provide an unexpected argument (i.e. a typo)? 

Yep! https://github.com/nf-core/modules/blob/d59f148cb848b4afabc6d6c2041651f28610bddb/modules/nf-core/deseq2/differential/templates/deseq_de.r#L67

> Are you expecting the contrast_meta to be constructed from rows of a csv file? 

Yep, that's what I have locally

> I'll take a proper look tomorrow and make suggestions.

Thanks!

modules/nf-core/deseq2/differential/main.nf
```diff
@@ -0,0 +1,30 @@
...
+        'https://depot.galaxyproject.org/singularity/bioconductor-deseq2%3A1.32.0--r41h399db7b_0' :
+        'quay.io/biocontainers/bioconductor-deseq2:1.32.0--r41h399db7b_0' }"
+
+    input:
+    path samplesheet
+    path counts
+    val contrast_meta
+
+    output:
+    tuple val(contrast_meta), path("deseq2.results.tsv")               , emit: results
```
SPPearce, line 16: Do we want a prefix on the output files? Currently every time you run this module, all the output files will be identically named. Which means that if you want to do any downstream combining of the outputs or just outputting them into a folder, you'll have to ensure they are renamed or put into a folder in some way anyway.

modules/nf-core/deseq2/differential/main.nf
```diff
@@ -0,0 +1,30 @@
...
+    tuple val(contrast_meta), path("normalised_counts.tsv")            , emit: normalised_counts
+    tuple val(contrast_meta), path("variance_stabilised_counts.tsv")   , emit: variance_stabilised_counts
+    tuple val(contrast_meta), path("R_sessionInfo.log")                , emit: session_info
+    tuple val(contrast_meta), path("versions.yml")                     , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    template 'deseq_de.r'
```
SPPearce, line 29: I know this is minor, but I think that the file extension should be `.R`, not `.r`. That is consistent with the [Bioconductor style guide](https://contributions.bioconductor.org/r-code.html#rcode-file-names) amongst others.

modules/nf-core/deseq2/differential/meta.yml
```diff
@@ -0,0 +1,63 @@
+name: "deseq2_differential"
+description: runs a differential expression analysis with DESeq2
+keywords:
+  - differential
+  - expression
+  - rna-seq
+
```
SPPearce, line 7:
```suggestion
  - deseq2
```

modules/nf-core/deseq2/differential/meta.yml
```diff
@@ -0,0 +1,63 @@
...
+        CSV-format sample sheet with sample metadata
+  - counts:
+      type: file
+      description: |
+        Raw TSV-format expression matrix as output from the nf-core RNA-seq workflow
+  - contrast_meta:
+      type: map
+      description: |
+        Groovy Map containing contrast information
+        e.g. [ variable:'treamtent', reference:'treated', control:'saline', blocking:'' ]
```
SPPearce, line 30:
```suggestion
        e.g. [ variable:'treatment', reference:'treated', control:'saline', blocking:'' ]
```

modules/nf-core/deseq2/differential/templates/deseq_de.r
```diff
@@ -0,0 +1,354 @@
...
+    'deseq2.dispersion.png',
+    width = 600,
+    height = 600
+)
+plotDispEsts(dds)
+dev.off()
+
+# R object for other processes to use
+
+save(dds, file = 'dds.rld.RData')
```
SPPearce, line 280: If you are saving just one object, then use `saveRDS` rather than `save`. Means the file can be read in with any filename, not always having to be read in with the name of the object you saved.

modules/nf-core/deseq2/differential/templates/deseq_de.r
```diff
@@ -0,0 +1,354 @@
...
+    file = 'deseq2.sizefactors.tsv',
+    col.names = TRUE,
+    row.names = FALSE,
+    sep = '\t',
+    quote = FALSE
+)
+
+# Write specified matrices
+
+if (opt\$write_normalised){
```
SPPearce, line 297: I don't see any point in making these optional. Just write the files.

modules/nf-core/deseq2/differential/templates/deseq_de.r
```diff
@@ -0,0 +1,354 @@
...
+        data.frame(gene_id=rownames(counts(dds)), counts(dds, normalized = TRUE)),
+        file = 'normalised_counts.tsv',
+        col.names = TRUE,
+        row.names = FALSE,
+        sep = '\t',
+        quote = FALSE
+    )
+}
+
+if (opt\$write_variance_stabilised){
```
SPPearce, line 308: I would also just export these two methods as two separate files, not ask.

modules/nf-core/deseq2/differential/templates/deseq_de.r
```diff
@@ -0,0 +1,354 @@
...
+################################################
+## Generate outputs                           ##
+################################################
+################################################
+
+# Common function to round numerics in data frames - mostly useful for testing,
+# since it doesn't seem to be possible to make outputs entirely reproducible
+# across machienes, even with set.seed(), but the differences are tiny.
+
+prepare_results <- function(x){
```
SPPearce, line 244: I would just round to 8dp whatever, don't need more than that.

modules/nf-core/deseq2/differential/templates/deseq_de.r
```diff
@@ -0,0 +1,354 @@
...
+    if (is.null(opt[[file_input]])) {
+        stop(paste("Please provide", file_input), call. = FALSE)
+    }
+
+    if (! file.exists(opt[[file_input]])){
+        stop(paste0('Value of ', file_input, ': ', opt[[file_input]], ' is not a valid file'))
+    }
+}
+
+if (opt\$random_seed > 0){
```
SPPearce, line 85: Does the set.seed actually make a difference? I didn't think DESeq used any randomness. Is this just a hangover from the precision issues?

SPPearce (commented):
Made some comments as promised. 
Does this shrink the log2FC values automatically? I can't see it.

modules/nf-core/deseq2/differential/templates/deseq_de.r
pinin4fjords: You're right, and it didn't help, I'll strip it out.

modules/nf-core/deseq2/differential/templates/deseq_de.r
pinin4fjords: That did cross my mind, but I came down on the side of keeping results 'as DESeq2 intended' except when we needed the rounding for testing. But happy to remove in the name of simplicity.

modules/nf-core/deseq2/differential/templates/deseq_de.r
pinin4fjords: Ok, will do!

modules/nf-core/deseq2/differential/templates/deseq_de.r
pinin4fjords: Yep, sounds good, I agree (think I only used `save()` because I saw it in use elsewhere in nf-core)

modules/nf-core/deseq2/differential/meta.yml
pinin4fjords: Thanks for the catch!

modules/nf-core/deseq2/differential/main.nf
pinin4fjords: I agree. I was just matching usage elsewhere: https://github.com/nf-core/rnaseq/blob/master/bin/deseq2_qc.r

Will change.

modules/nf-core/deseq2/differential/main.nf
pinin4fjords: Haha, I actually had prefixes at one point derived from concatenating the contrast meta. I removed it in favour of a `publishDir` doing the right thing from the workflow, but I can reinstate the prefix.

pinin4fjords:
> Does this shrink the log2FC values automatically? I can't see it.
I'm pretty sure it happens in `DESeq2()` for calculating the p values etc, but I haven't applied it to the fold changes in the output table (like [this](http://bioconductor.org/packages/devel/bioc/vignettes/DESeq2/inst/doc/DESeq2.html#log-fold-change-shrinkage-for-visualization-and-ranking)?). 

If you'd like that to happen should it be done all the time, or on request @SPPearce ?

modules/nf-core/deseq2/differential/templates/deseq_de.R
```diff
@@ -0,0 +1,338 @@
...
+write.table(
+    data.frame(gene_id=rownames(counts(dds)), counts(dds, normalized = TRUE)),
+    file = paste(output_prefix, 'normalised_counts.tsv', sep = '.'),
+    col.names = TRUE,
+    row.names = FALSE,
+    sep = '\t',
+    quote = FALSE
+)
+
+if (opt\$vs_method == 'vst'){
```
SPPearce, line 291: I would just export both vst counts and rlog counts.

SPPearce:
> > Does this shrink the log2FC values automatically? I can't see it.
> > I'm pretty sure it happens in `DESeq2()` for calculating the p values etc, but I haven't applied it to the fold changes in the output table (like [this](http://bioconductor.org/packages/devel/bioc/vignettes/DESeq2/inst/doc/DESeq2.html#log-fold-change-shrinkage-for-visualization-and-ranking)?).
> 
> If you'd like that to happen should it be done all the time, or on request @SPPearce ?

Mmm, I don't know. I'm not an expert in RNA-seq at all, mostly I work on DNA. I just know that it is something that I've done in the few times I've done any.

modules/nf-core/deseq2/differential/templates/deseq_de.R
pinin4fjords: Seems like we're creating a lot of excess files already, no?

That's why I had switches for the expression matrix output previously, a lot of the time people probably only care about the d/e results, and could keep the disk space burden small.

pinin4fjords:
> > > Does this shrink the log2FC values automatically? I can't see it.
> > > I'm pretty sure it happens in `DESeq2()` for calculating the p values etc, but I haven't applied it to the fold changes in the output table (like [this](http://bioconductor.org/packages/devel/bioc/vignettes/DESeq2/inst/doc/DESeq2.html#log-fold-change-shrinkage-for-visualization-and-ranking)?).
> > 
> > 
> > If you'd like that to happen should it be done all the time, or on request @SPPearce ?
> 
> Mmm, I don't know. I'm not an expert in RNA-seq at all, mostly I work on DNA. I just know that it is something that I've done in the few times I've done any. I would ask people with more experience than me of using DESeq2.

Sounds like something that should be controllable behaviour, anyway, whatever the default behaviour.

modules/nf-core/deseq2/differential/templates/deseq_de.R
SPPearce: Eh, they aren't large files in the grand scheme of things. They could be not published via `publishDir` option if people don't want them.

pinin4fjords: Actually, I was reminded by someone yesterday that the `rlog` func can be quite time consuming to run. So I really think we should keep that off by default, but I'm adding an option to run both.

pinin4fjords:
> Made some comments as promised. Does this shrink the log2FC values automatically? I can't see it.

I'm going to revise my earlier response with this from the DESeq docs:

"An option in DESeq2 is to provide maximum a posteriori estimates of the log2 fold changes in βi after incorporating a zero-centered Normal prior (betaPrior). While previously, these moderated, or shrunken, estimates were generated by DESeq or nbinomWaldTest functions, they are now produced by the lfcShrink function"

So I need to look into how to run the shrinkage effectively, I've a nasty feeling it's going to need something not in the DESeq2 biocontainer.

pinin4fjords:
(note: singularity failure is just because we're waiting for a build to propagate from biocontainers)

modules/nf-core/deseq2/differential/meta.yml
```diff
@@ -0,0 +1,64 @@
...
+        CSV-format sample sheet with sample metadata
+  - counts:
+      type: file
+      description: |
+        Raw TSV-format expression matrix as output from the nf-core RNA-seq workflow
+  - contrast_meta:
+      type: map
+      description: |
+        Groovy Map containing contrast information
+        e.g. [ variable:'treatmtent', reference:'treated', control:'saline', blocking:'' ]
```
WackerO, line 31:
```suggestion
        e.g. [ variable:'treatment', reference:'treated', control:'saline', blocking:'' ]
```

ggabernet (approved):
Looks good to me! As mentioned in slack there are some features missing that we would like to include but it's good to start with a minimal POC!

pinin4fjords:
Thanks @ggabernet! I'm just going to switch the inputs to be keyed by the meta- as per feedback in https://github.com/nf-core/modules/pull/2321, will do that quickly now.

pinin4fjords:
> One thing that could be done to also test the blocking variables and that the execution with multiple contrasts works fine would be to add another contrast in the example data with a blocking variable provided.

@ggabernet Good point- modifications to the test data here: https://github.com/nf-core/test-datasets/pull/670

ggabernet:
Actually this would be important only in the case that the files are initially provided to the pipeline with a samplesheet (e.g. multiple fastq files to analyze, each of them with individual sample metadata that is passed to the `meta` map). 
In the case of the differentialabundance pipeline, we expect to have just 1 count table with all samples to analyze. The samplesheet containing info about the different samples will be provided  to the module input. Therefore I would argue that we will not need the `meta` map here.

In the case of the other module that you mentioned #2321 that operates on reference data (e.g. reference `GTF` or `fasta`) and not on sample (sequencing) files, the meta maps would not be needed either.

pinin4fjords:
Thanks @ggabernet (and apologies I just saw this). I was confused in the other review - it seemed like the meta thing was pretty non-negotiable (though as it turned out I think there was a use for it there).  

In the latest commits I've put the inputs in a single tuple rather than a separate channel for the meta, which I think is nicer, but happy to reverse that if you like.

ggabernet (approved):
ok sure, looks good to me like this too!

pinin4fjords:
Thanks for the review @ggabernet !

PR #2176, opened 2022-10-07
[WIP] updating the sourmash sketch module

taylorreiter:
This PR updates the sourmash sketch module. 

In `main.nf`:
+ upgrades to the latest version of sourmash (4.5.0)
+ updates the params string to the standard set of parameters typically used by sourmash (including to sketch all of GenBank and the SRA). The test data should stay the same, and the way the param string is parameterized hasn't changed so can still be updated to the user if they wish to not follow this convention.

In `meta.yml`:
+ updates the doi to the correct citation (currently listed as mash, which is a different tool than sourmash)
+ updates the url to the correct sourmash github repo
+ updates the description from MinHash to FracMinHash
+ updates the types of data ingested to include protein sequences, transcriptomes, and fastq files

## PR checklist

- [ ] Add a resource `label`

This PR is a WIP because I'm not sure how to test if tests are passing locally, and I don't know if I followed the parameters requirement, I don't know what adding a resource label means, and I haven't tested docker/singularity


modules/nf-core/sourmash/sketch/meta.yml
```diff
@@ -20,8 +20,8 @@ input:
         e.g. [ id:'test', single_end:false ]
   - sequence:
       type: file
-      description: FastA file containing (genomic) sequence data
-      pattern: "*.{fna,fa,fasta}"
+      description: FASTA or FASTQ file containing (genomic, transcriptomic, or protein) sequence data
+      pattern: "*.{fna,fa,fasta,fastq,fq}"
```
taylorreiter, line 24: should I add `*gz` to these file format endings as well?

modules/nf-core/sourmash/sketch/main.nf
```diff
@@ -18,7 +18,7 @@ process SOURMASH_SKETCH {
     task.ext.when == null || task.ext.when
 
     script:
-    def args = task.ext.args ?: "dna --param-string 'scaled=1000,k=31'"
+    def args = task.ext.args ?: "dna --param-string 'scaled=1000,k=21,k=31,k=51,abund'"
```
Midnighter, line 21: Please note that this should be the absolute default arguments for this kind of process. Any more specialized application is expected to override `task.ext.args`. Are you sure these should be the default ones?

modules/nf-core/sourmash/sketch/meta.yml
Midnighter: Yes, absolutely, nf-core actually prefers compressed file formats so maybe drop the uncompressed ones. Should also add `.faa` for amino acid sequences then?

modules/nf-core/sourmash/sketch/meta.yml
```diff
@@ -20,8 +20,8 @@ input:
         e.g. [ id:'test', single_end:false ]
   - sequence:
       type: file
-      description: FastA file containing (genomic) sequence data
-      pattern: "*.{fna,fa,fasta}"
+      description: FASTA or FASTQ file containing (genomic, transcriptomic, or protein) sequence data
```
Midnighter, line 23: Maybe
```suggestion
      description: FASTA or FASTQ file containing (genomic, transcriptomic, or proteomic) sequence data
```

Midnighter (commented):
Update looks good to me. I have a slight doubt about the default arguments but everything else is 👍🏼

modules/nf-core/sourmash/sketch/meta.yml
taylorreiter: ooh love this

modules/nf-core/sourmash/sketch/meta.yml
taylorreiter: so maybe something like this?
```suggestion
      pattern: "*.{fna,fa,fasta,fastq,fq,faa}.gz"
```

modules/nf-core/sourmash/sketch/main.nf
ctb: yep.

(Except for `abund`, these are the default arguments inside of sourmash for `sketch dna` as well, but I think it's good to have them explicit here.)

taylorreiter: I checked in with other sourmash devs @ctb and @bluegenes and they agree that the default should be `scaled=1000,k=21,k=31,k=51,abund`. The idea is that signatures should be a unit that is kept together in a single file, and that you should only read the data once to get all of the sketches.

taylorreiter: oh woops sorry i didn't reload the page before I posted this to see that titus had responded! 😂

tests/modules/nf-core/sourmash/sketch/test.yml
```diff
@@ -5,6 +5,6 @@
     - sourmash/sketch
   files:
     - path: output/sourmash/test.sig
-      md5sum: 4b0dee307e35fc670cd0d416321e4961
+      md5sum: c23a5a251d2b6babba8ae0e3905e8d04
     - path: output/sourmash/versions.yml
```
nvnieuwk, line 9:
```suggestion
```

Checks for versions.yml now happen automatically

nvnieuwk: Could you also remove the `md5sum`? Github doesn't allow me to suggest a change there

modules/nf-core/sourmash/sketch/main.nf
nvnieuwk: Isn't it a bit weird that you should manually define the arguments that are seen as default? (just curious why it is this way :p)

nvnieuwk (changes requested):

modules/nf-core/sourmash/sketch/meta.yml
Midnighter: Neat, yes 🙂

modules/nf-core/sourmash/sketch/main.nf
Midnighter: Okay, just wanted to make sure. Thanks for the response @taylorreiter and @ctb 

@nvnieuwk that depends on your definition of manual. These modules are expected to be included in a variety of pipelines and it is in those pipeline configurations where the expected arguments should be set for the particular application of this module within the pipeline.

nvnieuwk: Aaah okay that makes sense! Thanks for clarifying!

Midnighter (approved):
Thanks for the update. I didn't know about the automatic test for `versions.yml` either. So just a little change and this is good to go.

modules/nf-core/sourmash/sketch/main.nf
jfy133: I had the same  question, @taylorreiter could you add a comment just above the line specify this (otherwise this might come up again, embedding `args` is very unusual/normally frowned upon).

ctb: (not familiar with nf-core policies, so take this as you will ;)

on reflection, since sourmash uses semantic versioning, `scaled=1000,k=21,k=31,k=51` are default and only `-p abund` is not, you might want to omit the first part - as long as this workflow is pinned to sourmash<5, you'll be good.

on the other hand it could be a nice matter of user education since these parameters are among the most likely ones to be tweaked by the user.

nvnieuwk (approved):

modules/nf-core/sourmash/sketch/main.nf
jfy133, line 21:
```suggestion
    // required defaults for the tool to run, but can be overridden
    def args = task.ext.args ?: "dna --param-string 'scaled=1000,k=21,k=31,k=51,abund'"
```

modules/nf-core/sourmash/sketch/meta.yml
```diff
@@ -1,15 +1,15 @@
 name: sourmash_sketch
-description: Create a signature (a hash sketch) of a sequence using sourmash
+description: Create a signature (a group of FracMinHash sketches) of a sequence using sourmash
 keywords:
   - hash sketch
   - signature
```
jfy133, line 5:
```suggestion
  - sourmash
  - genomics
  - metagenomics
  - taxonomic classification
  - taxonomic profiling
  - kmer
```

jfy133 (approved):

modules/nf-core/sourmash/sketch/meta.yml
jfy133: to sync with https://github.com/nf-core/modules/pull/2250/files, and make it more findable on https://nf-co.re/modules

PR #2199, opened 2022-10-10
Prototype `nf-test`

nvnieuwk:
Currently contains:
- A prototype of a simple process (rtgtools/pedfilter) => uses the process testing
- A prototype of a module that depends on other modules for its input data (multivcfanalyzer) => uses the workflow testing
   - To discuss: Define the input in the workflow itself or in the test file?
- A prototype of a workflow that needs multiple test cases (aka workflows) (multiqc) => uses the workflow testing
- A prototype where a specific config file is supplied (mostly helpful to define the `args`) (gatk4/printsvevidence) => uses workflow testing

IMO the last prototype seems the most suitable to replace all tests with. The developer also noted they will be adding `tags` and `snapshot testing` (https://github.com/askimed/nf-test/issues/29). `snapshot testing` could be a nice replacement for the current `yaml` implementation

Things that still need to be figured out or worked on:
- How to set up the tests in GitHub Actions => waiting for the `tags` feature
- How to let `nf-core/tools` generate test files => waiting the `yaml` or `snapshot testing` feature


nvnieuwk:
You need `nf-test` version 0.6.1 to run the code in this PR

PR #2204, opened 2022-10-10
Falco

### LINKED ISSUE #2203: new module: falco

lucacozzuto:
## PR checklist

Closes #2203 

- [ ] Add a resource `label`


lucacozzuto:
@nf-core-bot fix linting

lucacozzuto:
@nf-core-bot fix linting

modules/nf-core/falco/main.nf
```diff
@@ -0,0 +1,59 @@
...
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    if (meta.single_end) {
+        """
+        [ ! -f  ${prefix}.fastq.gz ] && ln -s $reads ${prefix}.fastq.gz
```
SPPearce, line 27: What is this doing?

tests/modules/nf-core/falco/test.yml
```diff
@@ -0,0 +1,29 @@
+- name: falco test_fastqc_single_end
+  command: nextflow run ./tests/modules/nf-core/falco -entry test_fastqc_single_end -c ./tests/config/nextflow.config -c ./tests/modules/nf-core/falco/nextflow.config
+  tags:
+    - falco
+  files:
+    - path: output/falco/fastqc_data.txt
+      md5sum: c5d0cd529d1d2f5a842bf3afd06a9567
+    - path: output/falco/fastqc_report.html
+      contains: "[ # TODO nf-core: file md5sum was variable, please replace this text with a string found in the file instead ]"
```
SPPearce, line 9: These need fixing (and below)

yuukiiwa: One can either remove this and keep the `md5sum` if the `md5sum` does not differ every time running the module

tests/modules/nf-core/falco/test.yml
```diff
@@ -0,0 +1,29 @@
...
+
+- name: falco test_fastqc_paired_end
+  command: nextflow run ./tests/modules/nf-core/falco -entry test_fastqc_paired_end -c ./tests/config/nextflow.config -c ./tests/modules/nf-core/falco/nextflow.config
+  tags:
+    - falco
+  files:
+    - path: output/falco/test_1.fastq.gz_fastqc_data.txt
+      md5sum: 36d989bb9e2d5a632e19452f4e6c2a4e
+    - path: output/falco/test_1.fastq.gz_fastqc_report.html
+      contains: "[ # TODO nf-core: file md5sum was variable, please replace this text with a string found in the file instead ]"
```
yuukiiwa, line 21: One can either remove this and keep the `md5sum` if the `md5sum` does not differ every time running the module

tests/modules/nf-core/falco/test.yml
```diff
@@ -0,0 +1,29 @@
...
+    - path: output/falco/test_1.fastq.gz_summary.txt
+      md5sum: a925aec214a83d2f6252847166f2ef3a
+    - path: output/falco/test_2.fastq.gz_fastqc_data.txt
+      md5sum: ad5c45dfc8f79754dd5d8029456b715b
+    - path: output/falco/test_2.fastq.gz_fastqc_report.html
+      contains: "[ # TODO nf-core: file md5sum was variable, please replace this text with a string found in the file instead ]"
```
yuukiiwa, line 27: One can either remove this and keep the `md5sum` if the `md5sum` does not differ every time running the module

modules/nf-core/falco/main.nf
```diff
@@ -0,0 +1,59 @@
...
+            falco: \$( falco --version | sed -e "s/falco v//g" )
+        END_VERSIONS
+        """
+    }
+
+    stub:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    touch ${prefix}.html
+    touch ${prefix}_fastqc_data.txt
```
SPPearce, line 52:
```suggestion
    touch ${prefix}_fastqc_data.txt
```

modules/nf-core/falco/main.nf
```diff
@@ -0,0 +1,59 @@
...
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    if (meta.single_end) {
+        """
+        [ ! -f  ${prefix}.fastq.gz ] && ln -s $reads ${prefix}.fastq.gz
+        falco $args --threads $task.cpus ${prefix}.fastq.gz
```
SPPearce, line 28: I don't understand why you need to rename the input files, can the tool not just take them directly like:
```suggestion
        falco $args --threads $task.cpus ${reads}
```

SPPearce (changes requested):
Tests need fixing and I'm confused about the linking.
The rest looks good though.

modules/nf-core/falco/main.nf
lucacozzuto: Hi, this is copy pasted by fastqc module. Since Falco is a C replacement of fastQC I just followed the same code

modules/nf-core/falco/main.nf
lucacozzuto: Again, I followed what is written in fastQC. We can change both actually

lucacozzuto: I suspect this is to help multiqc to make meaningful names... however again we should ask to people that make the fastQC module

SPPearce: Ah, yes, that may well be the reason.

SPPearce (approved):

SPPearce (changes requested):
Tests still failing (unapprove until then)

tests/modules/nf-core/falco/main.nf
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env nextflow
+
+nextflow.enable.dsl = 2
+
+include { FALCO } from '../../../../modules/nf-core/falco/main.nf'
+
+//
+// Test with single-end data
+//
+workflow test_fastqc_single_end {
```
matthdsm, line 10: Nitpicking, but shouldn't the test names be changed to `falco`

modules/nf-core/falco/main.nf
matthdsm: Can you check all of this linking etc. actually makes sense? It seems to me like `falco` can take an arbitrary number of sequence files to parse. I get that it's copy paste from `fastqc`, but that doesn't mean we can't make it better 😉

modules/nf-core/falco/main.nf
```diff
@@ -0,0 +1,59 @@
+process FALCO {
+    tag "$meta.id"
+    label 'process_medium'
```
matthdsm, line 3: Should be `process_single` since multithreading is not supported by `falco`

modules/nf-core/falco/main.nf
matthdsm: Nevermind, the linking and separation of SE and PE is done to ensure consistent naming across runs

modules/nf-core/falco/main.nf
```diff
@@ -0,0 +1,59 @@
...
+    tuple val(meta), path("*.txt") , emit: txt
+    path  "versions.yml"           , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    if (meta.single_end) {
```
matthdsm, line 25: Here I would check the size of the `reads` input. If it's more than one file, it should go to the second case, regardless of it being PE or SE. Endedness has no effect on the tool

matthdsm (changes requested):

modules/nf-core/falco/main.nf
matthdsm, line 3:
```suggestion
    label 'process_single'
```

modules/nf-core/falco/main.nf
lucacozzuto: true, but why you want to analyze 30 files at once? there is no multi-threading here

matthdsm: To each it's own I guess 🤷🏻  I don't see why anyone would, but we can't force people to think like we do.
Either way, it's more correct, since the current situation would break support for interleaved files for example

lucacozzuto: aha. So which change do you propose?

tests/modules/nf-core/falco/main.nf
lucacozzuto: Done

modules/nf-core/falco/main.nf
lucacozzuto: Ok, should it be in single end so? Another thing, Falco with one file will produce a standard output name. So you need to enforce the output name using -R, -D and -S. If you use more files those options will trigger an error. So this is the reason I put them only in single end. Should we indicate another meta info? Something that is not single end or paired end?

matthdsm: SE or PE don't matter for the tool. What you're doing with the settings is good. If there's one file, you need to set the naming options. If there's more than one file, you don't. Hence the check for the number of files instead of the endedness

lucacozzuto: But since this is a replacement for fastqc this will break the compatibility. People should be able to replace fastqc in the existing pipelines without changing the meta etc. Or you want I just ignore the meta and count the files? Not so sure is a good idea actually. Can we discuss on slack with more people, please? :)

matthdsm: I don't think this will break anything, since fastq's also doesn't take endedness into account (I think).
Lets take this to slack!

modules/nf-core/falco/main.nf
```diff
@@ -48,8 +45,9 @@ process FALCO {
     stub:
     def prefix = task.ext.prefix ?: "${meta.id}"
     """
-    touch ${prefix}.html
-    touch ${prefix}_fastqc_data.txt
+    touch ${reads}_fastqc_data.html
```
matthdsm, line 48: This needs to stay as `prefix` in the stub I think

matthdsm (changes requested):

modules/nf-core/falco/main.nf
lucacozzuto: yes

matthdsm (approved):

mashehu (approved):

PR #1978, opened 2022-08-28
New module - BBMap clumpify 

tamuanand:
## PR checklist
- [x] This comment contains a description of changes (with reason).
- [x] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [x] Remove all TODO statements.
- [x] Emit the `versions.yml` file.
- [x] Follow the naming conventions.
- [x] Follow the parameters requirements.
- [x] Follow the input/output options guidelines.
- [x] Add a resource `label`
- [x] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - [x] `PROFILE=docker pytest --tag bbmap/clumpify --symlink --keep-workflow-wd --git-aware`
  - [ ] `PROFILE=singularity pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [x] `PROFILE=conda pytest --tag bbmap/clumpify --symlink --keep-workflow-wd --git-aware`

PR #2206, opened 2022-10-10
Bbmap filterbyname

tokarevvasily:
## PR checklist

Closes #XXX 

- [ x ] This comment contains a description of changes (with reason).
- [ x ] If you've fixed a bug or added code that should be tested, add tests!
- [ x ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ x ] If necessary, include test data in your PR.
- [ x ] Remove all TODO statements.
- [ x ] Emit the `versions.yml` file.
- [ x ] Follow the naming conventions.
- [ x ] Follow the parameters requirements.
- [ x ] Follow the input/output options guidelines.
- [ x ] Add a resource `label`
- [ x ] Use BioConda and BioContainers if possible to fulfil software requirements.
  - [ x ] `PROFILE=docker pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [ x ] `PROFILE=singularity pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [ x ] `PROFILE=conda pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`


tokarevvasily:
@nf-core-bot fix linting

modules/nf-core/bbmap/filterbyname/main.nf
```diff
@@ -0,0 +1,53 @@
...
+    label 'process_single'
+
+    conda (params.enable_conda ? "bioconda::bbmap=39.00" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/bbmap:39.00--h5c4e2a8_0':
+        'quay.io/biocontainers/bbmap:39.00--h5c4e2a8_0' }"
+
+    input:
+    tuple val(meta), path(reads)
+    val output_extension
```
SPPearce, line 12: This could be set in the modules.config, like [samtools view ](https://github.com/nf-core/modules/blob/master/modules/nf-core/samtools/view/main.nf) does.

SPPearce (changes requested):
I would rather have the output extension in the config file than passed as a parameter.
The rest looks fine to me.

modules/nf-core/bbmap/filterbyname/main.nf
tokarevvasily: I do not think filterbyname.sh offers an option to pass a file extension as a flag (samtools view has `--output-fmt`). I could do all manipulations with it and then remove it from `args`, but I think it just overcomplicates things. What do you think?

Here is what I would have to do:
```
// if `output-fmt` was passed -> use it, otherwise treat everything after the
// first comma as prefered `output-fmt`
def file_type = args.toLowerCase().contains("--output-fmt fasta") ? "fasta" :
                    args.toLowerCase().contains("--output-fmt fa") ? "fa" :
                    args.toLowerCase().contains("--output-fmt fastq") ? "fastq" :
                    args.toLowerCase().contains("--output-fmt fq") ? "fq" :
                    meta.single_end ? reads.name.tokenize(".")[1..-1].join(".") :
                    reads[0].name.tokenize(".")[1..-1].join(".")

args = args.replaceAll("--output-fmt \\w+ ?", "")
```

nvnieuwk: Does the tool define the output format by this extension on the output file? Or is the output the same as the input?

nvnieuwk: Then I think using a dictionary `args` would be the best choice, see here for examples:
[nextflow.config](https://github.com/nf-core/modules/blob/2edb02c68b0d2e3b31dca6360f4a683d5b0f2970/tests/modules/nf-core/ascat/nextflow.config#L6-L15)
[main.nf](https://github.com/nf-core/modules/blob/2edb02c68b0d2e3b31dca6360f4a683d5b0f2970/modules/nf-core/ascat/main.nf#L34-L50)

Also don't forget to specify that the user should use it like this inside of the meta.yml, like [this](https://github.com/nf-core/modules/blob/2edb02c68b0d2e3b31dca6360f4a683d5b0f2970/modules/nf-core/ascat/meta.yml#L17-L40)

tokarevvasily: How would it work with passing `args` to the command itself (line 43)? Do you want me to manually write out every single CLI option available to solve that?

nvnieuwk: No, just the extension. You could add `args` again inside of `args` and use that for the CLI arguments :). (I should have been more clear on that sorry)

tokarevvasily: Okay, I am almost there. Here is what I have now:
```
def args               = task.ext.args   ?: [ args:"" ]
def prefix             = task.ext.prefix ?: "${meta.id}"
def output_extension   = args.output_extension ? "$args.output_extension" :
    meta.single_end ? reads.name.tokenize(".")[1..-1].join(".") :
    reads[0].name.tokenize(".")[1..-1].join(".")
def interleaved_output = true ? "$args.interleaved_output" : false

def input            = meta.single_end ? "in=${reads}" : "in=${reads[0]} in2=${reads[1]}"
def filtered         = meta.single_end || interleaved_output ? "out=${prefix}.${output_extension}" :
    "out1=${prefix}_1.${output_extension} out2=${prefix}_2.${output_extension}"

```

What should I put for output `tuple val(meta), path("*.${output_extension}"), emit: reads`? Because user can specify any extension and leaving it as is does not work (output_extension gets defined later)

nvnieuwk: That should work if you remove the `def` in front of `output_extension` (not 100% sure on this, but I remember it like this)

tokarevvasily: Removing `def` gives me a `Module compilation error` :(

I have tried defining `output_extension` outside the `script` block and changing `args.output_extension` to `ext.args.output_extension`, but from what I understand, config is not read in yet at that point, so I am getting `null` as a value. I was also toying with an idea of grabbing anything that is not `versions.yml` or `*.log` with regex, but group negation within a glob is not a thing from what I have learned (only character negation through `[!<SOME-CHARACTER(s)>]` is available).

I'll poke around more, but I welcome any ideas, unless we want to just stick with `output_extension` being one of the inputs :)

nvnieuwk: Like they do [here](https://github.com/nf-core/modules/blob/202683bfc98ddecdd456ea73268e330bca2e5c5a/modules/nf-core/samtools/convert/main.nf#L16)

tokarevvasily: I have tried that, but I do not think typing out every single possible extension written with different capitalizations is the best solution (capitalization is important because if `output_extension` is not passed, it uses input's extension over which I have no control).

tokarevvasily: I have two branches reads:
[One that uses `args` ](https://github.com/tokarevvasily/nf-core-modules/tree/bbmap-filterbyname-config). It has a bunch of predefined outputs, but I foresee issues if user does not specify output extension that they want and input's extension does not match any of the ones that I pass.

Here is [another branch that uses `input val`](https://github.com/tokarevvasily/nf-core-modules/tree/bbmap-filterbyname-vals). I like this approach significantly better and there is already a similar use case in [fastp](https://github.com/nf-core/modules/blob/202683bfc98ddecdd456ea73268e330bca2e5c5a/modules/nf-core/fastp/main.nf#L12), where `val save_trimmed_fail` is not directly used as an input for the tool, but rather controls how it runs.

For both I have added an option for interleaved output when using paired-end data!
Let me know which one you want me to push here!

nvnieuwk: I found a solution for the earlier problem with using `output_extension` in the output definition => see how they used `prefix` [here](https://github.com/nf-core/modules/blob/master/modules/nf-core/krakenuniq/build/main.nf)



> [One that uses `args` ](https://github.com/tokarevvasily/nf-core-modules/tree/bbmap-filterbyname-config). It has a bunch of predefined outputs, but I foresee issues if user does not specify output extension that they want and input's extension does not match any of the ones that I pass.

This is the one that follows the guidelines most closely. So I would suggest this one.



> Here is [another branch that uses `input val`](https://github.com/tokarevvasily/nf-core-modules/tree/bbmap-filterbyname-vals). I like this approach significantly better and there is already a similar use case in [fastp](https://github.com/nf-core/modules/blob/202683bfc98ddecdd456ea73268e330bca2e5c5a/modules/nf-core/fastp/main.nf#L12), where `val save_trimmed_fail` is not directly used as an input for the tool, but rather controls how it runs.

For this, an exception can be made, but that's not something I can decide by myself. Can you ping `@modules-team` in Slack and ask for some more opinions on this?

nvnieuwk: (I'll post a link to this in Slack)

mahesh-panchal: Is there a reason not to just stick with `fasta.gz` as the file suffix? Output should always be compressed.

tokarevvasily: @nvnieuwk I have tried this before and that's where I was getting `Module compilation error`. However, since your example clearly worked, I did some additional digging and realized that once I define a variable without `def` nothing after it can have `def`. I made appropriate changes, and everything worked! I am not sure if this behavior is expected, but it is definitely new to me.

As a note, I have also updated BBmap to the most recent version, however, singularity container is not yet on `https://depot.galaxyproject.org/`, but the new version is on `conda` and `quay.io/biocontainers`. As a result, tests with singularity will fail until new image gets uploaded.

nvnieuwk: Awesome, I didn't know that myself :)

The singularity container should be there soon (it usually takes a bit longer than the docker containers)

modules/nf-core/bbmap/filterbyname/main.nf
```diff
@@ -0,0 +1,59 @@
...
+    label 'process_single'
+
+    conda (params.enable_conda ? "bioconda::bbmap=39.01" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/bbmap:39.01--h311275f_0':
+        'quay.io/biocontainers/bbmap:39.01--h311275f_0' }"
+
+    input:
+    tuple val(meta), path(reads)
+    val names
```
nvnieuwk, line 12: It is also advised to specify the names like you did with `output_extension`

nvnieuwk (changes requested):
Almost there! Can you also add a test where you use another output extension?

modules/nf-core/bbmap/filterbyname/main.nf
tokarevvasily: Addressed in the [next commit](https://github.com/nf-core/modules/pull/2206/commits/247b6459f070daf2f02f36a4219d2406c43afa6c)

modules/nf-core/bbmap/filterbyname/main.nf
```diff
@@ -0,0 +1,59 @@
...
+        'https://depot.galaxyproject.org/singularity/bbmap:39.01--h311275f_0':
+        'quay.io/biocontainers/bbmap:39.01--h311275f_0' }"
+
+    input:
+    tuple val(meta), path(reads)
+
+    output:
+    tuple val(meta), path("*.$output_extension"), emit: reads
+    tuple val(meta), path('*.log')                , emit: log
+    path "versions.yml"                           , emit: versions
```
FriederikeHanssen, line 16:
```suggestion
    tuple val(meta), path("*.$output_extension"), emit: reads
    tuple val(meta), path('*.log')              , emit: log
    path "versions.yml"                         , emit: versions
```

modules/nf-core/bbmap/filterbyname/main.nf
```diff
@@ -0,0 +1,59 @@
...
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args               = task.ext.args   ?: [ args:"" ]
+    args.args = args.args ? "$args.args" : ""
```
FriederikeHanssen, line 23:
```suggestion
    def args               = task.ext.args   ?: ""
```

FriederikeHanssen: if args doesn't exists it will be evaluated anyways to an empty string, no?

modules/nf-core/bbmap/filterbyname/main.nf
```diff
@@ -0,0 +1,59 @@
...
+        "in=${reads[0]} in2=${reads[1]}"
+
+    def avail_mem = 3
+    if (!task.memory) {
+        log.info '[filterbyname] Available memory not known - defaulting to 3GB. Specify process memory requirements to change this.'
+    } else {
+        avail_mem = task.memory.giga
+    }
+
+    output_extension   = args.output_extension ? "$args.output_extension" :
```
FriederikeHanssen, line 37: For the ouput, this can be handled via the modules.config and passed in via the prefix, for example how we deal with bam input but cram output in GATK4 Markduplicates

FriederikeHanssen: either prefix can be evaluated with a closure for this or it can be fixed by the pipeline developer

FriederikeHanssen (dismissed):
I am not convinced a args map is absolutely necessary here. The otput extension could be handled via 

`prefix = ${meta.id}.fastq.gz` (or whatever he user would like to have, this can be defaulted in the module)

modules/nf-core/bbmap/filterbyname/main.nf
nvnieuwk: the arguments itself are defined as `args.args` (that's why it's parsed that way)

modules/nf-core/bbmap/filterbyname/main.nf
nvnieuwk: That's smart!

tokarevvasily: @FriederikeHanssen I am not following how this would solve an issue of "unknown output format".

tokarevvasily: Also, the way I interpret `prefix` - it is everything before extension. Controlling file extension through `prefix` is confusing for me.

tokarevvasily:
@nf-core-bot fix linting

tokarevvasily:
Not sure why singularity test fails. It works for me locally, and both conda and docker tests pass here. Input is welcome!

mahesh-panchal:
I'm not sure this is fixable. BBmap is producing some output that looks very yaml like and it's fooling the test parser into thinking it's yaml and throwing an error. I think this output is also singularity container specific, and I don't know why. 

@fabianegli @mirpedrol Do either of you know a way to get around this? 

We came across this error here too: https://github.com/nf-core/modules/pull/1978#issuecomment-1231548458, but we just ignored it there.

mirpedrol:
It seems that the problem is when obtaining the version, which ends up being `[0.001s][warning][os,cont ...` instead of the version number. Is there a different way to get the bbmap version? I can't find what `bbversion.sh` is/does.

tokarevvasily:
bbmap consists of a bunch of functions one of which is `bbversion.sh`. It print bbmap's version like so:
```
$ bbversion.sh 
39.01
```

Here is a [link](https://github.com/BioInfoTools/BBMap/blob/master/sh/bbversion.sh) to the code itself. It is an older version, but the code has not changed since then.

This same method is used in all bbmap/<subtools> on nf-core/modules.

mahesh-panchal:
The problem is to do with the Biocontainers singularity container. When you use that command inside, some extra stuff gets printed too. We can probably use `tail` or `sed` to make sure only the relevant stuff is printed. Our cluster is going through an Singularity to Apptainer update today, otherwise I would have checked for a solution.

SPPearce:
I've had a go at removing the extra stuff with `bbversion.sh | grep -v "Duplicate cpuset"`, will see if that works.

SPPearce (approved):

modules/nf-core/bbmap/filterbyname/main.nf
```diff
@@ -0,0 +1,59 @@
...
+    conda (params.enable_conda ? "bioconda::bbmap=39.01" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/bbmap:39.01--h5c4e2a8_0':
+        'quay.io/biocontainers/bbmap:39.01--h5c4e2a8_0' }"
+
+    input:
+    tuple val(meta), path(reads)
+
+    output:
+    tuple val(meta), path("*.$output_extension"), emit: reads
```
matthdsm, line 14:
```suggestion
    tuple val(meta), path("*.$suffix"), emit: reads
```
We're standardizing on `args.suffix` here

SPPearce:
I have been working on this PR, and almost finished a implementation

SPPearce:
Right, I think this is now ready.
Interesting slice of history here, we have gone full circle on the required inputs, now requiring that the mandatory inputs are value channels rather than passing them all through `ext.args` ;)

modules/nf-core/bbmap/filterbyname/meta.yml
```diff
@@ -0,0 +1,58 @@
...
+      type: file
+      description: The trimmed/modified fastq reads
+      pattern: "*${output_format}"
+  - log:
+      type: file
+      description: filterbyname.sh log file
+      pattern: "*.filterbyname.log"
+
+authors:
+  - "@tokarevvasily"
```
mahesh-panchal, line 58: Perhaps add yourself as author too Simon.

There's also no listed maintainer.

modules/nf-core/bbmap/filterbyname/main.nf
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    """
+    filterbyname.sh \\
+        -Xmx${avail_mem}g \\
+        $input \\
+        $output \\
+        $names_command \\
+        $args \\
+        &> ${prefix}.log
```
mahesh-panchal, line 47: Can we use `tee` for this please so the nextflow log/stdout can also see any logged error messages. https://nf-co.re/docs/guidelines/components/modules#capturing-stdout-and-stderr

modules/nf-core/bbmap/filterbyname/tests/nextflow.config
```diff
@@ -0,0 +1,24 @@
+process {
+    withName: 'test_bbmap_filterbyname_single_end:BBMAP_FILTERBYNAME' {
+        ext.args = [
```
mahesh-panchal, line 3: How is this working? `ext.args` should be a `String`, rather than a `Map`

modules/nf-core/bbmap/filterbyname/main.nf
SPPearce: Hopefully the tee I've just tried to write works.

modules/nf-core/bbmap/filterbyname/tests/nextflow.config
SPPearce: This wasn't being used any more

mahesh-panchal (approved):

PR #2227, opened 2022-10-11
fix: Remove contains with TODO

Emiller88:
Bug report incoming on tools.


Emiller88:
Nvm duplicate of #2226

PR #2250, opened 2022-10-11
Adding module sourmash/gather

vmikk:
## PR checklist

Closes https://github.com/nf-core/modules/issues/2214

- [x] Add a resource `label`
  - [x] `PROFILE=docker pytest --tag sourmash --symlink --keep-workflow-wd --git-aware`
  - [ ] `PROFILE=singularity pytest --tag sourmash --symlink --keep-workflow-wd --git-aware`
  - [x] `PROFILE=conda pytest --tag sourmash --symlink --keep-workflow-wd --git-aware`

There is an error in tests, associated with different module - `nf-core/fastq_align_dna`
Considering output guidelines - output files (signatures and csv are not compressed).


modules/nf-core/sourmash/gather/main.nf
```diff
@@ -0,0 +1,44 @@
...
+    tuple val(meta), path('*.csv')           , optional:true, emit: result
+    tuple val(meta), path('*_unassigned.sig'), optional:true, emit: unassigned
+    tuple val(meta), path('*_matches.sig')   , optional:true, emit: matches
+    path "versions.yml"                      , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: '--ksize 31'
```
jfy133, line 24: I don't think this should be here unless it's completely mandatory for the tool to run... I wonder if it would make sense to have a warning that if nothing supplied, 'here be dragons' style...

modules/nf-core/sourmash/gather/main.nf
```diff
@@ -0,0 +1,44 @@
...
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/sourmash:4.5.0--hdfd78af_0':
+        'quay.io/biocontainers/sourmash:4.5.0--hdfd78af_0' }"
+
+    input:
+    tuple val(meta), path(signature)
+    path(database)
+
+    output:
+    tuple val(meta), path('*.csv')           , optional:true, emit: result
```
jfy133, line 15: Is it correct this is optional? I would assume this would always contain some output? Or can it occur if there are no hits?

modules/nf-core/sourmash/gather/main.nf
```diff
@@ -0,0 +1,44 @@
...
+    tuple val(meta), path('*_matches.sig')   , optional:true, emit: matches
+    path "versions.yml"                      , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: '--ksize 31'
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    result     = "${prefix}.csv"
```
jfy133, line 26: This can be specified directly in the command itself

modules/nf-core/sourmash/gather/main.nf
```diff
@@ -0,0 +1,44 @@
...
+    matches    = "${prefix}_matches.sig"
+    unassigned = "${prefix}_unassigned.sig"
+
```
jfy133, line 29: These flags should be optional. You should make a (boolean) value channel which then inserts them into the command. [^1]

You will also need to add such comands for the `--save-prefetch*` parameters.

[^1]: Example: https://github.com/nf-core/modules/blob/174170689d08dc614316d90f15a73a3b0ba94521/modules/nf-core/fastp/main.nf#L30

modules/nf-core/sourmash/gather/main.nf
```diff
@@ -0,0 +1,44 @@
...
+        'quay.io/biocontainers/sourmash:4.5.0--hdfd78af_0' }"
+
+    input:
+    tuple val(meta), path(signature)
+    path(database)
+
+    output:
+    tuple val(meta), path('*.csv')           , optional:true, emit: result
+    tuple val(meta), path('*_unassigned.sig'), optional:true, emit: unassigned
+    tuple val(meta), path('*_matches.sig')   , optional:true, emit: matches
```
jfy133, line 17: Missing also outputs for `--save-prefetch` [^1].

[^1] https://nf-co.re/docs/contributing/modules#inputoutput-options point 1

jfy133 (changes requested):
Pretty good! Main thing is missing a few optional outputs

modules/nf-core/sourmash/gather/main.nf
vmikk: I think that there could be cases without any hits (depends on a sample and reference db)

modules/nf-core/sourmash/gather/main.nf
vmikk: True, will fix it.

modules/nf-core/sourmash/gather/main.nf
vmikk: Done.

modules/nf-core/sourmash/gather/main.nf
vmikk: Done

modules/nf-core/sourmash/gather/main.nf
vmikk: Done.

modules/nf-core/sourmash/gather/meta.yml
```diff
@@ -0,0 +1,60 @@
+name: "sourmash_gather"
+description: Search the reference database for the most highly contained genomes using sourmash
+keywords:
+  - hash sketch
+  - signature
+  - containment
```
jfy133, line 6:
```suggestion
  - containment
  - sourmash
  - genomics
  - metagenomics
  - taxonomic classification
  - taxonomic profiling
  - kmer
```

modules/nf-core/sourmash/gather/meta.yml
```diff
@@ -0,0 +1,60 @@
...
+      description: |
+        Groovy Map containing sample information
+        e.g. [ id:'test', single_end:false ]
+  - signature:
+      type: file
+      description: File containing signatures (hash sketches) of a sample
+      pattern: "*.{sig}"
+  - db:
+      type: file
+      description: Sourmash database (list of signatures or SBTs)
```
jfy133, line 28: Missing descriptions of new input calues

modules/nf-core/sourmash/gather/meta.yml
```diff
@@ -0,0 +1,60 @@
...
+      description: |
+        Signatures classified as belonging to any of the taxa
+        in the sourmash database.
+      pattern: "*{sig}"
+  - unassigned:
+      type: file
+      description: |
+        Signatures not classified as belonging to any of the taxa
+        in the sourmash database.
+      pattern: "*{sig}"
```
jfy133, line 57: Missing prefetch descriptions etc.

jfy133 (changes requested):
Almost there, mainly missing meta.yml entries for the newly added input/output channels

modules/nf-core/sourmash/gather/meta.yml
vmikk: Keywords updated.

modules/nf-core/sourmash/gather/meta.yml
vmikk: My bad, forgot to add them.
Done now.

modules/nf-core/sourmash/gather/meta.yml
vmikk: Done.

jfy133 (approved):

jfy133:
@nf-core-bot fix linting

modules/nf-core/sourmash/gather/main.nf
```diff
@@ -0,0 +1,53 @@
...
+    label 'process_low'
+
+    conda (params.enable_conda ? "bioconda::sourmash=4.5.0" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/sourmash:4.5.0--hdfd78af_0':
+        'quay.io/biocontainers/sourmash:4.5.0--hdfd78af_0' }"
+
+    input:
+    tuple val(meta), path(signature)
+    path(database)
```
taylorreiter, line 12: sourmash gather allow multiple databases to be passed per gather run. This is the preferred way to run e.g. all of the GenBank databases in one run. does encoding the database path this way allow multiple databases to be passed?

modules/nf-core/sourmash/gather/main.nf
```diff
@@ -0,0 +1,53 @@
...
+    val save_unassigned
+    val save_matches_sig
+    val save_prefetch
+    val save_prefetch_csv
+
+    output:
+    tuple val(meta), path('*.csv')           , optional:true, emit: result
+    tuple val(meta), path('*_unassigned.sig'), optional:true, emit: unassigned
+    tuple val(meta), path('*_matches.sig')   , optional:true, emit: matches
+    tuple val(meta), path('*_prefetch.sig')  , optional:true, emit: prefetch
```
taylorreiter, line 22: Here's the lines of code that control prefect save behavior:
https://github.com/sourmash-bio/sourmash/blob/569d212d65d20a4c94680dd451834046e825949a/src/sourmash/sourmash_args.py#L1188

Saving like this is going to emit some really large files...it might be worth saving the files as a `*.zip`, so they're compressed and packaged together. Zips are also correctly formatted to pass as databases to gather in the future

modules/nf-core/sourmash/gather/meta.yml
```diff
@@ -0,0 +1,92 @@
+name: "sourmash_gather"
+description: Search the reference database for the most highly contained genomes using sourmash
```
taylorreiter, line 2:
```suggestion
description: Search a metagenome sourmash signature against one or many reference databases and return the minimum set of genomes that contain the k-mers in the metagenome.
```

modules/nf-core/sourmash/gather/meta.yml
```diff
@@ -0,0 +1,92 @@
...
+keywords:
+  - hash sketch
```
taylorreiter, line 4:
```suggestion
  - FracMinHash sketch
```

modules/nf-core/sourmash/gather/meta.yml
```diff
@@ -0,0 +1,92 @@
...
+  - kmer
+  - containment
+  - sourmash
+  - genomics
+  - metagenomics
+  - taxonomic classification
+  - taxonomic profiling
+tools:
+  - "sourmash":
+      description: Compute and compare MinHash signatures for DNA data sets.
```
taylorreiter, line 15:
```suggestion
      description: Compute and compare FracMinHash signatures for DNA data sets.
```

modules/nf-core/sourmash/gather/meta.yml
```diff
@@ -0,0 +1,92 @@
...
+      homepage: https://sourmash.readthedocs.io/
+      documentation: https://sourmash.readthedocs.io/
+      tool_dev_url: https://github.com/dib-lab/sourmash
```
taylorreiter, line 18:
```suggestion
      tool_dev_url: https://github.com/sourmash-bio/sourmash
```

modules/nf-core/sourmash/gather/meta.yml
```diff
@@ -0,0 +1,92 @@
...
+      doi: "10.1186/s13059-016-0997-x"
```
taylorreiter, line 19: This is the Mash DOI, which is different than sourmash.
```suggestion
      doi: "10.21105/joss.00027"
```

Can two DOIs be listed? the sourmash tool asks for `10.21105/joss.00027` to be cited when sourmash is used, but gather was published in this preprint: 10.1101/2022.01.11.475838

modules/nf-core/sourmash/gather/meta.yml
```diff
@@ -0,0 +1,92 @@
...
+      description: |
+        Groovy Map containing sample information
+        e.g. [ id:'test', single_end:false ]
+  - signature:
+      type: file
+      description: File containing signatures (hash sketches) of a sample
+      pattern: "*.{sig}"
+  - db:
+      type: file
+      description: Sourmash database (list of signatures or SBTs)
```
taylorreiter, line 34: This can be a list of signatures, SBTs, or signature zip files.

modules/nf-core/sourmash/gather/meta.yml
```diff
@@ -0,0 +1,92 @@
...
+  - save_unassigned:
+      type: boolean
+      description: |
+        If true, output will contain a file with unassigned signatures from the query
```
taylorreiter, line 38:
```suggestion
        If true, output will contain a file that is a sourmash signature containing the unassigned hashes from the query
```

modules/nf-core/sourmash/gather/meta.yml
```diff
@@ -0,0 +1,92 @@
...
+  - save_matches_sig:
+      type: boolean
+      description: |
+        If true, output will contain a file with a query signatures matching the database
```
taylorreiter, line 42:
```suggestion
        If true, output will contain a file that is a sourmash signature composed of the FracMinHash sketches that were matched in the database and that matched the query
```

modules/nf-core/sourmash/gather/meta.yml
```diff
@@ -0,0 +1,92 @@
...
+  - save_prefetch:
+      type: boolean
+      description: |
+        If true, output will contain a file with all prefetch-matched signatures from the database
+  - save_prefetch_csv:
+      type: boolean
+      description: |
+        If true, output will contain a csv file with all prefetch-matched signatures
```
taylorreiter, line 50:
```suggestion
        If true, output will contain a csv file with the names of all prefetch-matched signatures
```

modules/nf-core/sourmash/gather/meta.yml
```diff
@@ -0,0 +1,92 @@
...
+        e.g. [ id:'test', single_end:false ]
+  - versions:
+      type: file
+      description: File containing software versions
+      pattern: "versions.yml"
+  - result:
+      type: file
+      description: |
+        Table with signatures classified as belonging to any of the taxa
+        in the sourmash database.
```
taylorreiter, line 66:
```suggestion
        Table with signatures classified as belonging to any of the genomes
        in the sourmash database(s).
```

modules/nf-core/sourmash/gather/meta.yml
```diff
@@ -0,0 +1,92 @@
...
+      pattern: "*{csv}"
+  - matches:
+      type: file
+      description: |
+        Signatures classified as belonging to any of the taxa
```
taylorreiter, line 71:
```suggestion
        A signature containing FracMinHash sketches of genomes
```

modules/nf-core/sourmash/gather/meta.yml
```diff
@@ -0,0 +1,92 @@
...
+        in the sourmash database.
+      pattern: "*{sig}"
+  - unassigned:
+      type: file
+      description: |
+        Signatures not classified as belonging to any of the taxa
+        in the sourmash database.
```
taylorreiter, line 78:
```suggestion
        A FracMinHash sketch containing hashes (k-mers) that did not match to any of the genomes
        in the sourmash database(s).
```

modules/nf-core/sourmash/gather/meta.yml
```diff
@@ -0,0 +1,92 @@
...
+      pattern: "*{sig}"
+  - prefetch:
+      type: file
+      description: |
+        All prefetch-matched signatures from the database.
+      pattern: "*{sig}"
+  - prefetchcsv:
+      type: file
+      description: |
+        All prefetch-matched signatures from the database in CSV format.
```
taylorreiter, line 88:
```suggestion
        The names of all prefetch-matched signatures from the database in CSV format.
```

taylorreiter (commented):
Added some comments! I also have a general question -- since this module is reliant on a database, what is the recommended way for a workflow to supply that database and to make sure that the k-mer sizes match between the database and the signature?

modules/nf-core/sourmash/gather/main.nf
Midnighter: `path` can indeed be a single file or multiple files. It might require different ways in the `script:` part to ensure correct command input.

modules/nf-core/sourmash/gather/main.nf
Midnighter: Compressed files are preferred in nf-core so that's a great hint 🙂

modules/nf-core/sourmash/gather/main.nf
Midnighter: Okay, had to look up the gather command help, since one or more databases are just separate arguments, the way the process is written now it can already work for one or more databases. 🎉

Midnighter (approved):
I think, the input parameters style relates to the [discussion on slack](https://nfcore.slack.com/archives/CJRH30T6V/p1665567652242099), so having an extensive configuration via `ext.args` might be preferred to the multiple inputs. Otherwise the module looks good to me!

vmikk:
Dear Taylor! Thank you so much for the valuable comments and corrections!
I will modify my PR accordingly soon (probably after the Nextflow Summit).

Considering the database parameter, I think it should work even with multiple databases if we'll use a channel with multiple values (e.g., with `collect` operator) - I will check it.

Also, I've almost drafted `sourmash/taxonomy` module. If you would agree, I will show it to you before the submission.

taylorreiter:
Thanks @vmikk! Yes I would love to review the sourmash taxonomy module as well!

modules/nf-core/sourmash/gather/main.nf
vmikk: `sourmash` is great, it supports compression out of the box.
I've updated the module - now the signatures are compressed with zip ([as recommended](https://sourmash.readthedocs.io/en/latest/databases-advanced.html#storing-signatures-in-zip-files-the-recommended-format)) and csv files are gzipped.

As a side effect, we cannot check md5 sums of the output now (I think it's because ZIP stores timestamps of the files).

PS. currently `--output-unassigned` argument ignores zip-compression of signatures, but I've reported this issue to the developers (https://github.com/sourmash-bio/sourmash/issues/2337).

Midnighter: Bummer about the timestamps. With the gzip tool that's easily avoided but I guess here we'd need a new version of sourmash to make that happen.

jfy133:
> I think, the input parameters style relates to the [discussion on slack](https://nfcore.slack.com/archives/CJRH30T6V/p1665567652242099), so having an extensive configuration via `ext.args` might be preferred to the multiple inputs. Otherwise the module looks good to me!

I think these are generally OK though, I think the discussion slack was related to non-file related mandatory options?

jfy133 (approved):

Midnighter:
I must admit, I'm confused which style is preferred. I think, all of the following inputs could be dropped in favor of using `task.ext.args`:

```nextflow
    val save_unassigned
    val save_matches_sig
    val save_prefetch
    val save_prefetch_csv
```

jfy133:
> I must admit, I'm confused which style is preferred. I think, all of the following inputs could be dropped in favor of using `task.ext.args`:
> 
> ```nextflow
>     val save_unassigned
>     val save_matches_sig
>     val save_prefetch
>     val save_prefetch_csv
> ```

Yeah I guess the discussion is still on going.

The reason why I don't mind these as channels is because they aren't just simply Boolean flags, but you have to supply the time file name with them.

Did consistency with the reason why we have the meta.id as a default information to construct the output file name, I personally like it to be explicit, as then a pipeline dev knows clearly what types of output files a module can generate - and how they are generated (like in fastp).

jfy133:
I would argue we can merge this in, unless you strongly disagree about the `val` channels @Midnighter ?

Midnighter:
Already approved, can definitely be merged 😃

PR #2260, opened 2022-10-11
Add AMPcombi module 

Darcy220606:
This PR adds the ampcombi module. 
It is now working on taking two kinds of inputs. The respective flags are called depending on whether the input is a directory or file.
(1) The results directory (e.g.amp_results) with a specific tree
(2) A file list (equivalent to the individual filepaths of the files generated by the AMP tools) and the sample ID provided through the meta

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`


louperelo:
@nf-core-bot fix linting

modules/nf-core/ampcombi/main.nf
```diff
@@ -0,0 +1,45 @@
...
+    val(outdir)
+
+    output:
+    tuple val(meta), path("${outdir}")          , emit: results_dir
+    path("${outdir}/*/*_diamond_matches.txt")   , emit: txt
+    path("${outdir}/*/*_ampcombi.csv")          , emit: csv
+    path("${outdir}/*/*_amp.faa")               , emit: faa
+    path("${outdir}/*.log")                     , optional:true, emit: log
+    path("${outdir}/*")                         , emit: results_subdir
+    path("${outdir}/amp_ref_database")          , optional:true, emit: results_subdir_db
```
Midnighter, line 22: Some of these should maybe also output a tuple with with `meta` map so that one can associate the result with the sample metainformation.

jfy133: But given this is aggregating a lot of input from different samples to aggregate, does it make sense to include meta? There is no actual consistency between files (i.e., shared DBs etc.)

modules/nf-core/ampcombi/main.nf
```diff
@@ -0,0 +1,45 @@
...
+
+    conda (params.enable_conda ? "bioconda::ampcombi=0.1.3" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/ampcombi:0.1.3--pyhdfd78af_0':
+        'quay.io/biocontainers/ampcombi:0.1.3--pyhdfd78af_0' }"
+
+    input:
+    tuple val(meta),  path(amp_input)
+    path(faa_folder)
+    val(outdir)
```
jfy133, line 13: Why a dedicted input channel and not with not `$prefix`?

modules/nf-core/ampcombi/main.nf
```diff
@@ -0,0 +1,45 @@
...
+
+    output:
+    tuple val(meta), path("${outdir}")          , emit: results_dir
+    path("${outdir}/*/*_diamond_matches.txt")   , emit: txt
+    path("${outdir}/*/*_ampcombi.csv")          , emit: csv
+    path("${outdir}/*/*_amp.faa")               , emit: faa
+    path("${outdir}/*.log")                     , optional:true, emit: log
```
jfy133, line 20: Why is the log optional? If anything one would expect this to be the only 'stable' output from  a tool (even if it can't _do_ anything with the input.

modules/nf-core/ampcombi/main.nf
```diff
@@ -0,0 +1,45 @@
...
+    path "versions.yml"                         , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def VERSION='0.1.3' // WARN: Version information not provided by tool on CLI. Please update this string when bumping container versions.
+    def fileoption = amp_input.isDirectory() ? "--amp_results $amp_input" : "--path_list $amp_input --sample_list ${prefix}"
```
jfy133, line 32: Is it required that the list of sample names comes from prefix? I find this behaviour inconsistent with typical nextflow. I would expect the sample name list should come from a dedicated input channel, and `${prefix}` describes the output file channel name (as this is typically derived from `meta`, which includes an `id` that is a singlular ID for all the given datset).

modules/nf-core/ampcombi/meta.yml
```diff
@@ -0,0 +1,83 @@
...
+      licence: "['MIT']"
+
+input:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. [ id:'test', single_end:false ]
+  - amp_input:
+      type: folder or filelist
```
jfy133, line 31: I don't think this fits with a schema.. .but we don't do much with this meta.yml so maybe can ignore.

Alternativeoly, why not two separate (but mutually exclusive) input channels, one for directory and one for the file list?

modules/nf-core/ampcombi/meta.yml
```diff
@@ -0,0 +1,83 @@
...
+      description: |
+        Groovy Map containing sample information
+        e.g. [ id:'test', single_end:false ]
+  - versions:
+      type: file
+      description: File containing software versions
+      pattern: "versions.yml"
+  - txt:
+      type: file
+      description: An alignment file containing the results from the Diamond alignment step done on all AMP hits.
```
jfy133, line 54:
```suggestion
      description: An alignment file containing the results from the DIAMOND alignment step done on all AMP hits.
```

modules/nf-core/ampcombi/meta.yml
```diff
@@ -0,0 +1,83 @@
...
+      pattern: "/*/*/*_diamond_matches.txt"
+  - csv:
+      type: file
+      description: A file containing the summary report of all predicted AMP hits from all AMP tools given as input and the corresponding taxonomic and functional classification from teh alignment step.
```
jfy133, line 58:
```suggestion
      description: A file containing the summary report of all predicted AMP hits from all AMP tools given as input and the corresponding taxonomic and functional classification from the alignment step.
```

tests/modules/nf-core/ampcombi/test.yml
```diff
@@ -0,0 +1,78 @@
...
+    - path: output/ampcombi/ampcombi_results/amp_ref_database/amp_ref.dmnd
+    - path: output/ampcombi/ampcombi_results/amp_ref_database/general_amps_2022_10_13.tsv
+    - path: output/ampcombi/ampcombi_results/amp_ref_database/general_amps_2022_10_13_clean.fasta
+    - path: output/ampcombi/ampcombi_results/sample_1/sample_1_amp.faa
+    - path: output/ampcombi/ampcombi_results/sample_1/sample_1_ampcombi.csv
+    - path: output/ampcombi/ampcombi_results/sample_1/sample_1_diamond_matches.txt
+    - path: output/ampcombi/ampcombi_results/sample_2/sample_2_amp.faa
+    - path: output/ampcombi/ampcombi_results/sample_2/sample_2_ampcombi.csv
+    - path: output/ampcombi/ampcombi_results/sample_2/sample_2_diamond_matches.txt
+    - path: output/untar1/test_files/ampir/sample_1/sample_1.ampir.faa
```
jfy133, line 15: Can you not include a `contains: ["<some_keyword>"]` for any o fthese?

modules/nf-core/ampcombi/main.nf
louperelo: Meta is related to the sample-name, right? Actually, we should include it to the results_subdir, which is created per sample.

modules/nf-core/ampcombi/main.nf
louperelo: We thought this way we relate the sample_name to the respective AMP-tool-outputs. Also, AMPcombi may have two kinds of input: the path to the folder which then contains all AMP-tools' outputs to be summarised or a list of file-paths to the output-files to be summarised. The second option still needs a test-workflow, but we need to fix a bug in AMPcombi first.

modules/nf-core/ampcombi/main.nf
louperelo: It is optional in the tool and would have to be activated, therefore optional. But we can activate it and make it mandatory in the module.

modules/nf-core/ampcombi/main.nf
jfy133: But a (typical) meta map has only a single ID, how would you assign meta to each sample separately?

modules/nf-core/ampcombi/main.nf
jfy133: Mmm ok... I don't think I'm following... Maybe we need a real time meeting and do live demos 😅

modules/nf-core/ampcombi/main.nf
jfy133: If you have to do a bug fix, I would suggest making it default
 Logs are _always_ useful (for provenance etc), so it would make more sense to make it opt-out of you want to have it optional still (--ignore-log or something)

louperelo: It prints it to the console otherwise, so you can see the progress of the tool

modules/nf-core/ampcombi/main.nf
louperelo: Agreed!

modules/nf-core/ampcombi/main.nf
Midnighter: The way I understood the latest use is that the process will be called per sample rather than for all samples together.

modules/nf-core/ampcombi/main.nf
```diff
@@ -0,0 +1,48 @@
...
+    label 'process_medium'
+
+    conda (params.enable_conda ? "bioconda::ampcombi=0.1.5" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/ampcombi:0.1.5--pyhdfd78af_0':
+        'quay.io/biocontainers/ampcombi:0.1.5--pyhdfd78af_0' }"
+
+    input:
+    tuple val(meta),  path(amp_input)
+    path(faa_folder)
```
adamrtalbot, line 12: An alternative here is to use `stageAs` with separate files: `path(faa_files, stageAs: "faa_folder/*")`
It's a different approach, not necessarily better.

tests/modules/nf-core/ampcombi/main.nf
```diff
@@ -0,0 +1,34 @@
...
+}
+
+workflow test_ampcombi_file_paths {
+    amp_input = [[id:'sample_1'],
+                 [file('https://github.com/nf-core/test-datasets/raw/modules/data/delete_me/ampcombi/test_files/ampir/sample_1/sample_1.ampir.tsv', checkIfExists: true),
+                  file('https://github.com/nf-core/test-datasets/raw/modules/data/delete_me/ampcombi/test_files/amplify/sample_1/sample_1_amplify.tsv', checkIfExists: true)]]
+
+    faa_folder = [
+        [ id:'sample_1' ],
+        file("https://github.com/nf-core/test-datasets/raw/modules/data/delete_me/ampcombi/test_faa.tar.gz", checkIfExists: true)
```
adamrtalbot, line 29: Is there a PR open on the test-datasets repo?

adamrtalbot (commented):
Well done on a complicated module. Just a few things need updating.

modules/nf-core/ampcombi/main.nf
```diff
@@ -0,0 +1,48 @@
...
+
+    output:
+    tuple val(meta), path("*/*")                       , emit: sample_dir
```
louperelo, line 15:
```suggestion
    tuple val(meta), path("*/${prefix}*")                       , emit: sample_dir
```

tests/modules/nf-core/ampcombi/main.nf
louperelo: Hi Adam! No, it's not open, it is this closed PR #649

louperelo (commented):
Thanks for the comments @adamrtalbot

modules/nf-core/ampcombi/main.nf
```diff
@@ -0,0 +1,48 @@
...
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/ampcombi:0.1.5--pyhdfd78af_0':
+        'quay.io/biocontainers/ampcombi:0.1.5--pyhdfd78af_0' }"
+
+    input:
+    tuple val(meta),  path(amp_input)
+    path(faa_folder)
+
+    output:
+    tuple val(meta), path("*/${prefix}*")                       , emit: sample_dir
```
louperelo, line 15:
```suggestion
    tuple val(meta), path("*/*")                       , emit: sample_dir
```

modules/nf-core/ampcombi/main.nf
```diff
@@ -0,0 +1,48 @@
+process AMPCOMBI {
+    tag "$meta.id"
+    label 'process_medium'
+
+    conda (params.enable_conda ? "bioconda::ampcombi=0.1.5" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/ampcombi:0.1.5--pyhdfd78af_0':
+        'quay.io/biocontainers/ampcombi:0.1.5--pyhdfd78af_0' }"
+
+    input:
```
jfy133, line 10: missing `amp_database`

modules/nf-core/ampcombi/main.nf
```diff
@@ -0,0 +1,48 @@
+process AMPCOMBI {
+    tag "$meta.id"
+    label 'process_medium'
```
jfy133, line 3:
```suggestion
    label 'process_single'
```

No threading (although that would probably be cool for the parsing step, if the tool can be slow)

modules/nf-core/ampcombi/main.nf
```diff
@@ -0,0 +1,48 @@
...
+    tuple val(meta), path("*/*diamond_matches.txt")    , emit: txt
+    tuple val(meta), path("*/*ampcombi.csv")           , emit: csv
+    tuple val(meta), path("*/*amp.faa")                , emit: faa
+    path("AMPcombi_summary.csv")                       , optional:true, emit: summary_csv
+    path("AMPcombi_summary.html")                      , optional:true, emit: summary_html
+    path("*.log")                                      , emit: log
+    path("*/amp_ref_database")                         , optional:true, emit: results_db
+    path("*/amp_ref_database/*.dmnd")                  , optional:true, emit: results_db_dmnd
+    path("*/amp_ref_database/*.clean.fasta")           , optional:true, emit: results_db_fasta
+    path("*/amp_ref_database/*.tsv")                   , optional:true, emit: results_db_tsv
```
jfy133, line 25: Why don't these have meta  if the earlier ones do?

modules/nf-core/ampcombi/main.nf
```diff
@@ -0,0 +1,48 @@
...
+    path "versions.yml"                                , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def fileoption = amp_input instanceof List ? "--path_list '${amp_input.collect{"$it"}.join("' '")}'" : "--amp_results $amp_input/"
```
jfy133, line 34: Is the `join` still meant to have the single quotes here, I though this was one of our bugs on Weds..?

modules/nf-core/ampcombi/meta.yml
```diff
@@ -0,0 +1,95 @@
...
+      type: folder
+      description: If the AMP reference database is not provided by the user using the flag `--amp_database', by default the DRAMP database will be downloaded, filtered and stored in this folder.
+      pattern: "/amp_ref_database"
+  - results_db_fasta:
+      type: file
+      description: AMP reference database fasta file, cleaned of diamond-uncompatible characters.
+      pattern: "/amp_ref_database/*.clean.fasta"
+  - results_db_dmd:
+      type: file
+      description: AMP reference database converted to Diamond database format.
```
jfy133, line 78:
```suggestion
      description: AMP reference database converted to DIAMOND database format.
```

tests/modules/nf-core/ampcombi/main.nf
jfy133: I think that is the wrong PR @louperelo but reegardless, the file is already on test-datasets based on the URL so looks fine to me

jfy133 (changes requested):
Almost there but starting to look pretty good!

modules/nf-core/ampcombi/main.nf
louperelo: We put medium because of the alignment step with Diamond, which requires threading. (There is a diamond module as well and it has 'process_medium' as well

modules/nf-core/ampcombi/main.nf
louperelo: The amp_database download is part of AMPcombi.

modules/nf-core/ampcombi/main.nf
louperelo: We did not put meta, because these files are not related specifically to one sample: The reference database may be used for several samples. The AMPcombi_summary files can only be generated, if AMPcombi parses a folder (like in the first test-workflow). I don't think the module will be applied with this option, but as the tool is able to do so, we included it.

modules/nf-core/ampcombi/main.nf
louperelo: Well, we tried all options (single, double, none) and it made no difference. The bug was in the AMPcombi-code ...

modules/nf-core/ampcombi/main.nf
Darcy220606: We stayed with the faa_folder without staging it

modules/nf-core/ampcombi/main.nf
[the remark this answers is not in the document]
Darcy220606: Thanks a lot for the suggestion @adamrtalbot  this was actually required because it was retaining unnecessary tar-d files.

modules/nf-core/ampcombi/main.nf
mahesh-panchal: The modules should also work in environments with no Internet. There should always be an option to supply a predownloaded database.

Darcy220606: The module indeed has an optional argument for the user to supply their own database indeed using the flag --amp-database instead of downloading it

mahesh-panchal: This needs to be an input so Nextflow can stage it properly. Optional inputs are handled by using an empty list when the input should be empty.
```nextflow
process TOOL {
    input:
    tuple val(meta), path(bam)
    path( opt_db )

    script:
    def db = optdb? "--db $opt_db": ""
    """
    cmd $db $bam
    """
}

workflow {
    TOOL (
        ch_bams,
        params.db ? file( params.db ) : []
    )
}
```

modules/nf-core/ampcombi/main.nf
jfy133: But we can't see that in a NXF pipeline, so I would suggest hardcoding it :)

modules/nf-core/ampcombi/main.nf
jfy133: Ok but where do you supply the threads to ampCOMBI to give to DIAMOND? I don't see any flag for this in the github repo/help? https://github.com/Darcy220606/AMPcombi

modules/nf-core/ampcombi/main.nf
jfy133: If it takes in meta you might as well pass out meta, you don't wnat to to assume what a user is doing.

You could have different batches of AMP outputs that you want to store separately.

modules/nf-core/ampcombi/main.nf
Darcy220606: Alright we added the threads fro diamond in the new ampcombi release

modules/nf-core/ampcombi/main.nf
Darcy220606: Alright, we added them to all outputs.

modules/nf-core/ampcombi/main.nf
Darcy220606: Alright, we included the amp_database as an optional input, but still waiting for the conda version to bump our conda env to v0.1.6

modules/nf-core/ampcombi/meta.yml
```diff
@@ -0,0 +1,99 @@
+name: "ampcombi"
+description: A tool to parse and suumarise results from antimicorbial peptides tools and provide the respective functional classification.
```
adamrtalbot, line 2:
```suggestion
description: A tool to parse and summarise results from antimicrobial peptides tools and present functional classification.
```

modules/nf-core/ampcombi/meta.yml
```diff
@@ -0,0 +1,99 @@
...
+
+input:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. [ id:'test', single_end:false ]
+  - amp_input:
+      type: folder or filelist
+      description: The path to the directory containing the results for the AMP tools for each sample processed or a list of files cooresponding to each file generated by AMP tools.
```
adamrtalbot, line 32:
```suggestion
      description: The path to the directory containing results for the AMP tools for each sample processed or a list of files corresponding to each file generated by AMP tools.
```

modules/nf-core/ampcombi/main.nf
```diff
@@ -0,0 +1,52 @@
...
+    path( opt_amp_db )
+
+    output:
+    tuple val(meta), path("${meta.id}*")                        , emit: sample_dir
+    tuple val(meta), path("${meta.id}/*diamond_matches.txt")    , emit: txt
+    tuple val(meta), path("${meta.id}/*ampcombi.csv")           , emit: csv
+    tuple val(meta), path("${meta.id}/*amp.faa")                , emit: faa
+    tuple val(meta), path("AMPcombi_summary.csv")               , optional:true, emit: summary_csv
+    tuple val(meta), path("AMPcombi_summary.html")              , optional:true, emit: summary_html
+    tuple val(meta), path("*.log")                              , emit: log
```
adamrtalbot, line 22:
```suggestion
    tuple val(meta), path("*.log")                              optional: true, emit: log
```

Just in case?

modules/nf-core/ampcombi/meta.yml
```diff
@@ -0,0 +1,99 @@
...
+  - faa_folder:
+      type: folder
+      description: The path to the folder containing the respective protein fasta files with '.faa' extension. Filenames have to contain the corresponding sample-name, i.e. sample_1.faa
```
adamrtalbot, line 35:
```suggestion
      description: The path to the folder containing protein fasta files with '.faa' extension. Filenames have to contain the corresponding sample name, i.e. sample_1.faa
```

adamrtalbot (approved):
I don't know anything about how this tool works, but I can follow the code so it gets a tick. I've made a few spelling corrections.

modules/nf-core/ampcombi/main.nf
Darcy220606: True

Midnighter:
Congrats 🎉

PR #2277, opened 2022-10-12
Hapibd

ashotmarg:
## PR checklist

Closes #XXX 

- [x] Add a resource `label`


modules/hapibd/meta.yml
```diff
@@ -0,0 +1,67 @@
...
+      pattern: "*.{vcf,vcf.gz}"
+  - map:
+      type: file
+      description: genetic map with cM units in PLINK format
+      pattern: "*.{map,map.gz,map.zip}"
+  - out:
+      type: string
+      description: the output filename prefix
+      pattern: "example: cohort.hapibd.out"
+  - excludesamples:
```
pawelciurkaardigen, line 36: input variable's name is `exclude`. I think it should be the same here.

modules/hapibd/meta.yml
```diff
@@ -0,0 +1,67 @@
...
+      description: phased VCF file with a GT FORMAT subfield with no missing alleles
+      pattern: "*.{vcf,vcf.gz}"
+  - map:
+      type: file
+      description: genetic map with cM units in PLINK format
+      pattern: "*.{map,map.gz,map.zip}"
+  - out:
+      type: string
+      description: the output filename prefix
+      pattern: "example: cohort.hapibd.out"
```
pawelciurkaardigen, line 35: It seems like this is not one of the inputs anymore

tests/modules/hapibd/main.nf
```diff
@@ -0,0 +1,28 @@
...
+
+include { HAPIBD } from '../../../modules/hapibd/main.nf'
+
+workflow test_hapibd {
+
+    input = [
+        [ id:'test', single_end:false ], // meta map
+        file("https://github.com/nf-core/test-datasets/raw/modules/data/delete_me/hapibd/target.truth.vcf.gz", checkIfExists: true)
+    ]
+    map = file("https://github.com/nf-core/test-datasets/raw/modules/data/delete_me/hapibd/target.map", checkIfExists: true)
```
pawelciurkaardigen, line 13: is this testdata file going to be moved outside of `delete_me` directory?

pawelciurkaardigen (changes requested):
Looks good, I left few comments

modules/hapibd/meta.yml
ashotmarg: You're right, I removed this!

tests/modules/hapibd/main.nf
ashotmarg: Yeah, James suggested me to put the test input files in there :)

ashotmarg:
Hi @pawelciurkaardigen, could you please have another look at the code. I have now addressed all your comments and also added a new docker (quay.io), singularity and conda images after these were generated in the bioconda.
I don't know why the nf-core linting is failing, is it because the folder structure was changed? Also, it seems that the singularity pull is also failing due to versions.yml, is that correct? Not sure why that is failing only in singularity.

modules/hapibd/main.nf
```diff
@@ -0,0 +1,50 @@
...
+
+    conda (params.enable_conda ? "bioconda::hap-ibd=1.0.rev20May22.818" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/hap-ibd:1.0.rev20May22.818--hdfd78af_0':
+        'quay.io/biocontainers/hap-ibd:1.0.rev20May22.818--hdfd78af_0' }"
+
+    input:
+        tuple val(meta), path(vcf)
+        path(map)
+        path(exclude)
```
nvnieuwk, line 13:
```suggestion
    tuple val(meta), path(vcf)
    path(map)
    path(exclude)
```

modules/hapibd/main.nf
```diff
@@ -0,0 +1,50 @@
...
+
+
+    output:
+        tuple val(meta), path("*.hbd.gz"),   emit: hbd
+        tuple val(meta), path("*.ibd.gz"),   emit: ibd
+        tuple val(meta), path("*.log"),      emit: log
+        path "versions.yml",                          emit: versions
```
nvnieuwk, line 20:
```suggestion
    tuple val(meta), path("*.hbd.gz"), emit: hbd
    tuple val(meta), path("*.ibd.gz"), emit: ibd
    tuple val(meta), path("*.log")   , emit: log
    path "versions.yml"              , emit: versions
```

modules/hapibd/main.nf
```diff
@@ -0,0 +1,50 @@
...
+
+    when:
+        task.ext.when == null || task.ext.when
```
nvnieuwk, line 23:
```suggestion
    task.ext.when == null || task.ext.when
```

modules/hapibd/main.nf
```diff
@@ -0,0 +1,50 @@
...
+
+    script:
+        def args = task.ext.args ?: ''
+        def prefix = task.ext.prefix ?: "${meta.id}"
+        def excludesamples_command = exclude ? "excludesamples=$exclude" : ""
```
nvnieuwk, line 28:
```suggestion
    def args = task.ext.args ?: ''
    def prefix = task.ext.prefix ?: "${meta.id}"
    def excludesamples_command = exclude ? "excludesamples=$exclude" : ""
```

modules/hapibd/meta.yml
```diff
@@ -0,0 +1,63 @@
...
+      description: text file containing samples one sample per line to be excluded from the analysis
+      pattern: "*.*"
+
+output:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing cohort information
+        e.g. [ id:'test' ]
+
```
nvnieuwk, line 43:
```suggestion
```

modules/hapibd/meta.yml
```diff
@@ -0,0 +1,63 @@
...
+  - versions:
+      type: file
+      description: File containing software versions
+      pattern: "versions.yml"
+
```
nvnieuwk, line 48:
```suggestion
```

nvnieuwk (changes requested):
LGTM, just some layout changes to stay consistent with all the other modules

nvnieuwk:
> Thanks for the review @nvnieuwk, well noticed, I've accepted the changes. However, do you know why two of the checks are failing?

Hi, the linting test is failing because your module isn't inside of `modules/nf-core/` you should move `modules/hapibd` to `modules/nf-core/hapibd`. You should also do the same for the module in the `test/modules` folder. And also change the file paths inside of `tests/config/pytest_modules.yml` for your module to these new file paths. (Did you use `nf-core modules create` to create the module?)

ashotmarg:
That's what I figured, thanks! Yes, I had used the `nf-core modules create` which is why I was confused. When I made my previous module (somalier), all my new module folders/files were in the same folder as the rest of the nf-core modules. However, this time `nf-core modules create` made module and test folders into their "own" folders, which seems to be creating the problem. I thought this was the  new way of doing things... Please check if this is a glitch only at my end, but regardless, I will change the folders/files accordingly, thanks.

nvnieuwk:
What version of the `nf-core` tools are you using? A change to the modules structure has been done not that long ago and the older versions of the tooling are invalid for module creation

ashotmarg:
Aaaah, ok, I am using version 2.4.1.

nvnieuwk:
That's the problem indeed, you'd better upgrade to version 2.6

ashotmarg:
I just did, thanks! Will now try with that one :)

ashotmarg:
Last time, with older version of nf-core the `nf-core modules create-test-yml hapibd` worked like charm for generating the test.yml file. However this time it's not the case unfortunately. I can't be sure it's the version differences though, I might have done something wrong this time/for this module...

Using docker with `nf-core modules create-test-yml hapibd` I am getting an error "Exception in thread "main" java.lang.IllegalArgumentException: File does not exist: target.truth.vcf.gz". I think, basically it cannot find the input file. I am guessing it is some sort of mounting issue with the docker? I managed to automatically create a test.yml providing "conda" for `nf-core modules create-test-yml hapibd`. This worked but I didn't have the output file "versions.yml" (along with its md5sum) in the generated test.yml file. It is part of the (standard) output, so I think I should have it. I think I got it automatically last time. Now I had to manually add the line "- path: output/hapibd/versions.yml" in the "test.yml".

The singularity check is still failing though, I will test this tomorrow on my other machine which has singularity installed.

nvnieuwk:
That's a strange error for docker, it seemed to work fine when I tried it yesterday on Gitpod... There have been some problems fetching the data from the test-datasets repository, so that could maybe be it? 

ˇversions.yml` is no longer added to `test.yml` since `nf-core` version 2.5 (I think). The tool now automatically checks for the presence of `versions.yml`.

Also strange that the `singularity` test fails when `docker` succeeds since the containers are based on each other...

tests/modules/nf-core/hapibd/test.yml
```diff
@@ -0,0 +1,25 @@
...
+  tags:
+    - hapibd
+  files:
+    - path: output/hapibd/test.hbd.gz # this is an empty file for this test example
+    - path: output/hapibd/test.ibd.gz
+      contains: ["I3019"]
+    - path: output/hapibd/test.log
+      contains: ["hap-ibd.jar"]
+    - path: output/hapibd/versions.yml
+      ## contains: ["hap-ibd.jar"]
```
nvnieuwk, line 12:
```suggestion
```

tests/modules/nf-core/hapibd/test.yml
```diff
@@ -0,0 +1,25 @@
...
```
nvnieuwk, line 25:
```suggestion
```

nvnieuwk (changes requested):

ashotmarg:
I just tried the singularity and it seems to work on my machine. The `PROFILE=singularity pytest --tag hapibd --symlink --keep-workflow-wd` command give the below output, but not sure why it is failing here at the PR stage. 

Screenshot 2022-10-18 at 10 41 46

nvnieuwk:
That's really weird. I'm not that experienced in using GHA tests, but I'll post on the Slack workspace. Somebody there will know :)

ashotmarg:
Hi @pawelciurkaardigen, could you perhaps check this again? The singularity check keeps failing and we (@nvnieuwk and I) don't really know why and how to proceed next. I can run the test locally and it's fine.

pawelciurkaardigen (approved):
Looks good to me. I also don't have an idea why there is a problem with singularity mode.

ashotmarg:
Hi @jfy133 , would you have time to replicate this singularity issue?

jfy133:
> Hi @jfy133 , would you have time to replicate this singularity issue?

Will look in the next 30m or so

jfy133:
Can confirm singularity looks OK to me, but this might be related:

```
"test_hapibd:HAPIBD":
    hapibd: [0.001s][warning][os,container] Duplicate cpuset controllers detected. Picking /sys/fs/cgroup/cpusetrevskipping /home/runner/go/src/github.com/hpcng/singularity/prefix/var/singularity/mnt/session/final/sys/fs/cgroup/cpuset.
    ```
    
This is from your previous failed test. this looks like an odd edge-case relate to GHA, so I would approve if code looks ok @nvnieuwk ?

**Review by nvnieuwk — APPROVED** [review]

**Comment by nvnieuwk** [comment]
Forgot to approve, you can merge :)

**Comment by jfy133** [comment]
Yeha, same again:

```
E           yaml.parser.ParserError: while parsing a block mapping
E             in "<unicode string>", line 2, column 5:
Warning: E                   hapibd: [0.001s][warning][os,con ... 
E                   ^
E           expected <block end>, but found '['
E             in "<unicode string>", line 2, column 21:
Warning: E                   hapibd: [0.001s][warning][os,container] Duplicat ... 
E                                   ^
```

**Comment by jfy133** [comment]
Can also confirm version reporting command works fine for me locally in singualriy: 

```
Singularity> echo $(hap-ibd 2>&1 | head -n1 | sed 's/^hap-ibd.jar  \[ version //; s/, /rev/; s/ \]//')
1.0rev20May22.818
```

jfy133:
@ashotmarg you can just merge, must be specific to GHA if no-one else is getting the wierd version error

ashotmarg:
Thanks a lot @jfy133 for having a look at this ;)

PR #2288, opened 2022-10-12
Abra2/realign module 

### LINKED ISSUE #2186: new module: ABRA2

alanhoyle:
## PR checklist

Closes #2186 

- [x] Add a resource `label`

### DIFF
--- a/modules/nf-core/abra2/realign/main.nf
+++ b/modules/nf-core/abra2/realign/main.nf
@@ -1,0 +1,60 @@
+
+process ABRA2_REALIGN {
+    tag "$meta.id" + (binding.variables["meta2"] ? ",$meta2.id" : ",none") + (binding.variables["meta3.id"] ? ",$meta3.id" : ",none")
+    label 'process_high'
+
+    conda (params.enable_conda ? "bioconda::abra2=2.24" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/quay.io/biocontainers/abra2:2.24--h9f5acd7_1':
+        'quay.io/biocontainers/abra2:2.24--h9f5acd7_1' }"
+
+    input:
+    tuple val(meta), path(bam),path(bam_bai)
+    tuple val(meta2), path(bam2),path(bam_bai2)
+    tuple val(meta3), path(bam3),path(bam_bai3)
+    tuple path(fasta), path(fasta_fai)
+    path(targets)
+
+    output:
+    tuple val(meta), path("${meta.id}.abra.bam"), emit: bam
+    tuple val(meta2), path("${meta2.id}.abra.bam"), optional : true, emit: bam2
+    tuple val(meta3), path("${meta3.id}.abra.bam"), optional : true, emit: bam3
+    path "versions.yml"           , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def VERSION = '2.24' // WARN: Version information not provided by tool on CLI. Please update this string when bumping container versions.
+
+    def targets_opt = targets ? "--targets ${targets[0]}" : ""
+
+    def bam_in = bam.getName()
+    def bam2_in = meta2.id ? ','+bam2.getName() : ''
+    def bam3_in = meta3.id ? ','+bam3.getName() : ''
+
+
+    def bam_out = meta.id + ".abra.bam"
+    def bam2_out = meta2 ? ','+meta2.id+".abra.bam" : ''
+    def bam3_out = meta3 ? ','+meta3.id+".abra.bam" : ''
+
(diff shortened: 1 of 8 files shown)


modules/nf-core/abra2/realign/main.nf
```diff
@@ -0,0 +1,60 @@
...
+    conda (params.enable_conda ? "bioconda::abra2=2.24" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/quay.io/biocontainers/abra2:2.24--h9f5acd7_1':
+        'quay.io/biocontainers/abra2:2.24--h9f5acd7_1' }"
+
+    input:
+    tuple val(meta), path(bam),path(bam_bai)
+    tuple val(meta2), path(bam2),path(bam_bai2)
+    tuple val(meta3), path(bam3),path(bam_bai3)
+    tuple path(fasta), path(fasta_fai)
```
nvnieuwk, line 15:
```suggestion
    path fasta
    path fasta_fai
```

Suggestion to be along the same lines as other modules that use `fasta` and `fasta_fai`

modules/nf-core/abra2/realign/main.nf
```diff
@@ -0,0 +1,60 @@
...
+    path(targets)
+
+    output:
+    tuple val(meta), path("${meta.id}.abra.bam"), emit: bam
```
nvnieuwk, line 19: Does the tool always need a maximum of 3 BAMs? I think it would maybe be better if a list of BAMs was used as input instead of separate input channels for every BAM

modules/nf-core/abra2/realign/main.nf
```diff
@@ -0,0 +1,60 @@
...
+    def bam_in = bam.getName()
+    def bam2_in = meta2.id ? ','+bam2.getName() : ''
+    def bam3_in = meta3.id ? ','+bam3.getName() : ''
+
+
+    def bam_out = meta.id + ".abra.bam"
+    def bam2_out = meta2 ? ','+meta2.id+".abra.bam" : ''
+    def bam3_out = meta3 ? ','+meta3.id+".abra.bam" : ''
+
+    def single_end = meta.single_end ? '--single' : ''
```
nvnieuwk, line 42: This should not be in the module itself. This is up to the pipeline developer to implement themselves when using the module

nvnieuwk (changes requested):
A couple suggestions. Does the tool also support CRAM?

nvnieuwk:
You should also add your module to `tests/config/pytest_modules.yml` for the CI tests to work

modules/nf-core/abra2/realign/main.nf
alanhoyle: The issue is that the bams need to be tracked in and out.  For example, we use it to realign tumor and normal bam together, and it's important that the tumor.abra.bam and normal.abra.bam aren't swapped.  It's not just taking an arbitrary list of bams and their relative metadata is important.

modules/nf-core/abra2/realign/main.nf
alanhoyle: See above:  We also have a potential use case with tumor DNA, normal DNA and tumor RNA

alanhoyle: We also have a use case with a tumor-only workflow where we run the tumor bams alone.

modules/nf-core/abra2/realign/main.nf
nvnieuwk: Will this be the case for every use case of the tool? (if so, then it's fine by me how you implement it now)

modules/nf-core/abra2/realign/main.nf
nvnieuwk: Yeah but the `nf-core` guidelines state that you should specify this in the `args` (https://nf-co.re/docs/contributing/modules#general). In your case you could specify `args` like this in the config file:

```
process {
    withName: ABRA2_REALIGN {
        ext.args = [
            {meta.single_end ? '--single' : ''},
            '' // Additional arguments
        ].join(' ')
    }
}
```

Feel free to ask me if something is not clear on this or other aspects of the module creation :)

modules/nf-core/abra2/realign/main.nf
alanhoyle: We use it in 3 ways.  alone, with T-N DNA bams, and with T-N DNA, and an RNA bam.  It might be okay to implement a bam-list-input version for use cases where we don't need to worry about the sample metadata, but that wouldn't work for the cases. we actually use in practice.

nvnieuwk: This is an interesting case, could you maybe post this on Slack in the #modules channel?

alanhoyle:
I realized the better way to do this would be to do a module for ABRA taking N bams, as input, and implement a subworkflow to handle the T-N dna and T-N-RNA. case.

PR #2289, opened 2022-10-12
Add auto cancelation

mashehu:


.github/workflows/pytest-workflow.yml
```diff
@@ -5,6 +5,11 @@ on:
   pull_request:
     branches: [master]
 
+# Cancel if a newer run is started
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
apeltzer, line 10: Doesn't this now also need the raw stuff?

jfy133 (approved):

PR #2291, opened 2022-10-13
refactor: Support container_registry in modules

Emiller88:
Motivated by: https://aws.amazon.com/blogs/hpc/biocontainers-are-now-available-in-amazon-ecr-public-gallery/

They don't have mulled images being pushed yet however, but this should support other urls as well so it's a nice change.


modules/nf-core/bowtie/align/main.nf
```diff
@@ -3,9 +3,12 @@ process BOWTIE_ALIGN {
     label 'process_high'
 
     conda (params.enable_conda ? 'bioconda::bowtie=1.3.0 bioconda::samtools=1.15.1' : null)
+    def docker_image = params.docker_url != null ?
+        '${params.docker_url}/mulled-v2-ffbf83a6b0ab6ec567a336cf349b80637135bca3:676c5bcfe34af6097728fea60fb7ea83f94a4a5f-0' :
+        'quay.io/biocontainers/mulled-v2-ffbf83a6b0ab6ec567a336cf349b80637135bca3:676c5bcfe34af6097728fea60fb7ea83f94a4a5f-0'
```
grst, line 8: we don't use this?

modules/nf-core/bowtie/align/main.nf
```diff
@@ -3,9 +3,12 @@ process BOWTIE_ALIGN {
...
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
         'https://depot.galaxyproject.org/singularity/mulled-v2-ffbf83a6b0ab6ec567a336cf349b80637135bca3:676c5bcfe34af6097728fea60fb7ea83f94a4a5f-0' :
-        'quay.io/biocontainers/mulled-v2-ffbf83a6b0ab6ec567a336cf349b80637135bca3:676c5bcfe34af6097728fea60fb7ea83f94a4a5f-0' }"
+        ${docker_url ?: 'quay.io/biocontainers'}/mulled-v2-ffbf83a6b0ab6ec567a336cf349b80637135bca3:676c5bcfe34af6097728fea60fb7ea83f94a4a5f-0' }"
```
grst, line 11:
```nextflow
container { workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
    'https://depot.galaxyproject.org/singularity/mulled-v2-ffbf83a6b0ab6ec567a336cf349b80637135bca3:676c5bcfe34af6097728fea60fb7ea83f94a4a5f-0' :
    "${docker_url ?: 'quay.io/biocontainers'}/mulled-v2-ffbf83a6b0ab6ec567a336cf349b80637135bca3:676c5bcfe34af6097728fea60fb7ea83f94a4a5f-0' }
```

Personally, I think we should get away with using a closure. But nested GStrings should work too. 

@mahesh-panchal usually *knows* these things, whereas I am just guessing.

Emiller88: This fails with 
```
    container { workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
        'https://depot.galaxyproject.org/singularity/mulled-v2-ffbf83a6b0ab6ec567a336cf349b80637135bca3:676c5bcfe34af6097728fea60fb7ea83f94a4a5f-0' :
        "${docker_url ?: 'quay.io/biocontainers'}/mulled-v2-ffbf83a6b0ab6ec567a336cf349b80637135bca3:676c5bcfe34af6097728fea60fb7ea83f94a4a5f-0' }
```

grst:
Is this accessible from outside aws?

modules/nf-core/bowtie/align/main.nf
```diff
@@ -3,9 +3,9 @@ process BOWTIE_ALIGN {
     label 'process_high'
 
     conda (params.enable_conda ? 'bioconda::bowtie=1.3.0 bioconda::samtools=1.15.1' : null)
-    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
-        'https://depot.galaxyproject.org/singularity/mulled-v2-ffbf83a6b0ab6ec567a336cf349b80637135bca3:676c5bcfe34af6097728fea60fb7ea83f94a4a5f-0' :
-        'quay.io/biocontainers/mulled-v2-ffbf83a6b0ab6ec567a336cf349b80637135bca3:676c5bcfe34af6097728fea60fb7ea83f94a4a5f-0' }"
+    container { workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        "https://depot.galaxyproject.org/singularity/mulled-v2-ffbf83a6b0ab6ec567a336cf349b80637135bca3:676c5bcfe34af6097728fea60fb7ea83f94a4a5f-0" :
+        "${params.docker_url ?: 'quay.io/biocontainers'}/mulled-v2-ffbf83a6b0ab6ec567a336cf349b80637135bca3:676c5bcfe34af6097728fea60fb7ea83f94a4a5f-0" }
```
Midnighter, line 8: I would use a different name for this variable since we want to change the registry only (I think).

Midnighter: Maybe it should be

```
"${params.docker_registry ?: 'quay.io'}/biocontainers/mulled-v2-ffbf83a6b0ab6ec567a336cf349b80637135bca3:676c5bcfe34af6097728fea60fb7ea83f94a4a5f-0"
```

```
params.docker_registry = "public.ecr.aws"
```

apeltzer:
When looking at the ECR gallery, @Emiller88 and I were unable to find the mulled containers there - need to ask AWS folks on whether this is not mirrored (yet?). 

@delagoya - can you share some insights on this?

modules/nf-core/bowtie/align/main.nf
mahesh-panchal: Configuration closures only work from the config. In the `process` block you need to have a GString leveraging the ternary expressions inside it. I really dislike lots of nested ternary expressions though as I think it makes the script look messy. With all these different registry options, I think they would be better off as profiles and the profile sets a `container_registry` parameter, which can then remove some of the existing logic since the image name and tag ( + build ) will same regardless of registry.

mahesh-panchal:
My suggestion is to use a profiles to set a parameter to supply the container registry and then simplify the ternary expression.
```nextflow
profiles {
    singularity {
         singularity.enabled = true
         params.container_registry = 'https://depot.galaxyproject.org/singularity'
     }
     docker {
         docker.enabled = true
         params.container_registry = 'quay.io/biocontainers'
     }
     aws {
          params.container_registry = 'public.ecr.aws'
     }
}
```
and then the `process` tag looks like:
```nextflow
process TOOL_SUBTOOL {
    container "${ (params.container_registry ? 'quay.io' ) + 'image:tag--build' }"
}
```

modules/nf-core/bowtie/align/main.nf
Emiller88: This does assume they have `biocontainers` in the name

mahesh-panchal:
Also one can just override from the command-line or `-params-file` if they want to use the docker registry from singularity or use a private repository.

Emiller88:
> Also one can just override from the command-line or `-params-file` if they want to use the docker registry from singularity or use a private registry.

Great point! I think that solves it.

Emiller88:
@mahesh-panchal Thoughts on:

```
"${ ... }"
```
vs. just a closure? `{}`

Emiller88:
> Container needs a string in this context I think.

https://github.com/nf-core/modules/pull/2291/commits/98f8156fb1171cc9bb0b585516db6cc15d5c2a54

This works :shrug:

Emiller88:
> If the string quotes are not needed, then you don't need the `{ }` either.

No dice 

```
Unknown process directive: `plus`

Did you mean of these?
        cpus

 -- Check script 'tests/modules/nf-core/bowtie/align/../../../../../modules/nf-core/bowtie/align/main.nf' at line: 6 or see '.nextflow.log' file for more details
```

Emiller88:
```
    def image = "mulled-v2-ffbf83a6b0ab6ec567a336cf349b80637135bca3:676c5bcfe34af6097728fea60fb7ea83f94a4a5f-0"
    conda (params.enable_conda ? 'bioconda::bowtie=1.3.0 bioconda::samtools=1.15.1' : null)
    container { (params.container_registry ?: 'quay.io' ) + image }
```
Thoughts? I think this is clean.

mahesh-panchal:
Maybe then

> > If the string quotes are not needed, then you don't need the `{ }` either.
> 
> No dice
> 
> ```
> Unknown process directive: `plus`
> 
> Did you mean of these?
>         cpus
> 
>  -- Check script 'tests/modules/nf-core/bowtie/align/../../../../../modules/nf-core/bowtie/align/main.nf' at line: 6 or see '.nextflow.log' file for more details
> ```

Hmm. Maybe then just `( )` is needed instead of `{ }`. This may be a parsing context thing. I don't see why a closure is needed here.

grst:
> If the string quotes are not needed, then you don't need the { } either.

I think one of the two is needed when using custom config files with `-c custom.config` to delay the evaluation until the params are ready. But I heard `-c` is discouraged nowadays anyway? 
And is there any difference between `"${}"` compared to `{  }` as far as the effect is concerned?

apeltzer:
Opened an issue here too - it will also "break" e.g. `nf-core modules create` in terms of that the biocontainers API will return a queried string with a full path to the container to use --> we can fix that to make it conveniently split base_url and the image id matching what is implemented in this PR, but should not forget to do so once this is merged ;-)

grst:
> container { (params.container_registry ?: 'quay.io' ) + image }

Slash after quay.io missing?

Emiller88:
> > container { (params.container_registry ?: 'quay.io' ) + image }
> 
> Slash after quay.io missing?

It's in the test config for now. Idk which place is better to put it, probably in the module? 

Also thoughts on `quay.io` vs `quay.io/biocontainers` as a default?

mahesh-panchal:
> > If the string quotes are not needed, then you don't need the { } either.
> 
> I think one of the two is needed when using custom config files with `-c custom.config` to delay the evaluation until the params are ready. But I heard `-c` is discouraged nowadays anyway? And is there any difference between `"${}"` compared to `{ }` as far as the effect is concerned?

The `"${ }"` is a groovy expression inside a GString, `{ }` is a closure/anonymous function (and in the context above a function that's just returning a string). The deferring of evaluation is only relevant from configuration files, so since this is from the script context, it does not have an effect.

mahesh-panchal:
> Also thoughts on `quay.io` vs `quay.io/biocontainers` as a default?

The default should be whatever makes a full path. I didn't check it, so I guess it should be `quay.io/biocontainers`

mahesh-panchal:
> It's in the test config for now. Idk which place is better to put it, probably in the module?

I think the best place is preceding the image name in the module. Most things can handle double slash, but break if it's not there.

Emiller88:
@mahesh-panchal I think this is good: https://github.com/nf-core/modules/pull/2291/commits/a8ce6a03cfc25602764b5ba87d59bd4379f65433

Emiller88:
Thoughts on reorging the conda and containers?
```
    def image = "/mulled-v2-ffbf83a6b0ab6ec567a336cf349b80637135bca3:676c5bcfe34af6097728fea60fb7ea83f94a4a5f-0"
    container { (params.container_registry ?: 'quay.io/biocontainers') + image }
    conda (params.enable_conda ? 'bioconda::bowtie=1.3.0 bioconda::samtools=1.15.1' : null)
```
It's not alphabetical :upside_down_face:

FriederikeHanssen:
So singularity would be set in the nextflow config?

mahesh-panchal:
Would it make adjusting the tooling easier if the code were:
```nextflow
    def image = "mulled-v2-ffbf83a6b0ab6ec567a336cf349b80637135bca3:676c5bcfe34af6097728fea60fb7ea83f94a4a5f-0"
    container [ params.container_registry ?: 'quay.io/biocontainers' ,  image ].join('/')
    conda (params.enable_conda ? 'bioconda::bowtie=1.3.0 bioconda::samtools=1.15.1' : null)
```

Emiller88:
Ooo I like that. I wonder if we can get https://docs.renovatebot.com/ working and drown ourselves in automated image update PRs (we could set them to automerge if the tests pass and it's not a major version bump)

apeltzer:
If we do this change entirely, that means also everyone has to use a separate singularity profile to keep pipelines from converting docker -> singularity images. Something to document to tell users upfront what is happening as well as to bear in mind for potential crashes when some (unaware) users now have to rely on the docker->singularity path again. Had quite a number of people with e.g. `squashfs not installed` issues in the past 👍🏻 

Otherwise looking good to me in general (didn't go through 600+ modules).

Emiller88:
They could just manually set the container image if they wanted the docker image right?

mahesh-panchal:
Not necessarily. `params.container_registry` could also be set in institutional profiles as well as common profiles like `singularity`. So both common use cases are covered. In cases where using neither profile are used then the user would likely be specifying their own config anyway.

apeltzer:
Should we then set the default of the singularity profile in nf-core to the `depot.galaxy...` url? I'm happy with the changes and agree that most of this should not cause issues also giving @mahesh-panchal 's comment above now too 👍🏻

mahesh-panchal:
Yes, I think that should be the plan, but also any institutional configs that enable containers need to also be updated to use the parameter too.

- update template profiles with `params.container_registry`
- update institutional configs with `params.container_registry`.

apeltzer:
Would vote for merging this then after some sanity checks? 🤠

maxulysse:
> Would vote for merging this then after some sanity checks? 🤠

I vote for updating template and releasing tools and then merging this PR

apeltzer:
> > Would vote for merging this then after some sanity checks? 🤠
> 
> I vote for updating template and releasing tools and then merging this PR

Or that way around ;-)

Emiller88:
I updated the template, and then realized I need to update the linter as well. Woof. :upside_down_face:

mahesh-panchal:
> ... (and in the context above a function that's just returning a string). The deferring of evaluation is only relevant from configuration files, so since this is from the script context, it does not have an effect.

So I just discovered this is not quite true, and oddly enough from the person I was explaining it to. 
https://grst.github.io/bioinformatics/2020/11/28/low-level-nextflow-hacking.html

ewels:
Some potential problems:

- Assumes that BioContainers always has the same base for docker + singularity. This would now be set at pipeline level and not module level, so if they start doing piecemeal different providers in the future, everything will break.
    - eg. if BioContainers moves _just_ FastQC to DockerHub, we have no way of dealing with this.
- Need to update tools in a way that it will work with both old versions of the syntax and new. It will take a long time for pipelines to update to this syntax (years).
    - Would be solved if we can have this at Nextflow level: https://github.com/nextflow-io/nextflow/issues/3340

I do not think that we should rush into merging this refactoring. It's a big change and it will cause a lot of pain if it goes ahead before we're fully ready.

mahesh-panchal:
> Assumes that BioContainers always has the same base for docker + singularity. This would now be set at pipeline level and not module level, so if they start doing piecemeal different providers in the future, everything will break.
> 
>     * eg. if BioContainers moves _just_ FastQC to DockerHub, we have no way of dealing with this.

The solution would be to change the default path in the module, but that would also mean needing to write a custom config with process specific container overwrite too as well as the params if needed.

muffato:
> Assumes that BioContainers always has the same base for docker + singularity. This would now be set at pipeline level and not module level, so if they start doing piecemeal different providers in the future, everything will break.
> * eg. if BioContainers moves _just_ FastQC to DockerHub, we have no way of dealing with this.

And the fact is that BioContainers are _already_ across quay.io and DockerHub. Legitimate modules may well refer to DockerHub for BioContainers

Emiller88:
Closing for now in favor of something along the lines of https://github.com/adamrtalbot/nextflow-customregistry

Leaving the branch just in case.

PR #2309, opened 2022-10-14
check for successful command completion

njspix:
Pileup might not always exit with non-zero status on errors. Add some bash to work around this.

### DIFF
--- a/modules/nf-core/biscuit/pileup/main.nf
+++ b/modules/nf-core/biscuit/pileup/main.nf
@@ -36,11 +36,11 @@ process BISCUIT_PILEUP {
         \$INDEX \\
         $input \\
         | bgzip -@ $bgzip_cpus $args2 > ${prefix}.vcf.gz
-    
-    # Pileup doesn't always exit with nonzero status in case of error. 
-    if grep -E "\\[main\\] Real time:" .command.err; then 
+
+    # Pileup doesn't always exit with nonzero status in case of error.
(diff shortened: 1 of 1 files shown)


nvnieuwk:
Standard output after a successful run contains something like this at the end:

´´´
[main] Real time: 0.046 sec; CPU: 0.047 sec
´´´

The module will always fail because of the code you added. Is there something else you could look for in the error message that defines a failure of the tool?

njspix:
Thanks for checking in. Unless I'm mistaken, log messages are printed to stderr, not stdout, so would appear in the .command.err file. 
The command can fail for different reasons, producing different error messages, and it might be difficult or impossible to capture all these by regex. I thought that conditioning on success would be a more robust approach. 
I'm unsure why the code is causing the module to fail; 
```
[main] Real time: 0.046 sec; CPU: 0.047 sec
```
also appears in stderr. Is my shellscript not standards-compliant, or am I missing something obvious? I've been using this on my system (CentOS7) for a while and I think it's been working fine...

nvnieuwk:
Ah, I understand. I've tried running the module in a gitpod environment to check what the problem with the tests was. 

Can you see that the module failed in the output files? Maybe you could add a check on the output files instead?

njspix:
I'm new to GitPod but I'm trying that as well, thanks for the tip!

Unfortunately the most likely result of failure is just a shorter-than-expected vcf, so there's really no way to tell from the output file.

nvnieuwk:
:crossed_fingers:

PR #2311, opened 2022-10-14
New Segemehl module

BarryDigby:
Segemehl index and align

- [x] Add a resource `label`


modules/nf-core/segemehl/align/main.nf
```diff
@@ -0,0 +1,54 @@
...
+        'https://depot.galaxyproject.org/singularity/segemehl:0.3.4--hc2ea5fd_5':
+        'quay.io/biocontainers/segemehl:0.3.4--hc2ea5fd_5' }"
+
+    input:
+    tuple val(meta), path(reads)
+    path(fasta)
+    path(index)
+
+    output:
+    tuple val(meta), path("${meta.id}/*"), emit: results
```
ramprasadn, line 16: Shouldn't this be prefix and not meta.id?

BarryDigby: Just tested it locally, 

`PROFILE=docker pytest --tag segemehl/align --symlink --keep-workflow-wd`

 - with `${prefix}`: failed
 - with `${meta.id}`: passed

🤔

modules/nf-core/segemehl/align/main.nf
```diff
@@ -0,0 +1,54 @@
...
+    """
+    mkdir -p $prefix
+
+    segemehl.x \\
+        -t $task.cpus \\
+        -d $fasta \\
+        -i $index \\
+        $reads \\
+        $args \\
+        -o ${prefix}/${prefix}.sam
```
ramprasadn, line 35: Looks like the tool could generate either a sam or a bam output so you should avoid hardcoding it to sam here. You could parse the args, find the appropriate format, and set the extension to that value. Something like this, https://github.com/nf-core/modules/blob/8ddc73f0a37e6be2e852fd828d3effe9734ffa62/modules/nf-core/ensemblvep/main.nf#L34-L43

tests/modules/nf-core/segemehl/align/test.yml
```diff
@@ -0,0 +1,25 @@
...
+  tags:
+    - segemehl
+    - segemehl/align
+  files:
+    - path: output/segemehl/test/test.sam
+      contains:
+        - "SRR5665260.1.36873006/1"
+        - "SRR5665260.1.28220282/1"
+        - "SRR5665260.1.12410541/1"
+        - "SRR5665260.1.23452355/1"
```
ramprasadn, line 25: Would be cool to include a test for stub-run as well 😄

modules/nf-core/segemehl/align/main.nf
BarryDigby: Pretty neat, thanks for sharing that. Torn between using `-b` or `--bam` in the config file, the former being an actual parameter of the tool, the latter being more human readable.

ramprasadn: huh.. I am not sure about this, but perhaps having something like `(args.contains("-b ") || args.contains("--bam ")) ? 'bam' : 'sam'` might work?

modules/nf-core/segemehl/align/main.nf
```diff
@@ -0,0 +1,55 @@
...
+    """
+    mkdir -p $prefix
+
+    segemehl.x \\
+        -t $task.cpus \\
+        -d $fasta \\
+        -i $index \\
+        $reads \\
+        $args \\
+        -o ${prefix}/${prefix}.{suffix}
```
ramprasadn, line 36:
```suggestion
        -o ${prefix}/${prefix}.${suffix}
```

modules/nf-core/segemehl/align/main.nf
```diff
@@ -0,0 +1,56 @@
...
+    path "versions.yml"                  , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def reads = meta.single_end ? "-q ${reads}" : "-q ${reads[0]} -p ${reads[1]}"
+    def suffix = args.contains("-b") ? "bam" : "sam"
```
ramprasadn, line 26:
```suggestion
    def suffix = (args.contains("-b ") || args.contains("--bamabafixoida")) ? 'bam' : 'sam'
```
Also, I ran some tests locally and '--bam' didn't work for me. But then I am not familiar with the tool so I don't if I screwed up my tests. Does `--bam` work for you?

modules/nf-core/segemehl/align/main.nf
ramprasadn:
```suggestion
     tuple val(meta), path("${prefix}/*"), emit: results
```
Sorry that I wasn't being very clear before. Because of the conditional you have in line 24, if a user defines his/her own prefix, then the prefix variable in the script will be set to that user-defined value and not meta.id. So when you have an user-defined prefix, if you are looking for a pattern that has the value of meta.id in it, you would miss capturing the output. But using `path("${prefix}/*")` should cover all scenarios.

modules/nf-core/segemehl/align/main.nf
ramprasadn: Also, I think it would be cool to include tests to make sure that the prefix and output format are set as expected. I am thinking something like [this](https://github.com/nf-core/modules/blob/master/tests/modules/nf-core/hmmer/hmmalign/main.nf).

BarryDigby: @ramprasadn I've added tests for bam file outputs that seem to be working well. 

Whilst adding a test for the prefix, I learned something interesting. I had to remove the `def` from the script preamble for nextflow to recognize `prefix` in the output declaration. The same behavior is observed here in `qualimap`:

https://github.com/nf-core/modules/blob/07c9080d7b66a325e8e67d9c57e10083ca8b9db4/modules/nf-core/qualimap/bamqc/main.nf#L15-L23

When using the canonical `def prefix = task.ext.prefix ?: "${meta.id}"`, I got the error `Missing output file(s) null/* expected by process test_segemehl_align_prefix:SEGEMEHL_PREFIX (test)`.

ramprasadn: That's right. I remember reading about this on nf-core slack a while ago. And my takeaway from that thread was that by adding def keyword in front of a variable we keep it in local scope. So if we want to access it outside of the script section then we need to omit `def`.

ramprasadn (approved):
Nice work! LGTM 🚀

matthdsm (approved):

PR #2318, opened 2022-10-17
Dedup subworkflow name

SPPearce:
I propose that we rename the `bam_dedup_stats_samtools_umitools` subworkflow to be just `bam_dedup_umitools`.
[Various](https://github.com/nf-core/modules/blob/a52cd7df53f3e88895362f2c8fd43cfe00fbf1a3/subworkflows/nf-core/fastq_align_bowtie2/main.nf#L6) [other](https://github.com/nf-core/modules/blob/master/subworkflows/nf-core/bam_markduplicates_picard/main.nf) [subworkflows](https://github.com/nf-core/modules/blob/master/subworkflows/nf-core/fastq_align_hisat2/main.nf) generate stats using `samtools`, without mentioning it in the name of the subworkflow.

I think the main tool should be mentioned in the title, not the auxiliary tools.


nvnieuwk (approved):

matthdsm:
Don't forget to edit `pytest_modules.yml` !

matthdsm (changes requested):
Also edit `tests/config/pytest_modules.yml`

matthdsm (approved):

Emiller88 (approved):
We might wait for @drpatelh's thoughts on this before we merge.

matthdsm:
That's a toughy since it's both samtools...

SPPearce:
> There's also `bam_sort_stats_samtools`? What do we want to do about that one?

That one does actually sort the bam and the gets the stats using `samtools`, so I'm less fussed about that.

matthdsm:
I'd even consider dropping `stats` from all the various workflows. I don't see much value in running it a million times when using subworkflows.
I would much rather have 1 qc workflow/step I could put where I want, without being forced into running `stats` every time.

matthdsm:
⏫ at some point we're just duplicating code IMO

Emiller88:
Yeah, that's the tough part with subworkflows, it's opinionated. We can't get down to something atomic like the modules, it's just going to have to be community consensus. But without the stats, you're just basically running `UMITOOLS_DEDUP ` :shrug:

nvnieuwk:
Maybe we could add an option to each subworkflow that defines whether the subworkflow runs the stats or not?

matthdsm:
I think we should get to a point where’s there guidelines as to what warrants a subworkflow so we dont get a bunch of tool x + stats or whatever.
Op 17 okt. 2022 om 16:22 heeft nvnieuwk ***@***.***> het volgende geschreven:

﻿

Maybe we could add an option to each subworkflow that defines whether the subworkflow runs the stats or not?

nvnieuwk:
Yep that would be useful :)

drpatelh:
Agree we should maybe have an option to switch the stats off. Still think we should keep it in the name though otherwise what distinguishes it from another subworkflow that doesn't include this at all? We should be as explicit as possible about what the subworkflow is actually doing so it's easier to find, use and reuse if required with other tools that could be drop in replacements. Happy to have another discussion about this next week maybe on a maintainers call?

SPPearce:
`bam_dedup_stats_samtools_umitools` sounds like it might be using [samtools markdup](http://www.htslib.org/doc/samtools-markdup.html), whereas the actual important part of the name is `umitools`.

matthdsm:
@drpatelh , @SPPearce 
What's the status of this PR? Are we good to merge or should we convert to draft and/or close?
We're cleaning up open PR's so we have nice place to start the upcoming Hackathon!

Cheers
Matthias - nf-core maintainers

SPPearce:
I'm still of the opinion that we should rename this subworkflow, and think that the other subworkflow names in the recent months are on the side of renaming it.

matthdsm:
We might want to merge https://github.com/nf-core/modules/pull/2979 before we proceed though

SPPearce:
Yes, and probably #2971 as well.

maxulysse:
What's happening with this PR, should we drop it and restart from scratch?

SPPearce:
Probably best to close this one, yes.
Are we renaming the subworkflow then?

maxulysse:
> Are we renaming the subworkflow then?

Let's make an issue for that, and discuss there, and then if needed we can have a PR

Issue #2316, opened 2022-10-17
new module: atlas-gene-annotation-manipulation/gtf2featureAnnotation


pinin4fjords:
### Is there an existing module for this?

- [X] I have searched for the existing module

### Is there an open PR for this?

- [X] I have searched for existing PRs

### Is there an open issue for this?

- [X] I have searched for existing issues

### Are you going to work on this?

- [X] If I'm planning to work on this module, I added myself to the `Assignees` to facilitate tracking who is working on the module

pinin4fjords:
See chat at https://nfcore.slack.com/archives/CJRH30T6V/p1665413183508719

PR #2321, opened 2022-10-17
Add gtf2featureannotation module

pinin4fjords:
## PR checklist

Adding module wrapper for https://github.com/ebi-gene-expression-group/atlas-gene-annotation-manipulation. See also https://nfcore.slack.com/archives/CJRH30T6V/p1665413183508719.

Apologies for the long name, didn't quite think that through at the time I wrote this tool!

Closes https://github.com/nf-core/modules/issues/2316

- [x] Add a resource `label`


pinin4fjords:
@nf-core-bot fix linting

modules/nf-core/atlasgeneannotationmanipulation/gtf2featureannotation/main.nf
```diff
@@ -0,0 +1,41 @@
...
+    tag '$gtf'
+    label 'process_single'
+
+    conda (params.enable_conda ? "bioconda::atlas-gene-annotation-manipulation=1.1.0" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/atlas-gene-annotation-manipulation%3A1.1.0--hdfd78af_0':
+        'quay.io/biocontainers/atlas-gene-annotation-manipulation:1.1.0--hdfd78af_0' }"
+
+    input:
+    path gtf
```
ramprasadn, line 11: I think you should add a meta map here because process inputs in nf-core pipelines typically carry sample specific information in them. 
https://nf-co.re/docs/contributing/modules#what-is-the-meta-map

modules/nf-core/atlasgeneannotationmanipulation/gtf2featureannotation/main.nf
```diff
@@ -0,0 +1,41 @@
...
+    path fasta
+
+    output:
+    path "feature_annotation.tsv"   , emit: feature_annotation
+    path "versions.yml"             , emit: versions
+    path "*.fa.gz"                  , emit: filtered_cdna, optional: true
```
ramprasadn, line 17: If you decide on adding meta to your input gtf channel, I'd also add meta to the ouput channel definitions here.

modules/nf-core/atlasgeneannotationmanipulation/gtf2featureannotation/main.nf
pinin4fjords: This is a gtf I imagined being kind of global to a workflow, which is why I didn't do that.

But maybe there could be multiple species/ annotation types. I'll add the meta.

pinin4fjords:
Thanks for review @ramprasadn! Requested changes done.

modules/nf-core/atlasgeneannotationmanipulation/gtf2featureannotation/main.nf
```diff
@@ -0,0 +1,43 @@
...
+    label 'process_single'
+
+    conda (params.enable_conda ? "bioconda::atlas-gene-annotation-manipulation=1.1.0" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/atlas-gene-annotation-manipulation%3A1.1.0--hdfd78af_0':
+        'quay.io/biocontainers/atlas-gene-annotation-manipulation:1.1.0--hdfd78af_0' }"
+
+    input:
+    val meta
+    path gtf
```
ramprasadn, line 12: I'd say its better to pack meta and gtf as a tuple [like this](https://github.com/nf-core/modules/blob/master/modules/nf-core/abacas/main.nf) because that's the default output structure in modules. That way if I want to pass an output from another process to this one, I can pass it as it is. Otherwise I would need to manipulate the channel to create a separate meta and gtf channels which is unnecessary.

modules/nf-core/atlasgeneannotationmanipulation/gtf2featureannotation/main.nf
```diff
@@ -0,0 +1,43 @@
...
+    tuple val(meta), path("*.anno.tsv")    , emit: feature_annotation
+    tuple val(meta), path("*.fa.gz")                    , emit: filtered_cdna, optional: true
+    tuple val(meta), path("versions.yml")               , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: ( meta.id ?: '' )
```
ramprasadn, line 25:
```suggestion
    def prefix = task.ext.prefix ?: meta.id 
```
As far as I know meta map always has an id (either sample or case) so in case the gtf doesn't have that I think we would want to know. Otherwise in this case, we might just end up creating a hidden file by the name ".anno.tsv" if meta.id isn't set 😅

modules/nf-core/atlasgeneannotationmanipulation/gtf2featureannotation/main.nf
```diff
@@ -0,0 +1,43 @@
+process ATLASGENEANNOTATIONMANIPULATION_GTF2FEATUREANNOTATION {
+    tag '$gtf'
```
ramprasadn, line 2:
```suggestion
    tag '${meta.id}'
```
This would be good in print the case/sample id in the log files in case someone wants to run this process multiple times in the same pipeline, no?

ramprasadn (commented):
Just a couple more minor suggestions. :)

modules/nf-core/atlasgeneannotationmanipulation/gtf2featureannotation/main.nf
pinin4fjords: Another good point- thanks!

pinin4fjords:
Thanks for the guidance @ramprasadn - how's this?

modules/nf-core/atlasgeneannotationmanipulation/gtf2featureannotation/main.nf
```diff
@@ -0,0 +1,42 @@
...
+    label 'process_single'
+
+    conda (params.enable_conda ? "bioconda::atlas-gene-annotation-manipulation=1.1.0" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/atlas-gene-annotation-manipulation%3A1.1.0--hdfd78af_0':
+        'quay.io/biocontainers/atlas-gene-annotation-manipulation:1.1.0--hdfd78af_0' }"
+
+    input:
+    tuple val(meta), path(gtf)
+    tuple val(meta), path(fasta)
```
ramprasadn, line 12: You should change it to  
```suggestion
    tuple val(meta2), path(fasta)
```
Otherwise it would clash with the meta information in the gtf channel.

modules/nf-core/atlasgeneannotationmanipulation/gtf2featureannotation/meta.yml
```diff
@@ -0,0 +1,31 @@
...
+tools:
+  - "atlasgeneannotationmanipulation":
+      description: "Scripts for manipulating gene annotation"
+      homepage: "https://github.com/ebi-gene-expression-group/atlas-gene-annotation-manipulation"
+      documentation: "https://github.com/ebi-gene-expression-group/atlas-gene-annotation-manipulation"
+      tool_dev_url: "https://github.com/ebi-gene-expression-group/atlas-gene-annotation-manipulation"
+      licence: "['Apache-2.0']"
+
+input:
+  - gtf:
```
ramprasadn, line 15: I suggest adding information about the meta map as well. [ex](https://github.com/nf-core/modules/blob/master/modules/nf-core/abacas/meta.yml)

modules/nf-core/atlasgeneannotationmanipulation/gtf2featureannotation/meta.yml
```diff
@@ -0,0 +1,31 @@
...
+      type: file
+      description: gtf annotation file
+      pattern: "*.{gtf}"
+
+output:
+  - versions:
```
ramprasadn, line 21: You should add information about the `filtered_cdna` output as well.

ramprasadn:
Module looks good, just a few changes to the docs and it should be 👌🏻

pinin4fjords:
> Module looks good, just a few changes to the docs and it should be 👌🏻

Done!

ramprasadn (approved):
Nice work! LGTM 🚀

ggabernet:
As I understand it, this module is for converting a GTF file to an annotation table. So within a pipeline, it will be called just once on the GTF file provided as reference. Actually in the case of operating on reference data the meta maps are not needed, as there is no sample-specific information to be forwarded to the module itself. See for example some genome indexing modules as the `salmon index` one (https://github.com/nf-core/modules/blob/master/modules/nf-core/salmon/index/main.nf).

pinin4fjords:
> As I understand it, this module is for converting a GTF file to an annotation table. So within a pipeline, it will be called just once on the GTF file provided as reference. Actually in the case of operating on reference data the meta maps are not needed, as there is no sample-specific information to be forwarded to the module itself. See for example some genome indexing modules as the `salmon index` one (https://github.com/nf-core/modules/blob/master/modules/nf-core/salmon/index/main.nf).

Haha- well, that was my original thinking, but you and @ramprasadn will have to fight it out :-D

ggabernet:
ok well if you can see the use to the `meta` map in the module, then it can always be provided as an empty list to the channel if not needed 👍

ggabernet (approved):
Looks good to me!

pinin4fjords:
Thanks for reviews @ggabernet @ramprasadn !

PR #1874, opened 2022-07-12
New module: vsearch/cluster

mirpedrol:
## PR checklist

Closes [#1747](https://github.com/nf-core/modules/issues/1747) 

- [x] This comment contains a description of changes (with reason).
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [x] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [x] Remove all TODO statements.
- [x] Emit the `versions.yml` file.
- [x] Follow the naming conventions.
- [x] Follow the parameters requirements.
- [x] Follow the input/output options guidelines.
- [x] Add a resource `label`
- [x] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - [ ] `PROFILE=docker pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [ ] `PROFILE=singularity pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [ ] `PROFILE=conda pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`

PR #2324, opened 2022-10-18
adding odgi/stats module

subwaystation:
## PR checklist

Contributes to https://github.com/nf-core/pangenome/issues/70.

- [x] Add a resource `label`

This PR adds the odgi/stats module which is required for the https://github.com/nf-core/pangenome pipeline.

### DIFF
--- a/modules/nf-core/odgi/stats/main.nf
+++ b/modules/nf-core/odgi/stats/main.nf
@@ -1,20 +1,3 @@
-// TODO nf-core: If in doubt look at other nf-core/modules to see how we are doing things! :)
-//               https://github.com/nf-core/modules/tree/master/modules
-//               You can also ask for help via your pull request or on the #modules channel on the nf-core Slack workspace:
-//               https://nf-co.re/join
-// TODO nf-core: A module file SHOULD only define input and output files as command-line parameters.
-//               All other parameters MUST be provided using the "task.ext" directive, see here:
-//               https://www.nextflow.io/docs/latest/process.html#ext
-//               where "task.ext" is a string.
-//               Any parameters that need to be evaluated in the context of a particular sample
-//               e.g. single-end/paired-end data MUST also be defined and evaluated appropriately.
-// TODO nf-core: Software that can be piped together SHOULD be added to separate module files
-//               unless there is a run-time, storage advantage in implementing in this way
-//               e.g. it's ok to have a single module for bwa to output BAM instead of SAM:
-//                 bwa mem | samtools view -B -T ref.fasta
-// TODO nf-core: Optional inputs are not currently supported by Nextflow. However, using an empty
-//               list (`[]`) instead of a file can be used to work around this issue.
-
 process ODGI_STATS {
     tag "$meta.id"
     label 'process_single'
--- a/modules/nf-core/odgi/stats/main.nf
+++ b/modules/nf-core/odgi/stats/main.nf
@@ -12,7 +12,7 @@ process ODGI_STATS {
     val(suffix)
 
     output:
-    tuple val(meta), path("${prefix}"."${suffix}"), emit: odgi_stats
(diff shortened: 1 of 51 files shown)


subwaystation:
@heuermh @ggabernet Could you please take a look at this?

The nf-core linting fails, because the md5sum is of an empty file. Which in this case is true, because running `odgi stats` without any arguments emits only an empty file.

However, I am puzzled about the failing workflows. What's the problem here? Thanks for any help!

ggabernet:
It seems like your `test.yml` does not follow the specifications. How did you create it? Can you try re-creating it with the `nf-core modules create-test-yml` command?

subwaystation:
It was automatically created when I did `nf-core modules create odgi/stats` following https://nf-co.re/developers/tutorials/dsl2_modules_tutorial#create-the-module-template.

subwaystation:
How shall I fix the md5sum problem @ggabernet? I can add an argument to this test case, so it does not produce an empty file. Would that be fine? Or is there a way to ignore it?

subwaystation:
As suggested by @mirpedrol I just don't check for the md5sum and only for file existence.

modules/nf-core/odgi/stats/main.nf
```diff
@@ -0,0 +1,36 @@
+process ODGI_STATS {
+    tag "$meta.id"
+    label 'process_single'
+
+    conda (params.enable_conda ? "bioconda::odgi=0.8.0" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/odgi:0.8.0--py39h2add14b_0':
+        'quay.io/biocontainers/odgi:0.8.0--py310hc8f18ef_0' }"
```
mirpedrol, line 8: The container version is different, is it possible to provide the same?

modules/nf-core/odgi/stats/main.nf
```diff
@@ -0,0 +1,36 @@
...
+
+    input:
+    tuple val(meta), path(graph)
+    val(suffix)
```
mirpedrol, line 12: Should be passed through `$args` (see https://nf-co.re/docs/contributing/modules#general) or maybe you can obtain it from the parameters in `$args` if I understand the documentation right

mirpedrol (commented):
Looks good! I only added a couple of comments :)

modules/nf-core/odgi/stats/main.nf
subwaystation: I thought the module itself should be able to run through without any optional arguments from `$args`. So the argument is mandatory in the sense that the module just won't run properly. But let me change.

modules/nf-core/odgi/stats/main.nf
subwaystation: There is no py310 version on singularity available, so let me downgrade.

modules/nf-core/odgi/stats/main.nf
subwaystation: @ggabernet I discussed a lot with @mirpedrol and the usual way would be to do something like:
`def suffix = args.contains("--tsv") ? ".tsv" : ".yaml"`. However, I already need the `suffix` when describing the outputs of the module. So using `$args` is not an option. Can I still specify an additional mandatory input channel as shown in the tutorial https://nf-co.re/developers/tutorials/dsl2_modules_tutorial#create-a-test-workflow?

What other solutions would you suggest? Thanks for any help!

subwaystation: As shown by @mirpedrol I am following her example https://github.com/nf-core/modules/pull/1874#pullrequestreview-1147515478 and use optional outputs. This is a straightforward solution and eliminates the requirement of an extra input channel.

mirpedrol (approved):
LGTM 🚀

PR #1149, opened 2021-12-09
Adding new module gfaffix

heuermh:
Fixes #1134 

## PR checklist

Closes #XXX 

- [ ] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [ ] Remove all TODO statements.
- [ ] Emit the `versions.yml` file.
- [ ] Follow the naming conventions.
- [ ] Follow the parameters requirements.
- [ ] Follow the input/output options guidelines.
- [ ] Add a resource `label`
- [ ] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
    - [ ] `PROFILE=docker pytest --tag <MODULE> --symlink --keep-workflow-wd`
    - [ ] `PROFILE=singularity pytest --tag <MODULE> --symlink --keep-workflow-wd`
    - [ ] `PROFILE=conda pytest --tag <MODULE> --symlink --keep-workflow-wd`

PR #2329, opened 2022-10-18
new odgi/viz module

subwaystation:
## PR checklist

This contributes to https://github.com/nf-core/pangenome/issues/70.

- [x] Add a resource `label`

This PR adds a new module `odgi viz` required for the https://github.com/nf-core/pangenome pipeline.

### DIFF
--- a/modules/nf-core/odgi/viz/main.nf
+++ b/modules/nf-core/odgi/viz/main.nf
@@ -23,7 +23,6 @@ process ODGI_VIZ {
     """
     odgi \\
         viz \\
-        --threads 1 \\
         --idx ${graph} \\
         --out ${prefix}.png \\
         $args
--- a/modules/nf-core/odgi/viz/main.nf
+++ b/modules/nf-core/odgi/viz/main.nf
@@ -23,6 +23,7 @@ process ODGI_VIZ {
     """
     odgi \\
         viz \\
+        --threads $task.cpus \\
         --idx ${graph} \\
         --out ${prefix}.png \\
         $args
--- a/modules/nf-core/odgi/viz/main.nf
+++ b/modules/nf-core/odgi/viz/main.nf
@@ -1,0 +1,36 @@
+process ODGI_VIZ {
+    tag "$meta.id"
+    label 'process_single'
+
+    conda (params.enable_conda ? "bioconda::odgi=0.8.0" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/odgi:0.8.0--py39h2add14b_0':
+        'quay.io/biocontainers/odgi:0.8.0--py310hc8f18ef_0' }"
+
+    input:
+    tuple val(meta), path(graph)
+
+    output:
+    tuple val(meta), path("*.png"), emit: odgi_viz
+    path "versions.yml"           , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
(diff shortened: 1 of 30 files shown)


modules/nf-core/odgi/viz/main.nf
```diff
@@ -0,0 +1,36 @@
+process ODGI_VIZ {
+    tag "$meta.id"
+    label 'process_single'
```
nvnieuwk, line 3: Is this the best label to use since you can define the amount of cores used?

nvnieuwk (changes requested):

modules/nf-core/odgi/viz/main.nf
subwaystation: Ah. If a module can only make use of 1 thread, I enter it directly into the `main.nf`?

nvnieuwk: Some modules are only able to use one core (that's why the `process_single` label exists). I don't know the tool this module is from. Is it a process intensive tool or do you think it will run well using just one module?

The label `process_single` specifies that the module will only get one core so you don't have to define it manually

subwaystation: Better now?

subwaystation: It can become memory intensive, but it can only make use of a single thread.

SPPearce: If the tool can be multithreaded, then allow it to be multithreaded by passing the `$task.cpus`. 
The `process_single` tag has a default setting of 6GB RAM, 1 core. 
The `process_low` tag is 12GB RAM, 2 cores and further scaling occurs with `process_medium` and `process_high`
So if the tool is likely to run out of RAM at 6GB, then default to a larger tag. But definitely using the $task.cpus.

subwaystation: So I removed `--threads 1` completely. Should be fine now, right?

nvnieuwk: you should keep `--threads $task.cpus` :) This allows other users to change the amount of cores the program uses

subwaystation: But it can only use 1 thread. Dependent on the input it might require several megabytes up to hundreds of gigabytes of RAM.

subwaystation: I also implemented parts of `odgi viz` so I know what's going on.

nvnieuwk: The RAM usage can be specified by the user in the configuration file. From what I hear from you, the `process_single` label is the best choice indeed :)

nvnieuwk (approved):

modules/nf-core/odgi/viz/main.nf
subwaystation: > If the tool can be multithreaded, then allow it to be multithreaded by passing the `$task.cpus`. The `process_single` tag has a default setting of 6GB RAM, 1 core. The `process_low` tag is 12GB RAM, 2 cores and further scaling occurs with `process_medium` and `process_high` So if the tool is likely to run out of RAM at 6GB, then default to a larger tag. But definitely using the $task.cpus.

@SPPearce We had the same discussion in https://github.com/nf-core/modules/pull/1149. So we could go down the same route and just keep `process_single`?

SPPearce: Yeah, I remember this now. `process_single` but with setting the threads so that the tool will use more if they are available is the way to go.

PR #2352, opened 2022-10-20
add module odgi/layout

subwaystation:
## PR checklist

This PR progresses https://github.com/nf-core/pangenome/issues/70.

- [x] Add a resource `label`

Adding a new module which is able to generate a 2D layout of a pangenome graph with `odgi layout`.

### DIFF
--- a/modules/nf-core/odgi/layout/main.nf
+++ b/modules/nf-core/odgi/layout/main.nf
@@ -9,6 +9,7 @@ process ODGI_LAYOUT {
 
     input:
     tuple val(meta), path(graph)
+    val(prefix)
 
     output:
     tuple val(meta), path("*.lay"), optional: true, emit: lay
--- a/modules/nf-core/odgi/layout/main.nf
+++ b/modules/nf-core/odgi/layout/main.nf
@@ -19,8 +19,7 @@ process ODGI_LAYOUT {
     task.ext.when == null || task.ext.when
 
     script:
-    def args = task.ext.args ?: ''
-    def prefix = task.ext.prefix ?: "${meta.id}"
+    def args = task.ext.args ?: "--out test.lay"
     """
     odgi \\
         layout \\
--- a/modules/nf-core/odgi/layout/main.nf
+++ b/modules/nf-core/odgi/layout/main.nf
@@ -19,7 +19,7 @@ process ODGI_LAYOUT {
     task.ext.when == null || task.ext.when
 
     script:
-    def args = task.ext.args ?: "--out test.lay"
+    def args = task.ext.args ?: "--out ${meta.id}.lay"
     """
     odgi \\
         layout \\
--- a/modules/nf-core/odgi/layout/main.nf
+++ b/modules/nf-core/odgi/layout/main.nf
@@ -9,7 +9,6 @@ process ODGI_LAYOUT {
 
     input:
     tuple val(meta), path(graph)
-    val(prefix)
 
     output:
     tuple val(meta), path("*.lay"), optional: true, emit: lay
--- a/modules/nf-core/odgi/layout/main.nf
+++ b/modules/nf-core/odgi/layout/main.nf
@@ -1,0 +1,35 @@
+process ODGI_LAYOUT {
+    tag "$meta.id"
+    label 'process_medium'
+
+    conda (params.enable_conda ? "bioconda::odgi=0.8.0" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/odgi:0.8.0--py310hc8f18ef_0':
+        'quay.io/biocontainers/odgi:0.8.0--py310hc8f18ef_0' }"
+
+    input:
+    tuple val(meta), path(graph)
+
+    output:
+    tuple val(meta), path("*.lay"), optional: true, emit: lay
+    tuple val(meta), path("*.tsv"), optional: true, emit: tsv
+    path "versions.yml"                           , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    odgi \\
+        layout \\
+        --threads $task.cpus \\
+        --idx ${graph} \\
+        $args
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        odgi: \$(echo \$(odgi version 2>&1) | cut -f 1 -d '-' | cut -f 2 -d 'v')
+    END_VERSIONS
+    """
+}
--- a/modules/nf-core/odgi/layout/meta.yml
+++ b/modules/nf-core/odgi/layout/meta.yml
@@ -24,6 +24,9 @@ input:
       type: file
       description: Pangenome graph in GFA v1.0 format or in ODGI binary format
       pattern: "*.{gfa,og}"
+  - prefix:
+      type: string
+      description: Prefix for output files
 
 output:
   - meta:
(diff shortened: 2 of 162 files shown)


nvnieuwk (changes requested):
How exactly can you specify the output format?

Can you also add a test that outputs a TSV file?

modules/nf-core/odgi/layout/main.nf
```diff
@@ -0,0 +1,35 @@
...
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    odgi \\
+        layout \\
+        --threads $task.cpus \\
+        --idx ${graph} \\
+        $args
```
nvnieuwk, line 29:
```suggestion
        --out ${prefix}.lay \\
        --tsv ${prefix}.tsv \\
        $args
```

tests/modules/nf-core/odgi/layout/nextflow.config
```diff
@@ -0,0 +1,9 @@
+process {
+
+    publishDir = { "${params.outdir}/${task.process.tokenize(':')[-1].tokenize('_')[0].toLowerCase()}" }
+
+
+    withName: ODGI_LAYOUT {
+        ext.args = '--out test.lay --tsv test.tsv'
+    }
```
nvnieuwk, line 8:
```suggestion
```

nvnieuwk (changes requested):
The output files should be created in the module itself using the prefix

modules/nf-core/odgi/layout/main.nf
subwaystation: But the output is optional. It could the module should only output a TSV and now LAY.
Creating the files directly in the module would mean we always create both files.

modules/nf-core/odgi/layout/main.nf
```diff
@@ -0,0 +1,35 @@
...
+        'https://depot.galaxyproject.org/singularity/odgi:0.8.0--py310hc8f18ef_0':
+        'quay.io/biocontainers/odgi:0.8.0--py310hc8f18ef_0' }"
+
+    input:
+    tuple val(meta), path(graph)
+
+    output:
+    tuple val(meta), path("*.lay"), optional: true, emit: lay
+    tuple val(meta), path("*.tsv"), optional: true, emit: tsv
+    path "versions.yml"                           , emit: versions
```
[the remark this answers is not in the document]
subwaystation: Since either lay, tsv, or both outputs are possible, we have to keep these channels optional.

modules/nf-core/odgi/layout/main.nf
nvnieuwk: That's okay, other modules do this too. It's up to the pipeline developer to choose what to save :)

modules/nf-core/odgi/layout/main.nf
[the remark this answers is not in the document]
nvnieuwk: Optional output means that the specified files aren't always part of the process. Since the tool will now create both files every time, it will always be there and isn't optional anymore

modules/nf-core/odgi/layout/main.nf
subwaystation: To give you a concrete example:

I have a pangenome graph and want to visualize it in 2D. So I do `odgi layout` with LAY output. Then I use `odgi draw` and have a 2D PNG of the graph.
Or I want to interactively explore the graph with e.g. https://github.com/chfi/gfaestus. Then I would only require the TSV output of `odgi layout`.

It is really rare I want both. We want to save disk space and IO here.

modules/nf-core/odgi/layout/main.nf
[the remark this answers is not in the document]
subwaystation: Yes, but I want these to stay optional :)

modules/nf-core/odgi/layout/main.nf
nvnieuwk: Yeah I understand what you mean! Maybe ask a second opinion on this, but I've always done it this way :)

subwaystation:
@nvnieuwk With my explanations, this PR is fine? ;)

JoseEspinosa (commented):
If you want to make the output optional, you can take an approach similar to the muscle module, see [here](https://github.com/nf-core/modules/blob/5e34754d42cd2d5d248ca8673c0a53cdf5624905/modules/nf-core/muscle/main.nf#L30-L36). I agree that especially if the files are big enough it is better just to produce one of them provided that this can be controlled by the tool arguments.

subwaystation:
That's a clever way. I will add accordingly.

subwaystation:
But then how to remove the arguments already in `$args`? Or is muscle able to ignore these?

subwaystation:
Assuming I have `--tsv` in args then I would have later in the cli:
```--tsv --tsv PREFIX.tsv
```

How would you solve this?

nvnieuwk:
That's mostly the reason why I wouldn't bother in using `args`. What's the size of the created files usually? Is it a huge problem if they both were created?

subwaystation:
There is no usual file size in pangenomics. For a 92 haplotypes HLA pangenome graph the file sizes are:
```
 - .og: 105MB
 - .lay:  8MB
 - .tsv: 20MB
 ```

However, the ODGI file for a full human pangenome graph from 92 haplotypes of all chromosomes is 120GB. So for this specific data set the TSV could be ~30GB. In the future we want to build graphs from ~300 haplotypes. So yes, the IO will be significant.

I think it is a huge problem that they are both created. And I think a module's functionality should stay as close to the underlying tool as possible.

**Comment by nvnieuwk** [comment]
The prefix should always be specified inside of the config file and should default to `meta.id`

File: modules/nf-core/odgi/layout/meta.yml [hunk 4]
```diff
@@ -0,0 +1,48 @@
...
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. [ id:'test', single_end:false ]
+  - graph:
+      type: file
+      description: Pangenome graph in GFA v1.0 format or in ODGI binary format
+      pattern: "*.{gfa,og}"
+
```
d4straub, line 27:
```suggestion
  - prefix:
      type: string
      description: Prefix for output files

```

subwaystation:
> The prefix should always be specified inside of the config file and should default to `meta.id`

@nvnieuwk But that's exactly not possible here. I need the prefix outside of the process. Luckily you said `should` ;)

nvnieuwk:
> @nvnieuwk But that's exactly not possible here. I need the prefix outside of the process. Luckily you said `should` ;)

After looking at this a bit closer, I think you don't even need a prefix. You can access the meta inside of the `args` and specify the prefix inside of the `args`

tests/modules/nf-core/odgi/layout/test.yml
```diff
@@ -0,0 +1,10 @@
+- name: "odgi layout"
+  command: nextflow run ./tests/modules/nf-core/odgi/layout -entry test_odgi_layout -c ./tests/config/nextflow.config -c ./tests/modules/nf-core/odgi/layout/nextflow.config
+  tags:
+    - "odgi"
+    - "odgi/layout"
+  files:
+    - path: "output/odgi/test.lay"
+    - path: "output/odgi/test.tsv"
+    - path: "output/odgi/versions.yml"
+      md5sum: 7705f6f91ca7f5cfe8665c6581694ea1
```
nvnieuwk, line 10:
```suggestion
```

nvnieuwk (changes requested):
I'll approve after these changes if you agree :)

nvnieuwk:
With these changes you can adjust the prefix inside `args` yourself

nvnieuwk (approved):

nvnieuwk:
Awesome! Hit it :partying_face:

PR #2393, opened 2022-10-24
Let DESeq2 module be flexible with CSV/ TSV, and use semicolons in blocking field

pinin4fjords:
This PR:

- Adds a basic method for being flexible with input files (CSV or TSV). There's a lot of inconsistency in what's used, and maybe it's better if the module is just flexible. This makes it respond to file extensions.
- Fixes a previous assumption about a comma-separated field in a comma-separated file. I think it's better if the blocking definition is semicolon-separated where multiple values are supplied, so that things don't break when that is stored in a CSV file like [the example](https://github.com/nf-core/test-datasets/blob/modules/data/genomics/mus_musculus/rnaseq_expression/SRP254919.contrasts.csv).

## PR checklist

- [x] Add a resource `label`

### DIFF
--- a/modules/nf-core/deseq2/differential/meta.yml
+++ b/modules/nf-core/deseq2/differential/meta.yml
@@ -19,11 +19,11 @@ input:
   - sample:
       type: file
       description: |
-        CSV-format sample sheet with sample metadata
+        CSV or TSV format sample sheet with sample metadata
   - counts:
       type: file
       description: |
-        Raw TSV-format expression matrix as output from the nf-core RNA-seq workflow
+        Raw TSV or CSV format expression matrix as output from the nf-core RNA-seq workflow
   - contrast_meta:
       type: map
       description: |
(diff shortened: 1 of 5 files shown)


Emiller88 (changes requested):
Looks good to me. I don't know if there's a more built-in way to do this than defining a function in R, but as long as it works.

Actually, could you add a test for a tsv? Just to make sure this functionality is covered and anything in the future doesn't break it.

pinin4fjords:
> Looks good to me. I don't know if there's a more built-in way to do this than defining a function in R, but as long as it works.

Will be glad to hear of a tidy way that doesn't involve additional libraries! It's a bit hacky I'll admit, but a thin layer around `read.delim()`, so not too naughty I think.

> Actually, could you add a test for a tsv? Just to make sure this functionality is covered and anything in the future doesn't break it.

Good idea. I've added an extra test that converts the matrix to CSV and tries it again. You'll see that the DESeq outputs in the test.yml are identical in their hashes.

Emiller88 (approved):
Awesome! Great work!

pinin4fjords:
Thanks for the review @Emiller88 !

PR #2414, opened 2022-10-28
Add fastq_align_chromap subworkflow

JoseEspinosa:
Adding in a subworkflow for FASTQ ALIGN using CHROMAP

## PR checklist

Closes #2413 

- [x] This comment contains a description of changes (with reason).
- [x] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [x] Remove all TODO statements.
- [x] Emit the `versions.yml` file.
- [x] Follow the naming conventions.
- [x] Follow the parameters requirements.
- [x] Follow the input/output options guidelines.
- [x] Add a resource `label`
- [x] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - [x] `PROFILE=docker pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [x] `PROFILE=singularity pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [x] `PROFILE=conda pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`

PR #2416, opened 2022-10-28
Add meta to fasta reference and index of chromap modules

JoseEspinosa:
As requested on #2414 add meta to reference channels of chromap modules.

### DIFF
--- a/modules/nf-core/chromap/chromap/main.nf
+++ b/modules/nf-core/chromap/chromap/main.nf
@@ -9,8 +9,8 @@ process CHROMAP_CHROMAP {
 
     input:
     tuple val(meta), path(reads)
-    tuple val(meta2), path (fasta)
-    tuple val(meta3), path(index)
+    path fasta
+    tuple val(meta2), path(index)
     path barcodes
     path whitelist
     path chr_order
--- a/modules/nf-core/chromap/chromap/main.nf
+++ b/modules/nf-core/chromap/chromap/main.nf
@@ -9,8 +9,8 @@ process CHROMAP_CHROMAP {
 
     input:
     tuple val(meta), path(reads)
-    path fasta
-    path index
+    tuple val(meta2), path (fasta)
+    tuple val(meta3), path(index)
     path barcodes
     path whitelist
     path chr_order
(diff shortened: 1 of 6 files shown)


modules/nf-core/chromap/chromap/main.nf
```diff
@@ -9,8 +9,8 @@ process CHROMAP_CHROMAP {
 
     input:
     tuple val(meta), path(reads)
-    path fasta
-    path index
+    tuple val(meta2), path (fasta)
```
matthdsm, line 12: nitpicking here, but shouldn't meta2 == meta3 here? I assume the aligner index is based on the the fasta file?

JoseEspinosa: yes, it's the same, what do you suggest that we get rid of the meta of the `fasta` file?

matthdsm: I tend to use the same variable for both, since they should be the same. Not that it matters much in to module, but it does indicate we expect it to be the same.

JoseEspinosa: 👉 [cfdc086](https://github.com/nf-core/modules/pull/2416/commits/cfdc086f72344be9051b09903af043efbdd41660)

modules/nf-core/chromap/index/meta.yml
```diff
@@ -15,6 +15,11 @@ tools:
       licence: ["GPL v3"]
 
 input:
+  - meta:
```
jfy133, line 18: missing meta2?

JoseEspinosa: This is for the indexing module and only has one meta for the `fasta` file, see [here](https://github.com/JoseEspinosa/nf-core-modules/blob/cfdc086f72344be9051b09903af043efbdd41660/modules/nf-core/chromap/index/main.nf#L10-L17)

modules/nf-core/chromap/chromap/main.nf
JoseEspinosa: OK, I have now used the same variable 👉 [1fc5a26](https://github.com/nf-core/modules/pull/2416/commits/1fc5a26ad04a6e7c64ace56f3eb29003d88d77a0)

JoseEspinosa:
I think I addressed your comments @matthdsm and @jfy133, would be great if you could give me your feedback 😄

matthdsm (approved):
Apologies for the delay!
LGTM 🚀

PR #2422, opened 2022-10-28
Adding module `sourmash taxonomy annotate`

### LINKED ISSUE #2261: new module: sourmash/taxonomy

vmikk:
## PR checklist

Closes #2261 (partially, as there are multiple commands, e.g., `sourmash tax annotate`, `sourmash tax genome`, `sourmash tax grep`, `sourmash tax metagenome`, `sourmash tax prepare`)

Currently, `sourmash tax annotate` does not support compressed input data (while the `sourmash gather` could output compressed files), therefore we use gunzip for now.

- [x] Add a resource `label`
  - [] `PROFILE=conda pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`

`PROFILE=conda` fails due to md5sum mismatch.


vmikk:
@taylorreiter , could you please check the module?
Currently, I've used taxonomy in CSV format. Does it make sense to run `sourmash tax prepare` to create SQL db prior the analysis?

Also, we can not use underscores in module names, therefore I used `taxannotate` instead of `tax_annotate`.

modules/nf-core/sourmash/taxannotate/meta.yml
```diff
@@ -0,0 +1,59 @@
...
+        Groovy Map containing sample information
+        e.g. [ id:'test', single_end:false ]
+  - gather_results:
+      type: file
+      description: |
+        Table with signatures classified as belonging to any of the genomes
+        in the sourmash database(s), result of `sourmash gather` command.
+  - taxonomy:
+      type: file
+      description: Database with lineages (in CSV format)
```
taylorreiter, line 36: Although it's not clear from the docs, the `--taxonomy` flag can take multiple taxonomy csv files, so maybe something like:
```suggestion
      description: One or more databases with lineages (in CSV format)
```

modules/nf-core/sourmash/taxannotate/main.nf
```diff
@@ -0,0 +1,51 @@
...
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args   = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+
+    # Currently, `sourmash tax annotate` does not support gz-compressed input,
+    # streaming is also not supported
+    # so we need to decomress the input file first
```
taylorreiter, line 28:
```suggestion
    # so we need to decompress the input file first
```

tests/modules/nf-core/sourmash/taxannotate/main.nf
```diff
@@ -0,0 +1,54 @@
...
+    database = [
+        [ id:'db', single_end:false ], // meta map
+        file(params.test_data['sarscov2']['genome']['genome_fasta'], checkIfExists: true)
+    ]
+
+    // Dummy taxonomy
+    File tax = new File("${workflow.workDir}/taxonomy.csv")
+    tax.write("ident,superkingdom,phylum,class,order,family,genus,species\n")
+    tax.append("db,d__Riboviria,p__Pisuviricota,c__Pisoniviricetes,o__Nidovirales,f__Coronaviridae,g__Betacoronavirus,s__SARS‑CoV‑2\n")
+    taxonomy = file(tax)
```
taylorreiter, line 28: ooooh very clever! I love this for the test file!

taylorreiter (changes requested):
This is really great! I have a few questions/typos.

modules/nf-core/sourmash/taxannotate/main.nf
```diff
@@ -0,0 +1,51 @@
...
+    label 'process_single'
+
+    conda (params.enable_conda ? "bioconda::sourmash=4.5.0" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/sourmash:4.5.0--hdfd78af_0':
+        'quay.io/biocontainers/sourmash:4.5.0--hdfd78af_0' }"
+
+    input:
+    tuple val(meta), path(gather_results)
+    path(taxonomy)
```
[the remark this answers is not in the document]
vmikk: Most likely, nf-core modules for sourmash will be used together (in a chain).
Currently, the output of the `gather` module is a single file.
Therefore, it's easier to provide it as input to `tax annotate`  via `--gather-csv`.
It would be nice to support `--from-file` as well. But it will require manipulating (`collect`ing) a channel that contains multiple gather results in the workflow. So this solution is not very general and depends on a project specifics.

So I think it's easier to use just a single input now.

taylorreiter (approved):
this looks really great! I made a couple comments on the docs but they aren't that big of a deal so i'm going to approve. Thank you for putting this together @vmikk! I'm excited to start using it soon 🎉

modules/nf-core/sourmash/taxannotate/main.nf
```diff
@@ -0,0 +1,43 @@
...
+
+    script:
+    def args   = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+
+    sourmash \\
+        tax annotate \
+        $args \\
+        --gather-csv ${gather_results} \\
```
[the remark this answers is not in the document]
vmikk: Yes, that's great that sourmash supports gzipped CSVs now!
I reported related issue in October (https://github.com/sourmash-bio/sourmash/issues/2337) and Titus fixed it literally the same day!

jfy133:
@nf-core-bot fix linting

modules/nf-core/sourmash/taxannotate/meta.yml
```diff
@@ -0,0 +1,59 @@
...
+input:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. [ id:'test', single_end:false ]
+  - gather_results:
+      type: file
+      description: |
+        Table with signatures classified as belonging to any of the genomes
```
jfy133, line 32:
```suggestion
        Mandatory table with signatures classified as belonging to any of the genomes
```

modules/nf-core/sourmash/taxannotate/meta.yml
```diff
@@ -0,0 +1,59 @@
...
+        in the sourmash database(s), result of `sourmash gather` command.
+  - taxonomy:
+      type: file
+      description: One or more databases with lineages (in CSV format)
```
jfy133, line 36:
```suggestion
      description: One or more databases with lineages (in CSV format, Mandatory)
```

modules/nf-core/sourmash/taxannotate/main.nf
```diff
@@ -0,0 +1,43 @@
+process SOURMASH_TAXANNOTATE {
+    tag "$meta.id"
+    label 'process_single'
+
+    conda (params.enable_conda ? "bioconda::sourmash=4.6.1" : null)
```
jfy133, line 5: Note that the `enable_conda` parameter has been deprecated, please check against the `master` on nf-core/modules to change this declaration.

modules/nf-core/sourmash/taxannotate/main.nf
```diff
@@ -0,0 +1,43 @@
...
+        tax annotate \
+        $args \\
+        --gather-csv ${gather_results} \\
+        --taxonomy ${taxonomy} \\
+        --output-dir "."
+
+    ## Output file name = ${prefix}.with-lineages.csv
+
+    ## Compress output
+    gzip "${prefix}.with-lineages.csv"
```
jfy133, line 36: I don't fully follow this - how does `sourmash` know to use `prefix` when it is not specifyied in the command?Is there some magic here? (otherwise I don't think the `## output file` name comment is necessary, to me (at least) it's self explanatory :)

modules/nf-core/sourmash/taxannotate/meta.yml
jfy133: Note updated docmentation guidelines [^1], but otherwise this documentation was already really good :+1: :muscle: 

[^1]: https://nf-co.re/docs/contributing/modules#documentation

tests/modules/nf-core/sourmash/taxannotate/main.nf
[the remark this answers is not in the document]
jfy133: Agreed! Clever!

jfy133 (changes requested):
The only blocker is the enable_conda deprecaion (which probably happened while this module was written), but other than the minor ocmments LGTM!

modules/nf-core/sourmash/taxannotate/main.nf
taylorreiter: ya there is magic here -- `sourmash tax annotate` chops `.csv` off of the gather file name and adds `.with-lineages.csv` to the end of it. (i don't have an opinion about the `## Output file name` comment

modules/nf-core/sourmash/taxannotate/main.nf
vmikk: Fixed

vmikk:
Pytest-workflow fails for conda, while for Docker and Singularity it works well.
```
Observed md5sum '001387d5fffaddded4764355bb7303a9' not equal to expected md5sum 'feb60b82ff625ffaf6584a1d3397c413'
```
Should we remove md5sum from the test for now?

modules/nf-core/sourmash/taxannotate/main.nf
vmikk: That's true, it looks cryptic.
Previously, as sourmash was not able to read gzipped-input, we unpacked input from `gather` and named it with `meta.id`, so it was clear. So now the process expects that `meta.id` is the same as `gather_results` (as inherited from `SOURMASH_GATHER`).

To fix it, maybe we can explicitly get the input file name and put it into a variable `prefix` inside the script itself? 
Please see the latest commit: https://github.com/nf-core/modules/pull/2422/commits/f85f107b9434279496247f4f66edf07fd2fcc0d7

jfy133: Hm, Ok. It still seems a bit over-complicated for me.

What about:

`gzip -cvf ${gather_results.getBaseName}.with-lineages.csv > ${prefix}.with-lineages.csv` [^1]

That way you can use exactly the output file name in the same way as sourmash (just using the native nextflow method), and that way you don't need any bash variables etc.

[^1]: getBaseName from Nextflow file attributes: https://www.nextflow.io/docs/latest/script.html#check-file-attributes

jfy133:
> Pytest-workflow fails for conda, while for Docker and Singularity it works well. Probably again something related not with the file content, but with file attributes, which leads to a different md5 of **gzipped** file.
> 
> ```
> Observed md5sum '001387d5fffaddded4764355bb7303a9' not equal to expected md5sum 'feb60b82ff625ffaf6584a1d3397c413'
> ```
> 
> Should we remove md5sum from the test for now?

Two options:

- try with the gzip option [`-n` or `--no-name`](https://nfcore.slack.com/archives/CJRH30T6V/p1634043714036200?thread_ts=1634043275.036000&cid=CJRH30T6V) it could be a timestamp issue -> preferred
- search for a string within the file (pytest can read into gzipped text files, maybe the column headers if it exists)

vmikk:
> Two options:
> 
>     * try with the gzip option [`-n` or `--no-name`](https://nfcore.slack.com/archives/CJRH30T6V/p1634043714036200?thread_ts=1634043275.036000&cid=CJRH30T6V) it could be a timestamp issue -> preferred
> 
>     * search for a string within the file (pytest can read into gzipped text files, maybe the column headers if it exists)

Thanks! The first option works like a charm!

modules/nf-core/sourmash/taxannotate/main.nf
vmikk: Sure, we can move `prefix` extraction to the Nextflow side.
However, I think we can not use just `getBaseName` as we need to remove both extensions (csv.gz), and it's not safe to use `getSimpleName` as it can remove part of the sample name if it contains period.
Also I tried to avoid unpacking the date into temporary file, as sourmash can read gz-comressed data.
So I've changed the code and used `prefix="${gather_results.getBaseName().replaceAll(/.csv/, '')}"`

jfy133: Ah fair point!

modules/nf-core/sourmash/taxannotate/main.nf
```diff
@@ -0,0 +1,43 @@
...
+
+    """
+    sourmash \\
+        tax annotate \
+        $args \\
+        --gather-csv ${gather_results} \\
+        --taxonomy ${taxonomy} \\
+        --output-dir "."
+
+    ## Output file name = "${prefix}".with-lineages.csv
```
jfy133, line 33:
```suggestion
    ## Output file name = "${gather_results}".with-lineages.csv
```

right? as this is what the tool actually generates.

However in this case I would rather not call it `prefix` as that should be reserved a user can customise via `ext.args` (which is not being used here). 

Actually do we need any of this inferring the file name at all? Are there any other ending in `with-lineages.csv`? Otherwise would simply not just after the annotate command, you run `gzip --no-name *.with-lineages.csv` ? No need to specify the full name?

jfy133 (commented):
Almogst there... one more question (sorry, I should probably just run the tool myself to see but I'm doing this review in between tasks 😅)

modules/nf-core/sourmash/taxannotate/main.nf
vmikk: Given `gather_results` = `tmp.csv.gz`, sourmash will produce `tmp.with-lineages.csv`, 
so `${gather_results}.with-lineages.csv` is different - it will be expanded as `tmp.csv.gz.with-lineages.csv`.
So if we want to name the output file explicitly, we must use `${prefix}.with-lineages.csv`.

However, I agree that there should be just a single `*.with-lineages.csv` file, so no need in the `prefix` at all.
I will fix the code in a minute.

jfy133: I see, OK! I'm going to give you the approval - sorry for complicating things!

jfy133 (approved):

modules/nf-core/sourmash/taxannotate/main.nf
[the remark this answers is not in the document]
jfy133: Technically we should support all input files as per the guidelines, however indeed from the descriptions `--from-file` does not sound particularly 'nextflow-y' so I'm going to avert my eyes from this for now ;)

PR #2427, opened 2022-10-28
new subworkflow bam_variant_calling_haplotypecaller

### LINKED ISSUE #2423: new subworkflow: bam_variantcalling_haplotypecaller
### Is there an existing subworkflow for this?

- [X] I have searched for the existing subworkflow

- [X] If I'm planning to work on this subworkflow, I added myself to the `Assignees` to facilitate tracking who is working on the subworkflow

nvnieuwk:
## PR checklist

Closes #2423  

- [ ] Add a resource `label`


nvnieuwk:
Things to do!
- [ ] Add support when no bed file is supplied
- [ ] Improve scatter/gather logic
- [ ] ... a lot of stuff

subworkflows/nf-core/bam_variant_calling_haplotypecaller/main.nf
```diff
@@ -0,0 +1,114 @@
...
+                                .map({ meta, bam, bai, full_bed, scattered_bed ->
+                                    new_meta = meta.clone()
+                                    new_meta.id = scattered_bed.baseName
+                                    new_meta.sample = meta.id
+                                    [ new_meta, bam, bai, scattered_bed ]
+                                })
+
+    } else {
+        ch_scattered_bams = ch_bam
+    }
```
FriederikeHanssen, line 47: So for each variantcalling subworkflow we have to resplit the bedfile?

FriederikeHanssen: and you need to know the number of splits you want to go for? Does the tool evenly split the bedfile into regions of similar size?

subworkflows/nf-core/bam_variant_calling_haplotypecaller/main.nf
```diff
@@ -0,0 +1,114 @@
...
+    if (strategy == "scatter" && scatter_count > 1){
+
+        ch_merge_input = GATK4_HAPLOTYPECALLER.out.vcf
+                            .combine(GATK4_HAPLOTYPECALLER.out.tbi, by:0)
+                            .map({ meta, vcf, tbi ->
+                                new_meta = [:]
+                                new_meta.id = meta.sample
+                                [ new_meta, vcf, tbi ]
+                            })
+                            .groupTuple()
```
FriederikeHanssen, line 86: Here we definitely need a group key magic or the merging will only happen after all haplotypecaller steps for all samples are run through making this a waiting bottleneck and unefficient

FriederikeHanssen (commented):
The logic for the gathering could be simplified with a branch operator, no?

subworkflows/nf-core/bam_variant_calling_haplotypecaller/main.nf
nvnieuwk: > So for each variantcalling subworkflow we have to resplit the bedfile?

Yeah or we could maybe adjust this workflow to also include multiple variant callers? (a bit similar to the uber alignment subworkflow)

> and you need to know the number of splits you want to go for? Does the tool evenly split the bedfile into regions of similar size?

bedtools split has two options for its splitting algorithm. The default uses a split to create regions that are similar in size with the downside that you could split on a variant region. The other algorithm keeps the regions specified in the BED file and tries to split them evenly (but there will be some variation in region length with this algorithm)

subworkflows/nf-core/bam_variant_calling_haplotypecaller/main.nf
nvnieuwk: That's interesting! I didn't think that was possible. I'll investigate 😎

nvnieuwk:
> The logic for the gathering could be simplified with a branch operator, no?

So you mean if there is a BED file you should perform the scatter, if not just skip it? Maxime asked for a functionality where you can skip the scatter overall that's why I added it this way but maybe a merge of the two ways would be best?

subworkflows/nf-core/bam_variant_calling_haplotypecaller/main.nf
FriederikeHanssen: See here for example: https://github.com/nf-core/sarek/blob/bcd7bf9cb98cddec27bb54fb47ee122c09388c02/subworkflows/nf-core/variantcalling/haplotypecaller/main.nf#L91 (and it is all over the sarek code base). In your case it might be slightly simpler to implement since you "know" the number of "scatters". In our case we have to count them in an earlier step in the pipeline, because they are changing depending on the bedfile.

FriederikeHanssen:
> > The logic for the gathering could be simplified with a branch operator, no?
> 
> So you mean if there is a BED file you should perform the scatter, if not just skip it? Maxime asked for a functionality where you can skip the scatter overall that's why I added it this way but maybe a merge of the two ways would be best?

Pretty sure @maxulysse and me are talking about the same thing just wording it differently. USing the "branch" operator would still allow to skip scattering altogether, but in my eyes would be a bit more nextflow-y. Basically, run merge on branched.multiple and then mix together branched.single with merged.out. If that makes sense.

subworkflows/nf-core/bam_variant_calling_haplotypecaller/main.nf
nvnieuwk: Allright, thanks! Should the subworkflow contain the code for dependent scatter counts? I don't know if it will make the subworkflow too bloated. It does seem like something that could improve the efficiency though.

nvnieuwk:
> > > The logic for the gathering could be simplified with a branch operator, no?
> > 
> > 
> > So you mean if there is a BED file you should perform the scatter, if not just skip it? Maxime asked for a functionality where you can skip the scatter overall that's why I added it this way but maybe a merge of the two ways would be best?
> 
> Pretty sure @maxulysse and me are talking about the same thing just wording it differently. USing the "branch" operator would still allow to skip scattering altogether, but in my eyes would be a bit more nextflow-y. Basically, run merge on branched.multiple and then mix together branched.single with merged.out. If that makes sense.

Yeah That sounds reasonable. So you mean if the user supplies a BED file to the subworkflow, it performs splitting and if the user doesn't supply a BED file with the input files, the subworkflow just runs the caller?

subworkflows/nf-core/bam_variant_calling_haplotypecaller/main.nf
FriederikeHanssen: since you have the fixed number, you can just supply that number as the second part in the groupkey argument. (or in this case might actually be able to use groupTuple and just set `size:scatter` . (best practice anyways to always set size for groupTuple. Otherwise the operator doesn't know when one group is "full" and when to advance.

FriederikeHanssen:
> Yeah That sounds reasonable. So you mean if the user supplies a BED file to the subworkflow, it performs splitting and if the user doesn't supply a BED file with the input files, the subworkflow just runs the caller?

yes essentially you can use something like:

``` 
haplotypecaller.out.vcf.branch {
       single: scatter <= 1
       multiple scatter > 1
     }.set(ch_output)
``` 

ch.output.single will be empty if scattering is done and `ch_output.multple` will be filed, which will the trigger the merging. And vice versa, if no scattering is done, onle `ch_output.single` will have any content, so implicelty no merging will be done as the `multiple` channel is empty. If you then mix the `single` and `merged` channel together you will automatically get the vcf output.

subworkflows/nf-core/bam_variant_calling_haplotypecaller/main.nf
nvnieuwk: Thanks for this! It really opened up a lot of possibilities 🥳

subworkflows/nf-core/bam_variant_calling_haplotypecaller/main.nf
```diff
@@ -0,0 +1,111 @@
...
+                        new_meta = meta.clone()
+                        new_meta.subwf_scatter_count = scatter_count ?: 1
+
+                        dragstr = dragstr_model ?: []
+
+                        multiple: new_meta.subwf_scatter_count > 1
+                            return [ new_meta, bam, bai, bed, dragstr ]
+                        single: new_meta.subwf_scatter_count <= 1
+                            return [ new_meta, bam, bai, bed, dragstr ]
+                    })
```
drpatelh, line 39:
```suggestion
    ch_bam
        .join(ch_dragstr_models ?: Channel.empty(), remainder: true)
        .join(ch_scatter ?: Channel.empty(), remainder: true)
        .branch( 
            { meta, bam, bai, bed, dragstr_model=[], scatter_count=[] ->
                new_meta = meta.clone()
                new_meta.subwf_scatter_count = scatter_count ?: 1

                dragstr = dragstr_model ?: []

                multiple: new_meta.subwf_scatter_count > 1
                    return [ new_meta, bam, bai, bed, dragstr ]
                single: new_meta.subwf_scatter_count <= 1
                    return [ new_meta, bam, bai, bed, dragstr ]
            }
        )
        .set { ch_input }
```

drpatelh: Much easier to read if you use `set`

drpatelh: And proper indents!

subworkflows/nf-core/bam_variant_calling_haplotypecaller/main.nf
```diff
@@ -0,0 +1,111 @@
...
+
+    BEDTOOLS_SPLIT (
+        ch_input.multiple.map({ meta, bam, bai, bed, dragstr_model -> [ meta, bed ]}),
+    )
+    ch_versions = ch_versions.mix(BEDTOOLS_SPLIT.out.versions.first())
+
+    ch_scattered_bams = ch_input.multiple.combine(BEDTOOLS_SPLIT.out.beds.transpose(), by:0)
```
drpatelh, line 46: Same here. Be nice to reformat this to use `set`

subworkflows/nf-core/bam_variant_calling_haplotypecaller/main.nf
```diff
@@ -0,0 +1,111 @@
...
+        ch_dbsnp,
+        ch_dbsnp_tbi
+    )
+    ch_versions = ch_versions.mix(GATK4_HAPLOTYPECALLER.out.versions.first())
+
+    //
+    // Optional: Merge scattered VCFs
+    //
+
+    ch_caller_output = GATK4_HAPLOTYPECALLER.out.vcf
```
drpatelh, line 73: Same here

subworkflows/nf-core/bam_variant_calling_haplotypecaller/main.nf
```diff
@@ -0,0 +1,111 @@
...
+        ch_fasta_fai
+    )
+    ch_versions = ch_versions.mix(BCFTOOLS_MERGE.out.versions.first())
+
+    TABIX_TABIX (
+        BCFTOOLS_MERGE.out.merged_variants
+    )
+    ch_versions = ch_versions.mix(TABIX_TABIX.out.versions.first())
+
+    ch_called_vcfs      = BCFTOOLS_MERGE.out.merged_variants
```
drpatelh, line 100: Same here

nvnieuwk: Thanks for the feedback! I'll adjust them now

matthdsm:
In the absence of an interval bed, could we use the new fangled GATK 4.3 `SplitCram`?  It's already implemented in the WARP Ultima genomics workflow.
https://broadinstitute.github.io/warp/docs/Pipelines/Ultima_Genomics_Whole_Genome_Germline_Pipeline/README/

nvnieuwk:
> In the absence of an interval bed, could we use the new fangled GATK 4.3 `SplitCram`? It's already implemented in the WARP Ultima genomics workflow. https://broadinstitute.github.io/warp/docs/Pipelines/Ultima_Genomics_Whole_Genome_Germline_Pipeline/README/

What would be the advantage in comparison with creating a BED file from the fasta index and using this to scatter the variant calling? (currently got that implemented in my local build)

matthdsm:
`SplitCram` splits the input cram by it's compression blocks, e.g. by batches of "similar" reads. This should improve read speed, since a cram file is read block by block. If we scatter per block, all those blocks can be reads and called in parallel. without requiring a bed file ( thats's my guess at least, I haven't checked out the WARP implementation thoroughly).

nvnieuwk:
> `SplitCram` splits the input cram by it's compression blocks, e.g. by batches of "similar" reads. This should improve read speed, since a cram file is read block by block. If we scatter per block, all those blocks can be reads and called in parallel. without requiring a bed file ( that's my guess at least, I haven't checked out the WARP implementation thoroughly).

Aaah that makes sense! That does sound better than the current way. @maxulysse and @FriederikeHanssen what are your opinions?

nvnieuwk:
> `SplitCram` splits the input cram by it's compression blocks, e.g. by batches of "similar" reads. This should improve read speed, since a cram file is read block by block. If we scatter per block, all those blocks can be reads and called in parallel. without requiring a bed file ( that's my guess at least, I haven't checked out the WARP implementation thoroughly).

`SplitCRAM` is currently still classified as an experiment tool, so we should probably wait a bit longer before adding it?

nvnieuwk:
> > `SplitCram` splits the input cram by it's compression blocks, e.g. by batches of "similar" reads. This should improve read speed, since a cram file is read block by block. If we scatter per block, all those blocks can be reads and called in parallel. without requiring a bed file ( that's my guess at least, I haven't checked out the WARP implementation thoroughly).
> 
> `SplitCRAM` is currently still classified as an experimental tool, so we should probably wait a bit longer before adding it?

Added with #2473

subworkflows/nf-core/bam_variant_calling_haplotypecaller/main.nf
```diff
@@ -0,0 +1,155 @@
...
+                [ meta, bam, bai, bed, dragstr ]
+            }
+        )
+        .mix(ch_input.multiple)
+        .set { ch_to_split }
+
+    BEDTOOLS_SPLIT (
+        ch_to_split.map({ meta, bam, bai, bed, dragstr_model -> [ meta, bed ]}),
+    )
+    ch_versions = ch_versions.mix(BEDTOOLS_SPLIT.out.versions.first())
```
FriederikeHanssen, line 75: should this go in a separate subworkflow? ths would be exactly yhe same for other variantcalling sw

matthdsm: 👍🏻 love the idea!

maxulysse: :exploding_head:

matthdsm: OTOH, I _think_ I remember a tool which is able to chunk a bed file into "equal" parts of sorts, which could replace this whole thing... I'm going to have to look it up

nvnieuwk: `bedtools split` already does that :). But I do agree that this should be in a different subworkflow!

maxulysse: so https://github.com/nf-core/modules/issues/2540 ?

nvnieuwk: Yeah I'm currently working on it :)

nvnieuwk: https://github.com/nf-core/modules/pull/2544

nvnieuwk: The big question now is: should we add scatter/gather to this subworkflow or let the pipeline developer do this on a pipeline level. 
An argument for not adding it here could be that the pipeline developer could choose which scatter/gather method he/she uses since there will probably be more scatter and gather subworkflows in the future. This way gives more flexibility, but makes this subworkflow quite redundant since it will only run haplotypecaller.

nvnieuwk:
Oh crap forgot that this one was still open

subworkflows/nf-core/bam_variant_calling_haplotypecaller/main.nf
nvnieuwk: I agree! What do you guys think, @maxulysse and @matthdsm? IMO this subworkflow is pretty pointless now since it would only run Haplotypecaller with no channel magic involved

PR #2437, opened 2022-10-31
SEQTK/SAMPLE supports more than 2 FASTQ files

adamrtalbot:
Inspired by recent update to FASTQC module, SEQTK/SAMPLE supports more than 2 FASTQ files, by looping through all of them. Previously it only had single or paired options.

- Supports wider variety of FASTQ files (single, ended, paired, paired + auxiliary).
- Removes dependency on `single_end` value in `meta`. 
- Allows possibility of dynamically setting N reads value based on input

As part of this, updated KRAKEN subworkflow which used the SEQTK/SAMPLE module. It now merges sample_size with FASTQ input channel. I include an update to the test where you can sample at multiple depths prior to running KRAKEN as a proof of concept, although MD5 for the two output files is the same.

## PR checklist

- [x] Add a resource `label`


Midnighter:
Auxiliary sequencing read files are typically much smaller than the primary (single-end or paired-end) files. I suspect that in most cases `sample_size` will be much larger than the total number of reads in those auxiliary files, making it meaningless to include them here. Or do you have another example where this isn't the case?

apeltzer:
`Observed md5sum '881e5bf7589efed7b6883c34[69](https://github.com/nf-core/modules/actions/runs/3385408709/jobs/5623520075#step:11:70)f07150' not equal to expected md5sum '75457ddceea3a5688d05438cdbffba24' for file '/home/runner/pytest_workflow_la2hchmq/seqtk_sample_test_seqtk_sample_paired_end/output/seqtk/test.sampled_test_2.fastq.gz.fastq.gz'`

apeltzer (approved):
Looks good to me.

modules/nf-core/seqtk/sample/main.nf
```diff
@@ -8,8 +8,7 @@ process SEQTK_SAMPLE {
         'quay.io/biocontainers/seqtk:1.3--h5bf99c6_3' }"
 
     input:
-    tuple val(meta), path(reads)
-    val sample_size
+    tuple val(meta), path(reads), val(sample_size)
```
FriederikeHanssen, line 11: if the module is overhauled anyway, I would argue for removing this val and putting it in the task.ext.args... 

(I am scared already and will hide somewhere :D )

adamrtalbot: I thought about this, but it's a big change for existing people who use the module. I've already introduced one breaking change but tried to not introdue a breaking change for the subworkflow.

apeltzer: The subworfklow is very new and only used (soon) in demultiplexing and my internal workflow - happy if this makes it more generic to use 👍🏻

adamrtalbot: Actually, the seqtk sample syntax is:
`seqtk sample [-2] [-s seed=11] <in.fa> <frac>|<number>`

So we'd have to use `args2` and make it compulsory? Or perhaps `task.ext.n`? This seems complicated and this has already stalled, I'll push that for a bit more investigation.

PR #2461, opened 2022-11-04
Prototype nf test

Emiller88:


Emiller88:
- Conda isn't working
- The whole `subworkflows/` thing has to go or we need to move all subworkflow tests under their own dir. I kinda like the second option.

matthdsm:
@Emiller88, is this still relevant or can we close?

PR #2475, opened 2022-11-07
Ensure string trimming is suffix at end of filepath

asgaines:
# Description

When working with a genome named something to the effect of `lambda.fa` (with accompanying `lambda.fa.amb` file), the following 'find`/`sed` commands in the `bwa` suite triggers a bug which breaks the workflow:

```bash
INDEX=`find -L ./ -name "*.amb" | sed 's/.amb//'`
```

The intention of the `sed` section is to strip away the file extension of the `.amb` file, but the regular expression is buggy when the rest of the filepath matches `.amb` (like `lamb` in `lambda.fa.amb`).

Examples:
```bash
echo hg38.fa.amb | sed 's/.amb//'  # yields: hg38.fa (works as intended)
echo lambda.fa.amb | sed 's/.amb//'  # yields: da.fa.amb (not what's intended)
```

The bug is due to the regex `.` character matching any character (in the case of `lambda`, the `l`). The period must be escaped to not be considered a special character. Furthermore, the trailing `$` is also necessary, otherwise this case is still possible:

```bash
echo ecoli.ambulatory.fa.amb | sed 's/.amb//'  # yields: ecoliulatory.fa.amb (not what's intended)
```

# Incomplete PR Checklist

I wanted to create tests to highlight this issue and prove the fix, but I couldn't get the tests running locally. I first needed to grant access to `/tmp` to `docker`, but then encountered a confusing permission denied issue, at which point I gave up, figuring it would be good to at least submit this PR. I did follow the instructions on this page: https://nf-co.re/developers/modules#writing-a-new-module-reference

## PR checklist

- [X] Add a resource `label`


muffato (approved):
I'm happy with your description and the change, it totally makes sense.

This way of setting `INDEX` is used in quite a few modules. Would you be able to search for the other ones and fix their `sed` the same way ?

muffato:
(I'm not worried about the CI failures, that's just Conda being Conda)

asgaines:
@muffato I got a chance to scan through the rest of the repo. I updated all of the regexes missing anchoring that I could find

muffato:
The change per se looks, and I'm looking at the CI errors

muffato:
All the biscuit errors were due to conda implicitly installing samtools 1.15.1 when asked for 1.15. My solution is to pin the build to guarantee the same version as in the mulled container.
All the other errors pertain to BAM files on Conda, which is a known issue.

PR #2399, opened 2022-10-26
Shinyngs/staticdifferential

pinin4fjords:
Just a simple module wrapper for https://github.com/pinin4fjords/shinyngs/blob/develop/exec/differential_plots.R, useful for making differential plots in differential analysis workflows. Right now it's just making volcano plots, but will be extended in future.

Notes;

- I've tried to adhere to the feedback given for related PRs (see https://github.com/nf-core/modules/pull/2315, https://github.com/nf-core/modules/pull/2364)
- The matrix data provided in the second channel is not currently used in the command. But I believe it will be necessary for future differential plots that leverage expression data (rather than just differential statistics as volcano plots do), so I've added the input channel to facilitate that.

## PR checklist

Closes https://github.com/nf-core/modules/issues/2398

- [x] This comment contains a description of changes (with reason).
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [x] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [x] If necessary, include test data in your PR.
- [x] Remove all TODO statements.
- [x] Emit the `versions.yml` file.
- [x] Follow the naming conventions.
- [x] Follow the parameters requirements.
- [x] Follow the input/output options guidelines.
- [x] Add a resource `label`
- [x] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - [x] `PROFILE=docker pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [x] `PROFILE=singularity pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [x] `PROFILE=conda pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`

PR #2476, opened 2022-11-07
Allow flexibility in meta map usage with Deseq2

pinin4fjords:
## PR checklist

This PR establishes some consistency with [feedback](https://github.com/nf-core/modules/pull/2399#discussion_r1015044350) from @jfy133 in another PR. Essentially we should not be 'baking in' assumptions about the content meta maps, but rather allowing the workflow-level usage to pass that. 

So, where the R template was previously reading the meta directly, it now receives the meta options like by setting the following in the workflow configs:

```ext.args = { "--contrast_variable $meta.variable --reference_level $meta.reference --treatment_level $meta.target --blocking_variables $meta.blocking "}```

I've also tweaked the bare-bones option parsing there to allow for blank values passed that way.

- [x] Add a resource `label`

### CODE REVIEW

File: modules/nf-core/deseq2/differential/templates/deseq_de.R [hunk 1]
```diff
@@ -14,9 +14,13 @@
 
 parse_args <- function(x){
     args_list <- unlist(strsplit(x, ' ?--')[[1]])[-1]
-    args_vals <- unlist(lapply(args_list, function(y) strsplit(y, ' +')))
+    args_vals <- unlist(lapply(args_list, function(y) strsplit(y, ' +')), recursive = FALSE)
 
-    as.list(structure(args_vals[c(FALSE, TRUE)], names = args_vals[c(TRUE, FALSE)]))
+    # Ensure the option vectors are length 2 (key/ value) to cactch empty ones
```
WackerO, line 19:
```suggestion
    # Ensure the option vectors are length 2 (key/ value) to catch empty ones
```

modules/nf-core/deseq2/differential/templates/deseq_de.R
```diff
@@ -187,13 +205,13 @@ if (!opt\$contrast_variable %in% colnames(sample.sheet)) {
...
-} else if (!is.null(opt\$blocking)) {
-    blocking.vars = unlist(strsplit(opt\$blocking, split = ';'))
+} else if (!is.null(opt\$blocking_variables)) {
+    blocking.vars = unlist(strsplit(opt\$blocking_variables, split = ';'))
     if (!all(blocking.vars %in% colnames(sample.sheet))) {
         stop(
             paste0(
                 'One or more of the blocking variables specified (',
-                opt\$blocking,
+                opt\$blocking_variables,
```
WackerO, line 214: I'd suggest maybe changing this error message to mention which variables are missing? For blocking, probably people will usually not list that many variables, I understand that, but generally, I try to make such error messages as fool-proof as possible by directly pointing to the offending values

WackerO (approved):
Other than those two comments, LGTM!

modules/nf-core/deseq2/differential/templates/deseq_de.R
pinin4fjords: Haha, yes, I hadn't thought the max 2 or 3 variables people could use would present a problem- but okay, I've it more explicit :-)

pinin4fjords:
Thanks for the review @WackerO

PR #2498, opened 2022-11-10
Bismark summary - BAM filenames, not files

ewels:
This bismark/summary module pulled in all BAM files to run. This staged them in the work directory and `bismark2summary` used these to guess the report filenames. That's all 😅  Quite an expensive strategy when staging files on the cloud.

Changed to instead accept a list of BAM _filenames_ instead, without staging the files.

## PR checklist

Closes #XXX 

### DIFF
--- a/modules/nf-core/bismark/summary/main.nf
+++ b/modules/nf-core/bismark/summary/main.nf
@@ -7,7 +7,7 @@ process BISMARK_SUMMARY {
         'quay.io/biocontainers/bismark:0.24.0--hdfd78af_0' }"
 
     input:
-    path(bam)
+    val(bam)
     path(align_report)
     path(dedup_report)
     path(splitting_report)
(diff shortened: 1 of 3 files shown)


modules/nf-core/bismark/summary/main.nf
```diff
@@ -7,7 +7,7 @@ process BISMARK_SUMMARY {
         'quay.io/biocontainers/bismark:0.24.0--hdfd78af_0' }"
 
     input:
-    path(bam)
+    val(bam)
```
FriederikeHanssen, line 10: the one time I like val channels for modules 😆

Emiller88:
This needs `nf-core modules patch`!

modules/nf-core/bismark/summary/main.nf
FelixKrueger: that one little change can probably save a **lot** of time (and money). To be honest though, I don't quite understand how a value channel would then be picked up the the `glob *.bam` that takes place in the work folder?

ewels: There is no glob? I updated the command so that it is given each explicit BAM filename. eg:

```
bismark_summary bam_1.bam bam2.bam bam3.bam
```

Issue #2502, opened 2022-11-12
ATAQV_ATAQV module requires mito_name as parameter (currently supplied via ext.args


robsyme:
### Have you checked the docs?

- [X] [nf-core website: troubleshooting](https://nf-co.re/usage/troubleshooting)
- [X] [nf-core modules documentation](https://nf-co.re/docs/contributing/modules)

### Description of the bug

# Description

The ATAQV_ATAQV process can add the `--mitochondrial-reference-name` argument when the mitochondrial name is known.

In nf-core/atacseq, where this module is used, the dev branch supplies this via the ext.args configuration option, where it relies on the `params.mito_name` configuration value to be available when this configuration is resolved:
```
ext.args   = [
    '--ignore-read-groups',
    params.mito_name ? "--mitochondrial-reference-name ${params.mito_name}" : ''
].join(' ').trim()
``` 

The `params.mito_name` value is set inside [main.nf#L29](https://github.com/nf-core/atacseq/blob/6b95c43e11f6e32c9de29baa4370ede6dac67a09/main.nf#L29), but this happens *after* the ext.args is resolved.

# Suggestion
My recommendation (well, @drpatelh's recommendation really) is that we move the mito_name to a process input and we remove the corresponding value from the configuration.

For all future work, we should be wary of using any parameter values in configuration if those parameters are set in `main.nf` using the `WorkflowMain.getGenomeAttribute()` method.

### Command used and terminal output

```console
$ nextflow run https://github.com/nf-core/atacseq -r dev 6b95c43 -profile test_full
N E X T F L O W  ~  version 22.10.1
Pulling nf-core/atacseq ...
 downloaded from https://github.com/nf-core/atacseq.git
Unknown config attribute `params.mito_name` -- check config file: /.nextflow/assets/nf-core/atacseq/nextflow.config
```

### System information

- Nextflow version (22.10.1)
- Hardware (Desktop, Cloud)
- Executor (local, awsbatch)
- Container engine and version: (AWS batch)
- OS and version: (CentOS Linux)

robsyme:
Resolved with merge

PR #2503, opened 2022-11-12
Add mito_name to ATAQV_ATAQV inputs

robsyme:
# Description

This PR is required to fix an issue in nf-core/atacseq, detailed in https://github.com/nf-core/modules/issues/2502.

Essentially, we cannot rely on ext.args to supply the `--mitochondrial-reference-name` argument, as the parameter is often set *after* the ext.args are resolved.

This commit introduces the `mito_name` as a process input.

### DIFF
--- a/modules/nf-core/ataqv/ataqv/main.nf
+++ b/modules/nf-core/ataqv/ataqv/main.nf
@@ -34,6 +34,7 @@ process ATAQV_ATAQV {
     """
     ataqv \\
         $args \\
+        $mito \\
         $peak \\
         $tss \\
         $excl_regs \\
--- a/modules/nf-core/ataqv/ataqv/main.nf
+++ b/modules/nf-core/ataqv/ataqv/main.nf
@@ -10,6 +10,7 @@ process ATAQV_ATAQV {
     input:
     tuple val(meta), path(bam), path(bai), path(peak_file)
     val organism
+    val mito_name
     path tss_file
     path excl_regs_file
     path autosom_ref_file
@@ -24,6 +25,7 @@ process ATAQV_ATAQV {
 
     script:
     def args = task.ext.args ?: ''
+    def mito = mito_name ? "--mitochondrial-reference-name ${mito_name}" : ''
     def prefix = task.ext.prefix ?: "${meta.id}"
     def peak        = peak_file        ? "--peak-file $peak_file"                       : ''
     def tss         = tss_file         ? "--tss-file $tss_file"                         : ''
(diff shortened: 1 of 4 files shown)


modules/nf-core/ataqv/ataqv/main.nf
```diff
@@ -24,6 +25,7 @@ process ATAQV_ATAQV {
 
     script:
     def args = task.ext.args ?: ''
+    def mito = mito_name ? "--mitochondrial-reference-name ${mito_name}" : ''
```
drpatelh, line 28: Need to add this to the command too 👀

tests/modules/nf-core/ataqv/ataqv/main.nf
```diff
@@ -51,7 +51,7 @@ workflow test_ataqv_ataqv_tss {
     ] 
     tss_file = file(params.test_data['sarscov2']['genome']['test_bed'], checkIfExists: true)
    
-    ATAQV_ATAQV ( input, 'human', tss_file, [], [] )
+    ATAQV_ATAQV ( input, 'human', [], tss_file, [], [] )
```
drpatelh, line 54: Seeing as it's a string can we pass an empty string instead?

modules/nf-core/ataqv/ataqv/main.nf
robsyme: 🤦

tests/modules/nf-core/ataqv/ataqv/main.nf
robsyme: Both will work. The string is probably a little cleaner
.

robsyme: The parameter might resolve to `null` in cases where it is not provided, as that's what the `getGenomeAttribute` when the parameter is not found. Given that, I think that `null` is a more representative "default" to what we'll see in the wild.

drpatelh: At this point, the module can be used in any workflow independent of that function so I think it makes more sense to default to the type specified in the `meta.yml`. Especially since these tests are unrelated to how the module is used in the workflow.

robsyme: Fair. Fixed

drpatelh: But we do default empty strings to `null` in the pipeline `nextflow.config` so you are right too 😅 In these scenarios I just find it easier to look at the tests and know what type the input channel is likely to be if we default to the native type, especially when you have loads of inputs.

drpatelh (approved):
🚀

PR #2506, opened 2022-11-14
Fix bbmap version string spurious text in singularity tests

SPPearce:
Fix the failing singularity tests in bbmap. See [https://github.com/nf-core/modules/pull/2206#issuecomment-1301828343](https://github.com/nf-core/modules/pull/2206#issuecomment-1301828343).

For some reason the singularity container on the github automated tests is producing some additional text after the `bbversion.sh` command is run, which confuses `pytest`. This just removes that string with a little `grep`.


Emiller88:
Don't merge until after #2510

SPPearce:
Conda is failing in three tests for whatever reason, with mismatched md5sums on files that aren't the `versions.yml` file.
I don't have time to look at this further today.

SPPearce:
Conda failing tests are due to what I believe is a common issue with samtools, where something in the header is different for files generated using conda.

mahesh-panchal (approved):

mahesh-panchal:
I think the `bam` issues are probably related to that, and the `gz` issues are related to that one where we found a four byte difference only because of the compression library, but the contents were exactly the same.

PR #2510, opened 2022-11-14
ci: Add apt-get update

Emiller88:
Also switch to apt-get

Fixes

```
Run sudo apt install bat > /dev/null

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/universe/r/rust-bat/bat_0.12.1-1ubuntu0.1_amd64.deb  404  Not Found [IP: 52.252.75.106 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
Error: Process completed with exit code 100.
```

### DIFF
--- a/.github/workflows/pytest-workflow.yml
+++ b/.github/workflows/pytest-workflow.yml
@@ -96,7 +96,8 @@ jobs:
       - name: Output log on failure
         if: failure()
         run: |
-          sudo apt install bat > /dev/null
+          sudo apt-get update > /dev/null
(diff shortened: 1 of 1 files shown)


adamrtalbot (approved):
As far as I can tell, this shouldn't work. But it does. So approved?

Emiller88:
> As far as I can tell, this shouldn't work. But it does. So approved?
Why shouldn't it?

Issue #2516, opened 2022-11-16
Reports channel form `KRAKEN_SCREEN` is empty


fmalmeida:
### Have you checked the docs?

- [X] [nf-core website: troubleshooting](https://nf-co.re/usage/troubleshooting)
- [X] [nf-core modules documentation](https://nf-co.re/docs/contributing/modules)

### Description of the bug

When running the `KRAKEN_SCREEN` sub-workflow the output channels from it, `KRAKEN_SCREEN.out.reports` are all empty.

I checked and the fix is having:

```groovy
ch_versions = ch_versions.mix(KRAKEN2.out.versions.first())
ch_reports  = ch_reports.mix(KRAKEN2.out.report)
```

Instead of:

https://github.com/nf-core/modules/blob/master/subworkflows/nf-core/fastq_contam_seqtk_kraken/main.nf#L40-L41

### System information

Nextflow version 22.10.1

PR #2517, opened 2022-11-16
Fix KRAKEN_SCREEN output channel definition

fmalmeida:
### Description of the bug

When running the `KRAKEN_SCREEN` sub-workflow the output channels from it, `KRAKEN_SCREEN.out.reports` are all empty. Related to issue #2516 

I checked and the fix is having:

```groovy
ch_versions = ch_versions.mix(KRAKEN2.out.versions.first())
ch_reports  = ch_reports.mix(KRAKEN2.out.report)
```

Instead of:

https://github.com/nf-core/modules/blob/master/subworkflows/nf-core/fastq_contam_seqtk_kraken/main.nf#L40-L41


grst (approved):
Lgtm

PR #2526, opened 2022-11-18
New subworkflow: vcf_joint_vc_gatk

matthdsm:
Joint genotyping workflow ported from Sarek.

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`


subworkflows/nf-core/vcf_joint_calling_germline_gatk/meta.yml
```diff
@@ -0,0 +1,48 @@
...
+  - csi:
+      type: file
+      description: CSI samtools index
+      pattern: "*.csi"
+  - versions:
+      type: file
+      description: File containing software versions
+      pattern: "versions.yml"
+authors:
+  - "None"
```
maxulysse, line 48: I think you can give credits to both @nickhsmith and yourself for that

matthdsm: Will do! This is still very much a WIP though. I'm waiting for the bed scatter subwf

matthdsm:
```suggestion
  - "nickhsmith"
  - "@matthdsm"
```

subworkflows/nf-core/vcf_joint_calling_germline_gatk/meta.yml
```diff
@@ -0,0 +1,49 @@
...
+  - csi:
+      type: file
+      description: CSI samtools index
+      pattern: "*.csi"
+  - versions:
+      type: file
+      description: File containing software versions
+      pattern: "versions.yml"
+authors:
+  - "nickhsmith"
```
FriederikeHanssen, line 48:
```suggestion
  - "@nickhsmith"
```

matthdsm: Nice catch!

subworkflows/nf-core/vcf_joint_calling_germline_gatk/main.nf
```diff
@@ -0,0 +1,73 @@
...
+    ch_versions    = Channel.empty()
+
+    // Merge bed files to create a single interval list
+    // BEDTOOLS_MULTIINTER( [meta, bed], chrom_sizes )
+    BEDTOOLS_MULTIINTER( ch_beds, [] )
+    ch_versions = ch_versions.mix(BEDTOOLS_MULTIINTER.out.versions).first()
+
+    // Split merged bed into individual intervals for genotyping
+    // BED_SCATTER_BEDTOOLS( [meta, bed, scatter_count] )
+    BED_SCATTER_BEDTOOLS(BEDTOOLS_MULTIINTER.out.bed, scatter_count)
```
nvnieuwk, line 33: I think you can leave the scatter/gather out of the subworkflow too as what's being discussed here: https://github.com/nf-core/modules/pull/2427#discussion_r1040655133

nvnieuwk: This way it's easier for the pipeline developer to just mix and match the subworkflows they prefer instead of making several subwfs with each method

matthdsm: I think the whole point of this subwf is to make it easier to have parallel joint GT in a pipeline, same for the haplotyper one.

By including the scatter/gather logic, a dev can import the subwf and be confident it runs in a parallel way without having to program any logic for it

matthdsm: if we keep reducing functionality to the bare minimum we just end up with modules again IMO 🤷🏻

nvnieuwk: The scatter/gather subworkflows I made don't actually need any more channel logic to fully use. You can just run them right after each other. But it's something that's still in discussion, we should probably discuss it with some more people. (I think designing the subworkflows so they work correctly with each other but keep them separate is the most flexible solution without giving up the ease of subworkflows)

nvnieuwk: I'll make a gather subwf with mergevcfs in the near future too :)

nvnieuwk: > if we keep reducing functionality to the bare minimum we just end up with modules again IMO 🤷🏻

Yep that's why we should probably fully discuss this first

matthdsm:
@nvnieuwk , you can pick this up when you're finished with your local version

PR #2530, opened 2022-11-18
Add matrixfilter module

pinin4fjords:
This PR adds a module for matrix filtering. It's an embedded R script in a template rather than calling an external library, happy to be directed if there's a better way of achieving this. The motivating use case is https://github.com/nf-core/differentialabundance, but the concept isn't specific to that workflow. 

Since this is not an external tool with externally documented optional params, I've documented the optional variables supplied via task.ext.args in the meta.yaml, in addition to the mandatory channel-supplied params.

The module is tested in 3 ways illustrating different use cases:

- Filter to require 1 count in at least 1 sample (default)
- Filter to require 1 count in at least 75% of samples
- Filter to require at least 1 count in a number of samples matching 100% of the size of the smallest group.

## PR checklist

- [x] Add a resource `label`


modules/nf-core/matrixfilter/templates/matrixfilter.R
```diff
@@ -0,0 +1,179 @@
...
+)
+
+opt <- parse_args('$task.ext.args', opt)
+
+abundance_matrix <- read_delim_flexible(opt\$abundance_matrix_file, row.names = 1)
+
+# Read the sample sheet and check against matrix
+
+samplesheet <- read_delim_flexible(opt\$sample_file, row.names = 1)
+if (any(! rownames(samplesheet) %in% colnames(abundance_matrix))){
```
WackerO, line 107: Could you again change this check to tell the user with values are missing?

WackerO (approved):
Apart from my comment lgtm!

modules/nf-core/matrixfilter/meta.yml
```diff
@@ -0,0 +1,71 @@
...
+      default: 1
+  - minimum_proportion:
+      type: numeric
+      description: |
+        A minimum proportion of observations that must pass the threshold.
+        Supplied via task.ext.args as --minimum_proportion. Overrides
+        minimum_samples
+      default: 0
+  - grouping_variable:
+      type: string
```
jfy133, line 47: I'm not sure about these to be honest, as this meta file is meant to just describe channels (but I don't have a better alternative place for it yet).

modules/nf-core/matrixfilter/main.nf
```diff
@@ -0,0 +1,22 @@
...
+    input:
+    tuple val(meta), path(samplesheet), path(abundance)
+
+    output:
+    tuple val(meta), path("*.filtered.tsv")             , emit: filtered
+    tuple val(meta), path("R_sessionInfo.log")          , emit: session_info
+    path "versions.yml"                                 , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
```
jfy133, line 18: How/where are the args passed to the script itself (I'm not familiar with templates admittedly)?

modules/nf-core/matrixfilter/main.nf
```diff
@@ -0,0 +1,22 @@
...
+    conda (params.enable_conda ? "r-base" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/r-base:4.2.1' :
+        'quay.io/biocontainers/r-base:4.2.1' }"
+
+    input:
+    tuple val(meta), path(samplesheet), path(abundance)
+
+    output:
+    tuple val(meta), path("*.filtered.tsv")             , emit: filtered
```
jfy133, line 13: This should just be `${prefix}` and specified via `ext.prefix` (should be used as with ext.args), and a check added for no name clash (to force a dev to supply prefix)

modules/nf-core/matrixfilter/meta.yml
```diff
@@ -0,0 +1,71 @@
...
+input:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing information on experiment, at a minimum an id.
+        e.g. [ id:'test' ]
+  - samplesheet:
+      type: file
+      description: |
+        CSV or TSV format sample sheet with sample metadata
```
jfy133, line 20: I would make this a separate optional input, I could imagine people just want to filter based on the entire dataset.

modules/nf-core/matrixfilter/main.nf
jfy133: Ah found it... maybe comment that `$task.ext.args` are specified on line XYZ in the scipt?

modules/nf-core/matrixfilter/main.nf
```diff
@@ -0,0 +1,22 @@
+process MATRIXFILTER {
+    tag "$meta"
+    label 'process_medium'
```
jfy133, line 3: I don't see anyy multithreading n the script
```suggestion
    label 'process_single'
```

modules/nf-core/matrixfilter/meta.yml
```diff
@@ -0,0 +1,71 @@
...
+  - abundance:
+      type: file
+      description: |
+        Raw TSV or CSV format expression abundance matrix with features (e.g.
```
jfy133, line 24:
```suggestion
        Raw TSV or CSV format abundance matrix with features (e.g.
```

modules/nf-core/matrixfilter/meta.yml
```diff
@@ -0,0 +1,71 @@
...
+        genes) by row and observations (e.g. samples) by column. All rowsnames
```
jfy133, line 25:
```suggestion
        genes) by row and observations (e.g. samples) by column. All rownames
```

modules/nf-core/matrixfilter/main.nf
jfy133: Although, why not use the same template boilerplate to psas arguments?

```
    def prefix = task.ext.prefix ?: meta.id
    ```

**Review by jfy133 — CHANGES_REQUESTED** [review]

File: modules/nf-core/matrixfilter/meta.yml (hunk 2 above)
Comment (pinin4fjords) [reply modules/nf-core/matrixfilter/meta.yml:47]: I agree. The need to document optional params because they're not documented as part of an underlying tool is one of the reasons I raised this as a tricky case.

File: modules/nf-core/matrixfilter/main.nf (hunk 3 above)
Comment (pinin4fjords) [reply modules/nf-core/matrixfilter/main.nf:18]: I _think_ I tried this and it didn't work in the template (newly defined variables weren't accessible in the template). But I'll double check.

File: modules/nf-core/matrixfilter/meta.yml (hunk 5 above)
Comment (pinin4fjords) [reply modules/nf-core/matrixfilter/meta.yml:20]: I actually started off doing it that way, but decided it was an unnecessary complication to the logic given that something like a sample sheet will always be around. But I can re-instate it.

File: modules/nf-core/matrixfilter/meta.yml [hunk 9]
```diff
@@ -0,0 +1,71 @@
...
+      type: numeric
+      description: |
+        A minimum proportion of observations that must pass the threshold.
+        Supplied via task.ext.args as --minimum_proportion. Overrides
+        minimum_samples
+      default: 0
+  - grouping_variable:
+      type: string
+      description: |
+        Supply a variable from the sample sheet that can be used to define
```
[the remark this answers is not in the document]
pinin4fjords: Ahh yes, it is, just didn't mark it - will do

modules/nf-core/matrixfilter/templates/matrixfilter.R
```diff
@@ -0,0 +1,186 @@
...
+opt <- list(
+    abundance_matrix_file = '$abundance',
+    sample_file = '$samplesheet',
+    minimum_abundance = 1,
+    minimum_samples = 1,
+    minimum_proportion = 0,
+    grouping_variable = NULL
+)
+
+opt <- parse_args('$task.ext.args', opt)
```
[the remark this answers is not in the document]
pinin4fjords: Again it's down to my having trouble accessing new variable definitions from the template, I'll try again to see if it was an earlier issue of mine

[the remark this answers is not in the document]
pinin4fjords: Confirmed. If I do this:

```
    script:                                                                     
    template 'matrixfilter.R'  
```

... I can't then refer to 'args' in matrixfilter.R. The relevant line resolves like:

```
opt <- parse_args('[]', opt)
```

modules/nf-core/matrixfilter/main.nf
pinin4fjords: Explanatory note added

modules/nf-core/matrixfilter/main.nf
pinin4fjords: Not 100% sure what you mean there @jfy133 . I've changed the output to be renamed according to the prefix derived from the meta, feel free to expand a bit and I'll try to match what you're after.

pinin4fjords:
@jfy133 I think I've hit everything we discussed (with the exception of the prefix thing I'm a little unclear on). Ready for re-review.

modules/nf-core/matrixfilter/main.nf
jfy133: Ah, sorry that is indeed unclear. I posted it on the wrong line.

In the R templtae script, you shoud use the module=-local `${prefix}` variable, and therefore here in the module in the `script:` block (like you normally would have for any other module to have the `prefx = task.ext.prefix ?: 'meta.id'` thing.

Does that make sense? Basically I'm trying to push this module back to as close as the module template as possible :)

modules/nf-core/matrixfilter/templates/matrixfilter.R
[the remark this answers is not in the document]
jfy133: Same concept as what I meant above for the prefix :)

jfy133 (commented):
Almost there! See my comment above on the thread :)

modules/nf-core/matrixfilter/main.nf
pinin4fjords: Thanks! Is the [template edit](https://github.com/pinin4fjords/nf-core-modules/blob/f2e2ae71685c7e8a6d153feff8ab71fd597fb948/modules/nf-core/custom/matrixfilter/templates/matrixfilter.R#L165) not fulfilling this? 

I'm intentionally not referring to `meta` explicitly in the template due to previous feedback about meta assumptions ;-). But it's done in the tests from [the nextflow.config](https://github.com/pinin4fjords/nf-core-modules/blob/f2e2ae71685c7e8a6d153feff8ab71fd597fb948/tests/modules/nf-core/custom/matrixfilter/nextflow.config#L17).

pinin4fjords: Again, not that setting `prefix` in the script block won't work here- the template doesn't seem to receive new variable definitions from `script` (not sure if that's a bug or a feature).

jfy133: Oh, schade :( ok!

jfy133 (approved):

pinin4fjords:
Thanks @jfy133 !

PR #2544, opened 2022-11-23
new subworkflow: bed_scatter_bedtools

nvnieuwk:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`


maxulysse (approved):
Love it

nvnieuwk:
@FriederikeHanssen and @matthdsm can you two also have a look if you got some free time? Definitely want to get this one right :)

subworkflows/nf-core/bed_scatter_bedtools/main.nf
```diff
@@ -0,0 +1,81 @@
...
+
+    take:
+    ch_bed          // channel: [ meta, bed, scatter_count ]
+    ch_fasta_fai    // channel: [ fasta_fai ]
+
+    main:
+
+    ch_versions = Channel.empty()
+
+    ch_bed
```
matthdsm, line 16: Instead of branching, shouldn't we make both inputs optional?
That way we split a bed file either way, but we generate one first if a `fai` is given.

FriederikeHanssen: agree here.

ch_input = bed.mix(GAWK(fasta).out.output)

maxulysse: Shouldn't that generation from fasta done at the pipeline level, outside of this subworkflow?

matthdsm: That would be preferable yes, but then the entire submodule is kind of moot, as it's just a bedtools split 😅

maxulysse: In term of processes yes I agree. But for me the most important part is the channel operations that's the magic we want to share in between pipelines.

matthdsm: Works for me! In that case, I'd toss the fai dependency and just go with the bed stuff 👍🏻

nvnieuwk: Allright! I'll change it when I got some time :)

nvnieuwk: There really isn't that much code left indeed, even operation wise it's pretty simple.

matthdsm:
@nvnieuwk , can you allow outside edits on this PR? That way I can finish it up and integrate in `germline`

matthdsm (approved):
nvm, it was already finished!

subworkflows/nf-core/bed_scatter_bedtools/main.nf
```diff
@@ -0,0 +1,46 @@
...
+    ch_bed          // channel: [ meta, bed, scatter_count ]
+
+    main:
+
+    ch_versions = Channel.empty()
+
+    ch_bed
+        .map(
+            { meta, bed, scatter_count ->
+                new_meta = meta.clone()
```
FriederikeHanssen, line 17: is this doing a deep copy? We have had many issues in the past with shallow copies

nvnieuwk: Fair point, I'll fix it

nvnieuwk: Added!

subworkflows/nf-core/bed_scatter_bedtools/main.nf
```diff
@@ -0,0 +1,45 @@
...
+    ch_bed          // channel: [ meta, bed, scatter_count ]
+
+    main:
+
+    ch_versions = Channel.empty()
+
+    ch_bed
+        .map(
+            { meta, bed, scatter_count ->
+                meta = meta + [subwf_scatter_count:scatter_count]
```
matthdsm, line 17: How about a more generic `scatter_count` key?

nvnieuwk: My main thought was that I should create a key that isn't in the input meta. `scatter_count` might be a bit too common? I mean I'm not sure, but that's the reasoning behind it (I'll use the same value in the variant calling subwf)

matthdsm: It's just that I'm not much of a fan of pushing extra meta keys on users 😅  get's real confusing real fast, especially if we intend to use the splitting function in conjunction with another subwf which isn't tailored to it.

nvnieuwk: Yeah, I understand. Would it be better if I were to just add it as a constant value to each bed? the user can do with it whatever they want then.

matthdsm: I think we should ask the #sarek overlords, they've got more experience with stuff like this.
@FriederikeHanssen , @maxulysse, opinions?

FriederikeHanssen: @nvnieuwk not sure I understand what you were hinting at, can you elaborate?

FriederikeHanssen: yes that works well. users can then put this in the meta map to use in other workflows: https://github.com/nf-core/sarek/blob/96749f742197e828f4632cc2a7481190e7f642ac/subworkflows/local/prepare_intervals/main.nf#L107

matthdsm: Allrighty then! +1 on the value output

nvnieuwk: allright, I'll fix it!

nvnieuwk: Added!

matthdsm:
LGTM!

subworkflows/nf-core/bed_scatter_bedtools/main.nf
FriederikeHanssen: I guess this would be a general discussion point here. Should the scatter count be rather supplied as a param by the user inthe modules.config. Currently I guess usage ouwld be, pass in param in the pipeline -> add to bed file channel --> add to meta here to use in modules.config --> remove from meta again, add back to channel. Might be easier to just have the param directly accessed in the modules.config?

matthdsm: I'm not really sure it should ever be added to meta in this workflow actually...

matthdsm:
In conjuntion with this, should we also create a `bed_scatter_gatk`, which would perhaps be better suited towards use with gatk, using `SplitIntervals`https://gatk.broadinstitute.org/hc/en-us/articles/9570513631387-SplitIntervals?

subworkflows/nf-core/bed_scatter_bedtools/main.nf
matthdsm: Since the scatter count is a mandatory input, I'd treat it like a `val` input for the workflow.

nvnieuwk: Yeah but the fact is that it could be dependent on the specific input file. That's the reason I put them together.

nvnieuwk:
> In conjuntion with this, should we also create a `bed_scatter_gatk`, which would perhaps be better suited towards use with gatk, using `SplitIntervals`https://gatk.broadinstitute.org/hc/en-us/articles/9570513631387-SplitIntervals?

Good idea! That one should just be a copy paste if this subwf :p

nvnieuwk:
Can this be merged? I think all issues are solved

PR #2603, opened 2022-12-01
Add tabular to GCT module for GSEA

pinin4fjords:
Per conversation with @drpatelh , adding a tabular to GCT format converter for preparing expression matrices for use with GSEA. Works with CSV or TSV format inputs, since there's some inconsistency about which one is used where. 

Questions: 

- Do I need to allow for non-numeric columns such as the symbol column the rnaseq workflow leaves in the matrix? (note that I stripped it in a process in the test workflow)? I decided that specifying that was overkill and best deferred to the workflow builders, but other opinions may differ.

## PR checklist

- [x] Add a resource `label`


modules/nf-core/custom/tabulartogseagct/main.nf
```diff
@@ -0,0 +1,33 @@
+process CUSTOM_TABULARTOGSEAGCT {
+    tag "$meta.id"
+    label 'process_single'
+
```
nvnieuwk, line 4: I would still add a container here for reproducibility. (It will work in most cases, but we gotta think of those edge cases :) )

modules/nf-core/custom/tabulartogseagct/main.nf
```diff
@@ -0,0 +1,33 @@
...
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def separator = tabular.getName().endsWith(".csv") ? ',' : '\\t'
+    """
+    n_columns=\$(head -n 1 $tabular | tr "$separator" "\\n" | wc -l)
```
nvnieuwk, line 20: Also this should be possible in Groovy code inside your pipeline, but would probably take a bit more code. For me it's okay to add this to `custom`, especially if it could be used by other people to work with the GSEA modules

tests/modules/nf-core/custom/tabulartogseagct/main.nf
```diff
@@ -0,0 +1,60 @@
...
+
+workflow test_custom_tabulartogseagct {
+    
+    input = strip_column(file(params.test_data['mus_musculus']['genome']['rnaseq_matrix'], checkIfExists: true))
+        .map{
+            tuple([ id:'test' ], it)
+        }
+
+    CUSTOM_TABULARTOGSEAGCT ( input )
+}
```
nvnieuwk, line 48: These could be simplified using Groovy code: https://www.nextflow.io/docs/latest/script.html in combination with CollectFile (https://www.nextflow.io/docs/latest/operator.html?highlight=collectfile#collectfile), this way these files won't be in the output directory (and the tests won't pick them up). This would also be better for reproducibility as these may not work on all systems

nvnieuwk (changes requested):

nvnieuwk (approved):
LGTM! Good job with the fixes! :rocket:

modules/nf-core/custom/tabulartogseagct/meta.yml
```diff
@@ -0,0 +1,37 @@
...
+  - "tabulartogseagct":
+      description: "Convert a TSV or CSV with features by row and observations by column to a GCT format file as consumed by GSEA"
+
+input:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing matrix information
+        e.g. [ id:'test' ]
+  - tsv:
```
jfy133, line 16: I'm too late, but this should be called `tabular` not `tsv.

modules/nf-core/custom/tabulartogseagct/meta.yml
```diff
@@ -0,0 +1,37 @@
...
+      type: file
+      description: TSV or CSV file
```
jfy133, line 18: Given this is very specific module for a very specific tool, I would make this more specific than just 'any' TSV/CSV file

PR #2617, opened 2022-12-05
Add GSEA module

pinin4fjords:
Adding a wrapper for the GSEA command line utility, run in GSEA mode (there are other modes that may need separate wrappers later).

## PR checklist

Closes https://github.com/nf-core/modules/issues/2584

- [x] Add a resource `label`


nvnieuwk (commented):
LGTM apart from the fixed `meta` fields (but it seems like a tricky one here)

modules/nf-core/gsea/gsea/main.nf
```diff
@@ -0,0 +1,76 @@
...
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    # Run GSEA
+    gsea-cli GSEA \
+        -res $gct \
+        -cls ${cls}#${meta.target}_versus_${meta.reference} \
```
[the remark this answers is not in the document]
jfy133: I agree with @nvnieuwk here! Is `-cls` mandatory here? If the tool fails if you dont provide it, you can just let the tool just that and the developer will learn ;) Otherwise you could just check that `-cls` exists in the `ext.args` as Nicolas suggests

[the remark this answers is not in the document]
pinin4fjords: @nvnieuwk @jfy133 Thanks for the feedback.

The issue is, I'd really like to be able to refer to the treatment and reference levels in the output files (I want the receiving workflows to 'know' which files refer to which side of the contrast), which kind of predicates meta or a `val()', right? I can't refer to an args map there.

Note that CLS is a mandatory file input providing the classes, combined with that contrast definition :-)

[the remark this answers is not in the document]
pinin4fjords: Or would the solution be to use static 'reference' and 'target' and rename the outputs?

modules/nf-core/gsea/gsea/main.nf
```diff
@@ -0,0 +1,77 @@
...
+    ln -s \$(ls ranked_gene_list_${args2.target}_versus_${args2.reference}_*.tsv) ranked_gene_list_${args2.target}_versus_${args2.reference}.tsv
+    ln -s \$(ls gsea_report_for_${args2.reference}_*.html) gsea_report_for_reference_${args2.reference}.html
+    ln -s \$(ls gsea_report_for_${args2.reference}_*.tsv) gsea_report_for_reference_${args2.reference}.tsv
+    ln -s \$(ls gsea_report_for_${args2.target}_*.html) gsea_report_for_target_${args2.target}.html
+    ln -s \$(ls gsea_report_for_${args2.target}_*.tsv) gsea_report_for_target_${args2.target}.tsv
+    popd
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        gsea: 4.3.2
```
jfy133, line 74: Don't embed directly into versions.yml! Please follow the convention/examples on what to do when a tool doesn't produce them itself [^1]

[^1]: https://nf-co.re/docs/contributing/modules#general point 4, homoe/annotatepeaks example

modules/nf-core/gsea/gsea/main.nf
```diff
@@ -0,0 +1,77 @@
...
+    def args = task.ext.args ?: ''
+    def args2 = task.ext.args2 ?: [ ]
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    # Run GSEA
+    gsea-cli GSEA \
+        -res $gct \
+        -cls ${cls}#${args2.target}_versus_${args2.reference} \
+        -gmx $gene_sets \
+        -out \$(pwd) \
```
jfy133, line 48: why not just `.`?

modules/nf-core/gsea/gsea/main.nf
```diff
@@ -0,0 +1,77 @@
...
+
+    input:
+    tuple val(meta), path(gct), path(cls), path(gene_sets)
+
+    output:
+    tuple val(meta), path("*/*.rpt")                                                                                                        , emit: param_log
+    tuple val(meta), path("*/index.html")                                                                                                   , emit: index
+    tuple val(meta), path("*/heat_map_corr_plot.html")                                                                                      , emit: heat_map_corr_plot
+    tuple val(meta), path("*/gsea_report_for_reference_*.tsv"),  path("*/gsea_report_for_target_*.tsv")                                     , emit: report_tsvs
+    tuple val(meta), path("*/gsea_report_for_reference_*.html"),  path("*/gsea_report_for_target_*.html")                                   , emit: report_htmls
```
jfy133, line 18: These should be separate outputs channels, you can combine them after [^1]

[^1]: https://nf-co.re/docs/contributing/modules#inputoutput-options point 5

modules/nf-core/gsea/gsea/main.nf
```diff
@@ -0,0 +1,77 @@
...
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/gsea:4.3.2--hdfd78af_0':
+        'quay.io/biocontainers/gsea:4.3.2--hdfd78af_0' }"
+
+    input:
+    tuple val(meta), path(gct), path(cls), path(gene_sets)
+
+    output:
+    tuple val(meta), path("*/*.rpt")                                                                                                        , emit: param_log
+    tuple val(meta), path("*/index.html")                                                                                                   , emit: index
```
jfy133, line 15: Ideally these should be emitted with channel names such as `rpt` and `index_html` (the latter to disabiduage from the `corr_plot_html` below

modules/nf-core/gsea/gsea/main.nf
```diff
@@ -0,0 +1,77 @@
...
+
+    script:
+    def args = task.ext.args ?: ''
+    def args2 = task.ext.args2 ?: [ ]
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    # Run GSEA
+    gsea-cli GSEA \
+        -res $gct \
+        -cls ${cls}#${args2.target}_versus_${args2.reference} \
```
jfy133, line 46: I still don't like this, but at least should be reversed so this is just `args` and then `args2` is for the 'raw' user stuff.

Is there a reason why you can't just hardcode `target` and `reference` in the name, but carry that information in the `meta map` (and then e.g. publishDir those files in a directory named after those, or even use the meta info  to insert into `prefix`?

I know you  really want to be able to refer to it, but I'm worried again it is not portable.

Or maybe even just `args`, `args2`, `args3`? I'm not really sure exactly where to document this though...

oooooooooooorrrrrrrrrrrrrrrrrrrrrr

we allow a `val` input channel exception here ;)

modules/nf-core/gsea/gsea/meta.yml
```diff
@@ -0,0 +1,138 @@
+name: "gsea_gsea"
+description: run the Broad Gene Set Enrichment tool in GSEA mode
+keywords:
+  - gene_set_analysis
```
jfy133, line 4: you need many more keywords here: `enrichment`, `gsea`, `gene set` etc.

modules/nf-core/gsea/gsea/meta.yml
```diff
@@ -0,0 +1,138 @@
...
+      homepage: "http://www.gsea-msigdb.org/gsea/index.jsp"
+      documentation: "https://software.broadinstitute.org/cancer/software/gsea/wiki/index.php/Main_Page"
+      doi: "10.1073/pnas.0506580102"
+      licence: "['BSD-3-clause']"
+
+input:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing information on inputs and contrast, at a minimum
```
jfy133, line 17: This is confused with `ext.args2

modules/nf-core/gsea/gsea/meta.yml
```diff
@@ -0,0 +1,138 @@
...
+      pattern: "*.{gct}"
+  - cls:
+      type: file
+      description: CL file with the classes of the samples in the GCT file
+      pattern: "*.{gct}"
+  - gene_sets:
+      type: file
+      description: GMX or GMT file with gene sets
+      pattern: "*.{gmx,gmt}"
+  - ext.args2:
```
jfy133, line 32: Not a channel component! I would rather you say this in `meta` as you've already done, and say how the `meta` info you need should be inserted into the respective args{1,2,3} in `module.config`

modules/nf-core/gsea/gsea/meta.yml
```diff
@@ -0,0 +1,138 @@
...
+        target groups. Can be a closure referencing the meta.
+
+output:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing information on inputs and contrast, at a minimum
+        an ID, reference and target e.g. [ id:'test', reference:"WT",
+        target:"MUT" ]
+  - report_tsvs:
```
jfy133, line 45: Please keep these entries in the same order as the outpu tchannels (tripped me up at first!, I thought you were missing abunch!)

jfy133 (changes requested):

modules/nf-core/gsea/gsea/main.nf
pinin4fjords: Ahh- thanks, good to know.

modules/nf-core/gsea/gsea/meta.yml
pinin4fjords: Thanks!

modules/nf-core/gsea/gsea/meta.yml
```diff
@@ -0,0 +1,143 @@
...
+
+input:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing metadata, at a minimum an id and the reference
+        and target levels e.g. e.g. [ id:'test', reference:"WT", target:"MUT" ].
+        Reference and target levels can then be provided to the module via an ext.args
+        specification for the process like: ext.args = { [ "reference":
+        "$meta.reference", "target": "$meta.target" ] }
```
jfy133, line 24: If we go with this method, then should specif that `reference` and `target` keys are mandatory

modules/nf-core/gsea/gsea/meta.yml
```diff
@@ -0,0 +1,143 @@
...
+      documentation: "https://software.broadinstitute.org/cancer/software/gsea/wiki/index.php/Main_Page"
+      doi: "10.1073/pnas.0506580102"
+      licence: "['BSD-3-clause']"
+
+input:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing metadata, at a minimum an id and the reference
+        and target levels e.g. e.g. [ id:'test', reference:"WT", target:"MUT" ].
```
jfy133, line 21:
```suggestion
        and target levels e.g. [ id:'test', reference:"WT", target:"MUT" ].
```

modules/nf-core/gsea/gsea/meta.yml
jfy133: Ah this is a good point, it is actually on the docs page, just not under guidelines 😅 [^1]

[^1]: https://nf-co.re/docs/contributing/modules#new-module-workflow point 6.2

jfy133 (commented):
We just need to sort out this reference/target business but otherwise looks pretty close to me

modules/nf-core/gsea/gsea/main.nf
maxulysse: We do that in sarek for some specific args/files combo:
https://github.com/nf-core/sarek/blob/master/conf/igenomes.config#L24

jfy133: So you would think it's acceptable?

modules/nf-core/gsea/gsea/main.nf
```diff
@@ -0,0 +1,80 @@
...
+    tuple val(meta), path("*/gset_rnd_es_dist*.png")                                                                                        , emit: gene_set_dist, optional: true
+    tuple val(meta), path("*/*.zip")                                                                                                        , emit: archive, optional: true
+    path "versions.yml"                                                                                                                     , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def VERSION = '4.3.2' // WARN: Version information not provided by tool on CLI. Please update this string when bumping container versions.
+    def args = task.ext.args ?: [ ]
```
Midnighter, line 42: Since you are referencing keys in `args`, I guess the default should be an empty map? That's quite unusual, though, it's almost always a simple string.

```suggestion
    def args = task.ext.args ?: [:]
```

modules/nf-core/gsea/gsea/main.nf
```diff
@@ -0,0 +1,80 @@
...
+    def args2 = task.ext.args2 ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    # Run GSEA
+    gsea-cli GSEA \
+        -res $gct \
+        -cls ${cls}#${args.target}_versus_${args.reference} \
```
mahesh-panchal, line 49:
```suggestion
        -cls ${cls}#${args.target}_versus_${args.reference} \
```
From the way `target` and `reference` are used throughout the code, they appear to be required inputs and to me should be `val` input channels as they're required for the function of the script.
Even if they're part of the meta map, they can be separated made into their own channel using `multiMap`. These kind of args should not be supplied by configuration files as they're for the user reconfigure tool settings.

modules/nf-core/gsea/gsea/main.nf
jfy133: Ok to summarise slack @pinin4fjords 

- Will the tool fail execute if you don't supply the `#reference_versus_target` string : if so use `input` channels to define `reference` and `target`
- If the tool can run without them: you should leave it to the pipeline developer to pass whatever string they want via an `args`

pinin4fjords: > * Will the tool fail execute if you don't supply the `#reference_versus_target` string : if so use `input` channels to define `reference` and `target`
> * If the tool can run without them: you should leave it to the pipeline developer to pass whatever string they want via an `args`

To reiterate from the [slack thread](https://nfcore.slack.com/archives/CJRH30T6V/p1670481521205589), yes, it will fail unless the .cls file has only two classes, which would be hard to ensure. So we're going with a separate channel to provide the reference and target.

modules/nf-core/gsea/gsea/main.nf
pinin4fjords: Thanks @mahesh-panchal - I've implemented those inputs as a new channel

modules/nf-core/gsea/gsea/main.nf
pinin4fjords: I'm not sure how the usual string method would have been useful in this context, I'd have had to parse the values out of the string (since the options were not passed through verbatim to gsea as is often the case), which the map helped me to avoid. 


But in any case I've now switched to the channel usage instead of the args map

modules/nf-core/gsea/gsea/main.nf
```diff
@@ -0,0 +1,80 @@
...
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    # Run GSEA
+    gsea-cli GSEA \
+        -res $gct \
+        -cls ${cls}#${target}_versus_${reference} \
+        -gmx $gene_sets \
+        -out . \
+        --rpt_label $prefix \
+        $args
```
jfy133, line 53: You need to escape the backslashes, no? I'm surprised the module tests are passing :eyes:

jfy133 (approved):
That'll do ;) honorary :cookie: for all the pummeling on this one

modules/nf-core/gsea/gsea/main.nf
mahesh-panchal: The single backslashes allow a single string to be written over multiple lines, but in the file, they will still be all one line. If one wants the backslashes to appear in the code ( we do ), then they should be escaped (`\\`)

modules/nf-core/gsea/gsea/main.nf
```diff
@@ -0,0 +1,80 @@
...
+    label 'process_single'
+
+    conda (params.enable_conda ? "bioconda::gsea=4.3.2" : null)
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/gsea:4.3.2--hdfd78af_0':
+        'quay.io/biocontainers/gsea:4.3.2--hdfd78af_0' }"
+
+    input:
+    tuple val(meta), path(gct), path(cls), path(gene_sets)
+    tuple val(meta2), val(reference), val(target)
```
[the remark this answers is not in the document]
pinin4fjords: I was just under the impression that tuple channels needed metas. I can take it out.

modules/nf-core/gsea/gsea/main.nf
```diff
@@ -0,0 +1,80 @@
...
+    tuple val(meta), path("*/global_es_histogram.png")                                                                                      , emit: histogram
+    tuple val(meta), path("*/heat_map_1.png")                                                                                               , emit: heatmap
+    tuple val(meta), path("*/pvalues_vs_nes_plot.png")                                                                                      , emit: pvalues_vs_nes_plot
+    tuple val(meta), path("*/ranked_list_corr_2.png")                                                                                       , emit: ranked_list_corr
+    tuple val(meta), path("*/[!gene_set_size|gsea_report|ranked_gene_list]*.tsv")                                                           , emit: gene_set_tsv, optional: true
+    tuple val(meta), path("*/[!gsea_report|heat_map_corr_plot|index|pos_snapshot|neg_snapshot]*.html")                                      , emit: gene_set_html, optional: true
+    tuple val(meta), path("*/[!butterfly|enplot|global_es_histogram|gset_rnd_es_dist|heat_map|pvalues_vs_nes_plot|ranked_list_corr]*.png")  , emit: gene_set_heatmap, optional: true
```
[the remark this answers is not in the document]
pinin4fjords: No, there's not, you get e.g. a .png for every gene set in the input listing. So what I'm doing here is taking out every png that is something else. I know it's messy and fragile, but it works, and I spent a fair amount of time testing.

modules/nf-core/gsea/gsea/main.nf
pinin4fjords: Thanks- sorted.

modules/nf-core/gsea/gsea/main.nf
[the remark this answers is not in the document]
mahesh-panchal: It's definitely not a requirement that all tuples use meta maps. It's only required for the first input, when a meta map is necessary for processing. There's no point forcing developers to add data that isn't used.

modules/nf-core/gsea/gsea/main.nf
```diff
@@ -0,0 +1,80 @@
...
+    (ls *.zip 2>/dev/null || true) | while read -r l; do
+        ln -s \$l ${prefix}.Gsea.rpt.zip
+    done
+
+    ln -s \$(ls ranked_gene_list_${target}_versus_${reference}_*.tsv) ranked_gene_list_${target}_versus_${reference}.tsv
+    ln -s \$(ls gsea_report_for_${reference}_*.html) gsea_report_for_reference_${reference}.html
+    ln -s \$(ls gsea_report_for_${reference}_*.tsv) gsea_report_for_reference_${reference}.tsv
+    ln -s \$(ls gsea_report_for_${target}_*.html) gsea_report_for_target_${target}.html
+    ln -s \$(ls gsea_report_for_${target}_*.tsv) gsea_report_for_target_${target}.tsv
+    popd
```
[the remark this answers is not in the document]
pinin4fjords: Thanks for this. The looping had been made more difficult by an earlier workaround to label things as 'reference' based on `args`, for the output pattern to pick up. That's no longer necessary since I can refer directly to `reference` and `target` since they're channel inputs- hooray! Obviously I shouldn't have been using subshells regardless.

Now fixed, and looped.

pinin4fjords:
Thanks @mahesh-panchal @jfy133 for that last batch of feedback and the approval. All final points addressed and code changes passing, merging now.

PR #2621, opened 2022-12-06
Set several QC processes to "single"

matthdsm:
Set selector to `process_single` for single core QC processes

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`

### DIFF
--- a/modules/nf-core/samtools/idxstats/main.nf
+++ b/modules/nf-core/samtools/idxstats/main.nf
@@ -1,6 +1,6 @@
 process SAMTOOLS_IDXSTATS {
     tag "$meta.id"
-    label 'process_low'
+    label 'process_single'
 
     conda (params.enable_conda ? "bioconda::samtools=1.16.1" : null)
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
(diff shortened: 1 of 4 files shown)


modules/nf-core/samtools/idxstats/main.nf
```diff
@@ -1,6 +1,6 @@
 process SAMTOOLS_IDXSTATS {
     tag "$meta.id"
-    label 'process_low'
+    label 'process_single'
```
Midnighter, line 3: This seems to be in conflict with the option given to the command `--threads ${task.cpus-1}`. The -1 doesn't make sense to me in general but are you saying that `samtools idxstats` cannot operate multithreaded or it doesn't need to? In that case I would also remove the `--threads` option or hard code it to 1.

Midnighter (changes requested):
Just the confusion on `samtools idxstats` otherwise this looks good.

modules/nf-core/samtools/idxstats/main.nf
Midnighter: Okay, going with your take, I guess this is fine.

Midnighter (approved):

muffato (approved):

modules/nf-core/samtools/idxstats/main.nf
matthdsm: I can confirm the `--threads` option is ignored in this tool. The `-1` value was to account for the 2 cpu's allotted through `process_low`. I actually should've removed it, good catch @Midnighter

matthdsm:
Thanks everyone!

PR #2622, opened 2022-12-06
Strip out meta from bracken to allow more flexibility for pipeline de…

jfy133:
…vs (as per guidelines)

x-ref: https://github.com/nf-core/taxprofiler/issues/168

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`

### DIFF
--- a/modules/nf-core/bracken/bracken/main.nf
+++ b/modules/nf-core/bracken/bracken/main.nf
@@ -21,12 +21,9 @@ process BRACKEN_BRACKEN {
     task.ext.when == null || task.ext.when
 
     script:
-    def threshold = meta.threshold ?: 10
-    def taxonomic_level = meta.taxonomic_level ?: 'S'
-    def read_length = meta.read_length ?: 150
-    def args = task.ext.args ?: "-l ${taxonomic_level} -t ${threshold} -r ${read_length}"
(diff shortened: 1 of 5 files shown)


Midnighter:
My idea was to add these as input channels then. Rather than removing them completely and only using `ext.args`. It's quite useful to run Bracken, for example, on family, genus, and species level. Plus you may have samples with different read lengths and still want to process them all in one pipeline run.

jfy133:
But the pipeline developer can still insert these via `args` no? That's exactly what we already do with taxprofiler with other tools. They don't need to be input channels

jfy133:
> I don't see how you can vary a parameter via `args`, e.g., how can I set for Bracken `-l F`, `-l G`, or `-l S` for the same sample in the same pipeline run?

Still in the meta! But you don't make any assumptions on the names. You duplicate the element in the channel, and modify the meta.

Midnighter:
Okay, fair enough. Most of these changes should be implemented on the `nf-core/modules` first, though, and then we update the module here on the pipeline?

jfy133:
> Okay, fair enough. Most of these changes should be implemented on the `nf-core/modules` first, though, and then we update the module here on the pipeline?

This is in the modules repo ;)

PR #2648, opened 2022-12-13
new module survivor/merge

### LINKED ISSUE #2647: new module: survivor/merge

nvnieuwk:
## PR checklist

Closes #2647 

- [x] Add a resource `label`


modules/nf-core/survivor/merge/main.nf
```diff
@@ -0,0 +1,51 @@
...
+
+    output:
+    tuple val(meta), path("*.vcf")   , emit: vcf
+    path "versions.yml"              , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: [:]
```
pinin4fjords, line 21: I was confused by the need to have args as a map- but I take it this is because SURVIVOR has positional arguments and you need to ensure the order?

I think if these are compulsory non-file arguments then we'd probably be better to formulate these as channel inputs (see https://github.com/nf-core/nf-co.re/pull/1486), assuming those are the accepted new guidelines.

pinin4fjords (changes requested):
I think, unless ongoing debate reverses, that the args should be part of a channel, like [this](https://github.com/nf-core/modules/blob/93a19352484d0a14fcae0b33ad24da7d99036583/modules/nf-core/gsea/gsea/main.nf#L12). 

Might require some more discussion ;-)

modules/nf-core/survivor/merge/main.nf
nvnieuwk: These also state that if there are defaults, they should be able to be overwritten by via `args`. I think that would be the best here. I'll add the defaults here

modules/nf-core/survivor/merge/main.nf
```diff
@@ -0,0 +1,51 @@
...
+    def prefix = task.ext.prefix ?: "${meta.id}"
+
+    vcfs.each{
+        println(it)
```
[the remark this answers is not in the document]
nvnieuwk: Good spot! Thanks!!

modules/nf-core/survivor/merge/main.nf
```diff
@@ -0,0 +1,56 @@
...
+    ls *.vcf > names.txt
+
+    SURVIVOR merge \\
+        names.txt \\
+        ${args.max_distance_breakpoints} \\
+        ${args.min_supporting_callers} \\
+        ${args.account_for_type} \\
+        ${args.account_for_sv_strands} \\
+        ${args.estimate_distanced_by_sv_size} \\
+        ${args.min_sv_size} \\
```
adamrtalbot, line 48: Caveat - I think these should all be supplied via `ext.args` as a string, at the cost of readability. However, if you do decide to go down this route I think it should be:
```suggestion
        ${max_distance_breakpoints} \\
        ${min_supporting_callers} \\
        ${account_for_type} \\
        ${account_for_sv_strands} \\
        ${estimate_distanced_by_sv_size} \\
        ${min_sv_size} \\
```

nvnieuwk: I've adjusted this on the last commit :), now with defaults too. Adding it to `args` as a string would be preferable indeed, but would make it a bit harder to really know what is what IMO

modules/nf-core/survivor/merge/main.nf
```diff
@@ -0,0 +1,75 @@
...
+    script:
+    def args = task.ext.args ?: [:]
+    def prefix = task.ext.prefix ?: "${meta.id}"
+
+    def max_distance_breakpoints        = args.max_distance_breakpoints         ?: 0.2
+    def min_supporting_callers          = args.min_supporting_callers           ?: 1
+    def account_for_type                = args.account_for_type                 ?: 0
+    def account_for_sv_strands          = args.account_for_sv_strands           ?: 0
+    def estimate_distanced_by_sv_size   = args.estimate_distanced_by_sv_size    ?: 0
+    def min_sv_size                     = args.min_sv_size                      ?: 20
```
jfy133, line 29: Given these are for positional arguments, I would say to make the module much less brittle I would use `input` args for these.

You should not set hidden defaults in pipelines either, which is why I tihnk this is a bad idea.

nvnieuwk: Okay fair point there. These arguments all need to be used in a flexible way so this will be a long `input` channel :grin:

modules/nf-core/survivor/merge/main.nf
```diff
@@ -0,0 +1,73 @@
...
+            error "Gzipped files are not supported by Survivor, please gunzip your VCF files first."
+            // https://github.com/fritzsedlazeck/SURVIVOR/issues/158
+        }
+    }
+
+    """
+    ls *.vcf > names.txt
+
+    SURVIVOR merge \\
+        names.txt \\
```
pinin4fjords, line 40: Bit of process substitution to save a write?

```suggestion
    SURVIVOR merge \\
        <(ls *.vcf) \\
```

nvnieuwk: Oh cool I didn't know this was possible, thanks!

pinin4fjords (approved):
LGTM in light of upcoming guideline changes on non-file madatory inputs (assuming my last suggestion doesn't break anything ;-) )

PR #2683, opened 2022-12-20
Plink recode

atrigila:
add plink/recode
 
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`


modules/nf-core/plink/recode/meta.yml
```diff
@@ -0,0 +1,145 @@
...
+  - bim:
+      type: file
+      description: PLINK extended MAP file
+      pattern: "*.{bim}"
+  - fam:
+      type: file
+      description: PLINK sample information file
+      pattern: "*.{fam}"
+
+output:
```
nvnieuwk, line 35: Could you maybe specify which arguments are needed to get the different output files?

modules/nf-core/plink/recode/main.nf
```diff
@@ -0,0 +1,59 @@
...
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+
+    """
+    plink \\
+        --bfile ${meta.id} \\
```
nvnieuwk, line 49: Why do you use `meta.id` here instead of `prefix`? :) (Just curious what this argument does, for most cases it is preffered to use `prefix` for the greater flexibility)

modules/nf-core/plink/recode/main.nf
```diff
@@ -0,0 +1,59 @@
...
+        --threads $task.cpus \\
+        --recode \\
```
nvnieuwk, line 51: Is this a mandatory argument that always needs to be given?

modules/nf-core/plink/recode/meta.yml
```diff
@@ -0,0 +1,145 @@
...
+  - "plink":
+      description: "Whole genome association analysis toolset, designed to perform a range of basic, large-scale analyses in a computationally efficient manner."
+      homepage: "https://www.cog-genomics.org/plink"
+      documentation: "https://www.cog-genomics.org/plink/1.9/data#recode"
+      tool_dev_url: "https://www.cog-genomics.org/plink/1.9/dev"
+      doi: ""
+      licence: "['GPL']"
+
+input:
+  # Only when we have meta
```
nvnieuwk, line 15:
```suggestion
```

modules/nf-core/plink/recode/meta.yml
```diff
@@ -0,0 +1,145 @@
...
+      type: file
+      description: PLINK extended MAP file
+      pattern: "*.{bim}"
+  - fam:
+      type: file
+      description: PLINK sample information file
+      pattern: "*.{fam}"
+
+output:
+  #Only when we have meta
```
nvnieuwk, line 36:
```suggestion
```

modules/nf-core/plink/recode/meta.yml
```diff
@@ -0,0 +1,145 @@
...
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. [ id:'test', single_end:false ]
+  #
```
nvnieuwk, line 42:
```suggestion
```

nvnieuwk (changes requested):
Very nice work for your first module :partying_face: here are some comments and suggestions!

modules/nf-core/plink/recode/main.nf
atrigila: I am not sure if I understand correctly, but `--recode` needs to be specified in order to produce the output. Otherwise no changes from input to output would happen.

modules/nf-core/plink/recode/main.nf
atrigila: Good question. In order to make PLINK_RECODE similar/consistent to previous PLINK modules (such as [plink_extract](https://github.com/nf-core/modules/blob/master/modules/nf-core/plink/extract/main.nf)) I used `${meta.id}`. I am not aware about the benefits of using prefix, though, would you suggest changing it?

nvnieuwk: What does the `-bfile` argument do exactly?

modules/nf-core/plink/recode/main.nf
nvnieuwk: Allright fine by me then :)

modules/nf-core/plink/recode/main.nf
atrigila: As per PLINK's documentation, "the` --bfile` flag causes the binary fileset plink.bed + plink.bim + plink.fam to be referenced.  If a prefix is given, it replaces all instances of 'plink'.". So tipically you add here the name/prefix of the these three files.

nvnieuwk: Ah! Okay, maybe you should specify this in the docs that the three input files need to have the same prefix. You could also automatically detect the name of these files and use this as input. `meta.id` won't always be the same as the prefix of the input files.
`bed.baseName` should do the trick in getting the name :)

atrigila: I was reading the reference and there is also: " `--bed`, `--bim`, and `--fam` let you specify the full name of one part of the PLINK 1 binary fileset, taking precedence over `--bfile`. "  I am going to think a bit more about how to handle this, then.

nvnieuwk: Ah I think it would be better if you can specify them separately :)

tests/modules/nf-core/plink/recode/main.nf
```diff
@@ -0,0 +1,23 @@
...
+
+workflow test_plink_recode {
+
+    input = [
+        [ id:'test', single_end:false ], // meta map
+        file(params.test_data['homo_sapiens']['genome']['syntheticvcf_short_vcf_gz'], checkIfExists: true)
+    ]
+    PLINK_VCF ( input )
+
+    bed_ch = PLINK_VCF.out.bed
```
matthdsm, line 16: Don't forget to update the tests!

modules/nf-core/plink/recode/main.nf
atrigila: Thank you for the suggestion! I addresed it here: https://github.com/nf-core/modules/pull/2683/commits/5c413babb4ea7d3af8390d5ecaacc051e6e60cf0 where each input is now specified separately.

tests/modules/nf-core/plink/recode/main.nf
atrigila: Modified the test so that it now works with 3 inputs: https://github.com/nf-core/modules/pull/2683/commits/6e6d4c666a8b4ca60f719bd25a373e09b87c69de

modules/nf-core/plink/recode/main.nf
nvnieuwk: Are these files different for each sample? If so I would leave the input channel as it was since they will always be linked to eachother that way (based on how the channel logic works in Nextflow)

atrigila: That's true. I haven't thought about that. So, if I understand correctly, the channel can be one, composed of a 'single' tuple but then I can call the elements of the channel individually in the script section. So instead of: 

```
    input:
    tuple val(meta), path(bed)
    tuple val(meta), path(bim)
    tuple val(meta), path(fam)
    
    script:
     """
    plink \\
        --bed ${bed}  \\
        --bim ${bim}  \\
        --fam ${fam}  \\ [etc]
      """
```
It is better to:

```
    input:
    tuple val(meta), path(bed),  path(bim), path(fam)
    
    script:
    """
    plink \\
        --bed ${bed}  \\
        --bim ${bim}  \\
        --fam ${fam}  \\ [etc]
   """
```

nvnieuwk: Yep that's correct!!!

atrigila: Here it is https://github.com/nf-core/modules/pull/2683/commits/50a3d4f1c2e6ed5c0819f4bc9d1769ee29183906 :)

nvnieuwk (approved):
LGTM :partying_face:

PR #2703, opened 2022-12-28
enhancement: add parameter for fastp to make writing of trimmed fastqs optional

### LINKED ISSUE #2700: [FEATURE] update fastp to support no-trim mode

it's possible to use the fastp to only produce QC output without the trimmed reads, but currently it is not possible to specify this mode for the nf-core fastp module. 

The module should accept an input parameter boolean called something like `perform_trimming` and if true, parameters indicating the paths of trimmed fastqs that will be written. 

i considered simply ignoring the output channel containing reads if i choose not to use them, but decided this might be problematic for storage reasons -- writing a new set of fastq that would not be used can take up a lot of unnecessary disk space. 

### Additional context

_No response_

anoronh4:
Closes #2700. with this PR, by defining the input channel `save_trimmed_pass` a user can define whether trimmed fastqs are written/saved, or not. This works for all three modes: interleaved, single_end, paired_end. This will help to save on storage when trimming is not necessarily desired but the qc provided by fastp is.  

## PR checklist

- [x] Add a resource `label`


modules/nf-core/fastp/main.nf
```diff
@@ -10,6 +10,7 @@ process FASTP {
     input:
     tuple val(meta), path(reads)
     path  adapter_fasta
+    val   save_trimmed_pass
     val   save_trimmed_fail
```
maxulysse, line 14: why pass before fail?

modules/nf-core/fastp/main.nf
```diff
@@ -30,6 +31,13 @@ process FASTP {
     def prefix = task.ext.prefix ?: "${meta.id}"
     def adapter_list = adapter_fasta ? "--adapter_fasta ${adapter_fasta}" : ""
     def fail_fastq = save_trimmed_fail && meta.single_end ? "--failed_out ${prefix}.fail.fastq.gz" : save_trimmed_fail && !meta.single_end ? "--unpaired1 ${prefix}_1.fail.fastq.gz --unpaired2 ${prefix}_2.fail.fastq.gz" : ''
+    def out_fq = ""
+    if ( save_trimmed_pass ) {
+        out_fq = meta.single_end ? "--out1 ${prefix}.fastp.fastq.gz" : "--out1 ${prefix}_1.fastp.fastq.gz --out2 ${prefix}_2.fastp.fastq.gz"
+        out_fq = task.ext.args?.contains('--interleaved_in') ? "${prefix}.fastp.fastq.gz" : out_fq
```
maxulysse, line 37:
```suggestion
        out_fq = task.ext.args?.contains('--interleaved_in') ? "${prefix}.fastp.fastq.gz" : ''
```

modules/nf-core/fastp/main.nf
```diff
@@ -30,6 +31,13 @@ process FASTP {
     def prefix = task.ext.prefix ?: "${meta.id}"
     def adapter_list = adapter_fasta ? "--adapter_fasta ${adapter_fasta}" : ""
     def fail_fastq = save_trimmed_fail && meta.single_end ? "--failed_out ${prefix}.fail.fastq.gz" : save_trimmed_fail && !meta.single_end ? "--unpaired1 ${prefix}_1.fail.fastq.gz --unpaired2 ${prefix}_2.fail.fastq.gz" : ''
+    def out_fq = ""
```
maxulysse, line 34:
```suggestion
    def out_fq = ''
```

modules/nf-core/fastp/main.nf
anoronh4: this would not give the proper result. I basically want line 37 to overwrite 36 only when `task.ext.args?.contains('--interleaved_in')`. it should not evaluate to `''`.

modules/nf-core/fastp/main.nf
anoronh4: not sure i understand the issue here. users would probably be more interested in the trimmed reads that pass filter, so i put it before for priority.

modules/nf-core/fastp/main.nf
anoronh4: rewrote logic for clarity here

anoronh4:
@maxulysse just wondering if there's any update on this PR. I have addressed or responded to each of the review comments, hope we can get this moving!

FriederikeHanssen:
sorry, maybe I am not fully understanding the issue. so in the case you are describing no output will be written, just the json and log files?

tests/modules/nf-core/fastp/main.nf
```diff
@@ -94,8 +100,52 @@ workflow test_fastp_paired_end_merged_adapterlist {
...
+//
+workflow test_fastp_interleaved_omit_trimmed {
+    input = [ [ id:'test', single_end:true ], // meta map
+              [ file(params.test_data['sarscov2']['illumina']['test_interleaved_fastq_gz'], checkIfExists: true) ]
+            ]
+    save_trimmed_pass = false 
+    save_trimmed_fail = false
+    save_merged       = false
+
+    FASTP ( input, [], save_trimmed_pass, save_trimmed_fail, save_merged )
```
Emiller88, line 150: Would it be too much to write these tests in nf-test instead?

maxulysse: I'd agree with @Emiller88 on that one

maxulysse:
@Emiller88 so do we want to merge, and rewrite all tests in nf-test?

jfy133:
This seems very broken with a lot merge conflict markers... It would need cleaning up ...

SPPearce:
Can I get some eyes on this now. 
I rewrote the tests, as I think they had been written at the start of our nf-test use.
I would like to test that the optional channels are indeed empty, but couldn't seem to manage that correctly at the moment. 
I don't think we need to check the contents of the other fastq files, and they are currently not working for me (commented out lines in the tests)

modules/nf-core/nonpareil/nonpareil/tests/nextflow.config
```diff
@@ -2,8 +2,13 @@ process {
 
     publishDir = { "${params.outdir}/${task.process.tokenize(':')[-1].tokenize('_')[0].toLowerCase()}" }
 
+<<<<<<< HEAD:tests/modules/nf-core/fastp/nextflow.config
```
jfy133, line 5: I don't think that's meant to be in there

jfy133 (approved):

jfy133:
I guess this is ok but I find that default behaviour is sub optimal.

The primary purpose of fastp is to do the read clean up, so by default I would expect to get those resulting reads.

I think just generating the QC and not using the FASTQ files is a rather odd edge case (imo) and would rather have that behaviour opt in....

SPPearce:
> I guess this is ok but I find that default behaviour is sub optimal.
> 
> The primary purpose of fastp is to do the read clean up, so by default I would expect to get those resulting reads.
> 
> I think just generating the QC and not using the FASTQ files is a rather odd edge case (imo) and would rather have that behaviour opt in....

I guess you mean that if you leave the channel empty then it will default to false? But the developer has to look at what the channels are to be able to plug the module in.

I'm running the demultiplex pipeline at the moment, and would like to get the QC report from fastp without the actual read trimming, because I'm going to pass them into a proprietary pipeline that is going to trim it anyway. 
Fastq

jfy133:
> > I guess this is ok but I find that default behaviour is sub optimal.
> > 
> > The primary purpose of fastp is to do the read clean up, so by default I would expect to get those resulting reads.
> > 
> > I think just generating the QC and not using the FASTQ files is a rather odd edge case (imo) and would rather have that behaviour opt in....
> 
> I guess you mean that if you leave the channel empty then it will default to false? But the developer has to look at what the channels are to be able to plug the module in.
> 
> I'm running the demultiplex pipeline at the moment, and would like to get the QC report from fastp without the actual read trimming, because I'm going to pass them into a proprietary pipeline that is going to trim it anyway. 
> Fastq

Actually no, more switch it to more like `dontsave_trimmed`. Agree empty channel is not a nice system, better to be an 'active' choice

Yeah I understand the justification, but I maintain most people would not be using a proprietary or alternative trimmer. The core  Fastp purpose is to do the read clean up, it's not just a QC reporter tool pole fastqc

SPPearce:
I would go for a negative argument (`skip` or `dontsave`) if it wasn't for the fact the other two channels are positive (`save`).
I don't see a large burden here on the developer, if they are passing reads in with four other empty channnels without looking at the module code/meta then they'll quickly discover it isn't making any reads.

jfy133:
Hmmm... I dunno. I'm still not convinced. Maybe you can ask for a vote on #team-maintainers

modules/nf-core/fastp/meta.yml
```diff
@@ -27,6 +27,9 @@ input:
       type: file
       description: File in FASTA format containing possible adapters to remove.
       pattern: "*.{fasta,fna,fas,fa}"
+  - save_trimmed_pass:
+      type: boolean
+      description: Specify true to save files that pass trimming thresholds, ending in `*.fastp.fastq.gz`
```
FriederikeHanssen, line 32: I am with James, but it's not a hill-to-die-on for me. However, if we go for this logic, I would like to have a big warning here, that it is not just about saving, which for me somewhat implies it is about the results directory only, BUT also to get any read output from the module.

SPPearce:
What about changing the parameter names to `write_` instead of `save_`?

modules/nf-core/fastp/main.nf
```diff
@@ -30,6 +31,9 @@ process FASTP {
     def prefix = task.ext.prefix ?: "${meta.id}"
     def adapter_list = adapter_fasta ? "--adapter_fasta ${adapter_fasta}" : ""
     def fail_fastq = save_trimmed_fail && meta.single_end ? "--failed_out ${prefix}.fail.fastq.gz" : save_trimmed_fail && !meta.single_end ? "--failed_out ${prefix}.paired.fail.fastq.gz --unpaired1 ${prefix}_1.fail.fastq.gz --unpaired2 ${prefix}_2.fail.fastq.gz" : ''
+    def out_fq  = save_trimmed_pass ? "${prefix}.fastp.fastq.gz" : '/dev/null'
```
maxulysse, line 34: where is this used?

mahesh-panchal:
What about `discard_fastq`?

modules/nf-core/fastp/meta.yml
maxulysse: Maybe allow by default, and have a `disable_save_trimmed_pass` instead?

maxulysse: I think everyone would rather have something instead of nothing

modules/nf-core/fastp/main.nf
SPPearce: Nowhere, well spotted.

SPPearce:
Following discussion we are going to change from `save_trimmed_pass` to `discard_trimmed_pass`. Then running with `false, false, false` will provide expected behaviour.

maxulysse (approved):

PR #2712, opened 2023-01-04
Simplify GSEA outputs, remove nasty symlinking, alow file-level prefixes

pinin4fjords:
## TL;DR;  

This fixes file prefixing in the GSEA module to avoid file name clashes downstream, and removes awkward symlinking. 

## Longer version:

The way I'd designed the GSEA output file handling has created problems for me in practice:

1. The prefix was used in such a way that it determined an output directory, but didn't prefix files themselves, making it difficult to work around file name clashes downstream. 
2. The symlinking I'd added to create predictable paths for use in CI for files otherwise containing a timestamp was creating issues with multiple matches in the output channels for certain patterns. 

I therefore propose this solution to do things differently:

1. Move files out of the results directory used by GSEA, and use the prefix at the file level.
2. Don't do the symlinking, but actually rename files, and correct cross-references in HTML files accordingly via use of sed.

I previously hesitated to actually rename the files themselves so I could avoid that correction of the cross references, but I think it's actually the lesser of the evils here. 

Note that the MD5Sums have changed where internal references to renamed files have been updated.

## PR checklist

- [x] Add a resource `label`

### DIFF
--- a/modules/nf-core/gsea/gsea/main.nf
+++ b/modules/nf-core/gsea/gsea/main.nf
@@ -66,7 +66,7 @@ process GSEA_GSEA {
     timestamp=\$(cat *.rpt | grep producer_timestamp | awk '{print \$2}')
 
     for pattern in _\${timestamp} .\${timestamp}; do
-        find . -name "*\${pattern}*" -execdir echo {} ';' | while read -r f; do
+        find . -name "*\${pattern}*" | sed "s|^\\./||" | while read -r f; do
             mv \$f \${f//\$pattern/}
         done
     done
--- a/modules/nf-core/gsea/gsea/main.nf
+++ b/modules/nf-core/gsea/gsea/main.nf
@@ -13,28 +13,28 @@ process GSEA_GSEA {
     path(chip) // Optional identifier mapping file
 
     output:
-    tuple val(meta), path("*.rpt")                                                                                                         , emit: rpt
-    tuple val(meta), path("*index.html")                                                                                                   , emit: index_html
-    tuple val(meta), path("*heat_map_corr_plot.html")                                                                                      , emit: heat_map_corr_plot
-    tuple val(meta), path("*gsea_report_for_${reference}.tsv")                                                                             , emit: report_tsvs_ref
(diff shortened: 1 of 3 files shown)


modules/nf-core/gsea/gsea/main.nf
```diff
@@ -43,43 +43,41 @@ process GSEA_GSEA {
     def VERSION = '4.3.2' // WARN: Version information not provided by tool on CLI. Please update this string when bumping container versions.
     def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
+
+    // Remove any trailing dots from prefix when passed as report label, so
+    // GSEA doesn't produce double-dotted top-level outputs
+    def rpt_label = prefix.replaceAll(".\$", "")
```
erikrikarddaniel, line 49: Shouldn't the pattern be "\.*$" to match trailing dots? This one looks like you're removing any character followed by a dollar.

erikrikarddaniel (approved):
Looks fine, although I have no experience of the tool. One tiny comment regarding a regexp that doesn't look right to me; fine if it works, so I'm approving. (The extremely long lines for emits made me wonder if it's so wise to _always_ line up the commas, wouldn't it be better to let the longest stick out?)

modules/nf-core/gsea/gsea/main.nf
pinin4fjords: Good catch- thanks. 

It actually needs to be `\\.$` to match a single trailing dot (`.` is interpreted as any character so would actually strip a single trailing character from all prefixes, which isn't what I was going for!). The escaping on the dollar was down to my double quotes:

```
> println('foo.bar.'.replaceAll(".", ''))
// (empty string)

> println('foo.bar.'.replaceAll("\\.", ''))
foobar

> println('foo.bar.'.replaceAll("\\.$", ''))
unknown recognition error type: groovyjarjarantlr4.v4.runtime.LexerNoViableAltException

> println('foo.bar.'.replaceAll('\\.$', ''))
foo.bar

> println('foo.bar'.replaceAll('\\.$', ''))
foo.bar
```

pinin4fjords:
Thanks for the review @erikrikarddaniel

PR #2715, opened 2023-01-04
Add --hardtrim* files as output for TrimGalore

drpatelh:
Partially closes https://github.com/nf-core/rnaseq/issues/922

`*report.txt` aren't produced by TrimGalore when `--hardtrim 3` or `--hardtrim 5` are set which means we can't get the read stats info. So I had to refactor the logic that filters channels if the FastQ files are empty after trimming.


erikrikarddaniel (approved):
LGTM.

PR #2738, opened 2023-01-11
Greater than 256

### LINKED ISSUE #391: CI fails if matrix is larger than 256

edmundmiller:
Closes #391

Progress


SPPearce:
@edmundmiller , is this needed any more?

maxulysse:
> @edmundmiller , is this needed any more?

I'm guessing no

edmundmiller:
Nope, hopefully we don't run into it now with https://github.com/adamrtalbot/detect-nf-test-changes, but I think if we just keep it to files it *should* be okay

PR #2757, opened 2023-01-20
add a new module sickle

BioInf2305:
## PR checklist

Closes #XXX 

- [x] Add a resource `label`


modules/nf-core/sickle/main.nf
```diff
@@ -0,0 +1,65 @@
...
+    tuple val(meta), path("*.log"), emit: log
+    path "versions.yml"           , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "${meta.id}"
+    def qual_type = "${meta.qual}"
```
nvnieuwk, line 26: Is `qual_type` a mandatory input? Otherwise you should add this to `ext.args`

nvnieuwk: It's also preffered to not force pipeline developers to use specific `meta` fields (`single_end` is the exception)

modules/nf-core/sickle/main.nf
```diff
@@ -0,0 +1,65 @@
...
+
+    if (meta.single_end){
+    """
+    sickle \\
+        se \\
+        $args \\
+        -f $reads \\
+        -t $qual_type \\
+        -o ${prefix}.se.trimmed.fastq.gz \\
+        -g \\
```
nvnieuwk, line 36: Is the `-g` argument mandatory here?

modules/nf-core/sickle/meta.yml
```diff
@@ -0,0 +1,53 @@
...
+
+input:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. [ id:'test', single_end:false, qual:'sanger'  ]
+  - reads:
+      type: file
+      description: List of input FastQ files of size 1 and 2 for single-end and paired-end data, respectively.
```
nvnieuwk, line 23: Can you also add here which file of the paired-end data has to be specified first?

nvnieuwk (changes requested):

modules/nf-core/sickle/main.nf
BioInf2305: > Is `qual_type` a mandatory input? Otherwise you should add this to `ext.args`

yes, "qual_type" is a mandatory argument (although by default it is set to "sanger" and it can take any of these three values: "solexa", "sanger" or "illumina". In the new PR, I removed "qual_type" from meta and put it as a separate input in input tuple. Do you have any suggestion on this? Should I completely remove "qual_type" ?

BioInf2305: > It's also preffered to not force pipeline developers to use specific `meta` fields (`single_end` is the exception)

I removed it from meta and put it as input in input tuple. Please review it. Thanks.

modules/nf-core/sickle/main.nf
BioInf2305: Nope, it is not mandatory argument but it gives compressed fastq output. I refer to couple of nf-core modules and noticed that it is preferred way to get output fastq file (notice "gzip" argument in here: https://github.com/nf-core/modules/blob/master/modules/nf-core/trimgalore/main.nf). However, I can remove it if it violates the standard templates of nf-core/modules. Please give your suggestion.

modules/nf-core/sickle/meta.yml
BioInf2305: It is done in the updated PR.

modules/nf-core/sickle/main.nf
nvnieuwk: I wouldn't add it as an input here if the tool automatically uses `sanger` as default and just let the user specify it in the `task.ext.args` then. The only reason `String` values should be given in the input is if the module 100% needs it to run (which isn't the case here)

modules/nf-core/sickle/main.nf
nvnieuwk: No that's perfect! Well done :)

modules/nf-core/sickle/main.nf
BioInf2305: Sorry, my mistake! It is not default and the tool does not work without quality type, therefore, I suggest to keep it as one of the input in tuple. Apart from this, is there any other change that you recommend me to carry out?

PR #2774, opened 2023-01-25
Update glimpse chunk

LouisLeNezet:
## PR checklist
Update glimpse chunk for better parallelization and version control.
The value of the region that will need to be iterated is now present in the input tuple (better for parallelization has prescribed by @nvnieuwk in previous PR).
The version number of the tools is also now dynamically obtained.


nvnieuwk (approved):
LGTM

PR #2776, opened 2023-01-25
Update: fix bug in conda scope check for antismash download databases

jfy133:
`.enabled` will not exist if the conda scope was never defined (e.g. if not using default nf-core profiles, but using a nf-core/config institutional profile)


jasmezz (approved):
:+1:

PR #2779, opened 2023-01-25
New module: glimpse ligate

LouisLeNezet:
## PR checklist
Here is all the files needed for the new module for the glimpse imputation software: "Ligate"

- [X] Add a resource `label`

### DIFF
--- a/modules/nf-core/glimpse/ligate/main.nf
+++ b/modules/nf-core/glimpse/ligate/main.nf
@@ -1,0 +1,35 @@
+process GLIMPSE_LIGATE {
+    tag "$meta.id"
+    label 'process_low'
+
+    conda "bioconda::glimpse-bio=1.1.1"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/glimpse-bio:1.1.1--hce55b13_1':
+        'quay.io/biocontainers/glimpse-bio:1.1.1--hce55b13_1' }"
+
+    input:
+    tuple val(meta), path(input)
+
+    output:
+    tuple val(meta), path("*.bcf"), emit: merged_bcf
+    path "versions.yml"           , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    GLIMPSE_ligate \\
+        $args \\
+        --input $input \\
+        --thread $task.cpus \\
+        --output ${prefix}_merged.bcf
+
+    cat <<-END_VERSIONS > versions.yml
+        "${task.process}":
+            glimpse: "\$(GLIMPSE_ligate --help | sed -nr '/Version/p' | grep -o -E '([0-9]+.){1,2}[0-9]')"
+    END_VERSIONS
+    """
+}
--- a/modules/nf-core/glimpse/ligate/main.nf
+++ b/modules/nf-core/glimpse/ligate/main.nf
@@ -11,7 +11,7 @@ process GLIMPSE_LIGATE {
     tuple val(meta), path(input)
 
     output:
-    tuple val(meta), path("*.bcf"), emit: merged_bcf
+    tuple val(meta), path("*.{vcf,bcf}"), emit: merged_variant
     path "versions.yml"           , emit: versions
 
     when:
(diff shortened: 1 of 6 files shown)


modules/nf-core/glimpse/ligate/main.nf
```diff
@@ -0,0 +1,35 @@
+process GLIMPSE_LIGATE {
+    tag "$meta.id"
+    label 'process_low'
+
+    conda "bioconda::glimpse-bio=1.1.1"
```
matthdsm, line 5: Latest version on github is 2.0.0. Could you look into updating bioconda and then your module (in the future)

modules/nf-core/glimpse/ligate/meta.yml
```diff
@@ -0,0 +1,45 @@
...
+        e.g. [ id:'test', single_end:false ]
+
+  - versions:
+      type: file
+      description: File containing software versions
+      pattern: "versions.yml"
+
+  - merged_bcf:
+      type: file
+      description: Output VCF/BCF file for the merged regions. Phased information (HS field) is updated accordingly for the full region.
```
matthdsm, line 41: Is the output vcf AND bcf? Or just bcf?
If both, can you add output channels for either option?

matthdsm:
xref https://github.com/bioconda/bioconda-recipes/pull/39013

modules/nf-core/glimpse/ligate/main.nf
LouisLeNezet: Hi!
I'm currently working on that [PR](https://github.com/bioconda/bioconda-recipes/pull/38994)
But there is some problem for the moment, and the two version are not exactly for the same use.

modules/nf-core/glimpse/ligate/meta.yml
LouisLeNezet: I think that both can be generated.
But I didn't tried yet.
I will test that !

modules/nf-core/glimpse/ligate/main.nf
matthdsm: Way ahead of me 👍🏻

modules/nf-core/glimpse/ligate/meta.yml
matthdsm: All possible outputs should be accounted for in the output channels, so we'd need to separate, optional outputs for vcf and bcf.
I suppose the output format can be set with some `ext.args` arguments?

LouisLeNezet: It doesn't seems to be the case. I think it is inferred from the --output extension.
There is more info [here](https://odelaneau.github.io/GLIMPSE/glimpse1/commands.html)

matthdsm: Is a vcf output at all possible? I think vcf is more "standard" than bcf, but it has to fit in your flow of course.

LouisLeNezet: @matthdsm I tried both, and it seems effectively to automatically detect the output format based on the extension given as `--output`.
How do I control for that ?
Do I force the format or do I add another argument that check the format the user want ?

matthdsm: _personally_ I'd default to vcf...
But since my opinion doesn't cover everyone's usecase, that doesn't matter. I think usually we check for the input format and output in the same format.

matthdsm: Then you can add 2 _optional_ output channels for both the vcf and bcf.

matthdsm: of course @nf-core/maintainers might have more opinions

nvnieuwk: You can use the `task.ext.suffix` to define the extension. Maybe set `vcf` as default on this :)

LouisLeNezet: Is it better to use 2 different output channels or just one that get both as in samtools convert ?

nvnieuwk: In this module it's one or the other, so I think you can keep it as one channel

SPPearce: I would agree to just have one channel which takes the output, whichever file type it is, as it only generates one single file.

modules/nf-core/glimpse/ligate/main.nf
```diff
@@ -0,0 +1,36 @@
...
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/glimpse-bio:1.1.1--hce55b13_1':
+        'quay.io/biocontainers/glimpse-bio:1.1.1--hce55b13_1' }"
+
+    input:
+    tuple val(meta), path(input)
+
+    output:
+    tuple val(meta), path("*.{vcf,bcf}"), emit: merged_variant
```
matthdsm, line 14: nitpicking, but is there a compressed option for vcf?

LouisLeNezet: That's a good question, you can compressed it with bgzip.
I didn't tried yet if it was possible directly in the software.

LouisLeNezet: So apparently it's possible you just need to specify the extension.

matthdsm (approved):
LGTM 🚀

LouisLeNezet:
Thanks @matthdsm

PR #2794, opened 2023-01-27
Update Glimpse phase for compressed output

LouisLeNezet:
## PR checklist

This PR allows the choice of file format for the output (compressed, binary or not) by the use of a suffix argument.


matthdsm (approved):
Compress all the things!

PR #1987, opened 2022-08-31
Improves kraken2 output glob

jfy133:
As per a request by @Midnighter 

- Previous classified_fastq output channel would include the unclassified and classifiedreads.txt files, this ensures they are not incorrectly picked up
- Also makes the FASTQ option  command injection variables clearer (distinguish from the atual pigz command)

## PR checklist

Closes #XXX 

- [x] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [x] Remove all TODO statements.
- [x] Emit the `versions.yml` file.
- [x] Follow the naming conventions.
- [x] Follow the parameters requirements.
- [x] Follow the input/output options guidelines.
- [x] Add a resource `label`
- [x] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - [ ] `PROFILE=docker pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [x] `PROFILE=singularity pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [ ] `PROFILE=conda pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`

PR #2796, opened 2023-01-27
Make untar work for non-gzipped tars

pinin4fjords:
## PR checklist

Currently the untar module assumes .tar.gz, so with a simple .tar we get:

```
gzip: unknown compression format
  tar: Child returned status 1
  tar: Error is not recoverable: exiting now
```
~~This allows detection of the 'gz' extension, and causes the 'z' flag to not be passed if it's absent.~~ We now use `-a` to handle exentions appropriately. I've also switched to the conventional `ext.prefix` usage per excellent review suggestion. We can still create the existing behaviour by setting prefix accordingly, as demonstrated in the tests.

If you, reviewer, would also care to approve https://github.com/nf-core/test-datasets/pull/738, then I could add a test for the non-gz situation as well :-)

- [ ] Add a resource `label`

### DIFF
--- a/modules/nf-core/untar/main.nf
+++ b/modules/nf-core/untar/main.nf
@@ -54,7 +54,8 @@ process UNTAR {
     stub:
     untar     = "${archive.simpleName}"
     """
-    touch $untar
+    mkdir $untar
+    touch ${untar}/file.txt
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
--- a/modules/nf-core/untar/main.nf
+++ b/modules/nf-core/untar/main.nf
@@ -11,8 +11,8 @@ process UNTAR {
     tuple val(meta), path(archive)
 
     output:
-    tuple val(meta), path("$untar"), emit: untar
-    path "versions.yml"            , emit: versions
+    tuple val(meta), path("$prefix"), emit: untar
+    path "versions.yml"             , emit: versions
 
     when:
     task.ext.when == null || task.ext.when
@@ -20,30 +20,28 @@ process UNTAR {
     script:
     def args  = task.ext.args ?: ''
     def args2 = task.ext.args2 ?: ''
-    untar     = "${archive.simpleName}"
+    prefix = task.ext.prefix ?: "${archive.simpleName}" 
 
     """
-    mkdir output
+    mkdir $prefix
 
     ## Ensures --strip-components only applied when top level of tar contents is a directory
-    ## If just files or multiple directories, place all in output
+    ## If just files or multiple directories, place all in prefix
     if [[ \$(tar -taf ${archive} | grep -o -P "^.*?\\/" | uniq | wc -l) -eq 1 ]]; then
         tar \\
-            -C output --strip-components 1 \\
+            -C $prefix --strip-components 1 \\
             -xavf \\
             $args \\
             $archive \\
             $args2
     else
         tar \\
-            -C output \\
+            -C $prefix \\
             -xavf \\
             $args \\
             $archive \\
             $args2
     fi
-
-    mv output ${untar}
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
@@ -52,10 +50,10 @@ process UNTAR {
     """
 
     stub:
-    untar     = "${archive.simpleName}"
+    prefix     = "${archive.simpleName}"
     """
-    mkdir $untar
-    touch ${untar}/file.txt
+    mkdir $prefix
+    touch ${prefix}/file.txt
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
--- a/modules/nf-core/untar/main.nf
+++ b/modules/nf-core/untar/main.nf
@@ -50,7 +50,7 @@ process UNTAR {
     """
 
     stub:
-    prefix = task.ext.prefix ?: "${meta.id}"
+    prefix    = task.ext.prefix ?: ( meta.id ? "${meta.id}" : archive.toString().replaceFirst(/\.[^\.]+(.gz)?$/, ""))
     """
     mkdir $prefix
     touch ${prefix}/file.txt
--- a/modules/nf-core/untar/main.nf
+++ b/modules/nf-core/untar/main.nf
@@ -50,7 +50,7 @@ process UNTAR {
     """
 
     stub:
-    prefix     = "${archive.baseName}"
+    prefix = task.ext.prefix ?: "${meta.id}"
     """
     mkdir $prefix
     touch ${prefix}/file.txt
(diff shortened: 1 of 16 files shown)


muffato:
What about using this option instead ?
```
       -a, --auto-compress
              Use archive suffix to determine the compression program.
```

pinin4fjords:
> What about using this option instead ? It would naturally bring support the bzip2, xz, etc
> 
> ```
>        -a, --auto-compress
>               Use archive suffix to determine the compression program.
> ```

@muffato haha- because I didn't know that existed. Now I do, thanks implemented accordingly

modules/nf-core/untar/main.nf
```diff
@@ -20,24 +20,24 @@ process UNTAR {
     script:
     def args  = task.ext.args ?: ''
     def args2 = task.ext.args2 ?: ''
-    untar     = archive.toString() - '.tar.gz'
+    untar     = "${archive.simpleName}"
```
adamrtalbot, line 23: How does this handle a filename with periods in it? `hello.world.tar.gz`

Then again, does it matter? Maybe it's worth the risk. Or unpack it to foldername `$prefix` to remove ambiguity.

modules/nf-core/untar/main.nf
```diff
@@ -52,7 +52,7 @@ process UNTAR {
     """
 
     stub:
-    untar     = archive.toString() - '.tar.gz'
+    untar     = "${archive.simpleName}"
     """
     touch $untar
```
adamrtalbot, line 57: Would this be more consistent with what the tool is doing?
```suggestion
    mkdir $untar
    touch ${untar}/file.txt
```

modules/nf-core/untar/main.nf
muffato: `simpleName` would return just `hello`. It strips out everything after the dot, cf https://www.nextflow.io/docs/latest/script.html?highlight=getname#check-file-attributes

This `untar` variable is in fact the name of the output directory, so I think it'd be fine setting it to the usual prefix as per the nf-core pipeline.
Otherwise, you could try something like this:
```suggestion
    untar     = archive.extension == "tar" ? "${archive.baseName}" : "${archive.baseName.baseName}"
```

pinin4fjords: Yeah, the issue is that there's no way of stripping only the extensions we want- since we could have .tgz, tar.gz etc. So I chose to be brutal. I guess it _could_ create issues if people were uncompressing multiple things like 'foo.one.tar.gz', 'foo.two.tar.gz' and ending up with multiple input folders named 'foo' in agglomerative processes. 

`$prefix` is a good call- I've done that (default is the simpleName).

pinin4fjords: Prefix is a good call. I've set the nextflow.config for the tests to override it to preserve existing behaviour (and any existing tools that rely on the directory name for some reason (e.g. gathering unzipped dirs in a process) can do the same.

pinin4fjords:
(note: ci failures are not related to the code changes here)

modules/nf-core/untar/main.nf
```diff
@@ -11,50 +11,49 @@ process UNTAR {
...
+    path "versions.yml"             , emit: versions
 
     when:
     task.ext.when == null || task.ext.when
 
     script:
     def args  = task.ext.args ?: ''
     def args2 = task.ext.args2 ?: ''
-    untar     = archive.toString() - '.tar.gz'
+    prefix = task.ext.prefix ?: "${meta.id}"
```
muffato, line 23: And nf-core said: Equal signs shall be aligned
```suggestion
    prefix    = task.ext.prefix ?: "${meta.id}"
```

muffato (commented):
I'm not so sure the CI failures are unrelated to the changes here:
kraken2/kraken2 is indeed marked as depending on untar https://github.com/nf-core/modules/blob/master/tests/config/pytest_modules.yml#L1676-L1679
because the test-case uses untar https://github.com/nf-core/modules/blob/master/tests/modules/nf-core/kraken2/kraken2/main.nf
All checks passed in the latest PR about kraken2/kraken2 I could find: https://github.com/nf-core/modules/pull/1987/checks

modules/nf-core/untar/main.nf
pinin4fjords: Ahh, that makes sense. I'm still learning, thank you. I'll fix that then.

modules/nf-core/untar/main.nf
pinin4fjords: Here's my revised proposal:

- Prefix if specified (which I've done in the tests to match previous behaviour)
- else if meta ID is set use that (per normal prefix behaviour)
- else if prefix is unset, generate a directory based on the archive, by removing extensions with or without an additional .gz extension (via regex).

This will preserve existing behaviour (foo.tar.gz -> foo) where there is no prefix or meta id (which seems to be the bulk of existing usage) and allow use of the conventional prefix.

modules/nf-core/untar/main.nf
pinin4fjords: There was just an error with a process select in the kraken subworkflow, which was causing a prefix declaration to apply to the tar process. This hadn't mattered previously (since tar was not using prefix), but does now. Fixed by increasing the selectivity.

pinin4fjords:
(fairly sure the current CI issues are conda-related things due to parallel environment creation processes in the subworkflow, and _really_ not anything to do with these changes now ;-) )

modules/nf-core/untar/main.nf
```diff
@@ -11,50 +11,49 @@ process UNTAR {
...
+    path "versions.yml"             , emit: versions
 
     when:
     task.ext.when == null || task.ext.when
 
     script:
     def args  = task.ext.args ?: ''
     def args2 = task.ext.args2 ?: ''
-    untar     = archive.toString() - '.tar.gz'
+    prefix    = task.ext.prefix ?: ( meta.id ? "${meta.id}" : archive.toString().replaceFirst(/\.[^\.]+(.gz)?$/, ""))
```
muffato, line 23: This should support all possible extensions (`tgz`, `tbz2`, `tar.gz`, etc) by assuming that there's always at least one extension to remove, and sometimes there's still `.tar` at the end.
```suggestion
    prefix    = task.ext.prefix ?: ( meta.id ? "${meta.id}" : archive.baseName.toString().replaceFirst(/\.tar$/, ""))
```

muffato (approved):
All tests pass locally

adamrtalbot (approved):

pinin4fjords:
> Is it because we've switched to using mamba as the conda engine recently?

@adamrtalbot I suspect not. It's just what happens when you do parallel environment creation for the same thing (I used to get this sort of thing (though maybe not the exact errors) previously when I ran workflows of workflows). There will be somewhere in the subworkflow tests where the untar environment is being created multiple times. 

But I am a little worried that's a problem inherent in subworkflows we need to address- I raised a discussion point in [#modules](https://nfcore.slack.com/archives/CJRH30T6V/p1674946758099229).

PR #2813, opened 2023-01-31
new module vg/construct

### LINKED ISSUE #2799: new module: vg/construct

nvnieuwk:
## PR checklist

Closes #2799 

- [ ] Add a resource `label`


modules/nf-core/vg/construct/main.nf
```diff
@@ -0,0 +1,68 @@
...
+    if(vcfs && (msa && fasta)) {
+        error("Please use either VCF files and reference FASTAs or an MSA file as input.")
+    }
+
+    vcf_input = vcfs ? vcfs.collect { "--vcf ${it}" }.join(" ") : ""
+    reference = fasta ? "--reference ${fasta}" : ""
+
+    msa_input = msa ? "--msa ${msa}" : ""
+
+    insertions = insertions_fasta ? "--insertions ${insertions_fasta}" : ""
```
pinin4fjords, line 35: Okay, I'm not familiar with this too, the docs don't match your usage and I'm having trouble accessing vg on my Mac (even docker) to view the inline help, so just to check, is this the situation?

Either:

- 'vcf mode' = vcfs, tbis, reference fasta (+fai?), plus optional insertion fa?
- 'msa mode' = msa only? with insertion fa?

So there's no 'core' set of inputs an basically everything is optional? If I've got the right end of that it feels like the msa should be a separate channel.

modules/nf-core/vg/construct/main.nf
```diff
@@ -0,0 +1,68 @@
...
+    label 'process_medium'
+
+    conda "bioconda::vg=1.45.0"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/vg:1.45.0--h9ee0642_0':
+        'quay.io/biocontainers/vg:1.45.0--h9ee0642_0' }"
+
+    input:
+    tuple val(meta), path(vcfs), path(tbis), path(msa), path(insertions_fasta)
+    tuple val(meta2), path(fasta)
```
pinin4fjords, line 12: Could we maybe just rename `fasta` to `reference`, since that's what the parameter is called? Would make the module easier to parse :-)

Does the fasta index need to be a separate channel? Seems like it relates closely to the fasta.

modules/nf-core/vg/construct/meta.yml
```diff
@@ -0,0 +1,66 @@
...
+      description: The index files for the VCF files
+      pattern: "*.tbi"
+  - msa:
+      type: file
+      description: A multiple sequence alignment file (cannot be used in combination with `vcfs` and `fasta`)
+      pattern: "*.{fa,fasta,fna,clustal}"
+  - insertions_fasta:
+      type: file
+      description: A FASTA file containing insertion sequences (referred to in the VCF file(s))
+      pattern: "*.vcf.gz"
```
pinin4fjords, line 40:
```suggestion
      pattern: "*.{fa,fasta,fna}"
```

pinin4fjords (commented):
I might make further suggestions about simplifying the args, but right now I don't understand how they're supposed to work :-)

modules/nf-core/vg/construct/main.nf
nvnieuwk: That's correct!! And yeah I've thought about splitting them up but then I need to create two different inputs for `insertions_fasta`. But we can discuss this, nothing is set in stone yet :)

modules/nf-core/vg/construct/main.nf
nvnieuwk: I'm following the way it's done in other modules. I think the fasta and its index should go together too, but I don't really want to do it different than in other modules for ease of development.

modules/nf-core/vg/construct/meta.yml
nvnieuwk: Good spot! Thanks!

modules/nf-core/vg/construct/main.nf
nvnieuwk: strange though that it isn't working on Mac...

pinin4fjords: Sorry, to double check- the insertions fasta applies to the msa mode as well? Despite what I might assume from:

```
 -I, --insertions FILE  a FASTA file containing insertion sequences 
                           (referred to in VCF) to add to graph.
```
?

modules/nf-core/vg/construct/main.nf
pinin4fjords: Fair enough!

modules/nf-core/vg/construct/main.nf
nvnieuwk: I haven't actually tested this, I'll do that right now!

nvnieuwk: It does work! I've added it to the tests.

modules/nf-core/vg/construct/main.nf
pinin4fjords, line 12:
```suggestion
    tuple val(meta), path(vcfs), path(tbis)     // Basic inputs for 'vcf mode'
    tuple val(meta2), path(fasta)                    // .. other vcf mode required input
    tuple val(meta3), path(fasta_fai)             //
    tuple val(meta4), path(msa)                     // Basic inputs for 'msa mode' 
    tuple val(meta5), path(insertions_fasta) // Optional inputs for both ethods
```
(sorry, couldn't figure out how to include the fai below the other comment!)

I feel like the inputs for different mutually exclusive modes should be separated, otherwise you'll always be forced to edit channels with `.map{}` etc to insert empty elements relating to the mode you don't want to use. This way you just have to add empty channels (`[]`) which feels simpler to me. 

But it's very subjective of me and I assume you'll be the main user, so I won't fight you on it!

pinin4fjords: I actually wonder if two different modules might be a simpler way of managing this complexity.

pinin4fjords (approved):
I think this is messy and difficult to understand as a consequence of how the tool works, but nothing I need to block you for, and I'll leave you to wrangle.

modules/nf-core/vg/construct/main.nf
nvnieuwk: Although maybe I could change the `vcfs` to `input` and check what file has been given to see what mode it should be run in

modules/nf-core/vg/construct/main.nf
```diff
@@ -0,0 +1,63 @@
...
+        vg: \$(echo \$(vg 2>&1 | head -n 1 | sed 's/vg: variation graph tool, version v//;s/ ".*"//' ))
+    END_VERSIONS
+    """
+
+    stub:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+
+    if(vcfs && msa) {
+        error("Please use either VCF files or an MSA file as input, not both.")
+    }
```
nvnieuwk, line 53:
```suggestion
```

modules/nf-core/vg/construct/meta.yml
```diff
@@ -0,0 +1,63 @@
+name: "vg_construct"
+description: write your description here
```
nvnieuwk, line 2:
```suggestion
description: Constructs a graph from a reference and variant calls or a multiple sequence alignment file
```

modules/nf-core/vg/construct/main.nf
pinin4fjords: I'm not sure that's much of an improvement, now there's `tbis` which don't apply to msa and would need an empty element in the channel.

I'm really not sure what the best way is, maybe a maintainer with more experience has a more cunning plan.

nvnieuwk: Yes, there's not really another option I think in this case. I'll ask in Slack for a second opinion, thanks for the review!!

nvnieuwk:
Thanks @maxulysse :partying_face:

modules/nf-core/vg/construct/main.nf
```diff
@@ -0,0 +1,60 @@
...
+    path "versions.yml"           , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+
+    def mode =  input instanceof ArrayList || input.toString().endsWith(".vcf.gz") ? 'vcf' : 'msa'
```
adamrtalbot, line 26: Does an uncompressed VCF work?

modules/nf-core/vg/construct/meta.yml
```diff
@@ -0,0 +1,63 @@
...
+      type: file(s)
+      description: The index files for the VCF files
+      pattern: "*.tbi"
+  - insertions_fasta:
+      type: file
+      description: A FASTA file containing insertion sequences (referred to in the VCF file(s))
+      pattern: "*.{fa,fasta,fna}"
+  - fasta:
+      type: file
+      description: The reference FASTA file (cannot be used in combination with `msa`, but is dependent on `vcfs`)
```
adamrtalbot, line 40:
```suggestion
      description: The reference FASTA file (cannot be used in combination with `msa`, but is required when using `vcfs`)
```

adamrtalbot (approved):
s'all good. I left a suggestion on the documentation and I've got one question about the input VCF but it isn't necessary for a merge.

modules/nf-core/vg/construct/main.nf
nvnieuwk: I haven't tested it since we don't really support unzipped VCFs, I can check real quick

nvnieuwk: It doesn't seem supported by the tool too

PR #2826, opened 2023-02-01
Update hifiasm

### LINKED ISSUE #2822: update module: hifiasm (to support hi-c phasing)
Hifiasm now has a mode whereby it can phase the contigs using Hi-C read.
Make the module support this mode

scorreard:
## PR checklist

Closes #2822 https://github.com/nf-core/modules/issues/2822

- [ ] Add a resource `label`

### DIFF
--- a/modules/nf-core/hifiasm/main.nf
+++ b/modules/nf-core/hifiasm/main.nf
@@ -2,16 +2,19 @@ process HIFIASM {
     tag "$meta.id"
     label 'process_high'
 
-    conda "bioconda::hifiasm=0.15.4"
+    conda "bioconda::hifiasm=0.18.5"
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
-        'https://depot.galaxyproject.org/singularity/hifiasm:0.15.4--h2e03b76_0' :
-        'quay.io/biocontainers/hifiasm:0.15.4--h2e03b76_0' }"
+        'https://depot.galaxyproject.org/singularity/hifiasm:0.18.5--h5b5514e_0' :
+        'quay.io/biocontainers/hifiasm:0.18.5--h5b5514e_0' }"
 
     input:
     tuple val(meta), path(reads)
     path  paternal_kmer_dump
     path  maternal_kmer_dump
     val   use_parental_kmers
+    path  hic_read1
+    path  hic_read2
+    val   use_hic
 
     output:
     tuple val(meta), path("*.r_utg.gfa")       , emit: raw_unitigs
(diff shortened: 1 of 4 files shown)


modules/nf-core/hifiasm/main.nf
```diff
@@ -41,6 +44,21 @@ process HIFIASM {
             -2 $maternal_kmer_dump \\
             $reads
 
+        cat <<-END_VERSIONS > versions.yml
+        "${task.process}":
+            hifiasm: \$(hifiasm --version 2>&1)
+        END_VERSIONS
+        """
+    } else if (use_hic) {
```
muffato, line 52: In principle, can the HiC mode be used at the same time as the parental-kmer mode ?

modules/nf-core/hifiasm/main.nf
```diff
@@ -2,16 +2,19 @@ process HIFIASM {
...
+        'quay.io/biocontainers/hifiasm:0.18.5--h5b5514e_0' }"
 
     input:
     tuple val(meta), path(reads)
     path  paternal_kmer_dump
     path  maternal_kmer_dump
     val   use_parental_kmers
+    path  hic_read1
+    path  hic_read2
+    val   use_hic
```
muffato, line 17: I know this is the same style as the parental-kmer mode above, but the nf-core guidelines advise that only files are passed through the input channels.
Could you remove `use_hic` here and infer it in the script section below instead ? At the same time you can check that you're not in a situation where only one of `hic_read1` and `hic_read2` is defined.

muffato (commented):
Thank you @scorreard ! Here are some suggestions

modules/nf-core/hifiasm/main.nf
scorreard: I don't think it can. I checked their paper and they don't mention the combination of both (but they don't say it's impossible neither). I think it would fail because it would try to do the phasing with two different technic. 
Maybe if it's something that can be done we can add the combined option later.

scorreard:
@muffato what do you think of having it this way? 
I can also update it for the trio if you think it's better.

modules/nf-core/hifiasm/main.nf
muffato: OK ! Thanks for checking :+1:

muffato:
That works, thanks ! Can you add another `else if` to catch cases where just one of `hic_read1`/`hic_read2` is defined and raise an error ? That'd be better for the module user rather than silently falling back to no phasing.
Similarly, the module should raise an error if it's being used with both parental kmers and Hi-C reads.

If you can remove the `use_parental_kmers` boolean to test the actual inputs `paternal_kmer_dump` and `maternal_kmer_dump` that'd be wonderful, but it's outside of the scope of your PR, which is adding the Hi-C mode. So, no obligation there.

scorreard:
@muffato, sorry for dropping this, I was at a conference all last week.
Let me know what you think of the latest updates :)

muffato (approved):
It looks nice and clean ! Great !

PR #2828, opened 2023-02-02
Ciriquant

BarryDigby:
I added CIRIquant to bioconda late last year, so it makes sense to follow up and add the nf-core module.

### DIFF
--- a/modules/nf-core/ciriquant/main.nf
+++ b/modules/nf-core/ciriquant/main.nf
@@ -1,0 +1,57 @@
+process CIRIQUANT {
+    tag "$meta.id"
+    label 'process_high'
+
+    conda "bioconda::ciriquant=1.1.2"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/ciriquant:1.1.2--pyhdfd78af_2' :
+        'quay.io/biocontainers/ciriquant:1.1.2--pyhdfd78af_2' }"
+
+    input:
+    tuple val(meta), path(reads)
+    path gtf
+    path fasta
+    path bwa
+    path hisat2
+
+    output:
+    tuple val(meta), path("${prefix}/${prefix}.gtf"), emit: gtf
+    path("${prefix}")                               , optional: true, emit: intermediates_directory
+    path "versions.yml"                             , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "${meta.id}"
+    bwa_prefix = fasta.toString() == 'genome.fa' ? fasta.toString() : fasta.toString() - ~/.(fa|fasta)$/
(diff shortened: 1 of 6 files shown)


SPPearce:
Are you still interested in getting this merged?

nictru:
@BarryDigby disappeared from the development of #circrna more than a year ago, since then I took over the main development. Did not notice this PR until now, but it looks quite outdated. However, adding CIRIquant to nf-core/modules is still a good idea. Will update the PR soon

SPPearce:
Ok, thanks @nictru .
I'm happy to help if you need some.
I suspect it'll need the environment.yml. I can sort that and swap to nf-test to get something towards working.

nictru:
That would be very great. I recently implemented changes to the local version in #circrna, which should be integrated here. 
After this, I will try to get the tests set up correctly, but as I don't have too much experience with this, I might need some help.

SPPearce:
Ok, I'll try and swap to nf-test later today.

nictru:
Very great, thank you!

SPPearce:
@nictru , I swapped to nf-test.
The log file and the bam files are complaining that they are not stable.
The log isn't surprising, as it presumably has a timestamp, so that will need excluding. The assertion will need changing from the default md5 of all files, see https://nf-co.re/docs/contributing/nf-test/assertions for examples.
The bam files are more surprising, is something non-deterministic in the process?

nictru:
Great, thanks! I have meetings this afternoon, but I will look at this in Detail ASAP

nictru:
The problematic BAM files are not used in #circrna anyway, so I removed the whole output for now. The stuff we actually use works and if we ever need this, we can still investigate.

SPPearce:
Odd, the versions file is different on conda, can you check what is different? Gitpod is helpful for this as it has all the container engineers installed

nictru:
For the command `echo $(bwa 2>&1) | sed 's/^.*Version: //; s/Contact:.*$//'` I receive different results:
- Docker/Biocontainers: `0.7.17-r1188`
- Conda: `0.7.18-r1243-dirty`

Which suggests to me that the Bioconda package was at some point updated without updating the biocontainer. Another explanation would be that the bwa version is not defined exact enough in the recipe, so that conda downloads what it wants.
I tested this using sequera containers - building a new image there gives me the same result as the conda version (`0.7.18-r1243-dirty`).

I see two solutions here:
1. Use the seqera image instead of biocontainers
2. Update the bioconda recipe

@SPPearce, do you have a preference?

nictru:
Ok I just checked, there are a lot of dependency versions not specified in [the recipe](https://github.com/bioconda/bioconda-recipes/blob/ec8bc357bd639682d6bc28afb241afbd8c066f1b/recipes/ciriquant/meta.yaml#L30)

SPPearce:
Ok, do you want to make a PR to pin them then?
Otherwise we could swap to seqera containers, but I don't know what the long-term plan is for how to use them.

famosab:
@nictru Do you want to finalize this?

SPPearce:
Seeing as nothing is happening with this for two years, lets close it. Feel free to reopen if you want to finish it off.

PR #2842, opened 2023-02-06
add --target-file argument

wikiselev:
Adding the `--target-file` argument requires an extra input to this module - a `bed` file. This commit adds the `bed` file input to the module.


nvnieuwk:
Can you add some tests? Also will this bed input be the same for all files in the same channel?

famosab:
I think this is very out of date and would close the issue. Feel free to open a new one if you want to change something on the samtools module :)

PR #2843, opened 2023-02-06
make fasta optional; remove fai

wikiselev:
According to [documentation](https://gatk.broadinstitute.org/hc/en-us/articles/360037052812-MarkDuplicates-Picard-) `--REFERENCE_SEQUENCE` argument is optional for `MarkDuplicates` and `fai` input is not used used in any argument at all.

This commit makes `fasta` input (same as `--REFERENCE_SEQUENCE`) optional and removes unnecessary `fai` input.

### DIFF
--- a/modules/nf-core/picard/markduplicates/main.nf
+++ b/modules/nf-core/picard/markduplicates/main.nf
@@ -10,7 +10,6 @@ process PICARD_MARKDUPLICATES {
     input:
     tuple val(meta), path(bam)
     path fasta
-    path fai
 
     output:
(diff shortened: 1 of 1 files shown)


projectoriented:
Hi @wikiselev, if I recall correctly- a lot of these tools may not explicitly require the need of `.fai` but it's most likely assumed it will be accompanied with the `.fasta` file for fast query/access. I recommend to keep the `.fai` as input to make it available for `MARK_DUPLICATES` in case it requires that file.

PR #2870, opened 2023-02-10
Glimpse_ligate input management

LouisLeNezet:
## PR checklist

The input of glimpse ligate is improved and take also in account the modification made on glimpse_phase previously.
Glimpse_ligate take now a path channel as input and not a file containing the path to the files.

### DIFF
--- a/modules/nf-core/glimpse/ligate/main.nf
+++ b/modules/nf-core/glimpse/ligate/main.nf
@@ -28,7 +28,7 @@ process GLIMPSE_LIGATE {
         $args \\
         --input all_files.txt \\
         --thread $task.cpus \\
-        --output ${prefix}.merged.${suffix}
+        --output ${prefix}.${suffix}
 
     cat <<-END_VERSIONS > versions.yml
         "${task.process}":
--- a/modules/nf-core/glimpse/ligate/main.nf
+++ b/modules/nf-core/glimpse/ligate/main.nf
@@ -8,7 +8,7 @@ process GLIMPSE_LIGATE {
         'quay.io/biocontainers/glimpse-bio:1.1.1--hce55b13_1' }"
 
     input:
-    tuple val(meta), path(input)
+    tuple val(meta), path(input_list)
 
     output:
     tuple val(meta), path("*.{vcf,bcf,vcf.gz,bcf.gz}"), emit: merged_variants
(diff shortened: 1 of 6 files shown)


modules/nf-core/glimpse/ligate/main.nf
```diff
@@ -22,11 +22,13 @@ process GLIMPSE_LIGATE {
...
     """
+    printf "%s\\n" $input_list | tr -d '[],' > all_files.txt
+
     GLIMPSE_ligate \\
         $args \\
-        --input $input \\
+        --input all_files.txt \\
         --thread $task.cpus \\
-        --output ${prefix}_merged.${suffix}
+        --output ${prefix}.merged.${suffix}
```
nvnieuwk, line 31:
```suggestion
        --output ${prefix}.${suffix}
```

It's preffered to not add any hardcoded extensions. See [here](https://nf-co.re/docs/contributing/modules#naming-conventions) part 6 :)

modules/nf-core/glimpse/ligate/main.nf
```diff
@@ -8,7 +8,7 @@ process GLIMPSE_LIGATE {
         'quay.io/biocontainers/glimpse-bio:1.1.1--hce55b13_1' }"
 
     input:
-    tuple val(meta), path(input)
+    tuple val(meta), path(input_list)
```
nvnieuwk, line 11: It would be better to split the files and value inputs in two different channels. (so one channel called `inputs` for example that contains all files and another called `regions` for example). That way all files get staged correctly while the values remains values.

nvnieuwk (changes requested):

modules/nf-core/glimpse/ligate/main.nf
LouisLeNezet: Okay I understand.

modules/nf-core/glimpse/ligate/main.nf
LouisLeNezet: I don't quite understand.
In this module I do not have any regions informations.
I just have multiple files that I need to merge in one.
Wouldn't it concern more the phasing modules ?

nvnieuwk: Oh yeah my bad it was `phase` that needed the regions, this is fine then!

nvnieuwk (approved):

nvnieuwk:
You should just recreate the `test.yml` now :)

PR #2889, opened 2023-02-17
copy fix for gridss

nvnieuwk:
Soft link copy of files from the BWA index (would give an error if the index was on another file system)


jfy133 (approved):

nvnieuwk:
Forgot to push the second commit, is that one also okay @jfy133 ?

jfy133 (approved):

PR #2875, opened 2023-02-15
New module Ilastik/pixelclassification

FloWuenne:
## PR checklist

Closes #2864 

- [x] This comment contains a description of changes (with reason).
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [x] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [x] Remove all TODO statements.
- [x] Emit the `versions.yml` file.
- [x] Follow the naming conventions.
- [x] Follow the parameters requirements.
- [x] Follow the input/output options guidelines.
- [x] Add a resource `label`
- [ ] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - [x] `PROFILE=docker pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [ ] `PROFILE=singularity pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [ ] `PROFILE=conda pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  
 
# Description of changes
  
This PR adds ilastik pixel classification as a DSL2 module. The module takes as input an h5 file as specified by ilastik and outputs one of 4 different output formats. Tests are included for all output formats, all non-file parameters are passed via ext.args. 
  
 ## Things to do before merging
- Add test data to module branch of https://github.com/nf-core/test-datasets
- Determine whether ilastik-forge conda channel can be used for conda
- Determine whether the biocontainer is built from most up-to-date version automatically:
https://hub.docker.com/r/biocontainers/ilastik/tags

PR #2894, opened 2023-02-19
New module Ilastik/multicut

### LINKED ISSUE #2865: new module: ilastik/multicut

FloWuenne:
## PR checklist

Closes #2865  
Uses the same tool (ilastik) as in #2875 

- [x] Add a resource `label`
  
 
# Description of changes
  
This PR adds ilastik Multicut as a DSL2 module. The module takes as input an image in .h5 format, a multicut ilastik project and a probability map from a previously trained ilastik pixel classification. It outputs a multicut segmentation mask in .tiff format.
  
 ## Things to do before merging
- Add test data to module branch of https://github.com/nf-core/test-datasets
- Determine whether ilastik-forge conda channel can be used for conda


modules/nf-core/ilastik/multicut/main.nf
```diff
@@ -0,0 +1,38 @@
+process ILASTIK_MULTICUT {
+    tag "$meta.id"
+    label 'process_low'
+
+    container "labsyspharm/mcmicro-ilastik:1.6.0"
+
+    input:
+    tuple val(meta), path(h5)
+    path ilp
+    path probs
```
nvnieuwk, line 10: Can you also add `meta2` and `meta3` to these inputs?

modules/nf-core/ilastik/multicut/main.nf
```diff
@@ -0,0 +1,38 @@
...
+
+    output:
+    tuple val(meta), path("*.tiff") , emit: out_tiff
+    path "versions.yml"           , emit: versions
```
nvnieuwk, line 14:
```suggestion
    tuple val(meta), path("*.tiff") , emit: out_tiff
    path "versions.yml"             , emit: versions
```

Alignments :p

modules/nf-core/ilastik/multicut/main.nf
```diff
@@ -0,0 +1,38 @@
...
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+
+    """
+    /ilastik-release/run_ilastik.sh \\
+        --headless \\
+        --readonly 1 \\
```
nvnieuwk, line 26: Are these mandatory arguments? And is there no way this module can be run without these arguments? Otherwise I would suggest removing them here and using `ext.args` to add these.

modules/nf-core/ilastik/multicut/main.nf
```diff
@@ -0,0 +1,38 @@
...
+        --project=$ilp \\
+        --raw_data=$h5 \\
+        --probabilities=$probs \\
+        --export_source="Multicut Segmentation" \\
```
nvnieuwk, line 30: Same here with `--export_source`

modules/nf-core/ilastik/multicut/main.nf
```diff
@@ -0,0 +1,38 @@
...
+        $args
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        ilastik: \$(/ilastik-release/run_ilastik.sh --headless --version)
+    END_VERSIONS
+    """
+}
```
nvnieuwk, line 38: Can you also add the stub?

tests/modules/nf-core/ilastik/multicut/main.nf
```diff
@@ -0,0 +1,17 @@
...
+include { ILASTIK_MULTICUT } from '../../../../../modules/nf-core/ilastik/multicut/main.nf'
+
+workflow test_ilastik_multicut {
+
+    input = [
+        [ id:'probabilities' ], // meta map
+        file("/Users/florian_wuennemann/1_Projects/nf_core/test_data/wga.h5")
+    ]
+    ilp = [file("/Users/florian_wuennemann/1_Projects/nf_core/test_data/multicut_140rc8_plant.ilp")]
+    probs = [file("/Users/florian_wuennemann/1_Projects/nf_core/test_data/wga-data_Probabilities.h5")]
```
nvnieuwk, line 14: Can you use online available data here? Preferably data that is in the `nf-core/test-datasets` repository. Otherwise nobody else can run these tests :)

tests/modules/nf-core/ilastik/multicut/nextflow.config
```diff
@@ -0,0 +1,18 @@
+process {
+
+    publishDir = { "${params.outdir}/${task.process.tokenize(':')[-1].tokenize('_')[0].toLowerCase()}" }
+
+    withName: "ILASTIK_MULTICUT" {
+        ext.args = '--output_filename_format=multicut_segmentation.tiff'
```
nvnieuwk, line 6: This should be in the main script (since this handles a file, all arguments in `ext.args` need to be optional, non-file arguments)

modules/nf-core/ilastik/multicut/main.nf
```diff
@@ -0,0 +1,38 @@
+process ILASTIK_MULTICUT {
+    tag "$meta.id"
+    label 'process_low'
+
+    container "labsyspharm/mcmicro-ilastik:1.6.0"
```
nvnieuwk, line 5: Can this be added to conda?

nvnieuwk (changes requested):

tests/modules/nf-core/ilastik/multicut/main.nf
FloWuenne: I just submitted a PR to add these data to test-datasets. Sorry, I made these PRs before submitting the test-data PR!

https://github.com/nf-core/test-datasets/pull/771

Will fix the file paths as soon as this is merged!

modules/nf-core/ilastik/multicut/main.nf
FloWuenne: Yes, these are in fact mandatory arguments. `--headless` is required to execute ilastik in command line mode and `--readonly 1` is basically to ensure, that the same model could be used by multiple nextflow processes.

Can these stay as is if mandatory or do I still need to move them somewhere else?

modules/nf-core/ilastik/multicut/main.nf
FloWuenne: So for Multicut, this is the only `--export_source` definition allowed and thus is the mandatory default argument, which still has to be provided as far as I understand the documentation:
https://www.ilastik.org/documentation/basics/headless

image

modules/nf-core/ilastik/multicut/main.nf
nvnieuwk: They can stay there if they will always be like this :)

modules/nf-core/ilastik/multicut/main.nf
nvnieuwk: That's strange they implemented it like this... But yeah it's fine if the arg is mandatory

modules/nf-core/ilastik/multicut/main.nf
nvnieuwk: You can see an example here: https://github.com/nf-core/modules/blob/9c8154a8dd4989ddcbd540f78ffcd98c04aec767/modules/nf-core/vcf2db/main.nf#L38-L49

I agree this could be better documented, but it's basically adding a little script that should be run when `-stub` is used in the nextflow command. It should create empty files that correspond what the module would output if it were fully run.

modules/nf-core/ilastik/multicut/main.nf
FloWuenne: So it's a bit tricky with ilastik. It is already on conda, but on it's own channel called [ilastik-forge](https://anaconda.org/ilastik-forge). There is also a biocontainer, but I don't see from where it is built : https://hub.docker.com/r/biocontainers/ilastik/tags. Additionally, the biocontainer is not the most recent release candidate. This is problematic, since MacOS users can't use the latest stable release to train models but can use the release candidate. That's why for now I put our mcmicro container, which we can update and uses the latest release.

image

nvnieuwk: Wow I've never seen a tool have it's own specific conda channel. Is it impossible to use this channel in Nextflow?

tests/modules/nf-core/ilastik/multicut/nextflow.config
FloWuenne: Actually I believe its optional. You can leave this argument out and ilastik will still write and output .tiff mask, I am not quite sure how it determines the filename if left empty though...

nvnieuwk: You can add this to the main script as an argument (`--output_filename_format=${prefix}.tiff`). This way the output file name can be adjusted according to the input by the user

modules/nf-core/ilastik/multicut/main.nf
FloWuenne: I actually have no idea. @drpatelh is a custom conda channel like ilastik-forge useable in the conda definition of modules?

modules/nf-core/ilastik/multicut/main.nf
FloWuenne: I added a stub to the file, is this the way it should look?

https://github.com/FloWuenne/modules/blob/3e58ed1c129c372e3e88c79046affa31e132b46b/modules/nf-core/ilastik/multicut/main.nf#L40-L50

tests/modules/nf-core/ilastik/multicut/main.nf
FloWuenne: Test-data is now available and paths have been fixed!

FloWuenne:
Any idea why the singularity build still fails? Is it due to this part of the nextflow.config in test?
```
profiles {
    standard {
        docker {
            enabled = true
            userEmulation = false
        }
    }
}
```

I had to add `userEmulation = false` to the config, because otherwise the workflow gives errors about puid tracking within the container.

FloWuenne:
> Any idea why the singularity build still fails? Is it due to this part of the nextflow.config in test?
> 
> ```
> profiles {
>     standard {
>         docker {
>             enabled = true
>             userEmulation = false
>         }
>     }
> }
> ```
> 
> I had to add `userEmulation = false` to the config, because otherwise the workflow gives errors about puid tracking within the container.

It seems that having `enabled = true` for docker as the standard profile was the issue. I removed it and now the singularity container also builds correctly!

nvnieuwk (commented):
Can you also add a check for pipelines running with conda like this?
```
    // Exit if running this module with -profile conda / -profile mamba
        exit 1, "ILASTIK_MULTICUT module does not support Conda. Please use Docker / Singularity / Podman instead."
    }
```

Otherwise LGTM

nvnieuwk (approved):

modules/nf-core/ilastik/multicut/main.nf
FloWuenne: I am gonna check with the ilastik developers, whether they plan to have a maintained container on dockerhub that tracks the newest version. In the meantime, I think having the ilastik modules run off the latest mcmicro-ilastik container is the best way to go. That way we can update the container whenever we need to.

FloWuenne:
Before merging, the ilastik authors have just released a "stable" release version 1.4.0. We might wanna update the mcmicro ilastik container and this here to that version. I will look into that now!

modules/nf-core/ilastik/multicut/main.nf
nvnieuwk: Allright, can you also ask whether or not they would want to add their tool to `biocontainers` instead of that separate channel? I think it would be best for everyone (although I don't know what their reason is so we'll have to see)

FloWuenne:
Should be good to merge now!

nvnieuwk (approved):

maxulysse:
I'm hoping I fixed the merge conflicts properly

maxulysse:

PR #2910, opened 2023-02-22
Fix SMNcaller stub

peterpru:
The SMNcopynumbercaller stub gives an error when running stub tests when the prefix is not defined. Here the output files have received a placeholder name instead, and successfully complete all tests.

Furthermore, Version has been defined in the module in quotes, as the version information is not provided by tool on CLI. A warning is present in the main.nf file to update version info when bumping program versions.

## PR checklist

- [x] Add a resource `label`

### DIFF
--- a/modules/nf-core/smncopynumbercaller/main.nf
+++ b/modules/nf-core/smncopynumbercaller/main.nf
@@ -22,7 +22,7 @@ process SMNCOPYNUMBERCALLER {
     manifest_text = bam.join("\n")
     def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
-    def VERSION = "1.1.2" // WARN: Version information not provided by tool on CLI. Please update version string below when bumping container versions.
+    // WARN: Version information not provided by tool on CLI. Please update version string below when bumping container versions.
     """
     echo "$manifest_text" >manifest.txt
     smn_caller.py \\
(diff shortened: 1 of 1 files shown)


modules/nf-core/smncopynumbercaller/main.nf
```diff
@@ -34,19 +34,19 @@ process SMNCOPYNUMBERCALLER {
...
     END_VERSIONS
     """
 
     stub:
     """
     mkdir out
-    touch out/${prefix}.tsv
-    touch out/${prefix}.json
+    touch out/smnstub.tsv
+    touch out/smnstub.json
```
nvnieuwk, line 45: The whole point of stub is to run the modules as if they were actually run. So the output files should have the same name as they would have if the module was run

modules/nf-core/smncopynumbercaller/main.nf
```diff
@@ -34,19 +34,19 @@ process SMNCOPYNUMBERCALLER {
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
-        SMNCopyNumberCaller: $VERSION
+        SMNCopyNumberCaller: "1.1.2"
```
nvnieuwk, line 37: The use of `$VERSION` is how all modules do it and the guidelines show that it should be like this: https://nf-co.re/docs/contributing/modules#general paragraph 4, the second part

nvnieuwk (commented):
I don't really get why you want to change it like this. The guidelines state it should be the way it was.

peterpru:
Thank you @nvnieuwk for the review, I realised the errors I get with the module is the way we have structured our test in our pipeline. I will close this PR.

PR #2913, opened 2023-02-22
add schema to validate meta.ymls

mashehu:
In order to avoid this I am proposing the following JSON-schema. We will use this then during `nf-core modules lint` to check the meta.ymls (and in the future also for subworkflows). It also is integrated in the pre-commit check. If you are using the yaml extension for VS code you get also hints about expected entries when working on this file (by adding the following to your settings.json
```
"yaml.schemas": {
"yaml-schema.json": "modules/**/meta.yml"
},
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,5 +7,5 @@ repos:
     rev: 0.21.0
     hooks:
       - id: check-jsonschema
-        files: ^modules/nf-core/*/meta.yml$
+        files: ^modules/nf-core/[^/]+/[^/]+/meta.yml$
         args: ["--schemafile", "yaml-schema.json"]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,3 +3,9 @@ repos:
     rev: "v2.7.1"
     hooks:
       - id: prettier
+  - repo: https://github.com/python-jsonschema/check-jsonschema
+    rev: 0.21.0
+    hooks:
+      - id: check-jsonschema
+        files: ^modules/nf-core/**/meta.yml$
+        args: ["--schemafile", "yaml-schema.json"]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,5 +7,6 @@ repos:
     rev: 0.21.0
     hooks:
       - id: check-jsonschema
-        files: ^modules/nf-core/[^/]+/[^/]+/meta.yml$
+        # match meta.ymls in one of the subdirectories of modules/nf-core
+        files: ^modules/nf-core/.*/meta\.yml$
         args: ["--schemafile", "yaml-schema.json"]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,5 +7,5 @@ repos:
     rev: 0.21.0
     hooks:
       - id: check-jsonschema
-        files: ^modules/nf-core/**/meta.yml$
+        files: ^modules/nf-core/*/meta.yml$
         args: ["--schemafile", "yaml-schema.json"]
--- a/yaml-schema.json
+++ b/yaml-schema.json
@@ -17,7 +17,8 @@
             "description": "Keywords for the module",
             "items": {
                 "type": "string"
-            }
+            },
+            "minItems": 3
         },
         "authors": {
             "type": "array",
@@ -29,6 +30,7 @@
         "input": {
             "type": "array",
             "description": "Input channels for the module",
+
             "items": {
                 "type": "object",
                 "patternProperties": {
@@ -37,7 +39,8 @@
                         "properties": {
                             "type": {
                                 "type": "string",
-                                "description": "Type of the input channel"
+                                "description": "Type of the input channel",
+                                "enum": ["map", "file", "directory", "string", "integer", "float"]
                             },
                             "description": {
                                 "type": "string",
@@ -45,7 +48,7 @@
                             },
                             "pattern": {
                                 "type": "string",
-                                "description": "Pattern of the input channel"
+                                "description": "Pattern of the input channel, given in Java glob syntax"
                             },
                             "default": {
                                 "type": ["string", "number", "boolean", "array", "object"],
@@ -68,7 +71,8 @@
                         "properties": {
                             "type": {
                                 "type": "string",
-                                "description": "Type of the output channel"
+                                "description": "Type of the output channel",
+                                "enum": ["map", "file", "directory", "string", "integer", "float"]
                             },
                             "description": {
                                 "type": "string",
@@ -76,7 +80,7 @@
                             },
                             "pattern": {
                                 "type": "string",
-                                "description": "Pattern of the output channel"
+                                "description": "Pattern of the input channel, given in Java glob syntax"
                             }
                         },
                         "required": ["type", "description"]
@@ -99,19 +103,23 @@
                             },
                             "homepage": {
                                 "type": "string",
-                                "description": "Homepage of the tool"
+                                "description": "Homepage of the tool",
+                                "pattern": "^(http|https)://.*$"
                             },
                             "documentation": {
                                 "type": "string",
-                                "description": "Documentation of the tool"
+                                "description": "Documentation of the tool",
+                                "pattern": "^(http|https)://.*$"
                             },
                             "tool_dev_url": {
                                 "type": "string",
-                                "description": "URL of the development version of the tool's documentation"
+                                "description": "URL of the development version of the tool's documentation",
+                                "pattern": "^(http|https)://.*$"
                             },
                             "doi": {
                                 "type": "string",
-                                "description": "DOI of the tool"
+                                "description": "DOI of the tool",
+                                "pattern": "^10.\\d{4,9}\\/[^,]+$"
                             },
                             "licence": {
                                 "type": ["array", "string"],
--- a/yaml-schema.json
+++ b/yaml-schema.json
@@ -48,7 +48,7 @@
                                 "description": "Pattern of the input channel"
                             },
                             "default": {
-                                "type": "string",
+                                "type": ["string", "number", "boolean", "array", "object"],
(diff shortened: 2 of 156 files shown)


yaml-schema.json
```diff
@@ -0,0 +1,137 @@
...
+                            { "required": ["documentation"] },
+                            { "required": ["tool_dev_url"] },
+                            { "required": ["doi"] }
+                        ]
+                    }
+                }
+            }
+        }
+    },
+    "required": ["name", "description", "keywords", "authors", "output", "tools"]
```
mirpedrol, line 136: Isn't input required too?
```suggestion
    "required": ["name", "description", "keywords", "authors", "output", "input", "tools"]
```

.pre-commit-config.yaml
```diff
@@ -3,3 +3,9 @@ repos:
     rev: "v2.7.1"
     hooks:
       - id: prettier
+  - repo: https://github.com/python-jsonschema/check-jsonschema
+    rev: 0.21.0
+    hooks:
+      - id: check-jsonschema
+        files: ^modules/nf-core/**/meta.yml$
```
mirpedrol, line 10:
```suggestion
        files: ^modules/nf-core/*/meta.yml$
```

mirpedrol (approved):
Looks good :) 

We will still have the problem of "wrong" indentation. When the indentation is not wrong but not in the way we like, this makes the command `nf-core modules info` fail. For example:
```
  - html:
    type: file
    description: FastQC report
    pattern: "*_{fastqc.html}"
```
instead of 
```
  - html:
      type: file
      description: FastQC report
      pattern: "*_{fastqc.html}"
```

yaml-schema.json
mashehu: Nope, there a some tools which don't have an input (usually the download their needed data), e.g. https://github.com/nf-core/modules/blob/master/modules/nf-core/krona/kronadb/meta.yml (I switched the input there to a comment)

mashehu: But this is maybe also not the optimal solution, because one might be confused about that. Maybe I will show a standard message on the website if there is no input needed.

mashehu:
> Looks good :)
> 
> We will still have the problem of "wrong" indentation. When the indentation is not wrong but not in the way we like, this makes the command `nf-core modules info` fail. For example:

I tried it out and my schema checker complains correctly about it (but doesn't fix it automatically, we still need to do that by ourselves).

mirpedrol:
I tested this pre-commit hook with Gitpod and did not fail, but now having a look at the follow up PR https://github.com/nf-core/tools/pull/2190 I see that the linting complains :) 
I get this error message "The meta.yml of the module fastqc is not valid: None is not of type 'object'. Check that the child entries of output.html are indented correctly."

mashehu:
> I tested this pre-commit hook with Gitpod and did not fail, but now having a look at the follow up PR [nf-core/tools#2190](https://github.com/nf-core/tools/pull/2190) I see that the linting complains :) I get this error message "The meta.yml of the module fastqc is not valid: None is not of type 'object'. Check that the child entries of output.html are indented correctly."

Yep, that is working as expected. Or were you thinking of something else?

mirpedrol:
> Yep, that is working as expected. Or were you thinking of something else?

No, works very nice 😄 
I think it's fine then, even if the pre-commit hook doesn't mention the indentation because linting will.

yaml-schema.json
mirpedrol: Oh, ok I see 👍

mashehu:
ah, you meant the git-hook doesn't do anything? let me check.

.pre-commit-config.yaml
```diff
@@ -3,3 +3,9 @@ repos:
     rev: "v2.7.1"
     hooks:
       - id: prettier
+  - repo: https://github.com/python-jsonschema/check-jsonschema
+    rev: 0.21.0
+    hooks:
+      - id: check-jsonschema
+        files: ^modules/nf-core/[^/]+/[^/]+/meta.yml$
```
mirpedrol, line 10: Sorry I didn't realise it was this regex. I think this should work
```suggestion
        files: ^modules/nf-core/.*/meta\.yml$
```

mashehu: jep, also didn't see it. Tested it now and works (error messages are not so helpful, but that is not on our side...

Emiller88:
Could we also add line to all the meta yamls with something like [this](https://github.com/onedr0p/home-ops/blob/e456e6ac5b772839039f065af30ff7c6f437569c/kubernetes/apps/rook-ceph/rook-ceph/app/helmrelease.yaml#L2)
```yaml
---
# yaml-language-server: $schema=https://path-to-schema.json
```

yaml-schema.json
```diff
@@ -0,0 +1,138 @@
...
+            "description": "Input channels for the module",
+            "items": {
+                "type": "object",
+                "patternProperties": {
+                    ".*": {
+                        "type": "object",
+                        "properties": {
+                            "type": {
+                                "type": "string",
+                                "description": "Type of the input channel"
```
jfy133, line 40: Should we set an enum based on: https://nf-co.re/docs/contributing/modules#documentation point 6?

yaml-schema.json
```diff
@@ -0,0 +1,138 @@
...
+                            },
+                            "description": {
+                                "type": "string",
+                                "description": "Description of the input channel"
+                            },
+                            "pattern": {
+                                "type": "string",
+                                "description": "Pattern of the input channel"
```
jfy133, line 48: Specify that this ideally should be based on a java glob?

yaml-schema.json
jfy133: (applies to output: too)

yaml-schema.json
```diff
@@ -0,0 +1,138 @@
...
+        "name": {
+            "type": "string",
+            "description": "Name of the module"
+        },
+        "description": {
+            "type": "string",
+            "description": "Description of the module"
+        },
+        "keywords": {
+            "type": "array",
```
jfy133, line 16: Can you specify a minimum number of elements in the array? Might be good to stop people just putting `samtools` or something not very useful (I would suggest at least 3)

yaml-schema.json
```diff
@@ -0,0 +1,138 @@
...
+                "patternProperties": {
+                    ".*": {
+                        "type": "object",
+                        "properties": {
+                            "description": {
+                                "type": "string",
+                                "description": "Description of the output channel"
+                            },
+                            "homepage": {
+                                "type": "string",
```
jfy133, line 101: Include a regex pattern starrting `http*` or something? (applies to URLs too)

yaml-schema.json
```diff
@@ -0,0 +1,138 @@
...
+                                "description": "Documentation of the tool"
+                            },
+                            "tool_dev_url": {
+                                "type": "string",
+                                "description": "URL of the development version of the tool's documentation"
+                            },
+                            "doi": {
+                                "type": "string",
+                                "description": "DOI of the tool"
```
jfy133, line 114: Example pattern you could use: https://github.com/SPAAM-community/AncientMetagenomeDir/blob/230bca6d7af66b52c41fa79fc55cf51a4fdfb62d/ancientmetagenome-hostassociated/samples/ancientmetagenome-hostassociated_samples_schema.json#L53

yaml-schema.json
```diff
@@ -0,0 +1,138 @@
...
+                            },
+                            "licence": {
+                                "type": ["array", "string"],
+                                "description": "Licence of the tool",
```
jfy133, line 118: Maybe an enum using: http://bio-tools.github.io/biotoolsSchema/#Link21 ?

jfy133 (commented):
Overall looks good but I would rather make it more strict :grimacing:

yaml-schema.json
mashehu: Added it now locally and I get quite a few deviations. Because I see quite a lot of "list" entries coming up, should we add it to the options? (if that is the correct groovy data object)

yaml-schema.json
mashehu: Hmm, that list is a bit too long and would make the schema quite unwieldy I think.

yaml-schema.json
mirpedrol: Looking for an example, I found [justrma](https://github.com/nf-core/modules/blob/master/modules/nf-core/affy/justrma/main.nf) and the [yml](https://github.com/nf-core/modules/blob/master/modules/nf-core/affy/justrma/meta.yml). Looks like `celfiles_dir` should be `directory` instead of `list`, no?
I don't know if there are other modules with different situations.

jfy133: I would rather say `list` isn't so informative for the type, it's the contents of the list is more important (is it a list of files?)

yaml-schema.json
jfy133: I thought with JSON schema you can refer to 'remote' enums?   But agree wouldn't want to embed the whole list.

jfy133: I wonder how bioconda/conda does it :thinking:

yaml-schema.json
mashehu: nested json schemas are not yet supported, but will be soon...

mirpedrol: I would opt for getting rid of `list` and just specifying the type of the list elements

yaml-schema.json
mashehu: ah, true! will try that

yaml-schema.json
mashehu: I think it is important to indicate if it is more then one item, so I would go for nested here, so we can actually check the items themselves as well.

jfy133: But shouldn't that be inthe description (that's normally how I do it)

type: file
description: A file or list of files that XYZ

mashehu: if we make it nested, we could then for example check the file extension of each individual file. More validation power 🦾

jfy133: :spaghetti:

PR #2917, opened 2023-02-23
Fix meta.ymls based on json schema 

mashehu:
Based on #2913 

- fix indentations, keywords, and typos 
- add missing types and tool urls
- remove empty fields


mashehu:
@nf-core-bot fix linting

modules/nf-core/custom/tabulartogseagct/meta.yml
```diff
@@ -6,6 +6,7 @@ keywords:
 tools:
   - "tabulartogseagct":
```
mirpedrol, line 7:
```suggestion
  - tabulartogseagct:
```

mirpedrol (approved):
Thank you! This looks like a lot of work!

mashehu:
These were the schema validation errors I could fix from my side. For the others, module authors need to handle it, when the want to update their modules

modules/nf-core/hapibd/meta.yml
```diff
@@ -10,8 +10,8 @@ tools:
       description: "Hap-ibd Detects identity-by-descent (IBD) segments and homozygosity-by-descent (HBD) segments in phased genotype data."
       homepage: "https://github.com/browning-lab/hap-ibd/blob/master/README.md"
       documentation: "https://github.com/browning-lab/hap-ibd/blob/master/README.md"
-      tool_dev_url: "None"
-      doi: "https://doi.org/10.1016/j.ajhg.2020.02.010"
```
mirpedrol, line 14: is the schema seeing the complete url as an error? I think it will be difficult to enforce this structure otherwise.

mashehu: yeah, the pattern is to only have the doi number. Means I don't have to check if it is there when rendering a link on the website.

matthdsm:
@mashehu , is the good to merge? If so, please feel free to hit the big green button!

mashehu:
First want to merge #2913 (because we added some more changes there) and then this one

PR #2927, opened 2023-02-27
Glimpse2 chunk

LouisLeNezet:
## PR checklist
Add new Glimpse2_chunk module.

- [X] Add a resource `label`


LouisLeNezet:
Should I check for the correct use off the model parameter ? (non / mandatory argument)

modules/nf-core/glimpse2/chunk/main.nf
```diff
@@ -0,0 +1,51 @@
...
+    output:
+    tuple val(meta), path("*.txt"), emit: chunk_chr
+    path "versions.yml"           , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def prefix    = task.ext.prefix ?: "${meta.id}"
+    def model_cmd = model=="rec" ? "--recursive" : model=="seq" ? "--sequential" : model=="unv" ? "--uniform-number-variants" : ""
```
pinin4fjords, line 32: Is there a reason there's an alias used for the model? 

If the value of `mode` was just 'recursive', 'sequential' or 'uniform-number-variants', you could pass directly in the command section like:

```
--${model}
```

You could lean on validation in the workflow schema to limit the options available.

LouisLeNezet: You are completely right :)
When you talk about the workflow schema to limit the options available, is it about the different pattern available in the meta.yml ?

pinin4fjords: So, when people write workflows that use this module, and if they provide an option to pass through to this parameter, they can restrict the available values like [this](https://github.com/nf-core/differentialabundance/blob/c5de0561506f73887979909274b9f7155ef1cf90/nextflow_schema.json#L27). 

A pattern in the meta.yml would be a very sensible thing to do for documentation anyway, but I'm not sure how enforced they are. You could add a different bit of groovy to the module where you had your code before to check that `model` is one of the allowed values, but I haven't seen a lot of that in module code and it's probably easier to let the tool die if it received a bad parameter name.

tests/modules/nf-core/glimpse2/chunk/nextflow.config
```diff
@@ -0,0 +1,9 @@
+process {
+    withName: GLIMPSE2_CHUNK {
+        ext.args = [
```
pinin4fjords, line 3: Take this out, since it's empty now?

LouisLeNezet: Done

modules/nf-core/glimpse2/chunk/meta.yml
```diff
@@ -0,0 +1,74 @@
...
+tools:
+  - "glimpse2":
+      description: "GLIMPSE2 is a phasing and imputation method for large-scale low-coverage sequencing studies."
+      homepage: "https://odelaneau.github.io/GLIMPSE"
+      documentation: "https://odelaneau.github.io/GLIMPSE/commands.html"
+      tool_dev_url: "https://github.com/odelaneau/GLIMPSE"
+      doi: "10.1038/s41588-020-00756-0"
+      licence: "['MIT']"
+
+requirement: AVX2
```
pinin4fjords, line 15:
```suggestion
requirements: 
  - AVX2
```

Sorry, was just reading https://nfcore.slack.com/archives/CE6SDBX2A/p1677231429855929 to understand what's going on here. This is the first time this has been used, so let's be consistent with the other fields and make it a plural and a list.

Otherwise I think we're good, but I'll ask for a second review as there's some new stuff here.

LouisLeNezet: No problem.
Thanks a lot for your review !

modules/nf-core/glimpse2/chunk/main.nf
```diff
@@ -0,0 +1,50 @@
+process GLIMPSE2_CHUNK {
+    tag "$meta.id"
+    label 'process_low'
+
+    beforeScript  """
+        if cat /proc/cpuinfo | grep avx2 -q
+        then
+            echo "Feature AVX2 present on host"
+        else
+            echo "Feature AVX2 not present on host"
```
SPPearce, line 10: Is the resulting error message clear that this module is the trigger?

modules/nf-core/glimpse2/chunk/meta.yml
```diff
@@ -0,0 +1,75 @@
...
+      type: file
+      description: File containing the genetic map.
+      pattern: "*.gmap"
+
+  - model:
+      type: string
+      description: |
+        Algorithm model to use :
+        "recursive": Recursive algorithm
+        "sequential": Recommended sequential algorithm
```
SPPearce, line 53:
```suggestion
        "sequential": Sequential algorithm (Recommended)
```

SPPearce: Who is actually recommending this?

modules/nf-core/glimpse2/chunk/meta.yml
```diff
@@ -0,0 +1,75 @@
...
+        "uniform-number-variants": Experimental. Uniform the number of variants in the sequential algorithm
```
SPPearce, line 54: I don't understand this.

modules/nf-core/glimpse2/chunk/meta.yml
LouisLeNezet: The developer of the software.
Everything can be find [here](https://odelaneau.github.io/GLIMPSE/docs/documentation/split_reference/)

modules/nf-core/glimpse2/chunk/main.nf
LouisLeNezet: The error look like this
```
Caused by:
  Process `test_shapeit5_phaserare:SHAPEIT5_PHASERARE (input)` terminated with an error exit status (1)

Command executed:

  SHAPEIT5_phase_rare \
       \
      --input-plain 1000GP.chr21.noNA12878.s.bcf \
      --scaffold input_chr21.vcf.gz \
       \
       \
      --input-region chr21:16600000-16800000 \
      --scaffold-region chr21:16600000-16800000 \
      --thread 2 \
      --output input_chr21_16600000-16800000.vcf.gz
  
      "test_shapeit5_phaserare:SHAPEIT5_PHASERARE":
          shapeit5: "$(SHAPEIT5_phase_rare | sed -nr '/Version/p' | grep -o -E '([0-9]+.){1,2}[0-9]' | head -1)"
  END_VERSIONS

Command exit status:
  1

Command output:
  (empty)

Command error:
  Feature AVX2 not present on node
```

modules/nf-core/glimpse2/chunk/meta.yml
LouisLeNezet: Their is 3 different algorithm that can be used.
You need to select one of those as mentioned [here](https://odelaneau.github.io/GLIMPSE/docs/documentation/chunk/)
That's the unconventional part of this module.
Do you have a better way of checking this ?

modules/nf-core/glimpse2/chunk/meta.yml
```diff
@@ -0,0 +1,75 @@
+name: "glimpse2_chunk"
+description: Defines chunks where to run imputation
+keywords:
+  - chunk
+  - imputation
```
nvnieuwk, line 5: Can you add some more keywords?

nvnieuwk (approved):
LGTM, only one small comment!

modules/nf-core/glimpse2/chunk/meta.yml
LouisLeNezet: Is it better with the new ones ?

nvnieuwk: Yes, thank you!

PR #2945, opened 2023-03-01
fix archive with subdirectories bug

mapo9:

Hi, I experienced some trouble with this module when trying to unzip an archive containing a main-directory with subdirectories.
In that case, all files that belong to the subdirs are located in the main-directory and the respective subdirs are empty.
The minor change we applied should fix this problem.


ggabernet:
yes I can corroborate that we looked at this together with @mapo9 and found this solution, but it would be great if @jfy133 could have a look as the original author of the module!

jfy133 (approved):
Sounds good to me I think, according to the modules page, the module is not yet used in any (released) pipelines anyway so we aren't going to necessarily break things by using it I guess

PR #2965, opened 2023-03-04
Update fastqc version

ekushele:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`


ewels (changes requested):
Looks like this release added a `--memory` option. I think it would probably be good to add that to the module.

Apparently it can also do SVG plots now. Might be a good idea to add this too?

ekushele:
Now it's failing also for docker and conda..

modules/nf-core/fastqc/main.nf
```diff
@@ -20,6 +20,10 @@ process FASTQC {
 
     script:
     def args = task.ext.args ?: ''
+    def cpu = task.cpus
```
matthdsm, line 23: Why all these calculations and not just `task.memory`?

modules/nf-core/fastqc/main.nf
```diff
@@ -20,6 +20,10 @@ process FASTQC {
...
+    def memory = task.memory.toMega() > 10000 ? 10000 : task.memory.toMega()
+    if (memory/250 < cpu)
+        cpu = Math.floor(memory/250).toInteger()
```
matthdsm, line 26: I don't think you should modify the number of cpu's inside the module 😨

ekushele: Because of this (from the manual):

>     -t --threads    Specifies the number of files which can be processed
>                 simultaneously.  Each thread will be allocated 250MB of
>                 memory so you shouldn't run more threads than your
>                 available memory will cope with, and not more than
>                 6 threads on a 32 bit machine

So I thought a thread must have at least 250MB memory, so if one gives 500Mb and 7 threads, there won't be enough memory to all the threads

modules/nf-core/fastqc/main.nf
ekushele: see below: https://github.com/nf-core/modules/pull/2965#discussion_r1129859990
And aslo when I ran I got the error: 
`Memory value 15000 MB was outside the allowed range (100 - 10000) at /usr/local/bin/fastqc line 203.`

Which I understand as limiting the memory to 10000

ekushele: @matthdsm I'm sure your busy.. When you have time to look at it, I'll be happy

matthdsm:
Hi @ekushele,

I'm unsure what the best way to go here is. I'm not a fan of messing with user defined resource requirements.
I think it's best to ask for some opinions from the @nf-core/maintainers here.

M

Midnighter:
I tend to agree with @matthdsm. It's usually better to try and not be too smart but rather give users free hand to configure the module. Maybe the important information that you shared here regarding memory and threads could be lifted to the `meta.yml` somehow? IMO, documenting it but still using `task.cpus` and `task.memory` directly seems better.

nvnieuwk:
> I tend to agree with @matthdsm. It's usually better to try and not be too smart but rather give users free hand to configure the module. Maybe the important information that you shared here regarding memory and threads could be lifted to the `meta.yml` somehow? IMO, documenting it but still using `task.cpus` and `task.memory` directly seems better.

I agree somehow documenting this should be the solution, maybe also add a comment or two in the code since some people only look at the `main.nf`.

SPPearce:
This is getting `process_medium` as a tag.
The default for that is 6 cpus and 36Gb. 
I can't really imagine a system setup that doesn't have at least 500Mb per cpu, that is **very** little.

But the default `process_medium` will go over 10Gb, so presumably that will give an error for the default amount? 
Hard capping that in the code sounds fine if that is a limit of the tool.

modules/nf-core/fastqc/main.nf
```diff
@@ -20,16 +20,18 @@ process FASTQC {
...
     def old_new_pairs = reads instanceof Path || reads.size() == 1 ? [[ reads, "${prefix}.${reads.extension}" ]] : reads.withIndex().collect { entry, index -> [ entry, "${prefix}_${index + 1}.${entry.extension}" ] }
     def rename_to = old_new_pairs*.join(' ').join(' ')
     def renamed_files = old_new_pairs.collect{ old_name, new_name -> new_name }.join(' ')
+    //from FastQC man: Each thread will be allocated 250MB of memory so you shouldn't run more threads than your available memory will cope with
     """
     printf "%s %s\\n" $rename_to | while read old_name new_name; do
         [ -f "\${new_name}" ] || ln -s \$old_name \$new_name
     done
-    fastqc $args --threads $task.cpus $renamed_files
+    fastqc $args --memory $memory --threads $task.cpu $renamed_files
```
maxulysse, line 34:
```suggestion
    fastqc $args --memory $memory --threads $task.cpus $renamed_files
```

ekushele:
How can the test pass for docker but fail for singularity and conda?

SPPearce:
Closed as we seem to already be on 0.12.1 with a cap on 10gb memory.

PR #2996, opened 2023-03-08
add module bcftools/split

matthdsm:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`

### DIFF
--- a/modules/nf-core/bcftools/split/main.nf
+++ b/modules/nf-core/bcftools/split/main.nf
@@ -1,0 +1,49 @@
+process BCFTOOLS_SPLIT {
+    tag "$meta.id"
+    label 'process_single'
+
+    conda "bioconda::bcftools=1.16"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/bcftools:1.16--hfe4b78e_1':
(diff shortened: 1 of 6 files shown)


maxulysse:
LGTM.
How does it works for species with only 1 chromosome?
How does it handles the weird chromosome names one can get?

matthdsm:
> How does it works for species with only 1 chromosome?

1 chromosome == 1 file, as is the case with the test data

> How does it handles the weird chromosome names one can get?

I didn't test with any weird alt chr names, but since it's only a string it shouldn't be an issue

maxulysse (approved):

matthdsm:
Thanks @maxulysse

maxulysse:
ok, then totally useless if only one chromosome?
Do we have way to count the number of chromosome before?

atrigila (approved):

matthdsm:
> ok, then totally useless if only one chromosome?
I guess so, but the overhead of this process is very small, so I don't think it's much of an issue.

> Do we have way to count the number of chromosome before?
Not from the top of my head no. We could use `bcftools index -s` but that's the same as this module

matthdsm:
Agreed.
But I don't see how we could get around that without another process, which would amount to the same issue, but even more so, because we'd have run 2 processes for the same result.

PR #3023, opened 2023-03-10
update bcftools norm

atrigila:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`

Update to the bcftools norm module. Previously, the output was fixed to be a vcf.gz. Now it allows for selection (bcf, vcf.gz, etc) inspired by the module bcftools sort https://github.com/nf-core/modules/pull/2920 . The default output is still a vcf.gz.

### DIFF
--- a/modules/nf-core/bcftools/norm/main.nf
+++ b/modules/nf-core/bcftools/norm/main.nf
@@ -12,19 +12,24 @@ process BCFTOOLS_NORM {
     path(fasta)
 
     output:
-    tuple val(meta), path("*.gz") , emit: vcf
-    path "versions.yml"           , emit: versions
+    tuple val(meta), path("*.{vcf,vcf.gz,bcf,bcf.gz}")  , emit: vcf
+    path "versions.yml"                                 , emit: versions
 
     when:
     task.ext.when == null || task.ext.when
 
     script:
-    def args = task.ext.args ?: ''
+    def args = task.ext.args ?: '--output-type z'
     def prefix = task.ext.prefix ?: "${meta.id}"
+    def extension = args.contains("--output-type b") || args.contains("-Ob") ? "bcf.gz" :
+                    args.contains("--output-type u") || args.contains("-Ou") ? "bcf" :
+                    args.contains("--output-type z") || args.contains("-Oz") ? "vcf.gz" :
+                    args.contains("--output-type v") || args.contains("-Ov") ? "vcf" :
+                    "vcf"
     """
     bcftools norm \\
         --fasta-ref ${fasta} \\
-        --output ${prefix}.vcf.gz \\
+        --output ${prefix}.${extension}\\
         $args \\
         --threads $task.cpus \\
         ${vcf}
@@ -36,9 +41,15 @@ process BCFTOOLS_NORM {
     """
 
     stub:
+    def args = task.ext.args ?: '--output-type z'
     def prefix = task.ext.prefix ?: "${meta.id}"
+    def extension = args.contains("--output-type b") || args.contains("-Ob") ? "bcf.gz" :
+                    args.contains("--output-type u") || args.contains("-Ou") ? "bcf" :
+                    args.contains("--output-type z") || args.contains("-Oz") ? "vcf.gz" :
+                    args.contains("--output-type v") || args.contains("-Ov") ? "vcf" :
+                    "vcf"
     """
-    touch ${prefix}.vcf.gz
+    touch ${prefix}.${extension}
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":


modules/nf-core/bcftools/norm/main.nf
```diff
@@ -12,19 +12,24 @@ process BCFTOOLS_NORM {
     path(fasta)
 
     output:
-    tuple val(meta), path("*.gz") , emit: vcf
-    path "versions.yml"           , emit: versions
+    tuple val(meta), path("*.{vcf,vcf.gz,bcf,bcf.gz}")  , emit: vcf
```
FriederikeHanssen, line 15: in line with other modules, I would suggest to divide between vcf and bcf  https://github.com/nf-core/modules/blob/master/modules/nf-core/samtools/view/main.nf

tests/modules/nf-core/bcftools/norm/test.yml
```diff
@@ -1,15 +1,19 @@
...
 - name: bcftools norm test_bcftools_norm_tbi
   command: nextflow run ./tests/modules/nf-core/bcftools/norm -entry test_bcftools_norm_tbi -c ./tests/config/nextflow.config -c ./tests/modules/nf-core/bcftools/norm/nextflow.config
   tags:
-    - bcftools
     - bcftools/norm
+    - bcftools
   files:
     - path: output/bcftools/test2.vcf.gz
+      md5sum: a7524541481cd4e7d38267ded10253e2
+    - path: output/bcftools/versions.yml
```
FriederikeHanssen, line 19: can you add a test for at least one other output type

modules/nf-core/bcftools/norm/main.nf
atrigila: Thanks for the suggestion! Would you also divide between the compressed and uncompressed versions (e.g. vcf and .vcf.gz)?

tests/modules/nf-core/bcftools/norm/test.yml
atrigila: Definitely! Thanks for the suggestion @FriederikeHanssen

modules/nf-core/bcftools/norm/main.nf
FriederikeHanssen: yes, it would ensure people really get out what they expect.

modules/nf-core/bcftools/norm/main.nf
```diff
@@ -12,19 +12,28 @@ process BCFTOOLS_NORM {
     path(fasta)
 
     output:
-    tuple val(meta), path("*.gz") , emit: vcf
-    path "versions.yml"           , emit: versions
+    tuple val(meta), path("*.vcf")      , emit: vcf     , optional: true
+    tuple val(meta), path("*.vcf.gz")   , emit: vcfgz   , optional: true
+    tuple val(meta), path("*.bcf")      , emit: bcf     , optional: true
+    tuple val(meta), path("*.bcf.gz")   , emit: bcfgz   , optional: true
```
nvnieuwk, line 18: I think it would be better to output everything in one output channel here since it's always one of the four. So use `path("*.{vcf,vcf.gz,bcf,bcf.gz}")` here :)

SPPearce: I agree.

nvnieuwk (approved):
LGTM!

modules/nf-core/bcftools/norm/main.nf
FriederikeHanssen: hm that is very inconsistent with how we treat the samtools modules for example. There it will also always be bam or cram, but we still output them in separate channels

nvnieuwk: But yeah wouldn't it be more flexible if we didn't do that? I've seen some subwfs that contain bottlenecks because of this during the spring cleaning

atrigila: Thanks for your recommendations! I made a new update.

nvnieuwk: Thanks can you wait with the merge until @FriederikeHanssen agrees too?

atrigila: Absolutely! Sorry I haven't seen the previous messages. In my opinion, I think @FriederikeHanssen suggestion is tidier but I do not have much experience and am not aware about the bottlenecks.

nvnieuwk: Here is an example: https://github.com/nf-core/modules/blob/ffa9641ee18f88aff974257cb50ba3cf8f7d143c/subworkflows/nf-core/bam_dedup_stats_samtools_umitools/main.nf#L30-L32.
Here the join statement use remainder true because it can't know whether a `.bai` or `.crai` file has been created. The pipeline will wait here until all `SAMTOOLS_INDEX` processes are run before creating the channels with `remainder`. The pipeline won't be on hold here if `SAMTOOLS_INDEX` were to have one channel for `bai` and `crai` (e.g. `index`).

FriederikeHanssen: ok, maybe you bring it up quickly in maintainers or modules, so we can add it as a guideline? I think this will impact a whole bunch of modules, and at that time we probably didn't consider subworkflows where you don't know the data type

nvnieuwk: I'll post about it on Slack, thanks!

maxulysse: You should know if you expect a bai or a csi, I'm assuming you have a params for that

maxulysse: but I see your point, and it could be interesting to have just index as output to samtools to capture both possiblity of bai/csi.
It's true that they are different files, but they are created in the same logic for the same purpose

nvnieuwk: Yeah but that's not the case in that subworkflow. And what about `.bai` and `.crai`? That's entirely dependent on the input file

nvnieuwk: > but I see your point, and it could be interesting to have just index as output to samtools to capture both possiblity of bai/csi. It's true that they are different files, but they are created in the same logic for the same purpose

Yes! But we should definitely get a lot of eyes on this :) as Rike said, it will be a breaking change

atrigila: What would you prefer me to do with this module for now?

PR #3027, opened 2023-03-13
Limit gatk4 heap size

matthdsm:
Limit gatk4 heap mem to 80% of allotted memory to account for overhead usage

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`

### DIFF
--- a/modules/nf-core/gatk4/annotateintervals/main.nf
+++ b/modules/nf-core/gatk4/annotateintervals/main.nf
@@ -36,7 +36,7 @@ process GATK4_ANNOTATEINTERVALS {
     if (!task.memory) {
         log.info '[GATK AnnotateIntervals] Available memory not known - defaulting to 3GB. Specify process memory requirements to change this.'
     } else {
-        avail_mem = task.memory.giga - 1
+        avail_mem = task.memory.giga * 0.8
     }
 
     """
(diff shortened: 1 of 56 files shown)


maxulysse:
why 80%?

maxulysse:
ok, so for your own infra then?

matthdsm:
So it seems a float value doesn't work and neither does `ceil`. Suggestions are welcome.

pontus:
80% is a good rule of thumb (at least for not very small runs). The current "pass the entire memory size as heap" is definitely wrong and is quiet likely to lead to unnecessary OOM kills (where resources are really adequate).

pontus:
If one wants to stick with Gigabytes, `(task.memory.giga*0.8).intValue()` should work, but I have a bit of a preference for switching to Megabytes instead, so I'd suggest `(task.memory.mega*0.8).intValue()` (and replacing `"-Xmx${avail_mem}g"` with `"-Xmx${avail_mem}M"` ).

matthdsm:
Thanks for the suggestion @pontus, I've applied it.

matthdsm:
image

pontus (approved):

pontus:
I can confidently say that passing the memory limit as max heap (-Xmx) is wrong. 

As for the actual value to use, the optimal value will be workload (i.e. application and sample) dependent, but I'd go with 80% if there are no indications of other values that would be better.

If this would be used for very small (say with a memory limit of a few hundred megabytes), it might be that additional tuning should be applied, but I don't think any pipeline does this.

maxulysse (approved):

matthdsm:
Thanks @pontus and @maxulysse !

PR #3032, opened 2023-03-14
New module: Plink exclude

atrigila:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
 
This module allows to exclude variants (stored in a file) from a plink binary set.


modules/nf-core/plink/exclude/main.nf
```diff
@@ -0,0 +1,40 @@
...
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    if( "$bed" == "${prefix}.bed" ) error "Input and output names are the same, use \"task.ext.prefix\" to disambiguate!"
+    """
+    plink \\
+        --bfile ${meta.id} \\
+        $args \\
+        --exclude $variants \\
+        --threads $task.cpus \\
+        --make-bed \\
```
nvnieuwk, line 32: Is `--make_bed` an argument that always has to be given when running this tool?

modules/nf-core/plink/exclude/meta.yml
```diff
@@ -0,0 +1,60 @@
+name: plink_exclude
+description: Exclude variant identifiers from plink bfiles
+keywords:
+  - exclude
+  - plink
+tools:
+  - plink:
+      description: Whole genome association analysis toolset, designed to perform a range of basic, large-scale analyses in a computationally efficient manner.
+      homepage: https://www.cog-genomics.org/plink
+
```
nvnieuwk, line 10:
```suggestion
```

nvnieuwk (changes requested):
Some small comments :)

modules/nf-core/plink/exclude/main.nf
atrigila: In fact, no. My first approach when designing this module was to pass it as an `ext.args`, following nf-core guidelines. However, I saw that `plink extract` has this argument explicit and I tried to be consistent between similar modules. As your input here is a bed-bim-fam and you would like to extract/exclude variants from this triset, the `--make-bed` forces the output to keep being a bed-bim-fam. I'd like your opinion on this, but I think that I'll be better to add the `--make-bed` to the `ext.args`.

nvnieuwk: I think you should keep it in `ext.args` and adjust your output channels to also work when this argument isn't given (I assume a BED file isn't created if this argument isn't given?)

atrigila: If you don't provide  `--make-bed` the program breaks with "`Warning: No output requested.  (Did you forget --make-bed?)  Exiting.` " `--make-bed`  makes the bed-bim-fam, not only the bed.

atrigila: Yes, you need to commonly pass  `--make-bed`  to get the output or the command `--recode` to transform the output to another file type. Thanks for your feedback! :)

nvnieuwk: Ah so there is an option to create another file type? Then you should adjust the module to work for both :)

nvnieuwk (approved):

PR #3033, opened 2023-03-14
added cellranger vdj/mkvdjref modules, with module test for vdj

klkeys:
## PR checklist

- [x] Add a resource `label`
~- [ ] Use BioConda and BioContainers if possible to fulfil software requirements.~ only works with `nfcore/cellranger` Docker image

adds new `cellranger` module `cellranger vdj` for single cell immune profiling

mostly cribbed from `cellranger count` and `cellranger mkref`

includes the corresponding `cellranger mkvdj` ref just in case

test data included in https://github.com/nf-core/test-datasets/pull/787


modules/nf-core/cellranger/vdj/main.nf
```diff
@@ -0,0 +1,54 @@
+process CELLRANGER_VDJ {
+    tag "$meta.gem"
```
SPPearce, line 2: Why are you introducing meta.gem here? The standard syntax is id.

modules/nf-core/cellranger/vdj/main.nf
```diff
@@ -0,0 +1,54 @@
...
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def sample_arg = meta.samples.unique().join(",")
+    def reference_name = reference.name
+    """
+    cellranger \\
+        vdj \\
+        --id='sample-${meta.gem}' \\
```
SPPearce, line 30: Why are you prefixing with "-sample"?

SPPearce (commented):
Is there a possibility to have a test for `mkvdjref`?

modules/nf-core/cellranger/mkvdjref/main.nf
```diff
@@ -0,0 +1,38 @@
...
+    path "versions.yml"     , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    """
+    cellranger \\
+        mkref \\
```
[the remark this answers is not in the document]
klkeys: ooh yes good catch

modules/nf-core/cellranger/vdj/main.nf
klkeys: that's how the `cellranger count` module is coded:

https://github.com/nf-core/modules/blob/ffaff8089fa2dec69e52a18c72384eee32bea2c4/modules/nf-core/cellranger/count/main.nf#L2

happy to change it to `"${meta.id}"` but in that case we should consider updating `cellranger count` too

modules/nf-core/cellranger/vdj/main.nf
klkeys: also the style from `cellranger count`:

https://github.com/nf-core/modules/blob/ffaff8089fa2dec69e52a18c72384eee32bea2c4/modules/nf-core/cellranger/count/main.nf#L30

also happy to change, but then we should change `cellranger count` too

modules/nf-core/cellranger/mkvdjref/main.nf
```diff
@@ -0,0 +1,38 @@
...
+        --genome=$reference_name \\
+        --fasta=$fasta \\
+        --genes=$gtf
```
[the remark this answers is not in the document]
klkeys: the online documentation doesn't list any other arguments, but it looks like a few (advanced) ones do exist

I'll add `$args` just in case

modules/nf-core/cellranger/mkvdjref/main.nf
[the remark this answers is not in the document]
klkeys: also need to update the process title:
```
process CELLRANGER_MKREF {
```

should be
```
process CELLRANGER_MKVDJREF {
```

modules/nf-core/cellranger/vdj/main.nf
klkeys: I'm inclined to still update `cellranger count` with `${meta.gem} --> ${meta.id}` since the latter is still available for that nf-core/scrnaseq workflow

maybe one of @grst @apeltzer can give a secondary opinion? 🙏🏼

grst: `gem` is used if a sample was sequenced across multiple flow cells. In that case cellranger expects multiple input folders. Each folder has a separate `id`, but the same `gem`. 

I think this could/should be solved differently by renaming input files, see also issues
https://github.com/nf-core/scrnaseq/issues/147
https://github.com/nf-core/scrnaseq/issues/195

grst:
1) Agree on using the `cellranger mkvdjref` with fasta/gtf. The `fetch-imgt` script requires an internet connection and in general our workflows should run offline. People can still run it independently and specify it as reference

2) Is there still any reason to run `cellranger vdj` over `cellranger multi`? For paired vdj+gex, `cellranger multi` is the recommended way. If there's only VDJ, `cellranger multi` can also be used. That way we would save one additional alignment route in the scrnaseq pipeline:
image

klkeys:
@grst no reason to prefer `cellranger vdj` over `cellranger multi`, I started with V(D)J simply because it met a need

but I'll start with `cellranger multi` once this is done

in case it matters, `cellranger multi` is different from `count` and `vdj` in that it uses a strange CSV-like input format

it will be some work (or maybe unwise to even try) to shoehorn into current nf-core samplesheet standards 😕

flipside is that a `cellranger multi` module wraps a fundamentally simple command:
```bash
    cellranger multi --id ${meta.id} --csv ${meta.cellranger_multi_csv_file}
````

and then `cellranger` handles all the other complexities (gene expression, feature barcode, cell multiplexing, V(D)J profiling, etc.)

**Comment by grst** [comment]
> in case it matters, cellranger multi is different from count and vdj in that it uses a strange CSV-like input format it will be some work (or maybe unwise to even try) to shoehorn into current nf-core samplesheet standards confused

I don't see a reason why we can't generate these files on a per-sample basis. In fact, @ErikaKvalem  has started to play around with this. It would, of course, require an extension of our samplesheet, but there was some discussion (and a proposed solution) here: https://github.com/nf-core/scrnaseq/issues/174#issuecomment-1308476027

File: modules/nf-core/cellranger/vdj/main.nf (hunk 1 above)
Comment (klkeys) [reply modules/nf-core/cellranger/vdj/main.nf]: ok @grst fair to say that we should leave `${meta.gem}` alone for now?

we can always update the modules later once consensus is reached about input file names

File: modules/nf-core/cellranger/vdj/main.nf (hunk 2 above)
Comment (klkeys) [reply modules/nf-core/cellranger/vdj/main.nf]: went ahead and removed `sample-` prefix 👍🏼

**Comment by klkeys** [comment]
added requested test for `cellranger mkvdjref` 👍🏼

**Review by grst — APPROVED** [review]
LGTM!

**Review by SPPearce — CHANGES_REQUESTED** [review]
The files look file, but two things:

1. The tests are not running, because you haven't added the modules to the `tests/config/pytest_modules.yml`. This should have been done by the nf-core modules create tool, did you not use that?
2. The general consensus is that we should not require additional fields in the meta map. I don't see why this is `meta.gem` not just `meta.id`.

**Review by jfy133 — COMMENTED** [review]
Hi @klkeys 

FYI we don't allow 'custom' per-module meta fields, as this should be left up to piepline developers how to encode such metadata (and can be inserted via `${meta.args}`).

This wasn't very well documented previously, soI've just now made a PR into the module guidelines to account for this.

Where you've used `meta.gem` I think what you want to use is [`${prefix}`](https://github.com/nf-core/tools/blob/99961bedab1518f592668727a4d692c4ddf3c336/nf_core/module-template/modules/main.nf#L52) which will need an ddition to the script variables

**Comment by klkeys** [comment]
@SPPearce apologies for the oversight, I updated `pytest_modules.yml`

in doing so, I noticed that the file refers to a `cellranger/gtf` module
https://github.com/nf-core/modules/blob/0591cad3d725d5c21337f72e638507abf709f75e/tests/config/pytest_modules.yml#L599

but that module doesn't seem to exist anywhere

my guess: it's an artifact from when `cellranger/mkgtf` was added

is it safe to clean up?

**Comment by jfy133** [comment]
> hello @jfy133 , I'll repeat some items that appeared in the conversation about this PR:
> 
> * the `${meta.gem}` variable appears in the previously approved `cellranger count` module, I merely mimicked it for consistency:
>   https://github.com/nf-core/modules/blob/0591cad3d725d5c21337f72e638507abf709f75e/modules/nf-core/cellranger/count/main.nf#L2
> * nf-core/scrnaseq "uses" `${meta.gem}` but in a way that's easily fixed:
>   [added cellranger vdj/mkvdjref modules, with module test for vdj #3033 (comment)](https://github.com/nf-core/modules/pull/3033#discussion_r1149682037)
> * I have no problem changing `${meta.gem}` to `${meta.id}`, and in fact did so but later reverted it given the conversation ([85caa77](https://github.com/nf-core/modules/commit/85caa77bdda2efec22d0a4b350ca51410b9ba236))
> * but if it's `${meta.id}` here and `${meta.gem}` in other `cellranger` modules, then the inconsistency will probably lead to future headaches if it's not fixed now
> 
> I'm happy to do the fix, and to propagate the change to the other `cellranger` modules and nf-core/scrnaseq, so long as others agree to it 👍🏼



> hello @jfy133 , I'll repeat some items that appeared in the conversation about this PR:
> 
> * the `${meta.gem}` variable appears in the previously approved `cellranger count` module, I merely mimicked it for consistency:
>   https://github.com/nf-core/modules/blob/0591cad3d725d5c21337f72e638507abf709f75e/modules/nf-core/cellranger/count/main.nf#L2
> * nf-core/scrnaseq "uses" `${meta.gem}` but in a way that's easily fixed:
>   [added cellranger vdj/mkvdjref modules, with module test for vdj #3033 (comment)](https://github.com/nf-core/modules/pull/3033#discussion_r1149682037)
> * I have no problem changing `${meta.gem}` to `${meta.id}`, and in fact did so but later reverted it given the conversation ([85caa77](https://github.com/nf-core/modules/commit/85caa77bdda2efec22d0a4b350ca51410b9ba236))
> * but if it's `${meta.id}` here and `${meta.gem}` in other `cellranger` modules, then the inconsistency will probably lead to future headaches if it's not fixed now
> 
> I'm happy to do the fix, and to propagate the change to the other `cellranger` modules and nf-core/scrnaseq, so long as others agree to it 👍🏼

Sorry I missed the earlier discussions! I was asked to chime in about this particular PR on slack ;)

I think that indeed switching all to `id` would be preferred from a modules standardisation point of view.  Even if scrnaseq currently relies on `meta.gem`, we shouldn't hardcode a pipeline-specific meta into a module (that was an earlier mistake I guess.., modules should be designed to serve _all_ pipelines not a single one ;)).

In short: it would be really appreciated if you could fix this across all related modules, and then the `scrnaseq` team can adjust this in the pipeline accordingly.

File: modules/nf-core/cellranger/mkfastq/main.nf [hunk 5]
```diff
@@ -14,18 +14,18 @@ process CELLRANGER_MKFASTQ {
     path csv
 
     output:
-    path "versions.yml", emit: versions
-    path "${bcl.getSimpleName()}/outs/fastq_path/*.fastq.gz"  , emit: fastq
+    path "${bcl.getSimpleName()}/outs/fastq_path/*.fastq.gz", emit: fastq
```
jfy133, line 17: Is there a reason why you're not using `meta.id` (rather than `.getSimpleName`)

modules/nf-core/cellranger/mkvdjref/main.nf
```diff
@@ -0,0 +1,39 @@
...
+
+    // Exit if running this module with -profile conda / -profile mamba
+    if (workflow.profile.tokenize(',').intersect(['conda', 'mamba']).size() >= 1) {
+        exit 1, "CELLRANGER_MKREF module does not support Conda. Please use Docker / Singularity / Podman instead."
+    }
+
+    input:
+    path fasta
+    path gtf
+    val reference_name
```
jfy133, line 15: Can you confirm all mandaotry but also optional inputs are covered here (not familiar with the tool)

modules/nf-core/cellranger/mkvdjref/main.nf
```diff
@@ -0,0 +1,39 @@
...
+
+    container "nfcore/cellranger:7.1.0"
+
+    // Exit if running this module with -profile conda / -profile mamba
+    if (workflow.profile.tokenize(',').intersect(['conda', 'mamba']).size() >= 1) {
+        exit 1, "CELLRANGER_MKREF module does not support Conda. Please use Docker / Singularity / Podman instead."
+    }
+
+    input:
+    path fasta
```
jfy133, line 13: Is there a possibiliy a pipeline may allow multiple references? If so I would suggest adding `meta` here (and you can then change the `tag` to meta.id)

modules/nf-core/cellranger/vdj/main.nf
```diff
@@ -0,0 +1,54 @@
...
+    output:
+    tuple val(meta), path("${meta.id}/outs/*"), emit: outs
+    path "versions.yml"                       , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def sample_arg = meta.samples.unique().join(",")
```
jfy133, line 25: Should be removed, the `--sample` flag can be presumably supplied directly via `ext.args` with wahtever meta field the pipeline deve may want to call it.

modules/nf-core/cellranger/vdj/meta.yml
```diff
@@ -0,0 +1,43 @@
...
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. [ id:'test', single_end:false ]
+  - reads:
+      type: file
+      description: |
+        List of input FastQ files of size 1 and 2 for single-end and paired-end data,
+        respectively.
```
jfy133, line 27: I see there is no if/else stratement based on meta for this. If there is a specific naming scheme taht must be  uysed for the tool to pick up the pairs I would add this here.

modules/nf-core/cellranger/vdj/meta.yml
```diff
@@ -0,0 +1,43 @@
...
+  - reference:
+      type: directory
+      description: Folder containing all the reference indices needed by Cell Ranger
+output:
+  - outs:
+      type: file
+      description: Files containing the outputs of Cell Ranger
+      pattern: "${meta.id}/outs/*"
```
jfy133, line 35: This isn't so informative as it's not hte asctual files that get output. Why not something like `*.{porefix1,prefix2,prefix3}` etc.?

tests/modules/nf-core/cellranger/mkvdjref/test.yml
```diff
@@ -0,0 +1,9 @@
+- name: cellranger mkvdjref test_cellranger_mkvdjref
+  command: nextflow run ./tests/modules/nf-core/cellranger/mkvdjref -entry test_cellranger_mkvdjref -c ./tests/config/nextflow.config -c ./tests/modules/nf-core/cellranger/mkvdjref/nextflow.config
+  tags:
+    - cellranger/mkvdjref
+    - cellranger
+  files:
+    - path: output/cellranger/homo_sapiens_chr22_reference/fasta/regions.fa
```
jfy133, line 7: Can add a `contains` with a string opf teh fasta entry

tests/modules/nf-core/cellranger/vdj/main.nf
```diff
@@ -0,0 +1,27 @@
...
+               file("https://github.com/nf-core/test-datasets/raw/modules/data/genomics/homo_sapiens/illumina/10xgenomics/cellranger_vdj/subsampled_sc5p_v2_hs_B_1k_b_fastqs/subsampled_sc5p_v2_hs_B_1k_b_S1_L001_R2_001.fastq.gz", checkIfExists: true)
+        ]
+    ]
+
+    reference_json = file("https://github.com/nf-core/test-datasets/raw/modules/data/genomics/homo_sapiens/illumina/10xgenomics/cellranger_vdj/refdata-cellranger-vdj-GRCh38-alts-ensembl-5.0.0/reference.json", checkIfExists: true)
+    reference_fasta = file("https://github.com/nf-core/test-datasets/raw/modules/data/genomics/homo_sapiens/illumina/10xgenomics/cellranger_vdj/refdata-cellranger-vdj-GRCh38-alts-ensembl-5.0.0/fasta/regions.fa", checkIfExists: true)
+    reference_suppfasta = file("https://github.com/nf-core/test-datasets/raw/modules/data/genomics/homo_sapiens/illumina/10xgenomics/cellranger_vdj/refdata-cellranger-vdj-GRCh38-alts-ensembl-5.0.0/fasta/supp_regions.fa", checkIfExists: true)
```
jfy133, line 17: These should be referred to via the nested list, as this is official nf-core/moduels data, rather than generic 'delete me' stuff.

You may have to update test_config.conf (I think it's called)

tests/modules/nf-core/cellranger/vdj/main.nf
```diff
@@ -0,0 +1,27 @@
...
+
+    reference_json.copyTo("${workDir}/vdj_reference/reference.json")
+    reference_fasta.copyTo("${workDir}/vdj_reference/fasta/regions.fa")
+    reference_suppfasta.copyTo("${workDir}/vdj_reference/fasta/supp_regions.fa")
```
jfy133, line 21: This looks scary, why is it not staged in the module?

jfy133 (dismissed):

modules/nf-core/cellranger/mkfastq/main.nf
klkeys: I didn't create the `cellranger/mkfastq` module, but I will update it now

klkeys: in case it matters @jfy133 , it doesn't look like metadata are passed into this module (no `meta` map is used here)

that might explain why `$meta.id` doesn't appear in the code

modules/nf-core/cellranger/mkvdjref/main.nf
klkeys: yes, any extra arguments are passed to `cellranger` via `task.ext.args`

tests/modules/nf-core/cellranger/vdj/main.nf
klkeys: the awkward staging here is to enforce the directory structure that `cellranger vdj` expects from its VDJ references

the output of `CELLRANGER_MKVDJREF` ref will already have this structure, but user-supplied references may not

and in this case, explicitly staging the three requisite files felt more transparent than trying to recursively trawl the [VDJ reference directory](https://github.com/nf-core/test-datasets/tree/modules/data/genomics/homo_sapiens/illumina/10xgenomics/cellranger_vdj) in the nf-core/test-datasets repo via HTTP

the module simply stages the directory (here `$workdir/vdj_reference`) which is much cleaner and mimics the behavior in `CELLRANGER_COUNT`

modules/nf-core/cellranger/vdj/meta.yml
klkeys: added pattern descriptor to match [Cellranger expected FASTQ format](https://support.10xgenomics.com/single-cell-vdj/software/pipelines/latest/using/fastq-input):
```
${Sample_Name}_S1_L00${Lane_Number}_${I1,I2,R1,R2}_001.fastq.gz
```

tests/modules/nf-core/cellranger/vdj/main.nf
klkeys: updated `test_config.conf` accordingly

`CELLRANGER_VDJ` tests now call test data from `params.test_data`

modules/nf-core/cellranger/vdj/main.nf
klkeys: note that this will affect downstream pipelines like [nf-core/scrnaseq(https://github.com/nf-core/scrnaseq/blob/c6ec2175a8f5d593c00136835c45c6291821fe11/subworkflows/local/align_cellranger.nf#L37)

however, in that case, `meta.samples` is simply a copy of `meta.id`, so it's probably safe to remove

to preclude `cellranger` loading extraneous FASTQs in the local directory (from `--fastqs=.`), the wisest change is probably to use `--sample=$meta.id`

that will ensure that Cellranger loads only the FASTQs relevant to the current sample in `meta.id`

tests/modules/nf-core/cellranger/mkvdjref/test.yml
klkeys: no that's not possible

the file is empty since the nf-core/test-datasets chr22 FASTQs don't have VDJ sequences

but the MD5 sum should be stable in that case

modules/nf-core/cellranger/mkfastq/main.nf
klkeys: ok after some thought it seems unwise to change this since it's unclear if `meta` is even necessary here

it seems like this module is usually run prior to executing any nf-core workflows anyways

to wit, nf-core/scrnaseq with the Cellranger aligner starts from FASTQs, not from BCLs

modules/nf-core/cellranger/vdj/meta.yml
klkeys: a reasonable compromise is to document whatever 10X thinks is an [average case output file set](https://support.10xgenomics.com/single-cell-vdj/software/pipelines/latest/using/vdj#success)

klkeys: ok it occurred to me that the `meta.yml` should match the code in the module

the code output is two lines:
```
tuple val(meta), path("${meta.id}/outs/*"), emit: outs
path "versions.yml" , emit: versions
```
and tabulating all possible consistent + optional outputs here doesn't seem wise

I'm going to punt and add a note to check official 10X docs for a full output file list

will do for both `cellranger count` and `cellranger vdj` 👍🏼

tests/modules/nf-core/cellranger/mkvdjref/test.yml
klkeys: MD5 sum is indeed stable but empty files enrage the linter, so current approach is to just check for file existence at `path:`

klkeys (commented):
I think that I have fixed all outstanding comments (so far)

cellranger/vdj passed linting and testing on local machine

should be ready for merging 👍🏼

modules/nf-core/cellranger/mkvdjref/main.nf
jfy133: > the generation of multiple references from separate organisms would simply make separate calls to cellranger mkvdjref

Yes but tha'ts exactly my point. How would you track which output comes from which call?

I'll leave it for now as I know you've been working on this for a while - but a reviewer may bring it up again in the future.

modules/nf-core/cellranger/vdj/main.nf
jfy133: That's fine - the pipeline needs to serve the module in a way (As the modules are what are standardised)!

modules/nf-core/cellranger/vdj/meta.yml
jfy133: Fair enough!

jfy133 (approved):
I think that's addressed all my comments sufficiently for now :+1:

SPPearce (changes requested):
I've made changes with respect to the prefix, hopefully will all pass and then I'm happy.

modules/nf-core/cellranger/mkvdjref/main.nf
```diff
@@ -0,0 +1,39 @@
+process CELLRANGER_MKVDJREF {
+    tag "$fasta"
+    label 'process_high'
+
+    container "nfcore/cellranger:7.1.0"
```
adamrtalbot, line 5:
```suggestion
    container "docker.io/nfcore/cellranger:7.1.0"
```

https://github.com/nf-core/modules/pull/3344

modules/nf-core/cellranger/vdj/main.nf
```diff
@@ -0,0 +1,54 @@
+process CELLRANGER_VDJ {
+    tag "${meta.id}"
+    label 'process_high'
+
+    container "nfcore/cellranger:7.1.0"
```
adamrtalbot, line 5:
```suggestion
    container "docker.io/nfcore/cellranger:7.1.0"
```

https://github.com/nf-core/modules/pull/3344

adamrtalbot (changes requested):
Could you use the full URI for the docker images please?

SPPearce:
@adamrtalbot , I have just updated to use the full path in all the modules.

SPPearce (approved):

SPPearce:
Right, I'm happy to go now. @adamrtalbot , if that was your only objection I think we are happy to merge?

jfy133 (approved):

adamrtalbot (approved):
apologies for being slow!

klkeys:
oof this sparked a lot of discussion 😵

thank you for the additional eyes and edits @SPPearce @jfy133 @adamrtalbot 

I will proceed with merging now 👍🏼

SPPearce:
Thanks for your hard work and patience @klkeys

klkeys:
premature celebration? it looks like nf-core linting failed on `cellranger/mkref` and `cellranger/mkvdjref` 😣 

not sure how this escaped me, but best not dwell on the past, I'll correct this in #3229

SPPearce:
That's strange, why did the linting tests not run? All the tests that ran were passing.

klkeys:
no clue 😣 

see here: https://github.com/nf-core/modules/actions/runs/4832726781/jobs/8611869687

I've fixed it in #3229

PR #3036, opened 2023-03-15
Add Bedtools unionbedg

ekushele:
## PR checklist

Closes #XXX 

- [v] This comment contains a description of changes (with reason).
- [v] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [v] Remove all TODO statements.
- [v] Emit the `versions.yml` file.
- [v] Follow the naming conventions.
- [v] Follow the parameters requirements.
- [v] Follow the input/output options guidelines.
- [v] Add a resource `label`
- [v] Use BioConda and BioContainers if possible to fulfil software requirements.


modules/nf-core/bedtools/unionbedg/main.nf
```diff
@@ -0,0 +1,42 @@
...
+
+    conda "bioconda::bedtools=2.30.0"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/bedtools:2.30.0--h7d7f7ad_2':
+        'quay.io/biocontainers/bedtools:2.30.0--h7d7f7ad_2' }"
+
+    input:
+    tuple val(meta), path(bedgraph, stageAs: "?/*")
+    path chrom_sizes
+    val(empty)
```
nvnieuwk, line 13: `val(empty)` is for an optional value. These shouldn't be in the main inputs but should be given to the module through `ext.args`. 
P.S. you forgot the meta for `chrom_sizes` again ;)

modules/nf-core/bedtools/unionbedg/main.nf
```diff
@@ -0,0 +1,42 @@
...
+    tag "$meta.id"
+    label 'process_low'
+
+    conda "bioconda::bedtools=2.30.0"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/bedtools:2.30.0--h7d7f7ad_2':
+        'quay.io/biocontainers/bedtools:2.30.0--h7d7f7ad_2' }"
+
+    input:
+    tuple val(meta), path(bedgraph, stageAs: "?/*")
```
nvnieuwk, line 11: Why do you use stageAs here?

modules/nf-core/bedtools/unionbedg/main.nf
```diff
@@ -0,0 +1,42 @@
...
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def sizes = chrom_sizes ? "-g ${chrom_sizes}" : ''
+    def empty = empty ? "-empty" : ''
+    if ("$bedgraph" == "${prefix}.bed") error "Input and output names are the same, use \"task.ext.prefix\" to disambiguate!"
```
nvnieuwk, line 27: I think it would be better to add a loop here to check every file separately. Here it checks all files in a string (which will never be the same as the output file)

nvnieuwk (changes requested):
Hi, good job. Here are some comments :)

modules/nf-core/bedtools/unionbedg/main.nf
ekushele: I didnt forget... it is an optional parameter, and when it was with meta I couldn't pass an empty value as []

modules/nf-core/bedtools/unionbedg/main.nf
ekushele: because without it-tests fail

ekushele: The unionbedg gets multiple files, likr samtools merge, I copied from there

modules/nf-core/bedtools/unionbedg/main.nf
ekushele: Or can I just remove this "if" statement?

modules/nf-core/bedtools/unionbedg/main.nf
ekushele: I removed the `chorm_sizes`, it's not mandatory, only if one specifies "-empty"

nvnieuwk: You should include all optional files in the input too (otherwise you have no way to give them when needed). If the output is optional you can use `[[], []]` as input to that channel :)

modules/nf-core/bedtools/unionbedg/main.nf
nvnieuwk: Aah I get it, we had a lot of discussions in the nf-core community about `stageAs` and the consensus is to avoid it as much as possible and make sure all your input files are names differently :). You can always rename the input files using the `.collectFile()` operator

modules/nf-core/bedtools/unionbedg/main.nf
nvnieuwk: I wouldn't do that, since this will give a warning before the module has been run, saving some precious compute time in the case of an error :)

ekushele: So how should I loop?

nvnieuwk: I think something like this should work:

```groovy
bedgraph.collect { if ("$it" == "${prefix}.bed") error "$it has the same name as the output, use \"task.ext.prefix\" to disambiguate!" }
```

Either that or with `.each { }` instead of `.collect { }`. I don't remember which one it was :)

modules/nf-core/bedtools/unionbedg/main.nf
ekushele: There is only one bedgraph file in test-datasets so @FriederikeHanssen  suggested I will use the same file twice... Seems problematic.

So can I add another bedGraph file to test-datasets? Just the same bedGraph with different values or even the same bedGraph with a different name?

nvnieuwk: I would just use the same Bedgraph file and rename it :) That's the least effort solution

ekushele: I created a PR: https://github.com/nf-core/test-datasets/pull/811
And updated the module to use the new bedGraph, but all tests fails because this file doesn't exists yet in the main test-datasets repository, so I'll be happy if you can review this PR

ekushele: OK, now all tests passed.
Thank you!

tests/config/test_data.config
```diff
@@ -76,6 +76,7 @@ params {
                 test_methylated_2_fastq_gz                     = "${params.test_data_base}/data/genomics/sarscov2/illumina/fastq/test.methylated_2.fastq.gz"
 
                 test_bedgraph                                  = "${params.test_data_base}/data/genomics/sarscov2/illumina/bedgraph/test.bedgraph"
+                test2_bedgraph                                 = "${params.test_data_base}/data/genomics/sarscov2/illumina/bedgraph/test2.bedgraph"
```
nvnieuwk, line 79:
```suggestion
```

nvnieuwk (approved):
A few small changes and then you are good to go 🥳

PR #3038, opened 2023-03-15
fix: redirect kraken2 output to /dev/null

Midnighter:
When the `--output` option is not given to kraken2, it prints out information for every classified read. This makes both the `.command.out` and `.command.log` grow to several GBs in size (depending on input). We ignore this output by redirecting to `/dev/null`. If a user is interested in this information they presumably chose the option `save_reads_assignment` anyway.

Prompted by https://github.com/nf-core/taxprofiler/issues/268


prototaxites (approved):
LGTM - did you double-check that .command.out/.command.log are fine? I don't have time to check myself!

Midnighter:
Depends what you mean by fine exactly. The `.command.out` is empty and the `.command.err` and `.command.log` are identical - containing the processing speed and percentages (U, R). So to me that's the expected outcome.

PR #3066, opened 2023-03-21
New Module: Windowmasker

### LINKED ISSUE #3048: new module: windowmasker/mk_counts

DLBPointon:
## PR checklist

Closes #3048

This PR adds windowmasker, this module contains both windowmasker mk_counts and ustat.

- [X] Add a resource `label`


modules/nf-core/windowmasker/main.nf
```diff
@@ -0,0 +1,44 @@
...
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args1   = task.ext.args1    ?: ""
+    def args2   = task.ext.args2    ?: ""
+    def prefix  = task.ext.prefix   ?: "${meta.id}"
+    def outfmt  = task.ext.outfmt   ?: "interval"
+    output  = "${prefix}.${outfmt}"
+
+    """
```
nvnieuwk, line 27: Can you add the `-mem` option to your commands? (You can access the specified memory like this: `task.memory.toMega()`

modules/nf-core/windowmasker/main.nf
```diff
@@ -0,0 +1,44 @@
+process WINDOWMASKER {
+    tag "$meta.id"
+    label 'process_medium'
```
nvnieuwk, line 3: Is this high of a label necessary for a module that runs on one thread?

modules/nf-core/windowmasker/main.nf
```diff
@@ -0,0 +1,44 @@
...
+    conda "bioconda::blast=2.13.0"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/blast:2.13.0--hf3cf87c_0':
+        'quay.io/biocontainers/blast:2.13.0--hf3cf87c_0' }"
+
+    input:
+    tuple val(meta), path(ref)
+
+    output:
+    tuple val(meta), path("${output}")  , emit: wm_intervals
```
nvnieuwk, line 14:
```suggestion
    tuple val(meta), path("${output}")  , emit: intervals
```
just intervals is good enough, no need to make the output names too complicated :)

modules/nf-core/windowmasker/main.nf
```diff
@@ -0,0 +1,44 @@
...
+    path "versions.yml"                 , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args1   = task.ext.args1    ?: ""
+    def args2   = task.ext.args2    ?: ""
+    def prefix  = task.ext.prefix   ?: "${meta.id}"
+    def outfmt  = task.ext.outfmt   ?: "interval"
```
nvnieuwk, line 24:
```suggestion
    def outfmt  = task.ext.suffix   ?: "interval"
```
We have a designated suffix variable for this, but this should only be used if the output extension cannot be derived from the arguments or other inputs. So does the module determine the output format based on the name of the output file or based on an argument?

modules/nf-core/windowmasker/main.nf
```diff
@@ -0,0 +1,44 @@
...
+    windowmasker -mk_counts \\
+        $args1 \\
+        -in ${ref} \\
+        -out ${prefix}.txt
+
+    windowmasker -ustat \\
+        ${prefix}.txt \\
+        $args2 \\
+        -in ${ref} \\
+        -out ${output}
```
nvnieuwk, line 37: It's preffered to limit a module to only one process, unless the tool will always be run like this

modules/nf-core/windowmasker/main.nf
```diff
@@ -0,0 +1,44 @@
...
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        windowmasker: \$(windowmasker -version-full 2>&1)
```
nvnieuwk, line 41: The version should only be the semantic form, like 1.0.0 for example (and not all the other text that `-version-full` prints out)

modules/nf-core/windowmasker/meta.yml
```diff
@@ -0,0 +1,39 @@
...
+  - ref:
+      type: file
+      description: An input nucleotide fasta file.
+output:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. [ id:'test', single_end:false ]
+  - wm_intervals:
```
nvnieuwk, line 28:
```suggestion
  - intervals:
```

tests/modules/nf-core/windowmasker/nextflow.config
nvnieuwk, line 1: Can you add some more tests here? Also the CI tests aren't running because the module hasn't been added to `tests/config/pytest_modules.yml`. If you use the `nf-core` tooling to create your module all this will be done for you :)

nvnieuwk (changes requested):

modules/nf-core/windowmasker/main.nf
nvnieuwk: Awesome!

modules/nf-core/windowmasker/main.nf
DLBPointon: It is based on an input argument
```
-outfmt <String, `fasta', `interval', `maskinfo_asn1_bin',
                  `maskinfo_asn1_text', `maskinfo_xml', `seqloc_asn1_bin',
                  `seqloc_asn1_text', `seqloc_xml'>
```

hence why making its own ext argument in the first place

nvnieuwk: To keep all modules compatible we have decided to not allow this in `nf-core`. Only a very select few `task.ext.args` are allowed (see [here](https://nf-co.re/docs/contributing/modules#what-are-the-ext-propertieskeys). I suggested suffix because that one can also be used when the output format is based on the type specified by the output file.

modules/nf-core/windowmasker/main.nf
DLBPointon: I assumed it was multi, done.

modules/nf-core/windowmasker/main.nf
DLBPointon: Would this be as simple as ` def memory  = task.memory.toMega()`?

nvnieuwk: that should work, in GATK modules they also add a check to see if `task.memory` is set. That could also be useful here I think

tests/modules/nf-core/windowmasker/nextflow.config
DLBPointon: I've added them manually to the pytest_modules.yml

nvnieuwk: Strange, it's been in there for a long time. What version are you using?

modules/nf-core/windowmasker/main.nf
DLBPointon: That makes sense. 
So in that case, the user would need to set both `ext.args = -outfmt interval` and `ext.suffix = interval`?

tests/modules/nf-core/windowmasker/nextflow.config
DLBPointon: Locally, we have 2.4.1 installed. This was the version we created our pipeline on so have stuck with it until we have a full release.

modules/nf-core/windowmasker/main.nf
DLBPointon: Done, changed from `-version-full 2>&1` to `-version-full | head -n 1`

modules/nf-core/windowmasker/main.nf
nvnieuwk: In this case I think it would be best to write some code that checks which outfmt is given in ext.args and use that one as extension. That would be the most user friendly solution I think

tests/modules/nf-core/windowmasker/nextflow.config
nvnieuwk: It's best to update to newer versions everytime a new version comes out to minimize the work you'll have to do in the end :) (It's gonna be a huge job now with the template merges 😁 )

DLBPointon: Oh no, well i'll prioritize it after this module 👍

modules/nf-core/windowmasker/mk_counts/main.nf
```diff
@@ -0,0 +1,43 @@
...
+
+    script:
+    def args    = task.ext.args     ?: ""
+    def prefix  = task.ext.prefix   ?: "${meta.id}"
+
+    def memory = 3072
+    if (!task.memory) {
+        log.info '[WINDOWMASKER: MK_COUNTS] Available memory not known - defaulting to 3GB. Specify process memory requirements to change this.'
+    } else {
+        memory = (task.memory.toMega()*0.8).intValue()
```
nvnieuwk, line 28: You can use the full memory here. The subsetting here is specifically for tools using Java :)

nvnieuwk: Unless it makes some sense to limit here too (I don't really know this tool)

modules/nf-core/windowmasker/mk_counts/main.nf
```diff
@@ -0,0 +1,43 @@
...
+        $args \\
+        -mem ${memory} \\
+        -in ${ref} \\
+        -out ${prefix}.txt
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        windowmasker: \$(windowmasker -version-full | head -n 1)
+    END_VERSIONS
+    """
```
nvnieuwk, line 42: Can you also add a stub section to both modules?

nvnieuwk (changes requested):

modules/nf-core/windowmasker/main.nf
DLBPointon: In that case, would something like this be ok?
```
    if (task.ext.args.contains('-outfmt')) {
        outfmt = task.ext.args.split('-outfmt')[1] // relies on -outfmt being the last arg in ext.args
    } else {
        log.info('WINDOWMASKER USTAT: outfmt not specified defaulting to "interval"')
        task.ext.args = task.ext.args + "-outfmt interval"
        outfmt = "interval"
    }
```

modules/nf-core/windowmasker/mk_counts/main.nf
DLBPointon: It's a C tool, so you're right. No point in limiting it.
Done

modules/nf-core/windowmasker/main.nf
nvnieuwk: Here is another way I've done it before for `bcftools`: https://github.com/nf-core/modules/blob/e73f327f6ef9f77a8e23dcf90340caa01f02954e/modules/nf-core/bcftools/convert/main.nf#L31-L35. This is indeed a bit more redundant but doesn't force the user to specifically place their outfmt at the end of args :) (yay for flexibility)

nvnieuwk: Otherwise using a regex is also a possibility :)

modules/nf-core/windowmasker/mk_counts/main.nf
DLBPointon: Done. Stub is just empty output right?

```
    stub:
    """
    mkdir windowmasker
    touch windowmasker/test.inteval
    """

    stub:
    """
    mkdir windowmasker
    touch windowmasker/test.txt
    """
```

nvnieuwk: Yes indeed!!! We also try to mimic the output generated with all the options given to the module (so in your case change the output extension for example)

modules/nf-core/windowmasker/main.nf
nvnieuwk: Perfect! Thanks!

modules/nf-core/windowmasker/mk_counts/main.nf
nvnieuwk: Can you also add the versions.yml creation to the stub? This can be the full code as it is in the main module (since it's a fairly simple calculation)

DLBPointon: Done. Thanks for the help over on Slack.

nvnieuwk: You are welcome! 🥳

nvnieuwk (approved):
All good except for this small comment. You can merge if you fixed that one 🚀

modules/nf-core/windowmasker/ustat/main.nf
```diff
@@ -0,0 +1,76 @@
...
+    output  = "${prefix}.${outfmt}"
+    """
+    touch ${prefix}.fasta
+    touch ${prefix}.interval
+    touch ${prefix}.maskinfo_asn1_bin
+    touch ${prefix}.maskinfo_asn1_text
+    touch ${prefix}.maskinfo_xml
+    touch ${prefix}.seqloc_asn1_bin
+    touch ${prefix}.seqloc_asn1_text
+    touch ${prefix}.seqloc_xml
```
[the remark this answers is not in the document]
DLBPointon: Just done that now. Thanks for the help!

Issue #3143, opened 2023-03-27
Allow decompression utilities to output files channels


pinin4fjords:
### Is your feature request related to a problem? Please describe

Right now it's difficult to use UNTAR and UNZIP to create a channel of files, you essentially need a separate process to hack your way. 

### Describe the solution you'd like

Can we just have a separate channel to emit the files, as well as the one that emits the directory? e.g untar would emit a `tuple(meta, [files])`, which we could transpose to a channel with a file per item.

### Describe alternatives you've considered

Hacky process-based solutions.

PR #3144, opened 2023-03-27
Untar files

pinin4fjords:
In this PR I propose an implementation of @nvnieuwk 's [suggestion](https://nfcore.slack.com/archives/CE6SDBX2A/p1674465188194259?thread_ts=1674461395.717019&cid=CE6SDBX2A) of separate channels in decompression utilities to facilitate the channeling of individual files from decompressed archives. 

## PR checklist

Closes https://github.com/nf-core/modules/issues/3143

- [ ] Add a resource `label`

### DIFF
--- a/modules/nf-core/untar/main.nf
+++ b/modules/nf-core/untar/main.nf
@@ -11,8 +11,9 @@ process UNTAR {
     tuple val(meta), path(archive)
 
     output:
-    tuple val(meta), path("$prefix"), emit: untar
-    path "versions.yml"             , emit: versions
+    tuple val(meta), path("$prefix")        , emit: untar
+    tuple val(meta), path("${prefix}/**")   , emit: files
+    path "versions.yml"                     , emit: versions
 
     when:
     task.ext.when == null || task.ext.when
--- a/modules/nf-core/unzip/main.nf
+++ b/modules/nf-core/unzip/main.nf
@@ -11,8 +11,9 @@ process UNZIP {
     tuple val(meta), path(archive)
 
     output:
-    tuple val(meta), path("${archive.baseName}/"), emit: unzipped_archive
-    path "versions.yml"                          , emit: versions
+    tuple val(meta), path("${archive.baseName}/")  , emit: unzipped_archive
+    tuple val(meta), path("${archive.baseName}/**"), emit: files
+    path "versions.yml"                            , emit: versions
 
     when:
     task.ext.when == null || task.ext.when


SPPearce:
Looks fine to me, but I strongly disagree with the names `args3` and `args2` to control the behaviour.

modules/nf-core/untar/main.nf
```diff
@@ -11,16 +11,19 @@ process UNTAR {
...
+    tuple val(meta), path("${prefix}_files/**") , emit: files, optional: true
+    path "versions.yml"                         , emit: versions
 
     when:
     task.ext.when == null || task.ext.when
 
     script:
     def args  = task.ext.args ?: ''
     def args2 = task.ext.args2 ?: ''
+    def args3 = task.ext.args3 ?: ''
```
nvnieuwk, line 24: I think this could be a boolean input. This seems a bit of weird workaround in my opinion. But again this is a difficult situation, we should get some more maintainer and core team eyes on it 😁

pinin4fjords:
> Looks fine to me, but I strongly disagree with the names `args3` and `args2` to control the behaviour.

@SPPearce me too- but this is the current convention, right?

modules/nf-core/untar/main.nf
pinin4fjords: Thanks! Could you clarify on how you think that boolean would be supplied?

nvnieuwk: As an actual input value. That seems the best solution here, but it's a bit of a grey line for the guidelines 😁

pinin4fjords: Well, it's non-mandatory as currently implemented (default is not to emit the files, as per current behaviour). So I don't think it qualifies for the `val` treatment, right?

nvnieuwk: Yeah but it's not an argument for a tool either... (oh god this will probably start another discussion 😁)

pinin4fjords: I could just revert the conditionality and emit the files all the time....

nvnieuwk: That's also a possibility, would you know if nextflow resolves the output channels when the module finishes or when they are actually used?

pinin4fjords: I don't really see how 'lazy globbing' might work, so I think we have to assume files channel would get filled from the glob regardless.

nvnieuwk: Yeah true better safe than sorry

SPPearce: Blah, I really don't understand the issue with adding other options that aren't the default set. They just make the code more readable.

nvnieuwk: I agree especially in this case!

tests/modules/nf-core/unzip/test.yml
```diff
@@ -8,3 +8,18 @@
...
+  files:
+    - path: output/unzip/ncbi_taxmap/
+    - path: output/unzip/ncbi_taxmap/ncbi.map
+      md5sum: de30dbba85f9070612b632e2a5a95952
+    - path: output/unzip/ncbi_taxmap/ncbi.tre
+      md5sum: 4029dd2091c685b9a86ddd9d0d870db0
+    - path: output/unzip/ncbi_taxmap_files/ncbi.map
+      md5sum: de30dbba85f9070612b632e2a5a95952
+    - path: output/unzip/ncbi_taxmap_files/ncbi.tre
+      md5sum: 4029dd2091c685b9a86ddd9d0d870db0
```
FriederikeHanssen, line 25: check for version.yml is missing

modules/nf-core/untar_files/main.nf
```diff
@@ -0,0 +1,63 @@
...
+    tuple val(meta), path("${prefix}/**") , emit: files, optional: true
+    path "versions.yml"                   , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args  = task.ext.args ?: ''
+    def args2 = task.ext.args2 ?: ''
+    prefix    = task.ext.prefix ?: ( meta.id ? "${meta.id}" : archive.baseName.toString().replaceFirst(/\.tar$/, ""))
```
nvnieuwk, line 23: I think we can assume an id will always be in the `meta`?

modules/nf-core/untar_files/main.nf
```diff
@@ -0,0 +1,63 @@
...
+
+    """
+    mkdir $prefix
+
+    ## Ensures --strip-components only applied when top level of tar contents is a directory
+    ## If just files or multiple directories, place all in prefix
+    if [[ \$(tar -taf ${archive} | grep -o -P "^.*?\\/" | uniq | wc -l) -eq 1 ]]; then
+        tar \\
+            -C $prefix --strip-components 1 \\
```
nvnieuwk, line 32: Shouldn't this be an `ext.args` argument? No need to make this module too complex (this should be something that should be resolved on a pipeline level?)

nvnieuwk (commented):
Some small comments to discuss before I continue my review :)

modules/nf-core/untar_files/main.nf
pinin4fjords: If you say so, that was a call made by my predecessors here :-)

pinin4fjords, line 23:
```suggestion
    prefix    = task.ext.prefix ?: meta.id
```

modules/nf-core/untar_files/main.nf
```diff
@@ -0,0 +1,63 @@
...
+            $archive \\
+            $args2
+    else
+        tar \\
+            -C $prefix \\
+            -xavf \\
+            $args \\
+            $archive \\
+            $args2
+    fi
```
pinin4fjords, line 44:
```suggestion
    tar \\
        -C $prefix \\
        -xavf \\
         $args \\
         $archive \\
         $args2
```

modules/nf-core/untar_files/main.nf
```diff
@@ -0,0 +1,63 @@
...
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        untar: \$(echo \$(tar --version 2>&1) | sed 's/^.*(GNU tar) //; s/ Copyright.*\$//')
+    END_VERSIONS
+    """
+
+    stub:
+    prefix    = task.ext.prefix ?: ( meta.id ? "${meta.id}" : archive.toString().replaceFirst(/\.[^\.]+(.gz)?$/, ""))
```
pinin4fjords, line 53:
```suggestion
    prefix    = task.ext.prefix ?: "${meta.id}"
```

modules/nf-core/unzip_files/main.nf
```diff
@@ -0,0 +1,37 @@
+process UNZIP_FILES {
+    tag "$archive"
```
pinin4fjords, line 2:
```suggestion
    tag "$meta.id"
```

modules/nf-core/untar_files/main.nf
nvnieuwk: I just think it's a bit weird to do it here as it's done nowhere else

nvnieuwk: I agree :p especially since this is a universally used one

pinin4fjords: Done- and tests corrected accordingly

SPPearce:
I don't like assuming that the meta has `id`, sometimes the meta will be made on the fly to ensure that the module has something, so I feel it is probably best to have a default for that.

pinin4fjords:
> I don't like assuming that the meta has `id`, sometimes the meta will be made on the fly to ensure that the module has something, so I feel it is probably best to have a default for that.

So, so far that's @nvnieuwk making me remove the default, @SPPearce asking me to put it back. Want to come to an agreement guys?

modules/nf-core/untar_files/main.nf
```diff
@@ -0,0 +1,52 @@
...
+    conda "conda-forge::sed=4.7 bioconda::grep=3.4 conda-forge::tar=1.34"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/ubuntu:20.04' :
+        'ubuntu:20.04' }"
+
+    input:
+    tuple val(meta), path(archive)
+
+    output:
+    tuple val(meta), path("${prefix}/**") , emit: files, optional: true
```
SPPearce, line 14: Why is this optional now?

modules/nf-core/unzip_files/main.nf
```diff
@@ -0,0 +1,37 @@
...
+    conda "conda-forge::p7zip=16.02"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/p7zip:15.09--h2d50403_4' :
+        'quay.io/biocontainers/p7zip:15.09--h2d50403_4' }"
+
+    input:
+    tuple val(meta), path(archive)
+
+    output:
+    tuple val(meta), path("${prefix}/**"), emit: files, optional: true
```
SPPearce, line 14: Again, why is this optional?

SPPearce (approved):
Hmmm.
This is being added as a new pair of modules, e.g. `UNTAR_FILES`, not as subtools of `UNTAR`.
But the nf-core tooling expects `TOOL_SUBTOOL1` and `TOOL_SUBTOOL2`, but here we don't have that structure. 
Are we confident that this isn't going to mess anything up? For instance the paths for the output files from the tests for `UNTAR_FILES` are just e.g. `output/untar/versions.yml`, not untar_files.
Is that a problem?

modules/nf-core/untar_files/main.nf
pinin4fjords: sorry, a relic from before, thanks for the spot

nvnieuwk (approved):
LGTM!

pinin4fjords:
Great! Thanks for the review @nvnieuwk @SPPearce @FriederikeHanssen

PR #3178, opened 2023-03-28
Addition of new module: interproscan. Issue #3103

toniher:
Interproscan new module, referring to Issue #3103


toniher:
Repo to be cleaned

PR #3191, opened 2023-03-28
nf-test Transition CI

Emiller88:
This will allow us to migrate tests to nf-test at our own pace. We can leave an overlap as well like I've shown currently, so tests can be written piecemeal, without us needed to redo everything all at once.

This will be ready once https://github.com/askimed/nf-test/pull/80 is released 🚀


lukfor:
We just released [v0.7.3](https://github.com/askimed/nf-test/releases/tag/v0.7.3) 🚀

maxulysse:
Can you add some docs about how to create the files for the tests until we make it into tools?

tests/modules/nf-core/fastqc/main.nf.test
```diff
@@ -0,0 +1,32 @@
+nextflow_process {
+
+    name "Test Process FASTQC"
+    script "modules/nf-core/fastqc/main.nf"
+    process "FASTQC"
+    tag "fastqc"
+
+    test("Should run without failures") {
```
sateeshperi, line 8:
```suggestion
    test("test_fastqc_single_end") {
```

sateeshperi (approved):
minor suggestion...otherwise lgtm

tests/modules/nf-core/fastqc/main.nf.test
Emiller88: Good catch, but I think we can do away with the _'s now. Thoughts? I had it actually just as "Single End", I guess I regenerated it?

sateeshperi: yup...that works too

PR #3205, opened 2023-03-29
Removed -1 in CPU size specification for samtools in BWA module

lukbut:
## PR checklist

Closes https://github.com/nf-core/nf-co.re/issues/1053

Removes the -1 in the CPU size specification for the BWA module for samtools.

- [x] Add a resource `label`

### DIFF
--- a/modules/nf-core/bwa/sampe/main.nf
+++ b/modules/nf-core/bwa/sampe/main.nf
@@ -31,7 +31,7 @@ process BWA_SAMPE {
         $read_group \\
         \$INDEX \\
         $sai \\
-        $reads | samtools sort -@ ${task.cpus - 1} -O bam - > ${prefix}.bam
+        $reads | samtools sort -@ ${task.cpus} -O bam - > ${prefix}.bam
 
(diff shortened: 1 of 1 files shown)


muffato:
I'm OK with it, but I want someone else to take a look. I've pinged them on Slack

nvnieuwk:
what about situations where you only use one CPU? You should maybe add a small check for that first

lukbut:
> what about situations where you only use one CPU? You should maybe add a small check for that first

@nvnieuwk Hmm I don't think that would be a problem now, after removing the `-1`? I think it was a problem previously as the resultant number of CPUs would result in 0. Now the number of CPUs would result to 1.

matthdsm (approved):

nvnieuwk:
> > what about situations where you only use one CPU? You should maybe add a small check for that first
> 
> @nvnieuwk Hmm I don't think that would be a problem now, after removing the `-1`? I think it was a problem previously as the resultant number of CPUs would result in 0. Now the number of CPUs would result to 1.

My bad, I looked at the wrong line 😁

PR #3225, opened 2023-03-30
2807 new module svtyper

kubranarci:
## PR checklist

Closes #XXX 

- [x] Add a resource `label`


nvnieuwk:

modules/nf-core/svtyper/svtyper/main.nf
```diff
@@ -0,0 +1,56 @@
...
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def vcf  = vcf ? "--input_vcf ${vcf}" : ""
+
+    """
+    svtyper \\
+        $vcf \\
+        --bam $bam \\
+        --lib_info ${prefix}.json \\
+        --output_vcf ${prefix}.gt.vcf \\
```
nvnieuwk, line 34:
```suggestion
        --output_vcf ${prefix}.vcf \\
```
Don't use custom prefixes aside from the file extension :) The pipeline developers should add this themselves through the prefix (or not if they don't want to)

nvnieuwk: You can also add a check for file name collisions as shown [here](https://nf-co.re/docs/contributing/modules#naming-conventions) part 6

modules/nf-core/svtyper/svtyper/main.nf
```diff
@@ -0,0 +1,56 @@
...
+        --ref_fasta $fasta \\
+        --write_alignment ${prefix}.relevant.bam \\
```
nvnieuwk, line 36:
```suggestion
        --write_alignment ${prefix}.bam \\
```

Same here :)

modules/nf-core/svtyper/svtyper/main.nf
```diff
@@ -0,0 +1,56 @@
...
+    "${task.process}":
+        svtyper: \$(echo \$(svtyper -h 2>&1 | grep "version:" | sed 's/^version: v//'))
+    END_VERSIONS
+    """
+    stub:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    touch ${prefix}.json
+    touch ${prefix}.gt.vcf
+    touch ${prefix}.relevant.bam
```
nvnieuwk, line 49:
```suggestion
    touch ${prefix}.vcf
    touch ${prefix}.bam
```

modules/nf-core/svtyper/svtyper/meta.yml
```diff
@@ -0,0 +1,66 @@
...
+      doi: "doi:10.1038/nmeth.3505"
+      licence: "['MIT']"
+
+input:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. [ id:'test']
+  - meta2:
```
nvnieuwk, line 22: Can you also add meta3 here?

nvnieuwk (changes requested):

modules/nf-core/svtyper/svtyper/main.nf
kubranarci: gt.vcf was the way that developers use. Besides, what if input and output names became the same at the end? How will the output be differentiated? That did not make sense to be

nvnieuwk: You can read the link I sent you in my previous message, this explains everything. You can still add the `.gt` yourself through prefix if you want to use it :)

kubranarci:

nvnieuwk:
> I went through a lot of new modules and I saw a lot of examples of adding extensions there. To be honest, l cannot find like any similar one you requested from me? Here, output files specific to the tool itself. I don't think naming would be an issue. Besides, It would make it hard to grep the correct output here.

This is a pretty recent guideline so not all modules are up to date yet. You should add a check that fails the process when both the input and the output files are called the same. I get why you are saying this, but by removing fixed suffices, we ensure the most flexibility for pipeline developers using this module, which is the most important thing for nf-core modules

kubranarci:
> > I went through a lot of new modules and I saw a lot of examples of adding extensions there. To be honest, l cannot find like any similar one you requested from me? Here, output files specific to the tool itself. I don't think naming would be an issue. Besides, It would make it hard to grep the correct output here.
> 
> This is a pretty recent guideline so not all modules are up to date yet. You should add a check that fails the process when both the input and the output files are called the same. I get why you are saying this, but by removing fixed suffices, we ensure the most flexibility for pipeline developers using this module, which is the most important thing for nf-core modules

The way you suggest prevents the usage of different prefixes for output types and makes it even more complicated. Seriously, even for recent pushes (3 hr ago) additional suffixes are there :D

nvnieuwk:
> > > I went through a lot of new modules and I saw a lot of examples of adding extensions there. To be honest, l cannot find like any similar one you requested from me? Here, output files specific to the tool itself. I don't think naming would be an issue. Besides, It would make it hard to grep the correct output here.
> > 
> > 
> > This is a pretty recent guideline so not all modules are up to date yet. You should add a check that fails the process when both the input and the output files are called the same. I get why you are saying this, but by removing fixed suffices, we ensure the most flexibility for pipeline developers using this module, which is the most important thing for nf-core modules
> 
> The way you suggest prevents the usage of different prefixes for output types and makes it even more complicated. Seriously, even for recent pushes (3 hr ago) additional suffixes are there :D

You can change output file names when publishing them to the output directory. But I get what you mean, I'm just saying what the guidelines are for this. @nf-core/maintainers can I get some more opinions on this?

modules/nf-core/svtyper/svtyper/main.nf
```diff
@@ -0,0 +1,59 @@
...
+    input:
+    tuple val(meta), path(bam), path(bam_index), path(vcf)
+    tuple val(meta2), path(fasta)
+    tuple val(meta2), path(fai)
+
+    output:
+    tuple val(meta), path("*.json")        , emit: json
+    tuple val(meta), path("*.vcf") , emit: gt_vcf
+    tuple val(meta), path("*.bam") , emit: bam
+    path "versions.yml"            , emit: versions
```
maxulysse, line 20:
```suggestion
    tuple val(meta), path("*.json"), emit: json
    tuple val(meta), path("*.vcf") , emit: gt_vcf
    tuple val(meta), path("*.bam") , emit: bam
    path "versions.yml"            , emit: versions
```

SPPearce:
The failure [https://github.com/nf-core/modules/actions/runs/4573550127/jobs/8074167388?pr=3225#step:12:54](is the md5sum for the bam file).
It is very common for a bam file to be different between docker/singularity and conda. The bam file contains paths in its header, and they end up slightly different on conda. 
Two options:

1. Don't test the md5sum for the bam file, if it isn't the main output of the tool.
2. Ignore the failing conda test and merge anyway.

SPPearce:
It looks like the bam file is not the most important part of this module, so I would be inclined to swap to just checking that the bam file exists, rather than its md5sum.

kubranarci:
> The failure [is on the md5sum for the bam file.](https://github.com/nf-core/modules/actions/runs/4573550127/jobs/8074167388?pr=3225#step:12:54) It is very common for a bam file to be different between docker/singularity and conda. The bam file contains paths in its header, and they end up slightly different on conda. Two options:
> 
> 1. Don't test the md5sum for the bam file, if it isn't the main output of the tool.
> 2. Ignore the failing conda test and merge anyway.

Thank you! Then I will be waiting for reviewers' approval to merge this.

nvnieuwk (approved):
LGTM!

PR #3232, opened 2023-03-31
Sentieon haplotyper module

asp8200:
This Sentieon-haplotyper-module is based on a similar module from NGC's Sentieon-augmented Sarek-pipeline.

Based on input from @SPPearce, I've tried to adjust the module to the nf-core style.

I've tried to make the module flexible enough to do "command chaining", ie. run two algorithms in one go:
```
sentieon driver OPTIONS --algo ALGORITHM ALGO_OPTION OUTPUT --algo ALGORITHM2 ALGO_OPTION2 OUTPUT2
```
https://support.sentieon.com/manual/usages/general/#driver-syntax

The chaining is done so that the module can output a vcf, a gvcf or **both** (which was needed at NGC).

In this draft-PR, the output is controlled by the input value channel `emit_mode`. When `emit_mode` is set to `gvcf` the module will output only gvcf; if it is set to `both`, then the module will output both a vcf and a gvcfm; if it is set to anything else the module will just output a vcf.

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`


SPPearce:
> In this draft-PR, the output is controlled by certain pipeline-parameters (params.haplotypecaller_emit_vcf and params.haplotypecaller_emit_gvcf). That will need to be changed, but how? How about replacing those parameters with say task.ext.emit_vcf and task.ext.emit_gvcf?

There is a policy of not allowing custom ext arguments, instead these should be value channels I think. That could potentially just be one value, e.g. `vcf`, `gvcf` or `both` rather than two (presumably there is no point running without emitting anything at all)? Or just keep it as two separate values.

I'm not sure on what your args2 and args3 are actually doing, are they not just part of one command?

maxulysse (approved):
Love the usage of args2 and args3 to tackle the vcf/gvcf command

PR #3250, opened 2023-04-03
Add new subworkflow and modify some mistakes in makewindows and phase…

LouisLeNezet:
## PR checklist

New subworkflow for phasing huge panel or single individual vcf with or without reference, scaffold and genome map.
Some small changes has also been done to the bedtools/makewindows module and to the shapeit5/phasecommon module.

- [ ] Add a resource `label`

### DIFF
--- a/subworkflows/nf-core/vcf_phase_shapeit5/main.nf
+++ b/subworkflows/nf-core/vcf_phase_shapeit5/main.nf
@@ -1,0 +1,81 @@
+include { BEDTOOLS_MAKEWINDOWS              } from '../../../modules/nf-core/bedtools/makewindows/main.nf'
+include { SHAPEIT5_PHASECOMMON              } from '../../../modules/nf-core/shapeit5/phasecommon/main'  
+include { SHAPEIT5_LIGATE                   } from '../../../modules/nf-core/shapeit5/ligate/main'
+include { BCFTOOLS_INDEX as VCF_INDEX1      } from '../../../modules/nf-core/bcftools/index/main.nf'
+include { BCFTOOLS_INDEX as VCF_INDEX2      } from '../../../modules/nf-core/bcftools/index/main.nf'
+
+workflow VCF_PHASE_SHAPEIT5 {
+
+    take:
+    ch_vcf        // channel (mandatory): [ [id, ref], vcf, csi, pedigree, region ]
+    ch_ref        // channel (optional) : [ meta, ref, csi ]
+    ch_scaffold   // channel (optional) : [ meta, scaffold, csi ]
+    ch_map        // channel (optional) : [ meta, map]
+
+    main:
+
+    ch_versions = Channel.empty()
+
+    // It is needed to generate a file containing the region to phase in a Chr \tab Start \tab End format
+    // The meta map needing to be conserved the following steps a required
+
+    // Keep the meta map and the region in two separated channel but keed id field to link them back
+    ch_vcf
+        .multiMap { m, vcf, csi, pedigree, region ->
+            metadata: [m.id, m]
+            region: [m.id,region ]
+        }
+        .set { ch_region }
+
+    // Create the File in bed format and use the meta id for the file name
+    ch_region.region
+        .collectFile { mid, region -> [ "${mid}.bed", region.replace(":","\t").replace("-","\t") ] }
+        .map { file -> [ file.simpleName, file ] }
+        .set { ch_merged_region }
+
+    // Link back the meta map with the file
+    ch_region.metadata
+        .join(ch_merged_region)
+        .map {mid, meta, region_file -> [meta, region_file]}
+        .set { ch_region_file }
+    
+    BEDTOOLS_MAKEWINDOWS(ch_region_file)
+    ch_versions = ch_versions.mix(BEDTOOLS_MAKEWINDOWS.out.versions.first())
+
+    chunk_output = BEDTOOLS_MAKEWINDOWS.out.bed
+                                .splitCsv(header: ['Chr', 'Start', 'End'], sep: "\t", skip: 0)
+                                .map { meta, it -> [meta, it["Chr"]+":"+it["Start"]+"-"+it["End"]]}
+
+    phase_input = ch_vcf.map{m, vcf, index, pedigree, region -> [m, vcf, index, pedigree]}
+                    .combine(chunk_output, by:0)
+
+    SHAPEIT5_PHASECOMMON ( phase_input,
+                            ch_ref,
+                            ch_scaffold,
+                            ch_map )
+    ch_versions = ch_versions.mix(SHAPEIT5_PHASECOMMON.out.versions.first())
+
+    VCF_INDEX1(SHAPEIT5_PHASECOMMON.out.phased_variant)
+    ch_versions = ch_versions.mix(VCF_INDEX1.out.versions.first())
+
+    ligate_input = SHAPEIT5_PHASECOMMON.output.phased_variant
+                                        .map{meta, vcf -> [meta, vcf]}
+                                        .groupTuple()
+                                        .combine(VCF_INDEX1.out.csi
+                                            .map{meta, vcf -> [meta, vcf]}
+                                            .groupTuple(),
+                                            by:0)
+
+    SHAPEIT5_LIGATE(ligate_input) 
+    ch_versions = ch_versions.mix(SHAPEIT5_LIGATE.out.versions.first())
+
+    VCF_INDEX2(SHAPEIT5_LIGATE.out.merged_variants)
+    ch_versions = ch_versions.mix(VCF_INDEX2.out.versions.first())
+
+    emit:
+    bed                 = BEDTOOLS_MAKEWINDOWS.out.bed           // channel: [ val(meta), bed ]
+    variants_phased     = SHAPEIT5_LIGATE.out.merged_variants    // channel: [ val(meta), vcf ]
+    variants_index      = VCF_INDEX2.out.csi                     // channel: [ val(meta), csi ]
+    versions            = ch_versions                            // channel: [ versions.yml ]
+}
+
--- a/subworkflows/nf-core/vcf_phase_shapeit5/main.nf
+++ b/subworkflows/nf-core/vcf_phase_shapeit5/main.nf
@@ -30,7 +30,7 @@ workflow VCF_PHASE_SHAPEIT5 {
     // Create the File in bed format and use the meta id for the file name
     ch_region.region
         .collectFile { mid, region -> [ "${mid}.bed", region.replace(":","\t").replace("-","\t") ] }
-        .map { file -> [ file.simpleName, file ] }
+        .map { file -> [ file.baseName, file ] }
         .set { ch_merged_region }
 
     // Link back the meta map with the file
--- a/subworkflows/nf-core/vcf_phase_shapeit5/main.nf
+++ b/subworkflows/nf-core/vcf_phase_shapeit5/main.nf
@@ -75,11 +75,11 @@ workflow VCF_PHASE_SHAPEIT5 {
         .groupTuple()
         .map{ meta, vcf, csi ->
                 [ meta,
-                  vcf.sort { a, b ->
+                vcf.sort { a, b ->
                     def aStart = a.getName().split('-')[1].toInteger()
                     def bStart = b.getName().split('-')[1].toInteger()
                     aStart <=> bStart},
-                  csi]}
+                csi]}
 
     SHAPEIT5_LIGATE(ch_ligate_input)
     ch_versions = ch_versions.mix(SHAPEIT5_LIGATE.out.versions.first())
--- a/subworkflows/nf-core/vcf_phase_shapeit5/main.nf
+++ b/subworkflows/nf-core/vcf_phase_shapeit5/main.nf
@@ -7,10 +7,10 @@ include { BCFTOOLS_INDEX as VCF_INDEX2      } from '../../../modules/nf-core/bcf
 workflow VCF_PHASE_SHAPEIT5 {
 
     take:
-    ch_vcf        // channel (mandatory): [ [id, ref], vcf, csi, pedigree, region ]
-    ch_ref        // channel (optional) : [ meta, ref, csi ]
-    ch_scaffold   // channel (optional) : [ meta, scaffold, csi ]
-    ch_map        // channel (optional) : [ meta, map]
+    ch_vcf        // channel (mandatory): [ val(meta), path(vcf), path(csi), path(pedigree), val(region) ]
+    ch_ref        // channel (optional) : [ val(meta), path(ref), path(csi) ]
+    ch_scaffold   // channel (optional) : [ val(meta), path(scaffold), path(csi) ]
+    ch_map        // channel (optional) : [ val(meta), path(map)]
 
     main:
 
--- a/subworkflows/nf-core/vcf_phase_shapeit5/main.nf
+++ b/subworkflows/nf-core/vcf_phase_shapeit5/main.nf
@@ -52,7 +52,7 @@ workflow VCF_PHASE_SHAPEIT5 {
             [meta, vcf, index, pedigree] }
         .combine(ch_chunk_output, by:0)
         .map { meta, vcf, index, pedigree, chunk ->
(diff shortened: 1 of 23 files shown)


subworkflows/nf-core/vcf_phase_shapeit5/main.nf
```diff
@@ -0,0 +1,81 @@
+include { BEDTOOLS_MAKEWINDOWS              } from '../../../modules/nf-core/bedtools/makewindows/main.nf'
+include { SHAPEIT5_PHASECOMMON              } from '../../../modules/nf-core/shapeit5/phasecommon/main'
+include { SHAPEIT5_LIGATE                   } from '../../../modules/nf-core/shapeit5/ligate/main'
+include { BCFTOOLS_INDEX as VCF_INDEX1      } from '../../../modules/nf-core/bcftools/index/main.nf'
+include { BCFTOOLS_INDEX as VCF_INDEX2      } from '../../../modules/nf-core/bcftools/index/main.nf'
+
+workflow VCF_PHASE_SHAPEIT5 {
+
+    take:
+    ch_vcf        // channel (mandatory): [ [id, ref], vcf, csi, pedigree, region ]
```
nvnieuwk, line 10:
```suggestion
    ch_vcf        // channel (mandatory): [ val(meta), path(vcf), path(csi), path(pedigree), val(region) ]
```
It's preffered to specify whether the variable is a value or path in this way. This makes it really clear to the end users what to do. Can you change all comments at the input and output to this format. Can you also change the structure in the meta.yml to this format?

subworkflows/nf-core/vcf_phase_shapeit5/main.nf
```diff
@@ -0,0 +1,81 @@
...
+        .multiMap { m, vcf, csi, pedigree, region ->
+            metadata: [m.id, m]
+            region: [m.id,region ]
+        }
+        .set { ch_region }
+
+    // Create the File in bed format and use the meta id for the file name
+    ch_region.region
+        .collectFile { mid, region -> [ "${mid}.bed", region.replace(":","\t").replace("-","\t") ] }
+        .map { file -> [ file.simpleName, file ] }
```
nvnieuwk, line 33:
```suggestion
        .map { file -> [ file.baseName, file ] }
```
`simpleName` will remove all extensions from the file name. If the `meta.id` were to have dots in it, all characters after these dots would be removed too. `baseName` only removes the last extension (`.bed` in this case), see [here](https://www.nextflow.io/docs/latest/script.html#check-file-attributes) for more info

subworkflows/nf-core/vcf_phase_shapeit5/main.nf
```diff
@@ -0,0 +1,81 @@
...
+        .set { ch_merged_region }
+
+    // Link back the meta map with the file
+    ch_region.metadata
+        .join(ch_merged_region)
```
nvnieuwk, line 38:
```suggestion
        .join(ch_merged_region, failOnMismatch:true, failOnDuplicate:true)
```
We started adding these checks to make sure the workflow don't create mismatches or duplicates with `.join()`, but that should be fine in your case

subworkflows/nf-core/vcf_phase_shapeit5/main.nf
```diff
@@ -0,0 +1,81 @@
...
+        .map {mid, meta, region_file -> [meta, region_file]}
+        .set { ch_region_file }
+
+    BEDTOOLS_MAKEWINDOWS(ch_region_file)
+    ch_versions = ch_versions.mix(BEDTOOLS_MAKEWINDOWS.out.versions.first())
+
+    chunk_output = BEDTOOLS_MAKEWINDOWS.out.bed
```
nvnieuwk, line 45: Please stay at between using `ch_name =` and `.set { ch_name }` over the whole subworkflow :)

subworkflows/nf-core/vcf_phase_shapeit5/main.nf
```diff
@@ -0,0 +1,81 @@
...
+                                .splitCsv(header: ['Chr', 'Start', 'End'], sep: "\t", skip: 0)
+                                .map { meta, it -> [meta, it["Chr"]+":"+it["Start"]+"-"+it["End"]]}
+
+    phase_input = ch_vcf.map{m, vcf, index, pedigree, region -> [m, vcf, index, pedigree]}
```
nvnieuwk, line 49:
```suggestion
    ch_phase_input = ch_vcf.map{m, vcf, index, pedigree, region -> [m, vcf, index, pedigree]}
```
Can you add the `ch_` prefix to the channel names that are missing them?

subworkflows/nf-core/vcf_phase_shapeit5/main.nf
```diff
@@ -0,0 +1,81 @@
...
+                            ch_scaffold,
+                            ch_map )
+    ch_versions = ch_versions.mix(SHAPEIT5_PHASECOMMON.out.versions.first())
+
+    VCF_INDEX1(SHAPEIT5_PHASECOMMON.out.phased_variant)
+    ch_versions = ch_versions.mix(VCF_INDEX1.out.versions.first())
+
+    ligate_input = SHAPEIT5_PHASECOMMON.output.phased_variant
+                                        .map{meta, vcf -> [meta, vcf]}
+                                        .groupTuple()
```
nvnieuwk, line 63: You should add a size indicator to `.groupTuple()` to make sure there are no bottlenecks in your subworkflow (currently the run will wait until all entries are in the channel and then it will execute the `groupTuple()`. You can define the size in two main ways:
1. If the size of the groups is the same for every sample, you can add `size:<size_of_group>` to `.groupTuple()`.
2. If the size of the groups is variable you can use `groupKey()` in the `.map()` before `.groupTuple()`. [here](https://github.com/nf-core/sarek/blob/c87f4eb694a7183e4f99c70fca0f1d4e91750b33/subworkflows/local/bam_variant_calling_haplotypecaller/main.nf#L78-L93) an example

subworkflows/nf-core/vcf_phase_shapeit5/main.nf
```diff
@@ -0,0 +1,81 @@
...
+                                        .combine(VCF_INDEX1.out.csi
+                                            .map{meta, vcf -> [meta, vcf]}
+                                            .groupTuple(),
```
nvnieuwk, line 66: Please also add the size to this `.groupTuple()`

nvnieuwk (changes requested):
Hi, very good job! Here are some comments, mostly to improve readability and efficiency of the workflow

subworkflows/nf-core/vcf_phase_shapeit5/main.nf
LouisLeNezet: Thanks !

subworkflows/nf-core/vcf_phase_shapeit5/main.nf
LouisLeNezet: Okay, I will try to remember that !

subworkflows/nf-core/vcf_phase_shapeit5/main.nf
```diff
@@ -58,15 +57,15 @@ workflow VCF_PHASE_SHAPEIT5 {
...
 
-    ligate_input = SHAPEIT5_PHASECOMMON.output.phased_variant
-                                        .map{meta, vcf -> [meta, vcf]}
-                                        .groupTuple()
-                                        .combine(VCF_INDEX1.out.csi
-                                            .map{meta, vcf -> [meta, vcf]}
-                                            .groupTuple(),
-                                            by:0)
+    ch_ligate_input = SHAPEIT5_PHASECOMMON.output.phased_variant
+        .map{meta, vcf -> [groupKey(meta), vcf]}
```
nvnieuwk, line 61: You should also add a size to the `groupKey` indicating how large the group with this meta should be. e.g:

```
        .map{meta, vcf, group_size -> [groupKey(meta, group_size), vcf]}
```

Where you defined before how many chunks there are for each group and stored that as the `group_size` value

nvnieuwk:
Also the name of the `bed` file create is not correct
```
Error: Artifact path is not valid: /pytest_workflow_ijb7stga/vcf_phase_shapeit5_test_vcf_phase_shapeit5_ind/work/73/2a16aefa143d0399fa6d5492d020aa/input1_chr21:16600115-16799989.bed. Contains the following character:  Colon :
```

subworkflows/nf-core/vcf_phase_shapeit5/main.nf
LouisLeNezet: Ah okay !

subworkflows/nf-core/vcf_phase_shapeit5/main.nf
```diff
@@ -0,0 +1,88 @@
...
+
+    ch_ligate_input = SHAPEIT5_PHASECOMMON.output.phased_variant
+        .combine(chunks_number, by:0)
+        .map{meta, vcf, chunks_num -> [groupKey(meta, chunks_num), vcf]}
+        .groupTuple()
+        .combine(VCF_INDEX1.out.csi
+            .combine(chunks_number, by:0)
+            .map{meta, csi, chunks_num -> [groupKey(meta, chunks_num), csi]}
+            .groupTuple(),
+            by:0)
```
nvnieuwk, line 74:
```suggestion
        .join(VCF_INDEX1.out.csi, failOnMismatch:true, failOnDuplicate:true)
        .combine(chunks_number, by:0)
        .map{meta, vcf, csi, chunks_num -> [groupKey(meta, chunks_num), vcf, csi]}
        .groupTuple()
```
This will do the same, but less verbose and a bit more flexible

nvnieuwk: You'll probably also need to create a temp ID field which you will change back afterwards to make sure the chunk files have different names (this is a good practice too for traceability)

subworkflows/nf-core/vcf_phase_shapeit5/main.nf
```diff
@@ -0,0 +1,88 @@
...
+    BEDTOOLS_MAKEWINDOWS(ch_region_file)
+    ch_versions = ch_versions.mix(BEDTOOLS_MAKEWINDOWS.out.versions.first())
+
+    ch_chunk_output = BEDTOOLS_MAKEWINDOWS.out.bed
+        .splitCsv(header: ['Chr', 'Start', 'End'], sep: "\t", skip: 0)
+        .map { meta, it -> [meta, it["Chr"]+":"+it["Start"]+"-"+it["End"]] }
+        .view()
+
+    // Count the number of chunks
+    chunks_number = ch_chunk_output
```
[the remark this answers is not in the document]
LouisLeNezet: True I forgot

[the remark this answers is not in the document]
nvnieuwk: No problem!

subworkflows/nf-core/vcf_phase_shapeit5/main.nf
LouisLeNezet: Shouldn't it be made at the nextflow.config level ?
If I understand, I should modify the meta.id in the phasing process (to add the region) and set it back to the origin after the ligation ?

nvnieuwk: Yes that would be ideal! We try to not use the config as much as possible in subworkflows, because those values can be easily overwritten which would break the whole workflow :)

nvnieuwk: Allright! Have a nice meal 😋

subworkflows/nf-core/vcf_phase_shapeit5/main.nf
```diff
@@ -0,0 +1,89 @@
...
+                            ch_map )
+    ch_versions = ch_versions.mix(SHAPEIT5_PHASECOMMON.out.versions.first())
+
+    VCF_INDEX1(SHAPEIT5_PHASECOMMON.out.phased_variant)
+    ch_versions = ch_versions.mix(VCF_INDEX1.out.versions.first())
+
+    ch_ligate_input = SHAPEIT5_PHASECOMMON.out.phased_variant
+        .join(VCF_INDEX1.out.csi, failOnMismatch:true, failOnDuplicate:true)
+        .map{ meta, vcf, csi -> 
+            meta.id = meta.id.substring(0, meta.id.indexOf('_'))
```
nvnieuwk, line 70: Will this work if there are other underscores in the `meta.id`?

subworkflows/nf-core/vcf_phase_shapeit5/main.nf
```diff
@@ -0,0 +1,89 @@
...
+                            ch_map )
+    ch_versions = ch_versions.mix(SHAPEIT5_PHASECOMMON.out.versions.first())
+
+    VCF_INDEX1(SHAPEIT5_PHASECOMMON.out.phased_variant)
+    ch_versions = ch_versions.mix(VCF_INDEX1.out.versions.first())
+
+    ch_ligate_input = SHAPEIT5_PHASECOMMON.out.phased_variant
+        .join(VCF_INDEX1.out.csi, failOnMismatch:true, failOnDuplicate:true)
+        .map{ meta, vcf, csi ->
+            meta.id = meta.id.substring(0, meta.id.indexOf('_'))
```
nvnieuwk, line 70: Can you make a new meta (e.g. `new_meta = meta + [id:<code_to_define_new_id>]`). It could be because of this you get some errors

subworkflows/nf-core/vcf_phase_shapeit5/main.nf
LouisLeNezet: Good question I will try !

LouisLeNezet: So it takes the first '_' as an index.

nvnieuwk (commented):
Some concerns with the meta.id but otherwise it LGTM

subworkflows/nf-core/vcf_phase_shapeit5/main.nf
```diff
@@ -0,0 +1,96 @@
...
+        .combine(ch_chunks_number, by:0)
+        .map{meta, vcf, csi, chunks_num ->
+            [groupKey(meta, chunks_num), vcf, csi]}
+        .groupTuple()
+        .map{ meta, vcf, csi ->
+                [ meta,
+                vcf.sort { a, b ->
+                    def aStart = a.getName().split('-')[1].toInteger()
+                    def bStart = b.getName().split('-')[1].toInteger()
+                    aStart <=> bStart},
```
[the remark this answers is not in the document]
LouisLeNezet: I have to admit I didn't quite know.
It's a spaceship operator and is used for comparison in Ruby. It returns -1 if the left operand is less than the right operand, 0 if they are equal, and 1 if the left operand is greater than the right operand.

subworkflows/nf-core/vcf_phase_shapeit5/main.nf
```diff
@@ -0,0 +1,96 @@
...
+                            ch_map )
+    ch_versions = ch_versions.mix(SHAPEIT5_PHASECOMMON.out.versions.first())
+
+    VCF_INDEX1(SHAPEIT5_PHASECOMMON.out.phased_variant)
+    ch_versions = ch_versions.mix(VCF_INDEX1.out.versions.first())
+
+    ch_ligate_input = SHAPEIT5_PHASECOMMON.out.phased_variant
+        .join(VCF_INDEX1.out.csi, failOnMismatch:true, failOnDuplicate:true)
+        .map{ meta, vcf, csi ->
+            newmeta = meta + [id: meta.id.substring(0, meta.id.indexOf('_'))]
```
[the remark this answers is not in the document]
LouisLeNezet: You're right I've tested it after the one of interest but not before.
It wouldn't work if it was the case.
I'll change the code accordingly.

subworkflows/nf-core/vcf_phase_shapeit5/main.nf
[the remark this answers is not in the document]
nvnieuwk: Oh interesting! Thanks for the explanation

nvnieuwk (approved):
Awesome job!

PR #3258, opened 2023-04-04
new module AnnotSV

nvnieuwk:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`


modules/nf-core/annotsv/Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM mambaorg/micromamba:1.4-focal
+
+LABEL authors="Nicolas Vannieuwkerke <nicolas.vannieuwkerke@ugent.be>" \
+    description="Docker image containing AnnotSV"
+
+ENV VERSION=3.3.2
+ENV ANNOTSV=/tmp/AnnotSV-${VERSION}
```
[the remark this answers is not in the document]
nvnieuwk: I've noticed 😁. Thanks for the suggestion I was failing miserably in fixing this.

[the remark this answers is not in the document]
pontus: Just to note; I don't know how their install looks, might be required to do a `mkdir -p "$ANNOTSV"` before installing.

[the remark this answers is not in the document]
nvnieuwk: Would you know why when converting to a singularity image, the files are deleted from the directory they were downloaded in (It's in `/home/mambauser` in the current version, because of permission issues with `/opt`). The tool needs to have access to the file structure like when it's downloaded to fully work. (For now when converting to singularity, the AnnotSV executable gets moved to `/usr/local/bin` instead)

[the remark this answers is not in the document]
nvnieuwk: > Just to note; I don't know how their install looks, might be required to do a `mkdir -p "$ANNOTSV"` before installing.

That variable should point to the directory containing the source code of AnnotSV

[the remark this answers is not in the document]
pontus: > > Just to note; I don't know how their install looks, might be required to do a `mkdir -p "$ANNOTSV"` before installing.
> 
> That variable should point to the directory containing the source code of AnnotSV

It doesn't look like it's used like that - instead the source is fetched and extracted to / - instead, the variable is used as installation prefix.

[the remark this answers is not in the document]
nvnieuwk: Yes indeed, but the main problem is that it's also used to find to location of a dependency which is in the source directory. It can't find this dependency if this variable points to another location :/



> Anything in /home is also bad, either switch back and forth with `USER` or make it multi-stage and copy over between?

What do you mean by multi-stage exactly?

nvnieuwk:
Singularity test is failing because the `quay` container isn't updated yet

modules/nf-core/annotsv/Dockerfile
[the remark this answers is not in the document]
pontus: > What do you mean by multi-stage exactly?

Multi-stage are builds that are done in multiple stages, to allow for using less layers in the final image/not including tooling needed for building.

Are you sure it's really expecting to look at the source directory?

https://github.com/pontus/nf-core-modules/tree/new-module-annotsv/modules/nf-core/annotsv is an example of how I think it should be after a quick look at AnnotSV (I'm not sure about licensing of the annotations or phenotypes fetched by make install and thus that may need to be addressed).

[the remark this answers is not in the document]
nvnieuwk: Thanks for the explanation it's clear to me now! I've been working with the developer to make the tool more compatible (and eventually add it to bioconda). We've made some major improvements which will make sure we won't have to do this 😄

nvnieuwk:
I'll wait with this merge until the tool has been added to bioconda

nvnieuwk:
Tool has been added to bioconda, now waiting for the containers to build :)

nvnieuwk:
A new build of the conda package is underway so I'll wait with the merge so I can use that one here. It won't impact this module though. (It was missing a dependency for the download of the annotations)

modules/nf-core/annotsv/annotsv/main.nf
```diff
@@ -0,0 +1,70 @@
...
+    tuple val(meta), path(variants), path(variants_index)
+    tuple val(meta2), path(annotations)
+    tuple val(meta3), path(candidate_genes)
+    tuple val(meta4), path(candidate_small_variants)
+    tuple val(meta5), path(false_positive_snv)
+    tuple val(meta6), path(gene_transcripts)
+
+    output:
+    tuple val(meta), path("*.tsv")              , emit: tsv
+    tuple val(meta), path("*.unannotated.tsv")  , emit: unannotated_tsv, optional:true
```
SPPearce, line 20:
```suggestion
    tuple val(meta), path("*.unannotated.tsv")  , emit: unannotated_tsv, optional: true
```

modules/nf-core/annotsv/annotsv/main.nf
```diff
@@ -0,0 +1,70 @@
...
+    tuple val(meta), path("*.vcf")              , emit: vcf, optional:true
```
SPPearce, line 21: Line the optional up? I don't know what has been done for other modules...

modules/nf-core/annotsv/annotsv/main.nf
```diff
@@ -0,0 +1,70 @@
...
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+
+    def cand_genes = candidate_genes ? "-candidateGenesFile ${candidate_genes}" : ""
+    def cand_small = candidate_small_variants ? "-candidateSnvIndelFiles ${candidate_small_variants}" : ""
+    def fp_snv = false_positive_snv ? "-snvIndelFiles ${false_positive_snv}" : ""
+    def transcripts = gene_transcripts ? "-txFile ${gene_transcripts}" : ""
```
SPPearce, line 34: Is this being used?

tests/modules/nf-core/annotsv/annotsv/main.nf
```diff
@@ -0,0 +1,53 @@
...
+    
+    input = [
+        [ id:'test', single_end:false ], // meta map
+        file(params.test_data["homo_sapiens"]["illumina"]["test_sv_vcf"], checkIfExists: true),
+        file(params.test_data["homo_sapiens"]["illumina"]["test_sv_vcf_tbi"], checkIfExists: true)
+    ]
+
+    annotations = [
+        [ id: 'annotations' ],
+        [] // For stub use only, this will fail if the actual module is run like this
```
SPPearce, line 17: Is this actually testing the module at all, or just a stub?

modules/nf-core/annotsv/annotsv/main.nf
nvnieuwk: I think only the emits are aligned (I prefer this too as it will become just some scattered text this way)

modules/nf-core/annotsv/annotsv/main.nf
nvnieuwk: Good spot! Thanks

tests/modules/nf-core/annotsv/annotsv/main.nf
nvnieuwk: It's a stub. The annotation files are 20 GB so there's no way I can add them to the test datasets repo, but I tested locally and it works like a charm

modules/nf-core/annotsv/annotsv/main.nf
```diff
@@ -0,0 +1,70 @@
...
+
+    """
+    AnnotSV \\
+        -annotationsDir ${annotations} \\
+        ${cand_genes} \\
+        ${cand_small} \\
+        ${fp_snv} \\
+        -outputFile ${prefix}.tsv \\
```
nvnieuwk, line 42:
```suggestion
        ${transcripts} \\
        -outputFile ${prefix}.tsv \\
```

modules/nf-core/annotsv/annotsv/main.nf
```diff
@@ -0,0 +1,71 @@
...
+    tuple val(meta2), path(annotations)
+    tuple val(meta3), path(candidate_genes)
+    tuple val(meta4), path(candidate_small_variants)
+    tuple val(meta5), path(false_positive_snv)
+    tuple val(meta6), path(gene_transcripts)
+
+    output:
+    tuple val(meta), path("*.tsv")              , emit: tsv
+    tuple val(meta), path("*.unannotated.tsv")  , emit: unannotated_tsv, optional: true
+    tuple val(meta), path("*.vcf")              , emit: vcf, optional:true
```
SPPearce, line 21:
```suggestion
    tuple val(meta), path("*.vcf")              , emit: vcf, optional: true
```

SPPearce (approved):

nvnieuwk:
Thanks @SPPearce!

PR #3269, opened 2023-04-06
Increased version of container to 0.4.1

FloWuenne:
Increase of release/container version to support compressed images using imagecodecs.

https://github.com/vanvalenlab/deepcell-applications/commit/0075eaa7f510a0da61f6b1c38023dace55d6eb79


erikrikarddaniel (approved):
That's a simple one! :-)

PR #3306, opened 2023-04-19
Fix subworkflows based on json schema

mashehu:
Yet another json schema, this time for subworkflows (needed so they are all nicely rendered on the website).
In order for the two schemas to behave nicely, I had to split them into two files and put them into the respective subfolders. It is very similar to the schema for modules, biggest difference is that I don't require types for the input and output, because it felt like 90% was some kind of `map` (you can still give them and they will be rendered on the homepage).

I fixed the broken meta.yaml according to  this new rules.

I also added the hint for the yaml-language server, to nicely show the problems already in the editor (thanks @Emiller88 for pointing this out)

### DIFF
--- a/subworkflows/yaml-schema.json
+++ b/subworkflows/yaml-schema.json
@@ -1,7 +1,7 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema",
     "title": "Meta yaml",
-    "description": "Validate the meta yaml file for an nf-core module or subworkflow",
+    "description": "Validate the meta yaml file for an nf-core subworkflow",
     "type": "object",
     "properties": {
         "name": {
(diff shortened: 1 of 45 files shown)


subworkflows/yaml-schema.json
```diff
@@ -0,0 +1,93 @@
...
+                                "description": "Type of the input channel"
+                            },
+                            "description": {
+                                "type": "string",
+                                "description": "Description of the input channel"
+                            },
+                            "pattern": {
+                                "type": "string",
+                                "description": "Pattern of the input channel, given in Java glob syntax"
+                            }
```
mirpedrol, line 57: We have also a "default" field in some of them, do we want to check this one too?

mirpedrol (approved):
LGTM :)

subworkflows/yaml-schema.json
mashehu: Good to know, I'll add it back then.

maxulysse:
cf https://github.com/nf-core/modules/pull/3340

PR #3310, opened 2023-04-21
new module: samtools sormadup

matthdsm:
This is a meta module combining several samtools tools to correctly markdup and sort bam files in one go.

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`

### DIFF
--- a/modules/nf-core/samtools/sormadup/main.nf
+++ b/modules/nf-core/samtools/sormadup/main.nf
@@ -2,10 +2,10 @@ process SAMTOOLS_SORMADUP {
     tag "$meta.id"
     label 'process_low'
 
-    conda "bioconda::samtools=1.16.1"
+    conda "bioconda::samtools=1.17"
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
-        'https://depot.galaxyproject.org/singularity/samtools:1.16.1--h6899075_1' :
-        'quay.io/biocontainers/samtools:1.16.1--h6899075_1' }"
+        'https://depot.galaxyproject.org/singularity/samtools:1.17--h00cdaf9_0' :
+        'quay.io/biocontainers/samtools:1.17--h00cdaf9_0' }"
 
     input:
     tuple val(meta), path(input)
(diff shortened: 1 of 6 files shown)


matthdsm:
xref to slack: https://nfcore.slack.com/archives/CJRH30T6V/p1682062426042809

mashehu:
Why is this a module and [bam_stats_samtools](https://github.com/nf-core/modules/blob/master/subworkflows/nf-core/bam_stats_samtools/main.nf) a subworkflow? Should we also port that one to a module?

matthdsm:
This is a module because it's a toolchain that's being piped to one another, thus eliminating the need for a bunch of intermediate files.
bam_stats_samtools are 3 distinct tools with separate (useful) outputs

mashehu:
makes kind of sense. We should then update our modules definition, which currently [says](https://nf-co.re/docs/contributing/modules#module):
> A module [...] is as atomic as possible i.e. cannot be split into another module.

modules/nf-core/samtools/sormadup/main.nf
```diff
@@ -0,0 +1,67 @@
...
+    output:
+    tuple val(meta), path("*.bam")      , emit: bam
+    tuple val(meta), path("*.metrics")  , emit: metrics
+    path "versions.yml"                 , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
```
muffato, line 23: Don't we need another args for at least the initial `samtools cat` ? (perhaps for all the other `samtools` commands too)

muffato (commented):
To be fair, there are already multiple non-atomic modules that for instance add a `samtools view` or a `samtools sort` in inputs/outputs, or do additional `gzip`, `bgzip`, `gunzip`, etc.

But I agree that this feels more like a sub-workflow. What about making a wrapper sub-workflow too ? As a user I wouldn't be looking `modules/` for such a sub-workflow.

modules/nf-core/samtools/sormadup/main.nf
matthdsm: I'm not really sure, I'd like to keep this as simple as possible to avoid any confusion. I do agree it would be good for me modifiability though.

matthdsm: My main target here it to get a samtools equivalent for `BIOBAMBAM_BAMSORMADUP`, that's why I went with the module route

matthdsm: Added the extra `ext.args`

PR #3331, opened 2023-04-25
New module : Pilon

### LINKED ISSUE #831: new module: pilon
I think it would be good to have a module for [pilon](https://bioconda.github.io/recipes/pilon/README.html)

This could be used to improve nf-core/bacass and is suggested [here](https://github.com/nf-core/bacass/issues/57) and used in nf-core/mag.

scorreard:
## PR checklist

Closes #831  

- [x] Add a resource `label`

### DIFF
--- a/modules/nf-core/pilon/main.nf
+++ b/modules/nf-core/pilon/main.nf
@@ -9,7 +9,7 @@ process PILON {
 
     input:
     tuple val(meta), path(fasta)
-    tuple val(meta), path(bam), path(bai)
+    tuple val(meta_bam), path(bam), path(bai)
     val pilon_mode
 
     output:
--- a/modules/nf-core/pilon/main.nf
+++ b/modules/nf-core/pilon/main.nf
@@ -29,12 +29,12 @@ process PILON {
     def valid_mode = ["frags", "jumps", "unpaired", "bam"]
     if ( !valid_mode.contains(pilon_mode) )  { error "Unrecognised mode to run Pilon. Options: ${valid_mode.join(', ')}" }
     """
-        pilon \\
-	--genome $fasta \\
-	--output ${meta.id} \\
-	--threads $task.cpus \\
+    pilon \\
+        --genome $fasta \\
+        --output ${meta.id} \\
+        --threads $task.cpus \\
         $args \\
-	--$pilon_mode $bam
+        --$pilon_mode $bam
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
--- a/modules/nf-core/pilon/main.nf
+++ b/modules/nf-core/pilon/main.nf
@@ -1,0 +1,43 @@
+process PILON {
+    tag "$meta.id"
+    label 'process_medium'
+
+    conda "bioconda::pilon=1.24"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/pilon:1.24--hdfd78af_0':
+        'quay.io/biocontainers/pilon:1.24--hdfd78af_0' }"
+
+    input:
+    tuple val(meta), path(fasta)
+    tuple val(meta), path(bam), path(bai)
+    val pilon_mode
+
+    output:
+    tuple val(meta), path("*.fasta") , emit: improved_assembly
+    tuple val(meta), path("*.vcf")   , emit: vcf               , optional : true
+    tuple val(meta), path("*.change"), emit: change_record     , optional : true
+    tuple val(meta), path("*.bed")   , emit: tracks_bed        , optional : true
+    tuple val(meta), path("*.wig")   , emit: tracks_wig        , optional : true
+    path "versions.yml"              , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def valid_mode = ["frags", "jumps", "unpaired", "bam"]
+    if ( !valid_mode.contains(pilon_mode) )  { error "Unrecognised mode to run Pilon. Options: ${valid_mode.join(', ')}" }
+    """
+        pilon \\
+	--genome $fasta \\
+	--output ${meta.id} \\
+	--threads $task.cpus \\
+        $args \\
+	--$pilon_mode $bam
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        pilon: \$(echo \$(pilon --version) | sed 's/^.*version //; s/ .*\$//' )
+    """
+}
--- a/modules/nf-core/pilon/meta.yml
+++ b/modules/nf-core/pilon/meta.yml
@@ -25,7 +25,7 @@ input:
       pattern: "*.{fasta}"
   - bam:
       type: file
-      description: BAM file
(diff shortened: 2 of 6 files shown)


tests/modules/nf-core/pilon/test.yml
```diff
@@ -0,0 +1,8 @@
+- name: pilon test_pilon
+  command: nextflow run ./tests/modules/nf-core/pilon -entry test_pilon -c ./tests/config/nextflow.config -c ./tests/modules/nf-core/pilon/nextflow.config
+  tags:
+    - pilon
+  files:
+    - path: output/pilon/test.fasta
+      md5sum: 2e881994820a5a641da9ea594ab4958f
+    - path: output/pilon/versions.yml
```
d4straub, line 8: shouldnt the version also be reproducible, i.e. the file should also have a stable md5sum??

tests/modules/nf-core/pilon/main.nf
```diff
@@ -0,0 +1,22 @@
...
+    ]
+
+    bam_tuple_ch = Channel.of([ [ id:'test', single_end:false ], // meta map
+                               file(params.test_data['homo_sapiens']['illumina']['test_paired_end_sorted_bam'], checkIfExists: true),
+                               file(params.test_data['homo_sapiens']['illumina']['test_paired_end_sorted_bam_bai'], checkIfExists: true),
+                                ])
+
+    pilon_mode = "bam"
+
+    PILON ( input, bam_tuple_ch, pilon_mode )
```
d4straub, line 21:
```suggestion
    PILON ( input, bam_tuple_ch, "bam" )
```
is fine as well I think, but shorter so I would prefer it.

d4straub (commented):
Looks good!
Formatting fails and I have a few suggestions, but fine overall I think.

modules/nf-core/pilon/main.nf
```diff
@@ -0,0 +1,43 @@
...
+    label 'process_medium'
+
+    conda "bioconda::pilon=1.24"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/pilon:1.24--hdfd78af_0':
+        'quay.io/biocontainers/pilon:1.24--hdfd78af_0' }"
+
+    input:
+    tuple val(meta), path(fasta)
+    tuple val(meta), path(bam), path(bai)
```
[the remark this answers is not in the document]
scorreard: You are right, it may be the same, but it may differ, so let's call it meta_bam.

modules/nf-core/pilon/main.nf
```diff
@@ -0,0 +1,43 @@
...
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def valid_mode = ["frags", "jumps", "unpaired", "bam"]
+    if ( !valid_mode.contains(pilon_mode) )  { error "Unrecognised mode to run Pilon. Options: ${valid_mode.join(', ')}" }
+    """
+        pilon \\
+	--genome $fasta \\
+	--output ${meta.id} \\
+	--threads $task.cpus \\
+        $args \\
+	--$pilon_mode $bam
```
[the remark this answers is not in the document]
scorreard: Should be better now, thanks for seeing that.

tests/modules/nf-core/pilon/test.yml
scorreard: I just checked in other test.yml files and the version.yml is usually not associated to a md5sum (I don't know why),, but I think that's ok.

Example with DeepVariant : https://github.com/nf-core/modules/blob/master/tests/modules/nf-core/deepvariant/test.yml
Example with Hifiasm : https://github.com/nf-core/modules/blob/master/tests/modules/nf-core/hifiasm/test.yml

modules/nf-core/pilon/meta.yml
```diff
@@ -0,0 +1,71 @@
...
+input:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. [ id:'test', single_end:false ]
+  - fasta:
+      type: file
+      description: FASTA file
+      pattern: "*.{fasta}"
```
d4straub, line 25: Sorry, I have always just a few minutes, I know that isnt ideal. But looking at the pilon documentation I found that this seem to be the assembly, so it might be good to detail that here:
```suggestion
  - fasta:
      type: file
      description: FASTA of the input genome
      pattern: "*.{fasta}"
```

modules/nf-core/pilon/meta.yml
```diff
@@ -0,0 +1,71 @@
...
+  - bam:
+      type: file
+      description: BAM file
+      pattern: "*.{bam}"
```
d4straub, line 29:
```suggestion
  - bam:
      type: file
      description: BAM file of reads aligned to the input genome
      pattern: "*.{bam}"
```

modules/nf-core/pilon/meta.yml
```diff
@@ -0,0 +1,71 @@
...
+  - bai:
+      type: file
+      description: BAI file
+      pattern: "*.{bai}"
```
d4straub, line 33:
```suggestion
  - bai:
      type: file
      description: BAI file (BAM index) of BAM reads aligned to the input genome
      pattern: "*.{bai}"
```

modules/nf-core/pilon/main.nf
```diff
@@ -0,0 +1,43 @@
...
+    label 'process_medium'
+
+    conda "bioconda::pilon=1.24"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/pilon:1.24--hdfd78af_0':
+        'quay.io/biocontainers/pilon:1.24--hdfd78af_0' }"
+
+    input:
+    tuple val(meta), path(fasta)
+    tuple val(meta_bam), path(bam), path(bai)
```
d4straub, line 12: Because fasta, bam and bai have to come all from the same sample, would it be rather:
```suggestion
    tuple val(meta), path(fasta), path(bam), path(bai)
```
I am not sure, thats just a question because I do not know better.

scorreard: That is an interesting suggestion. Both would work.
I prefer to have them split, and it works if meta and meta_bam are the same.
In the case of purge_dups, they concatenated all the inputs and I had to split them because sometimes my meta reflects the input sequencing type (SR, PacBio, ONT), not only the sample name.

d4straub: > I prefer to have them split, and it works if meta and meta_bam are the same.

Hm ok, I would prefer having them not split. Because I think the typical application is that before the module is some channel magic that joins the required input (genome = fasta, alignment = bam & bai) to have matching files by ensuring that meta is matching.

> In the case of purge_dups, they concatenated all the inputs and I had to split them because sometimes my meta reflects the input sequencing type (SR, PacBio, ONT), not only the sample name.

That seems to me rather the exception than the usual module output. But I might be wrong.

Unfortunately the nf-core modules guidelines do not specify this clearly, see https://nf-co.re/developers/modules#inputoutput-options, i.e.  `Directly associated auxiliary files to an input file MAY be defined within the same input channel alongside the main input channel (e.g. BAM and BAI).`, so it seems I shouldn't insist on it!

d4straub (approved):

PR #3348, opened 2023-04-28
Sentieon docker image

adamrtalbot:
Replace Sentieon Docker image with one from Quay.io

There might be a good reason for not using quay.io, but it's worth a try.

- Turns off Singularity tests because they don't work because of LD_PRELOAD (not sure why not) but Conda and Docker seem to work.

### DIFF
--- a/.github/workflows/pytest-workflow.yml
+++ b/.github/workflows/pytest-workflow.yml
@@ -107,14 +107,6 @@ jobs:
           - profile: "conda"
             tags: scimap/mcmicro
           - profile: "conda"
-            tags: sentieon/bwaindex
-          - profile: "conda"
-            tags: sentieon/bwamem
-          - profile: "conda"
-            tags: sentieon/dedup
-          - profile: "conda"
-            tags: sentieon/haplotyper
-          - profile: "conda"
             tags: universc
           - profile: "singularity"
             tags: universc
--- a/.github/workflows/pytest-workflow.yml
+++ b/.github/workflows/pytest-workflow.yml
@@ -110,6 +110,14 @@ jobs:
             tags: universc
           - profile: "singularity"
             tags: universc
+          - profile: "singularity"
+            tags: sentieon/bwaindex
+          - profile: "singularity"
+            tags: sentieon/bwamem
+          - profile: "singularity"
+            tags: sentieon/dedup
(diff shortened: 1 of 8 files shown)


.github/workflows/pytest-workflow.yml
```diff
@@ -107,17 +107,17 @@ jobs:
...
+            tags: universc
+          - profile: "singularity"
+            tags: universc
+          - profile: "singularity"
             tags: sentieon/bwaindex
-          - profile: "conda"
+          - profile: "singularity"
             tags: sentieon/bwamem
-          - profile: "conda"
+          - profile: "singularity"
```
maxulysse, line 117: What is happening there?

.github/workflows/pytest-workflow.yml
```diff
@@ -107,17 +107,17 @@ jobs:
...
             tags: sentieon/dedup
-          - profile: "conda"
-            tags: sentieon/haplotyper
-          - profile: "conda"
-            tags: universc
           - profile: "singularity"
-            tags: universc
+            tags: sentieon/haplotyper
```
maxulysse, line 120: Same question

maxulysse (commented):
Available on conda now?

.github/workflows/pytest-workflow.yml
adamrtalbot: Sentieon on conda works fine, Sentieon on singularity has a problem with LD_LIBRARY_PATH. It should be fine on _most_ machines but doesn't work on the GHA runners. See here for when I fixed the problem: https://github.com/bioconda/bioconda-recipes/pull/38781

asp8200: I can't run Sentieon with the singularity images introduced in this PR. I get the now well-known:
```
error while loading shared libraries: libstdc++.so.6: cannot open shared object file: No such file or directory
```
Similar issue with the subsequent images made by Don for Sentieon 202112.07.

I'm afraid this is a showstopper for this PR 😬

adamrtalbot: Darn!!! I don't know how singularity containers are built, where do they inherit `LD_LIBRARY` from?

maxulysse:
Not needed anymore

PR #3344, opened 2023-04-27
Default registry to quay.io

adamrtalbot:
Sets the default Docker registry to `quay.io`. 

Where necessary, specifies full URI for every Docker container that gets pulled from `docker.io`.

## PR checklist

- [x] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [ ] Remove all TODO statements.
- [ ] Emit the `versions.yml` file.
- [ ] Follow the naming conventions.
- [ ] Follow the parameters requirements.
- [ ] Follow the input/output options guidelines.
- [ ] Add a resource `label`
- [ ] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - [ ] `PROFILE=docker pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [ ] `PROFILE=singularity pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [ ] `PROFILE=conda pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`

PR #3352, opened 2023-04-28
fix-meta-liniting

sateeshperi:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`

### DIFF
--- a/modules/nf-core/cat/fastq/meta.yml
+++ b/modules/nf-core/cat/fastq/meta.yml
@@ -1,6 +1,7 @@
 name: cat_fastq
 description: Concatenates fastq files
 keywords:
+  - cat
   - fastq
   - concatenate
 tools:
@@ -16,7 +17,7 @@ input:
         Groovy Map containing sample information
         e.g. [ id:'test', single_end:false ]
   - reads:
-      type: list
+      type: file
       description: |
         List of input FastQ files to be concatenated.
 output:
--- a/modules/nf-core/kraken2/kraken2/meta.yml
+++ b/modules/nf-core/kraken2/kraken2/meta.yml
@@ -28,12 +28,12 @@ input:
       type: directory
       description: Kraken2 database
   - save_output_fastqs:
-      type: boolean
+      type: string
       description: |
         If true, optional commands are added to save classified and unclassified reads
         as fastq files
   - save_reads_assignment:
-      type: boolean
+      type: string
       description: |
         If true, an optional command is added to save a file reporting the taxonomic
         classification of each input read
--- a/.github/workflows/pytest-workflow.yml
+++ b/.github/workflows/pytest-workflow.yml
@@ -87,9 +87,17 @@ jobs:
           - profile: "conda"
             tags: fcs/fcsgx
           - profile: "conda"
+            tags: gatk4/baserecalibratorspark
+          - profile: "conda"
             tags: gatk4/cnnscorevariants
           - profile: "conda"
             tags: gatk4/determinegermlinecontigploidy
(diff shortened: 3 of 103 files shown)


modules/nf-core/cat/fastq/meta.yml
```diff
@@ -16,7 +17,7 @@ input:
         Groovy Map containing sample information
         e.g. [ id:'test', single_end:false ]
   - reads:
-      type: list
+      type: file
```
sateeshperi, line 20: is this supposed to be a map?

mashehu: I think file is correct here, because this type can also be multiple files as seen below

mashehu (approved):

modules/nf-core/kraken2/kraken2/meta.yml
```diff
@@ -28,12 +28,12 @@ input:
       type: directory
       description: Kraken2 database
   - save_output_fastqs:
-      type: boolean
+      type: string
```
JoseEspinosa, line 31: Is this not a boolean?

modules/nf-core/kraken2/kraken2/meta.yml
```diff
@@ -28,12 +28,12 @@ input:
...
       description: |
         If true, optional commands are added to save classified and unclassified reads
         as fastq files
   - save_reads_assignment:
-      type: boolean
+      type: string
```
JoseEspinosa, line 36: Same here

modules/nf-core/kraken2/kraken2/meta.yml
mashehu: we don't have boolean as a type, see (https://nf-co.re/docs/contributing/modules#documentation point 8). Not sure if that is a nextflow limitation (I would be for it)

ewels:
Merged into PR https://github.com/nf-core/modules/pull/3344 and then merged into `master` in https://github.com/nf-core/modules/commit/b9829e1064382745d8dff7f1d74d2138d2864f71.

Thanks @sateeshperi !

modules/nf-core/kraken2/kraken2/meta.yml
JoseEspinosa: I see that the PR is now closed, but I wanted to clarify this point anyway. I checked the modules repo and when I searched for `type: boolean` there are 72 occurrences in `meta.yml`s files both in modules and subworkflows. So if this is not allowed we should get rid of them. When you say this is a nextflow limitation what do you mean exactly? Thanks!

sateeshperi: ya i see those 72 too....i think we will have to change those too but,I am also curious as to why boolean is not supported @mashehu . now that we specify as string, will providing `"true"` be the same as `true` ?

mashehu: was just going off of the guidelines, which I think @jfy133 worked on.

mashehu: But since we don't validate these types anywhere (besides in json-schema for meta.yml) and don't use them anywhere AFAIK (besides rendering them on the website), I am all for adding `boolean` as a type to the guidelines.

mashehu: We had it already in the schema, added it also to the guidelines now: https://github.com/nf-core/nf-co.re/pull/1767/files

sateeshperi: will tools also need an update? as linting with boolean fails

PR #3385, opened 2023-05-04
Update vep subwf

nvnieuwk:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`


nvnieuwk:
@matthdsm, @maxulysse and @adamrtalbot let's break this down!

nvnieuwk:
Answer PR to issues found in #3006

tests/subworkflows/nf-core/vcf_annotate_ensemblvep/nextflow.config
```diff
@@ -2,12 +2,13 @@ process {
...
     publishDir = { "${params.outdir}/${task.process.tokenize(':')[-1].tokenize('_')[0].toLowerCase()}" }
 
-    withName: ENSEMBLVEP_VEP {
-        container = 'docker.io/nfcore/vep:108.2.WBcel235'
+    withName: BCFTOOLS_QUERY {
+        ext.args = "--list-samples"
     }
 
-    withName: vcf_annotate_ensemblvep_custom:VCF_ANNOTATE_ENSEMBLVEP_CUSTOM:ENSEMBLVEP_VEP {
-        ext.args   = '--custom test2.vcf,,vcf,exact,0,TOPMED --custom test3.vcf,,vcf,exact,0,TOPMED'
```
maxulysse, line 10: why removing all that?

matthdsm:
Personally, I think the merging of samples into one is out of scope for this subworkflow. The point here should be to abstract away the scatter-gather from workflow dev's, not manipulate the input channel.

Collecting the input vcf's into one creates a huge bottleneck and a workflow dev should be able to choose if they want to get that performance hit.

I'd vote to leaving it out completely and let dev's handle pre and postprocessing of the input vcf themselves, no magic required.

nvnieuwk:
A suggestion:

I'll remove the merging of the VCFs entirely from this subworkflow and make this purely scatter -> VEP -> gather. We can make another subworkflow that merges the VCFs before and calls the scatter-gather subworkflow with this VCF. This way custom sample-specific files will still be an option for the normal subworkflow, this will also prevent the bottlenecks pointed out by @matthdsm and only allow them if the developer of the pipeline wants them.

matthdsm:
> I'll remove the merging of the VCFs entirely from this subworkflow and make this purely scatter -> VEP -> gather.

I approve of this message

subworkflows/nf-core/vcf_annotate_ensemblvep/main.nf
```diff
@@ -2,44 +2,170 @@
...
+
+    //
+    // Run the annotation with EnsemblVEP
+    //
+
+    ch_vep_input = BCFTOOLS_PLUGINSCATTER.out.scatter
+        .transpose()
+        .map { meta, vcf ->
+            [ [id:vcf.simpleName], vcf, [] ]
+        }
```
adamrtalbot, line 92: I need a better explanation of what this is doing. I had to actually think and I'm, not very good at thinking. Either add some comments or flesh out the meta.yml

subworkflows/nf-core/vcf_annotate_ensemblvep/main.nf
```diff
@@ -2,44 +2,170 @@
...
+            name = vcf.name.replace(".bcf", "").replace(".vcf", "").replace(".gz", "")
+            [ name, vcf ]
+        } // Re-add the original meta to each VCF
+        .join(ch_vcf_map.meta, failOnDuplicate:true, failOnMismatch:true)
+        .branch { id, vcf, meta ->
+            bgzip: vcf.extension == "gz"
+                return [ meta, vcf ]
+            unzip: true
+                return [ meta, vcf, [] ]
+        }
```
adamrtalbot, line 150: Again, some more comments to tell me what it's doing. Don't make someone have to think.

subworkflows/nf-core/vcf_annotate_ensemblvep/main.nf
```diff
@@ -2,44 +2,170 @@
...
+        .groupTuple() // No size needed here because this will merge all VCFs in the workflow
+
+    BCFTOOLS_MERGE(
+        ch_merge_input,
+        [],
+        ch_fasta instanceof List ? [] : ch_fasta.map { it[1] },
+        ch_fasta_fai instanceof List ? [] : ch_fasta_fai.map { it[1] }
+    )
+    ch_versions = ch_versions.mix(BCFTOOLS_MERGE.out.versions.first())
```
adamrtalbot, line 68: Skip if there is only 1 VCF?

adamrtalbot (commented):
Two points:
 - It's a little hard to follow, so I'd like to see some additional documentation to clear make it easier for me
 - Is the scatter gather optional? Can I turn it off? VEP has a large overhead so scatter/gather only makes sense on HPCs.

nvnieuwk:
> * It's a little hard to follow, so I'd like to see some additional documentation to clear make it easier for me

Good point! I'll add more comments in the rework after the reviews 



> * Is the scatter gather optional? Can I turn it off? VEP has a large overhead so scatter/gather only makes sense on HPCs.

I can make it optional with a boolean :). The main question is do we want to keep all this in one subworkflow or split up the merging and the scatter/gather (cf https://github.com/nf-core/modules/pull/3385#issuecomment-1535835449). I'd like some feedback on this suggestion first before making any changes to the actual subwf

subworkflows/nf-core/vcf_annotate_ensemblvep/main.nf
nvnieuwk: Good idea!

matthdsm:
@adamrtalbot 
I suggest the following:

- We create a subworkflow that does **only** scatter gather, this would then be this PR
- Subsequently, we create another subworkflow that calls either this one or the plain module to get the optional scatter-gather. The second subworkflow takes care of the sample merge and split.

This way we keep the subworkflows more modular and pipeline devs can choose which one they like, without introducing too much complexity in the subworkflow code.

matthdsm:
> That would work, but you could also achieve the same thing with a flag and keep it in one subworkflow?

True, but I like to limit the number of flags as much as possible to maximise user friendliness, hence the extra subworkflow so those that only want the scatter-gather and not the bottleneck of merging samples.

nvnieuwk:
> Do we consider merging VCFs a bottleneck because it has to wait for every variant call to finish?

Indeed!

tests/subworkflows/nf-core/vcf_annotate_ensemblvep/test.yml
```diff
@@ -1,24 +1,61 @@
...
+    - path: output/bcftools/test_concat.vcf.gz
+    - path: output/ensemblvep/custom_test0_annotated.summary.html
+    - path: output/ensemblvep/custom_test0_annotated.vcf.gz
+    - path: output/ensemblvep/custom_test1_annotated.summary.html
+    - path: output/ensemblvep/custom_test1_annotated.vcf.gz
+    - path: output/ensemblvep/test0_annotated.summary.html
+    - path: output/ensemblvep/test0_annotated.vcf.gz
+    - path: output/ensemblvep/test1_annotated.summary.html
+    - path: output/ensemblvep/test1_annotated.vcf.gz
+    - path: output/tabix/custom_test.vcf.gz.tbi
```
adamrtalbot, line 31: You should include a contains statement here to check the contents of some of the files (it works with .gz files out of the box).

subworkflows/nf-core/vcf_annotate_ensemblvep/meta.yml
```diff
@@ -1,18 +1,22 @@
 # yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/subworkflows/yaml-schema.json
 name: vcf_annotate_ensemblvep
-description: Perform annotation with ensemblvep and bgzip + tabix index the resulting VCF file
+description: |
+  Perform annotation with ensemblvep and bgzip + tabix index the resulting VCF file.
+  Only VCF/BCF outputs are supported for now. Other outputs can be added later.
```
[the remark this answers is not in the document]
nvnieuwk: Added!

tests/subworkflows/nf-core/vcf_annotate_ensemblvep/test.yml
nvnieuwk: Done!

tests/subworkflows/nf-core/vcf_annotate_ensemblvep/test.yml
```diff
@@ -1,24 +1,61 @@
...
+    - path: output/bcftools/custom_test0.vcf
+    - path: output/bcftools/custom_test_concat.vcf.gz
+    - path: output/bcftools/test.vcf.gz
+    - path: output/bcftools/test0.vcf
+    - path: output/bcftools/test_concat.vcf.gz
+    - path: output/ensemblvep/custom_test0_annotated.summary.html
+    - path: output/ensemblvep/custom_test0_annotated.vcf.gz
+    - path: output/ensemblvep/test0_annotated.summary.html
+    - path: output/ensemblvep/test0_annotated.vcf.gz
+    - path: output/tabix/custom_test.vcf.gz.tbi
```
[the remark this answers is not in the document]
nvnieuwk: Done!

nvnieuwk:
Hi, is this PR okay now?

subworkflows/nf-core/vcf_annotate_ensemblvep/main.nf
```diff
@@ -2,44 +2,135 @@
 // Run VEP to annotate VCF files
 //
 
-include { ENSEMBLVEP_VEP } from '../../../modules/nf-core/ensemblvep/vep/main'
-include { TABIX_TABIX    } from '../../../modules/nf-core/tabix/tabix/main'
+include { ENSEMBLVEP_VEP            } from '../../../modules/nf-core/ensemblvep/vep/main'
+include { TABIX_TABIX               } from '../../../modules/nf-core/tabix/tabix/main'
+include { BCFTOOLS_PLUGINSCATTER    } from '../../../modules/nf-core/bcftools/pluginscatter/main'
+include { BCFTOOLS_CONCAT           } from '../../../modules/nf-core/bcftools/concat/main'
+include { BCFTOOLS_SORT             } from '../../../modules/nf-core/bcftools/sort/main'
```
maxulysse, line 9:
```suggestion
include { ENSEMBLVEP_VEP         } from '../../../modules/nf-core/ensemblvep/vep/main'
include { TABIX_TABIX            } from '../../../modules/nf-core/tabix/tabix/main'
include { BCFTOOLS_PLUGINSCATTER } from '../../../modules/nf-core/bcftools/pluginscatter/main'
include { BCFTOOLS_CONCAT        } from '../../../modules/nf-core/bcftools/concat/main'
include { BCFTOOLS_SORT          } from '../../../modules/nf-core/bcftools/sort/main'
```

subworkflows/nf-core/vcf_annotate_ensemblvep_snpeff/meta.yml
```diff
@@ -0,0 +1,75 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/subworkflows/yaml-schema.json
+name: vcf_annotate_ensemblvep_snpeff
+description: |
+  Perform annotation with ensemblvep and/or snpeff and bgzip + tabix index the resulting VCF file.
+  Only VCF/BCF outputs are supported for now. Other outputs can be added later.
+  This subworkflow uses the scatter-gather method to better parallellize the VEP execution.
+  In essence this means that the input VCF is split into multiple smaller VCFs, which are all annotated
+  in parallel and merged back together after the annotation.
```
adamrtalbot, line 8:
```suggestion
   Perform annotation with ensemblvep and/or snpeff and bgzip + tabix index the resulting VCF file. This subworkflow uses the scatter-gather method to run VEP in parallel to increase throughput. The input VCF is split into multiple smaller VCFs of fixed size, which are annotated separately and concatenated back together to a single output file per sample. Only VCF/BCF outputs are currently supported.
```

subworkflows/nf-core/vcf_annotate_ensemblvep_snpeff/main.nf
```diff
@@ -0,0 +1,161 @@
...
+
+    //
+    // Run the annotation with EnsemblVEP
+    //
+
+    ch_scatter = BCFTOOLS_PLUGINSCATTER.out.scatter
+        .map { meta, vcfs ->
+            is_list = vcfs instanceof ArrayList
+            count = is_list ? vcfs.size() : 1
+            [ meta, is_list ? vcfs : [vcfs], count ]
```
adamrtalbot, line 62: I'd be tempted to leave a comment explaining what this does.

adamrtalbot (approved):

subworkflows/nf-core/vcf_annotate_ensemblvep_snpeff/main.nf
nvnieuwk: done! thanks for the feedback!

adamrtalbot (approved):

maxulysse:
Oh, you removed the old annotation subworkflows.
I think @mashehu / @mirpedrol where using them for some tests in modules.
I do like where this PR is going, but I was actually more thinking of keeping these subworkflows.
I would really love doing this spread and gather strategy outside of these modules.
Also, is there any way to have that optional?
Similarly to how we do it on the preprocessing/variant calling in sarek.
I'm thinking on how to make it work for as many people as possible

nvnieuwk:
I think re-adding the old subworkflow would be the best solution in this case then. This way you can use these if you don't want to scatter/gather and the combined workflow if you want to scatter/gather. One small workaround to not scatter is to set the `val_sites_per_chunk` value very high, but that's not an ideal solution

maxulysse:
> I think re-adding the old subworkflow would be the best solution in this case then. This way you can use these if you don't want to scatter/gather and the combined workflow if you want to scatter/gather. One small workaround to not scatter is to set the `val_sites_per_chunk` value very high, but that's not an ideal solution

In the sarek subworkflows I'm planning to add to modules (let's take https://github.com/nf-core/sarek/blob/dev/subworkflows/local/bam_variant_calling_deepvariant/main.nf as an example), we have an intervals channel that contains the number of intervals, so if it's 0 or 1, then no scatter/gather, if it's more than one, we scatter/gather.

nvnieuwk:
> In the sarek subworkflows I'm planning to add to modules (let's take https://github.com/nf-core/sarek/blob/dev/subworkflows/local/bam_variant_calling_deepvariant/main.nf as an example), we have an intervals channel that contains the number of intervals, so if it's 0 or 1, then no scatter/gather, if it's more than one, we scatter/gather.

Good idea, what would you say to if `val_sites_per_chunk` is `-1` then no scattering is performed. This would add even more complexity to the subwf which already is getting pretty complex. Scattering on intervals doesn't really make a lot of sense here since you won't get an even distribution that way.

maxulysse:
> > In the sarek subworkflows I'm planning to add to modules (let's take https://github.com/nf-core/sarek/blob/dev/subworkflows/local/bam_variant_calling_deepvariant/main.nf as an example), we have an intervals channel that contains the number of intervals, so if it's 0 or 1, then no scatter/gather, if it's more than one, we scatter/gather.
> 
> Good idea, what would you say to if `val_sites_per_chunk` is `-1` then no scattering is performed. This would add even more complexity to the subwf which already is getting pretty complex. Scattering on intervals doesn't really make a lot of sense here since you won't get an even distribution that way.

Yeah, we definitively should be doing scatter on intervals, I agree with you. but it was more to give an idea about how we dealt with a similar matter before

nvnieuwk:
> If `val_sites_per_chunk` is an empty list or null can you just if-statement around that bit?

Good idea!! I'll look into it!

subworkflows/nf-core/vcf_annotate_ensemblvep_snpeff/main.nf
```diff
@@ -0,0 +1,191 @@
...
+    val_snpeff_db               //   value: the db version to use for snpEff
+    ch_snpeff_cache             // channel: [ path(cache) ] (optional)
+    val_tools_to_use            //   value: a list of tools to use options are: ["ensemblvep", "snpeff"]
+    val_sites_per_chunk         //   value: the amount of variants per scattered VCF
+
+    main:
+    ch_versions = Channel.empty()
+
+    // Check if val_sites_per_chunk is set and scatter if it is
+    if(val_sites_per_chunk) {
```
adamrtalbot, line 30:
```suggestion
    // Check if val_sites_per_chunk is set and scatter if it is
    if( val_sites_per_chunk ) {
```

So this is ignored if you just use `null`?

nvnieuwk: yes `null`, `[]` or `false` should all work

nvnieuwk:
@maxulysse and @matthdsm can I get your approvals too so this can finally get merged 😁

maxulysse (approved):

PR #3386, opened 2023-05-05
Use nf-core quay.io for ubuntu image

drpatelh:
Given the recent changes in https://github.com/nf-core/modules/pull/3380 and in nf-core/tools v.2.8 where we specify a default registry to `quay.io`, I have now created a mirror of the ubuntu base image in the nf-core quay.io account:
https://quay.io/repository/nf-core/ubuntu?tab=tags

This will make it easier to override `docker.registry` for all containers in the pipeline if we can slowly transition to copying all active containers in use on the nf-core Dockerhub account to quay.io instead.

### DIFF
--- a/modules/nf-core/cat/fastq/main.nf
+++ b/modules/nf-core/cat/fastq/main.nf
@@ -5,7 +5,7 @@ process CAT_FASTQ {
     conda "conda-forge::sed=4.7"
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
         'https://depot.galaxyproject.org/singularity/ubuntu:20.04' :
-        'docker.io/library/ubuntu:20.04' }"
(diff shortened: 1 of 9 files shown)


maxulysse (approved):

ewels (approved):

drpatelh:
Linting appears to be failing due to https://github.com/nf-core/tools/issues/2280

We need to fix this and other linting issues and ideally get a patch release out anyway so will merge this for now.

ewels:
That is already fixed on dev - maybe we should run CI on nf-core/tools dev on the modules repo again for a bit?

Or if we're happy we could push a patch release of tools now. 'cc @mirpedrol @mashehu

drpatelh:
A patch release would be awesome this week! Just need to double-check we aren't missing any other fixes for the registry stuff. It's bank holiday here today and @adamrtalbot and I will be back tomorrow.

PR #3443, opened 2023-05-22
fix: correct alignment extension for bowtie2/align

matq007:
In the currect setup, when setting samtools to view, a `bam` file is generated instead of `sam`.

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`


modules/nf-core/bowtie2/align/main.nf
```diff
@@ -38,6 +38,7 @@ process BOWTIE2_ALIGN {
     }
 
     def samtools_command = sort_bam ? 'sort' : 'view'
+    def extension = samtools_command == 'sort' ? 'bam' : 'sam'
```
nvnieuwk, line 41: Can you get the extension from the `args2` instead? `samtools view` and `samtools sort` both have the `--output-fmt SAM|BAM|CRAM` option from which you can easily define what the output format should be

matq007: Ah, I completely missed this option. My bad, of course, this makes more sense. Closing the PR then.

matq007:
As pointed out by @nvnieuwk, `args2` can contain `--output-fmt`.

matq007:
I wrote a regex trick, tested it on internal pipeline, was able to generate SAM files after setting `ext.args2 = "--output-fmt SAM"`

modules/nf-core/bowtie2/align/main.nf
```diff
@@ -38,6 +38,7 @@ process BOWTIE2_ALIGN {
     }
 
     def samtools_command = sort_bam ? 'sort' : 'view'
+    def extension = args2.contains("--output-fmt") ? (args2 =~ /--output-fmt+\s+(\S+)/)[0][1].toLowerCase() : ""
```
nvnieuwk, line 41: `--output-fmt` also has a short form `-O` can you also make sure this accounts for that too?

nvnieuwk (commented):
Can you add some tests for the different file output types? (preferable where you mix the use of `--output-fmt` and `-O`)

matq007:
And we're ready

```bash
==================================================== 94 passed, 1651 skipped in 189.09s (0:03:09) =====================================================
```

nvnieuwk (dismissed):
Awesome! Can you just fix the linting fails? Then you can merge!

matq007:
There is only 1 linting issue for `semibin/singleeasybin`. Shouldn't that be fixed with another PR? Not relevant for this PR, I haven't touched those files.

nvnieuwk:
> There is only 1 linting issue for `semibin/singleeasybin`. Shouldn't that be fixed with another PR? Not relevant for this PR, I haven't touched those files.

It's an easy fix (it's just asking for more keywords in the meta.yml)

nvnieuwk:
Tests are still failing can you rerender the tests for those modules/subworkflows? We try not to have failing modules/subworkflows on the main branch

matq007:
I see where the problem is. The alignment workflow always produces different `md5sum`, seems like there is always some slight deviation in the file. Can I remove the md5 checksum then? I know it doesn't sounds ideal, but as long you generate the aligned file it should be fine right?

nvnieuwk:
> I see where the problem is. The alignment workflow always produces different `md5sum`, seems like there is always some slight deviation in the file. Can I remove the md5 checksum then? I know it doesn't sounds ideal, but as long you generate the aligned file it should be fine right?

Yes that should be fine! Can you also add a 'contains' section to the output files where that's possible?

nvnieuwk (approved):
Awesome! Thank you for this! You can merge when the tests pass :)

matq007:
`singleeasybin` is not failing which was ok before. Can I merge anyway?

nvnieuwk:
Yeah go ahead! Conda tests usually differ in md5sum, it's fine as long as docker and singularity pass

PR #3453, opened 2023-05-24
add module: PhiSpy

### LINKED ISSUE #3452: new module: PhiSpy
[PhiSpy](https://github.com/linsalrob/PhiSpy)

jvfe:
Adds a module for running [PhiSpy](https://github.com/linsalrob/PhiSpy), a tool for predicting phage sequences in bacterial genomes.

## PR checklist

Closes #3452

- [x] Add a resource `label`

### DIFF
--- a/modules/nf-core/phispy/main.nf
+++ b/modules/nf-core/phispy/main.nf
@@ -11,10 +11,10 @@ process PHISPY {
     tuple val(meta), path(gbk)
 
     output:
-    tuple val(meta), path("${prefix}/")                        , emit: results
-    tuple val(meta), path("${prefix}/prophage_coordinates.tsv"), emit: coordinates
-    tuple val(meta), path("${prefix}/*.gb*")                   , emit: gbk
-    path "versions.yml"                                        , emit: versions
+    tuple val(meta), path("${prefix}/")              , emit: results
+    tuple val(meta), path("${prefix}/${prefix}.tsv") , emit: coordinates
+    tuple val(meta), path("${prefix}/${prefix}*.gb*"), emit: gbk
+    path "versions.yml"                              , emit: versions
 
     when:
     task.ext.when == null || task.ext.when
@@ -26,8 +26,11 @@ process PHISPY {
     PhiSpy.py \\
         $args \\
         --threads $task.cpus \\
+        -p $prefix \\
         -o $prefix \\
         $gbk
+
+    mv ${prefix}/${prefix}_prophage_coordinates.tsv ${prefix}/${prefix}.tsv
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
--- a/modules/nf-core/phispy/main.nf
+++ b/modules/nf-core/phispy/main.nf
@@ -13,7 +13,7 @@ process PHISPY {
     output:
     tuple val(meta), path("${prefix}/")              , emit: results
     tuple val(meta), path("${prefix}/${prefix}.tsv") , emit: coordinates
-    tuple val(meta), path("${prefix}/${prefix}*.gb*"), emit: gbk
+    tuple val(meta), path("${prefix}/${prefix}.gb*"), emit: gbk
     path "versions.yml"                              , emit: versions
 
     when:
@@ -30,7 +30,12 @@ process PHISPY {
         -o $prefix \\
         $gbk
 
-    mv ${prefix}/${prefix}_prophage_coordinates.tsv ${prefix}.tsv
+    mv ${prefix}/${prefix}_prophage_coordinates.tsv ${prefix}/${prefix}.tsv
+
+    filename=$gbk
+    gbk_extension="\${filename#*.}"
+    mv ${prefix}/${prefix}_${gbk} ${prefix}/${prefix}.\${gbk_extension}
+
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
--- a/modules/nf-core/phispy/main.nf
+++ b/modules/nf-core/phispy/main.nf
@@ -11,10 +11,18 @@ process PHISPY {
     tuple val(meta), path(gbk)
 
     output:
-    tuple val(meta), path("${prefix}.tsv"), emit: coordinates
-    tuple val(meta), path("${prefix}.gb*"), emit: gbk
-    tuple val(meta), path("${prefix}.log"), emit: log
-    path "versions.yml"                   , emit: versions
+    tuple val(meta), path("${prefix}.tsv")                     , emit: coordinates
+    tuple val(meta), path("${prefix}.gb*")                     , emit: gbk
+    tuple val(meta), path("${prefix}.log")                     , emit: log
+    tuple val(meta), path("${prefix}_prophage_information.tsv"), optional:true, emit: information
+    tuple val(meta), path("${prefix}_bacteria.fasta")          , optional:true, emit: bacteria_fasta
+    tuple val(meta), path("${prefix}_bacteria.gbk")            , optional:true, emit: bacteria_gbk
+    tuple val(meta), path("${prefix}_phage.fasta")             , optional:true, emit: phage_fasta
+    tuple val(meta), path("${prefix}_phage.gbk")               , optional:true, emit: phage_gbk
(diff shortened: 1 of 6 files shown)


modules/nf-core/phispy/main.nf
```diff
@@ -0,0 +1,37 @@
...
+    script:
+    def args = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    PhiSpy.py \\
+        $args \\
+        --threads $task.cpus \\
+        -o $prefix \\
+        $gbk
+
```
nvnieuwk, line 31: Can you move all files in the output folder and rename them to contain the prefix? This makes sure that the output has a unique name

modules/nf-core/phispy/main.nf
```diff
@@ -0,0 +1,37 @@
...
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        PhiSpy: \$(echo \$(PhiSpy.py --version 2>&1))
+    END_VERSIONS
+    """
+}
```
nvnieuwk, line 37: Can you add a stub here?

modules/nf-core/phispy/main.nf
jvfe: Sure! By that you mean add a 
```
mv $prefix/prophage_coordinates.tsv {prefix}/${prefix}_prophage_coordinates.tsv 
```
something like that?

nvnieuwk: Something like that yes, maybe also drop the `_prohpage_coordinates` so the file name is completely customizable

modules/nf-core/phispy/main.nf
jvfe: Done!

modules/nf-core/phispy/main.nf
```diff
@@ -0,0 +1,53 @@
...
+    conda "bioconda::phispy=4.2.21"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/phispy:4.2.21--py310h30d9df9_1':
+        'biocontainers/phispy:4.2.21--py310h30d9df9_1' }"
+
+    input:
+    tuple val(meta), path(gbk)
+
+    output:
+    tuple val(meta), path("${prefix}/")              , emit: results
```
nvnieuwk, line 14: Do you need the whole directory as input for another module? Otherwise I would suggest emitting all files separately

tests/modules/nf-core/phispy/test.yml
```diff
@@ -0,0 +1,12 @@
+- name: phispy test_phispy
+  command: nextflow run ./tests/modules/nf-core/phispy -entry test_phispy -c ./tests/config/nextflow.config -c ./tests/modules/nf-core/phispy/nextflow.config
+  tags:
+    - phispy
+  files:
+    - path: output/phispy/test/test_genome.gbff.gz
```
nvnieuwk, line 6: Can you also remove this `_genome` suffix?

modules/nf-core/phispy/main.nf
jvfe: My thinking here is that, since PhiSpy can generate additional output files ([Reference](https://github.com/linsalrob/PhiSpy#output-files)), if someone using the module is generating these optional files via task.ext.args, they can choose to access them by using the results channel.

jvfe: I'm doing something similar in a workflow with the [panaroo/run module](https://github.com/nf-core/modules/blob/master/modules/nf-core/panaroo/run/main.nf) for instance

tests/modules/nf-core/phispy/test.yml
jvfe: Sure, that will just take a bit more work because I'll have to capture the file extension, since it's based on the file extension in the input file.

modules/nf-core/phispy/main.nf
asp8200: > > I'm doing something similar in a workflow with the [panaroo/run module](https://github.com/nf-core/modules/blob/master/modules/nf-core/panaroo/run/main.nf) for instance
> 
> I get what you mean. @nf-core/maintainers anyone got an opinion on this?

I agree with you, @nvnieuwk. Using optional output-channels seems to be the standard for something like this.

jfy133: Exporting a whole directory can be OK IMO when there is a huge number of files and nested subdirectories.

Normally in such cases I would export the directory, _and_ additional dedicated channels for specific files too.

So I agree to both having the folder and the additional optional ones :+1:

modules/nf-core/phispy/main.nf
```diff
@@ -30,7 +30,12 @@ process PHISPY {
         -o $prefix \\
         $gbk
 
-    mv ${prefix}/${prefix}_prophage_coordinates.tsv ${prefix}.tsv
+    mv ${prefix}/${prefix}_prophage_coordinates.tsv ${prefix}/${prefix}.tsv
+
+    filename=$gbk
+    gbk_extension="\${filename#*.}"
+    mv ${prefix}/${prefix}_${gbk} ${prefix}/${prefix}.\${gbk_extension}
```
jvfe, line 37: I'm thinking if it's really viable to move all files to the top-level directory. Because the gbk output, for example, is basically a duplicate of the input gbk but with added prophage regions. So if I move it up and make it the name equal to the prefix, there's a chance it overwrites the input file. Not that it's necessarily an issue since it's just a symlink, but I just just wanted to point this out. What do you think, @nvnieuwk?

tests/modules/nf-core/phispy/test.yml
jvfe: Done - https://github.com/nf-core/modules/pull/3453#discussion_r1205514864

modules/nf-core/phispy/main.nf
nvnieuwk: You can add a check for filename collisions to avoid that (a lot of modules do this actually). The docs have a nice explanation on this ([here](https://nf-co.re/docs/contributing/modules#naming-conventions) part 6)

jvfe: Oh awesome. I hadn't noticed that. Thanks.

jvfe: Done [3122ddc](https://github.com/nf-core/modules/pull/3453/commits/3122ddc5cad39c0965c5a6600a6ff99f23de6692)

modules/nf-core/phispy/main.nf
jvfe: Thank you for the suggestions, I've added optional output-channels.

modules/nf-core/phispy/main.nf
```diff
@@ -0,0 +1,68 @@
...
+    mv ${prefix}_${gbk} ${prefix}${gbk_extension}
+    mv ${prefix}_phispy.log ${prefix}.log
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        PhiSpy: \$(echo \$(PhiSpy.py --version 2>&1))
+    END_VERSIONS
+    """
+
+    stub:
```
nvnieuwk, line 56: Can you update the stub to have the filename check? Also if possible can you add conditional creation of the optional files based on the arguments in the stub?

nvnieuwk (commented):
One more comment and then you are good to go!

nvnieuwk (approved):
LGTM

modules/nf-core/phispy/main.nf
jvfe: I've added the filename check to the stub. But, conditional creation of the optional output files would be difficult to implement - since there can be multiple possible combinations of output files, so I've opted to just create all optional outputs in the stub

nvnieuwk: That's fine :) It's not mandatory (just a nice to have)

jvfe:
> 

Thank you for a very thorough review!

PR #3480, opened 2023-06-02
FIX: remove mem assignement for SAMTOOLS_SORT

maxulysse:
As it's creating lots of issues


matthdsm (approved):
Apologies for the issues 😞  it was for the greater good ™️

adamrtalbot (approved):

maxulysse:
> Apologies for the issues disappointed it was for the greater good tm

No worries, I think it's a good idea, just not the best way to do it, let's fix it for now by reverting to how it was before as it's raising issues.

And when we have more time let's find ways to make it work properly, everywhere.

PR #3499, opened 2023-06-08
Add singularity.registry = 'quay.io' for tests

drpatelh:
See https://github.com/nf-core/tools/issues/2303

Looks like we need to add `singularity.registry` for those containers that aren't available for direct download via Galaxy as these will need to be converted from Docker -> Singularity.

Updated a couple of modules here to test that it's working.

We will need to bump the default version of NF to `23.04.0` because that is where the support for `singularity.registry` was added. Added in this commit https://github.com/nextflow-io/nextflow/commit/37c1aeb999ac1b1619682da701b49e25d3024dec


drpatelh:
Linting needs to be fixed in #tools. I think it is failing because we have moved the registry definition out into a config and tools can't find the container.

tests/config/nf-test.config
```diff
@@ -14,6 +14,7 @@ process {
 profiles {
     singularity {
         singularity.enabled = true
+        singularity.registry = 'quay.io'
```
ewels, line 17: Do these need to be in profile scope? I feel like the registry info should live in the config top-level, for anyone running with their own custom configurations.

drpatelh: Good point. Think for `nf-test` the idea is to always specify some sort of `--profile` when testing because you can't pass external configs as far as I can tell. Maybe @sateeshperi can confirm. Either way, I don't think there is any harm in adding it to the top-level like we do in pipelines.

drpatelh: Done https://github.com/nf-core/modules/pull/3499/commits/ea8564e9618b7c725bd3942efe84dda0e29854d3

sateeshperi (approved):

grst (approved):

PR #3502, opened 2023-06-08
FIX: Fix memory issue with some gatk4 modules

maxulysse:
Plus minimal code polish

### DIFF
--- a/modules/nf-core/gatk4/createsequencedictionary/main.nf
+++ b/modules/nf-core/gatk4/createsequencedictionary/main.nf
@@ -20,7 +20,7 @@ process GATK4_CREATESEQUENCEDICTIONARY {
     script:
     def args = task.ext.args ?: ''
 
-    def avail_mem = 6
+    def avail_mem = 6144
     if (!task.memory) {
         log.info '[GATK CreateSequenceDictionary] Available memory not known - defaulting to 6GB. Specify process memory requirements to change this.'
     } else {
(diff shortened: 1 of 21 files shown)


asp8200 (approved):

modules/nf-core/gatk4/createsequencedictionary/main.nf
```diff
@@ -20,7 +20,7 @@ process GATK4_CREATESEQUENCEDICTIONARY {
     script:
     def args = task.ext.args ?: ''
 
-    def avail_mem = 6
+    def avail_mem = 6144
```
FriederikeHanssen, line 23: Why so much for such a small process

FriederikeHanssen (approved):

modules/nf-core/gatk4/createsequencedictionary/main.nf
asp8200: `avail_mem` was accidentally decreased from 6GB to 6MB which - as you may already know - caused some problems over in Sarek. I guess the purpose of this PR is just to get things working again; not to find the lowest still-working value of `avail_mem`.

pontus (approved):

matthdsm (approved):

modules/nf-core/gatk4/createsequencedictionary/main.nf
FriederikeHanssen: yes well aware of the problem, just questioning why has 6GB by default assigned in comparison to all others. two birds one stone and so on :)

matthdsm: The picard equivalent only has 3GB. Also, and I've said this before, why do we have duplicate modules?

nvnieuwk (approved):

maxulysse:
The failed linting left will be fixed by https://github.com/nf-core/modules/pull/3499

maxulysse:
and same for the failing test

PR #3507, opened 2023-06-08
Fix inclusion of input bams in work directory

### LINKED ISSUE #3504: umitools/dedup converts input bam from symlink to hard link

### Description of the bug

when running umitools when `process.scratch=true`, i found that when the process is completed, the input bam is not only not cleaned up from the work directory but also the input bam is actually saved as a hard link rather than a symlink. the hard link's inode doesn't even match the inode of the original bam that it is linked from! when scratch is false, it is saved as a symlink, but when scratch is true, basically a redundant and unnecessary copy of the original bam is saved and can take up a significant amount of storage.

I think this is because the output channel matches all bams
https://github.com/nf-core/modules/blob/fd08c25c759d5d1f30422e73c9be153dbf657724/modules/nf-core/umitools/dedup/main.nf#L15
and `followLinks` is by default set to true (more info here: https://www.nextflow.io/docs/latest/process.html#output-type-path)

```console
PROFILE=singularity nextflow run ./tests/modules/nf-core/umitools/dedup -entry test_umitools_dedup_paired_end_no_stats -c ./tests/config/nextflow.config -c ./tests/modules/nf-core/umitools/dedup/nextflow.config -process.scratch=true
```
and the result in the work directory is:
```
$ ll work/bc/0910497b0c75f374e4ce91f2dd0d57/
total 65K
-rw-r----- 1 myuid mygid 19K Jun  8 12:03 test.bam
-rw-r----- 1 myuid mygid 19K Jun  8 12:03 test.dedup.bam
-rw-r----- 1 myuid mygid  79 Jun  8 12:03 versions.yml
```
```


### Relevant files

this is easily reproducible. 

### System information

I am running on HPC but still used the local execution mode. I tried with multiple nextflow versions as well.

anoronh4:
Staged input bam and bai in a subfolder using `stageAs`, in order to prevent the inputs from matching the wildcard in the bam output channel.

## PR checklist

Closes #3504  

- [x] Add a resource `label`

### DIFF
--- a/modules/nf-core/umitools/dedup/main.nf
+++ b/modules/nf-core/umitools/dedup/main.nf
@@ -8,7 +8,7 @@ process UMITOOLS_DEDUP {
         'biocontainers/umi_tools:1.1.4--py38hbff2b2d_1' }"
 
     input:
-    tuple val(meta), path(bam), path(bai)
+    tuple val(meta), path(bam, stageAs: 'input/*'), path(bai, stageAs: 'input/*')
     val get_output_stats
 
     output:
--- a/modules/nf-core/umitools/dedup/meta.yml
+++ b/modules/nf-core/umitools/dedup/meta.yml
@@ -3,6 +3,7 @@ description: Deduplicate reads based on the mapping co-ordinate and the UMI atta
 keywords:
   - umitools
   - deduplication
+  - dedup
 tools:
   - umi_tools:
       description: >


nvnieuwk:
Inputs files are always excluded from the output glob pattern, so this shouldn't be a problem. You can also change the output to `${prefix}.bam` instead if you want to be 100% certain that only the output BAM is in the output channel. Using `stageAs` should be seen as a last resort option, which isn't the case here :)

anoronh4:
> Inputs files are always excluded from the output glob pattern, so this shouldn't be a problem. You can also change the output to `${prefix}.bam` instead if you want to be 100% certain that only the output BAM is in the output channel. Using `stageAs` should be seen as a last resort option, which isn't the case here :)

@nvnieuwk yea it's occurring to me that there is a problem, but not exactly what i thought. the bam isn't exactly going into the output channel. nonetheless, the file is being copied from the scratch directory instead of being cleaned up, and becoming a hard link. I will put this issue on the nextflow github and see what they say.

Issue #3505, opened 2023-06-08
[FEATURE] umitools/dedup should output log for publishing and downstream analysis


anoronh4:
### Is your feature request related to a problem? Please describe

MultiQC has a module to parse the umitools/dedup log, however the umitools/dedup module doesn't provide this in the output channels.

### Describe the solution you'd like

I'd like to add a parameter to the commandline tool to produce the log, and add an output channel for the log file.

### Describe alternatives you've considered

None. This is the way.

PR #3506, opened 2023-06-08
Add log file in output channels of umitools/dedup

anoronh4:
Added log file parameter and output channel, and updated all pytests to check for the log file output. 

## PR checklist

Closes #3505 

- [x] This comment contains a description of changes (with reason).
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [x] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [x] If necessary, include test data in your PR.
- [x] Remove all TODO statements.
- [x] Emit the `versions.yml` file.
- [x] Follow the naming conventions.
- [x] Follow the parameters requirements.
- [x] Follow the input/output options guidelines.
- [x] Add a resource `label`
- [x] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - [x] `PROFILE=docker pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [x] `PROFILE=singularity pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [x] `PROFILE=conda pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`

PR #3526, opened 2023-06-13
Umitools/dedup adding log and bug workaround

anoronh4:
## PR checklist

Closes issue #3504: Stages input bam and bai in a subfolder using stageAs, in order to prevent the inputs from matching the wildcard in the bam output channel.
Closes issue #3505: Added log file parameter and output channel, and updated all pytests to check for the log file output.
 
This PR was created to combine PRs #3506 and #3507, based on a suggestion that the two PRs should be combined for easier review. Those PRs are now closed.

- [x] Add a resource `label`

### DIFF
--- a/modules/nf-core/umitools/dedup/main.nf
+++ b/modules/nf-core/umitools/dedup/main.nf
@@ -12,7 +12,7 @@ process UMITOOLS_DEDUP {
     val get_output_stats
 
     output:
-    tuple val(meta), path("*.bam")             , emit: bam
+    tuple val(meta), path("${prefix}.bam")     , emit: bam
     tuple val(meta), path("*.log")             , emit: log
     tuple val(meta), path("*edit_distance.tsv"), optional:true, emit: tsv_edit_distance
     tuple val(meta), path("*per_umi.tsv")      , optional:true, emit: tsv_per_umi
@@ -24,9 +24,10 @@ process UMITOOLS_DEDUP {
 
     script:
     def args = task.ext.args ?: ''
-    def prefix = task.ext.prefix ?: "${meta.id}"
+    prefix = task.ext.prefix ?: "${meta.id}"
     def paired = meta.single_end ? "" : "--paired"
-    def stats = get_output_stats ? "--output-stats $prefix" : ""
+    stats = get_output_stats ? "--output-stats ${prefix}" : ""
+    if ("$bam" == "${prefix}.bam") error "Input and output names are the same, set prefix in module configuration to disambiguate!"
 
     if (!(args ==~ /.*--random-seed.*/)) {args += " --random-seed=100"}
     """
--- a/modules/nf-core/umitools/dedup/main.nf
+++ b/modules/nf-core/umitools/dedup/main.nf
@@ -45,4 +45,18 @@ process UMITOOLS_DEDUP {
         umitools: \$(umi_tools --version 2>&1 | sed 's/^.*UMI-tools version://; s/ *\$//')
     END_VERSIONS
     """
+
+    stub:
+    """
+    touch ${prefix}.bam
+    touch ${prefix}.log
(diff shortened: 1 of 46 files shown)


anoronh4:
@nvnieuwk i have combined PRs #3506 and #3507, as you suggested. Hopefully we can get this moving soon.

modules/nf-core/umitools/dedup/main.nf
```diff
@@ -8,11 +8,12 @@ process UMITOOLS_DEDUP {
         'biocontainers/umi_tools:1.1.4--py38hbff2b2d_1' }"
 
     input:
-    tuple val(meta), path(bam), path(bai)
+    tuple val(meta), path(bam, stageAs: 'input/*'), path(bai, stageAs: 'input/*')
```
nvnieuwk, line 11: Do these files have to be in a folder for this to work? Otherwise it's preferred to leave them in the top directory

nvnieuwk: There actually is a part about this in the guidelines ([here](https://nf-co.re/docs/contributing/modules#naming-conventions) part 6). In short it tells developers that a check should be in the module which gives a custom error message if the input and output file name are the same. You can easily let the user fix this by letting them adjust the `task.ext.prefix`. This is not something that should be fixed on a module level :)

nvnieuwk: (We've had a lot of discussions about this last year and this is the consensus we reached)

anoronh4: great, sounds like i'll be changing the output file to `${prefix}.dedup.bam` and having the output file match `"*.dedup.bam"`. that should work around the initial issue while keeping everything at the top level.

nvnieuwk: No you can keep it like it is now, just add the error message. The pipeline developer can add the `.dedup` suffix to the prefix if they want to themselves, which makes the module way more flexible :)

anoronh4: ok, but not the way it is now, because we want stageAs to go away right? if that can stay in then I don't need to add the naming check either...

anoronh4: i'm having trouble getting this working. i took out the stageAs options as you suggested, and added the error message. however that still leaves the problem of the output channel matching both bams. however if i change the output channel to:
```
tuple val(meta), path("${prefix}.bam")     , emit: bam
```
it breaks with the following explanation:
```
Caused by:
  Missing output file(s) `null.bam` expected by process `test_umitools_dedup_no_umi:UMITOOLS_DEDUP (test)`
```
because the output declaration can't access `$prefix`. catching issues with name clashing is important but it doesn't solve the original problem i am facing. i don't know if i can follow the naming conventions and keep everything at the top-level while solving my problem. please let me know how i should proceed.

nvnieuwk: You will just need to change the prefix in the `nextflow.config` file in the test directory :) here an example of how another modules does that: https://github.com/nf-core/modules/blob/240937a2a9c30298110753292be041188891f2cb/tests/modules/nf-core/ultra/index/nextflow.config#L7

Also you can just remove the `def` of prefix if you want to use prefix in the output declaration

Most of this information can be found in the modules [guidelines](https://nf-co.re/docs/contributing/modules), it's an interesting read which will answer most of your questions for you :)

anoronh4: ok thanks. i tried removing `def` before but got frustrated because it caused other errors related to scope. i seem to have fixed it though and pushed my changes. so now there is no `stageAs` and the issue i noticed before should now be avoided.

nvnieuwk (approved):
Nice! This is perfect! Can you also quickly add a `stub` section before merging? :)

PR #3497, opened 2023-06-07
Move containers for sarek, demultiplex and scrnaseq to quay.io

drpatelh:
I have pulled and pushed the containers in these modules to the nf-core `quay.io` account. Updating modules now, will re-install in the pipeline and create patch releases.

PR #3537, opened 2023-06-20
Automatically rename input files in cellranger module

grst:
This PR implements automatic renaming for cellranger such that it can deal with
 * fastq files that don't follow the bcl2fastq naming convention (e.g. when downloaded from a sequencing archive) 
 * samples that got sequenced across mutliple flowcells and there are filename clashes

### How does this work
 * each file gets staged into a separate folder using `stageAs 'fastq_???/*`, where `???` will expand to a numeric folder, e.g 001. 
 * A python script retrieves all files from `fastq_*` and renames them to match the naming convention. 

See https://github.com/nf-core/scrnaseq/issues/241 for more details. 
Fixes issues such as https://github.com/nf-core/scrnaseq/issues/147, https://github.com/nf-core/scrnaseq/issues/195

I demonstrated that renaming the files does not affect the results, see the newly added README. 

## PR checklist

Closes #XXX 

- [x] Add a resource `label`


fmalmeida:
Generally, it looks good. I did not have the chance to run it myself, but the proposed code looks good.

I have one question tough. What happens if we have already samples named `{sample_name}_S{i}_L00{i}_{R1,R2}_001.fastq.gz` ? I mean, I did not catch precisely if the whole 'lane' numbering is only based on the multi-staged-folders.

If so, for example, do we expect any problem to arise from mixing this and end up renaming for example a input already properly named from test_S1_L001_R1_001.fastq.gz to test_S1_L002_R1_001.fastq.gz or test_S2_L001_R1_001.fastq.gz for example?

modules/nf-core/cellranger/count/templates/cellranger_count.py
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env python
```
klkeys, line 1:
```suggestion
#!/usr/bin/env python3
```
a blatantly nitpicky but paranoid suggestion here, based on personal experience

I still encounter systems and OSes with Python 2.7 as the default executable 👎🏼

klkeys (commented):
made several typo fix + style suggested changes

this fixes a problem that has festered for a while 🎉 

only downside that I see here is that maintaining the Python script will require some more care than the original shell call

klkeys:
@nf-core-bot fix linting

grst:
Thanks for fixing all my typos 🙏 

> I have one question tough. What happens if we have already samples named `{sample_name}_S{i}_L00{i}_{R1,R2}_001.fastq.gz` ? I mean, I did not catch precisely if the whole 'lane' numbering is only based on the multi-staged-folders.

Even if the files already follow the naming convention, they would be renamed anyway. I don't see that as a problem though, as I could show (see README.md) that it makes no difference whatsoever how fastq files are named and where they are placed, as long as cellranger picks them up. 

> only downside that I see here is that maintaining the Python script will require some more care than the original shell call

I was considering making it a separate module that only does the renaming (e.g. `CELLRANGER_RENAME_FILES`). In the end, I decided against it, because afaik this would mean an additional down- and upload of the fastq files on some cloud systems.

grst:
I now explicitly specified the container registry in the `container` directive and linting is happy. Is this the intended way nowadays? 

In that case we should fix this for the other cellranger containers as well. 
Also is there any preference of `quay.io/nfcore/cellranger:7.1.0` over `docker.io/nf-core/cellranger:7.1.0`?

fmalmeida:
> Even if the files already follow the naming convention, they would be renamed anyway. I don't see that as a problem though, as I could show (see README.md) that it makes no difference whatsoever how fastq files are named and where they are placed, as long as cellranger picks them up.

Sounds good :)

fmalmeida (approved):
I think it is looking good.

About the registry, I do not have any preference.

klkeys:
> Also is there any preference of quay.io/nfcore/cellranger:7.1.0 over docker.io/nf-core/cellranger:7.1.0?

functionally this shouldn't matter, but maybe it's best to pin the explicit container registry to whichever one the nf-core org wants to maintain longest, yes? is that `quay.io`?

also, it seems like one of the pytest CI checks failed on a Singularity pull, does the choice of container registry affect this?
>Caused by:
  Failed to pull singularity image
  command: singularity pull  --name nf-core-cellranger-7.1.0.img.pulling.1687285765645 docker://nf-core/cellranger:7.1.0 > /dev/null
  status : 255
  message:
    FATAL:   While making image from oci registry: error fetching image to cache: failed to get checksum for docker://nf-core/cellranger:7.1.0: Error reading manifest 7.1.0 in docker.io/nf-core/cellranger: errors:
    denied: requested access to the resource is denied
    unauthorized: authentication required

in case it matters, the Quay registry name has a hyphen ([nf-core](https://quay.io/repository/nf-core/cellranger?tab=info)) while the DockerHub registry doesn't ([nfcore](https://hub.docker.com/r/nfcore/cellranger))

grst:
@hpatel, I see the explicit registry was removed in https://github.com/nf-core/modules/pull/3497. 
 * Why? 
 * Singularity doesn't seem to be happy with it... what's the solution to that?

modules/nf-core/cellranger/count/main.nf
```diff
@@ -2,15 +2,15 @@ process CELLRANGER_COUNT {
     tag "$meta.id"
     label 'process_high'
 
-    container "nf-core/cellranger:7.1.0"
+    container "quay.io/nf-core/cellranger:7.1.0"
```
maxulysse, line 5:
```suggestion
    container "nf-core/cellranger:7.1.0"
```

PR #3542, opened 2023-06-21
Bcftools merge update

fellen31:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`

### DIFF
--- a/modules/nf-core/bcftools/merge/main.nf
+++ b/modules/nf-core/bcftools/merge/main.nf
@@ -52,7 +52,7 @@ process BCFTOOLS_MERGE {
                     args.contains("--output-type u") || args.contains("-Ou") ? "bcf" :
                     args.contains("--output-type z") || args.contains("-Oz") ? "vcf.gz" :
                     args.contains("--output-type v") || args.contains("-Ov") ? "vcf" :
-                    "vcf.gz"
+                    "vcf"
     """
     touch ${prefix}.${extension}
 
--- a/modules/nf-core/bcftools/merge/main.nf
+++ b/modules/nf-core/bcftools/merge/main.nf
@@ -29,7 +29,7 @@ process BCFTOOLS_MERGE {
                     args.contains("--output-type u") || args.contains("-Ou") ? "bcf" :
                     args.contains("--output-type z") || args.contains("-Oz") ? "vcf.gz" :
                     args.contains("--output-type v") || args.contains("-Ov") ? "vcf" :
-                    "vcf.gz"
+                    "vcf"
 
     """
     bcftools merge \\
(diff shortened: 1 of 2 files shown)


modules/nf-core/bcftools/merge/main.nf
```diff
@@ -27,16 +27,17 @@ process BCFTOOLS_MERGE {
     def regions = bed ? "--regions-file $bed" : ""
     def extension = args.contains("--output-type b") || args.contains("-Ob") ? "bcf.gz" :
                     args.contains("--output-type u") || args.contains("-Ou") ? "bcf" :
+                    args.contains("--output-type z") || args.contains("-Oz") ? "vcf.gz" :
                     args.contains("--output-type v") || args.contains("-Ov") ? "vcf" :
                     "vcf.gz"
```
nvnieuwk, line 32:
```suggestion
                    "vcf"
```

The default is uncompressed VCF :)

modules/nf-core/bcftools/merge/main.nf
```diff
@@ -49,6 +50,7 @@ process BCFTOOLS_MERGE {
     def prefix = task.ext.prefix ?: "${meta.id}"
     def extension = args.contains("--output-type b") || args.contains("-Ob") ? "bcf.gz" :
                     args.contains("--output-type u") || args.contains("-Ou") ? "bcf" :
+                    args.contains("--output-type z") || args.contains("-Oz") ? "vcf.gz" :
                     args.contains("--output-type v") || args.contains("-Ov") ? "vcf" :
                     "vcf.gz"
```
nvnieuwk, line 55:
```suggestion
                    "vcf"
```

modules/nf-core/bcftools/merge/main.nf
fellen31: Right, of course..Thanks! Leaves me with two questions though:

- Should bcftools convert look the same?
- Why does bcftools norm have `def args = task.ext.args ?: '--output-type z'`? To not break stuff?

nvnieuwk: Yeah there the bcftools modules are a bit of a wild wild west :grin: It doesn't really matter for now I think. You can update the other modules if you want or just leave them be, but I think this is the ideal setup that's the most flexible in every case. The reason `--output-type z` has been set as default in bcftools norm is because it's usually desired to create the bgzipped VCF by default. That's something you can do here too if you want that :)

fellen31: Thanks for the clarification. I think it's good like this, my main reason for the update was to allow other input files than *.vcf.gz and this solves it :)

nvnieuwk: Perfect! Thank you for the contribution!

nvnieuwk (approved):
LGTM!

PR #3551, opened 2023-06-22
Ashlar

### LINKED ISSUE #3550: new module: ASHLAR

RobJY:
## PR checklist

Closes #3550
This PR is for the addition of the new module ASHLAR.
ASHLAR performs image Alignment by Simultaneous Harmonization of Layer/Adjacency Registration

- [X] Add a resource `label`

I tested locally using `nf-core modules create-test-yml ashlar` and `nf-core modules lint ashlar` and we passed all tests.
Code pushed to GitHub has passed the CI tests as well.
Please let me know if I can provide any additional information.

This is our first nf-core module, so we are very open to any suggestions about ways to improve it.
Thanks in advance for your help!


RobJY:
Hmm, I'm unable to add the label 'ready for review'.  Where should I request this permission?  In the modules channel in Slack?

SPPearce (changes requested):
I don't understand what the zero_uuid process is doing, can you explain?
The tool looks like it can also take additional file arguments, `ffp` and `dfp`, can these be included as additional input channels?

tests/modules/nf-core/ashlar/main.nf
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env nextflow
+
+nextflow.enable.dsl = 2
+
+include { ASHLAR } from '../../../../modules/nf-core/ashlar/main.nf'
+include { ASHLAR as ASHLAR_TILE } from '../../../../modules/nf-core/ashlar/main.nf'
+
+// we zero out the UUID of output tiff images with ZERO_UUID so we get a consistent md5sum
+include { ZERO_UUID } from './zero_uuid.nf'
```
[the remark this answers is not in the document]
RobJY: The output from ASHLAR contains a UUID that is different between runs.  To get the same md5 sum for each file we use zero_uuid.nf to set all values of the UUID to zero.  The input to zero_uuid.nf is an offset to the UUID location in the file.
Is this okay or is there a better way of handling it?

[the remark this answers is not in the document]
SPPearce: Where is the UUID? Embedded into the file somehow?
Does that mean that the tool produces a different output every time it is ran? Because if so, that might be an issue in general for pipeline reproducibility.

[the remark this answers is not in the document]
RobJY: Yes, it creates a UUID in the header of the image file output.  That is the only thing that is different between runs because date and time are used in creating the UUID.  I realize we could have just checked for the existence of the file, but since this is the only element of the file that differs, we would like to make sure the actual image contents of the file have not changed between runs and this solution gives us that.

RobJY:
> he tool looks like it can also take additional file arguments, ffp and dfp, can these be included as additional input channels?

Those are optional parameters.  Should we include those as input channels?  I thought that was only for required input files.

RobJY:
Ah, looking at the documentation again I see:
> All mandatory and optional input files MUST be included in input channel definitions.

Ok, I'll look into making that change.  Thank you for pointing it out!

SPPearce:
The only way for files to get into the process is via the input channels, so in order for those options to be used they must be a channel in.

tests/modules/nf-core/ashlar/main.nf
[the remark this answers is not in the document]
jmuhlich: The OME-XML schema used to store metadata in the TIFF header has a UUID field and Ashlar fills that field with a random UUID. We had considered making a change to Ashlar itself to support passing an explicit UUID on the command line and/or disabling UUID generation completely (the XML schema does make the UUID optional) but concluded that just patching the nf-core module tests was more expedient. If and when Ashlar makes that change, we'll remove the zero_uuid code here.

tests/modules/nf-core/ashlar/main.nf
```diff
@@ -0,0 +1,73 @@
...
+}
+
+workflow test_ashlar_all_files_dfp_ffp {
+
+    input_list =  [ [ id:'test_all' ],
+               [file(params.test_data['imaging']['ome-tiff']['cycif_tonsil_cycle1'], checkIfExists: true),
+                file(params.test_data['imaging']['ome-tiff']['cycif_tonsil_cycle2'], checkIfExists: true),
+                file(params.test_data['imaging']['ome-tiff']['cycif_tonsil_cycle3'], checkIfExists: true)] ]
+
+    ASHLAR ( input_list, [file(params.test_data['imaging']['ome-tiff']['cycif_tonsil_dfp'], checkIfExists: true)],
```
SPPearce, line 55: Just for readability, can you put each of the 3 elements onto different lines to line them up and not wrap over (and below).

tests/modules/nf-core/ashlar/main.nf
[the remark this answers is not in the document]
SPPearce: Was there any progress on updating the tool on the command line so this isn't necessary?

tests/modules/nf-core/ashlar/main.nf
RobJY: Sure, no problem.  I just pushed an update.  Is that what you had in mind?

RobJY:
> Was there any progress on updating the tool on the command line so this isn't necessary?

No, that hasn't been updated yet.  I think there are other updates to ashlar that will be released soon, but I don't believe that's one of them. 
Is it okay to merge this module and make that change to ashlar at a later date?

modules/nf-core/ashlar/main.nf
```diff
@@ -0,0 +1,54 @@
...
+    """
+    if [ "$dfp_validated" = "false" ]; then
+        echo "Error: please input only zero or one dfp files"
+        exit 1
+    fi
+
+    if [ "$ffp_validated" = "false" ]; then
+        echo "Error: please input only zero or one ffp files"
+        exit 1
+    fi
```
adamrtalbot, line 41: If you run this in the shell script, the pipeline must submit a job, stage the files, _then_ discover it doesn't work. If you move it above into the script block it will error nice and early. Here's an example: https://github.com/nf-core/modules/blob/b9172e8c26a3db5009f7872654c44587e254f094/modules/nf-core/fq/subsample/main.nf#L28C1-L42

```suggestion
    if ( !dfp_validated ) { error "Please input only zero or one dfp files" }
    if ( !ffp_validated ) { error "Please input only zero or one ffp files" }
    """
```

modules/nf-core/ashlar/main.nf
```diff
@@ -0,0 +1,54 @@
...
+    tag '$meta.id'
+    label 'process_single'
+
+    conda "bioconda::ashlar=1.17.0"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/ashlar:1.17.0--pyh5e36f6f_0' :
+        'biocontainers/ashlar:1.17.0--pyh5e36f6f_0' }"
+
+    input:
+    tuple val(meta), path(file)
```
adamrtalbot, line 11: Let's use something more descriptive than `file`. `images`?

modules/nf-core/ashlar/meta.yml
```diff
@@ -0,0 +1,42 @@
...
+      doi: "10.1093/bioinformatics/btac544"
+      licence: ["MIT"]
+
+input:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. [ id:'test' ]
+  - file_in:
```
adamrtalbot, line 21: Needs to match the process input:
```suggestion
  - file:
```

But perhaps we change both to:
```suggestion
  - images:
```

modules/nf-core/ashlar/meta.yml
```diff
@@ -0,0 +1,42 @@
...
+      pattern: "*.{ome.tiff,ome.tif,rcpnl,btf,nd2,tiff,tif,czi}"
+output:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. [ id:'test' ]
+  - tif:
+      type: file
+      description: pyramidal OME-TIFF
```
adamrtalbot, line 33:
```suggestion
      description: A pyramidal, tiled OME-TIFF file created from input images.
```

tests/modules/nf-core/ashlar/main.nf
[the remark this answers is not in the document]
adamrtalbot: I think it's OK, not ideal but acceptable. I would include the process within this file instead of using a separate file. That way I can look at a single file and see exactly how the test works.
```suggestion
process ZERO_UUID {

    input:
    val(file_in)
    val(offset)

    when:
    file_in != "versions.yml"

    script:
    def file_path = file_in[1]

    """
    echo -n "00000000-0000-0000-0000-000000000000" | dd of=$file_path bs=1 seek=$offset conv=notrunc
    """

}
```

tests/modules/nf-core/ashlar/zero_uuid.nf
```diff
@@ -0,0 +1,21 @@
+nextflow.enable.dsl=2
+
+process ZERO_UUID {
+
+    input:
+    val(file_in)
```
adamrtalbot, line 6: Shouldn't this be `path(file_in)`?

tests/modules/nf-core/ashlar/main.nf
```diff
@@ -0,0 +1,103 @@
...
+workflow test_ashlar_1_file {
+
+    input_list =  [ [ id:'test_all' ],
+               [file(params.test_data['imaging']
+                                     ['ome-tiff']
+                                     ['cycif_tonsil_cycle1'], checkIfExists: true)] ]
+
+    ASHLAR ( input_list, [], [] )
+
+    ZERO_UUID ( ASHLAR.out[0], "8390123" )
```
adamrtalbot, line 20:
```suggestion
    ZERO_UUID ( ASHLAR.out.tif, "8390123" )
```

tests/modules/nf-core/ashlar/main.nf
```diff
@@ -0,0 +1,103 @@
...
+                file(params.test_data['imaging']
+                                     ['ome-tiff']
+                                     ['cycif_tonsil_cycle2'], checkIfExists: true),
+                file(params.test_data['imaging']
+                                     ['ome-tiff']
+                                     ['cycif_tonsil_cycle3'], checkIfExists: true)] ]
+
+    ASHLAR ( input_list, [], [] )
+
+    ZERO_UUID ( ASHLAR.out[0], "25169643" )
```
adamrtalbot, line 39:
```suggestion
    ZERO_UUID ( ASHLAR.out.tif, "25169643" )
```

tests/modules/nf-core/ashlar/main.nf
```diff
@@ -0,0 +1,103 @@
...
+                file(params.test_data['imaging']
+                                     ['ome-tiff']
+                                     ['cycif_tonsil_cycle2'], checkIfExists: true),
+                file(params.test_data['imaging']
+                                     ['ome-tiff']
+                                     ['cycif_tonsil_cycle3'], checkIfExists: true)] ]
+
+    ASHLAR_TILE ( input_list, [], [] )
+
+    ZERO_UUID ( ASHLAR_TILE.out[0], "12586923" )
```
adamrtalbot, line 58:
```suggestion
    ZERO_UUID ( ASHLAR_TILE.out.tif, "12586923" )
```

tests/modules/nf-core/ashlar/main.nf
```diff
@@ -0,0 +1,103 @@
...
+                                     ['cycif_tonsil_cycle3'], checkIfExists: true)] ]
+
+    ASHLAR ( input_list, [file(params.test_data['imaging']
+                                               ['ome-tiff']
+                                               ['cycif_tonsil_dfp'], checkIfExists: true)],
+                         [file(params.test_data['imaging']
+                                               ['ome-tiff']
+                                               ['cycif_tonsil_ffp'], checkIfExists: true)] )
+
+    ZERO_UUID ( ASHLAR.out[0], "25169643" )
```
adamrtalbot, line 82:
```suggestion
    ZERO_UUID ( ASHLAR.out.tif, "25169643" )
```

tests/modules/nf-core/ashlar/main.nf
```diff
@@ -0,0 +1,103 @@
...
+                                     ['cycif_tonsil_cycle1'], checkIfExists: true)] ]
+
+    ASHLAR ( input_list, [file(params.test_data['imaging']
+                                               ['ome-tiff']
+                                               ['cycif_tonsil_dfp'], checkIfExists: true)],
+                         [file(params.test_data['imaging']
+                                               ['ome-tiff']
+                                               ['cycif_tonsil_ffp'], checkIfExists: true)] )
+
+    ZERO_UUID ( ASHLAR.out[0], "8390123" )
```
adamrtalbot, line 100:
```suggestion
    ZERO_UUID ( ASHLAR.out.tif, "8390123" )
```

tests/modules/nf-core/ashlar/zero_uuid.nf
```diff
@@ -0,0 +1,21 @@
...
+
+    when:
+    file_in != "versions.yml"
+
+    script:
+    def file_path = file_in[1]
+
+    """
+
+    echo -n "00000000-0000-0000-0000-000000000000" | dd of=$file_path bs=1 seek=$offset conv=notrunc
```
adamrtalbot, line 17: Does this edit in place? If so try to avoid that, it wont work in quite a few situations.

modules/nf-core/ashlar/main.nf
RobJY: great idea. will do. thanks!

tests/modules/nf-core/ashlar/zero_uuid.nf
RobJY: I was using `val` because I want the path to the output tif file from the ashlar module so that I can run the `dd` command on it to zero out the UUID.  Perhaps I should change `file_in` to `meta` because it's really taking a meta map as input.  Actually, now that I think of it, it seems like the emit should be changed from `tif` to `meta` because that's really what's being output from the ashlar module.  What do you think @adamrtalbot?

This is our first module, so please let me know if any of this should be done differently.  Thanks!

adamrtalbot: Use `val(meta), path(tif)`.

Nextflow understands the difference between paths and values. By using the path to the existing file, it will try to point to the old path which may have moved, been deleted or simply not be available. This will work on a 'local' system like your laptop or a server, but will break on most other infrastructure. Nextflow will already cleverly stage in the file without copying so there's no need to try and work around it.

tests/modules/nf-core/ashlar/main.nf
```diff
@@ -0,0 +1,119 @@
...
+    file_in != "versions.yml"
+
+    script:
+    def file_path = file_in[1]
+
+    """
+    echo -n "00000000-0000-0000-0000-000000000000" | dd of=$file_path bs=1 seek=$offset conv=notrunc
+    """
+
+}
```
adamrtalbot, line 25:
```suggestion
process ZERO_UUID {

    input:
    tuple val(meta), path(file_in)
    val(offset)

    when:
    file_in != "versions.yml"

    script:
    """
    echo -n "00000000-0000-0000-0000-000000000000" | dd of=$file_in bs=1 seek=$offset conv=notrunc
    """

}
```

tests/modules/nf-core/ashlar/zero_uuid.nf
RobJY: Hmm, that gives me an error on `tif`:
```
CRITICAL Error running test workflow (exit code 1)                                                                       
         N E X T F L O W  ~  version 23.04.1                                                                             
         Launching `./tests/modules/nf-core/ashlar/main.nf` [festering_marconi] DSL2 - revision: ce57a5de86              
         ERROR ~ No such variable: tif                                                                                   
                                                                                                                         
          -- Check script 'tests/modules/nf-core/ashlar/main.nf' at line: 12 or see '.nextflow.log' file for more details
```
I've pushed the changes that I made.  Any idea what I did wrong?  Thanks!

I'm also clearly not understanding the difference between ASHLAR.out and ASHLAR.out.tif.  They both seem to be the whole meta map output from ASHLAR, but it's hard for me to tell.  I running tests with `nf-core modules create-test-yml ashlar`.  I guess I should just be running each individual test with `nextflow run`.  Is there a better way to run tests so that you can see more debugging information?

tests/modules/nf-core/ashlar/main.nf
```diff
@@ -0,0 +1,118 @@
...
+
+// we zero out the UUID of output tiff images with ZERO_UUID so we get a consistent md5sum
+process ZERO_UUID {
+
+    input:
+    val(meta), path(tif)
+    val(offset)
+
+    when:
+    tif != "versions.yml"
```
adamrtalbot, line 16:
```suggestion
```

Don't think this is needed any more.

tests/modules/nf-core/ashlar/main.nf
```diff
@@ -0,0 +1,118 @@
...
+include { ASHLAR } from '../../../../modules/nf-core/ashlar/main.nf'
+include { ASHLAR as ASHLAR_TILE } from '../../../../modules/nf-core/ashlar/main.nf'
+
+// we zero out the UUID of output tiff images with ZERO_UUID so we get a consistent md5sum
+process ZERO_UUID {
+
+    input:
+    val(meta), path(tif)
+    val(offset)
+
```
adamrtalbot, line 14:
```suggestion
    input:
    tuple val(meta), path(tif)
    val(offset)
    
    output:
    tuple val(meta), path(tif), includeInputs: true, emit: parsed_tif
```

Something like this!

RobJY: I got an error on the `includeInputs`.  I haven't used that before, so I wasn't sure how to fix it.  I just pushed code with it removed and it passes all tests.  Here's the error message:
```
CRITICAL Error running test workflow (exit code 1)                                                                       
         N E X T F L O W  ~  version 23.04.1                                                                             
         Launching `./tests/modules/nf-core/ashlar/main.nf` [compassionate_payne] DSL2 - revision: db9f537d46            
         [-        ] process > test_ashlar_1_file:ASHLAR -                                                               
         ERROR ~ No such variable: includeInputs                                                                         
                                                                                                                         
          -- Check script 'tests/modules/nf-core/ashlar/main.nf' at line: 16 or see '.nextflow.log' file for more details
```
Please let me know if I should make any changes.
Thanks for all you help with this @adamrtalbot!

tests/modules/nf-core/ashlar/main.nf
RobJY: I removed this as you suggested in my last commit and it seem good.  Thanks!

tests/modules/nf-core/ashlar/main.nf
RobJY: I totally missed the `tuple` above in my testing.  Seems to work great now.  Thanks for catching my mistake!!

tests/modules/nf-core/ashlar/zero_uuid.nf
RobJY: I see your updates correcting this further down the page.  It makes much more sense now.  Thanks for explaining it!

tests/modules/nf-core/ashlar/zero_uuid.nf
RobJY: Yes, this modifies the output tif from ashlar in place.  We didn't think this would be a problem because this is just in the test code.  When will this be a problem?  Is it because of the parallelism?

If I'm not operating on the output file from ashlar I think we're back to the same problem we're trying to work around in that the UUID varies for each of the output files from ashlar and we don't get a consistent md5sum when we test.  Am I not understanding a way to solve this?

Is it okay to have this workaround until ashlar is updated to accept a UUID as an input parameter?

adamrtalbot: No - you should never use the path to the original file in Nextflow. It breaks all sorts of stuff.

You can operate on the output file in Ashlar, but if you pass it to the process as a `path()` instead of a `val()`, Nextflow will correctly stage it into the working directory, edit it and pass it to the next process. Then test the resulting file for the md5sum.

RobJY: Apologies, I think I misunderstood your previous question.  Does 'original file' above refer to the input file to ashlar?

The zero_uuid process is not operating on the input file to the ashlar module.  It is operating on the output file from the ashlar module.  It's in place in that `dd` does not generate a new output file but instead overwrites the UUID in the output file from ashlar.

RobJY: Oh, I went back to the workflow documentation and I think I see what you're saying.  I'll try to modify the code and I'll let you know if I have problems.  Thanks!!

RobJY: Oh, I was looking at adding a local variable which seems to work, but in reading over your reply again it sounds like just having `path(tif)` is sufficient.  Is this correct?

RobJY: Hi @adamrtalbot, it sounds like changing `val(tif)` to `path(tif)` in the ZERO_UUID process is all we need for this to work correctly.  Is that correct?  Thanks!

SPPearce:
Does ashlar always only produce one output .tif file?

SPPearce:
Under the assumption that ashlar always produces only one file, I have modified the module to use `sed` to zero out the UUID, instead of `dd`.

SPPearce:
Gah, didn't notice that the create-test-yml had just pointed out all the md5sums were still variable!

SPPearce (approved):
Right, I think I'm happy with this now. 
@adamrtalbot, can you check my `sed` change please.

modules/nf-core/ashlar/main.nf
```diff
@@ -0,0 +1,53 @@
...
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def dfp = opt_dfp ? "--dfp $opt_dfp" : ""
+    def ffp = opt_ffp ? "--ffp $opt_ffp" : ""
+    def num_files = images instanceof List ? images.size() : 1
+    def opt_dfp_size = opt_dfp instanceof List ? opt_dfp.size() : 1
+    def opt_ffp_size = opt_ffp instanceof List ? opt_ffp.size() : 1
+    def dfp_validated = (opt_dfp_size == 0 || opt_dfp_size == 1 || opt_dfp_size == num_files) ? true : false
+    def ffp_validated = (opt_ffp_size == 0 || opt_ffp_size == 1 || opt_ffp_size == num_files) ? true : false
```
adamrtalbot, line 31: A slight tidy suggestion, not sure it makes it better but it sure does look nice like this:
```suggestion
    def args          = task.ext.args           ?: ''
    def prefix        = task.ext.prefix         ?: "${meta.id}"
    def dfp           = opt_dfp                 ? "--dfp $opt_dfp" : ""
    def ffp           = opt_ffp                 ? "--ffp $opt_ffp" : ""
    def num_files     = images instanceof List  ? images.size()    : 1
    def opt_dfp_size  = opt_dfp instanceof List ? opt_dfp.size()   : 1
    def opt_ffp_size  = opt_ffp instanceof List ? opt_ffp.size()   : 1
    def dfp_validated = (opt_dfp_size == 0 || opt_dfp_size == 1 || opt_dfp_size == num_files) ? true : false
    def ffp_validated = (opt_ffp_size == 0 || opt_ffp_size == 1 || opt_ffp_size == num_files) ? true : false
```

modules/nf-core/ashlar/main.nf
```diff
@@ -0,0 +1,53 @@
...
+    """
+
+    ashlar \\
+        -o ${prefix}.uuid.ome.tif \\
+        $images \\
+        $args \\
+        $dfp \\
+        $ffp
+
+    sed -E 's/UUID="urn:uuid:[[:xdigit:]]{8}-[[:xdigit:]]{4}-[[:xdigit:]]{4}-[[:xdigit:]]{4}-[[:xdigit:]]{12}"/UUID="urn:uuid:00000000-0000-0000-0000-000000000000"/g' ${prefix}.uuid.ome.tif > ${prefix}.ome.tif
```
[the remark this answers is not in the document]
adamrtalbot: Do you want to just remove the UUID instead of replacing it with zeros? Does that break stuff?

adamrtalbot (approved):
What a truly hideous regex. Still, it achieves what is necessary quickly and efficiently. I've dropped a few comments but none are deal breaking.

modules/nf-core/ashlar/main.nf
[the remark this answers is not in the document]
SPPearce: I have no idea whether any downstream tool may care about it or not.
I considered just finding and replacing everything after the `urn:uuid`, but didn't want to accidentally match anything else.

tests/modules/nf-core/ashlar/zero_uuid.nf
adamrtalbot: Yes - you should use path(tif) instead of the input. Then make sure to pass it as an output (with `includeInputs: true`). This makes sure Nextflow handles it properly and there are no janky workarounds which will break when used in certain contexts.

RobJY:
> Does ashlar always only produce one output .tif file?

It will produce one .ome.tif if the output path ends in .ome.tif, but it will output multiple single-channel files if the output path ends in .tif.

Here's a link to the repo if you'd like more info:
https://github.com/labsyspharm/ashlar

SPPearce:
> > Does ashlar always only produce one output .tif file?
> 
> It will produce one .ome.tif if the output path ends in .ome.tif, but it will output multiple single-channel files if the output path ends in .tif.
> 
> Here's a link to the repo if you'd like more info: https://github.com/labsyspharm/ashlar

Ok, the way that I am removing the UUID here will only do one file, but it is forcing that by using the `ome.tif` suffix. Do you expect people to want to make multiple single-channel files?

RobJY:
This should be fine.  We don't need to support multi-file output in the module.

RobJY:
In discussion we did come up with one thing that will need to be changed though.

We zero'd out the UUID in the testing just to quickly get matching md5 sums.  We shouldn't actually output all zeros in the UUID from the ashlar module though.  That may cause problems for users who make use of the UUID.

Could you change your sed command to replace all of the UUID field with spaces?  For example all of
`UUID="urn:uuid:00000000-0000-0000-0000-000000000000"` should be replaced with spaces.

RobJY:
I just pushed a proposed change.  It seems to work, but please let me know if there's a better way to do it.  Thanks!

RobJY:
Hmm, I spoke too soon. It passed my quick local test, but is failing the CI tests. Looking into it now.

RobJY:
Ah, I see I neglected to push the associated test.yml file with my last change and that was causing the problem.  I've got that fixed and we're passing all tests now.

RobJY:
@SPPearce and @adamrtalbot I see that the 'Merge when ready' is now green, so we're able to merge.  Thank you again for all your help with this PR!  Please let me know if you have any concerns about the last change I just made.

I'll reach out to our team for a final review and hopefully merge soon.

modules/nf-core/ashlar/main.nf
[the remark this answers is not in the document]
jmuhlich: @SPPearce Can you comment on why you changed from `sed -i` (in-place) to operating on a copy? These images can be hundreds of GBs so we try to avoid making copies where possible.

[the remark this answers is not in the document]
SPPearce: The `-i` flag didn't seem to be working on the CI for some reason. I definitely agree that would be better.

One thought is that the metadata is at the very end of the tif from what I could see, so there might be some clever sed (or similar) magic that could be used to start from the end of a file and only make one change.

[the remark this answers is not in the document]
jmuhlich: Oh, the test.yml was broken in your commit that failed CI. The file tests were changed from `md5sum` to `contains` with the placeholder TODO message. That's since been fixed so we're going to try again with `sed -i`.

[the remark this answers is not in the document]
RobJY: It now passes all tests.
Thank you for figuring that out @jmuhlich!

SPPearce:
Ahh, apparently I thought I had left a couple of comments but they are actually pending as a review 🤦 . 
I'm happy for this to be merged now, my one question would be whether it would be better to remove the entire UUID field rather than leave it empty (or at least leave it empty rather than a set of spaces). 
And the `sed -i` change is better, makes sense that it wasn't working on the CI if I hadn't recommitted the test yml afterwards.

SPPearce:
> TIFF files are binary with lots of file offsets stored in the headers. While this XML string happens to be at the very end for now, it could theoretically be at any offset in the file and there could be more data after it. We can't shift the XML content around without potentially breaking the offsets to other data elements.

Ok, fair enough, that's a good reason to leave it as spaces.

jmuhlich:
To explain more -- the UUID=... text is not a "field", it's an XML attribute inside a longer XML document embedded in the TIFF metadata. We can't delete everything after the UUID because that would damage the XML. We also need to keep the replacement text the exact same length as the matched text so as not to break offsets to other data elements that might be placed after the XML content.

RobJY:
Merging now.  Thanks everyone!!

PR #3555, opened 2023-06-24
Adding plasflow module

limrp:
A new module for the cram-size command of the samtools tool has been added.


modules/nf-core/samtools/cramsize/main.nf
```diff
@@ -0,0 +1,35 @@
...
+    """
+    samtools \\
+        cram-size \\
+        $args \\
+        -o ${cram.baseName}.size \\
+        $cram
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        : \$(echo \$(samtools --version 2>&1) | sed 's/^.*samtools //; s/Using.*\$//' ))
```
nvnieuwk, line 32:
```suggestion
        samtools: \$(echo \$(samtools --version 2>&1) | sed 's/^.*samtools //; s/Using.*\$//' ))
```

modules/nf-core/samtools/cramsize/main.nf
```diff
@@ -0,0 +1,35 @@
...
+    output:
+    tuple val(meta), path("*.size"), emit: size
+    path "versions.yml"            , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+
```
nvnieuwk, line 22:
```suggestion
    def prefix = task.ext.prefix ?: "$meta.id"
```
You should use prefix for your output files. This way the module is set up the most flexible

modules/nf-core/samtools/cramsize/main.nf
```diff
@@ -0,0 +1,35 @@
...
+    """
+    samtools \\
+        cram-size \\
+        $args \\
+        -o ${cram.baseName}.size \\
```
nvnieuwk, line 27:
```suggestion
        -o ${prefix}.size \\
```

modules/nf-core/samtools/cramsize/main.nf
```diff
@@ -0,0 +1,35 @@
...
+        $cram
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        : \$(echo \$(samtools --version 2>&1) | sed 's/^.*samtools //; s/Using.*\$//' ))
+    END_VERSIONS
+    """
+}
```
nvnieuwk, line 35: Can you add a stub section?

tests/modules/nf-core/samtools/cramsize/test.yml
```diff
@@ -0,0 +1,13 @@
+## TODO nf-core: Please run the following command to build this file:
+#                nf-core modules create-test-yml /cramsize
```
nvnieuwk, line 2: Can you generate the test yaml file?

modules/nf-core/samtools/cramsize/main.nf
limrp: Thanks a lot for your comments and suggestions @nvnieuwk!!!

modules/nf-core/samtools/cramsize/main.nf
limrp: Thanks a lot!!!

modules/nf-core/samtools/cramsize/main.nf
limrp: Thanks for all your advice @nvnieuwk ! I have just added a stub section and a test.yml file. I will be very grateful if you can review it, please. 

Thanks a lot in advance!

tests/modules/nf-core/samtools/cramsize/test.yml
limrp: Thanks for all your advice @nvnieuwk ! I have just added a stub section and a test.yml file. I will be very grateful if you can review it, please.

Thanks a lot in advance!

nvnieuwk:
Hi some tests are still failing, can you fix these before I do a final review? Also it's easier to get reviews when you have one module per PR (don't worry about it for now, just some advice for the future :wink:)

limrp:
> Hi some tests are still failing, can you fix these before I do a final review? Also it's easier to get reviews when you have one module per PR (don't worry about it for now, just some advice for the future wink)

Thank you for checking! I'll review the test that are failing :smiley:

limrp:
It seems I made another mistake again. My changes of the cramsize module are there, but with a lot of other modules that are not mine. 
I think it will be better to close this PR, make my changes in a new branch (like the documentation suggest) and do a new PR.

nvnieuwk (commented):
Tests for `plasflow` are still failing if you fixed those I'll give you a :heavy_check_mark: :p

limrp:
> Tests for `plasflow` are still failing if you fixed those I'll give you a ✔️ :p

Hi @nvnieuwk ! 
The only tests that are failing are the ones related to Plasflow from Conda. There is no issue with Plasflow from Docker and Singularity. There is also no problem with Samtools' cramsize.

I attempted to fix Plasflow's recipe in the Bioconda repository. I needed to specify the required Python version (3.5) for Plasflow to function.

Here is the link: https://github.com/bioconda/bioconda-recipes/pull/42697

However, I was given the following message:

> DEPRECATION: Python 3.5 reached the end of its life on September 13th, 2020. Please upgrade your Python as Python 3.5 is no longer maintained. pip 21.0 will drop support for Python 3.5 in January 2021. pip 21.0 will remove support for this functionality.
> ERROR: tensorflow-0.10.0rc0-cp35-cp35m-linux_x86_64.whl is not a supported wheel on this platform.

It seems that it will not be possible to make Plasflow work with conda because conda will not support Python 3.5 soon. But everything else works fine.
Do you think that in a case like this is it possible to accept the PR?

nvnieuwk:
Can you add the tool to be excluded from the conda tests in `.github/workflows/test.yml`? You'll see it in the `pytest` clause. This way conda won't be tested (this should only be used for tools that absolutely don't work with conda though). After that all tests should pass and you'll be ready to go. Thanks for all the effort in trying to fix it

limrp:
> Can you add the tool to be excluded from the conda tests in `.github/workflows/test.yml`? You'll see it in the `pytest` clause. This way conda won't be tested (this should only be used for tools that absolutely don't work with conda though). After that all tests should pass and you'll be ready to go. Thanks for all the effort in trying to fix it

Hi @nvnieuwk ! Thanks for your reply!

Are you refering to this section of the `test.yml` file?

image

Should I add Plasflow to that list?

Thanks a lot for all you advice!

nvnieuwk:
Yes, that's it!

SPPearce:
I'm going to close this for now, as there seems to be no response from @limrp .
I moved the samtools cramsize module out into a separate PR and merged that, but I can't immediately see a way to get plasflow to work properly

PR #3558, opened 2023-06-26
Bump bases2fastq

edmundmiller:
We're a couple of versions behind 😬 

@blajoie

### DIFF
--- a/modules/nf-core/bases2fastq/main.nf
+++ b/modules/nf-core/bases2fastq/main.nf
@@ -2,20 +2,20 @@ process BASES2FASTQ {
     tag "$meta.id"
     label 'process_high'
 
-    container "docker.io/elembio/bases2fastq:1.4.0"
+    container "docker.io/elembio/bases2fastq:1.8.0"
 
     input:
     tuple val(meta), path(run_manifest), path(run_dir)
 
     output:
-    tuple val(meta), path('output/Samples/*/*_R*.fastq.gz'), emit: sample_fastq
-    tuple val(meta), path('output/Samples/*/*.json')       , emit: sample_json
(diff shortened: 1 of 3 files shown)


Emiller88:
It's throwing an error sadly:
```
    what():  Run manifest file doesn't exist /home/runner/pytest_workflow_3q14yxch/bases2fastq/work/f8/4af9958a781006f17eeb699470581e/sim-data/RunManifest.json```

blajoie:
Andrew A and I will look @Emiller88  !

blajoie:
@Emiller88  - looks like we need to update the test-data being used here:
https://github.com/nf-core/test-datasets/raw/demultiplex/testdata/sim-data/sim-data.tar.gz

Can we be given access?  Or we can provide you a new tar ball and you could PR that in?  Then the test should succeed !

blajoie:
pushed a fix @Emiller88  - have a look!

Emiller88:
Currently waiting on the elembio quay account to be provisioned so we can point the docker image there, using docker hub breaks the linting.

maxulysse:
there's some conflicts to solve I'm afraid

SPPearce:
Is this a Pr that could be helped along with seqera containers to make the image?

edmundmiller:
No sadly, the opposite. I was chatting with @blajoie via slack and they'd prefer to keep everyone using their docker image for various reasons.

https://hub.docker.com/r/elembio/bases2fastq

I'll update the image to use that image and we can wrap this up!

SPPearce (approved):
Add the versions channel to the nf-test to pass the linting and it is good to go.

PR #3565, opened 2023-06-28
Version updates and yaml changes

DLBPointon:
Updated the versions and md5sums of 
bedtools map
busco <- just yaml def
windowmasker <- also changed yaml so it is acurately checking its own output file md5sum
seqtk cutn


JoseEspinosa (approved):
LGTM! 🚀

PR #3579, opened 2023-06-30
Update genmap modules

nvnieuwk:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`


nvnieuwk:
@jianhong can you also have a look at these since you are the original author?

maxulysse (approved):

modules/nf-core/genmap/index/main.nf
```diff
@@ -1,29 +1,44 @@
 process GENMAP_INDEX {
-    tag "$fasta"
-    label 'process_high'
+    tag "$meta.id"
```
jianhong, line 2: Is meta data is required for every module now?

modules/nf-core/genmap/index/main.nf
```diff
@@ -1,29 +1,44 @@
...
+    label 'process_low'
```
jianhong, line 3: The index process will take lots of resources like any other mapping tool. I suggest to put it as `process_high`

modules/nf-core/genmap/map/main.nf
```diff
@@ -0,0 +1,66 @@
...
+    tuple val(meta), path("*.txt")      , optional:true, emit: txt
+    tuple val(meta), path("*.csv")      , optional:true, emit: csv
+    path "versions.yml"                 , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "$meta.id"
```
jianhong, line 26: Will it better to also including meta2 info?
```suggestion
    def prefix = task.ext.prefix ?: meta2.id? "${meta.id}_${meta2.id}" : "$meta.id"
```

modules/nf-core/genmap/map/main.nf
```diff
@@ -0,0 +1,66 @@
...
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        genmap: \$(genmap --version 2>&1 | sed 's/GenMap version: //; s/SeqAn.*\$//')
+    END_VERSIONS
+    """
+
+    stub:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "$meta.id"
```
jianhong, line 47: similar to above.

jianhong (commented):
Looks good to me.

modules/nf-core/genmap/index/main.nf
nvnieuwk: Not necessarily but this module has reference data as input, which we are trying to add meta to. It's also more user friendly to be able to pass some meta along with the files.

modules/nf-core/genmap/index/main.nf
nvnieuwk: Ah okay, I assumed this because there is no way to set how many resources the tool uses

modules/nf-core/genmap/map/main.nf
nvnieuwk: Not entirely because the second input is optional :)

PR #3593, opened 2023-07-04
Use `gzip -cd` because `gunzip` preserves the group ownership

muffato:
Hello,

I've found out that `gunzip` creates the uncompress file under the same group ownership as the compressed file. This may cause some issues when the work directory is on a filesystem where this group doesn't have a quota. It also breaks the assumption that all the files created by Nextflow are under the primary group of the user (or the group of the holding directory if the setgid permission is used) and is painful to debug.

Here, I propose passing through the shell to get the file created to guarantee the group ownership follows the usual rules.

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`

### DIFF
--- a/modules/nf-core/gunzip/main.nf
+++ b/modules/nf-core/gunzip/main.nf
@@ -21,6 +21,9 @@ process GUNZIP {
     def args = task.ext.args ?: ''
     gunzip = archive.toString() - '.gz'
     """
+    # Not calling gunzip itself because it creates files
+    # with the original group ownership rather than the
+    # default one for that user / the work directory
     gzip \\
         -cd \\
         $args \\
--- a/modules/nf-core/gunzip/main.nf
+++ b/modules/nf-core/gunzip/main.nf
@@ -21,10 +21,11 @@ process GUNZIP {
     def args = task.ext.args ?: ''
     gunzip = archive.toString() - '.gz'
     """
-    gunzip \\
-        -f \\
+    gzip \\
+        -cd \\
         $args \\
-        $archive
+        $archive \\
+        > $gunzip
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":


maxulysse:
yeah, but then it's not a gunzip module anymore, but a gzip one

muffato:
> yeah, but then it's not a gunzip module anymore, but a gzip one

One may argue it was already the case as `gunzip` is just a shell script that runs `gzip -d` (at least on Ubuntu) 😉

maxulysse:
> > yeah, but then it's not a gunzip module anymore, but a gzip one
> 
> One may argue it was already the case as `gunzip` is just a shell script that runs `gzip -d` (at least on Ubuntu) wink

Then let's rename the module ;-)

muffato:
That doesn't make sense to me: I don't see how someone wanting to uncompress a file would even consider using a module named `gzip`. Similarly, the `untar` module runs the `tar` command, but it's still called "untar" because that's the name of the logical operation it does.

maxulysse:
Good point

JoseEspinosa (approved):
I agree it makes sense should we add a comment in the code for future reference?

muffato:
> I agree it makes sense should we add a comment in the code for future reference?

Oh yes, let me do that. Otherwise someone will see the module and won't understand why it's not a straightforward `gunzip`.

modules/nf-core/gunzip/main.nf
```diff
@@ -21,10 +21,11 @@ process GUNZIP {
     def args = task.ext.args ?: ''
     gunzip = archive.toString() - '.gz'
     """
-    gunzip \\
-        -f \\
+    gzip \\
```
muffato, line 24:
```suggestion
    # Not calling gunzip itself because it creates files
    # with the original group ownership rather than the
    # default one for that user / the work directory
    gzip \\
```

PR #3595, opened 2023-07-04
New module:  mmseqs/search

Joon-Klaps:
## PR checklist
New module mmseqs search https://github.com/soedinglab/MMseqs2/wiki#searching

Closes [#2332 ](https://github.com/nf-core/modules/issues/2332)

- [X] Add a resource `label`


modules/nf-core/mmseqs/search/main.nf
```diff
@@ -0,0 +1,66 @@
...
+    script:
+    def args = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "${meta.id}"
+
+    if ("$db_query" == "${prefix}" || "$db_target" == "${prefix}"  ) error "Input and output names of databases are the same, set prefix in module configuration to disambiguate!"
+    """
+    mkdir -p ${prefix}
+
+    # Extract basename of DB
+    DB_QUERY_PATH_NAME=\$(find -L "$db_query/" -maxdepth 1 -name "*.dbtype" | sed -e 'N;s/^\\(.*\\).*\\n\\1.*\$/\\1\\n\\1/;D' )
```
JoseEspinosa, line 31: This will work in the case that the DBs have been generated by the `MMSEQS_CREATEDB` module but could it be that if the DB has been generated otherwise, this line is not general enough?

Joon-Klaps: I'm not certain about that, i think it's quite general. It might not work if the same directory contains multiple mmseqs databases. The `sed` command will identify the longest common substring which corresponds to the sequence database.  
What do u suggest to change to make it more general? Looking for the `.dbtype` is also the the strategy of  mmseqs (see all the `if notExists` statements https://github.com/soedinglab/MMseqs2/blob/f5f780acd64482cd59b46eae0a107f763cd17b4d/data/workflow/nucleotide_clustering.sh#L53). 
Also for the module `mmseqs/cluster` I used the nf-core test database of mmseqs and it is still able to identify the sequence database as it is the one with the shortest string there. 
Maybe the best solution here is to add a `args2` & `args3` to define a search pattern?

JoseEspinosa: I think that if looking for the `.dbtype` is the strategy of mmseqs it should be fine. Actually, I was referring more to the fact of calling the directory and the db files the same way which could not be always the case is it?

JoseEspinosa (approved):
LGTM Nice job! :rocket:

JoseEspinosa (approved):
LGTM Great job!

PR #3610, opened 2023-07-07
Update md5sum module to be able to take multiple files

SPPearce:
Update md5sum module to loop over multiple files when given them, for instance R1 and R2 of a fastq pair.
Someone with more linux-fu than me should probably check that it will work with files with spaces in their names etc.

I have made the branch off a previous PR apparently, rather than the main, which is why there are some spurious commits in the history.


mashehu (approved):

muffato (commented):
Hi @SPPearce . I would suggest an alternative mode whereby there is a single `.md5` file with the checksums of _all_ input files.

SPPearce:
> Hi @SPPearce . I would suggest an alternative mode whereby there is a single `.md5` file with the checksums of _all_ input files.

As an optional mode controlled by a value input?

muffato:
> > Hi @SPPearce . I would suggest an alternative mode whereby there is a single `.md5` file with the checksums of _all_ input files.
> 
> As an optional mode controlled by a value input?

I think that'd be the way to go given my understanding of the nf-core guidelines.

modules/nf-core/md5sum/main.nf
```diff
@@ -22,10 +22,16 @@ process MD5SUM {
...
+    do
+
     md5sum \\
         $args \\
-        ${file} \\
-        > ${file}.md5
+        \${FILE} \\
+        > "\${FILE}.md5"
+
+    done
```
mahesh-panchal, line 34: Here's my suggestion based on Matthieu's suggestion
```bash
    input:
    tuple val(meta), path(files)
    val as_separate_files

    script:
    if ( as_separate_files ) {
        """
        find -L * -type f \\
            ! -name '*.md5' \\
            -exec sh -c 'md5sum "\$1" > "\$1.md5"' _ "{}" \\;
        """
    } else {
        """
        find -L * -type f \\
            ! -name '*.md5' \\
            -exec md5sum "{}" + \\
            > all.md5
        """
    }

    output:
    tuple val(meta), path("*.md5"), emit: checksum

```
Only caveat is that `files` must be files and not directories, otherwise the first instance will write out md5's to input subdirectories.
This should also handle spaces.

SPPearce:
@mahesh-panchal, @muffato, I've incorporated Mahesh's code and the ability to generate a single file.

modules/nf-core/md5sum/main.nf
```diff
@@ -19,29 +20,55 @@ process MD5SUM {
...
+            -exec sh -c 'md5sum $args "\$1" > "\$1.md5"' _ "{}" \\;
 
-    """
-    md5sum \\
-        $args \\
-        ${file} \\
-        > ${file}.md5
+        cat <<-END_VERSIONS > versions.yml
+        "${task.process}":
+            md5sum: \$(echo \$(md5sum --version 2>&1 | head -n 1| sed 's/^.*) //;' ))
```
mahesh-panchal, line 32:
```suggestion
            md5sum: \$( md5sum --version | sed '1!d; s/.* //' )
```

muffato: Same in the other occurrences below

modules/nf-core/md5sum/main.nf
```diff
@@ -19,29 +20,55 @@ process MD5SUM {
 
     script:
     def args = task.ext.args ?: ''
-    def prefix = task.ext.prefix ?: "${meta.id}"
+    def prefix = task.ext.prefix ?: "${meta.id}" // will only use when as_separate_files = false
+    if ( as_separate_files ) {
+        """
+        find -L * -type f \\
```
muffato, line 26: I would suggest this to make it explicit that only input _files_ are supported.
```suggestion
        find -L * -maxdepth 0 -type f \\
```

SPPearce:
Thanks both!

PR #3627, opened 2023-07-12
[pairtools/sort] Update memory usage

nservant:
Update the `pairtools/sort` memory usage.  
This tools is running a classical `sort` command in background, and define how much memory it can use.
However, for large dataset, it seems that `pairtools/sort` goes beyond memory.  
I guess that `pairtools` itself is using part of the memory in addition to the `sort` sub-command.

A simple fix here is to give to `pairtools/sort` soemthing like 80% of the authorized memory to avoid memory crash


nservant:
@jianhong , looks all good now. Thanks

jianhong (approved):

PR #3649, opened 2023-07-18
Hisat2-build: index construction w/ truly optional inputs 

### LINKED ISSUE #3648: Hisat2 build optional inputs

### Description of the bug

The Hisat2 build module decides whether to use information about splicing based on the available memory as higher memory is required to build splice aware indices. However, if this memory requirement is satisfied then splicesite and annotation files are made required arguments rather than optional. 

I suspect the optimal functionality would be for these arguments to be optional regardless of memory availability

_No response_

### Relevant files

_No response_

### System information

_No response_

JackCurragh:
## PR checklist

Closes #3648 

I have adapted the existing module to only use these optional inputs if they are provided. Previously the `--exons` and `--ss` parameters were optional dependent on available memory. Simpler indices were constructed if the memory requirement was not satisfied. This PR makes it so that nf-core users can intentionally build these 'non-splice aware' indices if they so choose (example use case is where there is no splicing).

### DIFF
--- a/tests/modules/nf-core/hisat2/build_test/main.nf
+++ b/tests/modules/nf-core/hisat2/build_test/main.nf
@@ -7,10 +7,10 @@ include { HISAT2_BUILD              } from '../../../../../modules/nf-core/hisat
 
 workflow test_hisat2_build {
     fasta = [ [id:'genome'],
-              file(params.test_data['homo_sapiens']['genome']['genome_fasta'], checkIfExists: true)
+              file(params.test_data['sarscov2']['genome']['genome_fasta'], checkIfExists: true)
             ]
     gtf   = [ [id:'genome'],
-              file(params.test_data['homo_sapiens']['genome']['genome_gtf'], checkIfExists: true)
+              file(params.test_data['sarscov2']['genome']['genome_gtf'], checkIfExists: true)
             ]
     HISAT2_EXTRACTSPLICESITES ( gtf )
     HISAT2_BUILD ( fasta, gtf, HISAT2_EXTRACTSPLICESITES.out.txt )
@@ -18,21 +18,21 @@ workflow test_hisat2_build {
 
 workflow test_hisat2_build_fasta_only {
     fasta = [ [id:'genome'],
-              file(params.test_data['homo_sapiens']['genome']['genome_fasta'], checkIfExists: true)
+              file(params.test_data['sarscov2']['genome']['genome_fasta'], checkIfExists: true)
             ]
     gtf   = [ [id:'genome'],
-              file(params.test_data['homo_sapiens']['genome']['genome_gtf'], checkIfExists: true)
+              file(params.test_data['sarscov2']['genome']['genome_gtf'], checkIfExists: true)
             ]
-    HISAT2_BUILD ( fasta, [[],[]], [[],[]] )
+    HISAT2_BUILD ( fasta, [[:],[]], [[:],[]] )
 }
 
 workflow test_hisat2_build_fasta_ss_only {
     fasta = [ [id:'genome'],
-              file(params.test_data['homo_sapiens']['genome']['genome_fasta'], checkIfExists: true)
+              file(params.test_data['sarscov2']['genome']['genome_fasta'], checkIfExists: true)
             ]
     gtf   = [ [id:'genome'],
-              file(params.test_data['homo_sapiens']['genome']['genome_gtf'], checkIfExists: true)
+              file(params.test_data['sarscov2']['genome']['genome_gtf'], checkIfExists: true)
             ]
     HISAT2_EXTRACTSPLICESITES ( gtf )
-    HISAT2_BUILD ( fasta, [[],[]], HISAT2_EXTRACTSPLICESITES.out.txt )
+    HISAT2_BUILD ( fasta, [[:],[]], HISAT2_EXTRACTSPLICESITES.out.txt )
 }
--- a/tests/modules/nf-core/hisat2/build_test/main.nf
+++ b/tests/modules/nf-core/hisat2/build_test/main.nf
@@ -23,7 +23,6 @@ workflow test_hisat2_build_fasta_only {
     gtf   = [ [id:'genome'],
               file(params.test_data['homo_sapiens']['genome']['genome_gtf'], checkIfExists: true)
             ]
-    HISAT2_EXTRACTSPLICESITES ( gtf )
     HISAT2_BUILD ( fasta, [[],[]], [[],[]] )
 }
 
--- a/tests/modules/nf-core/hisat2/build_test/main.nf
+++ b/tests/modules/nf-core/hisat2/build_test/main.nf
@@ -15,3 +15,35 @@ workflow test_hisat2_build {
     HISAT2_EXTRACTSPLICESITES ( gtf )
     HISAT2_BUILD ( fasta, gtf, HISAT2_EXTRACTSPLICESITES.out.txt )
 }
+
+workflow test_hisat2_build_fasta_only {
+    fasta = [ [id:'genome'],
+              file(params.test_data['sarscov2']['genome']['genome_fasta'], checkIfExists: true)
+            ]
+    gtf   = [ [id:'genome'],
+              file(params.test_data['sarscov2']['genome']['genome_gtf'], checkIfExists: true)
+            ]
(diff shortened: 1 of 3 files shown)


jfy133 (changes requested):
Please could you add more tests to check this is working?

JackCurragh:
I actually completely missed that this was adding new functionality so didn't realise I needed tests. Getting the hang of things now...

jfy133:
I guess it's not _new_ functionality, but if the tests had been written correctly in the first place the issue would've been picked up, so we do the good work here ;)

JackCurragh:
Okay yes I get you. So now Linting is failing as the `test.yml` contains md5sums for empty files. These empty files are the output of `hisat2_extract_exons.py` on the[ sarscov2 gtf](https://github.com/nf-core/test-datasets/blob/modules/data/genomics/sarscov2/genome/genome.gtf) which does not contain any exons. So this is in fact the desired behaviour. 

This is failing for the one original test and then 2 of the 3 that I have added so I am not sure how this was passing the linting tests previously. Any idea how I should handle this? Change the test dataset? [For example.](https://github.com/nf-core/test-datasets/blob/modules/data/genomics/homo_sapiens/genome/genome.gtf)

jfy133 (commented):
Could you 'clean up' the test.yml so it only has the files that come out of BUILD (i.e., not splicesites)?

And are your previous comments still valid now you've swwitched from SARS-COV to human?

jfy133 (commented):
OK linting fail is API rate, but that's OK.

But I'm now concerned that `genome.splice_sites.txt` is not appearing in any of the test output not?


What's going on there... or was this module not working in the first place :eyes: 

We need someoneone from a pipeline using the module to help here I think...

@BarryDigby @drpatelh (but they might be on holiday, so might take a bit).

What you could do in the meantime for your own pipeline is to copy and paste the changes to `main.nf` inyour your pipeline's copy of the `main.nf` and run `nf-core modules patch hisat2/build` until this is solved.

JackCurragh:
Oh, I removed that splicesite file from the test output because I thought that was what you were suggesting previously! Those files don't come from the build module itself. 

My understanding is that the intermediate processes (eg. HISAT2_EXTRACTSPLICESITES) should be tested in their own module tests so then just this modules outputs should be checked when testing these modules (eg. HISAT2_BUILD) 

I can add them back if you prefer @jfy133? 

p.s thank you for your patience :)

tests/modules/nf-core/hisat2/build_test/main.nf
```diff
@@ -7,11 +7,33 @@ include { HISAT2_BUILD              } from '../../../../../modules/nf-core/hisat
...
 }
+
+workflow test_hisat2_build_fasta_only {
+    fasta = [ [id:'genome'],
+              file(params.test_data['homo_sapiens']['genome']['genome_fasta'], checkIfExists: true)
+            ]
+    gtf   = [ [id:'genome'],
+              file(params.test_data['homo_sapiens']['genome']['genome_gtf'], checkIfExists: true)
+            ]
+    HISAT2_EXTRACTSPLICESITES ( gtf )
```
jfy133, line 26: Not needed to be executed here if not passed to HISAT2_BUILD in this test (and `gtf` could be removed I guess)

jfy133 (approved):
Ahh sorry, I guess that the original test.yml was bad then :sweat_smile:  (I guess I wasn't concentrating properly, my bad)! Yes my original request was still correct! I thin this looks OK to me now - Harshil or other RNA people can shout at me later if this is wrong ;)

JackCurragh:
Okay, @jfy133, we are getting there! I have removed that unnecessary call from the tests, changes were approved and have all green ticks on actions. 

I think this means I can merge but just want to double check (and will know for next time)

tests/modules/nf-core/hisat2/build_test/main.nf
```diff
@@ -7,11 +7,32 @@ include { HISAT2_BUILD              } from '../../../../../modules/nf-core/hisat
 
 workflow test_hisat2_build {
     fasta = [ [id:'genome'],
-              file(params.test_data['sarscov2']['genome']['genome_fasta'], checkIfExists: true)
+              file(params.test_data['homo_sapiens']['genome']['genome_fasta'], checkIfExists: true)
             ]
     gtf   = [ [id:'genome'],
-              file(params.test_data['sarscov2']['genome']['genome_gtf'], checkIfExists: true)
+              file(params.test_data['homo_sapiens']['genome']['genome_gtf'], checkIfExists: true)
```
maxulysse, line 13: why changing from sarscov2 to homo_sapiens?

maxulysse: ok, I read the comments, so makes sense

jfy133: all OK then?

tests/modules/nf-core/hisat2/build_test/main.nf
```diff
@@ -7,11 +7,32 @@ include { HISAT2_BUILD              } from '../../../../../modules/nf-core/hisat
...
 }
+
+workflow test_hisat2_build_fasta_only {
+    fasta = [ [id:'genome'],
+              file(params.test_data['homo_sapiens']['genome']['genome_fasta'], checkIfExists: true)
+            ]
+    gtf   = [ [id:'genome'],
+              file(params.test_data['homo_sapiens']['genome']['genome_gtf'], checkIfExists: true)
+            ]
+    HISAT2_BUILD ( fasta, [[],[]], [[],[]] )
```
JoseEspinosa, line 26: Maybe change it to make it clearer that is the meta map?
```suggestion
    HISAT2_BUILD ( fasta, [[:],[]], [[:],[]] )
```

tests/modules/nf-core/hisat2/build_test/main.nf
```diff
@@ -7,11 +7,32 @@ include { HISAT2_BUILD              } from '../../../../../modules/nf-core/hisat
...
+
+workflow test_hisat2_build_fasta_ss_only {
+    fasta = [ [id:'genome'],
+              file(params.test_data['homo_sapiens']['genome']['genome_fasta'], checkIfExists: true)
+            ]
+    gtf   = [ [id:'genome'],
+              file(params.test_data['homo_sapiens']['genome']['genome_gtf'], checkIfExists: true)
+            ]
+    HISAT2_EXTRACTSPLICESITES ( gtf )
+    HISAT2_BUILD ( fasta, [[],[]], HISAT2_EXTRACTSPLICESITES.out.txt )
```
JoseEspinosa, line 37: Same here?
```suggestion
    HISAT2_BUILD ( fasta, [[:],[]], HISAT2_EXTRACTSPLICESITES.out.txt )
```

JoseEspinosa (commented):
Maybe I am missing something but isn't the test for the `gtf` input missing?

jfy133:
> Maybe I am missing something but isn't the test for the `gtf` input missing?

That's the first test no?

JoseEspinosa:
> > Maybe I am missing something but isn't the test for the `gtf` input missing?
> 
> That's the first test no?

Yes, indeed. I think I need some holidays 😛

JoseEspinosa (approved):
LGTM! 🚀 
I just dropped a comment about the empty map declaration but this is more how I would do it but not really an issue 😄

jfy133:
@JackCurragh you're GTG once making Jose's changes (assuming it doesn't break the testS)

JoseEspinosa:
> @JackCurragh you're GTG once making Jose's changes (assuming it doesn't break the testS)

If you think it could break the test, then just go ahead

tests/modules/nf-core/hisat2/build_test/main.nf
JackCurragh: The sarscov2 annotation doesn't contain any exons to be extracted so the linting was complaining about having expected (intermediate) outputs that had empty md5sums. Indeed, this actually is a redundant change when I remove them from the `test.yml` as James advised. 

Should I revert back?

jfy133: No, it's fine to switch! Doesn't amke a difference

JackCurragh: Oh haha now I have switched back.

PR #3666, opened 2023-07-24
New module: galah

prototaxites:
Adds Galah for genome dereplication.

Couple of unsureties:
* I've included the meta map, but I'm not sure if this is necessary or not - presumably this might be desireable in the case that the input genomes come from a single source (e.g. a metagenome), but not if they are from multiple sources. 

## PR checklist

- [x] Add a resource `label`


modules/nf-core/galah/main.nf
```diff
@@ -0,0 +1,57 @@
...
+    path "versions.yml"                               , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "${meta.id}"
+    def qc_input = qc_table ? ((qc_format == "checkm") ? "--checkm-tab-table ${qc_table}" : "--genome-info ${qc_table}") : ""
+    if( qc_table && !(qc_format in ["checkm", "genome_info"]) ) error "Invalid qc_format supplied! qc_format should be either 'checkm' or 'genome_info'."
```
adamrtalbot, line 27: Could this be spaced out or organised a bit better? I can't read it!

prototaxites: How does that look now? Split these multi-part if-elses across two extra lines.

modules/nf-core/galah/main.nf
```diff
@@ -0,0 +1,59 @@
...
+
+    conda "bioconda::galah=0.3.1"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/galah%3A0.3.1--h031d066_3':
+        'biocontainers/galah:0.3.1--h031d066_3' }"
+
+    input:
+    tuple val(meta), path(bins)
+    path(qc_table)
+    val(qc_format)
```
adamrtalbot, line 13: One final thing - I'm not familiar with the tool, does qc_table and qc_format relate to an individual sample? Or all samples? If they're per sample I would suggest this (easier for grouping and submitting):
```suggestion
    tuple val(meta), path(bins), path(qc_table), val(qc_format)
```

If it's one qc_table and format for all samples I would suggest this:
```suggestion
    tuple val(meta), path(bins)
    tuple val(meta2), path(qc_table)
    val(qc_format)
```

Alternatively, you could move `qc_format` to the same channel as the `qc_table`:
```suggestion
    tuple val(meta), path(bins)
    tuple val(meta2), path(qc_table), val(qc_format)
```

adamrtalbot (commented):
Nice one, just got a question about input structure (mostly because I'm not familiar with the tool).

modules/nf-core/galah/main.nf
prototaxites: The `qc_table` should contain information relating to all the files in `bins`, with a given input `qc_format`. If you were running the process more than once on multiple collections of bins, then each input would have its own `qc_table` - so I think the first option you suggest is appropriate!

prototaxites:
Fixed and good to go! 👍

adamrtalbot (approved):

PR #3674, opened 2023-07-27
Fix centrifuge: reduce number of input channels

jfy133:
The `sam_format` was unnecessary as it can be supplied via `args` (whereas for the FASTQ files we need to dynamically construct the output filenames to ensure they get picked up by the output globs).

Also running `--out-fmt 'sam'` has a weird bug in centrifuge itself so that the output profile/report from the tool is missing the header (versus with the head when no `--out-fmt` is supplied).

This PR removes the superfluous channel for `--out-fmt` and also captures the other 'alignment' format in the optional output glob.

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`

### DIFF
--- a/modules/nf-core/centrifuge/centrifuge/main.nf
+++ b/modules/nf-core/centrifuge/centrifuge/main.nf
@@ -16,7 +16,7 @@ process CENTRIFUGE_CENTRIFUGE {
     output:
     tuple val(meta), path('*report.txt')                 , emit: report
     tuple val(meta), path('*results.txt')                , emit: results
-    tuple val(meta), path('*.sam')                       , optional: true, emit: sam
+    tuple val(meta), path('*.{sam,tab}')                 , optional: true, emit: sam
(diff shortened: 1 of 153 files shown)


modules/nf-core/centrifuge/centrifuge/main.nf
```diff
@@ -49,9 +47,7 @@ process CENTRIFUGE_CENTRIFUGE {
         --report-file ${prefix}.report.txt \\
         -S ${prefix}.results.txt \\
         $unaligned \\
-        $aligned \\
-        $sam_output \\
-        $args
+        $aligned
```
LilyAnderssonLee, line 50: So we remove $args?

LilyAnderssonLee (approved):

modules/nf-core/centrifuge/centrifuge/main.nf
jfy133: Oops! Deleted too much

PR #3679, opened 2023-07-28
Hierarchy of tests applied in order 

adamrtalbot:
Unify Github Actions into single linear workflow.

This runs the tests in a single workflow:

1. Check for basic linting (Prettier, editorconfig)
1. Check for changes in the code base, match to tags in `pytest_modules.yml`
1. Perform `nf-core lint` on modules found in `pytest_modules.yml`
1. pytest-workflow on modules in `pytest_modules.yml`
1. Repeat steps (2 and 4 for `nftest_modules.yml`)
1. Confirms successful completion of all above steps with a final check

Visualisation of testing structure: 
image

Unfortunately nf-test needs a separate check for file changes. It still runs within the same workflow but requires an additional check. I couldn't fix this with clever matrix operations but it might be possible in future with a more unified testing structure.

**Note:** We can run the tests in linear, which might be useful for hackathons etc. We can do this by modifying the `needs:` block, e.g.:

Linear:
``` yaml
    needs: [pytest-changes, Prettier, EditorConfig, nf-core-lint]
```

Parallel:
``` yaml
    needs: [pytest-changes]
```

This would cause all tests to execute one after the other.


modules/nf-core/gatk4/haplotypecaller/main.nf
```diff
@@ -50,6 +50,8 @@ process GATK4_HAPLOTYPECALLER {
         --tmp-dir . \\
         $args
 
+    # Fake stuff
```
FriederikeHanssen, line 53: 😆 whats it for though?

adamrtalbot: It kicks off some actual pipeline test! It will be removed before merging.

FriederikeHanssen: ahhh, smart 🤯 💡

mirpedrol (approved):
LGTM

SPPearce:
> It's more clear if you look at the interactive version, unfortunately you can't edit it: https://github.com/nf-core/modules/actions/runs/5717939183?pr=3679

Oh wow, that is really bad design as it definitely looks like they are leading into completely different boxes.

SPPearce:
Is there a reason that `Prettier` and `EditorConfig` are UpperCamelCase, but everything else is kebab-case?

PR #3753, opened 2023-08-17
Module quast update to meet guidelines again

Joon-Klaps:
## PR checklist

Quast was not up to nf-core guideline standards for modules. Mostly, the output filename was hardcoded and additional arguments were redundant. 

- [X] Add a resource `label`

### DIFF
--- a/modules/nf-core/quast/main.nf
+++ b/modules/nf-core/quast/main.nf
@@ -13,12 +13,12 @@ process QUAST {
     tuple val(meta3), path(gff)
 
     output:
-    tuple val(meta), path("${prefix}")                   , emit: results
-    tuple val(meta), path("${prefix}.tsv")               , emit: tsv
-    tuple val(meta), path("${prefix}_transcriptome.tsv") , optional: true , emit: transcriptome
-    tuple val(meta), path("${prefix}_misassemblies.tsv") , optional: true , emit: misassemblies
-    tuple val(meta), path("${prefix}_unaligned.tsv")     , optional: true , emit: unaligned
-    path "versions.yml"                                  , emit: versions
+    tuple val(meta), path("${prefix}")                                                 , emit: results
+    tuple val(meta), path("${prefix}/report.tsv")                                      , emit: tsv
+    tuple val(meta), path("${prefix}/contigs_reports/all_alignments_transcriptome.tsv"), optional: true , emit: transcriptome
+    tuple val(meta), path("${prefix}/contigs_reports/misassemblies_report.tsv")        , optional: true , emit: misassemblies
+    tuple val(meta), path("${prefix}/contigs_reports/unaligned_report.tsv")            , optional: true , emit: unaligned
+    path "versions.yml"                                                                , emit: versions
 
     when:
     task.ext.when == null || task.ext.when
@@ -26,8 +26,8 @@ process QUAST {
     script:
     def args      = task.ext.args   ?: ''
     prefix        = task.ext.prefix ?: "${meta.id}"
-    def features  = gff ? "--features $gff" : ''
-    def reference = fasta ? "-r $fasta" : ''
+    def features  = gff             ?  "--features $gff" : ''
+    def reference = fasta           ?  "-r $fasta"       : ''
     """
     quast.py \\
         --output-dir $prefix \\
@@ -36,11 +36,6 @@ process QUAST {
         --threads $task.cpus \\
         $args \\
         ${consensus.join(' ')}
-
-    ln -s ${prefix}/report.tsv ${prefix}.tsv
-    [ -f  ${prefix}/contigs_reports/all_alignments_transcriptome.tsv ] && ln -s ${prefix}/contigs_reports/all_alignments_transcriptome.tsv ${prefix}_transcriptome.tsv
-    [ -f  ${prefix}/contigs_reports/misassemblies_report.tsv         ] && ln -s ${prefix}/contigs_reports/misassemblies_report.tsv ${prefix}_misassemblies.tsv
-    [ -f  ${prefix}/contigs_reports/unaligned_report.tsv             ] && ln -s ${prefix}/contigs_reports/unaligned_report.tsv ${prefix}_unaligned.tsv
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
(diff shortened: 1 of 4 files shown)


modules/nf-core/quast/main.nf
```diff
@@ -7,25 +8,26 @@ process QUAST {
...
 
     input:
-    path consensus
-    path fasta
-    path gff
-    val use_fasta
-    val use_gff
+    tuple val(meta)  , path(consensus)
+    tuple val(meta2) , path(fasta)
+    tuple val(meta3) , path(gff)
```
adamrtalbot, line 13:
```suggestion
    tuple val(meta) , path(consensus)
    tuple val(meta2), path(fasta)
    tuple val(meta3), path(gff)
```

modules/nf-core/quast/main.nf
```diff
@@ -7,25 +8,26 @@ process QUAST {
...
     output:
-    path "${prefix}"    , emit: results
-    path '*.tsv'        , emit: tsv
-    path "versions.yml" , emit: versions
+    tuple val(meta), path("${prefix}")                   , emit: results
+    tuple val(meta), path("${prefix}.tsv")               , emit: tsv
+    tuple val(meta), path("${prefix}_transcriptome.tsv") , optional: true , emit: transcriptome
+    tuple val(meta), path("${prefix}_misassemblies.tsv") , optional: true , emit: misassemblies
+    tuple val(meta), path("${prefix}_unaligned.tsv")     , optional: true , emit: unaligned
+    path "versions.yml"                                  , emit: versions
```
adamrtalbot, line 21:
```suggestion
    tuple val(meta), path("${prefix}")                  , emit: results
    tuple val(meta), path("${prefix}.tsv")              , emit: tsv
    tuple val(meta), path("${prefix}_transcriptome.tsv"), optional: true , emit: transcriptome
    tuple val(meta), path("${prefix}_misassemblies.tsv"), optional: true , emit: misassemblies
    tuple val(meta), path("${prefix}_unaligned.tsv")    , optional: true , emit: unaligned
    path "versions.yml"                                 , emit: versions
```

modules/nf-core/quast/main.nf
```diff
@@ -7,25 +8,26 @@ process QUAST {
...
 
     script:
-    def args = task.ext.args   ?: ''
-    prefix   = task.ext.prefix ?: 'quast'
-    def features  = use_gff ? "--features $gff" : ''
-    def reference = use_fasta ? "-r $fasta" : ''
+    def args      = task.ext.args   ?: ''
+    prefix        = task.ext.prefix ?: "${meta.id}"
+    def features  = gff ? "--features $gff" : ''
+    def reference = fasta ? "-r $fasta" : ''
```
adamrtalbot, line 30: This one is a bit unnecessary:
```suggestion
    def args      = task.ext.args   ?: ''
    prefix        = task.ext.prefix ?: "${meta.id}"
    def features  = gff             ?  "--features $gff" : ''
    def reference = fasta           ?  "-r $fasta"       : ''
```

modules/nf-core/quast/main.nf
```diff
@@ -35,7 +37,96 @@ process QUAST {
         $args \\
         ${consensus.join(' ')}
 
-    ln -s ${prefix}/report.tsv
+    ln -s ${prefix}/report.tsv ${prefix}.tsv
+    [ -f  ${prefix}/contigs_reports/all_alignments_transcriptome.tsv ] && ln -s ${prefix}/contigs_reports/all_alignments_transcriptome.tsv ${prefix}_transcriptome.tsv
+    [ -f  ${prefix}/contigs_reports/misassemblies_report.tsv         ] && ln -s ${prefix}/contigs_reports/misassemblies_report.tsv ${prefix}_misassemblies.tsv
+    [ -f  ${prefix}/contigs_reports/unaligned_report.tsv             ] && ln -s ${prefix}/contigs_reports/unaligned_report.tsv ${prefix}_unaligned.tsv
```
adamrtalbot, line 43: What does this achieve? Symlinking can break on some filesystems, why not just specify these paths in the output directive? e.g.:
```
outputs:
    tuple val(meta), path("${prefix}/contigs_reports/all_alignments_transcriptome.tsv"), optional: true, emit: thing
```

adamrtalbot (commented):
Just one query about the symlinking stuff.

modules/nf-core/quast/main.nf
Joon-Klaps: I did it to make the files most likely to be used in publihsdir more easily accesible. 
I wasn't aware that linking of files can break stuff more easily, is there an issue I can lecture myself on? 

It was also done in the original module of quast and just continued the logic.

adamrtalbot: Cool, it made sense but my feeling was it added another thing to go wrong. How is the new solution?

Joon-Klaps: All links were removed, if one would like to export the files in the publishdir they would have to either publish the entire result directory or specify it as `{prefix}/contigs_reports/all_alignments_transcriptome.tsv`, ...

adamrtalbot: Ah so you want to publish them in the base directory? That makes sense, having subfolders you don't control can be messy.

You could set `--output-dir .` and/or modify the publishDir to achieve that.

Joon-Klaps: I'm not entirely certain what you are suggesting.  The problem with setting `--output dir .` would mean that all contigs will have the same output name `report.tsv` and there won't be an easy way to distinguish the files from each other in the published directory no? Maybe when combining it with `saveAs` and `ext.prefix` but then you would basically do the same thing as what is written now.   

The problem I want to solve is how can we fix the multiple subfolders of quast's output without making mess by having 3 different subfolders with just one file and without the use of symlinks?

adamrtalbot: OK I see, what an annoying tool output.

Maybe we leave it as you started now (symlinking). I'll give it a stress test on some different file systems and if I encounter a problem I'll let you know?

Joon-Klaps: That's a good idea! Thanks!

adamrtalbot (approved):

Issue #3756, opened 2023-08-17
new module: stitch


saulpierotti:
I am working on developping a module for STITCH that will be used in the phaseimpute nf-core pipeline

Upstream: https://github.com/rwdavies/STITCH

famosab:
Closed by  #3766

PR #3766, opened 2023-08-21
Stitch

saulpierotti:
## PR checklist

New module STITCH, closing issue #3756. This will be part of the phaseimpute pipeline.

- [x] Add a resource `label`


saulpierotti:
Not sure why conda returns different hashes - should I address it or Docker and Singularity passing is sufficient?

modules/nf-core/stitch/meta.yml
```diff
@@ -0,0 +1,146 @@
...
+      licence: "['GPL v3']"
+
+input:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing information about the set of positions to run the imputation over
+        e.g. `[ id:'test' ]`
+
+  - meta3:
```
LouisLeNezet, line 29: meta3 need to be deleted here.

modules/nf-core/stitch/meta.yml
```diff
@@ -0,0 +1,146 @@
...
+      type: directory
+      description: |
+        Folder of pre-generated input RData objects used when STITCH is called with the `--regenerateInput FALSE` flag. It is generated by running STITCH with the `--generateInputOnly TRUE` flag.
+      pattern: "input"
+
+  - rdata:
+      type: directory
+      description: |
+        Folder of pre-generated input RData objects used when STITCH is called with the `--regenerateInput FALSE` flag. It is generated by running STITCH with the `--generateInputOnly TRUE` flag.
+      pattern: "RData_in"
```
LouisLeNezet, line 51: Shouldn't the pattern be '.RData' only ?
It is staged afterwards as 'Rdata_in'

tests/modules/nf-core/stitch/main.nf
```diff
@@ -0,0 +1,52 @@
...
+    .collectFile( name: "cramlist.txt", newLine: true, sort: true )
+
+    reads = Channel.of( reads ).combine( cramlist )
+
+    STITCH (
+        stitch_input,
+        reads,
+        reference,
+        seed
+    )
```
LouisLeNezet, line 51: Could you also add a test with the RData synchronisation ?
You could do that by getting the result of the previous test and feed it to the second ?

LouisLeNezet: It would also be nice to test without the seed and check for only the presence of the files.

LouisLeNezet:
> Not sure why conda returns different hashes - should I address it or Docker and Singularity passing is sufficient?

Could you try it locally and check if the image / files are really that different ?

tests/modules/nf-core/stitch/main.nf
saulpierotti: Thanks for the comments!

I think better to run it with a different test - first generate the input and then imputing. The current test is a single step imputation. Would you test without the seed only or in addition to the current test?

LouisLeNezet: Both imputation in one step and imputation in two would be nice to be tested.
I would put the test without the seed as an addition.

tests/modules/nf-core/stitch/main.nf
```diff
@@ -0,0 +1,71 @@
...
+def posfile         = file(params.test_data['homo_sapiens']['genome']['genome_21_stitch_posfile'], checkIfExists: true)
+def input           = []
+def rdata           = []
+def chromosome_name = "chr21"
+def K               = 2
+def nGen            = 1
+def stitch_input    = [ [ id: "test_positions" ], posfile, input, rdata, chromosome_name, K, nGen ]
+
+// sequencing data
+def crams = [
```
atrigila, line 17: STITCH also accepts bams as input. Could you also test that?

modules/nf-core/stitch/meta.yml
```diff
@@ -0,0 +1,146 @@
...
+      type: file
+      description: List of sorted BAM/CRAM/SAM file
+      pattern: "*.{bam,cram,sam}"
+
+  - collected_crais:
+      type: file
+      description: List of BAM/CRAM/SAM index files
+      pattern: "*.{bai,crai,sai}"
+
+  - cramlist:
```
atrigila, line 81: The input could be crams or bams. Is there a particular reason why you use crams only?

modules/nf-core/stitch/main.nf
```diff
@@ -0,0 +1,83 @@
+process STITCH {
+    tag "$meta.id"
+    label 'process_high'
```
atrigila, line 3: Is there a particular reason why you use process_high?

atrigila (commented):
Thanks a lot for your contribution. I have left some comments in the review.

modules/nf-core/stitch/meta.yml
saulpierotti: Good point will implement this. Just for the module I need to use either cram or bam for the name of the input, even though it can accept both

modules/nf-core/stitch/main.nf
saulpierotti: In general it is very resource-heavy as a method, probably would be the heaviest step in most pipelines. Do you think would be better to use something different? The memory/time scale quadratically with number of haplotypes (K), and I think approximately linearly with number of SNPs and number of samples

saulpierotti: Also memory scales ~linearly with number of threads

LouisLeNezet: Regarding the process tag configuration.
Do you think it's feasible to run it with less than 12 CPU, 84 Gb in 10 hours ?

saulpierotti: Depends on the sample size and parameters, for what I personally do with it no, but with small sample sets I think yes

saulpierotti:
I implemented the suggested changes - still need to test with conda. Let me know if you have other suggestions :)

modules/nf-core/stitch/main.nf
LouisLeNezet: @nf-core/core Could we have a bit of help here ?
We would like to know what's the best to do when the process tag "medium" is maybe a bit limit to work.

saulpierotti:
Ok so with conda locally I get the same hashes - I guess it is up to some difference in some graphics software shipped with the stitch container and the container used for the conda test.

LouisLeNezet:
> Ok so with conda locally I get the same hashes as with docker - I guess it is up to some difference in some graphics software shipped with the stitch container and the container used for the conda test. Should we remove the hashes from the images, investigate it further, or accept that the conda test fails?

That's strange that it doesn't work the same locally and on github with the different containers...
Maybe we should wait for the @nf-core/core to tell us their point of view ?

saulpierotti:
Makes sense! I am running the test locally with `nf-core module test stitch` and then I select conda in the prompt

modules/nf-core/stitch/main.nf
saulpierotti: I think we can set it to medium then and the user can override as required, there can be use cases where medium is enough I guess

atrigila (approved):
Thank you for the changes and the additional tests. LGTM! :tada:

PR #3785, opened 2023-08-29
Fix BCL Convert Container Name

adalpino-illumina:
There is a typo in the container name for the BCL Convert Process module. It should be `nfcore/bclconvert:4.0.3` instead of `nf-core/bclconvert:4.0.3`. Notice the additional dash.

https://github.com/nf-core/modules/blob/master/modules/nf-core/bclconvert/Dockerfile#L2

As a result, the following error occurs ...

>  docker: Error response from daemon: pull access denied for nf-core/bclconvert, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.

This change fixes the name of the container.

### DIFF
--- a/modules/nf-core/bclconvert/main.nf
+++ b/modules/nf-core/bclconvert/main.nf
@@ -2,7 +2,7 @@ process BCLCONVERT {
     tag {"$meta.lane" ? "$meta.id"+"."+"$meta.lane" : "$meta.id" }
     label 'process_high'
 
-    container "nf-core/bclconvert:4.0.3"
+    container "nfcore/bclconvert:4.0.3"
 
     input:
     tuple val(meta), path(samplesheet), path(run_dir)


Emiller88 (commented):
Appreciate the catch! However, this is using the quay image(`quay.io/nf-core/bclconvert`) which has the dash

https://quay.io/repository/nf-core/bclconvert

adalpino-illumina:
Ok @Emiller88, So it doesn't work currently, what changed should we make or what am I doing wrong on my end?

adalpino-illumina:
Also

$ docker pull quay.io/nf-core/bclconvert
Using default tag: latest
Error response from daemon: manifest for quay.io/nf-core/bclconvert:latest not found: manifest unknown: manifest unknown

Emiller88:
Could you elaborate on how it doesn't work?

And try 

```bash
$ docker pull quay.io/nf-core/bclconvert:4.0.3
```
It doesn't seem like there's a latest tag for the quay repo.

adalpino-illumina:
Yeah I forgot the tag when I tried to pull the image. Thanks.

it doesn't work because the container definition in the module does not point to `quay.io/nf-core/bclconvert:4.0.3` which as you say is the correct image.

Should it not point to `quay.io/nf-core/bclconvert:4.0.3` instead of `nf-core/bclconvert:4.0.3`?

Emiller88:
> Should it not point to quay.io/nf-core/bclconvert:4.0.3 instead of nf-core/bclconvert:4.0.3?

No, sadly it's a whole thing 🫠
We moved all the modules to use to `docker.registry` config setting, so you could swap in something like 
https://github.com/nf-core/modules/blob/90038ef7750628aa43992c07101f04dbf5c4032a/tests/config/nextflow.config#L32-L34

Then you can point at the biocontainers on ECR so you don't hit rate limits and whatnot https://aws.amazon.com/blogs/hpc/biocontainers-are-now-available-in-amazon-ecr-public-gallery/

For context, they announced it last year, and then the `registry` config got added to Nextflow proper, and we didn't have to hack it out with something like: https://github.com/nf-core/modules/pull/2291

adalpino-illumina:
Oh I see thanks! I'll close this guy too then since it sounds like we just need to add `docker.registry = 'quay.io'` to our config file.

PR #3788, opened 2023-08-29
Fix Pangolin Docker container name

adalpino-illumina:
Similarly to https://github.com/nf-core/modules/pull/3785, the Pangolin Docker container name was incorrect. This change updates the container name to the correct one.


adalpino-illumina:
Adding `docker.registry = 'quay.io'` to the config file should fix this.

See https://github.com/nf-core/modules/pull/3785

PR #3795, opened 2023-08-31
ci: Switch to setup-nextflow

edmundmiller:
This might break everything, but we'll at-least fix https://github.com/nf-core/setup-nextflow/issues/19

dumpster-fire

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -97,24 +97,13 @@ jobs:
       - name: Install nf-core tools development version
         run: python -m pip install --upgrade --force-reinstall git+https://github.com/nf-core/tools.git@dev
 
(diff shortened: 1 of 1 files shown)


modules/nf-core/fastqc/main.nf
```diff
@@ -30,6 +30,7 @@ process FASTQC {
         [ -f "\${new_name}" ] || ln -s \$old_name \$new_name
     done
 
+
```
maxulysse, line 33: don't forget to revert it

maxulysse:
LGTM, ok to merge once you revert the fake triggering

maxulysse (approved):
Just revert the changes in fastqc

PR #3798, opened 2023-08-31
ci: Try setup-apptainer

Emiller88:
Gonna remove the space commit, but I think this worked.

@adamrtalbot Any thoughts on the Setup Singularity steps with the `$NXF_SINGULARITY_CACHEDIR` and what not?

Also, not sure if there are other apptainer/singularity things going on, the singularity setup action we were using just said to start using the apptainer setup action. So if anyone knows more about it, chime in!

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -272,11 +272,9 @@ jobs:
         with:
           version: "${{ matrix.NXF_VER }}"
 
-      - name: Set up Singularity
+      - name: Setup apptainer
         if: matrix.profile == 'singularity'
-        uses: eWaterCycle/setup-singularity@v7
-        with:
-          singularity-version: 3.7.1
+        uses: eWaterCycle/setup-apptainer@main
 
       - name: Set up Singularity
         if: matrix.profile == 'singularity'
@@ -456,11 +454,9 @@ jobs:
           wget -qO- https://code.askimed.com/install/nf-test | bash
           sudo mv nf-test /usr/local/bin/
 
-      - name: Set up Singularity
+      - name: Setup apptainer
         if: matrix.profile == 'singularity'
-        uses: eWaterCycle/setup-singularity@v7
-        with:
-          singularity-version: 3.7.1
+        uses: eWaterCycle/setup-apptainer@main
 
       - name: Set up Singularity
         if: matrix.profile == 'singularity'


pontus:
The eWaterCycle action seems ok.

(Packages have been available for some time from both apptainer and singularityce, so the alternative would mostly be to install from those. Either should be fine, though.)

pontus (approved):

pontus:
There's no reason for nf-core to mess with singularity/apptainer internal caching (`NXF_SINGULARITY_CACHEDIR` is regarded by nextflow and works as well for apptainer as for singularityce, it just helps nextflow know where it should store images).

pontus:
Unless one changes the profile to use `apptainer` *instead of* singularity, one shouldn't need to change anything more (and I think there are still enough issues with that to make it premature).

If docker images are pulled directly, that actually uses the apptainer cache (but we really should strive not to, and we don't seem to have bothered with that for singularity).

pontus:
There are two different caches involved - nextflow can use a cache it looks in and uses instead - without that it'd store images inside the work directory (`work/singularity`), so if it does different runs with different work directories that can quickly fill up the disk. This is fixed by `NXF_SINGULARITY_CACHEDIR` (simply installing apptainer doesn't matter here - for all nextflow knows, it's still singularity and `NXF_SINGULARITY_CACHEDIR` is considered by nextflow, not singularity).

There's also a cache within singularityce/apptainer for OCI-conversions. That would not be duplicated over different runs, but could possibly be considered pointing somewhere if we wanted to use github caching (`actions/cache`) for those. I don't think that's a big problem today though - and the proper fix is really getting readymade images that don't need user conversion.

But I'm convinced nothing more is needed here (and if the eWaterCycle team has abandoned there singularity action, we don't want to use that, so either switch to their apptainer action or pull it directly I guess).

Emiller88:
> But I'm convinced nothing more is needed here (and if the eWaterCycle team has abandoned there singularity action, we don't want to use that, so either switch to their apptainer action or pull it directly I guess).

@adamrtalbot Yeah there was no loop, I just saw they were moving on and wondered if it would work out of the box 😄 

Just so I'm clear, everything is still good in this PR, right? I'll add a comment linking the comment from @pontus above those, so they don't get deleted.

PR #3803, opened 2023-09-03
Update bamcmp

SPPearce:
Bamcmp module currently doesn't allow specification of the filenames for the two bam files it generates, nor does it check if there is a collision in these.


modules/nf-core/bamcmp/main.nf
```diff
@@ -8,26 +8,37 @@ process BAMCMP {
...
+    path "versions.yml"                    , emit: versions
 
     when:
     task.ext.when == null || task.ext.when
 
     script:
     def args = task.ext.args ?: ''
-    def prefix = task.ext.prefix ?: "${meta.id}"
+    prefix = task.ext.prefix ?: "${meta.id}_primary"
+    prefix2 = task.ext.prefix2 ?: "${meta.id}_contaminant"
```
nvnieuwk, line 24: Seems controversial, but I like this solution :partying_face:

modules/nf-core/bamcmp/main.nf
nvnieuwk, line 1: Can you add a stub section?

nvnieuwk (commented):
One comment :p

modules/nf-core/bamcmp/main.nf
SPPearce: stub added

nvnieuwk (approved):
LGTM

PR #3804, opened 2023-09-03
New module: quilt/quilt

atrigila:
## PR checklist

Closes [#3783 ](https://github.com/nf-core/modules/issues/3783). Part of phaseimpute pipeline.  

- [x] Add a resource `label`


modules/nf-core/quilt/quilt/main.nf
```diff
@@ -0,0 +1,62 @@
+process QUILT_QUILT {
+    tag "$meta.id"
+    label 'process_single'
+
+    conda "bioconda::r-quilt=1.0.4"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/r-quilt:1.0.4--r43h06b5641_2':
+        'biocontainers/r-quilt:1.0.4--r43h06b5641_2' }"
+
+    input:
```
LouisLeNezet, line 10: It could be nice to have all the parameter with the same case convention.

modules/nf-core/quilt/quilt/main.nf
```diff
@@ -0,0 +1,62 @@
...
+    tuple val(meta), path("*.vcf.gz.tbi"),          emit: tbi,   optional:true
+    tuple val(meta), path("RData", type: "dir"),    emit: rdata, optional:true
+    tuple val(meta), path("plots", type: "dir"),    emit: plots, optional:true
+    path "versions.yml",                            emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
```
LouisLeNezet, line 27: Could you align the "="

modules/nf-core/quilt/quilt/main.nf
```diff
@@ -0,0 +1,62 @@
...
+    tag "$meta.id"
+    label 'process_single'
+
+    conda "bioconda::r-quilt=1.0.4"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/r-quilt:1.0.4--r43h06b5641_2':
+        'biocontainers/r-quilt:1.0.4--r43h06b5641_2' }"
+
+    input:
+    tuple val(meta), path(bams), path(bais), path(bamlist), val(chr), val(regionStart), val(regionEnd), val(buffer), val(nGen), path(reference_haplotype_file), path(reference_legend_file), path(genetic_map_file)
```
LouisLeNezet, line 11: Is it necessary to have the bams and bamlist argument ?
Wouldn't it be possible to create the bamlist/cramlist from the bams argument ?

tests/modules/nf-core/quilt/quilt/test.yml
```diff
@@ -0,0 +1,133 @@
...
+
+- name: quilt quilt test_quilt_optional_outputs
+  command: nextflow run ./tests/modules/nf-core/quilt/quilt -entry test_quilt_optional_outputs -c ./tests/config/nextflow.config -c ./tests/modules/nf-core/quilt/quilt/nextflow.config
+  tags:
+    - quilt
+    - quilt/quilt
+  files:
+    - path: output/quilt/RData/QUILT_prepared_reference.chr20.2000001.2100000.RData
+      md5sum: 74e273dca98556c07e9134107456c4a8
+    - path: output/quilt/plots/haps.NA12878.chr20.2000001.2100000_igs.1.0.truth.png
```
LouisLeNezet, line 33: Does the md5sum change for all the png ?

LouisLeNezet (commented):
Quite the nice module.
I'm excited to test it on my data !

saulpierotti:
Nice module! I see that QUILT has a CLI interface in the same way as STITCH, with this script:

`https://github.com/rwdavies/QUILT/blob/master/QUILT.R`

Maybe it would be nice to update the bioconda package to include this script if it does not already and use the cli in the module.
What do you think?

modules/nf-core/quilt/quilt/main.nf
atrigila: yes! I will correct that

tests/modules/nf-core/quilt/quilt/test.yml
atrigila: unfortunately yes

modules/nf-core/quilt/quilt/main.nf
atrigila: The argument required by the module is the bamlist only, but it won't work if you don't have the bams in the same directory. The solution I thought of for a future implementation (in the pipeline or subworkflow) is to create the bamlist from the bams (which I think Saul did for the testing of STITCH). But you need some sort of channel manipulation, which is more appropriately done in the subworkflow context.

atrigila:
Thanks for the suggestion @saulpierotti! I like the idea. I used your recipe as reference but I was not able to pass the tests. Would you mind taking a look at my [attempt adding the CLI](https://github.com/bioconda/bioconda-recipes/pull/42826)?

modules/nf-core/quilt/quilt/main.nf
```diff
@@ -0,0 +1,62 @@
+process QUILT_QUILT {
+    tag "$meta.id"
+    label 'process_single'
```
SPPearce, line 3: Is this an appropriate label for the tool? It looks like it is multithreaded and I suspect requires more than 6Gb of RAM in a lot of circumstances.

modules/nf-core/quilt/quilt/main.nf
```diff
@@ -0,0 +1,62 @@
...
+
+    conda "bioconda::r-quilt=1.0.5"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/r-quilt:1.0.5--r43h06b5641_0':
+        'biocontainers/r-quilt:1.0.5--r43h06b5641_0' }"
+
+    input:
+    tuple val(meta), path(bams), path(bais), path(bamlist), val(chr), val(regions_start), val(regions_end), val(buffer), val(ngen), path(reference_haplotype_file), path(reference_legend_file), path(genetic_map_file)
```
SPPearce, line 11: That is a really big tuple...
Is the order specific to anything? Can the mandatory file paths come before the optional arguments?
What are the value channels doing?

modules/nf-core/quilt/quilt/meta.yml
```diff
@@ -0,0 +1,99 @@
...
+      type: file
+      description: (Mandatory) Reference haplotype file in IMPUTE format (file with no header and no rownames, one row per SNP, one column per reference haplotype, space separated, values must be 0 or 1)
+      pattern: "*.{hap.gz}"
+  - reference_legend_file:
+      type: file
+      description: (Mandatory) Reference haplotype legend file in IMPUTE format (file with one row per SNP, and a header including position for the physical position in 1 based coordinates, a0 for the reference allele, and a1 for the alternate allele).
+      pattern: "*.{legend.gz}"
+  - genetic_map_file:
+      type: file
+      description: (Mandatory) File with genetic map information, a file with 3 white-space delimited entries giving position (1-based), genetic rate map in cM/Mbp, and genetic map in cM. If no file included, rate is based on physical distance and expected rate (expRate).
```
SPPearce, line 69: Is this mandatory or not?

modules/nf-core/quilt/quilt/meta.yml
```diff
@@ -0,0 +1,99 @@
...
+tools:
+  - "quilt":
+      description: "Read aware low coverage whole genome sequence imputation from a reference panel"
+      homepage: "https://github.com/rwdavies/quilt"
+      documentation: "https://github.com/rwdavies/quilt"
+      tool_dev_url: "https://github.com/rwdavies/quilt"
+      doi: "10.1038/s41588-021-00877-0"
+      licence: "['GPL v3']"
+
+input:
```
SPPearce, line 18: Can you order these in the same order as the input to the tool, with the meta2 and meta3 as well please.

tests/modules/nf-core/quilt/quilt/main.nf
```diff
@@ -0,0 +1,93 @@
...
+
+nextflow.enable.dsl = 2
+
+include { QUILT_QUILT                   } from '../../../../../modules/nf-core/quilt/quilt/main.nf'
+include { QUILT_QUILT as QUILT_OPTIONAL } from '../../../../../modules/nf-core/quilt/quilt/main.nf'
+
+
+    // input sequencing data (bam)
+    bam = [
+        file('https://github.com/nf-core/test-datasets/raw/modules/data/delete_me/quilt/NA12878.haplotagged.1.0.bam', checkIfExists: true),
```
SPPearce, line 11: Are these files going to be moved out of delete_me?

tests/modules/nf-core/quilt/quilt/main.nf
```diff
@@ -0,0 +1,93 @@
...
+        file('https://github.com/nf-core/test-datasets/raw/modules/data/delete_me/quilt/ALL.chr20_GRCh38.genotypes.20170504.chr20.2000001.2100000.noNA12878.legend.gz', checkIfExists: true)
+    ]
+
+    genetic_map_file = [
+        file('https://github.com/nf-core/test-datasets/raw/modules/data/delete_me/quilt/CEU-chr20-final.b38.txt.gz', checkIfExists: true)
+    ]
+
+    // parameters
+
+    def chr = "chr20"
```
SPPearce, line 44: Why are these parameters all being set in value channels? They feel like they should just be being supplied in ext.args in general, or could be given as part of the meta.map.

modules/nf-core/quilt/quilt/main.nf
```diff
@@ -0,0 +1,62 @@
...
+
+    """
+
+
+    QUILT.R \\
+        --chr=$chr \\
+        $list_param \\
+        --posfile=$posfile \\
+        --phasefile=$phasefile \\
+        --reference_haplotype_file=$reference_haplotype_file \\
```
SPPearce, line 43: If any of these are optional, the command needs to have an if, e.g.

`def reference_haplotype_command = ${reference_haplotype_file} ? "--reference_haplotype_file = ${reference_haplotype_file}" : ""`

SPPearce (commented):
I don't think all these value channels are needed here. Can they not be controlled by the ext-args?

modules/nf-core/quilt/quilt/main.nf
atrigila: Yes, the tuple contains a substantial number of arguments, which we see is a common requirement for imputation tools. A similar structure can be found in tools such as [STITCH](https://github.com/nf-core/modules/blob/master/modules/nf-core/stitch/main.nf) and [GLIMPSE 1](https://github.com/nf-core/modules/blob/master/modules/nf-core/glimpse/phase/main.nf) and [2](https://github.com/nf-core/modules/blob/master/modules/nf-core/glimpse2/phase/main.nf).

The value channels within the tuple serve the purpose of defining essential variable information, including the chromosome and region to be processed. While nGen, seed, and buffer are currently included in the tuple, I appreciate your suggestion to potentially provide them as ext.args. I will work on implementing this improvement.

modules/nf-core/quilt/quilt/meta.yml
atrigila: Thanks for noticing this. It has been corrected.

modules/nf-core/quilt/quilt/meta.yml
atrigila: Yes, thank you for your suggestion.

tests/modules/nf-core/quilt/quilt/main.nf
atrigila: As of now, the files will continue to reside in the "delete_me" directory until a consensus is reached on the most suitable method for organizing files used to test a module. It's worth noting that upon the publication of the complete #phaseimpute pipeline, we anticipate having our dedicated repository to manage test data. At that point, we will adjust the file storage accordingly.

tests/modules/nf-core/quilt/quilt/main.nf
atrigila: The decision to set these parameters in value channels serves two key reasons. Firstly, it aligns with the design and functionality of other nf-core modules, such as [STITCH](https://github.com/nf-core/modules/blob/master/modules/nf-core/stitch/main.nf), which is a tool developed by the same author. We wanted to have consistency in design that would facilitate ease of use and familiarity for users. 

Secondly, these parameters cannot be supplied via ext.args because the chromosome and chunks are dynamically read from the channel output of another tool, specifically, the "glimpse_chunk" tool. This approach allows for the simultaneous launch of several parallel processes, each handling a distinct section of the genome. However, I agree with you that some parameters (buffer, nGen, seed) could be supplied via ext.args and I am working on this implementation.

modules/nf-core/quilt/quilt/main.nf
atrigila: Thanks for your suggestion.

modules/nf-core/quilt/quilt/main.nf
SPPearce: Ok, that's fine, just wanted to ask the question.

modules/nf-core/quilt/quilt/main.nf
SPPearce: Ok, so if they are based on a previous tool output then a value makes sense.

tests/modules/nf-core/quilt/quilt/main.nf
SPPearce: Ok, just wanted to highlight that delete_me isn't for long-term storage.

SPPearce (commented):
One more question. Is the seed argument required for ensuring that the tool is deterministic? If so, might be sensible to ensure that a default is set, [like this](https://github.com/nf-core/modules/blob/f9c512bea532dbe9c043c2925cc50b9f27f28e37/modules/nf-core/umitools/dedup/main.nf#L32).

tests/modules/nf-core/quilt/quilt/nextflow.config
```diff
@@ -0,0 +1,18 @@
+process {
+
+    publishDir = { "${params.outdir}/${task.process.tokenize(':')[-1].tokenize('_')[0].toLowerCase()}" }
+    cpus = 1 // More than 1 cpu may lead to different md5sum
```
SPPearce, line 4: That is interesting, just because things end up in a different order in the output?

SPPearce (approved):
Looks good to me now!
The comment about multiple cpus makes me slightly nervous, does that make a difference downstream?

modules/nf-core/quilt/quilt/main.nf
saulpierotti: Just wanted to add here that nGen is a mandatory parameter for the tool and per nf-core guidelines it should be supplied as a value channel and not in ext.args 😄

SPPearce: Hmm, ok, if they are not set to any kind of default then they should be value channels: https://nfcore.slack.com/archives/C043UU89KKQ/p1677581560661679
What else is required for the tool to run and has no default set at all?

atrigila: `nGen` default is NA
`buffer` default is NA
`seed` default is also NA (but we added a default)

saulpierotti: @atrigila  can quilt run without specifying those parameters? I know that STITCH requires nGen so I supposed that also QUILT does but probably you know better than me? :) I think buffer is probably optional?

atrigila: I've tested them and unfortunately they are not optional, the code breaks. So I added them again back as val channels.

atrigila: Thank you both for your reviews! @SPPearce @saulpierotti

PR #3839, opened 2023-09-15
add MAGUS

lrauschning:
Added a module with two submodules for [MAGUS](https://github.com/vlasmirnov/MAGUS).
MAGUS has a pending [pull request](https://github.com/bioconda/bioconda-recipes/pull/43039) to being added to bioconda.
The module has been tested with a local conda install from the recipe submitted to bioconda.

## PR checklist

- [x] Add a resource `label`
- [x] Use BioConda and BioContainers if possible to fulfil software requirements. (comp. [https://github.com/bioconda/bioconda-recipes/pull/43039])
  - [x] `PROFILE=conda pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware` (tested with a local conda recipe)


tests/config/pytest_modules.yml
```diff
@@ -2246,6 +2246,14 @@ mageck/test:
   - modules/nf-core/mageck/test/**
   - tests/modules/nf-core/mageck/test/**
 
+magus/align:
+  - modules/nf-core/magus/align/test/**
+  - tests/modules/nf-core/magus/align/test/**
+
+magus/guidetree:
+  - modules/nf-core/magus/guidetree/test/**
+  - tests/modules/nf-core/magus/guidetree/test/**
```
maxulysse, line 2255:
```suggestion
  - modules/nf-core/magus/align/**
  - tests/modules/nf-core/magus/align/**

magus/guidetree:
  - modules/nf-core/magus/guidetree/**
  - tests/modules/nf-core/magus/guidetree/**
```

modules/nf-core/magus/align/main.nf
```diff
@@ -0,0 +1,52 @@
+process MAGUS_ALIGN {
+    tag "$meta_fasta.id"
+    label 'process_medium'
+
+    conda "bioconda::magus-msa=0.1.2a"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/magus-msa:0.1.2a':
+        'biocontainers/magus-msa:0.1.2a' }"
```
maxulysse, line 8:
```suggestion
    conda "bioconda::magus-msa=0.1.2"
    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
        'https://depot.galaxyproject.org/singularity/magus-msa:0.1.2--pyhdfd78af_0 ':
        'biocontainers/magus-msa:0.1.2--pyhdfd78af_0 ' }"
```

modules/nf-core/magus/guidetree/main.nf
```diff
@@ -0,0 +1,48 @@
+process MAGUS_GUIDETREE {
+    tag "$meta.id"
+    label 'process_medium'
+
+    conda "bioconda::magus-msa=0.1.2a"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/magus-msa:0.1.2a':
+        'biocontainers/magus-msa:0.1.2a' }"
```
maxulysse, line 8:
```suggestion
    conda "bioconda::magus-msa=0.1.2"
    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
        'https://depot.galaxyproject.org/singularity/magus-msa:0.1.2--pyhdfd78af_0 ':
        'biocontainers/magus-msa:0.1.2--pyhdfd78af_0 ' }"
```

maxulysse (approved):

SPPearce (changes requested):
Within the maintainers team we've had a discussion recently, and we want to:
-  keep the names of the meta maps as meta, meta2, meta3 etc across all modules for consistency.
- leave editing the meta maps to the workflow logic, not within modules.

modules/nf-core/magus/guidetree/main.nf
```diff
@@ -0,0 +1,48 @@
...
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/magus-msa:0.1.2--pyhdfd78af_0 ':
+        'biocontainers/magus-msa:0.1.2--pyhdfd78af_0 ' }"
+
+    input:
+    tuple val(meta), path(fasta)
+
+    output:
+    tuple val(meta), path("*.tree"), emit: tree
+    path "versions.yml"           , emit: versions
```
SPPearce, line 15:
```suggestion
    path "versions.yml"            , emit: versions
```

modules/nf-core/magus/guidetree/main.nf
```diff
@@ -0,0 +1,48 @@
...
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    magus \\
+        -np $task.cpus \\
+        -i $fasta \\
+        -o ${prefix}.tree \\
+        --onlyguidetree TRUE \\
```
SPPearce, line 28: I don't know what the tool does, what other options could this be?

SPPearce (commented):
Some minor tweaks, looks generally good.

SPPearce:
It isn't obvious to me why all the tests are failing though.

lrauschning:
I'm also not sure why they are failing – locally they work. I think its related to vlasmirnov/MAGUS#12, but I'm not sure what to do about that.

modules/nf-core/magus/align/meta.yml
```diff
@@ -0,0 +1,61 @@
...
+
+  - fasta:
+      type: file
+      description: Input sequences in FASTA format.
+      pattern: "*.{fa,fna,fasta}"
+
+  - tree:
+      type: file
+      description: Optional path to a file containing a guide tree in newick format to use as input.
+        If empty, or overwritten by passing `-t [fasttree|fasttree-noml|clustal|parttree]`, MAGUS will construct its own guide tree. If empty, `fasttree` is used as a default.
```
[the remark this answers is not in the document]
lrauschning: Yes, it works, at least when running the tests locally, as tested by the test using the guidetree module to generate a guidetree file.
When `-t` is passed multiple times, the last occurence overwrites the value, so a `-t` in `args` would overwrite the module input guidetree – I think that's reasonable behaviour, although I wouldn't expect that to be a very common scenario.

modules/nf-core/magus/guidetree/main.nf
lrauschning: The parameters mostly have to do with tweaking the way the multiple seuqence alignment is constructed.
They are documented in the help message (s. also https://github.com/vlasmirnov/MAGUS/blob/master/magus.py#L43 )

lrauschning:
Locally, both the align and guidetree tests work. Will run again when the CI is not being rate-limited :D

modules/nf-core/magus/guidetree/main.nf
SPPearce: But you are hardcoding this one to`--onlyguidetree TRUE`, could someone actually want the other outputs?

lrauschning: The other output would be the alignment, which can be obtained by the sister module magus/align.
Theoretically, one could also pass `--onlyguidetree FALSE` to args which would overwrite this setting, but I don't see much point in doing that rather than calling the correct module.

maxulysse:
@mashehu Can I get your help on that one?

pontus:
(5 second look) https://github.com/nf-core/modules/actions/runs/6546256778/job/17776379524?pr=3839#step:12:218 looks broken and likely needs to be worked around somehow.

lrauschning:
Heya,
if I recall correctly the result of our conversation at the hackathon was that the module works on a typical linux environment with glibc, but the busybox environment of the AWS runners on which the CI is running is causing the CI to fail.
is there a way to switch these CI tasks to GitHub runners?

pontus:
There might be problems with that as well (haven't checked), but as magus tries to change permissions for some files it provides at runtime, it won't work with singularity or docker (if trying running non-root and some other user than the one used for installation).

I'd suggest amending the bioconda recipe so it sets these files executable at installation time.

lrauschning:
Hm, I'm not sure that is related. It fails differently in the conda vs docker testcases, in the conda testcase the call to mafft is failing:
```
'magus guidetree test_magus_guidetree' exited with exit code '1' instead of '0'.
stderr: 
stdout: _test_magus_guidetree/work/conda/env-07f10ca1a16e2d07815b5569d6fde9df/lib/python3.12/subprocess.py", line 502, in check_returncode
      raise CalledProcessError(self.returncode, self.args, self.stdout,
  subprocess.CalledProcessError: Command '/home/ubuntu/pytest_workflow_znkd99vj/magus_guidetree_test_magus_guidetree/work/conda/env-07f10ca1a16e2d07815b5569d6fde9df/bin/mafft --localpair --maxiterate 1000 --ep 0.123 --quiet --thread 2 --anysymbol magus_working_dir/decomposition/initial_tree/skeleton_sequences.txt > magus_working_dir/decomposition/initial_tree/temp_initial_align.txt' returned non-zero exit status 1.
```
I think that is related to mafft only working in a glibc-based environment. I could maybe try adding glibc to the conda dependencies of magus, but I don't think that's a very neat solution, as nearly all boxes this will be running on will be using glibc anyway.
In the docker testcase the error is:
 ```
 'magus guidetree test_magus_guidetree' exited with exit code '1' instead of '0'.
stderr: 
stdout: 47651a2bd0146e6f1e69a992ee09c6ae7e8b5847bed5a
  Status: Downloaded newer image for quay.io/biocontainers/magus-msa:0.1.2--pyhdfd78af_0
  WARNING: Your kernel does not support CPU shares or the cgroup is not mounted. Shares discarded.
  chown: unrecognized option '--from'
  BusyBox v1.32.1 (2021-04-13 11:15:36 UTC) multi-call binary.
 ```
I think in this case docker itself is trying to `chown` a newly downloaded image and failing.
In addition, when testing with a fresh conda install of magus, I am not seeing any call to chown being made by MAGUS:
```
alias chown="echo CHOWN"
$ strace -o tmp.txt magus -i setoxin.fa -o setoxin-magus.fa | grep CHOWN
[nothing found]
$ grep -e "chown" tmp.txt
[nothing found]
```

PR #3840, opened 2023-09-15
Integrate FASTA and guidetree meta in famsa module

lrauschning:
Changes the metadata handling in the famsa module to no longer discard the guidetree meta information but to merge it with the metadata  from the input fasta, preferring the fasta information in case of a conflict, the same as it is handled in the MAGUS PR (#3839).

## PR checklist


lrauschning:
Eh whoops, will resubmit based on the current main master branch.

PR #3841, opened 2023-09-15
Merge metas of guide tree and input fasta in famsa module (resubmitted) 

lrauschning:
Changes the metadata handling in the famsa module to no longer discard the guidetree meta information but to merge it with the metadata from the input fasta, preferring the fasta information in case of a conflict, the same as it is handled in the MAGUS PR (#3839).

## PR checklist
 * [x]  This comment contains a description of changes (with reason).

### DIFF
--- a/modules/nf-core/famsa/align/main.nf
+++ b/modules/nf-core/famsa/align/main.nf
@@ -1,7 +1,7 @@
 
 
 process FAMSA_ALIGN {
-    tag "$meta.id"
+    tag "$meta_fasta.id"
     label 'process_medium'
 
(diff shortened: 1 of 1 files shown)


SPPearce:
Within the maintainers team we've had a discussion recently, and we want to:
-  keep the names of the meta maps as meta, meta2, meta3 etc across all modules for consistency.
- leave editing the meta maps to the workflow logic, not within modules.

lrauschning:
Okay, thanks for the clarification – should I close this, then?

SPPearce:
Yes, I think this can be closed.

PR #3848, opened 2023-09-20
Add meta to bowtie build

SPPearce:
Add a meta channel to the genome fasta for bowtie (bowtie1 not bowtie2) to bring it in line with the other tools.


modules/nf-core/bowtie/build/main.nf
```diff
@@ -8,20 +8,21 @@ process BOWTIE_BUILD {
...
-    path "versions.yml" , emit: versions
+    tuple val(meta), path('bowtie') , emit: index
+    path "versions.yml"             , emit: versions
 
     when:
     task.ext.when == null || task.ext.when
 
     script:
     def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${fasta.baseName}"
```
maxulysse, line 22: why not use meta.id?

maxulysse (approved):

modules/nf-core/bowtie/build/main.nf
SPPearce: All I did was add the meta at the moment, didn't look at the rest of the tool.

modules/nf-core/ngscheckmate/patterngenerator/tests/main.nf.test
```diff
@@ -16,7 +16,10 @@ nextflow_process {
             script "../../../bowtie/build/main.nf"
             process {
                 """
-                input[0] = file(params.test_data['sarscov2']['genome']['genome_fasta'], checkIfExists: true)
+                input[0] = [
+                    [ id: 'sarscov2' ],
+                    file(params.test_data['sarscov2']['genome']['genome_fasta'], checkIfExists: true)
```
mahesh-panchal, line 21: Has there been a formal switch to the `params.modules_testdata_base_path + hard_path` syntax now?

modules/nf-core/bowtie/build/tests/main.nf.test
```diff
@@ -0,0 +1,56 @@
...
+    tag "bowtie/build"
+
+    test("sarscov2 - fasta") {
+
+        when {
+            process {
+                """
+                input[0] = [[id: 'sarscov2'],
+                             file(params.test_data['sarscov2']['genome']['genome_fasta'], checkIfExists: true)
+                           ]
```
mahesh-panchal, line 19: I'm never sure whether to enforce nicer formatting on these.
```suggestion
                input[0] = [
                    [id: 'sarscov2'],
                    file(params.test_data['sarscov2']['genome']['genome_fasta'], checkIfExists: true)
                ]
```

modules/nf-core/ngscheckmate/patterngenerator/main.nf
```diff
@@ -25,7 +25,9 @@ process NGSCHECKMATE_PATTERNGENERATOR {
     if ("$fasta" == "${prefix}.fasta") error "makesnvpattern.pl generates a fasta file with the same name as the input fasta, use \"task.ext.prefix\" to disambiguate!"
 
     """
-    makesnvpattern.pl ${bed} ${fasta} ${bowtie_index}/${fasta.getBaseName()} . ${prefix}
+    INDEX=`find -L ./ -name "*.3.ebwt" | sed 's/\\.3.ebwt\$//'`
```
mahesh-panchal, line 28: Can you remind me what the list of files here is? I recall another instance where the pattern match wasn't unique because it also matched the `rev` something `ebwt`.

Also could we update it to the modern bash syntax for command substitution?` `\$( ... )` rather than backticks?

mahesh-panchal (commented):
No major issues aside from the tests. The rest are just passing comments that you can decide on, but not conditional to the approval.

modules/nf-core/ngscheckmate/patterngenerator/main.nf
SPPearce: There are two rev files, 4 ebwt files, which is why it is looking for number 3. I took this from the bowtie align module.

modules/nf-core/bowtie/build/tests/main.nf.test
SPPearce: I'm never sure what is nicer formatting on them ;)

modules/nf-core/ngscheckmate/patterngenerator/main.nf
SPPearce, line 28:
```suggestion
    INDEX=\$(find -L ./ -name "*.3.ebwt" | sed 's/\\.3.ebwt\$//')
```

modules/nf-core/ngscheckmate/patterngenerator/tests/main.nf.test
SPPearce: Yes... although I'm also assuming at some point there is a bulk update.

modules/nf-core/bowtie/align/tests/main.nf.test.snap
```diff
@@ -0,0 +1,192 @@
+{
+    "sarscov2 - single_end": {
+        "content": [
+            [
+                "versions.yml:md5,96e36b0b99c80da0be8239d03db30ecc"
+            ],
+            [
+                "7bdcfc6f54ae6e8f4570395cc85db9a3"
+            ],
+            [
```
mashehu, line 10: maybe stupid nf-test question: why do the content entries of the stub tests get keys, but not the non-stub?

modules/nf-core/bowtie/align/tests/nextflow.config
mashehu, line 1: file can be deleted

modules/nf-core/bowtie/build/meta.yml
```diff
@@ -15,13 +15,23 @@ tools:
...
 output:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing nformation about the genome fasta
+        e.g. [ id:'test' ]
   - index:
       type: file
-      description: Bowtie genome index files
+      description: Folder containing bowtie genome index files
```
mashehu, line 34:
```suggestion
      description: bowtie genome index files
```
If I see it correctly, you get all the `ebwt` files in `index`, not just the `bowtie` folder containing them

mashehu (approved):

modules/nf-core/bowtie/align/tests/main.nf.test.snap
SPPearce: It means the test isn't just the md5 on the whole channel

mashehu: interesting thanks!

modules/nf-core/bowtie/build/meta.yml
SPPearce: They are all inside the `bowtie` folder, so the downstream process receives the whole folder. Not sure why the nf-test doesn't show that on the md5sums

PR #3860, opened 2023-09-22
Fix initialise linting

mashehu:
needed to update the json schema to allow empty components/modules


adamrtalbot (approved):
Nice, wasn't aware this was an issue so thank for the fix.

mashehu:
no idea, why the CI didn't catch this...

PR #3861, opened 2023-09-22
add linting for subworkflows to ci

mashehu:


adamrtalbot (approved):
Nice one, that would have taken me ages!

PR #3884, opened 2023-09-27
Propr logratio

suzannejin:
This module computes the logratio transformation of one data matrix.
In concrete, it computes the ratios between each data value and a reference.
This reference could be the sample geometric mean, or the value from another variable.
In the first case, it is called centred logratio transformation (CLR), and in the second additive logratio transformation (ALR).

It is a common data transformation technique used in compositional data analysis (CoDA).

## PR checklist

- [x] Add a resource `label`


modules/nf-core/propr/logratio/main.nf
```diff
@@ -0,0 +1,23 @@
+process PROPR_LOGRATIO {
+    tag "$meta.id"
+    label 'process_single'
```
JoseEspinosa, line 3: Won't it need to be better to set higher default resources for the process? e.g.:
```suggestion
    label 'process_medium'
```

modules/nf-core/propr/logratio/templates/logratio.R
```diff
@@ -0,0 +1,299 @@
...
+################################################
+
+
+# read matrix
+mat <- read_delim_flexible(
+    opt\$count,
+    header = TRUE,
+    row.names = opt\$gene_id_col,
+    check.names = FALSE
+    )
```
JoseEspinosa, line 211:
```suggestion
)
```

modules/nf-core/propr/logratio/templates/logratio.R
```diff
@@ -0,0 +1,299 @@
...
+    if (opt\$alpha == 'NA'){
+        logratio <- get_logratio(mat, opt\$ivar)
+    }else{
+        logratio <- get_boxcox(mat, opt\$ivar, opt\$alpha)
+    }
+
+    # set the reference column to NA
+    logratio[,opt\$ivar] = NA
+
+}else if (opt\$transform == 'clr'){
```
JoseEspinosa, line 239:
```suggestion
} else if (opt\$transform == 'clr'){
```

modules/nf-core/propr/logratio/templates/logratio.R
```diff
@@ -0,0 +1,299 @@
...
+
+    # get clr
+    if (opt\$alpha == 'NA'){
+        logratio <- get_logratio(mat, 'clr')
+    }else{
```
JoseEspinosa, line 244:
```suggestion
    } else {
```

modules/nf-core/propr/logratio/templates/logratio.R
```diff
@@ -0,0 +1,299 @@
...
+    gene_id_col = 'gene_id'
+)
+opt_types <- lapply(opt, class)
+
+# Apply parameter overrides
+args_opt <- parse_args('$task.ext.args')
+for ( ao in names(args_opt)){
+    if (! ao %in% names(opt)){
+        stop(paste("Invalid option:", ao))
+    }else{
```
JoseEspinosa, line 157:
```suggestion
    } else {
```

modules/nf-core/propr/logratio/templates/logratio.R
```diff
@@ -0,0 +1,299 @@
...
+        file,
+        sep = separator,
+        header = header,
+        row.names = row.names,
+        check.names = check.names
+    )
+
+    if ( (row.names == 'gene_id') & ('gene_name' %in% colnames(mat)) ){
+        mat <- mat[, -which(colnames(mat) == 'gene_name')]
+    }else if ( (row.names == 'gene_name') & ('gene_id' %in% colnames(mat)) ){
```
JoseEspinosa, line 59:
```suggestion
    } else if ( (row.names == 'gene_name') & ('gene_id' %in% colnames(mat)) ){
```

modules/nf-core/propr/logratio/templates/logratio.R
```diff
@@ -0,0 +1,299 @@
...
+#' @param ivar Reference variable given by user.
+#' If it is 'null', then set the last column as reference (default).
+#' Otherwise, it should refer to either gene name or gene index.
+#' If the gene name is given, find its index.
+#'
+#' @return The reference gene index
+set_reference <- function(ivar, mat){
+    if (ivar == 'NA'){
+        ivar <- ncol(mat)
+    }else{
```
JoseEspinosa, line 90:
```suggestion
    } else {
```

modules/nf-core/propr/logratio/templates/logratio.R
```diff
@@ -0,0 +1,299 @@
...
+
+    # get reference set
+    opt\$ivar <- set_reference(opt\$reference, mat)
+    gene_name <- colnames(mat)[opt\$ivar]
+    opt\$reference <- gene_name
+
+    # get alr
+    if (opt\$alpha == 'NA'){
+        logratio <- get_logratio(mat, opt\$ivar)
+    }else{
```
JoseEspinosa, line 232:
```suggestion
    } else {
```

tests/modules/nf-core/propr/logratio/main.nf
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env nextflow
+
+nextflow.enable.dsl = 2
+
+include { PROPR_LOGRATIO } from '../../../../../modules/nf-core/propr/logratio/main.nf'
+
+workflow test_propr_logratio {
```
JoseEspinosa, line 7: Please implement a test workflow for each of the possible parameterizations, e.g. for testing `alr` etc.

tests/modules/nf-core/propr/logratio/nextflow.config
```diff
@@ -0,0 +1,10 @@
+process {
+
+    publishDir = { "${params.outdir}/${task.process.tokenize(':')[-1].tokenize('_')[0].toLowerCase()}" }
+
+
```
JoseEspinosa, line 5:
```suggestion
```

JoseEspinosa (changes requested):
I leave for @pinin4fjords the more extensive review to accommodate the module in the differentialabundance pipeline. Mainly I suggested format issues and asked for more tests.

modules/nf-core/propr/logratio/main.nf
suzannejin: But I don't want to allocate too many cpus for it. Should I create a custom label or should I declare the resources in the test/nextflow.config?

modules/nf-core/propr/logratio/main.nf
```diff
@@ -0,0 +1,23 @@
...
+
+    conda "conda-forge::r-propr=4.2.6"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/r-propr:4.2.6':
+        'biocontainers/r-propr:4.2.6' }"
+
+    input:
+    tuple val(meta), path(count), val(transformation), val(reference), val(alpha)
```
pinin4fjords, line 11: You should separate the mandatory configuration options in the the vals from the files to which the `meta` pertains- see [this example](https://github.com/nf-core/modules/blob/master/modules/nf-core/salmon/quant/main.nf) and [the docs](https://nf-co.re/docs/contributing/modules).

modules/nf-core/propr/logratio/main.nf
```diff
@@ -0,0 +1,23 @@
...
+
+    output:
+    tuple val(meta), path("*.logratio.tsv"), emit: logratio
```
pinin4fjords, line 14: silly thing- but could you align the emits as you'll see in most other modules? Makes things easier to read.

pinin4fjords (changes requested):
Looking pretty tidy already- nice.

I think you should refine the interface based on conventions, and I agree a couple more tests would be nice.

modules/nf-core/propr/logratio/main.nf
JoseEspinosa: The idea will be that you should allocate a sensible number of resources for running the module in a real case scenario. For module tests the resources are already limited by [this](https://github.com/nf-core/modules/blob/36cfa4c2a7e538832c50f40cb16d159e84cd3305/tests/config/nextflow.config#L8-L12) as they are also limited in the `test.config` for pipelines tests e.g. https://github.com/nf-core/differentialabundance/blob/3a849f046990707cad2a0df751c371ac7f1165ab/conf/test.config#L21-L23

modules/nf-core/propr/logratio/main.nf
```diff
@@ -11,9 +11,9 @@ process PROPR_LOGRATIO {
     tuple val(meta), path(count)
 
     output:
-    tuple val(meta), path("*.logratio.tsv"), emit: logratio
-    path "*.R_sessionInfo.log", emit: session_info
-    path "versions.yml", emit: versions
+    tuple val(meta), path("*.logratio.tsv"),    emit: logratio
+    path "*.R_sessionInfo.log",                 emit: session_info
+    path "versions.yml",                        emit: versions
```
JoseEspinosa, line 16: Take a look to the comma convention [here](https://nf-co.re/docs/contributing/modules#comma-example)
```suggestion
    tuple val(meta), path("*.logratio.tsv"), emit: logratio
    path "*.R_sessionInfo.log"             , emit: session_info
    path "versions.yml"                    , emit: versions
```

modules/nf-core/propr/logratio/meta.yml
```diff
@@ -0,0 +1,70 @@
...
+        This can be used at the workflow level to pass optional parameters to the module.
+        [id: 'test', ...]
+  - count:
+      type: file
+      description: |
+        Count matrix, where rows = variables or genes, columns = samples or cells.
+        This matrix should not contain zeros. Otherwise, they will be first replaced by the minimun value.
+        You may want to handle the zeros with a different method beforehand.
+      pattern: "*.{csv,tsv}"
+  - transformation:
```
pinin4fjords, line 34: You should remove anything from this file that's not in the inputs.

modules/nf-core/propr/logratio/meta.yml
```diff
@@ -0,0 +1,70 @@
...
+  - alpha:
+      type: string
+      description: |
+        (Optional) Alpha parameter for Box-Cox transformation. Default = null, standard log transformation is applied.
+
+output:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information.
```
pinin4fjords, line 52: Make sure this is appropriate to the matrix input (the meta is probably not sample info in this case)

modules/nf-core/propr/logratio/meta.yml
```diff
@@ -0,0 +1,70 @@
+name: "propr_logratio"
+description: ALR/CLR transform the data matrix
```
pinin4fjords, line 2: Could you expand this a bit? The user needs to know what the module is for, and probably doesn't know what ALR or CLR means.

pinin4fjords (changes requested):
Almost there! Looks like your `val` inputs were optional, so you've moved them to `ext.args`, which is fine.

Mostly just fixing up the meta required now. Could you also make sure this PR actually has a description describing what the module is for?

modules/nf-core/propr/logratio/meta.yml
suzannejin: even though if the arguments like transformation, alpha, etc would be useful for the user to declare ext.args?
Do you suggest another way or place to put this information for the users?

JoseEspinosa (approved):
LGTM! :rocket: 
Feel free to merge after addressing @pinin4fjords suggestions

modules/nf-core/propr/logratio/meta.yml
pinin4fjords: Yep, current policy is that we only document things in channels, with those being the minimal things required to operate a module. So if a value is mandatory it goes in via channels (like the Salmon example I indicated earlier) and gets documented in the meta.yml. Everything else is considered optional, goes in via ext.args, and we defer to the tool's native documentation (to avoid maintaining duplicate info here).

pinin4fjords (approved):
Thanks for addressing all the feedback!

PR #3911, opened 2023-10-04
Different approach to handling sralite files when using prefetch.

robsyme:
## Overview

The SRA uses two file formats used to store the short read data: sra and sralite, and we want to support both formats.

When using `prefetch ID`, the behaviour of the resulting files differs depending on whether the format of `ID` stored at NCBI is sra or sralite. 

If the backend format is sra (using `ID=SRR6242565` as an example), the `prefetch $ID` command writes into a directory keyed by the ID:
```
SRR6242565
└── SRR6242565.sra
```

If the backend format is sralite (using `ID=SRR042042` as an example), the `prefetch $ID` command writes a single file to the current working directory:
```
SRR042042.sralite
```

Previously, we simply moved the sralite file into a new directory to match the sra format output, but this can break downstream tools. The most important is `fasterq-fetch` which requires the path to the file in the case of sralite, and the path to the parent directory in the case of sra.  If we simply emit as output the file for sralite and the parent directory for sra, the downstream `fasterq-fetch` doesn't need any additional logic to differentiate the format.

This variability requires two changes (included in this PR):
1) The path provided to the `vdb-validate` command inside the script block
2) The path required in the `output` block of the process

## PR checklist

Closes #XXX 

- [x] Add a resource `label`

### DIFF
--- a/modules/nf-core/sratools/fasterqdump/main.nf
+++ b/modules/nf-core/sratools/fasterqdump/main.nf
@@ -2,10 +2,10 @@ process SRATOOLS_FASTERQDUMP {
     tag "$meta.id"
     label 'process_medium'
 
-    conda "bioconda::sra-tools=2.11.0 conda-forge::pigz=2.6"
+    conda "bioconda::sra-tools=3.0.8 conda-forge::pigz=2.6"
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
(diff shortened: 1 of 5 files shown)


robsyme:
Note: I might as well update the sra-tools version at the same time. Currently waiting on the sra-tools+pigz mulled container to be accepted into Biocontainers, and when that is done I'll fix up the container links and append one last commit to this PR.

Please don't merge until that version bump is in.

maxulysse (approved):
LGTM

maxulysse (approved):

maxulysse (approved):

PR #3920, opened 2023-10-06
INITIALISE subworkflow corrected scoping and parameters are made more consistent

adamrtalbot:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`

### DIFF
--- a/subworkflows/nf-core/initialise/main.nf
+++ b/subworkflows/nf-core/initialise/main.nf
@@ -9,7 +9,7 @@ workflow INITIALISE {
     take:
         version         // bool
         help            // bool
-        validate_params // bool
+        validateParams  // bool
         logo            // bool
 
     main:
@@ -47,13 +47,13 @@ workflow INITIALISE {
         log.info citation
 
         // Print help message if needed
-        if (params.help) {
+        if (help) {
             def String command = "nextflow run ${workflow.manifest.name} --input samplesheet.csv -profile docker"
             log.info paramsHelp(command)
             System.exit(0)
         }
 
-        if ( params.validate_params != false ){
+        if ( validateParams != false ){
             validateParameters()
         }
 


subworkflows/nf-core/initialise/main.nf
```diff
@@ -9,7 +9,7 @@ workflow INITIALISE {
     take:
         version         // bool
         help            // bool
-        validate_params // bool
+        validateParams  // bool
```
maxulysse, line 12: why `validateParams` and not `validate_params`?

maxulysse: I'd rather we keep what we take to reflect the params on the other side as much as possible

adamrtalbot: Consistency, here is every other tool in the template:
```
    // Schema validation default options
    validationFailUnrecognisedParams = false
    validateParams                   = true
    validationLenientMode            = false
    validationSchemaIgnoreParams     = 'genomes'
    validationShowHiddenParams       = false
    validationSkipDuplicateCheck     = false
```

maxulysse: these are comming from plugins

maxulysse: This subworkflow is not a plugin, and for me, is you have
`INITIALISE(params.help, params.validate_params...)`
we should reflect these in the input we `take,` so `help` and `validate_params`

maxulysse: All params that come from plugins follow Nextlfow's logic, so they are `camelCase`, all params coming from nf-core are `snake_case`

adamrtalbot: That isn't true, nf-validation isn't very consistent right now:
`params.monochrome_logs`:
https://github.com/nextflow-io/nf-validation/blob/7ca22a47c6504f78fe900cc1caba8bdf927825c4/plugins/nf-validation/src/main/nextflow/validation/SchemaValidator.groovy#L223-L225

`params.parameters_schema`:
https://github.com/nextflow-io/nf-validation/blob/7ca22a47c6504f78fe900cc1caba8bdf927825c4/plugins/nf-validation/src/main/nextflow/validation/SchemaValidator.groovy#L158

I have a PR for supporting `monochromeLogs` and `parametersSchema` but it will be backwards compatible and keep the old format parameters: https://github.com/nextflow-io/nf-validation/pull/101

I don't think the user cares where the parameters are coming from, but they should be consistent within a purpose. Having a bunch of `validateThing` then suddenly having `validate_other_thing` isn't intuitive and the moment you explain that it comes from a different plugin the user gets bored and starts thinking about something else. I'm bored just thinking about it.

maxulysse (approved):

PR #3943, opened 2023-10-10
test self-hosted

mashehu:
This should run the CI tests on AWS

### DIFF
--- a/modules/nf-core/dshbio/exportsegments/main.nf
+++ b/modules/nf-core/dshbio/exportsegments/main.nf
@@ -2,10 +2,10 @@ process DSHBIO_EXPORTSEGMENTS {
     tag "${meta.id}"
     label 'process_medium'
 
-    conda "bioconda::dsh-bio=2.3"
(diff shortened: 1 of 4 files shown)


modules/nf-core/dshbio/exportsegments/main.nf
```diff
@@ -2,10 +2,10 @@ process DSHBIO_EXPORTSEGMENTS {
     tag "${meta.id}"
     label 'process_medium'
 
-    conda "bioconda::dsh-bio=2.3"
+    conda "bioconda::dsh-bio=2.1"
```
heuermh, line 5: I don't understand the downgrade here, is that a merge issue?

heuermh (changes requested):

modules/nf-core/dshbio/exportsegments/main.nf
mashehu: I just needed a PR to test our self-hosted runners. So, nothing to see here 🙂

mashehu:
Working now

PR #3955, opened 2023-10-11
allow tools with no DOI available

mashehu:
With this PR we make the DOI check less strict and allow the value "no DOI available" as an alternative.


mahesh-panchal (approved):
Thank you.

PR #3969, opened 2023-10-12
use dorny/paths-filter with git commands instead of GitHub API

mirpedrol:
From the [docs](https://github.com/dorny/paths-filter):

>     # Personal access token used to fetch a list of changed files
>     # from GitHub REST API.
>     # It's only used if action is triggered by a pull request event.
>     # GitHub token from workflow context is used as default value.
>     # If an empty string is provided, the action falls back to detect
>     # changes using git commands.


mashehu (approved):

Emiller88:
@adamrtalbot any idea why this is failing? 

What's the point of the check for the checks just for the debug print?

mirpedrol:
It is detecting too many changed files, no? Testing dumpsoftwareversions, fastqc, snakemake and samtools

mashehu:
> It is detecting too many changed files, no? Testing dumpsoftwareversions, fastqc, snakemake and samtools

Yes, but somehow only from nf-test, so something not perfectly clean with the setup there.

mirpedrol:
After some research, I found that there is a small problem when using git commands to retrieve changed files. There's a [PR](https://github.com/dorny/paths-filter/pull/133) opened with the fix already. Same user created a fork of paths-filter with the fix. 
This is now detecting only files changed in this PR, but it includes the modules coming from merging branch master to the current branch.
Do you think it's a good option to use this fork until the original GHA is fixed?

PR #4024, opened 2023-10-16
bismark: nf-test migration

phue:
`nf-test` migration of bismark modules

- [x] bismark/genomepreparation
- [x] bismark/align
- [x] bismark/coverage2cytosine
- [x] bismark/deduplicate
- [x] bismark/methylationextractor
- [x] bismark/summary
- [x] bismark/report


phue:
Test coverage is still rather low though

`bismark/align` allows for different aligners (default is `bowtie2`, but also `hisat2`)
We do implement `hisat2` as a choice in `nf-core/methylseq` but pass that option via `ext.args`.

Not sure how to deal with that in the context of `nf-test` (Would probably have to change the `bismark/align` to have an actual `param` for the aligner choice)

sateeshperi:
could you not remove the pytest files rather just delete the tags in `pytest_modules.yml` ?

we will remove the pytest files later once we have all the nf-tests in

Emiller88 (approved):
Looks good! We'll ignore conda for now.

PR #4039, opened 2023-10-16
added dumpsoftwareversions.groovy

ThiboVanderkam:

### DIFF
--- a/modules/nf-core/custom/dumpsoftwareversions/templates/dumpsoftwareversions.groovy
+++ b/modules/nf-core/custom/dumpsoftwareversions/templates/dumpsoftwareversions.groovy
@@ -1,0 +1,132 @@
+String dedent(String input) {
+    def lines = input.tokenize('\n')
+    def minIndent = Integer.MAX_VALUE
+
+    for (line in lines) {
+        if (line.trim()) {
+            def indent = line.indexOf(line.trim())
+            if (indent < minIndent) {
+                minIndent = indent
+            }
+        }
+    }
+
+    def dedentedLines = []
+    for (line in lines) {
+        dedentedLines << line.substring(minIndent)
+    }
+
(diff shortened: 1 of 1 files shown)


adamrtalbot (commented):
Could you change this so that it can be imported directly in 'native Nextflow'?

Changes:
 - Move this into an Nextflow script as a function
 - Add an nf-test

Thank you!

modules/nf-core/custom/dumpsoftwareversions/templates/dumpsoftwareversions.groovy
```diff
@@ -0,0 +1,132 @@
...
+    def versionsMqc = [
+        "id": "software_versions",
+        "section_name": "${workflow.manifest.name} Software Versions",
+        "section_href": "https://github.com/${workflow.manifest.name}",
+        "plot_type": "html",
+        "description": "are collected at run time from the software output.",
+        "data": _make_versions_html(versionByModule),
+    ]
+
+    @Grab('org.yaml:snakeyaml:1.29') // Add SnakeYAML library dependency
```
grst, line 114: Could this lead to problems when someone tries to run the pipeline on a host not connected to the internet? I believe snakeyaml is anyway shipped with nextflow, but requesting a specific version could lead to a download request nevertheless?

modules/nf-core/custom/dumpsoftwareversions/templates/dumpsoftwareversions.groovy
```diff
@@ -0,0 +1,132 @@
+String dedent(String input) {
```
grst, line 1: Doesn't groovys native `.stripIndent()` do exactly this?

See also: http://docs.groovy-lang.org/latest/html/documentation/index.html#_triple_single_quoted_string

PR #4042, opened 2023-10-16
Set snapshots as generated

Emiller88:


maxulysse (approved):

adamrtalbot (approved):
I didn't think you could do better after the previous PR, but this is the single greatest piece of code ever written.

PR #4079, opened 2023-10-17
Move conda environment to yml

### LINKED ISSUE #4078: Move conda to yaml env files
Since we're shipping "self contained modules"(TODO Tag the issue linking those), we could move the conda envs to use [conda environment files](https://www.nextflow.io/docs/latest/conda.html#use-conda-environment-files). 

The end goal would be [building from wave](https://github.com/nf-core/modules/issues/4077), and then #3797 picking up the package updates and making the PRs automagically for us.

Emiller88:
Closes #4078


maxulysse (approved):

Emiller88:
Okay I have the changes for all of the modules locally. I'm not pushing it up yet because I don't want another PR with 100 reviewers 😬

mirpedrol (approved):

Emiller88:
Merging sorry for all the requests 😬

ajtock:
Hi @Emiller88 

Would it be valid to make the conda directive path absolute rather than relative by adding `"${projectDir}/"` (e.g., `conda "${projectDir}/modules/nf-core/trimgalore/environment.yml"`)? I encountered a potential issue when running a pipeline locally, where the relative path can be resolved only if the launch directory is the project directory containing `modules/`. I'm not too familiar with nf-test yet, but is it possible that adding `"${projectDir}/"` to the path might also fix [**Process relative paths aren't resolved correctly** askimed/nf-test#146](https://github.com/askimed/nf-test/issues/146)?

Thanks!
Andy

nvnieuwk:
I think this is a better solution for the whole pipeline runs, but it doesn't fix nf-test since it creates a small pseudo workflow to test the process. That pipeline is located in the `.nf-test` folder so the `$projectDir` is there instead of the top folder. this will need a change in nf-test to work

grst:
what about "${moduleDir}/environment.yaml"?

nvnieuwk:
Wow this works! Thanks for the suggestion @grst !!

nvnieuwk:
Fix PR: https://github.com/nf-core/modules/pull/4252

PR #4080, opened 2023-10-17
Add CI to build mulled containers with Wave

### LINKED ISSUE #4077: Build "mulled" containers with wave
And push to quay.io. Since I don't think wave requires tower anymore. 

We can build them from conda envs ourselves, and then set up a GitHub action whenever the modules get updated to push the containers.

From a lunch conversation with @mahesh-panchal and @bentsherman @phue

edmundmiller:
Closes #4077


maxulysse (approved):

maxulysse:
Why don't we do that already?

Emiller88:
> Why don't we do that already?

Just hadn't gotten around to it 😆 

I'm also having trouble authenticating with the registry. I didn't think it would have to go through tower if it was public. @pditommaso could you confirm?

I think if not we can use Tower to build the images in CI using `nextflow inspect` if the `Dockerfile` or `environment.yml` changes and push the image for people.

tests/config/nextflow.config
```diff
@@ -28,6 +30,9 @@ if ("$PROFILE" == "singularity") {
     docker.userEmulation = false
     docker.fixOwnership = true
     docker.runOptions = '--platform=linux/amd64 -u $(id -u):$(id -g)'
+    wave.enabled = true
+    wave.strategy = ['dockerfile', 'conda', 'container']
+    wave.build.repository = 'docker.io/emiller88/modules'
```
adamrtalbot, line 35: It's very kind of you to fund nf-core.

MatthiasZepper:
How does that impact offline use of pipelines? A mulled container can easily be downloaded with `nf-core download`. How about those built with wave? Will this still work?

Emiller88:
> How does that impact offline use of pipelines? A mulled container can easily be downloaded with `nf-core download`. How about those built with Wave? Will this still work?

That was @ewels concern as well! I think `nextflow inspect` might be able to handle this in theory. 

But I think I misunderstood that the wave builds are happening locally, but I believe they're happening on a web server.

Ideally this would just be for building the containers, and then end users will be able to pull the container normally, unless they modify the environment.

MatthiasZepper:
I am not in Barcelona and probably somewhat out of the loop. Additionally, I have no experience using Wave, @mribeirodantas Bytesize talk is essentially all I know about it.  

In general, I have no objections against abandoning Bioconda's CI in favour of Wave. Coincidentally, I learned yesterday why Oxford Nanopore [has little confidence in Bioconda](https://github.com/nanoporetech/modkit/pull/65#issuecomment-1766850967) and that they are strongly favouring own distribution channels. Moving away from Bioconda [allows supporting AArch64 natively](https://labs.epi2me.io/inkling) and for a better provenance in dependencies. Taking into account the importance of ONT software in nf-core pipelines, I think that alone justifies allowing for other software and container sources in modules, including building them with Wave.

> I believe this is a supported feature of Wave Freeze and nextflow inspect -concretize, however I think if it can't run offline this feature is a no-go, so we should triple check.

The ability to run offline is indeed paramount for us at NGI. We would need to maintain our own forks of the pipelines with patched modules, if that option was dropped from the nf-core pipelines/modules. From which registry the container is downloaded/pulled is pretty much irrelevant for us (_a non-free/non-sponsored service might be a slight downer, though_), but might matter for scientists in other countries that are under international sanctions (e.g. Iran, Russia etc.)

> I believe this is a supported feature of Wave Freeze and nextflow inspect -concretize. Build time is happening at the Wave server, but this should be functionally equivalent to running nf-core download. 

Tools 2.10 is simply extracting the container definitions from the modules & configs with Regexes. Subsequently, it is downloading all that start with `http://` paths and pulling the remainder. I am not aware of any code that would support Wave builds as of now. 

But the decision of abandoning the regexes in favour of `nextflow inspect` for future releases is set in stone and I presume you are actually implementing that already @adamrtalbot & Marcel? In that case, `nf-core download` could probably also be enabled to resolve any Wave instructions and also kick off builds and obtain a dynamically created container URI if needed? 

My only concern would be that the pipeline, when executed, will not find the appropriate container without contacting the Wave server asking for its hash respectively the required layers. But I believe this concern is actionable by the Nextflow/Wave plugin developers?

tests/config/nextflow.config
```diff
@@ -14,6 +14,8 @@ process {
 if ("$PROFILE" == "singularity") {
     singularity.enabled = true
     singularity.autoMounts = true
+    wave.enabled = true
+    wave.strategy = ['dockerfile', 'conda', 'container']
```
pditommaso, line 18:
```suggestion
    wave.enabled = true
    wave.strategy = ['conda', 'container']
    wave.freeze = true
    wave.build.repository = 'docker.io/emiller88/modules'
```

Singularity native requires the *freeze* mode, which in turn requires also the target build repository

tests/config/nextflow.config
```diff
@@ -28,6 +30,9 @@ if ("$PROFILE" == "singularity") {
     docker.userEmulation = false
     docker.fixOwnership = true
     docker.runOptions = '--platform=linux/amd64 -u $(id -u):$(id -g)'
+    wave.enabled = true
+    wave.strategy = ['dockerfile', 'conda', 'container']
```
pditommaso, line 34:
```suggestion
    wave.strategy = 'conda'
```

Likely you want only rely on `conda` packages

tests/config/nextflow.config
pditommaso, line 35: This is not strictly necessary. if omitted it used Wave default registry (tho images will only be cached for one week)

modules/nf-core/bowtie/align/main.nf
```diff
@@ -2,10 +2,7 @@ process BOWTIE_ALIGN {
     tag "$meta.id"
     label 'process_high'
 
-    conda "bioconda::bowtie=1.3.0 bioconda::samtools=1.16.1"
-    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
-        'https://depot.galaxyproject.org/singularity/mulled-v2-ffbf83a6b0ab6ec567a336cf349b80637135bca3:c84c7c55c45af231883d9ff4fe706ac44c479c36-0' :
-        'biocontainers/mulled-v2-ffbf83a6b0ab6ec567a336cf349b80637135bca3:c84c7c55c45af231883d9ff4fe706ac44c479c36-0' }"
+    conda "modules/nf-core/bowtie/align/environment.yml"
```
pditommaso, line 5: Would not using `bioconda::bowtie=1.3.0 bioconda::samtools=1.16.1` make it more readable?

modules/nf-core/bowtie/align/environment.yml
```diff
@@ -0,0 +1,9 @@
+name: nf-core_bowtie-align
+channels:
+  - conda-forge
+  - bioconda
+  - defaults
+
+dependencies:
+  - bioconda::bowtie=1.3.0
+  - bioconda::samtools=1.16.1
```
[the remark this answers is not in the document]
pditommaso: This is a tricky point, to include package compatible with ARM64, we are maintaining a [`seqera` channel](https://github.com/seqeralabs/conda-recipes) forked from bioconda. Therefore it be useful to add `seqera` in the channels and remove the `bioconda::` prefix. 

However, once bioconda will support ARM64, the goal is to merge those recipes there.

pditommaso:
> I'm also having trouble authenticating with the registry. I didn't think it would have to go through tower if it was public. @pditommaso could you confirm?

Currently, without credentials you will be limited to 25 builds per day. We are discussing how to allow a community registry for nf-core pipelines.

modules/nf-core/bowtie/align/main.nf
Emiller88: Main thing is wanting to incorporate Renovate to auto update or packages or us. It supports conda envs out of the box, it doesn't support a line of conda packages.

tests/config/nextflow.config
Emiller88: We have a few modules with dockerfiles (bclconvert off the top of my head) and a few that use images only and don't have conda packages (parabricks I think is this way)

tests/config/nextflow.config
Emiller88: Ideally we'd want to not rebuild them unless the environment changes.

The other issue is it would make more sense to have one container repo per module. 

Currently it looks like wave creates a new tag with the env name rather than a image per env.

MatthiasZepper:
> Currently, without credentials you will be limited to 25 builds per day. We are discussing how to allow a community registry for nf-core pipelines.
[
Would a combination of Wave with Kraken be an option](https://github.com/uber/kraken)? There are enough academic institutions represented in the nf-core community that could sponsor a server/mirror or two, so we could host the images in a peer-to-peer network after they were build with Wave?

Emiller88:
Okay I think this works pretty well. We can also swap to just plain docker builds if Wave is too much headache.

Only problem is I can't tell why wave isn't working. I'm guessing the tower instance doesn't have access to the GH container registry, but I'd hope it'd fail gracefully.

pditommaso:
Yes, it looks like the creds provided do not have enough permissions 

```
error checking push permissions -- make sure you entered the correct tag name, and that you are authenticated correctly, and try again: checking push permission for "[docker.io/emiller88/modules:bowtie-1.3.0--6627003a47795810](http://docker.io/emiller88/modules:bowtie-1.3.0--6627003a47795810)": creating push check transport for [index.docker.io](http://index.docker.io/) failed: GET https://auth.docker.io/token?scope=repository%3Aemiller88%2Fmodules%3Apush%2Cpull&service=registry.docker.io: unexpected status code 401 Unauthorized: {"details":"access token has insufficient scopes"}
```

pditommaso:
> Would a combination of Wave with Kraken be an option? There are enough academic institutions represented in the nf-core community that could sponsor a server/mirror or two, so we could host the images in a peer-to-peer network after they were build with Wave? Unfortunately, this does seemingly not support Singularity images...

@MatthiasZepper I had missed this. Kraken looks indeed an interesting option. In relation to Singularity as long as it's Kraken is OCI compliant it should work

.github/workflows/wave.yml
```diff
@@ -0,0 +1,78 @@
...
+          sudo mv wave-${WAVE_VER}-linux-x86_64 /usr/local/bin/wave
+          chmod +x /usr/local/bin/wave
+
+      - name: Build container
+        if: matrix.profile == 'docker'
+        run: |
+          wave --conda-file "${{ matrix.files }}" \
+              --freeze \
+              --await \
+              --build-repo ghcr.io/nf-core/modules/bowtie \
```
maxulysse, line 63:
```suggestion
              --build-repo quay.io/nf-core/modules/bowtie \
```

.github/workflows/wave.yml
```diff
@@ -0,0 +1,78 @@
...
+              --tower-token ${{ secrets.TOWER_ACCESS_TOKEN }} \
+              --tower-workspace-id ${{ secrets.TOWER_WORKSPACE_ID }}
+
+      - name: Build Singularity
+        if: matrix.profile == 'singularity'
+        run: |
+          wave --conda-file "${{ matrix.files }}" \
+              --freeze \
+              --await \
+              --build-repo ghcr.io/nf-core/modules/bowtie \
```
maxulysse, line 73:
```suggestion
              --build-repo quay.io/nf-core/modules/bowtie \
```

Emiller88:
https://github.com/nextflow-io/nextflow/commit/f5362a7b067173a29d684663df22bb48fbbf5659

I think this might fix my singularity or docker hash issue? I thought it was already a thing for some reason.

Emiller88:
Ah we're waiting on `23.11.0-edge` now, I thought it might have been released.

.github/workflows/wave.yml
```diff
@@ -0,0 +1,78 @@
...
+          sudo mv wave-${WAVE_VER}-linux-x86_64 /usr/local/bin/wave
+          chmod +x /usr/local/bin/wave
+
+      - name: Build container
+        if: matrix.profile == 'docker'
+        run: |
+          wave --conda-file "${{ matrix.files }}" \
+              --freeze \
+              --await \
+              --build-repo quay.io/nf-core/modules/bowtie \
```
nvnieuwk, line 63: Is it normal that it points to `bowtie` here?

modules/nf-core/bowtie/align/main.nf
```diff
@@ -3,9 +3,7 @@ process BOWTIE_ALIGN {
     label 'process_high'
 
     conda "${moduleDir}/environment.yml"
-    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
-        'https://depot.galaxyproject.org/singularity/mulled-v2-ffbf83a6b0ab6ec567a336cf349b80637135bca3:c84c7c55c45af231883d9ff4fe706ac44c479c36-0' :
-        'biocontainers/mulled-v2-ffbf83a6b0ab6ec567a336cf349b80637135bca3:c84c7c55c45af231883d9ff4fe706ac44c479c36-0' }"
+    container "nf-core/modules/bowtie:bowtie_align--6c5b9c93546643d8"
```
adamrtalbot, line 6: If I'm correct, we have to hard code the Wave built container every time we update (as you have done here). Correct? So it's always going to be a 2-step process which is a fairly rubbish development experience, although not as bad as waiting for mulled containers to build.

drpatelh: Would using this instead be an option where we can create a container config directly without needing to hard-code in the module? The config could be added to sit alongside the module and be updated automatically (I think).

```
nextflow inspect -concretize -format config -profile docker,wave . > container.config
```

Where we just add a Wave profile to the modules nextflow.config here somewhere:
https://github.com/nf-core/modules/blob/f3da70f025bd7138ee6e88135c489e201ec56c9a/tests/config/nextflow.config#L14-L31

```
    wave {
        fusion.enabled = true
        wave.enabled = true
        wave.strategy = ['conda']
        wave.build.repository = 'quay.io/nf-core/modules/'
        wave.freeze = true
    }
```

Don't think you need to specify the name of the container with this approach and would mean we can also build and push containers from the CLI if required by using the same profile if we have access.

Thoughts?

.github/workflows/wave.yml
Emiller88: I just hardcoded it for the moment. I didn't want to play the CI game with that yet.

nvnieuwk: Okay! Fair enough :p

modules/nf-core/bowtie/align/main.nf
Emiller88: >  So it's always going to be a 2-step process which is a fairly rubbish development experience

Ideally we'll never bump these by hand. It'll look like:
1. Renovate will bump the conda packages
2. Wave will build the new container
3. Renovate will bump the container in the same PR.

That is _if_ we don't do the inspect method.

> Thoughts?
https://github.com/nf-core/modules/pull/4080#discussion_r1364142755

Yeah I initially tried that, we didn't know how it'd work with offline. I guess does `nextflow inspect` work without being able to ask wave for the hash?

Yeah I just tried it by turning off my wi-fi, it just sits there and waits. So even if the containers are downloaded ahead of time it still needs to check with wave(for now)

Emiller88:
Ah, this might be dead in the water unless we spin up a singularity registry. 

But it also doesn't look like [wave supports any singularity registry](https://docs.seqera.io/wave/faq/#which-container-registries-are-supported-by-wave) 

Could someone educate me on what was the issue with letting it pull the docker images? Space?

edmundmiller:
Nevermind after updating the wave version and Nextflow version everything worked!

edmundmiller:
Things we'd need to do to use these in pipelines:

1. Update nf-core tools linting
2. Add `singularity.ociAutoPull` and `apptainer.ociAutoPull`
3. Bump Nextflow versions to `23.12.0-edge`

Bonus points:

1. Use the biocontainers script to make containers public automatically
2. Maybe consolidate images to be one environment per tool (ie we don't really need 2 different bowtie2 environments, it's just one extra dependancy)
3. Point at a different container backend that has a better uptime.
4. Clean up #4306 and make version bumping happen automatically.

JoseEspinosa (approved):
Awesome 😍

.github/workflows/wave.yml
```diff
@@ -0,0 +1,87 @@
...
+            return '${{ matrix.files }}'.replace('modules/nf-core/', '').replace('/environment.yml', '').replace('/', '_');
+
+      - name: Build container
+        if: matrix.profile == 'docker'
+        run: |
+          wave --conda-file "${{ matrix.files }}" \
+              --freeze \
+              --await \
+              --build-repo quay.io/nf-core/modules/${{steps.registry-name.outputs.result}} \
+              --cache-repository quay.io/nf-core/wave-cache \
```
ewels, line 71: Given the `/singularity` below?

```suggestion
              --cache-repository quay.io/nf-core/wave-cache/docker \
```

ewels:
> 3\. Bump Nextflow versions to `23.12.0-edge`

I feel a bit nervous about this for production pipelines. Can do the prep work but might be nice to wait for the next stable release at least.

mahesh-panchal:
Awesome work 🚀

modules/nf-core/bowtie/align/main.nf
```diff
@@ -3,9 +3,7 @@ process BOWTIE_ALIGN {
     label 'process_high'
 
     conda "${moduleDir}/environment.yml"
-    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
-        'https://depot.galaxyproject.org/singularity/mulled-v2-ffbf83a6b0ab6ec567a336cf349b80637135bca3:c84c7c55c45af231883d9ff4fe706ac44c479c36-0' :
-        'biocontainers/mulled-v2-ffbf83a6b0ab6ec567a336cf349b80637135bca3:c84c7c55c45af231883d9ff4fe706ac44c479c36-0' }"
+    container "nf-core/modules/bowtie_align:bowtie_align--d797c0ccd2d330fd"
```
maxulysse, line 6: WHAT IS THIS MAGIC?

MatthiasZepper: Please mind that it is compulsory for all magicians to deposit samples of all their spells in the grimoire "_Magical Modules Mastery_" 🪄. 

[The tome can be found in the tool's repository,](https://github.com/nf-core/tools/tree/master/tests/data/mock_module_containers/modules) and the [corresponding wizardry ordeal is codified in over here](https://github.com/nf-core/tools/blob/master/tests/test_download.py#L185-L239).

edmundmiller:
> I feel a bit nervous about this for production pipelines. Can do the prep work but might be nice to wait for the next stable release at least.

Agreed, hopefully 24.04.0 will be out at the beginning of April rather than the end of April 😉

modules/nf-core/bowtie/align/main.nf
edmundmiller: > Please mind that it is compulsory for all magicians to deposit samples of all their spells in the grimoire "Magical Modules Mastery" 🪄.

Didn't even know that was a thing 🙃

MatthiasZepper: Few people do, and luckily the container definitions in modules don't change frequently either. 

Whether it's a significant alteration or a minor linting oversight like not using the correct quote characters, the impact might unfortunately not be apparent until much later. Unless tested right away with `nf-core download`, it will take weeks to months until the first pipeline is released that uses an updated module. Failures due to a change in June 2023, for example, [were just noticed recently](https://github.com/nf-core/tools/pull/2768). 

Once an issue is identified, the problematic code has already been released in a pipeline revision and can't be retracted for fixing a quirky module. Therefore, the only viable option is to adapt `nf-core download` accordingly, which is cumbersome and sometimes not even possible any more, [because two pipelines require mutually exclusive adaptions](https://github.com/nf-core/tools/issues/2392#issuecomment-1682393518). Ironically, it was exactly an unconventional Bowtie module, that broke the downloadability for particular [chipseq](https://github.com/nf-core/chipseq/issues/318) and [atacseq](https://github.com/nf-core/atacseq/issues/270) revisions beyond repair.

The ideal solution would be a CI test running right away in the modules repo. However, I don't know if that is feasible, because individual Python functions from the tools package would need to be applied selectively to a module.

pditommaso:
I think we can resume this effort as of nextflow 24.03.0-edge nextflow support wave community  registry which allows you to build images without providing a custom (password-protected) container registry. Just remove `wave.build.repository` in your setting

edmundmiller:
https://github.com/seqeralabs/wave-cli/commit/c798f283f5d5c0b94de3f2a44455f6c68e4d7462 Might be useful.

Currently swapped the environment names:

```diff
+ name: htslib_samtools
- name: samtools_view
```

Personally think this is a better naming structure so we don't have an environment that could be the same. `samtools_view` `samtools_sort`, etc.

edmundmiller:
| setting           | Image                                                                    |
|-------------------|--------------------------------------------------------------------------|
| imageSuffix       | community.wave.seqera.io/library:htslib_samtools--9bc8b2f12954060b       |
| tagPrefix         | community.wave.seqera.io/library/build:htslib_samtools--a81ec36ed9046273 |
| none              | community.wave.seqera.io/library/build:htslib_samtools--a81ec36ed9046273 |
|--|--|
| No env name       | community.wave.seqera.io/library/htslib_samtools:1.20--11a4e6daa46930ec  |
| env name*          | community.wave.seqera.io/library/htslib_samtools:9bc8b2f12954060b        |
| Seqera containers | community.wave.seqera.io/library/htslib_samtools:1.20--11a4e6daa46930ec  |

So there are two options here.

1. Change the Seqera containers default to remove the 1.20.
2. Add the option to wave-cli to generate it with a version in the tag even with env_name.
3. Get rid of the env names.

I think 2 is best option to avoid `work/conda/environment-764620c334b2a2c557b6a7deec235610` which can make them difficult to find. 

But I also see the arguement to drop the env_names, let Nextflow figure out if the environments are the same, and then we're future proofed against and Seqera containers default changes.

> [!IMPORTANT]
> Main reason for wanting to stay in line with the website is to avoid having duplicate containers for the same thing (ie `htslib_samtools:1.20--11a4e6daa46930ec` & `htslib_samtools:9bc8b2f12954060b`)

edmundmiller:
🙃

https://github.com/nf-core/modules/issues/4234

That cuts out option 3.

I think this sounds like a Nextflow bug because the two modules should be able to use the same environment.

The plot thickens:

| env name          | community.wave.seqera.io/library/bowtie_samtools:76ee1ea1b2f95f97 |   |
| Seqera containers | community.wave.seqera.io/library/bowtie_samtools:92cb3a5587de75d7 |   |

With bowtie_samtools, there's no version tag(`1.3.1--hash`) with Seqera containers(I suspect because htslib is baked into samtools in the `samtools_view` env?) but the hashes don't match?

edmundmiller:
Okay, discussed with @pditommaso in [Slack](https://nextflow.slack.com/archives/C0477AS31T5/p1718147660190939).

The issue with the inconsistent hash is the `name` field.

Determined the best way forward:
1. Remove all the env names to match Seqera containers
2. If the errors come back up, swap tests to use mamba instead of conda and hope that fixes the conda bug

`2` did happen, we tested it in https://github.com/nf-core/sarek/pull/1564/ and https://github.com/nf-core/modules/pull/5802

Figured the easiest path is actually to just stop testing conda and pytest-workflow together. I think it's a worthy sacrifice to move this forward, since we're phasing pytest-workflow out soon(™️).

.github/workflows/wave.yml
```diff
@@ -0,0 +1,83 @@
...
+          script: |
+            return '${{ matrix.files }}'.replace('modules/nf-core/', '').replace('/environment.yml', '').replace('/', '_');
+
+      - name: Build container
+        if: matrix.profile == 'docker'
+        run: |
+          wave --conda-file "${{ matrix.files }}" \
+              --freeze \
+              --await \
+              --cache-repository quay.io/nf-core/wave-cache \
```
ewels, line 68: Do we need `--cache-repository` now that Singularity Containers is live?

.github/workflows/wave.yml
```diff
@@ -0,0 +1,83 @@
...
+      # - name: Build Singularity
+      #   if: matrix.profile == 'singularity'
+      #   run: |
+      #     wave --conda-file "${{ matrix.files }}" \
+      #         --freeze \
+      #         --await \
+      #         --cache-repository quay.io/nf-core/wave-cache/singularity \
+      #         --tower-token ${{ secrets.TOWER_ACCESS_TOKEN }} \
+      #         --tower-workspace-id ${{ secrets.TOWER_WORKSPACE_ID }} \
+      #         --singularity
```
ewels, line 81: I would like to still build Singularity natively, rather than using `oras`. This will be quicker for the end user and also allow https downloads, so that `nf-core download` still works for offline use.

ewels:
Copying over dump [from Slack](https://nfcore.slack.com/archives/CJRH30T6V/p1718274585966869?thread_ts=1718230777.984439&cid=CJRH30T6V):

> Main thing is for Singularity: I would like to keep `https://` downloads for Singularity `.sif` files if possible, so that we don't break `nf-core download`
> That's currently in-development for Wave / Wave-CLI and should hopefully be ready any day
> Having native Singularity images is also better because it's quicker / less compute resources for the end user rather than recompiling at run time
>
> If we're doing this then we also need to figure out the best nextflow syntax for `container` in the processes, as we're back to having two separate strings
>
> And finally, I would also like to be building for ARM as well where possible

.github/workflows/wave.yml
adamrtalbot: Why use quay.io now? Why not switch to Seqera community containers?

.github/workflows/wave.yml
adamrtalbot: Agreed, no need to use quay.io now if we can avoid it.

edmundmiller: Done! Sorry forgot to cut that out after removing the build repo.

modules/nf-core/bowtie/align/main.nf
```diff
@@ -3,9 +3,10 @@ process BOWTIE_ALIGN {
     label 'process_high'
 
     conda "${moduleDir}/environment.yml"
+    container ""
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
-        'https://depot.galaxyproject.org/singularity/mulled-v2-ffbf83a6b0ab6ec567a336cf349b80637135bca3:c84c7c55c45af231883d9ff4fe706ac44c479c36-0' :
-        'biocontainers/mulled-v2-ffbf83a6b0ab6ec567a336cf349b80637135bca3:c84c7c55c45af231883d9ff4fe706ac44c479c36-0' }"
+        'oras://community.wave.seqera.io/library/bowtie_samtools:16f00b34cfc72399' :
```
ewels, line 8: I think this will break `nf-core download`. I'd prefer the https links if possible, and should test the download if we can.

modules/nf-core/bowtie/align/environment.yml
```diff
@@ -1,7 +1,8 @@
-name: bowtie_align
```
ewels, line 1: Are we ok to remove these? Given the conda concurrent env build bug?

modules/nf-core/bowtie/align/main.nf
```diff
@@ -3,9 +3,10 @@ process BOWTIE_ALIGN {
     label 'process_high'
 
     conda "${moduleDir}/environment.yml"
+    container ""
```
ewels, line 6: Accidental?

.github/workflows/wave.yml
```diff
@@ -0,0 +1,118 @@
...
+          sudo mv wave-${WAVE_VER}-linux-x86_64 /usr/local/bin/wave
+          chmod +x /usr/local/bin/wave
+
+      - name: Create a registry name
+        uses: actions/github-script@v7
+        id: registry-name
+        with:
+          result-encoding: string
+          script: |
+            return '${{ matrix.files }}'.replace('modules/nf-core/', '').replace('/environment.yml', '').replace('/', '_');
```
ewels, line 60: This is to have `name` in the conda environment, right? We can remove if so.

ewels (commented):
We should add in ARM builds as well (may need to tolerate + log failures).

Also, is it possible to parallelise the different requests in separate jobs? Would go 4x faster then.

ewels:
It'd be nice to also write the container URIs to the `meta.yml` file. No problem for it to be duplicated for now, means we can just remove from the `main.nf` file later. See also https://github.com/nf-core/modules/pull/6659#issuecomment-2357587262

Note that we haven't decided on a spec for how this should look in the meta YAML file yet, so need to do that. Should be easy to agree though, I have a rough idea in my head already.

.github/workflows/wave.yml
edmundmiller: No, this was to have a different wave repo per module. We can keep this for anything built by a Dockerfile for now(bclconvert, cellranger, etc.)

edmundmiller:
> We should add in ARM builds as well (may need to tolerate + log failures). 

Sounds good! We can log it in the CI definitely, or maybe we output it to files like bioconda does.

modules/nf-core/bowtie/align/main.nf
edmundmiller: Is there a Seqera feedback link for the https links?

edmundmiller:
Going to merge this finally. There isn't anything in here that should break anything.

Things that ended up in this PR:
1. Requests images on environment.yml changes
2. Switched bowtie modules and samtools/view to Seqera containers(which we've been doing)
3. Added Renovate comments

This is so we can start making some smaller PRs to the system and testing automations out.

mahesh-panchal:
I'm not sure if it's related to this PR in particular, but I'm getting notifications of wave workflow run failed.

https://github.com/nf-core/modules/actions/runs/11737926211

edmundmiller:
Yeah, thanks @mahesh-panchal fixed that in #6954 (hopefully)

PR #4088, opened 2023-10-17
Umitools/extract nf test

laramiellindsey:
nf-tests PR review checklist
- [ ] relative paths used in test file
- [ ] correct tags used (<tool>, <tool/sub-tool>, <modules> , <modules_nfcore>)
- [ ] all assertions enclosed in assertAll()
- [ ] /tests folder contains .nf.test , .nf.test.snap, tags.yml , nextflow.config-(optional)
- [ ] make sure the tags have been removed from tests/config/pytest_modules.yml so that pytests will be skipped

### DIFF
--- a/modules/nf-core/umitools/extract/tests/tags.yml
+++ b/modules/nf-core/umitools/extract/tests/tags.yml
@@ -1,0 +1,2 @@
+umitools/extract:
+  - modules/nf-core/umitools/extract/**
--- a/modules/nf-core/umitools/extract/tests/main.nf.test
+++ b/modules/nf-core/umitools/extract/tests/main.nf.test
@@ -5,6 +5,7 @@ nextflow_process {
     process "UMITOOLS_EXTRACT"
     config "./nextflow.config"
     tag "modules_nfcore"
+    tag "modules"
     tag "umitools"
     tag "umitools/extract"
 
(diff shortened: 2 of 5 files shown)


laramiellindsey:
Is this the best way to handle differences in snapshot, or do we need to add more assertions?

SPPearce:
> Is this the best way to handle differences in snapshot, or do we need to add more assertions?

What are there differences in the files?

laramiellindsey:
> > Is this the best way to handle differences in snapshot, or do we need to add more assertions?
> 
> What are there differences in the files?

The fastq md5 and log md5 do not match when rerunning nf-test using `assert snapshot(process.out).match()` . The versions md5 do match so `snapshot(process.out.versions).match("versions")` works.

SPPearce:
> > > Is this the best way to handle differences in snapshot, or do we need to add more assertions?
> > 
> > 
> > What are there differences in the files?
> 
> The fastq md5 and log md5 do not match when rerunning nf-test using `assert snapshot(process.out).match()` . The versions md5 do match so `snapshot(process.out.versions).match("versions")` works.

Yes, but what are the differences? The fastq should be the same I would have thought. The log might contain a date/time.

modules/nf-core/umitools/extract/tests/tags.yml
```diff
@@ -0,0 +1,2 @@
+umitools/extract:
+  - modules/nf-core/umitools/extract/**
```
maxulysse, line 2:
```suggestion
umitools/extract:
  - modules/nf-core/umitools/extract/**
```

maxulysse:
versions is failing with singularity, but ok for me to merge

PR #4118, opened 2023-10-17
Rename test.yml in getchromsizes

SPPearce:
The pytest tests are currently failing because `custom/getchromsizes` has a file called test.yml for nf-test.


sateeshperi (approved):
oops...sorry abt that...thanks for renaming 🙏

PR #4122, opened 2023-10-18
Newmodule yak

scorreard:
## PR checklist

- [x] Add a resource `label`

### DIFF
--- a/.github/workflows/pytest-workflow.yml
+++ b/.github/workflows/pytest-workflow.yml
@@ -122,7 +122,8 @@ jobs:
             tags: subworkflows/bcl_demultiplex
           - profile: "conda"
             tags: subworkflows/vcf_annotate_ensemblvep
-
+          - profile: "conda"
+            tags: subworkflows/fasta_clean_fcs
     env:
       NXF_ANSI_LOG: false
       SENTIEON_LICENSE_BASE64: ${{ secrets.SENTIEON_LICENSE_BASE64 }}
--- a/modules/nf-core/yak/main.nf
+++ b/modules/nf-core/yak/main.nf
@@ -1,0 +1,48 @@
+process YAK {
+    tag "$meta.id"
+    label 'process_medium'
+
+    conda "bioconda::yak=0.1"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/yak:0.1--h7132678_2':
+        'biocontainers/yak:0.1--h7132678_2' }"
+
+    input:
+    tuple val(meta), path(reads)
+
+    output:
+    tuple val(meta), path("*.yak"), emit: yak
+    path "versions.yml"           , emit: versions
+
+    when:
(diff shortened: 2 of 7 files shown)


.github/workflows/pytest-workflow.yml
```diff
@@ -122,7 +122,8 @@ jobs:
             tags: subworkflows/bcl_demultiplex
           - profile: "conda"
             tags: subworkflows/vcf_annotate_ensemblvep
-
+          - profile: "conda"
+            tags: subworkflows/fasta_clean_fcs
```
mahesh-panchal, line 126: I think this shouldn't be here.

modules/nf-core/yak/main.nf
```diff
@@ -0,0 +1,48 @@
...
+    yak \\
+        count \\
+        $args \\
+        -t${task.cpus} \\
+        -o ${prefix}.yak <(zcat $reads)
+
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        yak : \$(echo \$(yak --version 2>&1) | sed 's/^.*yak //' )
```
mahesh-panchal, line 33: You don't need this `echo`

scorreard:
```suggestion
        yak :  \$(yak --version 2>&1) | sed 's/^.*yak //' 
```

modules/nf-core/yak/main.nf
```diff
@@ -0,0 +1,48 @@
...
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    yak \\
+        count \\
+        $args \\
+        -t${task.cpus} \\
+        -o ${prefix}.yak <(zcat $reads)
```
scorreard, line 28:
```suggestion
    if [[ $reads == *.gz ]]; then
        yak \\
            count \\
            $args \\
            -t${task.cpus} \\
            -o ${prefix}.yak <(gzip -dc $reads)
     else
        yak \\
            count \\
            $args \\
            -t${task.cpus} \\
            -o ${prefix}.yak < $reads
     fi            
```

modules/nf-core/yak/main.nf
mahesh-panchal: The suggestion has the closing bracket in the wrong place.

modules/nf-core/yak/main.nf
mahesh-panchal: To keep the script code shorter I usually do something like:
```nextflow
script:
def useGzip = !( assembly instanceof List ? assembly.every{ it.name.endsWith(".gz") } : assembly.name.endsWith(".gz") )
"""
${useGzip ? 'gzip -c' : 'cat'} $assembly >output.gz
"""
```
The idea is a test is written in groovy, and then the value of it is used in an expression inside the script code to determine which command to use.

scorreard:
Closing this PR without merging as a lot of changes were made to module creation and nf-test, I think it would be better to strat from scratch.

PR #4142, opened 2023-10-18
modified proteus to output both raw and log2 intensities

suzannejin:
currently, proteus module only outputs log2 and normalized intensities.
Sometimes, we do want to raw intensities.

## PR checklist

- [x] Add a resource `label`


pinin4fjords:
I would say that, since this is a straightforward transform, it would be simpler to just output the raw values. An additional output consumes resources for no reason- users downstream can apply log2 if they wish.

suzannejin:
> I would say that, since this is a straightforward transform, it would be simpler to just output the raw values. An additional output consumes resources for no reason- users downstream can apply log2 if they wish.

hello Jonathan! 
I kept the log2 transform, because the original code needs the log2 transform for the downstream limma.
But yes, if you prefer, we can only keep the raw counts, and then inside limma log2 transform it. I actually rather prefer so too.

PR #4177, opened 2023-10-19
fix `no locator available for file` error

mashehu:
we added this trying to fix the AWS runners, but without success. now it seems to also break github runners so removing it again.


Emiller88 (approved):
🤠

PR #4181, opened 2023-10-20
update yaml schema with new maintainers field

mashehu:


Emiller88 (approved):
Didn't know this had a schema too...

PR #4183, opened 2023-10-20
Clean up after the merge cowboy

edmundmiller:
- Exclude a conda test
- Add note about version snaps


.github/workflows/test.yml
```diff
@@ -406,6 +406,7 @@ jobs:
             tags: universc
           - profile: conda
             tags: vt/decompose
+
```
SPPearce, line 409:
```suggestion
```

SPPearce (approved):

PR #4193, opened 2023-10-23
Change logic of mitohifi/mitohifi api

ksenia-krasheninnikova:
Input is specified directly by mitohifi -c/-r options via ext.args2

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`


muffato:
@ksenia-krasheninnikova . It now works ! I just added a Singularity option to mount `$HOME`.

You can ignore the `prettier` error as it comes from pre-existing files. By the way, I undid your merge from master because it was bringing more errors.

muffato:
@scorreard @mahesh-panchal . You may be interested in this MitoHiFi module change. Would you be able to review it ?

modules/nf-core/mitohifi/mitohifi/main.nf
```diff
@@ -41,11 +41,12 @@ process MITOHIFI_MITOHIFI {
...
-    def run_type = reads ? "-r ${reads}" :
-                    contigs ? "-c ${contigs}" :
-                    exit("Reads or contigs must be specified")
+    def args2 = task.ext.args2 ?: ''
+    def script_input = args2.equals("-r") ? "-r ${input}" :
+                args2.equals("-c") ? "-c ${input}" :
+                exit("-r for reads or -c for contigs must be specified")
     """
-    mitohifi.py ${run_type} -f ${ref_fa} -g ${ref_gb} -o ${mito_code} -t $task.cpus ${args}
+    mitohifi.py ${script_input} -f ${ref_fa} -g ${ref_gb} -o ${mito_code} -t $task.cpus ${args}
```
scorreard, line 49:
```suggestion
    mitohifi.py ${script_input} \\
        -f ${ref_fa} \\
        -g ${ref_gb} \\
        -o ${mito_code} \\
        -t $task.cpus ${args}
```

tests/modules/nf-core/mitohifi/mitohifi/nextflow.config
```diff
@@ -1,5 +1,7 @@
 process {
-
+    ext.args2 = '-c'
```
scorreard, line 2: Can you do a test with -r too or is that too complicated?

scorreard:
It is indeed a great improvement! I did a couple of suggestions, but other than that it looks good.
Can you be more specific about the reason prettier fail? What previous files?

muffato:
Prettier complains about some Conda environment.yml files that have been introduced recently at the hackathon. I noticed in this PR and another one that merging `master` in was in fact aggravating the situation by bringing a ton of failures in unrelated modules. That's why I'd rather wait for things to be solved on the master branch

tests/modules/nf-core/mitohifi/mitohifi/nextflow.config
ksenia-krasheninnikova: Yes, we can do it. Thank you for suggestion.

ksenia-krasheninnikova:
@muffato I've added another test for raw reads as @scorreard suggested. Github CI now evokes 182 check runs. I wonder if it happens because this branch is out of sync with the master. It also runs 'py-test changes' instead of 'pytest' and the run checks lack reports about the two mitohifi pytest runs. However both entry points are being picked up when I run `nf-core modules test mitohifi/mitohifi` locally.

scorreard (approved):
LGTM! Well done!!

muffato (approved):

ksenia-krasheninnikova:
Thank you both! @muffato @scorreard

mahesh-panchal:
Sorry, I was away.
However, doesn't this go against nf-core guidelines? The `-r` or `-c` is mandatory, and so must be an `input:` channel.
Also the `exit` is going to kill the workflow too no?

muffato:
Thanks @mahesh-panchal . I've opened #4248 to fix the `exit 1`.

I didn't think of getting the `-c`/`-r` through a channel. The guidelines say _SHOULD_, not _MUST_ 😉  :
> Non-file mandatory arguments or arguments needed to modify the command to make the module run with no error, SHOULD be provided as value channels (for example lib_type in [salmon/quant](https://github.com/nf-core/modules/blob/master/modules/nf-core/salmon/quant/main.nf)) - see ‘Input/output options’ below.
>
> Input channel `val` declarations SHOULD be defined for all mandatory non-file inputs that are essential for the functioning of the tool (e.g. parameters, flags etc).
> - Mandatory non-file inputs are options that the tool MUST have to be able to be run.
> - These non-file inputs are typically booleans or strings, and must be documented as such in the corresponding entry in the `meta.yaml`.

but I think using a separate channel would indeed be better here:

1. It'd make it absolutely clear that the option must be passed from the [documentation](https://nf-co.re/modules/mitohifi_mitohifi) and the input channel structure.
2. The `-c`/`-r` mode is very tightly coupled to the data that flow around the module that it makes sense to have it defined it right in the sub-workflow rather than in `modules.config`. It also mitigates the risk that someone overrides it by accident.

@ksenia-krasheninnikova . Could you do that in another PR ?

PR #4199, opened 2023-10-24
Genome gtf update

JackCurragh:
I re-did the changes to module tests using the new sorted genome.gtf. 

I am much more confident in these changes.


JackCurragh:
Apologies for the messy review requests. I didn't know they would automatically assign like that

PR #4200, opened 2023-10-24
Nf-test rtgtools/rocplot

nvnieuwk:


modules/nf-core/rtgtools/rocplot/tests/main.nf.test
```diff
@@ -0,0 +1,69 @@
...
+    tag "modules"
+    tag "modules_nfcore"
+    tag "rtgtools"
+    tag "rtgtools/rocplot"
+
+    test("homo_sapiens - tsv") {
+
+        setup {
+            run("UNTAR") {
+                script "../../../untar/main.nf"
```
Aratz, line 15:
```suggestion
                script "modules/nf-core/untar/main.nf"
```
Probably more robust and a bit clearer with an absolute path since these two modules are unrelated.

modules/nf-core/rtgtools/rocplot/tests/main.nf.test
```diff
@@ -0,0 +1,69 @@
...
+                process {
+                    """
+                    input[0] = [
+                        [ id:'test' ],
+                        file(params.test_data['homo_sapiens']['genome']['genome_21_sdf'])
```
Aratz, line 20: Is there a reason to skip `checkIfExists` here but not below?

modules/nf-core/rtgtools/rocplot/tests/main.nf.test
```diff
@@ -0,0 +1,69 @@
...
+            }
+        }
+
+        then {
+            assertAll (
+                { assert process.success },
+                { assert snapshot(
+                    file(process.out.png[0][1]).exists(),
+                    file(process.out.svg[0][1]).exists(),
+                    file(process.out.versions[0]).exists()
```
Aratz, line 62: Are these files md5sum-able? If yes, you should remove `exists()`, otherwise, you should probably use `assert file(process.out.png[0][1]).exists()`.

Aratz (commented):
Nice! Just had a few questions for you.

Also you can probably remove the old tests in `tests/modules/nf-core/rtgtools/rocplot`

modules/nf-core/rtgtools/rocplot/tests/main.nf.test
nvnieuwk: I followed the steps detailed here: https://hackmd.io/@nf-core/ryWsFLwyT I'm not sure what's the best option, pinging @sateeshperi for assistance :grin:

modules/nf-core/rtgtools/rocplot/tests/main.nf.test
nvnieuwk: Not really, I just copy pasted from the pytest files. But good idea to add it!

modules/nf-core/rtgtools/rocplot/tests/main.nf.test
nvnieuwk: Nope the md5sums differ everytime, that's why I added these checks :p

modules/nf-core/rtgtools/rocplot/tests/main.nf.test
nvnieuwk: It was intentional it seems. Somehow the check doesn't work even though the file exists... weird

modules/nf-core/rtgtools/rocplot/tests/main.nf.test
Aratz: For reference, in this module an absolute path was used: https://github.com/nf-core/modules/blob/master/modules/nf-core/sratools/fasterqdump/tests/main.nf.test

modules/nf-core/rtgtools/rocplot/tests/main.nf.test
nvnieuwk: I've added the file name as a snapshot value, I think this is better than to dynamically check if the output exists (without checking the filename). this way we know for sure the right files are in the right channel

modules/nf-core/rtgtools/rocplot/tests/main.nf.test
sateeshperi: we are going to stick with absolute paths for now to keep it consistent.....I will update the fasterqdump module

sateeshperi (approved):

modules/nf-core/rtgtools/rocplot/tests/main.nf.test
```diff
@@ -0,0 +1,69 @@
...
+                    input[0] = [
+                        [ id:'test' ],
+                        file(params.test_data['homo_sapiens']['genome']['genome_21_sdf'])
+                    ]
+                    """
+                }
+            }
+
+            run("RTGTOOLS_VCFEVAL") {
+                script "../../../rtgtools/vcfeval/main.nf"
```
nvnieuwk, line 27:
```suggestion
                script "modules/nf-core/rtgtools/vcfeval/main.nf"
```

modules/nf-core/rtgtools/rocplot/tests/main.nf.test
nvnieuwk: So absolute path for dependency modules and relative path for the module that actually needs testing?

PR #4248, opened 2023-10-30
Use "error" rather than "exit 1", and removed `quay.io/`

muffato:
1. @mahesh-panchal found an instance of `exit 1` in `mitohifi/mitohifi` after #4193 got merged. I found a few other instances across modules and decided to open a separate pull-request. I believe the rationale for using "error" instead of "exit 1" is expressed in #3683.
2. One of those modules was still using `quay.io/` as a container prefix, so I added another commit that removes the prefix from the few modules that still had it (see #3380).

## PR checklist

Closes #XXX 

- [ ] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [ ] Remove all TODO statements.
- [ ] Emit the `versions.yml` file.
- [ ] Follow the naming conventions.
- [ ] Follow the parameters requirements.
- [ ] Follow the input/output options guidelines.
- [ ] Add a resource `label`
- [ ] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - [ ] `PROFILE=docker pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [ ] `PROFILE=singularity pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [ ] `PROFILE=conda pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`

PR #4047, opened 2023-10-16
New module: NCBI datasets download

koenbossers:
## PR checklist

Add new module ncbidatasets/download

Contains tests with nf-test

Closes #3420 

- [x] This comment contains a description of changes (with reason).
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [x] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [x] Remove all TODO statements.
- [x] Emit the `versions.yml` file.
- [x] Follow the naming conventions.
- [x] Follow the parameters requirements.
- [x] Follow the input/output options guidelines.
- [x] Add a resource `label`
- [x] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - [ ] `PROFILE=docker pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [ ] `PROFILE=singularity pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [ ] `PROFILE=conda pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`

PR #4254, opened 2023-10-31
Mitohifi channel input

ksenia-krasheninnikova:
Changes to input according to 

https://github.com/nf-core/modules/pull/4193
https://github.com/nf-core/modules/pull/4248

by @muffato and @mahesh-panchal 

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`

### DIFF
--- a/modules/nf-core/mitohifi/mitohifi/main.nf
+++ b/modules/nf-core/mitohifi/mitohifi/main.nf
@@ -42,11 +42,11 @@ process MITOHIFI_MITOHIFI {
     }
 
     def args = task.ext.args ?: ''
-    if (! ["-c", "-r"].contains(input_mode)) {
-        error "-r for reads or -c for contigs must be specified"
+    if (! ["c", "r"].contains(input_mode)) {
+        error "r for reads or c for contigs must be specified"
     }
     """
-    mitohifi.py ${input_mode} ${input} \\
+    mitohifi.py -${input_mode} ${input} \\
         -f ${ref_fa} \\
         -g ${ref_gb} \\
         -o ${mito_code} \\
--- a/modules/nf-core/mitohifi/mitohifi/main.nf
+++ b/modules/nf-core/mitohifi/mitohifi/main.nf
@@ -10,6 +10,7 @@ process MITOHIFI_MITOHIFI {
     tuple val(meta), path(input)
     path ref_fa
     path ref_gb
+    val input_mode
     val mito_code
 
     output:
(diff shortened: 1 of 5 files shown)


modules/nf-core/mitohifi/mitohifi/meta.yml
```diff
@@ -25,6 +25,10 @@ input:
       type: file
       description: Reference annotation
       pattern: "*.{gb}"
+  - input_mode:
```
mahesh-panchal, line 28: I would change this to `use_reads` and have it be a boolean value. This means that `input_mode` doesn't need to be a string that has to be checked for it's value, and if it's true, use `-r`, otherwise `-c`. It would remove the need for the error message too.

ksenia-krasheninnikova: I'd rather keep it as is to opt for readability in the module calls. Also, boolean version suggests there should be a default value, while in this case setting the value explicitly is preferable. Also as @muffato points out boolean value is more off and on in its nature, while this case we're choosing between two different workflow types.

mahesh-panchal: Sorry, I'm not sure I understand. Can you show me how you're using it? 
In a module call, you can also use variables or comments to enhance readability. 
```nextflow
use_reads = true // could potentially be a params too. 
SOME_PROCESS (
    ch_input,
    use_reads // true: input are reads, false: input are contigs
)
```
In the current state, one needs to add a check in the module that `input_mode` should be either `-r` or `-c`.

I think this case is not different from https://github.com/nf-core/modules/blob/8fc1d24c710ebe1d5de0f2447ec9439fd3d9d66a/modules/nf-core/bwa/mem/main.nf#L13 for example.

modules/nf-core/mitohifi/mitohifi/main.nf
```diff
@@ -37,16 +38,17 @@ process MITOHIFI_MITOHIFI {
     script:
     // Exit if running this module with -profile conda / -profile mamba
     if (workflow.profile.tokenize(',').intersect(['conda', 'mamba']).size() >= 1) {
-        exit 1, "MitoHiFi module does not support Conda. Please use Docker / Singularity instead."
+        error "MitoHiFi module does not support Conda. Please use Docker / Singularity instead."
     }
 
     def args = task.ext.args ?: ''
     def args2 = task.ext.args2 ?: ''
```
mahesh-panchal, line 45: args2 should also be dropped as it's no longer being used.

modules/nf-core/mitohifi/mitohifi/meta.yml
muffato: I do see a difference with flags such as this `sort_bam`. I read booleans as extra steps that are typically off by default and explicitly turned on. In the case of MitoHifi, it's not an on/off thing, it's a alternative mode. I think a more accurate comparison is the NCBI datasets module (#4047) which does different things depending on the mode

mahesh-panchal: OK. I get the idea better. What about then making the modes just `r` or `c` and having the code as `-$mode`.

modules/nf-core/mitohifi/mitohifi/main.nf
```diff
@@ -37,16 +38,16 @@ process MITOHIFI_MITOHIFI {
...
+        error "MitoHiFi module does not support Conda. Please use Docker / Singularity instead."
     }
 
     def args = task.ext.args ?: ''
-    def args2 = task.ext.args2 ?: ''
-    def script_input = args2.equals("-r") ? "-r ${input}" :
-                args2.equals("-c") ? "-c ${input}" :
-                exit("-r for reads or -c for contigs must be specified")
+
+    if ( ! input_mode or ! (input_mode in [ '-r', '-c' ]) ) {
```
mahesh-panchal, line 46:
```suggestion
    if ( input_mode !in [ '-r', '-c' ] ) {
```

muffato:
Manual rebase + merge to resolve the conflict with `master`

modules/nf-core/mitohifi/mitohifi/main.nf
muffato: The merge from `master` brought an alternative way of doing this, via `contains`

mahesh-panchal (approved):
Thanks for updating.

PR #4267, opened 2023-11-01
Update genomescope

mahesh-panchal:
## PR checklist

Extracts the Kmer coverage from the model summary and emit's it into a channel. This is useful for certain tools that would like to use the kmer coverage estimate as input.

Closes #XXX 

- [ ] Add a resource `label`

### DIFF
--- a/modules/nf-core/genescopefk/main.nf
+++ b/modules/nf-core/genescopefk/main.nf
@@ -15,6 +15,7 @@ process GENESCOPEFK {
     tuple val(meta), path("*_summary.txt")                , emit: summary
     tuple val(meta), path("*_transformed_linear_plot.png"), emit: transformed_linear_plot
     tuple val(meta), path("*_transformed_log_plot.png")   , emit: transformed_log_plot
+    tuple val(meta), env(KMERCOV)                         , emit: kmer_cov
     path "versions.yml"                                   , emit: versions
 
     when:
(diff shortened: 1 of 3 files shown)


modules/nf-core/genescopefk/main.nf
```diff
@@ -36,6 +37,8 @@ process GENESCOPEFK {
         --output . \\
         --name_prefix ${prefix}
 
+    printf -v KMERCOV "%.2f\\n" \$( grep "^kmercov" *_model.txt | cut -d" " -f2 )
```
muffato, line 40: TIL I learned there is `printf -v` to create new environment variables. 👍🏼 

As far as I can tell, `printf -v` is a bash builtin. Are we certain this script section _always_ runs in bash ?

modules/nf-core/genescopefk/meta.yml
```diff
@@ -54,6 +54,10 @@ output:
       type: file
       description: GeneScope histogram summary
       pattern: "*_summary.txt"
+  - kmer_cov:
+      type: string
```
muffato, line 58: It's always a float, isn't it ?
```suggestion
      type: float
```

ksenia-krasheninnikova: Yes, it is always float

mahesh-panchal: The groovy class is a String, but that string is always a float as that's what the format string prints.

modules/nf-core/genescopefk/main.nf
mahesh-panchal: It works in ksh and a few other shells too, but I can add the directive at the top.

muffato (approved):

PR #4284, opened 2023-11-02
ci: One changes job to rule them all

edmundmiller:
Kinda just wondered if we could do this, makes it a little more linear to follow.

Need to fix the linting step.


edmundmiller:
Think this is going to get superseded by #6286, so closing it out.

PR #4285, opened 2023-11-02
Added a meta map for the reference genome of minimap2

muffato:
The minimap2/align module only supports a meta map for the reads to align. In this PR I add support for a meta map for the reference too - as in many other alignment modules.

As usual, the Conda test fails because the output BAM file differs "binarily", though its content (`samtools view`) is the same.

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`


modules/nf-core/minimap2/align/meta.yml
```diff
@@ -25,6 +25,11 @@ input:
       description: |
         List of input FASTA or FASTQ files of size 1 and 2 for single-end
         and paired-end data, respectively.
+  - meta2:
+      type: map
+      description: |
+        Groovy Map containing reference information
```
jfy133, line 31: I wonder if it's worth mentioning it's not used for anything

jfy133 (approved):

modules/nf-core/minimap2/align/meta.yml
muffato: Is it mentioned in other modules ? I don't remember seeing it.

That said, I think it could be misleading. Yes, there is no written occurrence of `meta2` in the module, but it doesn't prevent it from still being used, for instance in a closure for `task.ext.args`, `task.memory`, etc (which is the way I intend to use it).

jfy133: No not always. I seem to remember that the original modules from Harshil didn't include an entry for each one.

However you make a good point, so fine as it is. My approval stands as before :+1:

PR #4297, opened 2023-11-03
Add renovate config check

Emiller88:
- ci(renovate): Add pre-commit hook
- ci: Add a pre-commit GitHub action workflow

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -16,3 +16,8 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v3
       - uses: pre-commit/action@v3.0.0
+        # FIXME Flip this off once we get to less than a couple hundred. Adding
+        # this so it will only run against changed files. It'll make it much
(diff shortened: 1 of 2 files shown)


maxulysse (approved):

mirpedrol:
I saw in the action repo that they recommend using [pre-commit.ci](https://pre-commit.ci/), I'm not familiar with either of them, but should we consider this instead?

maxulysse:
pre-commit test are failing

Emiller88:
> I saw in the action repo that they recommend using [pre-commit.ci](https://pre-commit.ci/), I'm not familiar with either of them, but should we consider this instead?

Up to y'all, they recommend it because it's their product 😁 I'd rather keep it all unified and not find out whether their ci is free or not.

Emiller88:
> pre-commit test are failing

Yeah I'll see if I can get around to fixing them. Mainly our fault for not checking it and therefore not enforcing it 😬

Emiller88:
Forced the merge because bwa/mem fails with conda for now.

PR #4313, opened 2023-11-08
NONPAREIL:Add gzipped input support, and nf-test 🚀

jfy133:
Simply adds gunzipping of input FASTA/Q if gzipped, as the tool only accepts uncompressed, but it doesn't make sense to force a separate GUNZIP in the pipeline in all cases

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`


tests/modules/nf-core/pyrodigal/main.nf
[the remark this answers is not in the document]
mashehu: Same with the test.yml

modules/nf-core/nonpareil/tests/main.nf.test
```diff
@@ -0,0 +1,87 @@
...
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert process.out.npa.get(0).get(1) ==~ ".*/test.npa" },
+                { assert process.out.npc.get(0).get(1) ==~ ".*/test.npc" },
+                { assert path(process.out.npo.get(0).get(1)).getText().contains("# @impl: Nonpareil") },
+                { assert path(process.out.npl.get(0).get(1)).md5 == "be95453a272a9f685d5695d2b391fde8" },
```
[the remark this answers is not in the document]
jfy133: Can a snapshot be performed on a specific file?

modules/nf-core/nonpareil/tests/main.nf.test
```diff
@@ -0,0 +1,87 @@
...
```
nvnieuwk, line 36: Maybe also add these to the snapshot to reduce the maintenance on them in the future :)

modules/nf-core/nonpareil/tests/main.nf.test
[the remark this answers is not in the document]
nvnieuwk: Yes the snapshot is quite flexible, you can snapshot everything

tests/modules/nf-core/pyrodigal/main.nf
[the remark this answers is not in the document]
nvnieuwk: We decided to let them stay for now and only remove the tags in `pytest_modules.yml`. In case we need to revert back for some reason. When the conversion is fully done, we can bulk remove them all

modules/nf-core/nonpareil/meta.yml
```diff
@@ -23,8 +23,8 @@ input:
         e.g. `[ id:'test', single_end:false ]`
   - reads:
       type: file
-      description: Uncompressed FASTQ or FASTA file
-      pattern: "*.{fasta,fna,fas,fa,fastq,fq}"
+      description: FASTQ or FASTA file (ideally uncompressed, but not required)
+      pattern: "*.{fasta,fna,fas,fa,fastq,fq,fasta.gz,fna.gz,fas.gz,fa.gz,fastq.gz,fq.gz}},"
```
Emiller88, line 27:
```suggestion
      pattern: "*.{fasta,fna,fas,fa,fastq,fq,fasta.gz,fna.gz,fas.gz,fa.gz,fastq.gz,fq.gz}"
```

I think this is a typo?

modules/nf-core/nonpareil/tests/main.nf.test
Emiller88: Just thinking in the comments here. I think we need to make this a requirement. If there's a reason not to, then a comment with a note about why it's not in the snapshot?

tests/modules/nf-core/pyrodigal/main.nf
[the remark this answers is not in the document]
Emiller88: I think we were leaving them because it broke the linting 🤔 

So I'd rather we deleted them piecemeal, so we can easily see they were one to one. Whenever we can get around to the linting update.

Emiller88 (changes requested):
What this PR needs, is more snapshots!

modules/nf-core/nonpareil/tests/main.nf.test
nvnieuwk: I agree!!

tests/modules/nf-core/pyrodigal/main.nf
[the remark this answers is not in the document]
nvnieuwk: I'm not sure because linting wasn't even running for nf-test modules back then (I could be wrong though I don't remember the exact reason, just thought it was that one)

modules/nf-core/nonpareil/tests/main.nf.test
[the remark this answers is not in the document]
jfy133: Ooooooooooooooooooooooooooooooh nice! That wasn't clear to me from the docs

modules/nf-core/nonpareil/tests/main.nf.test
```diff
@@ -0,0 +1,87 @@
...
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert process.out.npa.get(0).get(1) ==~ ".*/test.npa" },
+                { assert process.out.npc.get(0).get(1) ==~ ".*/test.npc" },
+                { assert path(process.out.npo.get(0).get(1)).getText().contains("# @impl: Nonpareil") },
+                { assert snapshot(path(process.out.npl.get(0).get(1))).match() },
```
nvnieuwk, line 36:
```suggestion
                { assert snapshot(
                    file(process.out.npa[0][1]).name,
                    file(process.out.npc[0][1]).name,
                    path(process.out.npo[0][1]).readLines()[0],
                    path(process.out.npl[0][1]
                    ).match()
                }
                    
```

I think something like this might be easier to maintain :) This will snapshot the filename (without the full path) of the first two files, the first line of the 3rd file (please change this to whatever line you want to check since I don't know this format) and the md5sum of the last file

jfy133: Ooooh much more readable!

nvnieuwk (approved):
Awesome! LGTM :rocket:

modules/nf-core/nonpareil/tests/main.nf.test
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(
+                        file(process.out.npa[0][1]).name,
+                        file(process.out.npc[0][1]).name,
+                        path(process.out.npo[0][1]).readLines()[0],
+                        path(process.out.npl[0][1])
```
mashehu, line 37: assert snapshot(process.out.npa, ...) is not stable enough?

jfy133: npa/npc files change a lot (slightly stochastic numbering), as is npo.

Unless Id on't understand what you mean by 'stable enough'?

mashehu: that's exactly what I meant with stable enough. so good to go!

jfy133:
Ah just need @Emiller88 s ok now!

Emiller88 (approved):
Looks great!

PR #4318, opened 2023-11-10
require at least two elements in licence field if it is an array

mashehu:


modules/yaml-schema.json
```diff
@@ -159,15 +157,27 @@
                                 "items": {
                                     "type": "string"
                                 },
-                                "uniqueItems": true
+                                "uniqueItems": true,
+                                "minItems": 2,
```
mahesh-panchal, line 161: Maybe a silly question, but does this only apply when it's of type "array", or also when it's of type "string" ( and in which case what does that mean?).

mashehu: tested it and this option is ignored for strings.

mahesh-panchal (approved):
Without having much experience with this, LGTM

mirpedrol (approved):

modules/yaml-schema.json
```diff
@@ -154,20 +152,43 @@
...
+                                    }
+                                ],
                                 "description": "Licence of the tool",
                                 "items": {
                                     "type": "string"
                                 },
-                                "uniqueItems": true
+                                "uniqueItems": true,
+                                "minItems": 2,
+                                "message": "Licence must either be a string, e.g. \"MIT\" or have at least 2 items otherwise"
```
maxulysse, line 175:
```suggestion
                                "message": "Licence must either be a string, e.g. \"MIT\" or be an array with at least 2 items otherwise"
```

modules/yaml-schema.json
```diff
@@ -154,20 +152,43 @@
...
+                            },
+                            {
+                                "required": ["documentation"]
+                            },
+                            {
+                                "required": ["tool_dev_url"]
+                            },
+                            {
+                                "required": ["doi"]
+                            }
```
maxulysse, line 191: prettier?

PR #4323, opened 2023-11-13
add environment-schema.json, rename yaml-schema to meta-schema

mashehu:
In preparation for the 2.11 tools release


modules/environment-schema.json
```diff
@@ -0,0 +1,34 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "title": "conda environment yaml",
+    "description": "Validate the environment.yml file for an nf-core module",
+    "type": "object",
+    "properties": {
+        "name": {
```
mirpedrol, line 7: We don't add the name now, linting will fail, should we add it to all modules with this PR?

mirpedrol (approved):

modules/environment-schema.json
mashehu: ah, yes. let's wait for @Emiller88's PR to add the names first then

Emiller88 (approved):
Awesome! Adding this issue just as another check: https://github.com/nf-core/tools/issues/2515

https://github.com/nf-core/modules/pull/4327 is ready!

PR #4355, opened 2023-11-17
.nf-test -> nf-test

maxulysse:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -567,6 +567,7 @@ jobs:
       # Test the module
       - name: Run nf-test
         run: |
+          NXF_TEST_DIR=~ \
           nf-test test \
             --profile=${{ matrix.profile }} \
             --tag ${{ matrix.tags }} \
--- a/modules/nf-core/sratools/fasterqdump/tests/main.nf.test
+++ b/modules/nf-core/sratools/fasterqdump/tests/main.nf.test
@@ -22,7 +22,7 @@ nextflow_process {
         when {
             process {
                 """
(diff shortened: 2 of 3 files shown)


Emiller88:
Did we mean to add `NXF_TEST_DIR` as an environment variable and forgot? I think the `TMPDIR=~` is for singularity 🤔

Emiller88:
Ah, you said that in Slack
> also we don't set up NXF_TEST_DIR in the GHA, should we set it up to ~ as we used to do with TMPDIR?

I think we should do this, not change that, because it changes how the local tests are ran then.

maxulysse:
> Ah, you said that in Slack
> 
> > also we don't set up NXF_TEST_DIR in the GHA, should we set it up to ~ as we used to do with TMPDIR?
> 
> I think we should do this, not change that, because it changes how the local tests are ran then.

main issue is that if we run with `.nf-test` the tests fails locally unless we set up the `NXF_TEST_DIR` locally too

maxulysse:
I'd rather we have that in `/tmp` or whatever is equivalent as I had set it up at the beginning

sateeshperi (approved):

Emiller88:
> I'd rather we have that in `/tmp` or whatever is equivalent as I had set it up at the beginning

Looks like it was @adamrtalbot so we'll have to wait for an answer.

maxulysse:
`NXF_TEST_DIR=~` is good enough for me, that works on GHA, and whoever can set up the env to work how they like locally.
But I'm wondering if we should have used `TMP_DIR` instead actually

PR #4356, opened 2023-11-18
fix: correct output to be FASTA

Midnighter:
* KrakenUniq saved reads (classified and unclassified) were assumed to be in FASTQ format just as the input. However, in reality they are FASTA.
* Saved reads were assumed to be paired-end in case of paired input, in reality, paired input is merged for a single FASTA output file.
* Adjust filenames and patterns accordingly.
* Only supply the output option once to KrakenUniq, not twice.
* Describe the new expected input and output.
* Adapt the tests to the new output files.
* Remove an unused process section from the test configuration.

## PR checklist

Closes https://github.com/nf-core/taxprofiler/issues/420


modules/nf-core/krakenuniq/preloadedkrakenuniq/main.nf
```diff
@@ -11,17 +11,16 @@ process KRAKENUNIQ_PRELOADEDKRAKENUNIQ {
     tuple val(meta), path(fastqs)
     path  db
     val ram_chunk_size
-    val save_output_fastqs
+    val save_output_fasta
```
jfy133, line 14: I wounder if coudl be called `save_output_reads` to make it more generic?

modules/nf-core/krakenuniq/preloadedkrakenuniq/main.nf
```diff
@@ -30,13 +29,13 @@ process KRAKENUNIQ_PRELOADEDKRAKENUNIQ {
     def args = task.ext.args ?: ''
     def args2 = task.ext.args ?: ''
 
-    def classified   = meta.single_end ? '"\${PREFIX}.classified.fastq"'   : '"\${PREFIX}.classified#.fastq"'
-    def unclassified = meta.single_end ? '"\${PREFIX}.unclassified.fastq"' : '"\${PREFIX}.unclassified#.fastq"'
-    def classified_option = save_output_fastqs ? "--classified-out ${classified}" : ''
-    def unclassified_option = save_output_fastqs ? "--unclassified-out ${unclassified}" : ''
+    def classified   = meta.single_end ? '"\${PREFIX}.classified.fasta"'   : '"\${PREFIX}.merged.classified.fasta"'
+    def unclassified = meta.single_end ? '"\${PREFIX}.unclassified.fasta"' : '"\${PREFIX}.merged.unclassified.fasta"'
```
jfy133, line 33: Are they actually merged? I wonder if we need to include that in the file name... it might confuse people that the module is doin some internal module (as in, outside centrifuge itself), I think I would rather have just `un/classified.fasta` as sufficent. As far as I can see from your comments there still is not any difference between single or paired end output?

tests/modules/nf-core/krakenuniq/preloadedkrakenuniq/test.yml
```diff
@@ -4,24 +4,22 @@
     - krakenuniq/preloadedkrakenuniq
     - krakenuniq
   files:
-    - path: output/krakenuniq/test.classified_1.fastq.gz
-    - path: output/krakenuniq/test.classified_2.fastq.gz
     - path: output/krakenuniq/test.krakenuniq.classified.txt
     - path: output/krakenuniq/test.krakenuniq.report.txt
-    - path: output/krakenuniq/test.unclassified_1.fastq.gz
-    - path: output/krakenuniq/test.unclassified_2.fastq.gz
+    - path: output/krakenuniq/test.merged.classified.fasta.gz
```
jfy133, line 9: Don't want to take a stab at converting to nf-test? :grimacing: https://nf-co.re/docs/contributing/modules#migrating-from-pytest-to-nf-test

modules/nf-core/krakenuniq/preloadedkrakenuniq/main.nf
Midnighter: Of course

modules/nf-core/krakenuniq/preloadedkrakenuniq/main.nf
Midnighter: Not sure what centrifuge has to do with this.

When the input is paired-end, the reads are merged and output into a single FASTA file. So it makes sense to me to name it that way.

tests/modules/nf-core/krakenuniq/preloadedkrakenuniq/test.yml
Midnighter: Hmm, maybe.

modules/nf-core/krakenuniq/preloadedkrakenuniq/main.nf
Midnighter: Also, see my comment here https://github.com/nf-core/taxprofiler/issues/420#issuecomment-1817625086

jfy133 (approved):

tests/modules/nf-core/krakenuniq/preloadedkrakenuniq/test.yml
Midnighter: Well, wasn't so hard after all. (Famous last words.)

jfy133:
Aha, well we were both in both counts - it can be FASTA _or_ FASTQ depending on the input type :sweat_smile:

Midnighter:
I don't fully get your comment. Isn't output always FASTA? Do I need to change the modules again because I missed a case?

jfy133:
> I don't fully get your comment. Isn't output always FASTA? Do I need to change the modules again because I missed a case?

No it appears not.

As far as I can tell, the output is whatever the input format is.

If input is fastq, then output is FASTQ. If input is fasta, the output is fasta. I guess this why it forces us to specify the entire file name (with extension)...

Midnighter:
I'll run some tests and try to suggest a fix.

PR #4360, opened 2023-11-20
Removed the --offline flag

GallVp:
## PR checklist

Thank you for developing an excellent module for BUSCO.

- [x] Removed the `--offline` flag from `busco_lineage_dir`. This flag required that the BUSCO lineages are already downloaded and present at `busco_lineages_path`. If this is the case and the user wants to prohibit BUSCO from downloading the database, `--offline` flag can be added via `ext.args`. However, if the user wants a persistent `busco_lineages_path` across successive pipeline runs by a user with different inputs and wants BUSCO to manage the download of the required lineage(s), `busco_lineages_path` can be provided without adding the `--offline` flag through `ext.args`. This is documented in BUSCO help as: "--offline: To indicate that BUSCO cannot attempt to download files"


mahesh-panchal (commented):
I'm ok with this. @muffato I think this was something you specifically included.

muffato:
Hiya. I think it makes sense, @GallVp . We can always add `--offline` to `ext.args` indeed

muffato (approved):

mahesh-panchal:
Oh no... The maintainer update made it so that now the tool needs nf-test tests. @mashehu What's the protocol here?

mashehu:
we have a interim solution incoming, where we downgrade the fails to warnings for pytest https://github.com/nf-core/tools/pull/2524. will be ready today

mashehu:
The above mentioned PR is merged now

mashehu (approved):

PR #4361, opened 2023-11-21
Skipped cleaned_assembly in the absence of contamination

GallVp:
## PR checklist

- [x] When an assembly is clean, the `adaptor_report` is empty. Disk space can be saved by not emitting `cleaned_assembly` which in the case of a clean assembly is essentially a zipped copy of the input assembly.


d4straub (commented):
Thanks for the PR and sorry for the long delay.

The code looks fine to me.

But wouldn't it be better for any downstream application to receive the cleaned genome of the process? Following your proposed changes, the output here would be empty, i.e. any next steps would be _not_ executed in a nextflow pipeline.
I would argue that the metagenome assembly file of a pipeline such as nf-core/mag is not particularly big compared to the intermediary files. Is it? Whats your application?

d4straub:
Hi,

> I generally work with plant genomes which are typically 600 MBs ~ 2.4 GBs.

metagenome assemblies range in a similar size, and temp files dwarf that. But any reduction in size is good, I agree.

I agree that your pseudo code might work and solves the problem. But the change means every single pipeline that has any downstream processing of cleaned assemblies will need that pseudo code, correct? And (almost) every developer will waste time figuring this out. 

Maybe it would be more intuitive to the user of the module to not have to use such code but instead do in the module:
```
    num_contamination_lines=\$(cat "output/fcs_adaptor_report.txt" | wc -l)
    if [[ \$num_contamination_lines -gt 1 ]]; then
        find output/cleaned_sequences/  -type f ! -name "*.gz" -exec gzip {} \\;
        cp output/cleaned_sequences/*     "${prefix}.cleaned_sequences.fa.gz"
    else
        ln -s $assembly "${prefix}.cleaned_sequences.fa.gz"
    fi
```
(and remove `optional: true`)
therefore a symbolic link will be generated to the original assembly file and the module will always have an output.
I am not aware of a disadvantage of that approach, but maybe there is?

GallVp:
Hi @d4straub 

Symlink is a good idea. I am not sure if there is a way to exclude `clean_assembly` from being published to the `publishDir` if it is clean. Second, if the `assembly` extension is fasta and the `clean_assembly` extension is always fa.gz, it can create confusion and malfunction in downstream modules for clean assemblies.

Should we invite more people for review so that other users of this module can also chip in?

d4straub:
> I am not sure if there is a way to exclude clean_assembly from being published to the publishDir if it is clean. 

Hm yes indeed. Could be solved by naming it in two different ways, one of which is picked up by the publishing regex, but both need to be picked up by the output regex. But I agree that is a complication.

> Second, if the assembly extension is fasta and the clean_assembly extension is always fa.gz, it can create confusion and malfunction in downstream modules for clean assemblies.

I think (not entirely sure) that only fa.gz should arrive, at least from the pipeline itself, but I think there is now the possibility for inputting an assembly, if that input doesnt have the requirements that we need here, then I agree.

> Should we invite more people for review so that other users of this module can also chip in?

Absolutely, more opinions are helpful I think. Maybe advertise the discussion in nf-core slack if you dont have someone already in mind?

muffato:
Hi. I asked around at the Sanger Tree of Life department, and people said they're happy with the change because they only use the report.

SPPearce:
What is the decision going forward with this PR?

GallVp:
Hi @SPPearce 

I have version bumped fcsadaptor and have migrated it to nf-test. CI tests should pass now.

muffato:
It'd be good to record the snippet https://github.com/nf-core/modules/pull/4361#issuecomment-1828509600 in a place where users of this module could see it. Is there space for that in `meta.yml` ?

GallVp:
> It'd be good to record the snippet [#4361 (comment)](https://github.com/nf-core/modules/pull/4361#issuecomment-1828509600) in a place where users of this module could see it. Is there space for that in `meta.yml` ?

Hi @muffato 

I have added the comment to man.nf below the output channels.

https://github.com/nf-core/modules/blob/f0943fd59ed8576722f90fce047287177ee947e1/modules/nf-core/fcs/fcsadaptor/main.nf#L21

muffato (approved):
Thank you @GallVp !

PR #4378, opened 2023-11-22
add missing closing quote in deprecation comment

mashehu:


sateeshperi (approved):

.github/workflows/test.yml
```diff
@@ -392,9 +392,11 @@ jobs:
               issue_number: context.issue.number,
               owner: context.repo.owner,
               repo: context.repo.repo,
-              body: `Hello! We're in the process of migrating to [nf-test](https://code.askimed.com/nf-test)!🚀
+              body: 'Hello! We're in the process of migrating to [nf-test](https://code.askimed.com/nf-test)!🚀
```
mirpedrol, line 395: Will it fail if we have single quotes in the text? I think we should use `

mashehu: Using double quotes now

PR #4394, opened 2023-11-23
nf-test and snapshots for stringtie suite

sateeshperi:


maxulysse (approved):
LGTM

modules/nf-core/stringtie/merge/tests/main.nf.test
```diff
@@ -0,0 +1,82 @@
...
+                input[1] = file(params.test_data['homo_sapiens']['genome']['genome_gtf'], checkIfExists: true)
+                """
+            }
+        }
+
+        then {
+            assertAll(
+            { assert process.success },
+            { assert snapshot(process.out.gtf).match("fs_gtf") },
+            { assert snapshot(process.out.versions).match("fs_versions") }
```
maxulysse, line 42: I think @mashehu preferred to just snapshot `process.out` to be more concise.

sateeshperi: there are few emits that are optional outputs...so cannot do `process.out`

PR #4412, opened 2023-11-24
abacas nf-test snapshot

sateeshperi:


maxulysse (approved):

modules/nf-core/abacas/tests/main.nf.test
```diff
@@ -0,0 +1,38 @@
...
+                           ]
+                input[1] = file(params.test_data['sarscov2']['genome']['genome_fasta'], checkIfExists: true)
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out.results[0][1].findAll {
```
mashehu, line 27: I think we should check at least the filename for the files excluded here.

mashehu (approved):

modules/nf-core/abacas/tests/main.nf.test
sateeshperi: good point...don't think its done implicitly....will add it in...give me few min

sateeshperi: plz check and merge if ready. thanks

mashehu (approved):

PR #4419, opened 2023-11-26
fastq_subsample_fq_salmon nf-test snapshot

sateeshperi:


sateeshperi:
can't get salmon index and results into snapshot....if anybody can give it a shot.....else we can check for any important files in publishDir as last resort

test.tap
```diff
@@ -1,2 +0,0 @@
-1..1
```
sateeshperi, line 1: this file sneaked in somehow...so deleting

adamrtalbot (approved):
Not the cleanest, but it'll do.

Issue #4451, opened 2023-11-28
[FEATURE] nf-test for vcf2maf


stevekm:
### Is your feature request related to a problem? Please describe

nf-test for vcf2maf

PR #4463, opened 2023-11-28
4451 add vcf2maf nf test

stevekm:
## PR checklist

Closes [#4451](https://github.com/nf-core/modules/issues/4451) 

- [ ] Add a resource `label`

### DIFF
--- a/modules/nf-core/vcf2maf/tests/main.nf.test
+++ b/modules/nf-core/vcf2maf/tests/main.nf.test
@@ -2,12 +2,13 @@ nextflow_process {
     name "Test vcf2maf"
     process "VCF2MAF"
     script "../main.nf"
-    config "./nextflow.config"
+
     tag "modules"
     tag "modules_nfcore"
     tag "vcf2maf"
 
     test("Run without VEP") {
+        config "./nextflow.config"
         when {
             process {
                 """
@@ -28,4 +29,43 @@ nextflow_process {
             )
         }
     }
-}
+
+    test("Run with VEP"){
+        tag "vcf2maf-withVEP"
+        config "./nextflow.withVEP.config"
+
+        setup {
+            run("UNTAR") {
+                script "../../../../modules/nf-core/untar/main.nf"
+                process {
+                """
+                input[0] = [ [ id:'test_genome_vep' ], file(params.test_data['homo_sapiens']['genome']['vep_cache'], checkIfExists: true) ]
+                """
+            }
+        }
+        }
+
+        when {
+            process {
+                """
+                // define inputs of the process here. Example:
+                input[0] = [
+                                [ id:'test'], // meta map
+                                file(params.test_data['homo_sapiens']['illumina']['test_genome_vcf'], checkIfExists: true)
+                            ]
+                input[1] = [ file(params.test_data['homo_sapiens']['genome']['genome_fasta'], checkIfExists: true) ]
+                input[2] = UNTAR.out.untar.map { it[1] }
+                """
+            }
+        }
+
+        then{
+            assertAll(
+            { assert process.success },
+            { assert snapshot(process.out).match() }
+            )
+        }
+    }
+
+    }
+
--- a/modules/nf-core/vcf2maf/tests/main.nf.test
+++ b/modules/nf-core/vcf2maf/tests/main.nf.test
@@ -1,0 +1,31 @@
+nextflow_process {
+    name "Test vcf2maf"
+    process "VCF2MAF"
+    script "../main.nf"
+    config "./nextflow.config"
+    tag "modules"
+    tag "modules_nfcore"
+    tag "vcf2maf"
+
+    test("Run without VEP") {
+        when {
+            process {
+                """
+                // define inputs of the process here. Example:
+                input[0] = [
+                                [ id:'test'], // meta map
+                                file(params.test_data['homo_sapiens']['illumina']['test_genome_vcf'], checkIfExists: true)
+                            ]
+                input[1] = [ file(params.test_data['homo_sapiens']['genome']['genome_fasta'], checkIfExists: true) ]
+                input[2] = []
+                """
+            }
+        }
+        then {
+            assertAll(
+            { assert process.success },
+            { assert snapshot(process.out).match() }
+            )
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


ahvigil:
you can also remove the old tests at https://github.com/nf-core/modules/tree/master/tests/modules/nf-core/vcf2maf

modules/nf-core/vcf2maf/tests/nextflow.config
```diff
@@ -0,0 +1,12 @@
+// config for running without VEP
+params.species = "homo_sapiens"
+params.build   = "GRCh38"
+
+process {
+            publishDir = { "${params.outdir}/${task.process.tokenize(':')[-1].tokenize('_')[0].toLowerCase()}" }
+            withName: 'VCF2MAF' {
+                ext.args = "--ncbi-build ${params.build} --species ${params.species} --inhibit-vep"
+                // NOTE: using a different container here than in the main.nf for VCF2MAF
```
johnoooh, line 9: Maybe just add a note in the pull request documenting why a different container was used for the test.   Was there an error with the other one?  Its fine to do this IMO but just specify why somewhere.

johnoooh (approved):

stevekm:
so the problem with this PR is that when it runs in docker / singularity, it uses an alternate container for running with / without VEP, which influences the output

when it runs with conda, it uses the same environment for both, which causes the outputs to be inconsistent with the test cases

not sure what the best way to resolve this is. I feel like the conda is wrong, but I am not sure how to fix it to make the test pass. I suggest just ignoring the conda, because changing the conda env would likely impact users who are using the module with the specified env.

@sateeshperi

modules/nf-core/vcf2maf/tests/nextflow.withVEP.config
```diff
@@ -0,0 +1,10 @@
+// config for running with VEP
+params.species = "homo_sapiens"
+params.build   = "GRCh38"
```
maxulysse, line 3: Can we use Wbcel235 for the tests, that should be faster

maxulysse:
> so the problem with this PR is that when it runs in docker / singularity, it uses an alternate container for running with / without VEP, which influences the output
> 
> when it runs with conda, it uses the same environment for both, which causes the outputs to be inconsistent with the test cases
> 
> not sure what the best way to resolve this is. I feel like the conda is wrong, but I am not sure how to fix it to make the test pass. I suggest just ignoring the conda, because changing the conda env would likely impact users who are using the module with the specified env.
> 
> @sateeshperi

How about we alway use the container with VEP?

modules/nf-core/vcf2maf/tests/nextflow.withVEP.config
adamrtalbot: The VCF, the VEP cache and the genome FASTA all need to match up so they will all need to be changed.

stevekm:
its still failing because the output file is different when running with only conda

```
                "test.maf:md5,c7f357efa774fd5c4003d84f05212ed0"					   |	                "test.maf:md5,b89c598af3489ac987ed6c537470e1d1"
```

stevekm:
also note that its not possible to debug this locally because the conda env used here does not work for M1 Mac

stevekm:
> Run actions/setup-python@v4
> Installed versions
>   Version 3.11 was not found in the local cache
>   Error: API rate limit exceeded for installation ID 1565271. If you reach out to GitHub Support for help, please include the request ID E16B:8E76:1BFFBBEF:1C524193:6567A021.

:(

SPPearce:
Can we figure out a way to finish this off?

modules/nf-core/vcf2maf/tests/main.nf.test
```diff
@@ -0,0 +1,79 @@
...
+        }
+        then {
+            // NOTE: need this for debugging but remove for the PR when ready
+            for (line in path(process.out.maf[0][1]).readLines()){
+                println line
+            }
+            assertAll(
+            { assert process.success },
+            { assert snapshot(process.out).match() },
+            { assert path(process.out.maf[0][1]).readLines().size() == 132 },
```
adamrtalbot, line 33: We could consider slicing the contents of the VCF and adding them to the snapshot? See the documentation: https://www.nf-test.com/docs/assertions/files/#filter-lines

Untested, the syntax might be a bit wrong:
```suggestion
            { assert snapshot(
                  process.out,
                  process.out.maf[0][1].gzip.get(0)).linesGzip[5..132]
              ).match() 
            },
```

adamrtalbot (commented):
Well done, VEP and VEP related tools are such a pain to work with.

SPPearce:
Bumpity @stevekm or @adamrtalbot  or anyone else willing to do it, can we get this finished?

modules/nf-core/vcf2maf/tests/main.nf.test
```diff
@@ -0,0 +1,76 @@
...
+    tag "modules_nfcore"
+    tag "vcf2maf"
+    tag "untar"
+
+    test("Run without VEP") {
+        config "./nextflow.config"
+        when {
+            process {
+                """
+                // define inputs of the process here. Example:
```
SPPearce, line 16:
```suggestion
```

modules/nf-core/vcf2maf/tests/main.nf.test
```diff
@@ -0,0 +1,76 @@
...
+                input[0] = [ [ id:'test_genome_vep' ], file(params.test_data['homo_sapiens']['genome']['vep_cache'], checkIfExists: true) ]
+                """
+            }
+        }
+        }
+
+        when {
+            process {
+                """
+                // define inputs of the process here. Example:
```
SPPearce, line 56:
```suggestion
```

SPPearce (approved):

SPPearce:
Note that on conda this is testing with VEP in both tests, but that is fine.

SPPearce:
> Note that on conda this is testing with VEP in both tests, but that is fine.

By this I mean it is using a conda environment with VEP as well as vcf2maf in both tests, whereas the docker/singularity tests use a container without VEP for one of the test.

SPPearce:
> > > Note that on conda this is testing with VEP in both tests, but that is fine.
> > 
> > 
> > By this I mean it is using a conda environment with VEP as well as vcf2maf in both tests, whereas the docker/singularity tests use a container without VEP for one of the test.
> 
> I think I've fixed it with [61a9d20](https://github.com/nf-core/modules/commit/61a9d20599e21e79beaa26e10c4d1629c2a1dd37) but not certain...

Well, it has made a difference to the snapshot, so is now presumably using the different environment.

SPPearce:
Oh, thought we'd finished this, but apparently not quite.

SPPearce (approved):

PR #4478, opened 2023-11-28
ariba/run nf-test

buehlere:
:heavy_check_mark: adding test folder nf-test
:heavy_check_mark: removing from pytest_modules.yml

## PR checklist

Closes #XXX 

- [x] Adding nf-test to ariba run
- [x] Add a resource `label`


modules/nf-core/ariba/run/tests/main.nf.test.snap
```diff
@@ -0,0 +1,18 @@
...
+                "versions.yml:md5,98363d30a08faf8ebda127cf44584ebe"
+            ]
+        ],
+        "timestamp": "2023-11-28T16:56:05.056901"
+    },
+    "illumina - fatsq": {
+        "content": [
+            [
+                
+            ]
```
sateeshperi, line 14: why is this empty ?

rpetit3: Haha beat me to it @sateeshperi 

Here's the previous pytest outputs: https://github.com/nf-core/modules/blob/master/tests/modules/nf-core/ariba/run/test.yml

rpetit3:
@nf-core/modules are the 49 CI checks on this PR expected?

nvnieuwk:
Seems like the branch was outdated, should be fixed now :)

modules/nf-core/ariba/run/tests/main.nf.test.snap
buehlere: I was having some trouble with the MD5s yesterday. But figured out what I was doing wrong. Working on matching the original test now!

buehlere: Okay! I believe I'm checking all expected outputs in the snapshot now: https://github.com/nf-core/modules/blob/e5459cf998c1c3ada532d48b0fb2365e0041e6b5/modules/nf-core/ariba/run/tests/main.nf.test.snap

I also added in the [contain tests](https://github.com/nf-core/modules/blob/e5459cf998c1c3ada532d48b0fb2365e0041e6b5/modules/nf-core/ariba/run/tests/main.nf.test#L48-L59) as well.

rpetit3:
Thank you @nvnieuwk nice to know!

buehlere:
> Seems like the branch was outdated, should be fixed now :)

thanks for the catch!

modules/nf-core/ariba/run/tests/main.nf.test.snap
rpetit3: Awesome @buehlere! this is learning experience for me as well. I'll defer to @sateeshperi 

@sateeshperi I'm wondering if `contains` can be given a list like in pytest. What would your recommendations be here? 

Haha I'm thinking of all the pytest tests that over use contains

apetkau (commented):
This looks awesome 😄. Should the files under `tests/modules/nf-core/ariba/run` be removed since this is switching to nf-test away from pytest?

buehlere:
> This looks awesome 😄. Should the files under `tests/modules/nf-core/ariba/run` be removed since this is switching to nf-test away from pytest?

Good catch! I just removed.

modules/nf-core/ariba/run/tests/main.nf.test
```diff
@@ -0,0 +1,65 @@
...
+    script "../main.nf"
+    process "ARIBA_RUN"
+    config "./nextflow.config"
+
+    tag "modules"
+    tag "modules_nfcore"
+    tag "ariba"
+    tag "ariba/run"
+
+    test("illumina - fatsq") {
```
sateeshperi, line 14:
```suggestion
    test("sarscov2 illumina  paired-end [fastq]") {
```

modules/nf-core/ariba/run/tests/main.nf.test
```diff
@@ -0,0 +1,65 @@
+// nf-core modules test ariba/run
```
sateeshperi, line 1:
```suggestion
```

modules/nf-core/ariba/run/tests/main.nf.test
```diff
@@ -0,0 +1,65 @@
...
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(file(process.out.results[0][1][0]).name).match("assembled_genes.fa.gz") },
+                { assert snapshot(file(process.out.results[0][1][1]).name).match("assembled_seqs.fa.gz") },
+                { assert snapshot(file(process.out.results[0][1][2]).name).match("assemblies.fa.gz") },
```
sateeshperi, line 44: why aren't any of these gzipped fasta files generating any consistent md5sums ?

sateeshperi: @buehlere :point_up:

buehlere: @sateeshperi I think this is because the compression done by ariba keeps the time stamp on the gzipped files:

image

I was reading that this can cause problems with the md5sums [here](https://unix.stackexchange.com/questions/31008/why-does-the-gzip-version-of-files-produce-a-different-md5-checksum). This might be why they only checked for the path on the gzipped files in [the original test.](https://github.com/nf-core/modules/blob/master/tests/modules/nf-core/ariba/run/test.yml#L7-L15)?

Is there a way to save the original and re-generated nf-test files for a comparison to prove this? I'm not sure I see each distinct run in the work directory: 

image

@apetkau I maybe remember you hitting a similar issue with time stamps on gz files. Is there a good way to demonstrate this?

apetkau: @buehlere yes, I encountered a similar issue with md5's changing for gzip files. It was resolved for me by using the `-n` option to `gzip`:

> -n
When compressing, do not save the original file name and timestamp by default. (The original name is always saved if the name had to be truncated.) When decompressing, do not restore the original file name if present (remove only the gzip suffix from the compressed file name) and do not restore the original timestamp if present (copy it from the compressed file). This option is the default when decompressing.
https://www.gnu.org/software/gzip/manual/html_node/Invoking-gzip.html

However, I believe `ariba` is gzipping the files internally, so if this was the issue it would have to be fixed within `ariba` itself (or md5 sums ignored for those particular files/other tests performed).

PR #4496, opened 2023-11-29
remove parameter --bcl-num-parallel-tiles

FelixKrueger:
Remove parameter `--bcl-num-parallel-tiles` as it causes poor performance or run failures

https://github.com/nf-core/demultiplex/issues/156

## PR checklist

Closes https://github.com/nf-core/demultiplex/issues/156 


Emiller88 (approved):
So close to a negative PR!

PR #4521, opened 2023-11-30
Remove pytest-workflow tests for modules covered by nf-test

Emiller88:
Stuff that we didn't cut out during the Barcelona Hackathon because the linting wasn't up to date yet.

If anyone wants to volunteer to go through and make sure all of the tests had equivalent coverage, have at it!


grst (approved):
Looks good! 

Btw, as the pytests are being removed, did you come up with a solution to substitute the [valid version yaml checker](https://github.com/nf-core/modules/blob/master/tests/test_versions_yml.py) in nf-test?

PR #4545, opened 2023-12-06
Condense testing and pre-commit workflow into single workflow

adamrtalbot:


Emiller88 (approved):
Awesome, thanks for the detailed explaination, I'm caught up now!

Documented it here https://github.com/nf-core/ops/commit/a76929cf2e11777d360c319554ca895c9899683f for other to stumple on.

PR #4553, opened 2023-12-07
Revert "ariba/run nf-test"

buehlere:
Reverts nf-core/modules#4478 I accidentally added this to the merge queue.


sateeshperi:
not sure if we want to revert and merge again or just correct already merged one....thoughts @mashehu

sateeshperi:
related PR #4478

mashehu:
Agree no need to revert (since the previous version looks like it had failing tests). Just fix ASAP. 🙂

PR #4554, opened 2023-12-07
Condense linting workflows to one job

adamrtalbot:
- Condense multiple linting jobs into single linting job
- Condense two check-changes jobs to single job with two outputs

Will require a change on the branch protection.


mashehu:
Why don't we run these linting steps as part of pre-commit?

PR #4563, opened 2023-12-10
adapterremoval nf-test snapshot

sateeshperi:


kpadm (commented):
Looks good to me! @sateeshperi

modules/nf-core/adapterremoval/tests/nextflow.config
```diff
@@ -0,0 +1,4 @@
+process {
```
SPPearce, line 1: We don't need the empty config?

SPPearce (approved):
Seems to have an empty config file, but fine apart from that

modules/nf-core/adapterremoval/tests/nextflow.config
sateeshperi: don't need an empty file

sateeshperi:
thank you @kpadm @SPPearce

PR #4564, opened 2023-12-10
Multiqc integration tests PoC

edmundmiller:
Just had a thought that we could maybe do some better integration testing with MultiQC.

I'm thinking it might be mutually beneficial to test all of the MultiQC modules in one place.

I'm not sure if this should live here, or in another repo, or in MultiQC. Just wanted to start the conversation. Probably not a top priority.

### DIFF
--- a/modules/nf-core/multiqc/tests/e2e/main.nf.test
+++ b/modules/nf-core/multiqc/tests/e2e/main.nf.test
@@ -4,6 +4,7 @@ nextflow_pipeline {
     script "./main.nf"
 
     tag "fastqc"
+    tag "bowtie2/align"
 
     test("Single-Read") {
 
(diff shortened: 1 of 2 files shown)


modules/nf-core/multiqc/tests/e2e/main.nf.test
```diff
@@ -0,0 +1,29 @@
...
+        when {
+            params {
+                outdir   = "$outputDir"
+            }
+        }
+
+        then {
+            assertAll (
+            { assert process.success },
+            // NOTE The report contains the date inside it, which means that the md5sum is stable per day, but not longer than that. So you can't md5sum it.
```
ewels, line 20: There's a MultiQC option to omit the date from the report, I think

Emiller88: That was just copied from the stock test 🙃 

> I think
😆

luisas:
@edmundmiller Is this one still relevant? :)

PR #4569, opened 2023-12-11
unset `$JAVA_TOOL_OPTIONS` in gitpod

mirpedrol:


mashehu (approved):

mashehu:
I guess we need the same for tools and the tempalte?

mirpedrol:
> I guess we need the same for tools and the pipeline template?
I will open the PR now 👍 
I have tried to add this to the Dockerfile, but I learnt that there isn't a way to unset an ENV variable from a Dockerfile, so we have to do it when creating the Gitpod environment. If you know a better way, let me know! 😄

PR #4571, opened 2023-12-11
Added a SAMTOOLS_PIPELINE to run multiple samtools commands at once

muffato:
I've recently spent some time optimising the resource usage of some Nextflow pipelines and I found that the [samtools workflow to mark duplicates](https://www.htslib.org/algorithms/duplicate.html#workflow) takes 5x longer to run and 3x more disk space when implemented the nf-core way, i.e. as 4 modules put together in a sub-workflow, compared to having a single local module with a bash pipeline.

Last time we discussed this sort of module, #3310 and https://nfcore.slack.com/archives/CJRH30T6V/p1682062426042809 , it was decided against though we didn't know the actual improvements. But both @drpatelh and @SPPearce suggested to reopen the discussion after seeing the figures. After all, the nf-core guidelines don't _forbid_ it (bold type mine).
> Software that can be piped together SHOULD be added to separate module files **unless there is a run-time, storage advantage in implementing in this way**.

However, @drpatelh said that there should be options to disable some of the 4 commands, because not anyone may need them. Here is my take on this.

I introduce a new module that can run a [bash _pipeline_](https://www.gnu.org/software/bash/manual/html_node/Pipelines.html) (hence the name) of any commands the user wants in any order, as long as they take BAM/CRAM as inputs as outputs. It wasn't as trivial as I thought it would because samtools commands have different ways of defining the input and output files 🙃 

It works like this:
```nextflow
workflow {
    input = [
        file(params.test_data['sarscov2']['illumina']['test_paired_end_bam'], checkIfExists: true)
    ]
    commands = ['collate', 'fixmate', 'sort', 'markdup']
    SAMTOOLS_PIPELINE ( input, [[],[]], commands )
    SAMTOOLS_PIPELINE.out.output
}
process {
    withName: SAMTOOLS_PIPELINE {
        ext.args = [
            fixmate: '-m',
        ]
    }
}
```

What I like about the approach:

- No mulled container involved. Just the regular samtools, which is much easier to understand and update.
- Easy to add more samtools commands to the module.
- Maximum flexibility for the user in terms of chaining samtools commands.

What I dislike about the implementation:

- I couldn't find how to dynamically get the value of `task.ext.args${index}`. So instead I make `ext.args` a map where the key is the command name and the value is the string to add to the command-line. I don't know if/how closures can be used. I would very much prefer to use the regular `task.ext.args`, `task.ext.args2`, etc.
- Because of the above, my implementation doesn't handle the same command being used multiple times in the pipeline with different command-line options.
- The configuration of the module is actually split between the `.nf` (`commands = ['collate', 'fixmate', 'sort', 'markdup']`) and the `.config` (`ext.args = [ ... ]`). Maybe I should move the definition of the commands to the `.config` ?
- Indentation of the resulting `.command.sh` is very fragile. What I've got works but for instance I tried `pipeline_command += "samtools ${this_command} ${all_args[index]} \\ \n" and that caused the `cat <<-END_VERSIONS > versions.yml` to be indented, which broke the generation of the yml.

What it needs:

- [x] testing with CRAM. I'm no sure I put the `--reference` in the right place.
- [ ] migration to nf-test.

What it may need:

- The selected commands often have command-line options to take extra files as inputs or output more files. Those could be added to the module, though to be fair the existing modules often don't list every single possible option either.
- Some samtools-based sub-workflows could be rewritten to use this module, bringing an instant performance boost to their workflows.
- A different name ? I chose "pipeline" because it's a bash pipeline under the hood. @mashehu doesn't like it :) Alternatives: pipe, chain, any, multi, combine. Ideally it should be name that stands out from the already-crowded list of samtools commands
- A simple sub-workflow wrapper so that people looking for samtools sub-workflows still find about this. (they may not think about checking modules)
- Alternatively, only expose this capability as a sub-workflow, either by allowing "local" modules in nf-core sub-workflows and moving the file over there, or by embedding this `process` block in the sub-workflow's `main.nf`.

## PR checklist

- [x] Add a resource `label`

### DIFF
--- a/modules/nf-core/samtools/pipeline/main.nf
+++ b/modules/nf-core/samtools/pipeline/main.nf
@@ -46,29 +46,29 @@ process SAMTOOLS_PIPELINE {
             case !"reheader":
                 // "reheader" is the only command not to offer these
                 command << "-@ $task.cpus"
-                command << fasta && last ? "--reference ${fasta}" : ''
-                command << !last ? '-u' : ''
+                command << (fasta && last ? "--reference ${fasta}" : '')
+                command << (!last ? '-u' : '')
                 // NOTE: no "break" here because we want to run the next batch of "case"
 
             //// Then the input/ouput parameters, which differ between commands
             case "collate":
                 // [-o OUTPUT|-O] [INPUT|-]
-                command << last ? "-o ${prefix}.${extension}" : "-O"
-                command << first ? input : '-'
+                command << (last ? "-o ${prefix}.${extension}" : "-O")
+                command << (first ? input : '-')
                 break
             case ["addreplacerg", "sort", "view"]:
                 // [-o OUTPUT] [INPUT|-]
-                command << last ? "-o ${prefix}.${extension}" : ""
-                command << first ? input : '-'
+                command << (last ? "-o ${prefix}.${extension}" : "")
+                command << (first ? input : '-')
                 break
             case "reheader":
                 // [INPUT|-]
-                command << first ? input : '-'
+                command << (first ? input : '-')
                 break
             case ["fixmate", "markdup"]:
                 // [INPUT|-] [OUTPUT|-]
-                command << first ? input : '-'
-                command << last ? "${prefix}.${extension}" : "-"
+                command << (first ? input : '-')
+                command << (last ? "${prefix}.${extension}" : "-")
                 break
             default:
                 assert false: "$cmd is not supported"
--- a/modules/nf-core/samtools/pipeline/main.nf
+++ b/modules/nf-core/samtools/pipeline/main.nf
@@ -86,7 +86,15 @@ process SAMTOOLS_PIPELINE {
     """
 
     stub:
-    def prefix = task.ext.prefix ?: "${meta.id}"
+    def prefix    = task.ext.prefix ?: "${meta.id}"
+    def cmd_size  = commands.size()
+    def last_args = task.ext."args$cmd_size" ?: ''
+    def extension = last_args.contains("--output-fmt sam") ? "sam" :
+                    last_args.contains("--output-fmt bam") ? "bam" :
+                    last_args.contains("--output-fmt cram") ? "cram" :
+                    input.extension
+    assert "$input" != "${prefix}.${extension}" : "Input and output names are the same, use \"task.ext.prefix\" to disambiguate!"
+
     """
     touch ${prefix}.${extension}
 
--- a/modules/nf-core/samtools/pipeline/main.nf
+++ b/modules/nf-core/samtools/pipeline/main.nf
@@ -40,12 +40,15 @@ process SAMTOOLS_PIPELINE {
             task.ext."args${first ? '' : index+1}" ?: ''
         ]
         switch(cmd){
+            //// First the common options
             case !"reheader":
-                // The reheader has no useful option
+                // "reheader" is the only command not to offer these
                 command << "-@ $task.cpus"
                 command << fasta && last ? "--reference ${fasta}" : ''
                 command << !last ? '-u' : ''
-            // samtools commands have slightly different syntax
+                // NOTE: no "break" here because we want to run the next batch of "case"
+
+            //// Then the input/ouput parameters, which differ between commands
             case "collate":
                 // [-o OUTPUT|-O] [INPUT|-]
                 command << last ? "-o ${prefix}.${extension}" : "-O"
--- a/modules/nf-core/samtools/pipeline/main.nf
+++ b/modules/nf-core/samtools/pipeline/main.nf
@@ -89,4 +89,15 @@ process SAMTOOLS_PIPELINE {
         samtools: \$(echo \$(samtools --version 2>&1) | sed 's/^.*samtools //; s/Using.*\$//')
     END_VERSIONS
     """
+
+    stub:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    touch ${prefix}.${extension}
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        samtools: \$(echo \$(samtools --version 2>&1) | sed 's/^.*samtools //; s/Using.*\$//')
+    END_VERSIONS
+    """
 }
--- a/modules/nf-core/samtools/pipeline/main.nf
+++ b/modules/nf-core/samtools/pipeline/main.nf
@@ -33,6 +33,8 @@ process SAMTOOLS_PIPELINE {
     assert "$input" != "${prefix}.${extension}" : "Input and output names are the same, use \"task.ext.prefix\" to disambiguate!"
     // Compose pipe
     def cmds = commands.indexed().collect { index, cmd ->
+        def first = index == 0
+        def last = index == cmd_size-1
         def command = [
             "samtools $cmd",
             task.ext."args${index!=0 ? index+1 : ''}" ?: ''
@@ -41,27 +43,27 @@ process SAMTOOLS_PIPELINE {
             case !"reheader":
                 // The reheader has no useful option
                 command << "-@ $task.cpus"
-                command << fasta && index == cmd_size ? "--reference ${fasta}" : ''
-                command << index != cmd_size ? '-u' : ''
+                command << fasta && last ? "--reference ${fasta}" : ''
+                command << !last ? '-u' : ''
             // samtools commands have slightly different syntax
             case "collate":
                 // [-o OUTPUT|-O] [INPUT|-]
-                command << index == cmd_size ? " -o ${prefix}.${extension}" : " -O"
-                command << index == 0 ? index : '-'
+                command << last ? "-o ${prefix}.${extension}" : "-O"
+                command << first ? index : '-'
                 break
             case ["addreplacerg", "sort", "view"]:
                 // [-o OUTPUT] [INPUT|-]
-                command << index == cmd_size ? "-o ${prefix}.${extension}" : ""
-                command << index == 0 ? index : '-'
+                command << last ? "-o ${prefix}.${extension}" : ""
+                command << first ? index : '-'
                 break
             case "reheader":
                 // [INPUT|-]
-                command << index == 0 ? index : '-'
+                command << first ? index : '-'
                 break
             case ["fixmate", "markdup"]:
                 // [INPUT|-] [OUTPUT|-]
-                command << index == 0 ? index : '-'
-                command << index == cmd_size ? "${prefix}.${extension}" : "-"
+                command << first ? index : '-'
+                command << last ? "${prefix}.${extension}" : "-"
                 break
             default:
                 assert false: "$cmd is not supported"
--- a/modules/nf-core/samtools/pipeline/main.nf
+++ b/modules/nf-core/samtools/pipeline/main.nf
@@ -14,6 +14,7 @@ process SAMTOOLS_PIPELINE {
 
     output:
     tuple val(meta), path("*.{bam,cram,sam}"), emit: output
+    tuple val(meta), path("*.{bai,csi,crai}"), emit: index, optional: true
     path "versions.yml",                       emit: versions
 
     when:
--- a/modules/nf-core/samtools/pipeline/main.nf
+++ b/modules/nf-core/samtools/pipeline/main.nf
@@ -1,0 +1,92 @@
+process SAMTOOLS_PIPELINE {
+    tag "$meta.id"
+    label 'process_medium'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/samtools:1.18--h50ea8bc_1':
+        'biocontainers/samtools:1.18--h50ea8bc_1' }"
+
+    input:
+    tuple val(meta), path(input)
+    tuple val(meta2), path(fasta)
+    val commands
+
+    output:
+    tuple val(meta), path("*.{bam,cram,sam}"), emit: output
+    path "versions.yml",                       emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+
+    // Check that we are asked to run a pipeline
+    def n_commands = commands.size()
+    if (n_commands < 2) error "SAMTOOLS_PIPELINE is used to chain 2 or more samtools commands"
+
+    // Fetch the arguments
+    def all_args = []
+    for (int index = 0; index < n_commands; index++) {
+        all_args.add( task.ext.args?[ commands[index] ] ?: '' )
+    }
+    def last_args = all_args[-1]
+
+    // Output file
+    def prefix = task.ext.prefix ?: "${meta.id}"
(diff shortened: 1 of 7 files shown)


maxulysse:
Love the idea

mashehu (commented):
Please don't call it pipeline, that is already too overloaded. 
I am also not 100% sure about the universality of this. Feels like we are basically moving away from atomic modules again and one needs to dig through several configs to read through several configs and docs to figure out what the module/command is  actually doing

muffato:
Hi guys. You were quick 😉  ! I am still testing a few things locally. I was then going to fill in the body of the PR to explain how this came up and share the link on Slack for discussion. I know we already discussed "samtools sormadup" and decided against, so I don't expect an easy ride 😅 but I think there are good reasons to still do something on this topic.

SPPearce:
As someone whose cluster is currently running out of scratch space, I definitely support the idea of chaining multiple samtools operations into one module, producing lots of individual bam/cram files uses a lot of memory when not necessary.
We already pipe `bwa` directly into `samtools` for this exact reason, bam/cram files can be BIG.

modules/nf-core/samtools/pipeline/main.nf
```diff
@@ -0,0 +1,112 @@
...
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        samtools: \$(echo \$(samtools --version 2>&1) | sed 's/^.*samtools //; s/Using.*\$//')
+    END_VERSIONS
+    """
+
+    stub:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    touch ${prefix}.${extension}
```
nvnieuwk, line 105: Not really what you wanted comments on, but this stub won't work since the `extension` isn't defined in the stub scope (all output files will be `${prefix}.null`)

muffato: I noticed that too :)

mahesh-panchal:
I definitely approve of the idea. 

In terms of implementation, I would skip the for loop and only use if then else to make an array, and then `join` the string array with the pipe symbol when evaluating the string array in it's execution place. I.e.
L94 would look like:
```
${samtools_commands.join(' | ')}
```

I'm starting to think we need to revisit also how args are passed to modules. I would prefer that the args are indexed with the 'tool_subtool' key in this case for readability.

muffato:
> In terms of implementation, I would skip the for loop and only use if then else to make an array, and then `join` the string array with the pipe symbol when evaluating the string array in it's execution place. I.e.

I like the idea of `join`. Will definitely implement it. But I still need a `for` loop of some sort to go over the command names, right ? If not an actual `for`, then perhaps a `.collect` with a closure ?

mahesh-panchal:
From what I see in the tests, the commands are a particular order, which means they can be dealt with linearly in that order with an if/elseif, or am I missing something? Within each section, construct the command and push it on to the array ( or add at position if dealing with out of order ).

muffato:
The user asks for a particular order, but as far as the module in concern, all combinations are valid. The module does not try to make sense of the user request

mahesh-panchal:
One can get the index of the command being constructed and insert the constructed command at the same position in the output array. This is all assuming a command can only be listed once in the input array, otherwise you would need a loop of some sort.

muffato:
The listing of the input and outputs differs would differ depending on whether it's the first, in the middle, or the last. As far as I understand, that's 3 positions times 7 commands = 21 scenarios that would have to be pre-generated so that the final combination can be as easy as a get and join.
I think it's possible to get rid of the `for` loop, but it feels it'd just be for the sake of it. I'm not seeing yet how it would make the module easier to read

mahesh-panchal:
You could use the same array join syntax to make the commands too, like is done with some of the ext.args sometimes when params are checked. I'll try and make some time tomorrow to make a code suggestion (or realise the idea is not worth it).

modules/nf-core/samtools/pipeline/main.nf
```diff
@@ -0,0 +1,112 @@
...
+            error "${this_command} is not supported"
+        }
+
+        if (!is_last_command) {
+            pipeline_command += " |"
+        }
+    }
+
+    """
+    $pipeline_command
```
mahesh-panchal, line 94:
```suggestion
    script:
    // Check that we are asked to run more than 1 command
    assert commands.size() > 1
    def prefix = task.ext.prefix ?: "${meta.id}"
    def cmd_size  = commands.size()
    def last_args = task.ext."args$cmd_size" ?: ''
    def extension = last_args.contains("--output-fmt sam") ? "sam" :
                    last_args.contains("--output-fmt bam") ? "bam" :
                    last_args.contains("--output-fmt cram") ? "cram" :
                    "bam"
    assert "$input" != "${prefix}.${extension}" : "Input and output names are the same, use \"task.ext.prefix\" to disambiguate!"

    // Compose pipe
    def cmds = commands.indexed().collect { index, cmd ->
        def command = [
            "samtools $cmd",
            task.ext."args${index!=0 ? index+1 : ''}" ?: ''
        ]
        switch(cmd){
            case !"reheader":
                // The reheader has no useful option
                command << "-@ $task.cpus"
                command << fasta && index == cmd_size ? "--reference ${fasta}" : ''
                command << index != cmd_size ? '-u' : ''
            // samtools commands have slightly different syntax
            case "collate":
                // [-o OUTPUT|-O] [INPUT|-]
                command << index == cmd_size ? " -o ${prefix}.${extension}" : " -O"
                command << index == 0 ? index : '-'
                break
            case ["addreplacerg", "sort", "view"]:
                // [-o OUTPUT] [INPUT|-]
                command << index == cmd_size ? "-o ${prefix}.${extension}" : ""
                command << index == 0 ? index : '-'
                break
            case "reheader":
                // [INPUT|-]
                command << index == 0 ? index : '-'
                break
            case ["fixmate", "markdup"]:
                // [INPUT|-] [OUTPUT|-]
                command << index == 0 ? index : '-'
                command << index == cmd_size ? "${prefix}.${extension}" : "-"
                break
            default:
                assert false: "$cmd is not supported"
        }
        command.join(' ')
    }
    """
    ${cmds.join(' | ')}
```

Since the keys are strings you can quote for dynamic key construction. As that made it necessary to have
an index, it was easier to use an indexed collect. My idea with the if/else of the command wasn't useful after looking closer at what you had implemented, but it seemed very amenable to a switch. 

I'm also starting to like `assert`, as power assertions might be more useful than just responding with an error string.

Anyway, that would be my suggestion. I'm not 100% on it all, but I think it's correct.

muffato: 🤩 That looks great, thank you so much ! There's a lot of Groovy I just didn't know. I can spot a few issues / typos. I'll fix it and push the new version.

How does `assert` bode with Tower ? I think Tower was the reason we switched from `exit 1` to `error`

mahesh-panchal: Ah yeah. Good old off by one error. Just put back the booleans. 

No idea about Seqera Platform and assert.

mahesh-panchal: Suggestion updated

muffato: It needed a few more changes, but it now works !

muffato:
Major rewrite of the module - courtesy of @mahesh-panchal ! It's much nicer to read and easier to maintain.

Also:

- I've checked that CRAM works
- I've fixed the sub

SPPearce:
The structure is looking great, but what name do we actually want? I agree that `pipeline` is overloaded.

mahesh-panchal:
I think to sum it up, what we want is some mechanism to include more efficient modules that allow piping of several commands, or batch processing, etc. This is example is a case where it saves so much disk space and other resources, by having a module such as this. The issue is nf-core modules should be atomic, particularly as it makes workflow composition clearer to newcomers. However, nf-core workflows are also intended to be best practice and in terms of resources, they are really bad at utilizing resources effectively. Some of us think we need to revise this atomicity guideline and have a mechanism for some better practice modules. At least that is my take.

mahesh-panchal:
I can't read properly on the phone ... 
I'm not sure what to call these non-atomic modules. I think the name should reflect the composition, but perhaps following the subworkflow naming scheme works best here.

modules/nf-core/samtools/pipeline/main.nf
```diff
@@ -0,0 +1,106 @@
...
+    // Compose pipe
+    def cmds = commands.indexed().collect { index, cmd ->
+        def first = index == 0
+        def last = index == cmd_size-1
+        def command = [
+            "samtools $cmd",
+            task.ext."args${first ? '' : index+1}" ?: ''
+        ]
+        // First the common options
+        if (cmd != "reheader") {
```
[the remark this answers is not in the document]
muffato: `case !"reheader"` didn't work.
I think `case {it != "reheader"}` would have worked, but skipping `break` on purpose is a bit unusual. That's why I moved it outside of the `switch`

SPPearce:
How does something very explicit, like `samtools multicommand` sound?
I don't think we want a lot of these type of modules, only very common ones like `samtools` (`bedtools` perhaps).

mahesh-panchal:
I think it would be better if we separated these aggregate commands in their own folder, like subworkflows are. It makes them easier to find, and highlights they're  not atomic. Perhaps call the folder `aggregates`. I like the `samtools_multicommand`. I was thinking `samtools_composepipe`, or `samtools_streamcommands`, but I think they sound odd.

muffato:
I think there's too much emphasis on the (non-)atomicity of this. The nf-core guidelines don't prohibit multiple commands running in a module. There is already *a lot* of modules that run multiple commands (the best proxy is to search for mulled containers: they typically indicate that the module needs another tool from another package).

In my mind, I map "modules" to Nextflow `process` and "sub-workflow" to Nextflow `workflow`, and I'm not keen on introducing another term for non-atomic modules, which are still `process`. Or I would argue to move all the existing non-atomic modules over there ;)

As an alternative, what about naming this: `modules/nf-core/custom/multisamtools` ? Still a "module" (because a `process`) but under the `custom` group because it's obviously customised.

mahesh-panchal:
While the guidelines don't prohibit multiple commands the reason is because the modules would be impractical without them. The modules are supposed to be as simple as possible without making them impractical. There may also be examples where this isn't the case, but there several modules that don't fully follow guidelines. 

For findability, I do think this needs to be separated in some way. This kind of process definitely follows what the purpose of subworkflows are supposed to be.

SPPearce:
I think this could slot into the `subworkflows` folder, rather than into a separate folder. I don't expect to be making too many of these chained modules like this.

muffato:
Well, I did mention sub-workflows in my OP 

> * A simple sub-workflow wrapper so that people looking for samtools sub-workflows still find about this. (they may not think about checking modules)
> * Alternatively, only expose this capability as a sub-workflow, either by allowing "local" modules in nf-core sub-workflows and moving the file over there, or by embedding this `process` block in the sub-workflow's `main.nf`.

The first technical question is: can nf-core sub-workflows have "local" modules ? I can't really see room for that in the repository

mahesh-panchal:
@nf-core/core we need a decision where to place this, because this is a necessity.

muffato:
@adamrtalbot . Then maybe #3310 could be reopened because it's exactly what you're advocating for ?

This PR is for a follow-up module that allows enabling/disabling the components of the command, without having to write a new module for every combination.

FYI my PR is currently only for BAM→BAM transformations. `samtools fastq` is not yet supported.

mahesh-panchal:
> We've got a bit dogmatic about "1 tool, 1 module", to the point of losing sight of why that rule exists. In my mind, a module should be an single unit of work, not a piece of software. 

This would be a positive change, but is everyone in agreement (to generally allow multiple commands)?

> > These processes could also help serve to improve developer skills by demonstrating practical examples showing groovy tips and tricks, and better resource usage.
> 
> Nah, they'll just make a local module in their pipeline.

Then I would argue we're not annotating modules well enough and not making them findable. I think to help others, it would be useful if we added more comments to modules explaining anything that might be complicated. 

> To be fair, I quite like this module, think it's really good code and _love_ the flexibility. I'd happy accept this under the golden rule: Minimum Useful Thing™️, but I'm firmly in the camp we should be simplifying and reducing nf-core, not adding complexity.

I think we're in agreement here to make things as simple as possible. I don't think anyone wants to add complexity unless it's a necessity.

muffato:
FYI, there is already a module that runs samtools commands with a pipe: https://github.com/nf-core/modules/blob/master/modules/nf-core/samtools/collatefastq/main.nf
It does `samtools collate | samtools fastq`, which could very well be achieved with the two existing modules SAMTOOLS_COLLATE and SAMTOOLS_FASTQ, but is of course more efficient as a pipe in a single process

mahesh-panchal:
In my opinion that's one of the modules that doesn't follow guidelines and shouldn't be in the `modules` folder ( or more specifically should not have been approved - the placement is still up for discussion). However, generally, this too should have a place along side the process in this PR and the other optimized processes suggested.

mahesh-panchal:
That's only for read pairs, that the input needs to be collated.

mahesh-panchal:
But I also hope none of our guidelines cause people to make broken pipelines. At the end of the day, it is up to the developer to check that their code is correct, hopefully through rigorous testing and peer-review.

matthdsm:
Can anybody add what was decided in the last maintainers meeting? I couldn't make it, but I'm very interested in either this or my `sormadup` [proposal](https://github.com/nf-core/modules/pull/3310).

mahesh-panchal:
See https://github.com/nf-core/website/pull/2327 for a basic summary and start on guidelines. @maxulysse is in charge of this. Basically this PR might not be maintainable by the average nf-core user (most maintainers were not in favor of it), and maintainers would like the script to be explicit with the tool, in particular so argsN maps to a particular tool/subcommand (i.e. like yours). However, they're also worried about maintenance burden, and since one can make local modules, these compound processes should likely go there unless they can be used across multiple pipelines. There's still space to provide more opinions though.

matthdsm:
Thanks @mahesh-panchal!
@maxulysse, In light of those decisions, should I put more work towards the sormadup module? I'm sure it can be used elsewhere, seeing there's an entire subworkflows that does the same, albeit slower and more resource intensive.
I think @priyanka-surana would be interesed too for the the sanger-tol pipelines.

priyanka-surana:
@matthdsm We (@muffato) are already using your implementation as a [local module](https://github.com/sanger-tol/readmapping/blob/main/modules/local/samtools_sormadup.nf). Would be great to have this directly from nf-core.

muffato:
Yes @matthdsm . In sanger-tol we've got sormadup already (almost exactly your version, the only change I did was removing the `sort_memory`), but also soon `collate | fasta` and `view | fastq`

maxulysse:
I'm in favor of resurrecting https://github.com/nf-core/modules/pull/3310, that sounds like a good candidate for a meta module that will be wonders in term of data footprint savings

SPPearce:
I was a dissenting voice on this, I like the flexibility of this "module" and would merge this in.
But yes, modules chaining together multiple tools are acceptable, it was just the complexity of this one that is not currently in favour.

PR #4578, opened 2023-12-12
ncbigenomedownload updated container and added input parameters

a4000:
I updated the container version, because I was getting a bug with the previous version.

I added the taxids as a path input so the user has the option of using a text file with taxids

I added a groups input because that's a mandatory parameter

I also added "-p $task.cpus \\" so that multiple cpus can be used

### DIFF
--- a/modules/nf-core/ncbigenomedownload/main.nf
+++ b/modules/nf-core/ncbigenomedownload/main.nf
@@ -4,12 +4,14 @@ process NCBIGENOMEDOWNLOAD {
 
     conda "${moduleDir}/environment.yml"
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
-        'https://depot.galaxyproject.org/singularity/ncbi-genome-download:0.3.1--pyh5e36f6f_0' :
-        'biocontainers/ncbi-genome-download:0.3.1--pyh5e36f6f_0' }"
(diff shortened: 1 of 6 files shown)


modules/nf-core/ncbigenomedownload/main.nf
```diff
@@ -31,15 +33,19 @@ process NCBIGENOMEDOWNLOAD {
...
+    def taxids_opt     = taxids ? "-t ${taxids}" : ""
     """
     ncbi-genome-download \\
         $args \\
         $accessions_opt \\
+        $taxids_opt \\
         --output-folder ./ \\
-        --flat-output
+        --flat-output \\
+        -p $task.cpus \\
```
mashehu, line 47:
```suggestion
        --parallel $task.cpus \\
```
we try to use the long parameter names to make things more readable

mashehu (approved):

mashehu (commented):
it would also be great if you could do a quick nf-test migration from pytest, see the docs on how to: https://nf-co.re/docs/contributing/modules#migrating-from-pytest-to-nf-test

a4000:
> it would also be great if you could do a quick nf-test migration from pytest, see the docs on how to: https://nf-co.re/docs/contributing/modules#migrating-from-pytest-to-nf-test

I'm merging the pull request for now, but I plan on migrating to nf-test once I understand it better. I'm having trouble following the documentation. I have nf-core 2.10 which I believe is the latest version, but when I run `nf-core modules create <tool>/<subtool> --migrate-pytest`, I get the message "No such option: --migrate-pytest"

PR #4579, opened 2023-12-12
xz/compress: adds stub and stub test, improves tests and meta.yml

leoisl:
This PR adds stub and stub test, improves tests and meta.yml for module xz/compress

### DIFF
--- a/modules/nf-core/xz/compress/main.nf
+++ b/modules/nf-core/xz/compress/main.nf
@@ -21,11 +21,20 @@ process XZ_COMPRESS {
     def args = task.ext.args ?: ''
     archive = raw_file.toString() + ".xz"
     """
-    set -e
-
     # needs --stdout for xz to avoid the following issue:
     # xz: ${raw_file}: Is a symbolic link, skipping
     xz -T $task.cpus --stdout ${args} ${raw_file} > ${archive}
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        xz: \$(xz --version | head -n1 | awk '{print \$NF}')
+    END_VERSIONS
+    """
+
+    stub:
+    archive = raw_file.toString() + ".xz"
+    """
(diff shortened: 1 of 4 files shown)


lrauschning (changes requested):
Looks good! I think testing the stubs is a bit over the top tbh, but it doesn't hurt to have it in any case.
Only thing that strikes me is why you removed the meta map output from meta.yml

leoisl:
> Only thing that strikes me is why you removed the meta map output from meta.yml

Yeah, this was a bit weird to me as well when I first found this in another module, but then realised that the meta map is not actually part of the output, is part of the `archive` tuple emitted by the module. Linting actually complains about this, see for e.g. the linting warning for the first PR for `xz/compress` that we worked on last week: https://github.com/nf-core/modules/actions/runs/7114166474/job/19367662421#step:9:104 . By removing the meta map from the meta output makes linting issues no warnings: https://github.com/nf-core/modules/actions/runs/7178698590/job/19547402154?pr=4579#step:9:45

lrauschning:
I think in nfcore it's standard to include it, even if that causes a warning during linting, though I see your point.
Looking at a few modules randomly, I only saw one that doesn't include it (cat/cat).
So I propose not deleting it – though it might be worth seeing if the linting can be configured to not emit a warning in this case (you may want to raise this point on the slack).

leoisl:
> Looking at a few modules randomly, I only saw one that doesn't include it (cat/cat).
> So I propose not deleting it

I agree, I'd rather follow the standard than make linting happy.

> (you may want to raise this point on the slack).

Just asked this on slack just to be sure before re-updating the other 3 PRs

leoisl:
output.meta added back to meta.yml: https://github.com/nf-core/modules/pull/4579/commits/6217ac515fdf7d3e5740470c63151d5f55335a73

lrauschning (approved):

PR #4582, opened 2023-12-12
Update freebayes to add metas to all channels. also nf-tests migration

TCLamnidis:
Add metas to all input channels of freebayes module.
Also migrate to nf-tests.
Thanks to @sateeshperi for their help with migration to nf-tests.

## PR checklist

- [x] Add a resource `label`


modules/nf-core/freebayes/main.nf
```diff
@@ -9,11 +9,11 @@ process FREEBAYES {
 
     input:
     tuple val(meta), path(input_1), path(input_1_index), path(input_2), path(input_2_index), path(target_bed)
```
mashehu, line 11:
```suggestion
    tuple val(meta1), path(input_1), path(input_1_index), path(input_2), path(input_2_index), path(target_bed)
```
to make it the same like the rest. anyway you can avoid just numbering them, by giving more insightful names?

TCLamnidis: added the suggestion. 
Regarding better names for the metas: I have tried giving information on the meta.yml, but I worry that naming them things like `input_meta, reference_meta`, `samples_meta`, `populations_meta`, `cnv_meta` might be equally confusing. Specifically since "sample" and "population" might be information typically associated with the input channel in many uses 😬

TCLamnidis: actually reverted that. I think it makes sense to keep the first meta as just meta and name the rest, for consistency with modules that only have one meta.
Otherwise, when using this module in a pipeline the dev will need to supply things like `prefix=${meta1.id}` in the `modules.config`, while for most other modules (incl. single meta modules) it would be `prefix=${meta.id}`. 
@mashehu Thoughts? Happy add it back.

mashehu: i would then give others clearer name like you said above. makes it also easier to rememeber which meta belongs to which input file.

TCLamnidis: Done!

modules/nf-core/freebayes/meta.yml
```diff
@@ -22,34 +22,67 @@ input:
...
+        e.g. [ id:'test', single_end:false ]
   - samples:
       type: file
       description: Optional - Limit analysis to samples listed (one per line) in the FILE.
       pattern: "*.txt"
+  - popualtions_meta:
+      type: map
+      description: |
+        Groovy Map containing meta information for the populations file.
+        e.g. [ id:'test', single_end:false ]
```
mashehu, line 76: is the example really a probable use case for the populations file?

modules/nf-core/freebayes/tests/nextflow.config
mashehu, line 1: this file can be removed

mashehu (approved):
One more file which needs to be removed, then it's good to go from my side. Consider also adding a stub mode for the module (will probably be required soon)

modules/nf-core/freebayes/tests/main.nf.test
```diff
@@ -0,0 +1,179 @@
...
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                // { assert snapshot(process.out).match() }, // Output VCF includes a timestamp, so snapshot not consistent past a day.
+                { assert snapshot(file(process.out.vcf.get(0).get(1)).name).match("test.vcf.gz") },
+                { assert path(process.out.vcf.get(0).get(1)).linesGzip.toString().contains('MT192765.1\t10214\t.\tATTTAC\tATTAC\t29.8242') },
+                { assert snapshot(file(process.out.versions.get(0)).name).match("versions.yml") },
```
sateeshperi, line 39:
```suggestion
                { assert snapshot(process.out.versions).match() },
```

why capture only the name `versions.yml`......and not the hash of the file as above?

modules/nf-core/freebayes/tests/main.nf.test.snap
```diff
@@ -0,0 +1,14 @@
+{
+    "test.vcf.gz": {
+        "content": [
+            "test.vcf.gz"
+        ],
+        "timestamp": "2023-12-12T13:33:01.830845"
+    },
+    "versions.yml": {
+        "content": [
+            "versions.yml"
```
sateeshperi, line 10: @mashehu can we lint for presence of md5 for versions in the snap ?

mashehu: Yes, please write an issue

modules/nf-core/freebayes/tests/main.nf.test
TCLamnidis: Fixed versions assertions in all tests. Thanks!

PR #4589, opened 2023-12-13
usable with every configuration

mapo9:
Hi,

the old module was not usable for every possible configuration. 
It contained flags that were not required and made other configurations impossible to run.

The current form 
`
cellranger \
        mkvdjref \
        --genome=$reference_name \
        --fasta=$fasta \
       --genes=$gtf \
        $args
`
contains `--fasta` and `--genes` which are not required.

The current format makes running the other configuration impossible

`
cellranger \
        mkvdjref \
        --genome=$reference_name \
        --seqs=$seqs 
`

which is necesary e.g. when trying to specify the IMGT reference for the `cellranger vdj` command.

So, I would suggest to remove the unrequired ones from the module and rather add them in the respective config file of the pipelines. 

See also [here](https://support.10xgenomics.com/single-cell-vdj/software/pipelines/latest/advanced/references#header) for the tool's description.

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`

### DIFF
--- a/modules/nf-core/cellranger/mkvdjref/environment.yml
+++ b/modules/nf-core/cellranger/mkvdjref/environment.yml
@@ -3,3 +3,5 @@ channels:
   - conda-forge
   - bioconda
   - defaults
+dependencies:
+  - ""
--- a/modules/nf-core/cellranger/mkvdjref/environment.yml
+++ b/modules/nf-core/cellranger/mkvdjref/environment.yml
@@ -5,4 +5,3 @@ channels:
   - defaults
 dependencies:
   - ""
-  
--- a/modules/nf-core/cellranger/mkvdjref/environment.yml
+++ b/modules/nf-core/cellranger/mkvdjref/environment.yml
@@ -5,3 +5,4 @@ channels:
   - defaults
 dependencies:
   - ""
+  
--- a/modules/nf-core/cellranger/mkvdjref/main.nf
+++ b/modules/nf-core/cellranger/mkvdjref/main.nf
@@ -9,7 +9,6 @@ process CELLRANGER_MKVDJREF {
     path gtf            // optional
     path seqs           // optional
     val reference_name
-    
 
     output:
     path "${reference_name}", emit: reference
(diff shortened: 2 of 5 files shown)


klkeys:
this PR ignores an explicit and intentional design choice for the module:

* https://github.com/nf-core/modules/pull/3033#issuecomment-1485685399

using a FASTA + GTF file pair is also an entirely acceptable way to create a VDJ reference:

https://support.10xgenomics.com/single-cell-vdj/software/pipelines/latest/advanced/references#genome

nf-core does not control the IMGT script or resource. that is managed upstream by 10x Genomics. if you want to implement that capacity and are willing to maintain it, then I suggest that you create a separate sister module, call it something like `cellranger/mkvdjref_imgt`, and proceed with the PR there.

(cc @grst)

mapo9:
Ok, I didnt know that the discussion was done beforehand. 

Personally, I dont see how the `fetch-imgt`script is connected to this module as it is sth one has to be run independently before calling `mkvdjref`. 
So, I dont see the maintenance problem here, could you elaborate a little further, so I can imagine better?

Of course using a FASTA + GTF file pair is also an entirely acceptable way to create a VDJ reference, but the idea of my PR is that it's not the only one and people that want to do it differently cannot use the module at all and adding the unnecessary flags in the respective config would be an easy thing to do when using the module, wouldnt it?

grst:
Agree the `fetch-imgt` script is independent of this module. 

> Of course using a FASTA + GTF file pair is also an entirely acceptable way to create a VDJ reference, but the idea of my PR is that it's not the only one and people that want to do it differently cannot use the module at all and adding the unnecessary flags in the respective config would be an easy thing to do when using the module, wouldnt it

But how would you pass the FASTA and GTF to this module without the input channels?

mapo9:
Thanks for your input @grst . I changed the structure of the module so that all the optional input remained as input and checks are implemented to see whether a specific input is empty (`[]`). Based on these checks  either possibility for the `mkvdjref`is run. 
I think that's the most convenient way to account for the different input possibilities.

Again I would argue againts making individual modules for each input but rather implementing this little change to make the one module we already have cover all possibilities.

mapo9:
Great, then I'll merge. Thx for the reviews!

PR #4594, opened 2023-12-13
Update GATK4_HAPLOTYPECALLER module

TCLamnidis:
This corrects the assertion of versions.yml in nf-tests for GATK4_HAPLOTYPECALLER to also check the md5sum of the file (#4409 only checked presence of the `versions.yml`). This was suggeted as the right way to assert the versions files in #4582 
Also, renamed all metas in the module and meta.yml to give them more informative names.

## PR checklist

- [x] Add a resource `label`


modules/nf-core/gatk4/haplotypecaller/main.nf
CarsonJM, line 1: It looks like a lot of these inputs (i.e. input_index, fai, dict, and dbsnp_tbi) are not used. Is there a reason they are included in the input section? If they are removed it could clean it up a bit

TCLamnidis: I am not too familiar with the tool itself, but generally, the index for each file is meant to have the same file name as the input file it is for, with a different suffix. I would guess that GATK HC checks for their presence by file name instead of needing the indices provided with specific paths. So the inputs are probably used implicitly, even if they are not in the command.

maxulysse: they are needed

modules/nf-core/gatk4/haplotypecaller/main.nf
```diff
@@ -9,11 +9,11 @@ process GATK4_HAPLOTYPECALLER {
...
-    tuple val(meta2), path(fasta)
-    tuple val(meta3), path(fai)
-    tuple val(meta4), path(dict)
-    tuple val(meta5), path(dbsnp)
-    tuple val(meta6), path(dbsnp_tbi)
+    tuple val(ref_meta), path(fasta)
+    tuple val(ref_idx_meta), path(fai)
+    tuple val(ref_dict_meta), path(dict)
+    tuple val(dbsnp_meta), path(dbsnp)
+    tuple val(dbsnp_tbi_meta), path(dbsnp_tbi)
```
maxulysse, line 16: Like the idea, but not sure about its implementation, I'd reckon it's something we should talk more about it

TCLamnidis: with things like multi-reference mapping, this kind of approach becomes necessary, to have a way to associate the correct fai, dict, dbsnp to the correct reference. It's something I've implemented in other nf-core modules without issue ( https://github.com/nf-core/modules/pull/3660 ).
Is there a reason not to do it like this?

maxulysse: sorry wasn't clear at all.

I like the `meta5` -> `dbsnp_meta` change, just wondering if `meta_dbsnp` would be better, and if we can agree on a guideline

TCLamnidis: oh! I see. It is following the suggestion of @mashehu here https://github.com/nf-core/modules/pull/4582#discussion_r1424019809
Just trying to keep the changes I end up making across different modules a bit consistent :D I'm all for having a convention on this for sure.

TCLamnidis:
@maxulysse @mashehu Is there a decision on meta naming? I need these changes to add the module in eager. 
I am happy to revert to meta2-meta6 for consistency across modules, or keep informative names like in #4582.

maxulysse:
> @maxulysse @mashehu Is there a decision on meta naming? I need these changes to add the module in eager. I am happy to revert to meta2-meta6 for consistency across modules, or keep informative names like in #4582.

Sorry, forgot to report back after the maintainers meeting last week.
So we agreed to keep meta2-meta6, because we're hoping to get rid of the supplementary meta maps in the future, so there's no point into introducing a new convention for that

TCLamnidis:
ok. thanks! I renamed the metas back to 2-6 :)

jfy133 (approved):

PR #4595, opened 2023-12-13
4374 new module truvari_consistency

kubranarci:
## PR checklist

Closes #XXX 

- [x] Add a resource `label`


modules/nf-core/truvari/consistency/main.nf
```diff
@@ -0,0 +1,51 @@
...
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def vcf3   = vcf3 ?: "$vcf3"
+
+    """
+    truvari \\
+        consistency \\
+        $args \\
+        $vcf1 \\
+        $vcf2  \\
+        $vcf3 > ${prefix}.consistency
```
nvnieuwk, line 32: How is the output extension defined here? I would keep it at `.txt` or `.json` instead of a new unknown extension :)

modules/nf-core/truvari/consistency/main.nf
```diff
@@ -0,0 +1,51 @@
...
+    tag "$meta.id"
+    label 'process_single'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/truvari:4.1.0--pyhdfd78af_0':
+        'biocontainers/truvari:4.1.0--pyhdfd78af_0' }"
+
+    input:
+    tuple val(meta), path(vcf1), path(vcf2), path(vcf3)
```
nvnieuwk, line 12: Can this only be calculated on 2 or 3 VCFs?

nvnieuwk (commented):
Hi some comments :)

modules/nf-core/truvari/consistency/main.nf
kubranarci: it is either json or txt depending on the arguments given, that is why I couldn't decide.

modules/nf-core/truvari/consistency/main.nf
kubranarci: yes, either 2 or 3 vcfs

nvnieuwk: okay then this is fine :)

modules/nf-core/truvari/consistency/main.nf
nvnieuwk: You can parse the `args` to see which one it is. that way you can give the right extension to the output. See here for an example of how we do this in bcftools: https://github.com/nf-core/modules/blob/5074bd37c59454497e790d9210e2f7a876f2c24f/modules/nf-core/bcftools/sort/main.nf#L23-L27

kubranarci: done :)

kubranarci:
> Hi some comments :)

thanks for the review!

modules/nf-core/truvari/consistency/main.nf
```diff
@@ -0,0 +1,54 @@
...
```
nvnieuwk, line 12: Hi I've checked the tool out while reviewing the second time and noticed that it's possible to supply as many VCFs as you can. Can you please change the input to account for this? Something like this should work:
```
tuple val(meta), path(vcfs)
```

Where `vcfs` is a tuple of many VCFs

kubranarci: well I thought it could be nonconventional to use many vcfs but I tried and it seems to be working for at least 4. So I changed it now.

kubranarci: Should we finalize this? @nvnieuwk

nvnieuwk: Yeah I think it's better to have the option and not use it then having to use it, but at the cost of changing the modules again. And sorry I took some days off, I'll review now

modules/nf-core/truvari/consistency/tests/main.nf.test
nvnieuwk, line 1: Can you also add a stub test here?

nvnieuwk (commented):
Looking good! One comment left!

nvnieuwk (approved):
LGTM!

PR #4622, opened 2023-12-15
Create new NARFMAP modules

Emiller88:
I couldn't be bothered to make a mulled container so I made it with auto-wave 😬


nvnieuwk (approved):
LGTM!

nvnieuwk:
> I couldn't be bothered to make a mulled container so I made it with auto-wave 😬

I fully support that :grin:

PR #4645, opened 2023-12-22
Format with black

grst:
## PR checklist

Some files were not formatted with black which makes problems with linting in the pipelines. 

Closes #XXX 

- [ ] Add a resource `label`


Emiller88 (approved):
Great catch!

I don't think this one is going to pass CI.

mashehu:
Double check with ruff format please (we switched to it in the meantime and shouldn't do much different)

modules/nf-core/cellranger/count/templates/cellranger_count.py
```diff
@@ -58,13 +59,19 @@ def chunk_iter(seq, size):
...
+        "${prefix}",
+        "--fastqs",
+        str(fastq_all),
+        "--transcriptome",
+        "${reference.name}",
+        "--localcores",
+        "${task.cpus}",
+        "--localmem",
+        "${task.memory.toGiga()}",
+        *shlex.split("""${args}"""),
```
grst, line 74: @edmundmiller, does ruff not respect the `# fmt: off` comments? 
I added those comments, because it's more readable to have the parameter name and the associated parameter in the same line. 

With black, this worked just fine

mahesh-panchal: Is there a reason you don't format it as
```
"--transcriptome ${reference.name}"
```

modules/nf-core/mygene/templates/mygene.py
```diff
@@ -158,9 +197,9 @@ def query2idmap(self) -> dict:
             species=self.species,
             entrezonly=self.entrezonly,
             ensemblonly=self.ensemblonly,
-            returnall=True
+            returnall=True,
```
mahesh-panchal, line 200: Why does this comma get added here?

modules/nf-core/cellranger/count/templates/cellranger_count.py
mashehu: you need to do it outside the expression: https://docs.astral.sh/ruff/formatter/#format-suppression

modules/nf-core/mygene/templates/mygene.py
grst: Optional commas after the last expression are a fantastic invention:
 * you can comment/uncomment any line and it will still work
 * you can't forget to add a comma when adding an entry.

The goal of black/ruff is also to minimize diffs. With the extra comma, when an item gets added, only one line has changed, while it would be two if you'd need to add the comma first.

modules/nf-core/cellranger/count/templates/cellranger_count.py
grst: > `"--transcriptome ${reference.name}"`

I don't think this would work. In that case it would be treated as a single argument, i.e. the same as if you pass
```
cellranger count "--transcriptome ${reference.name}"
```
in bash.

mashehu:
@nf-core-bot fix linting

PR #4651, opened 2023-12-23
affy/justrma add stub, nf-test and snapshot

sateeshperi:


modules/nf-core/affy/justrma/environment.yml
```diff
@@ -5,3 +5,4 @@ channels:
   - defaults
 dependencies:
   - bioconda::bioconductor-affy=1.78.0
+  - conda-forge::r-base=4.3.1
```
jfy133, line 8: Doesn't that come with a bioconductor package!? Or is it unpinned in conda?

modules/nf-core/affy/justrma/main.nf
```diff
@@ -21,5 +21,14 @@ process AFFY_JUSTRMA {
     task.ext.when == null || task.ext.when
 
     script:
+    prefix = task.ext.prefix ?: "${meta.id}"
     template 'affy_justrma.R'
+
+    stub:
+    """
+    touch test_eset.rds
+    touch test_matrix.tsv
```
jfy133, line 30:
```suggestion
    touch ${prefix}_eset.rds
    touch ${prefix}_matrix.tsv
```

modules/nf-core/affy/justrma/tests/main.nf.test
```diff
@@ -0,0 +1,97 @@
...
+            }
+        }
+    }
+
+    test("homo_sapiens genome affy_array [celfiles]") {
+
+        when {
+            process {
+                """
+                ch_samplesheet = Channel.of([ [ id:'test' ], file(params.test_data['homo_sapiens']['genome']['affy_array_samplesheet'], checkIfExists: true) ])
```
jfy133, line 32: Please use new `test_data_path` notation thingy

jfy133: (as in the rest)

modules/nf-core/affy/justrma/tests/nextflow.config
```diff
@@ -0,0 +1,6 @@
+process {
+    withName: 'test_affy_justrma:AFFY_JUSTRMA' {
+        ext.args = '--sample_name_col name --build_annotation TRUE'
```
jfy133, line 3: Where does this get parsed? I don't see any further reference to it

jfy133 (approved):
Generally OK, a few thigns I don't fuilly understand

modules/nf-core/affy/justrma/environment.yml
SPPearce: Correct, it is unpinned in conda, was using 4.3.3 in the conda recipe.

modules/nf-core/affy/justrma/tests/nextflow.config
SPPearce: It was being read for the second test, but then not actually used as the name of the process had changed from pytest. Actually using it now.

PR #4653, opened 2023-12-26
agrvate stub, nf-test, snap

sateeshperi:


modules/nf-core/agrvate/tests/main.nf.test
```diff
@@ -0,0 +1,62 @@
+nextflow_process {
+
+    name "Test Process AGRVATE"
+    script "../main.nf"
+    process "AGRVATE"
```
Joon-Klaps, line 5: Reason why you make a config but don't import it? 
```suggestion
    script "../main.nf"
    config "./nextflow.config"
    process "AGRVATE"
```

Joon-Klaps (approved):
Looks good, had one question though

PR #4654, opened 2023-12-26
agat/sqstatbasic stub, nf-test, snap

sateeshperi:


modules/nf-core/agat/sqstatbasic/tests/nextflow.config
Joon-Klaps, line 1: Is an empty config necessary to have, was it generated by the template? Or did you forget you made one and forgot to delete?

Joon-Klaps (approved):
One little question, great job!

modules/nf-core/agat/sqstatbasic/tests/nextflow.config
sateeshperi: good catch....forgot to delete

PR #4655, opened 2023-12-26
agat/spstatistics stub, nf-test, snap

sateeshperi:


modules/nf-core/agat/spstatistics/tests/nextflow.config
```diff
@@ -0,0 +1,4 @@
+process {
+
+
+}
```
vagkaratzas, line 4: Maybe remove this file completely since it is not used?

modules/nf-core/agat/spstatistics/tests/tags.yml
```diff
@@ -0,0 +1,2 @@
+agat/spstatistics:
+  - "modules/nf-core/agat/spstatistics/**"
```
vagkaratzas, line 2: Should we remove the double quotes from the second line here? What's the better practice?

vagkaratzas (approved):
All good. Left two minor comments

modules/nf-core/agat/spstatistics/tests/tags.yml
sateeshperi: I think the double quotes are meant to be....pinging @mashehu for confirmation

mashehu: not a strict rule, but would be good if we have the same style for all tags.yml and going by the [template for them](https://github.com/nf-core/tools/blob/master/nf_core/module-template/tests/tags.yml), please use double-quotes in tags.yml files.

PR #4658, opened 2023-12-26
agat/convertspgff2gtf stub & nf-test snapshot

sateeshperi:


toniher (approved):
Thanks for adding stub and migrating to nf-test

PR #4661, opened 2023-12-26
amplify/predict add stub, update test and snap

sateeshperi:


modules/nf-core/amplify/predict/main.nf
```diff
@@ -2,7 +2,6 @@ process AMPLIFY_PREDICT {
     tag "$meta.id"
     label 'process_single'
 
-    // WARN: Version information not provided by tool on CLI. Please update version string below when bumping container versions.
```
SPPearce, line 5: Did the tool gain a version output?

SPPearce (approved):

modules/nf-core/amplify/predict/main.nf
sateeshperi: looks like it....im not seeing the version being set anywhere explicitly.... @louperelo can confirm

SPPearce:
Completed in #5436

PR #4665, opened 2023-12-28
sourmash/compare stub, nf-test, snapshot

sateeshperi:


sateeshperi:
hmmm...I had tests passing locally and conda passes on the CI...strange

modules/nf-core/sourmash/compare/main.nf
```diff
@@ -14,16 +14,17 @@ process SOURMASH_COMPARE {
     val save_csv
 
     output:
-    tuple val(meta), path("*comp.npy.labels.txt"), path("*comp.npy"), optional:true, emit: matrix
-    path "*comp.csv"       , optional:true, emit: csv
-    path "versions.yml"    , emit: versions
+    tuple val(meta), path("*comp.npy")           , optional:true, emit: matrix
+    tuple val(meta), path("*comp.npy.labels.txt"), optional:true, emit: labels
```
sateeshperi, line 18: need to confirm if this will break anything down-stream or it is ok to separate out the outputs this way

modules/nf-core/sourmash/compare/main.nf
```diff
@@ -14,21 +14,22 @@ process SOURMASH_COMPARE {
...
+    path "*comp.csv"                             , optional:true, emit: csv
```
nvnieuwk, line 19: Why no `meta` here?

modules/nf-core/sourmash/compare/main.nf
```diff
@@ -14,21 +14,22 @@ process SOURMASH_COMPARE {
...
+    path "versions.yml"                                         , emit: versions
```
JoseEspinosa, line 20:
```suggestion
    path "*comp.csv"                             , optional:true, emit: csv
    path "versions.yml"                          , emit: versions
```

modules/nf-core/sourmash/compare/main.nf
SPPearce: I have no idea...

JoseEspinosa (approved):
LGTM! Just a small aligning suggestion 😄

modules/nf-core/sourmash/compare/main.nf
SPPearce: Adding now.

nvnieuwk: Thanks!

modules/nf-core/sourmash/compare/main.nf
SPPearce: I've moved the emits first, optional later to make it look nicer in my eyes

PR #4669, opened 2023-12-29
fix fastqc tests n snap

sateeshperi:
migrating existing pytests but, would be gr8 to have more specific assertions for each test


grst (approved):
LGTM

maxulysse (approved):

PR #4675, opened 2024-01-03
Change Space Ranger count module publishDir

fasterius:
## PR checklist

The Space Ranger count module currently publishes its results into `{meta.id}/outs/`, where the `{meta.id}` part is _sort of_ required due to the need for the `--id` flag in Space Ranger itself. This makes it hard for pipelines using this module to modify the publish directory in `modules.config` to _e.g._ `{meta.id}/spaceranger/outs/`, which would better adhere to the `{outdir}/{SOFTWARE/CONTEXT}/[...]` pattern commonly seen elsewhere in nf-core pipelines.

This small PR makes the module publish into `outs/` instead, which is more easily amenable to pipeline-specific changes in `modules.config`. See [this commit at nf-core/spatialtranscriptomics](https://github.com/nf-core/spatialtranscriptomics/commit/70726cd8e4aca8ab147ee4b9b4ceb3c9187b691b) for a temporary implementation while waiting for this PR to be merged for what I was looking for.

I don't know if there were specific reasons to include `{meta.id}` in the publish directory that I'm missing, but if there are any this PR might be better ignored.

(The PR also includes some minor spelling fixes I happened to see as I was editing.)


grst (approved):
LGTM

I faintly remember that I struggled to make this work back then, but probably the 
```
mv ${prefix}/outs outs
```

is the part I was missing. 
If there are no issues with the downstream processes in the pipeline, I'm happy to go with this change.

fasterius:
Okay, great! It's working downstream in the pipeline, I'm just running the tests locally to update the snapshots and then we should be good to go.

Regarding the failing pytest tests, I'm not sure what's going on here. I didn't change any of the final output files, just their locations, and this only for the `count` module, but even `mkref` and `mkgtf` yields different md5sums. Does pytest use the contents of `meta.yml` when calculating the md5sums? Maybe this is obvious to somebody that have used pytest before, but it's my first time 😅 I don't want to update the md5sums before checking.

Regarding the failing linting tests in this PR, should they be ignored? They all fail due to not using the MIT licence and that their respective `environment.yml` file is empty of dependencies. I assume you had these failures before as well?

grst:
I think that is at least in part due to the fact that module tests are being migrated to `nf-test`. 
I'm a bit out-of-the loop though if with any module update it is now mandatory to port the tests. Ideally yes, I suppose.

fasterius:
Okay, so does that mean we can't merge this until we migrate to using `nf-test`? If that's the case then I won't bother updating the checksums and see what's needed for `nf-test`. What about the linting errors?

grst:
> What about the linting errors?

The license needs to be an array, so
```
license:
  - 10x Genomics Eula
```
instead of
```
license: 10x Genomics Eula
```

It complaining about the `environment.yml` I don't know how this is supposed to behave for modules without conda. Maybe best ask on slack.

maxulysse (approved):

maxulysse:
Looking good, but none of the tests are working

fasterius:
Yeah, I didn't bother updating them, as I'm working on a migration to `nf-test` instead (which will be a separate PR, I guess). This PR should then be merged after the migration is complete.

maxulysse:
Oh I see

maxulysse:
ping me if you need anything

PR #4679, opened 2024-01-06
add icdiff to gitpod.yml

sateeshperi:


sateeshperi:
unless @edmundmiller has a better PR with difftastic!

edmundmiller (approved):
Nah let's do it! I tried it out yesterday , but it needs a bit of tweaking.

No reason to hold up progress, it'll just be a swap then anyways!

PR #4682, opened 2024-01-08
use pdiff also on gitpod for nf-test

mashehu:
Same setup as in nf-core/tools and the test.yml.


mashehu:
Will add this to the gitpod docker image instead

PR #4703, opened 2024-01-10
Kraken2/add module

### LINKED ISSUE #2953: new module: kraken2/build

alxndrdiaz:
## PR checklist

Closes #2953 

- [ ] Add a resource `label`


alxndrdiaz:
Files created in output directory: 

```
        └── [4.0K]  test
            └── [4.0K]  library
                └── [4.0K]  added
                    ├── [ 564]  prelim_map_Mkw71vtzg1.txt
                    ├── [ 13K]  SYkwAEUmm7.fna
                    └── [   0]  SYkwAEUmm7.fna.masked
```

jfy133:
I'm not sure about the `chown` error, but for the nf-test error, I seem to get this regularly, and have to run e.g., in your case:

```
sudo rm -rf /home/path/nf-core/modules/.nf-test/tests/cf01ae0d03dee5f3be28629449d283d6
```

And then run the `nf-core module test` command again to get it to work.

alxndrdiaz:
> I'm not sure about the `chown` error, but for the nf-test error, I seem to get this regularly, and have to run e.g., in your case:
> 
> ```
> sudo rm -rf /home/path/nf-core/modules/.nf-test/tests/cf01ae0d03dee5f3be28629449d283d6
> ```
> 
> And then run the `nf-core module test` command again to get it to work.

Thanks, this is helpful to clean the nf-test output. One thing I noticed after this output was cleaner is that `$prefix` is not assigned as the name of the output database as expected: 

```
  Nextflow stdout:                                                                                                                                            │
│                                                                                                                                                               │
│   ERROR ~ Error executing process > 'KRAKEN2_ADD (test)'                                                                                                      │
│                                                                                                                                                               │
│   Caused by:                                                                                                                                                  │
│     Missing output file(s) `null/` expected by process `KRAKEN2_ADD (test)`                                                                                   │
│                                                                                                                                                               │
│   Command executed:                                                                                                                                           │
│                                                                                                                                                               │
│     kraken2-build \                                                                                                                                           │
│         --add-to-library \                                                                                                                                    │
│         proteome.fasta \                                                                                                                                      │
│         --db test \                                                                                                                                           │
│         --threads 2 \                                                                                                                                         │
│                                                                                                                                                               │
│     cat <<-END_VERSIONS > versions.yml                                                                                                                        │
│     "KRAKEN2_ADD":                                                                                                                                            │
│         kraken2: $(echo $(kraken2 --version 2>&1) | sed 's/^.*Kraken version //; s/ .*$//')                                                                   │
│     END_VERSIONS                                                                                                                                              │
│                                                                                                                                                               │
│   Command exit status:                                                                                                                                        │
│     0                                                                                                                                                         │
│                                                                                                                                                               │
│   Command output:                                                                                                                                             │
│     (empty)    
```

modules/nf-core/kraken2/add/main.nf
```diff
@@ -0,0 +1,38 @@
+process KRAKEN2_ADD {
```
jfy133, line 1: ~Is there a reason why this is called `KRAKEN2_ADD` rather than `KRAKEN2_BUILD`? I would argue for the latter - the only difference between add/build etc is the `--add-to-library` which can be supplied either via a prefix, or as it's an essential 'parameter' can be specified as an input channel (e.g. called `mode`?).~

modules/nf-core/kraken2/add/main.nf
```diff
@@ -0,0 +1,38 @@
...
+    tag "$meta.id"
+    label 'process_medium'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/mulled-v2-5799ab18b5fc681e75923b2450abaa969907ec98:87fc08d11968d081f3e8a37131c1f1f6715b6542-0' :
```
jfy133, line 7: What is in this mulled container exactly, when the conda environment is just kraken2? I wonder if this is also possibly causing the issue with the `chown` error message?

modules/nf-core/kraken2/add/main.nf
jfy133: Although looking at the parametmer list maybe this inded makes more sense as after `--build` I would immediately run the `--clean`  flag in same module...

OK so this is fine :sweat_smile: ignore the comment above (leaving for food for thought anyway)

jfy133: Or you could do the `mode` input channel method above, and wrap the `--clean` in a bash `if` statement... but I guess then someone couldn't then use the module for taxonomy downloading (I think?... it's 6:30am here... sorry :sweat_smile: )...

modules/nf-core/kraken2/add/main.nf
```diff
@@ -0,0 +1,38 @@
...
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    kraken2-build \\
+        --add-to-library \\
+        ${fasta} \\
+        --db ${prefix} \\
+        --threads ${task.cpus} \\
+        $args
+    cat <<-END_VERSIONS > versions.yml
```
jfy133, line 33:
```suggestion

    cat <<-END_VERSIONS > versions.yml
```

modules/nf-core/kraken2/add/main.nf
```diff
@@ -0,0 +1,38 @@
...
+        'biocontainers/mulled-v2-5799ab18b5fc681e75923b2450abaa969907ec98:87fc08d11968d081f3e8a37131c1f1f6715b6542-0' }"
+
+    input:
+    tuple val(meta), path(fasta)
+    path taxonomy_names
+    path taxonomy_nodes
+    path accession2taxid
+
+    output:
+    tuple val(meta), path("$prefix/"), emit: db
```
jfy133, line 17:
```suggestion
    tuple val(meta), path("$prefix"), emit: db
```

I seem to remember you can't specify directories with `/` but lets see... this comment suggestion is optional :)

modules/nf-core/kraken2/add/main.nf
```diff
@@ -0,0 +1,38 @@
...
+    path "versions.yml"              , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
```
jfy133, line 25:
```suggestion
    prefix = task.ext.prefix ?: "${meta.id}"
```

that's the source of your error: if you use `def` it means that the variable can only be used within the scope it's defined (in this case `script`). Remove the `def` and you can use it in e.g. the `output:` scope.

modules/nf-core/kraken2/add/meta.yml
```diff
@@ -0,0 +1,54 @@
+name: kraken2_add
+description: Adds fasta files to a Kraken2 taxonomic database
+keywords:
+  - metagenomics
+  - db
+  - classification
```
jfy133, line 6:
```suggestion
  - classification
  - build
```

modules/nf-core/kraken2/add/main.nf
```diff
@@ -0,0 +1,38 @@
...
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/mulled-v2-5799ab18b5fc681e75923b2450abaa969907ec98:87fc08d11968d081f3e8a37131c1f1f6715b6542-0' :
+        'biocontainers/mulled-v2-5799ab18b5fc681e75923b2450abaa969907ec98:87fc08d11968d081f3e8a37131c1f1f6715b6542-0' }"
+
+    input:
+    tuple val(meta), path(fasta)
+    path taxonomy_names
+    path taxonomy_nodes
+    path accession2taxid
```
jfy133, line 14: IF we stick with the standalone `ADD` module - would this still be necessary? I'm assuming yes, but we need to be careful how this works as these would have to be staged into an existing `$prefix/taxonomy` directory, and make sure this doesn't get overwritten on every `add`...

Which also leads to the question, what do we do if we want to add multiple FASTAs in one (so `path(fasta)` is a list)...? The examples in the kraken2 manual suggest using for loops or `find/xargs`  combination

modules/nf-core/kraken2/add/meta.yml
```diff
@@ -0,0 +1,54 @@
...
+input:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. `[ id:'sample1', single_end:false ]`
+  - fasta:
+      type: file
+      description: fasta file that will be added to the database
+      pattern: "*.{fa,fasta,fna,ffn}"
```
jfy133, line 24: Are gzipped files allowed?

modules/nf-core/kraken2/add/tests/main.nf.test
```diff
@@ -0,0 +1,56 @@
...
+
+        when {
+            process {
+                """
+                input[0] = [
+                    [ id:'test' ], // meta map
+                    file(params.test_data['sarscov2']['genome']['proteome_fasta'], checkIfExists: true)
+                    ]
+                input[1] = file(params.test_data['sarscov2']['metagenome']['prot_names_dmp'], checkIfExists: true)
+                input[2] = file(params.test_data['sarscov2']['metagenome']['prot_nodes_dmp'], checkIfExists: true)
```
jfy133, line 39: Does this really work with protein accessions!?

jfy133 (commented):
I have a lot of comments here, but it's more structural/design rather than the quality of the module against the guidelines (which is good) .

I also identified the error and the reason why you couldn't build the snapshot (I thnk)

Can chat about this on slack if it doesn't make sense.

modules/nf-core/kraken2/add/meta.yml
alxndrdiaz: > Are gzipped files allowed?

Not mentioned in the documentation, so I guess is not possible: [Kraken2: custom databases](https://github.com/DerrickWood/kraken2/blob/master/docs/MANUAL.markdown#custom-databases)

modules/nf-core/kraken2/add/tests/main.nf.test
alxndrdiaz: > Does this really work with protein accessions!?

It should work, but it should be specified until the build step using the --protein option:

```
--protein                  Build a protein database for translated search
```

modules/nf-core/kraken2/add/main.nf
alxndrdiaz: > IF we stick with the standalone `ADD` module - would this still be necessary? I'm assuming yes, but we need to be careful how this works as these would have to be staged into an existing `$prefix/taxonomy` directory, and make sure this doesn't get overwritten on every `add`...
> 
> Which also leads to the question, what do we do if we want to add multiple FASTAs in one (so `path(fasta)` is a list)...? The examples in the kraken2 manual suggest using for loops or `find/xargs` combination

That's correct, I forgot to put all those files in the taxonomy directory (db/taxonomy) as expected by kraken2.

modules/nf-core/kraken2/add/main.nf
alxndrdiaz: Tried a different container: 

```
        'https://depot.galaxyproject.org/singularity/kraken2:2.1.2--pl5321h9f5acd7_2' :
        'biocontainers/kraken2:2.1.2--pl5321h9f5acd7_2' }"
```
But the error persists...

alxndrdiaz:
In summary there are three things that should be updated to (possibly) fix the nf-tests: 

1. Put all taxonomy files in the `db/taxonomy` directory usig `stageAs`.
2. Use the `--protein` flag as mentioned in the documentation.
3.  Fix the `.accession2taxid` file which is being converted to directory after using `untar`*

*the file is compressed in `test-datasets`: `prot.accession2taxid.gz`, when running the nf-test this is decompressed using the `untar` module, and it should be a plain text file containing only: 
```
accession.version	taxid
YP_009724389.1	2697049
```
but it is somehow transformed to a directory:

```
Command error:                                                                                                                                                   │
│     cp: omitting directory 'prot.accession2taxid'
```

alxndrdiaz:
> In summary there are three tasks that could be made to (possibly) fix the nf-tests:
> 
>     1. Put all taxonomy files in the `db/taxonomy` directory usig `stageAs`.
> 
>     2. Use the `--protein` flag as mentioned in the documentation.
> 
>     3. Fix the `.accession2taxid` file which is being converted to directory after using `untar`*
> 
> 
> *the file is compressed in `test-datasets`: `prot.accession2taxid.gz`, when running the nf-test this is decompressed using the `untar` module, the result should be a plain text file containing only:
> 
> ```
> accession.version	taxid
> YP_009724389.1	2697049
> ```
> 
> but it is somehow transformed to a directory:
> 
> ```
> Command error:                                                                                                                                                   │
> │     cp: omitting directory 'prot.accession2taxid'
> ```
> 
> ```
> cat /home/nf-core/modules/.nf-test/tests/cf01ae0d03dee5f3be28629449d283d6/work/9c/fdcf7b81546be339cddf8df84b5683/prot.accession2taxid
> cat: /home/nf-core/modules/.nf-test/tests/cf01ae0d03dee5f3be28629449d283d6/work/9c/fdcf7b81546be339cddf8df84b5683/prot.accession2taxid: Is a directory
> ```

1. Done.
2. This flag is only used in the build step (apparently), `kraken2 `will throw an error if used with `add`.
3.  Using `gunzip` module instead of `untar` fixes this error.

`chown` error persists after these changes were merged.

jfy133:
Huh interesting about the CHOWN... it's not coming up in the actions tests being ran?

jfy133:
I'm a bit overloaded atm so can't look further atm :(

alxndrdiaz:
Running the test command with the `--update` flag has a weird behavior , the first test seems to work:
```
╭─────────────────────────────────────────────────────────────────────── nf-test output ───────────────────────────────────────────────────────────────────────╮
│                                                                                                                                                              │
│ 🚀 nf-test 0.8.3                                                                                                                                             │
│ https://code.askimed.com/nf-test                                                                                                                             │
│ (c) 2021 - 2023 Lukas Forer and Sebastian Schoenherr                                                                                                         │
│                                                                                                                                                              │
│ Found 1 files in test directory.                                                                                                                             │
│ Warning: every snapshot that fails during this test run is re-record.                                                                                        │
│                                                                                                                                                              │
│ Test Process KRAKEN2_ADD                                                                                                                                     │
│                                                                                                                                                              │
│   Test [cf01ae0d] 'sarscov2 protein_db' PASSED (7.552s)                                                                                                      │
│   Snapshots:                                                                                                                                                 │
│     1 updated [sarscov2 protein_db]                                                                                                                          │
│                                                                                                                                                              │
│                                                                                                                                                              │
│ Snapshot Summary:                                                                                                                                            │
│   1 updated                                                                                                                                                  │
│                                                                                                                                                              │
│ SUCCESS: Executed 1 tests in 7.553s                                                                                                                          │
│                                                                                                                                                              │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

But then the second tests fails: 

```
INFO     Generating nf-test snapshot again to check stability                                                                                                   
╭─────────────────────────────────────────────────────────────────────── nf-test output ───────────────────────────────────────────────────────────────────────╮
│                                                                                                                                                              │
│ 🚀 nf-test 0.8.3                                                                                                                                             │
│ https://code.askimed.com/nf-test                                                                                                                             │
│ (c) 2021 - 2023 Lukas Forer and Sebastian Schoenherr                                                                                                         │
│                                                                                                                                                              │
│ Found 1 files in test directory.                                                                                                                             │
│ Error: java.io.IOException: Launch Directory '/home/nf-core/modules/.nf-test/tests/cf01ae0d03dee5f3be28629449d283d6' could not be deleted or created: │
│ java.nio.file.AccessDeniedException:                                                                                                                         │
│ /home/nf-core/modules/.nf-test/tests/cf01ae0d03dee5f3be28629449d283d6/work/b2/91774fbbba57b0795577eea6b0ec3e/test/library/added/QH3Hp6nib5.fna        │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
CRITICAL Ran, but found errors:                                                                                                                                 
          - nf-test failed                                                                                                                                      

```

jfy133:
OK my suspicion is the container has the 'wrong' version of `chown`, which Kraken2 (undocumented) needs..
So I think we maybe need to try building a mulled container but with a compatible version of chown
This is based on

```
$ bash .command.run 
Masking low-complexity regions of new file... done.
Added "proteome.fasta" to library (db)
chown: unrecognized option '--from'
BusyBox v1.32.1 (2021-04-13 11:15:36 UTC) multi-call binary.

Usage: chown [-RhLHPcvf]... USER[:[GRP]] FILE...

Change the owner and/or group of each FILE to USER and/or GRP

        -R      Recurse
        -h      Affect symlinks instead of symlink targets
        -L      Traverse all symlinks to directories
        -H      Traverse symlinks on command line only
        -P      Don't traverse symlinks (default)
        -c      List changed files
        -v      List all files
        -f      Hide errors
```
Which then presumably results in the 'can't access file' jav aexception

jfy133:
I will try requesting a coreutils + kraken2 multi-package container and see what happens...

jfy133:
And fundamnetally I think its' something do with `shutil` used in teh kraken2 add python script... given the massive warning at the top...

https://docs.python.org/3/library/shutil.html

jfy133:
But hopefuly the mulled container fixes the issue!

jfy133:
similar issues here: https://github.com/common-workflow-language/cwltool/issues/1913

jfy133:
@alxndrdiaz can you try with this container?

quay.io/biocontainers/mulled-v2-4e263fb896be122f4594792ae66a3020b4f9059f:2e0b144854b4a3d69b5df7a0340a60db846cc8bf-0

jfy133:
OK, no something appears to be wrong with the mulled container, `chown` is still the original...

alxndrdiaz:
### Test 1: run kraken2 (2.1.2) locally using the test data as follows:

```
tree -ash protdb
[4.0K]  protdb
└── [4.0K]  taxonomy
    ├── [  48]  prot.accession2taxid
    ├── [  82]  prot_names.dmp
    └── [1.0K]  prot_nodes.dmp
``` 

```
#!/bin/bash
dbname=protdb
fasta=proteome.fasta
kraken2-build --add-to-library $fasta --db $dbname --threads 2
```
Output (no errors are raised): 

```
Masking low-complexity regions of new file... done.
Added "proteome.fasta" to library (protdb)
```

File permissions

```
ls -rl protdb/library/added/
total 24
-rw------- 1 user user   564 feb 13 22:26 prelim_map_QkFQKmeB2i.txt
-rw-rw-r-- 1 user user     0 feb 13 22:26 6cK2jYAkM1.fna.masked
-rw-rw-r-- 1 user user 17975 feb 13 22:26 6cK2jYAkM1.fna
```

### Test 2: with the same test data run kraken2 using the Docker image: 

```
#!/bin/bash

# docker image
img='quay.io/biocontainers/kraken2:2.1.2--pl5321h9f5acd7_2'
docker pull $img
# input
dbname=protdb
fasta=proteome.fasta
# run kraken2 
docker run \
   --rm \
   -v $(pwd):$(pwd) \
   -w $(pwd) \
   -t ${img} \
   kraken2-build --add-to-library $fasta --db $dbname --threads 2
```

Output is also OK and no errors are raised: 

```
Masking low-complexity regions of new file... done.
Added "proteome.fasta" to library (protdb)
```

The file permissions change using the Docker image:

```
ls -rl protdb/library/added/
total 20
-rw-r--r-- 1 root root     0 feb 13 22:32 SvaT8uDmc3.fna.masked
-rw-r--r-- 1 root root 13383 feb 13 22:32 SvaT8uDmc3.fna
-rw------- 1 root root   564 feb 13 22:32 prelim_map_70VAV5EhNl.txt
```
If this happens when running the module tests it is probably messing with any read file operations.

jfy133:
So you mean it works manually in the docker image with the original?!

alxndrdiaz:
> So you mean it works manually in the docker image with the original?!

Yes. The only problem is that the permissions of the output files are changed to `root` when using the Docker image (still don't know why), it seems that this prevents any further file operations. Also running the test with: 

```
nf-core modules test kraken2/add --update 
```
Generates the output files but the permissions are changed as well: 

```
# test directory
ldir="/home/nf-core/modules/.nf-test/tests/cf01ae0d03dee5f3be28629449d283d6/work/5a/3c65cbd2f4e9e57e0cde9a618d9e48" 
# check permissions of output files
ls -l $ldir/test/library/added
total 20
-rw-r--r-- 1 root root 13383 feb 14 08:59 mg7bS0iePD.fna
-rw-r--r-- 1 root root     0 feb 14 08:59 mg7bS0iePD.fna.masked
-rw------- 1 root root   564 feb 14 08:59 prelim_map_Wokcrk78XD.txt
```

jfy133:
I built a mulled container that works with the module :tada: 

I just need to request the official one :), and then you can go about fighting nf-test (the output isn't stable :sweat_smile: )

```
 java.lang.RuntimeException: Different Snapshot:                                                                                                                              │
│ Found:                                                                                                                                                                       │
│ [                                                                                                                                                                            │
│     [                                                                                                                                                                        │
│         [                                                                                                                                                                    │
│             {                                                                                                                                                                │
│                 "id": "test"                                                                                                                                                 │
│             },                                                                                                                                                               │
│             [                                                                                                                                                                │
│                 [                                                                                                                                                            │
│                     [                                                                                                                                                        │
│                         "prelim_map_9AtSQCgKaG.txt:md5,229620e9ad30fc411f025597f4e6379f",                                                                                    │
│                         "qIS93ZUMlY.fna:md5,befde39f5f4d5cebb25e57328830f321",                                                                                               │
│                         "qIS93ZUMlY.fna.masked:md5,d41d8cd98f00b204e9800998ecf8427e"                                                                                         │
│                     ]                                                                                                                                                        │
│                 ],                                                                                                                                                           │
│                 [                                                                                                                                                            │
│                     "prot.accession2taxid:md5,c0f96ba5dbb00150b4b805ba6dab7bea",                                                                                             │
│                     "prot_names.dmp:md5,130f9132095562e09c732679c562f5e9",                                                                                                   │
│                     "prot_nodes.dmp:md5,c471c27a4ce85ae74d2c63633c9ce1e3"                                                                                                    │
│                 ]                                                                                                                                                            │
│             ]                                                                                                                                                                │
│         ]                                                                                                                                                                    │
│     ],                                                                                                                                                                       │
│     [                                                                                                                                                                        │
│         "versions.yml:md5,9a7b40921622d8c873fb3bfd8bac4c3d"                                                                                                                  │
│     ]                                                                                                                                                                        │
│ ]                                                                                                                                                                            │
│                                                                                                                                                                              │
│ Expected:                                                                                                                                                                    │
│ [                                                                                                                                                                            │
│     [                                                                                                                                                                        │
│         [                                                                                                                                                                    │
│             {                                                                                                                                                                │
│                 "id": "test"                                                                                                                                                 │
│             },                                                                                                                                                               │
│             [                                                                                                                                                                │
│                 [                                                                                                                                                            │
│                     [                                                                                                                                                        │
│                         "aABaLor5yN.fna:md5,befde39f5f4d5cebb25e57328830f321",                                                                                               │
│                         "aABaLor5yN.fna.masked:md5,d41d8cd98f00b204e9800998ecf8427e",                                                                                        │
│                         "prelim_map_AE_Kctf91Q.txt:md5,229620e9ad30fc411f025597f4e6379f"                                                                                     │
│                     ]                                                                                                                                                        │
│                 ],                                                                                                                                                           │
│                 [                                                                                                                                                            │
│                     "prot.accession2taxid:md5,c0f96ba5dbb00150b4b805ba6dab7bea",                                                                                             │
│                     "prot_names.dmp:md5,130f9132095562e09c732679c562f5e9",                                                                                                   │
│                     "prot_nodes.dmp:md5,c471c27a4ce85ae74d2c63633c9ce1e3"                                                                                                    │
│                 ]                                                                                                                                                            │
│             ]                                                                                                                                                                │
│         ]                                                                                                                                                                    │
│     ],                                                                                                                                                                       │
│     [                                                                                                                                                                        │
│         "versions.yml:md5,9a7b40921622d8c873fb3bfd8bac4c3d"                                                                                                                  │
│     ]                                                                                                                                                                        │
│ ]   
```

jfy133:
@alxndrdiaz there you go! Singularity will come tomorrow I guess.

in the meantime (if you have time) can you:

1. Add a bash for loop, so it runs through all possible fastas in the fasta input channel
2. Complete the nf-test assertions, but note it is unstable (there are files made with random strings... might have to ignore those...)

alxndrdiaz:
> @alxndrdiaz there you go! Singularity will come tomorrow I guess.
> 
> in the meantime (if you have time) can you:
> 
>     1. Add a bash for loop, so it runs through all possible fastas in the fasta input channel
> 
>     2. Complete the nf-test assertions, but note it is unstable (there are files made with random strings... might have to ignore those...)

@jfy133 [2]  One (dummy) solution is to ignore the output directory (this makes the Docker test to work) in the `main.nf.test` file : 

```
  assertAll(
                { assert process.success }
            )
```
[1] Tried to use to pass multiple FASTA files but still need to find out how to pass a `path() ` within the `main.nf.test`  file instead of `file()`, alsoaneed to specify the file extensions (`fastaext`) to allow the command to know which files to pass, but I am still trying to find out how to do this with nf-test : 

```
 find -L $ {fastas} -name ${fastaext} -print0 | \\
    xargs -0 -I{} -n1 kraken2-build --add-to-library {}  --db ${prefix} --threads $task.cpus $args
```

alxndrdiaz:
> > @alxndrdiaz there you go! Singularity will come tomorrow I guess.
> > in the meantime (if you have time) can you:
> > ```
> > 1. Add a bash for loop, so it runs through all possible fastas in the fasta input channel
> > 
> > 2. Complete the nf-test assertions, but note it is unstable (there are files made with random strings... might have to ignore those...)
> > ```
> 
> @jfy133 [2] One (dummy) solution is to ignore the output directory (this makes the Docker test to work) in the `main.nf.test` file :
> 
> ```
>   assertAll(
>                 { assert process.success }
>             )
> ```
> 
> [1] Tried to use to pass multiple FASTA files but still need to find out how to pass a `path() ` within the `main.nf.test` file instead of `file()`, also need to specify the file extensions (`fastaext`) to allow the command to know which files to pass, but I am still trying to find out how to do this with nf-test :
> 
> ```
>  find -L $ {fastas} -name ${fastaext} -print0 | \\
>     xargs -0 -I{} -n1 kraken2-build --add-to-library {}  --db ${prefix} --threads $task.cpus $args
> ```

jfy133:
I should have got the (1) parallle adding now; so I think we can go back to making (2) properly :)

alxndrdiaz:
> I should have got the (1) parallle adding now; so I think we can go back to making (2) properly :)

These assertions only check [1] if the process is run successfully, [2] an output test directory is created, [3] versions file.

```
  then {
            assertAll(
                { assert process.success },
                { assert process.out.db.get(0).get(1) ==~ ".*/test" },
                { assert snapshot(
                        process.out.versions
                    ).match() }
            )
        }
```

Trying to use  `assert process.out.db.get(0).get(1).get(0)` does not work (probably because the channel is a directory and not a single file), the following error is raised in this case: 

```
 groovy.lang.MissingMethodException: No signature of method: java.lang.String.get() is applicable for argument types: (Integer) values: [0]                         │
│ Possible solutions: getAt(int), getAt(int), grep(), next(), getAt(groovy.lang.Range), getAt(groovy.lang.Range) 
```

jfy133:
I tweaked the tests slightly to make them more explicit, but I think we are there!!!

modules/nf-core/kraken2/add/meta.yml
```diff
@@ -0,0 +1,55 @@
+name: kraken2_add
+description: Adds fasta files to a Kraken2 taxonomic database
+keywords:
+  - metagenomics
+  - db
+  - classification
+  - build
```
TCLamnidis, line 7:
```suggestion
  - build
  - kraken2
  - add
```

modules/nf-core/kraken2/add/meta.yml
```diff
@@ -0,0 +1,55 @@
...
+        Groovy Map containing sample information
+        e.g. `[ id:'sample1', single_end:false ]`
+  - versions:
+      type: file
+      description: File containing software versions
+      pattern: "versions.yml"
+  - db:
+      type: directory
+      description: contains required files to build the database
+      pattern: "*/"
```
TCLamnidis, line 51:
```suggestion
  - db:
      type: directory
      description: contains required files to build the database
      pattern: "*/"
  - versions:
      type: file
      description: File containing software versions
      pattern: "versions.yml"
```
Rearranged since the meta and db are part of the same output channel

modules/nf-core/kraken2/add/tests/main.nf.test
```diff
@@ -0,0 +1,92 @@
+nextflow_process {
+
+    name "Test Process KRAKEN2_ADD"
+    script "../main.nf"
+    process "KRAKEN2_ADD"
+    tag "kraken2"
+    tag "kraken2/add"
+   tag "gunzip"
+    tag "modules"
```
TCLamnidis, line 9:
```suggestion
    tag "kraken2/add"
    tag "gunzip"
    tag "modules"
```

PR #4702, opened 2024-01-09
Update modules required for fetchngs pipeline

drpatelh:
This PR forms part of updates required to transition the way we use test data on nf-core as discussed in Slack [here](https://nfcore.slack.com/archives/CJRH30T6V/p1704755682903089).

The corresponding updates at the pipeline-level can be seen in [this commit](https://github.com/nf-core/fetchngs/pull/245/commits/f9eb52509483a646bb53ef7e39b95cfdcc138b25).

PR #4706, opened 2024-01-10
Minor fixes to bases2fastq module

adamrtalbot:
Changes by @drpatelh from PR #4702 

Minor fixes to syntax of bases2fastq module.

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`


maxulysse (approved):

maxulysse (approved):

maxulysse (approved):

pontus:
Seems not sufficient :(

I don't see where the image comes from, other than being in nf-core, do you know its repo?

pontus:
I'm fairly certain that's not the way (except for being good to verify that the problem is the no-home mount). (It might get tests to pass but then users will have issues later on).

pontus:
Looking at it some more, it appears the binary is licensed with the terms at https://www.elementbiosciences.com/legal/bases2fastq-software-end-user-license-agreement?utm_source=elembio&utm_medium=custom-url&utm_content=bases2fastq-software-end-user-license-agreement&utm_campaign=legal, which don't look very free to me, so I'm leaning towards this module should be removed/reworked into using software that can legally be used.

maxulysse:
> I believe it's required for demultiplexing sequencing from Element biosciences, who offered support for nf-core so would rather keep this in. The key question is why has it started to fail now? did it ever work?

@edmundmiller any idea?

pontus:
Ping @blajoie @ajaltomare, is this your tool we're working with? Is it trying to use the user's home directory some way and can it be controlled. It seems the module has been broken at some point and I think we're still working on the theory that it's due to newer nextflow defaulting to asking singularity not to bind the user home directory, meaning it's not writable.

@adamrtalbot Dropping tests doesn't help but is rather counterproductive if usage later on is broken.

Also, Element biosciences may very helpfully be offering this, but my impression from a very quick look there is no paper trail for that, possibly exposing end-users to risk we shouldn't. If they want to offer it, maybe it can be made clear.

I'm also unsure why the container image is kept in the nf-core quay rather than using the images offered directly (at https://hub.docker.com/r/elembio/bases2fastq).

pontus:
I think this is only used by demultiplex, so maybe not huge usage, but I still view it as dropping the test is the wrong thing to do and give pain to end users needlessly (and if we're not concerned about usage, we could just remove it :)

The license at that page might have changed since, but without trail and documentation of an earlier, more liberal license, I still think we expose end-users to legal risks, again needlessly. (This isn't related to if the image is pulled from quay nf-core or Element biosciences dockerhub.)

maxulysse (approved):

PR #4725, opened 2024-01-11
Switch ci to pre-commit

mashehu:


.github/workflows/fix-linting.yml
```diff
@@ -14,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       # Use the @nf-core-bot token to check out so we can push later
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
```
mirpedrol, line 16: do we want to ping the commit instead of the major version?

mirpedrol (approved):

.github/workflows/fix-linting.yml
mashehu: didn't like it at first either, but there are valid reasons for it: https://docs.renovatebot.com/upgrade-best-practices/#extends-helperspingithubactiondigests and it's anyway handled by renovate, so we don't need to think to much about the versioning, imo

mashehu:
probably better to restart this another time

PR #4738, opened 2024-01-14
Remove modules/ suffix from params.testdata_base_path

drpatelh:
Inspired by https://github.com/nf-core/fetchngs/pull/251 so we can use the `params.testdata_base_path` for both module and pipeline test data if required


maxulysse (approved):
Makes sense

PR #4754, opened 2024-01-17
Add compression to MSA modules

lrauschning:
Reopened, this time without duplicating commits.
This PR introduces output compression to the FAMSA, MTMalign, MAFFT, CLUSTALO and MUSCLE5 (EDIT: and TCOFFEE) modules.
This is required at 7. in https://nf-co.re/docs/contributing/modules#general
Behaviour of the modules is not otherwise changed.


lrauschning:
Tested locally on conda, let's see if singularity also works.

modules/nf-core/tcoffee/alncompare/main.nf
```diff
@@ -27,8 +27,8 @@ process TCOFFEE_ALNCOMPARE {
     """
     export TEMP='./'
     t_coffee -other_pg aln_compare \
-        -al1 ${ref_msa} \
-        -al2 ${msa} \
+        -al1 <(unpigz -cdf ${ref_msa})\
```
JoseEspinosa, line 30: Maybe you should check whether the file is compressed or not before to avoid unnecessary processing

modules/nf-core/tcoffee/irmsd/main.nf
```diff
@@ -24,14 +24,16 @@ process TCOFFEE_IRMSD {
     """
     export TEMP='./'
 
+    unpigz -cdf $msa > msafile &
```
JoseEspinosa, line 27: Here as well something like [this](https://github.com/nf-core/modules/blob/8fc1d24c710ebe1d5de0f2447ec9439fd3d9d66a/modules/nf-core/abritamr/run/main.nf#L30-L32) would be better in my opinion

lrauschning:
For MTMalign, the file is written first to disk, and then compressed on disk, as I wasn't able to cleanly isolate MTMalign's output (it has a hardcoded outfile name for the FASTA output, and pollutes stdout with debug messages).
I think it still makes sense to implement it, so that the modules have a (somewhat) standard interface.

modules/nf-core/tcoffee/alncompare/main.nf
lrauschning: yes, good point. pigz does autodetect this, but the detection might add some overhead compared to uncompressing based on the file extension.

modules/nf-core/clustalo/align/main.nf
```diff
@@ -4,45 +4,54 @@ process CLUSTALO_ALIGN {
...
 
     script:
     def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
+    def write_output = compress ? "--force -o >(pigz -cp ${task.cpus} > ${prefix}.aln.gz)" : "> ${prefix}.aln"
+    // using >() is necessary to preserve the return value,
+    // so nextflow knows to display an error when it failed
+    // the --force -o is necessary, as clustalo expands the commandline input,
+    // causing it to treat the pipe as a parameter and fail
+    // this way, the command expands to /dev/fd/<id>, and --force allows writing output to an already existing file
```
adamrtalbot, line 30: I'm not a huge fan of this but I guess most pipeline developers will leave it on `true` and forget about it, so why not?

modules/nf-core/clustalo/align/main.nf
```diff
@@ -4,45 +4,54 @@ process CLUSTALO_ALIGN {
...
+        pigz: \$(echo \$(pigz --version 2>&1) | sed 's/^.*pigz\\w*//' ))
     END_VERSIONS
     """
 
     stub:
     def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
     """
-    touch ${prefix}.aln
+    touch ${prefix}.aln.gz
```
adamrtalbot, line 49: This should change based on the `compress` value. Something like this:
```
    stub:
    def output = compress ? "${prefix}.aln.gz" : "${prefix}.aln"
    """
    touch ${output}
```

modules/nf-core/famsa/align/main.nf
```diff
@@ -10,26 +10,29 @@ process FAMSA_ALIGN {
...
     def prefix = task.ext.prefix ?: "${meta.id}"
     def options_tree = tree ? "-gt import $tree" : ""
     """
     famsa $options_tree \\
+        $compress_args \\
         $args \\
         -t ${task.cpus} \\
         ${fasta} \\
-        ${prefix}.aln
+        ${prefix}.aln${compress ? '.gz':''}
```
adamrtalbot, line 35: Hmm I'm coming around to this idea a bit more, it seems to be cleaner and harder to mess up.

adamrtalbot (commented):
You need to update the stubs. You could add tests to check them as well.

modules/nf-core/clustalo/align/main.nf
lrauschning: Ah good catch, didn't change that since introducing the compress input channel. Might also affect some of the other modules, I'll have a look.

modules/nf-core/famsa/align/main.nf
lrauschning: Yes, it being the most clean and straightforward to understand/document is I think the main advantage.
Especially for tools like FAMSA which natively support compression its also cleaner than the output format changing based on a parameter passed via ext.args.

modules/nf-core/mafft/main.nf
```diff
@@ -4,32 +4,35 @@ process MAFFT {
...
-    def add          = add             ? "--add ${add}"                   : ''
-    def addfragments = addfragments    ? "--addfragments ${addfragments}" : ''
-    def addfull      = addfull         ? "--addfull ${addfull}"           : ''
-    def addprofile   = addprofile      ? "--addprofile ${addprofile}"     : ''
-    def addlong      = addlong         ? "--addlong ${addlong}"           : ''
+    def add          = add             ? "--add <(unpigz -cdf ${add})"                   : ''
+    def addfragments = addfragments    ? "--addfragments <(unpigz -cdf ${addfragments})" : ''
+    def addfull      = addfull         ? "--addfull <(unpigz -cdf ${addfull})"           : ''
+    def addprofile   = addprofile      ? "--addprofile <(unpigz -cdf ${addprofile})"     : ''
+    def addlong      = addlong         ? "--addlong <(unpigz -cdf ${addlong})"           : ''
```
adamrtalbot, line 33: Do we need pigz here? I'm just thinking gzip would remove a dependency.

adamrtalbot: If there's a massive speed benefit then it's worth it, but sometimes I've found pigz to be about the same speed as gzip in many situations.

lrauschning: Ah sorry, missed this on Friday.
I used pigz here to maintain consistency with the output file writing, and as it's already a dependency for processing the output. If gzip is replaced by pigz in the base container, this might become relevant as an implementation using gzip would then require an additional dependency.
Decompression in pigz is by default not parallelized (except for separating the threads for reading/writing), so I wouldn't expect this to have any significant effect on performance.

adamrtalbot: OK no worries.

adamrtalbot (approved):

PR #4760, opened 2024-01-17
CHORES: update fasqc tests with new data organisation

maxulysse:


maxulysse:
can you add an extra entry in the meta.yml to fix linting?

maxulysse:
> can you add an extra entry in the meta.yml to fix linting?

yeah, that comment was meant for another PR...
I usually don't ask @maxulysse to do much...

modules/nf-core/fastp/tests/main.nf.test
```diff
@@ -19,11 +19,10 @@ nextflow_process {
                 save_trimmed_fail = false
                 save_merged       = false
 
-                input[0] = [
+                input[0] = Channel.of([
```
grst, line 22: Is `Channel.of` necessary? 
I usually wouldn't care, but since this will be propagated to a lot of modules, I'd try to keep it as minimal as possible.

modules/nf-core/fastp/tests/main.nf.test
```diff
@@ -266,10 +262,10 @@ nextflow_process {
                 save_trimmed_fail = false
                 save_merged       = false
 
-                input[0] = [ [ id:'test', single_end:true ], // meta map
-                            [ file(params.test_data['sarscov2']['illumina']['test_interleaved_fastq_gz'], checkIfExists: true) ]
-                        ]
-
+                input[0] = Channel.of([
+                    [ id:'test', single_end:true ], // meta map
+                    [ file(params.test_data['sarscov2']['illumina']['test_interleaved_fastq_gz'], checkIfExists: true) ]
```
grst, line 267: why does this stil use the old system?

modules/nf-core/fastp/tests/main.nf.test
```diff
@@ -336,10 +332,10 @@ nextflow_process {
...
```
grst, line 337: same here

modules/nf-core/fastp/tests/main.nf.test
maxulysse: No, but I'd rather be explicit

modules/nf-core/fastp/tests/main.nf.test
maxulysse: :facepalm:

adamrtalbot (approved):

modules/nf-core/fastp/tests/main.nf.test
adamrtalbot: I quite like using `Channel.of` here, it's being explicit and using the 'proper Nextflow'.

PR #4769, opened 2024-01-18
add optional bam index to outputs for sambamba markdup

DavyCats:
## PR checklist

The current sambamba markdup task does not specify the bam index as an output. Many downstream tools will need this file as an input, so it's useful to add here.


BioInf2305:
sambamba markdup does not produce the index file by default. So how come this optional output is appropriate? The user might need to run sambamba index or samtools index separately to generate the index file (?).

DavyCats:
> sambamba markdup does not produce the index file by default. So how come this optional output is appropriate?

Because it's an *optional* output. Isn't the entire premise behind an optional output that it may or may not be produced depending on the circumstances? Sambamba produces a BAM index any time it writes a coordinate sorted BAM file (as mentioned by the sambamba author here: https://github.com/biod/sambamba/issues/298) and markdup will write a coordinate sorted BAM file if the input is coordinate sorted. So, in many cases it will be produced, even within the nf-core ecosystem. After all, the nf-core BWA MEM process specifically has an option to output coordinate sorted BAM files.

I can understand that one might not want to support every obscure output file that a program might produce in some minute fraction of use-cases. However, the BAM index will be required by basically every down stream tool. It, therefore, seems like it would be beneficial to support it. That way people won't have to call an additional process to create a file which in many cases would already have been produced by this one.

BioInf2305:
You are absolutely right about the optional output tag but I did not know that "sambamba markdup" sometimes also generates the index file; thank you for bringing this to my attention.

BioInf2305:
I think you should put the "optional" directive after "emit" (as mentioned here: https://github.com/nextflow-io/nextflow/discussions/1980).

DavyCats:
Thanks for the tip! They're switched around now.

DavyCats:
I don't really understand why the singularity test is failing. Locally the test is succeeding. It looks like the sambamba markdup process is failing due to a read-only file system:

The `.command.err` from the test artifacts:
```
INFO:    Environment variable SINGULARITYENV_TMPDIR is set, but APPTAINERENV_TMPDIR is preferred
INFO:    Environment variable SINGULARITYENV_NXF_TASK_WORKDIR is set, but APPTAINERENV_NXF_TASK_WORKDIR is preferred

sambamba 1.0.0
 by Artem Tarasov and Pjotr Prins (C) 2012-2022
    LDC 1.28.1 / DMD v2.098.1 / LLVM12.0.0 / bootstrap LDC - the LLVM D compiler (1.28.1)

sambamba-markdup: /home/runner/sambamba-pid17-markdup-qegd: Read-only file system

```

SPPearce (approved):

PR #4787, opened 2024-01-23
umicollapse module: Set initial heap size and modificaitons to log file creation

MatthiasZepper:
This PR suggests two changes to the upstream nf-core module that I had to introduce to make it work [within an evaluation version the rnaseq pipeline using this module](https://github.com/MatthiasZepper/rnaseq/tree/eval_umicollapse).

The first modification adds a `-Xmx` parameter to the command that specifies the maximum heap size of the virtual machine. The parameter is used exactly as it is found in several other nf-core modules wrapping Java tools.

The second modification writes a logfile from stdout, since moving the `.command.log` doesn't seem to be a valid approach. The file seemingly does not yet exist in the work directory when the command is about to be executed.

For me, it was fine to not set an explicit stack size with `-Xss`, but @grst [apparently had to do it for the smrnaseq pipeline](https://github.com/nf-core/smrnaseq/blob/d5192d48a3491504b41fe1af51860d1d9773c632/modules/nf-core/umicollapse/main.nf), so maybe it would be helpful to add this as well?

The additional label `process_high` was added because module lint complained about the lack of standard process labels. If `process_high_memory` is declared after `process_high` in the config, i[t will take precedence over the memory setting](https://github.com/MatthiasZepper/Nextflow_withLabel_prio). 

## PR checklist

- [ ] Add a resource `label`

### DIFF
--- a/modules/nf-core/umicollapse/main.nf
+++ b/modules/nf-core/umicollapse/main.nf
@@ -26,13 +26,10 @@ process UMICOLLAPSE {
     """
     umicollapse \\
         bam \\
-        -Xms${String.format("%.0f", Math.rint(0.2 * task.memory.toGiga()))}g  \\
         -Xmx${task.memory.toGiga()}g  \\
         -i $bam \\
         -o ${prefix}.bam \\
-        $args
-
-    mv .command.log ${prefix}_UMICollapse.log
+        $args | tee ${prefix}_UMICollapse.log
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
--- a/modules/nf-core/umicollapse/main.nf
+++ b/modules/nf-core/umicollapse/main.nf
@@ -22,10 +22,22 @@ process UMICOLLAPSE {
     script:
     def args   = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
-    def VERSION = '1' // WARN: Version information not provided by tool on CLI. Please update this string when bumping container versions.
-
+    def VERSION = '1.0.0-1' // WARN: Version information not provided by tool on CLI. Please update this string when bumping container versions.
+    // Memory allocation: We need to make sure that both heap and stack size is sufficiently large for
+    // umicollapse. We set the stack size to 5% of the available memory, the heap size to 90%
+    // which leaves 5% for stuff happening outside of java without the scheduler killing the process.
+    def max_heap_size_mega = (task.memory.toMega() * 0.9).intValue()
+    def max_stack_size_mega = (task.memory.toMega() * 0.05).intValue()
     """
-    umicollapse \\
+    # Getting the umicollapse jar file like this because `umicollapse` is a Python wrapper script generated
+    # by conda that allows to set the heap size (Xmx), but not the stack size (Xss).
+    # `which` allows us to get the directory that contains `umicollapse`, independent of wheather we
+    # are in a container or conda environment.
+    UMICOLLAPSE_JAR=$(dirname $(which umicollapse))/../share/umicollapse-${VERSION}/umicollapse.jar
+    java \\
+        -Xmx${max_heap_size_mega}M \\
+        -Xss${max_stack_size_mega}M \\
+        -jar \$UMICOLLAPSE_JAR \\
         bam \\
         -Xmx${task.memory.toGiga()}g  \\
         -i $bam \\
--- a/modules/nf-core/umicollapse/main.nf
+++ b/modules/nf-core/umicollapse/main.nf
@@ -26,6 +26,8 @@ process UMICOLLAPSE {
     """
     umicollapse \\
         bam \\
+        -Xms${String.format("%.0f", Math.rint(0.2 * task.memory.toGiga()))}g  \\
+        -Xmx${task.memory.toGiga()}g  \\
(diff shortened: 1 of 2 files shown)


grst:
> For me, it was fine to not set an explicit stack size with -Xss, but @grst [apparently had to do it for the smrnaseq pipeline](https://github.com/nf-core/smrnaseq/blob/d5192d48a3491504b41fe1af51860d1d9773c632/modules/nf-core/umicollapse/main.nf), so maybe it would be helpful to add this as well?

Yeah, we did get a stackoverflow error without that. Unfortunately, this can't be passed directly to the `umicollapse` command (which turned out to be a python wrapper script generated by conda), which is why I ended up calling the `jar` directly: 
https://github.com/nf-core/smrnaseq/blob/d5192d48a3491504b41fe1af51860d1d9773c632/modules/nf-core/umicollapse/main.nf#L27-L34

MatthiasZepper:
> Yeah, we did get a stackoverflow error without that. Unfortunately, this can't be passed directly to the `umicollapse` command (which turned out to be a python wrapper script generated by conda), which is why I ended up calling the `jar` directly: https://github.com/nf-core/smrnaseq/blob/d5192d48a3491504b41fe1af51860d1d9773c632/modules/nf-core/umicollapse/main.nf#L27-L34

I saw it, but wondered if the path `/usr/local/share/umicollapse-${VERSION}/umicollapse.jar` is compatible with conda? 

Also, I wondered if hardcoding the stack size to 99M advisable? Should that not also depend on the task memory? I am not good at Groovy, but maybe choosing some fraction of it, like `-Xss${String.format("%.0f", Math.rint(0.2 * task.memory.toGiga()))}g` could work?

grst:
> I saw it, but wondered if the path /usr/local/share/umicollapse-${VERSION}/umicollapse.jar is compatible with conda?

good point, I just tried in the container. But it should be possible to get this path somehow, maybe
```
$(dirname $(which umicollapse))/../share/umicollapse-${VERSION}.jar
```

Definitely a good idea to make this dependent on the task memory. We are still debugging tbh.

grst:
We had to increase the stack size even further. By setting it to `1G` all our jobs succeeded. So setting this to a few percent of the heapspace probably makes sense. 

I pushed a commit trying this out and also trying to obtain the jar file independent of wheather we are using conda or a container.

grst:
ok, seems to work in principle. The tests fail because of some checksum mismatch

modules/nf-core/umicollapse/main.nf
```diff
@@ -33,7 +33,7 @@ process UMICOLLAPSE {
     # by conda that allows to set the heap size (Xmx), but not the stack size (Xss).
     # `which` allows us to get the directory that contains `umicollapse`, independent of wheather we
     # are in a container or conda environment.
-    UMICOLLAPSE_JAR=$(dirname $(which umicollapse))/../share/umicollapse-${VERSION}/umicollapse.jar
+    UMICOLLAPSE_JAR=\$(dirname \$(which umicollapse))/../share/umicollapse-${VERSION}/umicollapse.jar
```
MatthiasZepper, line 36: Don't you have to escape the `$` for `${Version}` as well then?

Also mind that I had already added the `-Xmx` parameter, so you need to remove my declaration then.

modules/nf-core/umicollapse/main.nf
```diff
@@ -21,16 +22,27 @@ process UMICOLLAPSE {
...
+    # by conda that allows to set the heap size (Xmx), but not the stack size (Xss).
+    # `which` allows us to get the directory that contains `umicollapse`, independent of wheather we
+    # are in a container or conda environment.
+    UMICOLLAPSE_JAR=\$(dirname \$(which umicollapse))/../share/umicollapse-${VERSION}/umicollapse.jar
+    java \\
+        -Xmx${max_heap_size_mega}M \\
+        -Xss${max_stack_size_mega}M \\
+        -jar \$UMICOLLAPSE_JAR \\
         bam \\
+        -Xmx${task.memory.toGiga()}g  \\
```
grst, line 42:
```suggestion
```

modules/nf-core/umicollapse/main.nf
grst: VERSION is a groovy variable, so no ;) 
But good catch with the double Xmx!

apeltzer:
I know this might become a rabbit hole by itself, but should we look into migrating pytest to `nf-test`?

grst:
We totally should. The tests are already broken, no point in fixing pytest. 
@MatthiasZepper, do you want to go for it? Otherwise I can give it a shot

MatthiasZepper:
> We totally should. The tests are already broken, no point in fixing pytest. @MatthiasZepper, do you want to go for it? Otherwise I can give it a shot

I have 0 experience with nf-test and not really the capacity to familiarize myself with it at the moment. So if you volunteer, I am more than grateful to accept that generous offer.

apeltzer:
I have started working on nf-test tests, we can continue on the branch here on modules as we can all push there directly and I'm lacking permissions on your repo @MatthiasZepper

apeltzer:
Opened this one for collaborative editing: https://github.com/nf-core/modules/pull/4795

PR #4802, opened 2024-01-25
cellranger multi: Convert module in python template to add an automatic renaming step

fmalmeida:
Hi @grst and @klkeys ,

This relates to the following [comment](https://github.com/nf-core/scrnaseq/pull/276#discussion_r1461575657) in the scrnaseq pipeline PR where we are adding cellranger multi to it.

So, I had to change the module in order to use a 'python templates' so that it was easier to integrate the renaming step into the module.

However, there are a few points to check/discuss:

* Are all functionalities kept (I guess the testings shall check it)
* Is it ok to go forward like that or it is adding much complexity and would be better to find another way? If so, which other way?


fmalmeida:
Has the way how the testing runs are configured changed? It seems that the tests failed even before running.

modules/nf-core/cellranger/multi/templates/cellranger_multi.py
```diff
@@ -0,0 +1,187 @@
...
+# target directory in which the renamed fastqs will be placed
+fastq_all = Path("./fastq_all")
+fastq_all.mkdir(exist_ok=True)
+
+# Match R1 in the filename, but only if it is followed by a non-digit or non-character
+# match "file_R1.fastq.gz", "file.R1_000.fastq.gz", etc. but
+# do not match "SRR12345", "file_INFIXR12", etc
+filename_pattern =  r'([^a-zA-Z0-9])R1([^a-zA-Z0-9])'
+
+for i, (r1, r2) in enumerate(chunk_iter(fastqs, 2)):
```
grst, line 40: enumeration should start at 1 (my bad actually), otherwise we'll have `L000` instead of `L001` as first file.

modules/nf-core/cellranger/multi/templates/cellranger_multi.py
```diff
@@ -0,0 +1,187 @@
...
+if len("${include_frna}") > 0:
+    with open("${config}", 'a') as file, open("${frna_csv_text}", 'r') as input_conf:
+        file.write("${include_frna}\\n")
+        for line in input_conf:
+            file.write(line + "\\n")
+
+# Remove blank lines from config file
+with open('tmp.txt', 'w') as file:
+    file.write( run(['grep', '-v', '-e', '^[[:space:]]*\$', "${config}"], capture_output=True, text=True).stdout )
+run(['mv', 'tmp.txt', "${config}"])
```
grst, line 157: I'd use `run` only with `check=True`. But in this case it would probably be easier/more elegant to just build the entire file in python and only write the final file to disk.

grst (commented):
In general I think this is fine... If we are using Python already there's probably a couple of things that could be made more readable by moving operations from groovy to Python, but I'm not sure it's worth it.

The more general question is if we really want to have the file renaming as part of the module, or separately. 

**Pro in the same module**
 * performance gain when using S3 buckets

**Pro separately**
 * easier to maintain
 * easier to reuse (probably a module CELLRANGER_RENAME_READS could be used for all tools of the *ranger family, including spaceranger and cellranger arc
 * module can be used without renaming (if you don't want a potential additional source of errors and anyway only use files that follow the naming conventions). 


A full up- and download of 10s of GB of files just for renaming is pretty bad (which is why I coulpled it tightly with the module in cellranger), but it could be off by default in most pipelines and with more recent developments like sequera fusion it might not be that bad anymore.

grst:
as for the tests: 
image

fmalmeida:
> as for the tests: image

In the latest version, @klkeys added a new input file which I did not have in my copy, thus by not checking I had mistakenly removed those.

I have now tried to add them again which should get us into being able to run the tests.

I had also to modify the tests because before, the module was independent of sample-IDs, meaning it would use together whatever data was given.

However, so it can be integrated to the scrnaseq pipeline where multiple samples can be given at once, the module was changed to require that the IDs of all input meta maps are the same, which would mean they come from the same sample in the pipeline run ... then, the renaming, corrects the filenames if necessary, to have the meta.id as prefix so that cellranger can find all data.

fmalmeida:
The thing is that, by changing the meta.ids, some files which had the previous ids in it, changed md5sum, I am not sure how to update them using the tooling as it was changed to nf-test.

fmalmeida:
Hi @grst ,

As described in my last [comment](https://github.com/nf-core/modules/pull/4802#issuecomment-1910126642), I had to modify the sample IDs so that it could be integrated in the scrnaseq pipeline, so now the module **expects** all the data to be from the same sample, and thus, must have the same `meta.id` which is what will be used to data tuple collection in the pipeline.

But, because I had to alter the meta.id values in the testings, the md5sum of a few files that contained such information has changed.

You will see that I have now altered the value of these files and thus the testing passes.

What is to note and supports analysis correcteness, is that these were **the only** files that required update because they contain the sample id values. We can see that the results are the same as none of the others were touched, including the resulting `.mtx` files which contain the counts.

FYI: @klkeys 

Now the only thing remaining to decide is whether we go or not for such behaviour of having a renaming. What we can probably add is a `--skip_renaming` parameter to only copy/move the data without playing with the names if desired by the user.

apeltzer:
No strong opinion on bundling with module vs having a separate module. I see benefits / and also issues with not having it with the module as it adds potentially quite a bit of transfer impact on s3... but its both pros and cons and I dont have a clear favorite solution here.

modules/nf-core/cellranger/multi/templates/cellranger_multi.py
fmalmeida: Hi @grst ,
I have modified this one and I think it might be resolved now.

modules/nf-core/cellranger/multi/templates/cellranger_multi.py
fmalmeida: Hi @grst ,
I have changed to parse all the multiline string first and then save the file just once.
If you can take a look to see if it is resolved.

fmalmeida:
Hi @grst ,

I have now updated the code of the pipeline of the module, resolving the comments you have added and keeping the renaming inside the module as per the last comments from @klkeys and @apeltzer .

I have now also added the option to skip the renaming as a module parameter, which here in the testing is defaulted to false and in the pipeline, when integrated, can become a parameter.

Can you take another review to see if it is good for merging?

grst:
@nf-core-bot fix linting

grst (approved):
I could think of a couple of things that could be solved more elegantly now that we are using Python instead of Bash, but I can live with the current solution.

fmalmeida:
Hi @grst ,
Since this one currently allows one to perform the renaming only when desired, I will now merge it so I can continue the development of the scrnaseq sub-workflow.
Good things with it are:

* other enhancements as you mentioned can happen in parallel once merged
* development of sub-workflow can continue
* if needed, the previous version of the module, with the BASH workflow, is still accessible and usable.

👍🏼

PR #4809, opened 2024-01-29
Add umicollapse capability to run with fastq

apeltzer:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`

### DIFF
--- a/modules/nf-core/umicollapse/main.nf
+++ b/modules/nf-core/umicollapse/main.nf
@@ -9,10 +9,11 @@ process UMICOLLAPSE {
         'biocontainers/umicollapse:1.0.0--hdfd78af_1' }"
 
     input:
-    tuple val(meta), path(bam), path(bai)
+    tuple val(meta), path(input), path(bai)
 
     output:
-    tuple val(meta), path("*.bam"), emit: bam
+    tuple val(meta), path("*.bam"), emit: bam,             optional: true
+    tuple val(meta), path("*.fastq.gz"), emit: fastq,      optional: true
     tuple val(meta), path("*_UMICollapse.log"), emit: log
     path  "versions.yml"          , emit: versions
 
@@ -28,6 +29,10 @@ process UMICOLLAPSE {
     // which leaves 5% for stuff happening outside of java without the scheduler killing the process.
     def max_heap_size_mega = (task.memory.toMega() * 0.9).intValue()
     def max_stack_size_mega = (task.memory.toMega() * 0.05).intValue()
+    def file_type = args.contains("fastq") ? "fastq" :
+                    args.contains("bam") ? "bam" :
+                    input.getExtension()
+    def extension = args.contains("fastq") ? "fastq.gz" : "bam"
     """
     # Getting the umicollapse jar file like this because `umicollapse` is a Python wrapper script generated
     # by conda that allows to set the heap size (Xmx), but not the stack size (Xss).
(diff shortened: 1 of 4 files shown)


modules/nf-core/umicollapse/main.nf
```diff
@@ -28,6 +29,10 @@ process UMICOLLAPSE {
     // which leaves 5% for stuff happening outside of java without the scheduler killing the process.
     def max_heap_size_mega = (task.memory.toMega() * 0.9).intValue()
     def max_stack_size_mega = (task.memory.toMega() * 0.05).intValue()
+    def file_type = args.contains("fastq") ? "fastq" :
+                    args.contains("bam") ? "bam" :
+                    input.getExtension()
```
mahesh-panchal, line 34: Might this end up with `gz` as the extension?

apeltzer: Yes, good catch! I will remove the third clause.

MatthiasZepper:
Making two modules, one for each subcommand, would possibly have been easier for users of the module...but a smart solution. For the ordinary mortals like me, a bit of documentation explaining the trick with the empty `[]` as _bai_ input for _fastq_ that was mentioned by Mahesh on Slack would be great!

modules/nf-core/umicollapse/main.nf
```diff
@@ -28,6 +29,8 @@ process UMICOLLAPSE {
     // which leaves 5% for stuff happening outside of java without the scheduler killing the process.
     def max_heap_size_mega = (task.memory.toMega() * 0.9).intValue()
     def max_stack_size_mega = (task.memory.toMega() * 0.05).intValue()
+    def file_type = input.getExtension().contains("fastq") ? "fastq" : "bam"
```
MatthiasZepper, line 32: What happens if somebody uses `fq` instead of `fastq` for the file names? Before, you wanted to create an argument I believe, but relying on file naming conventions to select the correct subcommand seems error-prone to me?

mahesh-panchal: Agreed. Would it better to check if it ends in `bam`?

MatthiasZepper: Possibly. But technically, I see no use case for the module to dynamically decide which subcommand needs to be run. The authors of the pipeline incorporating the module will decide. So you could probably just define a value channel as input to hardcode the run mode?

mahesh-panchal: True. This is also how some other modules like mitohifi and busco do things. User provides the mode.

MatthiasZepper: Since one probably want to allow for easier configuration without messing with complicated dynamic `ext.args`, the de facto pipeline usage will likely look like this anyway?

```
include { UMICOLLAPSE_FASTQ  } from '../../../modules/nf-core/umicollapse/main'
include { UMICOLLAPSE_BAM  } from '../../../modules/nf-core/umicollapse/main'
```

modules/nf-core/umicollapse/tests/main.nf.test
```diff
@@ -150,6 +152,32 @@ nextflow_process {
...
+
+            when {
+                config "./nextflow_SE.config"
+                process {
+                    """
+                    input[0] = [
+                    [ id:'test', single_end:true ], // meta map
+                    file(params.test_data['sarscov2']['illumina']['test_1_fastq_gz'], checkIfExists: true),
+                    []
+                    ]
```
mahesh-panchal, line 165:
```suggestion
                    input[0] = [
                        [ id:'test', single_end:true ], // meta map
                        file(params.test_data['sarscov2']['illumina']['test_1_fastq_gz'], checkIfExists: true),
                        []
                    ]
```

mahesh-panchal (approved):
LGTM

PR #4817, opened 2024-01-30
Propr propr

suzannejin:
Add propr/propr module to perform compositional data analysis on rna-seq data

## PR checklist

Closes #XXX 

- [x] Add a resource `label`


modules/nf-core/propr/propr/tests/main.nf.test
```diff
@@ -0,0 +1,158 @@
...
+                    file(params.test_data['mus_musculus']['genome']['rnaseq_matrix'], checkIfExists: true)
+                ]
+                """
+            }
+        }
+        
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out.matrix).match("Test propr/propr using default options") },
```
pinin4fjords, line 30: Why only snapshotting the matrix?

tests/config/pytest_modules.yml
```diff
@@ -1,2619 +1,3577 @@
 adapterremovalfixprefix:
   - modules/nf-core/adapterremovalfixprefix/**
   - tests/modules/nf-core/adapterremovalfixprefix/**
+
```
pinin4fjords, line 4: Don't think you need to be changing this file- probably just something funny on your branch.

modules/nf-core/propr/propr/tests/main.nf.test
suzannejin: The FDR file is calculated based on a random sampling, so I don´t think we can snapshot.
Then the rds is the R object including the matrix, and the fdr, and more stuff that could be useful to run propr again if the user wants. So, same problem cannot be snapshot and have the same md5

tests/config/pytest_modules.yml
suzannejin: Cuz it always give me merging conflicts with the master... what should I do then?

modules/nf-core/propr/propr/tests/main.nf.test
pinin4fjords: Can you use set.seed() to make the FDR reproducible? You're right on the rds of course

tests/config/pytest_modules.yml
pinin4fjords: Probably something funny in your git history. Try literally copying in the file from master in a new commit?

modules/nf-core/propr/propr/tests/main.nf.test
suzannejin: That would require to change the package code and rebuild container and so on...
It is really a test that shuffles the input data and run multiple times the function used to create the matrix output, so I think if the matrix output is consistent, this step should not have problem too

suzannejin: @pinin4fjords now it should be fine. I added a way to fix the seed for fdr reproducibility

modules/nf-core/propr/propr/tests/main.nf.test
```diff
@@ -0,0 +1,163 @@
...
+                ]
+                """
+            }
+        }
+        
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out.matrix).match("Test propr/propr using default options - matrix") },
+                { assert snapshot(process.out.versions).match("versions") }
```
pinin4fjords, line 31: missing fdr check here

modules/nf-core/propr/propr/tests/main.nf.test
```diff
@@ -0,0 +1,163 @@
...
+                ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out.matrix).match("Test propr/propr while running clr+pcor.bshrink explicitly - matrix")},
+                { assert snapshot(process.out.fdr).match("Test propr/propr while running clr+pcor.bshrink explicitly - fdr")}
```
pinin4fjords, line 56: Conventionally, we have a version check in every test. But I won't insist.

pinin4fjords (changes requested):
Thanks for sorting out the seed - I think that really helps. You're just missing the FDR check on the first test.

modules/nf-core/propr/propr/tests/main.nf.test
suzannejin: hey there! fdr is an optional output actually. Here in the default test, I did not ask to compute fdr.

pinin4fjords: Ahh, my bad, thanks for explanation

pinin4fjords (approved):

suzannejin:
@pinin4fjords Thank you for the review!

PR #4827, opened 2024-01-31
Update custom/sratoolsncbisettings to only run when sra ids are provided

drpatelh:
This is to prevent the `FASTQ_DOWNLOAD_PREFETCH_FASTERQDUMP_SRATOOLS:CUSTOM_SRATOOLSNCBISETTINGS` process from running unnecessarily when no sra ids are provided to the workflow. If the sra ids channel is empty that process will no longer be run.


robsyme (approved):
LGTM

Is this just to avoid executing CUSTOM_SRATOOLSNCBISETTINGS when no IDs are provided?

maxulysse (approved):

drpatelh:
> Is this just to avoid executing CUSTOM_SRATOOLSNCBISETTINGS when no IDs are provided?

Yep, any other solutions you can think of?

PR #4838, opened 2024-02-02
CHORES: update samtools tests with new data organisation

maxulysse:


modules/nf-core/samtools/ampliconclip/tests/main.nf.test
```diff
@@ -9,94 +9,94 @@ nextflow_process {
...
-                { assert snapshot(
-                    file(process.out.bam[0][1]).name,
-                    process.out.stats,
-                    process.out.rejects_bam,
-                    process.out.versions
-                ).match() }
+                { assert snapshot(file(process.out.bam[0][1]).name).match("bam_bed_name") },
+                { assert snapshot(process.out.stats).match("bam_bed_stats") },
+                { assert snapshot(process.out.rejects_bam).match("bam_bed_rejects_bam") },
+                { assert snapshot(process.out.versions).match("bam_bed_versions") }
```
adamrtalbot, line 37: Any reason for individual snapshots?

adamrtalbot (commented):
Just one q?

modules/nf-core/samtools/ampliconclip/tests/main.nf.test
maxulysse: I find it's easier to read

adamrtalbot: Fair enough.

adamrtalbot (approved):

modules/nf-core/samtools/ampliconclip/tests/main.nf.test
pinin4fjords: Also, it prevents duplication in the snapshots. Individual named ones like this only outputs the snapshot once. If you just use `.out` you get the snapshot twice, in named and numbered form.

maxulysse: Oooh, nice, I never noticed that

PR #4855, opened 2024-02-06
Add test for initialize breaking the template

edmundmiller:
If we're moving stuff out of the template that can break the template we need to check for changes.

Can include any other files that might break the template as well.

### DIFF
--- a/.github/workflows/tools.yml
+++ b/.github/workflows/tools.yml
@@ -3,16 +3,16 @@ on:
   push:
     branches: [master]
     paths:
-      - "**/initialise**"
+      - "subworkflows/nf-core/initialise/**"
   pull_request:
     branches: [master]
     paths:
-      - "**/initialise**"
+      - "subworkflows/nf-core/initialise/**"
   merge_group:
     types: [checks_requested]
     branches: [master]
     paths:
-      - "**/initialise**"
+      - "subworkflows/nf-core/initialise/**"
 
 # Cancel if a newer run is started
 concurrency:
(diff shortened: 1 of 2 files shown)


edmundmiller:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-an-action-in-the-same-repository-as-the-workflow

edmundmiller:
> no way updating a subworkflow or module breaks the template

Love it, that's the goal!

> It's hideously convoluted to test a different repo within another repo.

Yeah it's a bit much to think about. 

How does tools currently install/update fastqc?

mashehu:
> How does tools currently install/update fastqc?

It's a manual step in the release: https://github.com/nf-core/tools/blob/master/.github/RELEASE_CHECKLIST.md?plain=1#L6

edmundmiller:
Via @mashehu -bot https://github.com/nf-core/tools/commit/8d1094ca285c433880f8fd81fa409d064417fb1b

So we'd catch it there, might just go a few months without noticing, or be something that breaks when the dependancies get bumped before release and stales it.

Maybe a cron job that runs the module and subworkflow update and bumps it in modules so it can get fixed ahead of time and not block a release?

mashehu:
> Maybe a cron job that runs the module and subworkflow update and bumps it in modules so it can get fixed ahead of time and not block a release?

couldn't renovate do this?

edmundmiller:
> couldn't renovate do this?

Probably, I think we debated it. But I think that's a whole thing to make a [renovate datasource](https://docs.renovatebot.com/modules/datasource/) for nf-core/modules.

mashehu:
closing this, because we will use https://github.com/nf-core/tools/pull/2733 instead

PR #4876, opened 2024-02-07
Add quartonotebook module

### LINKED ISSUE #4812: new module: quartonotebook

fasterius:
This PR adds a new QUARTONOTEBOOK module, which renders parametrised [Quarto](https://quarto.org/) documents to HTML. This modules re-uses some code from the RMARKDOWNNOTEBOOK and JUPYTERNOTEBOOK modules, namely part of the code that parametrizes the reports. It re-uses some of the test datasets from those modules, but also adds its own test data; Quarto can render R Markdown as well as Jupyter notebooks.

A difference between this module and its notebook-relatives is how the parametrisation is done. This module does not output a parametrised report with the relevant parameters changed from the default, but rather outputs a parametrised `params.yml` file, which is equivalent to the hidden `.params.yml` in the other modules. The main reason for this is that Quarto is built with multi-language support (Python, R, Julia and Observable, of which the former two are tested for) and has an option (`--execute-params`) that allows taking parameters from a YAML file in a language-agnostic way using `papermill` (JUPYTERNOTEBOOK also uses `papermill`). The way that the other modules do it yields a preferable output from a user-perspective, but is harder to achieve without language-specific code - I did attempt this, but failed, so if anybody has any good ideas on how to achieve this I'm interested!

## PR checklist

Closes #4812 

- [X] If necessary, include test data in your PR. (https://github.com/nf-core/test-datasets/pull/1089)
- [X] Add a resource `label`
- [X] Ensure that the test works with either Docker / Singularity.


modules/nf-core/quartonotebook/main.nf
```diff
@@ -0,0 +1,97 @@
...
+    path "versions.yml"                 , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    // Exit if running this module with -profile conda / -profile mamba
+    if (workflow.profile.tokenize(',').intersect(['conda', 'mamba']).size() >= 1) {
+        exit 1, "The QUARTONOTEBOOK module does not support Conda/Mamba, please use Docker / Singularity / Podman instead."
+    }
```
grst, line 31: why? I think quarto is on conda-forge?

modules/nf-core/quartonotebook/tests/with-parametrization.config
```diff
@@ -0,0 +1,10 @@
+env {
+    XDG_CACHE_HOME = "./.xdg_cache_home"
+    XDG_DATA_HOME = "./.xdg_data_home"
```
grst, line 3: Wouldn't it make sense for this to be set by default, i.e. add it to the `script` section of the process?

grst (commented):
Thanks for adding this! I wish quarto existed when I was working on the jupyternotebook and rmarkdown modules. 

Just two minor comments. I also thought there was something broken with the dumping the parameters to yaml in more recent versions of nextflow, but the tests seem to pass.

modules/nf-core/quartonotebook/tests/with-parametrization.config
fasterius: Good point! I have update the module with this suggestion.

modules/nf-core/quartonotebook/main.nf
fasterius: It's complicated 😅 I wrote a bit about this in commit [6c9a653](https://github.com/nf-core/modules/pull/4876/commits/6c9a653ca1f02a3dd4ab0d5996cbc5726b82a90e), but in essence it's because of several problems with architecture differences and Pandoc. Quarto is indeed on conda-forge (albeit an outdated version), but this does not work on ARM64 due to Pandoc version incompatibilities. Manually installing Quarto from their `tar.gz` releases manually inside a Docker container works for both AMD64 and ARM64, though, so that means the module will work on either architecture.

The problem comes with having a Conda directive with a different environment than the container-based one; the difference being in how Quarto and its dependencies are installed. I did retain the Conda directive (plus a disallowance if-clause for ARM64 systems) until the previously mentioned commit, but the snapshot for testing always turned out different than the container-based environment (even with the same Quarto version installed).

So, ideally, Conda will at some point in the future update their Quarto and Pandoc versions such that the dependency tree can be solved for ARM64 systems, at which point we can add the Conda directive back in.

fasterius:
As I was working on your comments I realised that I had forgotten to add the ability to use Quarto extensions for the module! I have now added commits for this, as well as outputting the original report to better reflect how RMARKDOWNNOTEBOOK and JUPYTERNOTEBOOK works - this one already outputs the parameter YAML for parametrisation, so it felt natural to also give the original report for users who want to edit it directly.

I have previously made a Quarto nf-core extension, which is here: https://github.com/fasterius/nf-core-quarto-template. We are already using it in the spatialtranscriptomics pipeline.

modules/nf-core/quartonotebook/main.nf
grst: I see, would be great to add a short comment in the code for future reference.

grst (approved):

modules/nf-core/quartonotebook/main.nf
fasterius: Good point, I have added this now.

edmundmiller:
@fasterius any links to the issue with pandoc on arm64 with amd64 emulation? Trying to use Seqera containers in https://github.com/nf-core/modules/pull/5561

edmundmiller:
It seems to? The tests pass at least. Let me know if there's something else to look for!

PR #4851, opened 2024-02-05
Update hamronization/abricate

jasmezz:
- Update hamronization version
- Add nf-test + stub sections

## PR checklist

- [x] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [ ] Remove all TODO statements.
- [ ] Emit the `versions.yml` file.
- [ ] Follow the naming conventions.
- [ ] Follow the parameters requirements.
- [ ] Follow the input/output options guidelines.
- [ ] Add a resource `label`
- [ ] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - [ ] `PROFILE=docker pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [x] `PROFILE=singularity pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [ ] `PROFILE=conda pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`

PR #4878, opened 2024-02-07
Update hamronization submodules to 1.1.4

jasmezz:
- Bumping hamronization version 1.1.1 > 1.1.4 in all five sub-modules
- (Except for hamronization/abricate with its own PR #4851 which serves as precedence case for nf-test migration in hamronization submodules)
- Migrating all five submodules to nf-test

## PR checklist

- [ ] Add a resource `label`


jasmezz:
@nf-core-bot fix linting

pinin4fjords (approved):
Nothing too controversial here. You might consider using individual named snapshots for each output, which would prevent the duplication you see in the .snap files, but not a blocking concern.

jasmezz:
@pinin4fjords Thanks for the review! :rocket: Hmm you don't mean an individual name for each test? Because that should already be the case, all tests have a descriptive name and the corresponding stub ends with "stub". Is there another way to name outputs/tests, you have a link to an example?

pinin4fjords:
> @pinin4fjords Thanks for the review! 🚀 Hmm you don't mean an individual name for each test? Because that should already be the case, all tests have a descriptive name and the corresponding stub ends with "stub". Is there another way to name outputs/tests, you have a link to an example?

So, you're currently doing e.g.:

```
```

Which snapshots all the outputs at once. It also duplicates in the snapshot such that you get one numbered  and once named entry.

Doing something like:

```
{ assert snapshot(process.out.tsv).match('tsv_testname') }
{ assert snapshot(process.out.json).match('json_testname') }
{ assert snapshot(process.out).match('versions_testname') }
```

is more explicit, and you only get one entry in the snapshot.

But as I say, not a blocking thing.

jasmezz:
Ahh alright! Thanks. Makes sense, on the other hand, the `test.nf` looks clean with just snapshotting _everything_ in 1 line. Will merge for now, but keep in mind for future updates!

PR #4891, opened 2024-02-09
add checkProfileProvided to utils_nfcore_pipeline

mirpedrol:
Add `checkProfileProvided` to `utils_nfcore_pipeline` according to the changes in the pipeline template https://github.com/nf-core/tools/pull/2736

### DIFF
--- a/subworkflows/nf-core/utils_nfcore_pipeline/tests/main.function.nf.test
+++ b/subworkflows/nf-core/utils_nfcore_pipeline/tests/main.function.nf.test
@@ -21,6 +21,26 @@ nextflow_function {
         }
     }
 
+    test("Test Function checkProfileProvided") {
+
+        function "checkProfileProvided"
+
+        when {
+            function {
+                """
+                input[0] = []
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert function.success },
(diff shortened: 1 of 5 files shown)


subworkflows/nf-core/utils_nfcore_pipeline/tests/main.function.nf.test
```diff
@@ -21,6 +21,26 @@ nextflow_function {
         }
     }
 
+    test("Test Function checkProfileProvided") {
```
mashehu, line 24: How about a test with a faulty input?

mashehu (approved):

subworkflows/nf-core/utils_nfcore_pipeline/tests/main.function.nf.test
mirpedrol: I couldn't find a way to provide a `workflow.profile` in the nf-test docs, nor how to check the content of the warning logging, so I'm not sure how to test a case of failure in this case.

mashehu: Okay, maybe I am thinking too much in a pytest way. 🙂 Good as it is then

PR #4892, opened 2024-02-10
Update utils-nfcore-pipeline: add multiqc report to completion email

mirpedrol:
Change according to the nf-core pipeline template https://github.com/nf-core/tools/pull/2736

### DIFF
--- a/subworkflows/nf-core/utils_nfcore_pipeline/main.nf
+++ b/subworkflows/nf-core/utils_nfcore_pipeline/main.nf
@@ -249,7 +249,7 @@ def logColours(monochrome_logs=true) {
 //
 // Construct and send completion email
 //
-def completionEmail(summary_params, email, email_on_fail, plaintext_email, outdir, monochrome_logs=true) {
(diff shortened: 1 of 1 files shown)


subworkflows/nf-core/utils_nfcore_pipeline/main.nf
```diff
@@ -287,6 +287,24 @@ def completionEmail(summary_params, email, email_on_fail, plaintext_email, outdi
     email_fields['projectDir']   = workflow.projectDir
     email_fields['summary']      = summary << misc_fields
 
+    // On success try attach the multiqc report
+    def mqc_report = null
```
mashehu, line 291:
```suggestion
    def mqc_report = []
```
So it's the same type as the `multiqc_report` parameter

mashehu (approved):

mirpedrol:
Moved attaching the mutiqc report to a separate function with https://github.com/nf-core/modules/pull/4892/commits/fa89e991a2fb7413f693141f71bfe12a1e3ef7ae. 
I have tested that the value of mqc_report is a file path when we use multiqc. In cases where we don't use multiqc we can pass null instead of the output channel.

PR #4895, opened 2024-02-12
NARFMAP segfault

edmundmiller:
For some reason this fails in nascent, but not here...

### DIFF
--- a/modules/nf-core/narfmap/align/tests/main.nf.test
+++ b/modules/nf-core/narfmap/align/tests/main.nf.test
@@ -7,6 +7,7 @@ nextflow_process {
     tag "modules_nfcore"
     tag "narfmap"
     tag "narfmap/align"
+    tag "narfmap/hashtable"
 
     test("sarscov2 - fastq, hashtable, false") {
 
--- a/modules/nf-core/narfmap/align/tests/main.nf.test
+++ b/modules/nf-core/narfmap/align/tests/main.nf.test
@@ -42,10 +42,12 @@ nextflow_process {
             assertAll (
                 { assert process.success },
                 { assert snapshot(
+                    process.out.bam,
+                    process.out.versions,
                     file(process.out.bam[0][1]).name,
                     file(process.out.log[0][1]).readLines().findAll { it.startsWith("decompHash") },
-                    file(process.out.versions[0]).name
-                ).match() }
+                ).match()
+                }
             )
         }
 
@@ -84,10 +86,12 @@ nextflow_process {
             assertAll (
                 { assert process.success },
                 { assert snapshot(
+                    process.out.bam,
+                    process.out.versions,
                     file(process.out.bam[0][1]).name,
                     file(process.out.log[0][1]).readLines().findAll { it.startsWith("decompHash") },
-                    file(process.out.versions[0]).name
-                ).match() }
+                ).match()
+                }
             )
         }
 
@@ -129,10 +133,12 @@ nextflow_process {
             assertAll (
                 { assert process.success },
                 { assert snapshot(
+                    process.out.bam,
+                    process.out.versions,
                     file(process.out.bam[0][1]).name,
                     file(process.out.log[0][1]).readLines().findAll { it.startsWith("decompHash") },
-                    file(process.out.versions[0]).name
-                ).match() }
+                ).match()
+                }
             )
         }
 
@@ -174,10 +180,12 @@ nextflow_process {
             assertAll (
                 { assert process.success },
                 { assert snapshot(
+                    process.out.bam,
+                    process.out.versions,
                     file(process.out.bam[0][1]).name,
                     file(process.out.log[0][1]).readLines().findAll { it.startsWith("decompHash") },
-                    file(process.out.versions[0]).name
-                ).match() }
+                ).match()
+                }
             )
         }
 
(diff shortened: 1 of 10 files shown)


nvnieuwk (approved):
LGTM, just gotta fix those linting issues

edmundmiller:
Sorry that wasn't clear. This test shouldn't pass 😆 It fails in Nascent, need to try to reproduce with the sarek test as well 🙃

modules/nf-core/narfmap/align/tests/segfault.nf.test
```diff
@@ -0,0 +1,55 @@
...
+            }
+        }
+
+        then {
+            assertAll (
+                { assert process.success },
+                { assert snapshot(
+                    file(process.out.bam[0][1]).name,
+                    file(process.out.log[0][1]).readLines().findAll { it.startsWith("decompHash") },
+                    file(process.out.versions[0]).name
```
maxulysse, line 48: process.out.versions[0] ? why not process.out.versions?

edmundmiller: Because I started this PR in February 😅

modules/nf-core/narfmap/align/tests/segfault.nf.test
```diff
@@ -0,0 +1,55 @@
...
+    script "../main.nf"
+    process "NARFMAP_ALIGN"
+    tag "modules"
+    tag "modules_nfcore"
+    tag "narfmap"
+    tag "narfmap/align"
+    tag "narfmap/hashtable"
+    tag "bug"
+
+    test("This should segfault") {
```
maxulysse, line 13:
```suggestion
    test("Human - bam") {
```

modules/nf-core/narfmap/align/tests/main.nf.test
```diff
@@ -42,10 +42,12 @@ nextflow_process {
             assertAll (
                 { assert process.success },
                 { assert snapshot(
+                    process.out.bam,
+                    process.out.versions,
                     file(process.out.bam[0][1]).name,
```
maxulysse, line 47: I think you can remove this line as well,since you're snapshotting process.out.bam already

modules/nf-core/narfmap/align/tests/main.nf.test
```diff
@@ -84,10 +86,12 @@ nextflow_process {
...
```
maxulysse, line 91: I think you can remove this line as well,since you're snapshotting process.out.bam already

modules/nf-core/narfmap/align/tests/main.nf.test
```diff
@@ -129,10 +133,12 @@ nextflow_process {
...
```
maxulysse, line 138: I think you can remove this line as well,since you're snapshotting process.out.bam already

modules/nf-core/narfmap/align/tests/segfault.nf.test
edmundmiller: This is a regression test, not a standard test. I don't know if it's related to the sample being Human or not. But I think it should break naming conventions.

modules/nf-core/narfmap/align/tests/segfault.nf.test
```diff
@@ -0,0 +1,55 @@
...
+
+        setup {
+            run("NARFMAP_HASHTABLE") {
+                script "../../hashtable/main.nf"
+                process {
+                    """
+                    input[0] = [
+                        [id:'test'],
+                        file('https://raw.githubusercontent.com/nf-core/test-datasets/nascent/reference/GRCh38_chr21.fa', checkIfExists: true)
```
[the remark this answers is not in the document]
edmundmiller: No, that's the issue. The data from modules works. It used to fail for whatever reason with this test dataset specifically, so this is checking for a regression until I get proper testing set up in the NARFMAP repo.

modules/nf-core/narfmap/align/tests/main.nf.test
edmundmiller: I just copied what was at bwa/mem 🤷🏻‍♂️ 

https://github.com/nf-core/modules/blob/58ebf878bd5285a3de23cd52f14a343b7cd47d98/modules/nf-core/bwa/mem/tests/main.nf.test#L45-L53

modules/nf-core/narfmap/align/tests/segfault.nf.test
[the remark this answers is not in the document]
maxulysse: Oooh I see.
I think the data we have in Sarek fails too, it's coming from modules if I remember well.
Just not the regular sarscov2, but human data

modules/nf-core/narfmap/align/tests/main.nf.test
maxulysse: I see

maxulysse: but in your case, since you snapshot `process.out.bam`, you don't need to snapshot `file(process.out.bam[0][1]).name` as well

maxulysse: I think we need to update the bwa/mem tests with the nft-bam plugins

edmundmiller: Made an issue for updating those and cleaned this one up for now.

maxulysse:
ok, so we update this back in nascent and sarek now?

edmundmiller:
No it didn't work 😞

Issue #4934, opened 2024-02-16
Dockerfile or container image provenance information missing from deepvariant module


mahesh-panchal:
### Have you checked the docs?

- [X] [nf-core website: troubleshooting](https://nf-co.re/usage/troubleshooting)
- [X] [nf-core modules documentation](https://nf-co.re/docs/contributing/modules)

### Description of the bug

https://github.com/nf-core/modules/blob/master/modules/nf-core/deepvariant/main.nf points to a container located at
```
container "nf-core/deepvariant:1.5.0"
```
However, unlike Cellranger, there is no Dockerfile in the same directory, or information where the image comes from. 

Slack discussion here: https://nfcore.slack.com/archives/CJRH30T6V/p1708079375154689 between me and @maxulysse .

maxulysse:
note, add info for the gatk containers from the broad too

PR #4900, opened 2024-02-13
bump gatk4 modules

matthdsm:
Bumping GATK4 to v4.5.0.0

## PR checklist

Closes #XXX 

- [ ] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [ ] Remove all TODO statements.
- [ ] Emit the `versions.yml` file.
- [ ] Follow the naming conventions.
- [ ] Follow the parameters requirements.
- [ ] Follow the input/output options guidelines.
- [ ] Add a resource `label`
- [ ] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - [ ] `PROFILE=docker pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [ ] `PROFILE=singularity pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [ ] `PROFILE=conda pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`

PR #4935, opened 2024-02-16
Add README to modules build with Docker

maxulysse:
close https://github.com/nf-core/modules/issues/4934


mahesh-panchal (commented):
The Dockerfiles are duplicates no? Why not put the GATK Dockerfile in the parent so we know it's common to all?

maxulysse:
> The Dockerfiles are duplicates no? Why not put the GATK Dockerfile in the parent so we know it's common to all?

because we use this Dockerfile for just 4 gatk modules over 59, for all the rest we just use biocontainers as usual

mahesh-panchal:
> > The Dockerfiles are duplicates no? Why not put the GATK Dockerfile in the parent so we know it's common to all?
> 
> because we use this Dockerfile for just 4 gatk modules over 59, for all the rest we just use biocontainers as usual

Ah, but there's still the issue the file could become out of sync between copies. Can they be symlinks instead?

maxulysse:
I'll make sure to check that whenever there's an update on the modules, and given that @matthdsm is actually making that change right now in https://github.com/nf-core/modules/pull/4900
I actually already used gatk 4.5, so let's wait for https://github.com/nf-core/modules/pull/4900 before merging this one please

matthdsm:
I'm not sure I like just copy pasting a dockerfile from another repo. I'd much rather just refer to the original file through a permalink or whatever.

Ideally we're able to fix the bioconda recipe/biocontainer and we can leave the bloated broad dockerfile behind

edmundmiller:
Yeah I'm against the Dockerfile copy pasting, they're probably just going to bitrot.

Could we build them with Spack and wave?

edmundmiller:
https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/gatk/package.py

Yeah maybe we just toss those in there?

matthdsm:
Not good enough, still missing the necessary custom python pkgs

modules/nf-core/gatk4/postprocessgermlinecnvcalls/Dockerfile
```diff
@@ -0,0 +1,102 @@
...
+
+# Start GATK Python environment
+
+WORKDIR /gatk
+ENV PATH $CONDA_PATH/envs/gatk/bin:$CONDA_PATH/bin:$PATH
+RUN conda env create -n gatk -f /gatk/gatkcondaenv.yml && \
+    echo "source activate gatk" >> /gatk/gatkenv.rc && \
+    echo "source /gatk/gatk-completion.sh" >> /gatk/gatkenv.rc && \
+    conda clean -afy && \
+    rm -rf /root/.cache/pip
```
edmundmiller, line 96: @matthdsm Are these the lines?

matthdsm: Nah, I think its a pip install somewhere at the bottom of the dockerfile. 
Can’t look it right now, on mobile

matthdsm: It’s this line here. 
https://github.com/broadinstitute/gatk/blob/cfd4d87ec29ac45a68f13a37f30101f326546b7d/scripts/gatkcondaenv.yml.template#L66
Afaik, this should install the custom scripts, but for some reason, these aren’t detected when the conda recipe is used

maxulysse:
Replaced the Dockerfiles with self-explanatory READMEs

matthdsm (approved):
Looks good!

maxulysse:
germlinecnvcaller tests are failing, did you fix theses in your PR?

matthdsm:
Yep, tests are passing in my PR

maxulysse:
Ok, then let's wait until yours is merged to merge this one

edmundmiller:
Got wave working with Dockerfiles FYI https://github.com/nf-core/modules/actions/runs/7944492424/job/21690068698(spack not so much)

PR #4937, opened 2024-02-16
Dp24 tabix update

DLBPointon:
Updating the tabix modules to include nf-test


DLBPointon:
Forgot the tags!

DLBPointon:
Fails due to the existence of intermediary full file paths being included in the snap, also a known issue with gzipped files and nf-test.

modules/nf-core/tabix/bgziptabix/tests/main.nf.test
```diff
@@ -0,0 +1,83 @@
+nextflow_process {
+
+    name "Test Process TABIX_BGZIPTABIX"
+    script "modules/nf-core/tabix/bgziptabix/main.nf"
+    process "TABIX_BGZIPTABIX"
```
sateeshperi, line 5:
```suggestion
    process "TABIX_BGZIPTABIX"
    config "./tabix_tbi.config"
```

if the config is same for all tests...apply here to make it global

modules/nf-core/tabix/bgzip/tests/main.nf.test
```diff
@@ -0,0 +1,95 @@
...
+                                [ file(params.test_data['sarscov2']['illumina']['test_vcf'], checkIfExists: true) ]
+                ]
+                """
+            }
+        }
+
+        then {
+            assertAll (
+                { assert process.success },
+                { assert snapshot(process.out).match() }
```
nvnieuwk, line 91: There currently is a bug in `nf-test` that fails when you `touch` a file ending in `.gz`. The tool thinks it's gzipped because of this, which is not the case and it fails on that. You have two possible solutions:
1. You change the `touch ${prefix}.gz` command in the stub to `echo "" | gzip > ${prefix}.gz`
2. You only check the filenames with `nf-test` like this:
```
snapshot(
   file(process.out[0][1]).name
).match("stub_module")
```
(You should add the `file(...).name` code for every file in the output channel)

sateeshperi: can confirm this is a bug.....thanks @nvnieuwk

DLBPointon: Thanks both for the input. I can now get most of the testing working, tuples in the format `tuple val(meta), path("*.gz"), path("*.tbi")` are evading me for the bgziptabix module, I'm wondering whether the format should be changed to 2 channels in place of this 1.

Anywhoo...

Looks like the only major issue now is conda.

Where it gets interesting is that tabix 1.11 is defined everywhere, and used nowhere. 1.11 is the only version on Galaxy and Conda too as the only post-version 1 release.

I can only assume this is a change that was implemented after the original modules were made but there is no evidence of that.

Containers using 1.12 and conda using 1.19. There's also some small change in the output causing a change in the d5sum, however, I've unzipped a couple of the output files and they are visually identical. Index files between versions are identical so I'm guessing a newline character is to blame.

I guess it's time to contact the dev?

nvnieuwk: `tabix` was officially moved to `htslib` from version 1.12 and on (if I remember correctly). Maybe it's just time to use `htslib` instead?

maxulysse: Oh I see, then the issue is the same that we had with samtools.
You can just add htslib to pin it to the proper version

maxulysse: cf https://github.com/nf-core/modules/pull/4773
https://github.com/nf-core/modules/pull/4831

nvnieuwk (approved):
LGTM! Maybe wait for another approval since these modules are wildly used

modules/nf-core/tabix/bgzip/environment.yml
```diff
@@ -4,4 +4,4 @@ channels:
   - bioconda
   - defaults
 dependencies:
-  - bioconda::tabix=1.11
+  - bioconda::htslib=1.19.1
```
maxulysse, line 7: I'd rather we use the tabix recipe and pin htslib similarly to what is done for the samtools modules, as this is still a tabix modules

DLBPointon:
@maxulysse Could I have an approval if the changes are ok?

matthdsm (approved):
LGTM!

PR #4940, opened 2024-02-16
Wave with Dockerfiles

edmundmiller:
For @matthdsm on #4935

Branched off from #4080

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -12,3 +12,4 @@ modules/nf-core/pints/caller/** @edmundmiller
 modules/nf-core/preseq/ccurve/** @edmundmiller
 modules/nf-core/preseq/lcextrap/** @edmundmiller
 subworkflows/nf-core/homer/** @edmundmiller
+modules/nf-core/**/Dockerfile @edmundmiller @maxulysse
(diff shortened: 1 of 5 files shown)


matthdsm:
Which part of this is for me exactly?

edmundmiller:
Just in hopes of making the GATK custom images easier to generate!

matthdsm:
No apologies needed! I just wasn't following what what happening exactly 😛 
Nice work on the wave stuff!

edmundmiller:
@maxulysse This was from the maintainer meeting on Friday. https://github.com/nf-core/modules/commit/116adce113b45050bdf437405147e7b7feeac02f

Idk if that's what you were thinking about too, but it was at least what was in my head. 🙃

matthdsm:
That's exactly what the Broad does. Does it work??

matthdsm:
Apparently not 😞 
```
Requirement 'gatkPythonPackageArchive.zip' looks like a filename, but the file does not exist
Processing ./gatkPythonPackageArchive.zip\nERROR: Could not install packages due to an OSError: [Errno 2] No such file or directory: '/tmp/gatkPythonPackageArchive.zip
```

edmundmiller:
I think we'd have to get the path right, and even then I'm not sure if it'll play nicely with the conda install

edmundmiller:
I stand corrected. https://simonwillison.net/2022/Apr/24/pip-install-github/

edmundmiller:
So it worked... but it broke wave...

edmundmiller:
This should be ready to go! Dockerfiles would need reworking, but it's there to be used now 🤷🏻‍♂️

maxulysse (approved):

PR #4950, opened 2024-02-19
Updates to gnu sort module to include nf-test

DLBPointon:
Updated the gnu/sort module to include nf-test
Changed the stub - looks like I just copy-pasted the command into the stub when I originally wrote it.


nvnieuwk (approved):
LGTM

PR #4964, opened 2024-02-21
Updates for Seqtk

DLBPointon:
Updated the 6 seqtk modules. All are now 1.4 rather than a mix of 1.3 and 1.4.

Added a tuple in and out to subseq.

Added nf-tests.

4 modules had no stub, now added and tested.


modules/nf-core/seqtk/cutn/tests/main.nf.test
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        then {
+            assertAll (
+                { assert process.success },
+                { assert snapshot(process.out).match() },
+                { assert snapshot(
+                            file(process.out.bed[0][1]).name
+                                ).match("genome_cut")
+                }
```
maxulysse, line 32: I don't think you need that if you already snapshot `process.out`

DLBPointon: I thought that the above was specifically for the file where as the `process.out` was there to catch everything. Happy to be wrong and remove the `process.out` though.

DLBPointon: Hi @maxulysse,
You were right, so i've updated to just use process out as that is better for static output. The other method is for when output has content varying between runs. Just a misunderstanding on my part to include both.

There's now an issue with the runner but that's under investigation.

DLBPointon:
@maxulysse Could I have an approval if the above is ok?

DLBPointon:
NF-Test API limit hit once again.

PR #4968, opened 2024-02-22
Added parallel version of ltrharvest

GallVp:
## PR checklist

Closes #XXX 

- [x] Added parallel version of ltrharvest
- [x] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/ltrharvest/meta.yml
+++ b/modules/nf-core/ltrharvest/meta.yml
@@ -12,13 +12,12 @@ keywords:
   - transposons
   - retrotransposons
 tools:
-  - "edta":
-      description: Extensive de-novo TE Annotator (EDTA)
-      homepage: "https://github.com/oushujun/EDTA"
-      documentation: "https://github.com/oushujun/EDTA"
-      tool_dev_url: "https://github.com/oushujun/EDTA"
-      doi: "10.1186/s13059-019-1905-y"
-      licence: ["GPL v3"]
+  - "LTR_HARVEST_parallel":
+      description: A Perl wrapper for LTR_harvest
+      homepage: "https://github.com/oushujun/EDTA/tree/v2.2.0/bin/LTR_HARVEST_parallel"
+      documentation: "https://github.com/oushujun/EDTA/tree/v2.2.0/bin/LTR_HARVEST_parallel"
+      tool_dev_url: "https://github.com/oushujun/EDTA/tree/v2.2.0/bin/LTR_HARVEST_parallel"
+      licence: ["MIT"]
   - "gt":
       description: "The GenomeTools genome analysis system"
       homepage: "https://genometools.org/index.html"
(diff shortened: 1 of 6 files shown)


modules/nf-core/ltrharvest/meta.yml
```diff
@@ -0,0 +1,60 @@
...
+      tool_dev_url: "https://github.com/oushujun/EDTA"
+      doi: "10.1186/s13059-019-1905-y"
+      licence: ["GPL v3"]
+  - "gt":
+      description: "The GenomeTools genome analysis system"
+      homepage: "https://genometools.org/index.html"
+      documentation: "https://genometools.org/documentation.html"
+      tool_dev_url: "https://github.com/genometools/genometools"
+      doi: "10.1109/TCBB.2013.68"
+      licence: ["ISC"]
```
mahesh-panchal, line 28: This is unusual, but I guess it makes sense. 
Include an entry specifically for 'ltrharvest' so it's origin is clear

GallVp: Thank you @mahesh-panchal 

I have added an entry for LTR_harvest_parallel.

mahesh-panchal (approved):

modules/nf-core/ltrharvest/tests/main.nf.test
```diff
@@ -0,0 +1,60 @@
...
+                ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() },
+                { assert snapshot(path(process.out.versions[0]).text).match("stub_versions") }
```
ahvigil, line 54: the `stub_versions` and `script_versions` snapshots look the same. I think you can share those snapshots between test cases (`.match("versions")`) since the script and stub blocks use the exact same logic

ahvigil (approved):
minor nitpick about simplifying test snapshot, otherwise looks good

modules/nf-core/ltrharvest/tests/main.nf.test
ahvigil: it would also be ok to remove this check entirely or loosen it a bit to just check for existence, so you don't need to update the snapshots for minor version bumps... but its fine if you want to be conservative here

GallVp: Hi @ahvigil 

Thank you for the feedback. It is not possible anymore to use the same snapshot name. nf-test doesn't allow that anymore.

ahvigil: ugh thats too bad but it seems you're right: https://github.com/askimed/nf-test/pull/188

however, there is an open issue to provide an opt in for cases like this where it might be useful to remove repetition: https://github.com/askimed/nf-test/issues/197. Hope they can implement something like that!

PR #4972, opened 2024-02-22
fix tiny bug in gawk module

MariekeVromman:
adds missing brackets

### DIFF
--- a/modules/nf-core/gawk/main.nf
+++ b/modules/nf-core/gawk/main.nf
@@ -22,7 +22,7 @@ process GAWK {
     def args  = task.ext.args  ?: '' // args is used for the main arguments of the tool
     def args2 = task.ext.args2 ?: '' // args2 is used to specify a program when no program file has been given
     prefix = task.ext.prefix ?: "${meta.id}"
-    suffix = task.ext.suffix ?: "${input.getExtension}"
+    suffix = task.ext.suffix ?: "${input.getExtension()}"
 
     program = program_file ? "-f ${program_file}" : "${args2}"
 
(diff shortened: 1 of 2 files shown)


mashehu (approved):

nvnieuwk (approved):
LGTM if tests pass :rocket:

MariekeVromman:
linting is failing due to the license

```
The meta.yml of the module gawk is not valid: 
Licence must be an array of one or more entries, e.g. ["MIT"]
Check the entry for tools.
The current value is ['GPL v3'].
```

@mashehu , @nvnieuwk , how do I know which values are allowed here? I tried looking in the docs but did not immediately find it. According to this [site](https://www.gnu.org/software/gawk/manual/gawk.html), gawk falls under the GNU license, but I don't know if 'GNU' is ok for this field. Or maybe 'GLP' without the version...?

nvnieuwk:
You should change the `licence: "['GPL v3']"` to `licence: ['GPL v3']`. That's an artifact from the older tools versions that has been removed (@mashehu can probably go into more detail on the why)

MariekeVromman:
Sooo, the tests were all passing, and I added the PR to the merge queue, but then the docker test started failing (looks like it can not pull the image?).
I don't really know how to go forward now 😅
@nvnieuwk , @mashehu , could you point me to the next steps? Thanks a lot!

mashehu:
You unfortunately hit one of the wonky API call limits, or as @apeltzer put it on #nf-core-memes:
image

The Solution: rerun the failed job and hope that your API call gets through 🤞🏻

MariekeVromman:
thanks a lot!

PR #4981, opened 2024-02-23
Bump kallistobustools to 0.28.2

gennadyFauna:
Changes:

- Bumped kb version to 0.28.2.
- Added support for the new `nac` workflow.
- Updated `main.nf` for `ref` and `count` to match nomenclature in `meta.yml` files: `intron` instead of `introns`
- Updated `count` to match `ref` and take in a `workflow_mode` variable. This fixes a bug: the results of the, e.g., `lamanno` pipeline are incorrect if the workflow is not specified; `kb` generates one matrix instead of two and ignores the intronic content.
- Updated tests to account for new quantification logic, removing the `nucleus` option and adding the `nac` one, and passing the workflow variable directly into `count`.
- Updated the test outputs.

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


gennadyFauna:
The upstream conda distributable has been rebuilt on the relevant repositories.

CarsonJM (commented):
Great work! At the moment we are trying to migrate modules from pytest to nf-test (https://nf-co.re/docs/contributing/modules#migrating-from-pytest-to-nf-test) would you be able to follow this guide and add nf-test?

gennadyFauna:
I appreciate it! Unfortunately, at this time, I do not think I am comfortable enough to convert the pytest checks to nf-test, and would prefer to handle that in a future PR.

CarsonJM:
@gennadyFauna I went through and added the nf-tests if you are interested in taking a look!

CarsonJM:
Here is a brief summary of how I went about this:

1. Ran the following code to add the appropriate files:
`nf-core modules create kallistobustools/ref --migrate-pytest`
``nf-core modules create kallistobustools/count --migrate-pytest`
2. I added the files that were used as input in the pytest to the nf-test
3. I ran the nf-tests using the following command to see if there were any inconsistent hashes in the output
`nf-core modules test kallistobustools/ref`
`nf-core modules test kallistobustools/count`
4. There were inconsistent hashes (when the tests fail is will show you which files had changed md5sums). So I added a line to test all files except those with inconsistent hashes:
`path(process.out.count.get(0).get(1)).list().findAll { file(it.toString()).name != "run_info.json" && file(it.toString()).name != "kb_info.json" },`  
5. I separate tests for the inconsistent files where I read the lines that don't contain dates/times/runtimes (these are typically what cause inconsistent hashes)
`file(path(process.out.count.get(0).get(1)).list().find { file(it.toString()).name == "kb_info.json" }.toString()).readLines()[2..10],`
6. After going through this process for all inconsistent files and both modules, the tests passed!

Totally okay if this doesn't make sense yet, I just wanted to explain the changes I made and hopefully provide an intro to how nf-tests work!

modules/nf-core/kallistobustools/count/main.nf
```diff
@@ -36,12 +37,29 @@ process KALLISTOBUSTOOLS_COUNT {
...
+        kallistobustools: \$(echo \$(kb --version 2>&1) | sed 's/^.*kb_python //;s/positional arguments.*\$//')
+    END_VERSIONS
+    """
+
+    stub:
+    def args    = task.ext.args ?: ''
+    def prefix  = task.ext.prefix ?: "${meta.id}"
+    def cdna    = t1c ? "-c1 $t1c" : ''
+    def intron  = t2c ? "-c2 $t2c" : ''
+    def memory  = task.memory.toGiga() - 1
```
jasmezz, line 59:
```suggestion
    def prefix  = task.ext.prefix ?: "${meta.id}"
```
Since you don't use these variables in the stub code, you can get rid of them. Unless you prefer to have it as similar to the script block as possible?

modules/nf-core/kallistobustools/count/tests/main.nf.test
```diff
@@ -0,0 +1,98 @@
...
+
+        when {
+            process {
+                """
+                input[0] = Channel.of(
+                    [
+                        [id:'test'], // meta map
+                        [
+                        file(params.test_data['homo_sapiens']['10xgenomics']['cellranger']['test_10x_10k_pbmc_cmo_gex1_fastq_1_gz'], checkIfExists: true),
+                        file(params.test_data['homo_sapiens']['10xgenomics']['cellranger']['test_10x_10k_pbmc_cmo_gex1_fastq_2_gz'], checkIfExists: true)
```
jasmezz, line 36:
```suggestion
                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/10xgenomics/cellranger/10k_pbmc_cmo/fastqs/gex_1/subsampled_SC3_v3_NextGem_DI_CellPlex_Human_PBMC_10K_1_gex_S2_L001_R1_001.fastq.gz', checkIfExists: true),
                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/10xgenomics/cellranger/10k_pbmc_cmo/fastqs/gex_1/subsampled_SC3_v3_NextGem_DI_CellPlex_Human_PBMC_10K_1_gex_S2_L001_R2_001.fastq.gz', checkIfExists: true)
```

To keep it consistent...

modules/nf-core/kallistobustools/count/tests/main.nf.test
```diff
@@ -0,0 +1,98 @@
...
```
jasmezz, line 77:
```suggestion
                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/10xgenomics/cellranger/10k_pbmc_cmo/fastqs/gex_1/subsampled_SC3_v3_NextGem_DI_CellPlex_Human_PBMC_10K_1_gex_S2_L001_R1_001.fastq.gz', checkIfExists: true),
                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/10xgenomics/cellranger/10k_pbmc_cmo/fastqs/gex_1/subsampled_SC3_v3_NextGem_DI_CellPlex_Human_PBMC_10K_1_gex_S2_L001_R2_001.fastq.gz', checkIfExists: true)
```

modules/nf-core/kallistobustools/ref/main.nf
```diff
@@ -62,4 +62,33 @@ process KALLISTOBUSTOOLS_REF {
         END_VERSIONS
         """
     }
+
+    stub:
+    def args = task.ext.args ?: ''
```
jasmezz, line 67:
```suggestion
```

Unused variable like above

modules/nf-core/kallistobustools/ref/tests/main.nf.test
```diff
@@ -0,0 +1,120 @@
...
+                { assert snapshot(
+                    process.out.versions,
+                    process.out.t2g,
+                    process.out.cdna,
+                    process.out.intron,
+                    process.out.cdna_t2c,
+                    process.out.intron_t2c,
+                    ).match()
+                },
+                { assert file(process.out.index.get(0)).exists() }
```
jasmezz, line 36: Cannot check for file content for these index files?

jasmezz (approved):
Great, another nf-test migration :tada: Check my comments and update if you find necessary.

modules/nf-core/kallistobustools/ref/main.nf
gennadyFauna: probably just putting $args into the kb ref call.

modules/nf-core/kallistobustools/count/main.nf
gennadyFauna: sorry, I am not familiar with these conventions. How would this work if all of these variables are required for kb count?

gennadyFauna: Not sure I understand this: all of these variables go into the count call.

jasmezz: Look closely: Those variables are required for the `kb count` call in the the `script:` block above, but not here in the `stub:` block. Only `$prefix` is used below. That's why I can tell even without knowing the kallisto tool, that they are not required here.

modules/nf-core/kallistobustools/ref/main.nf
jasmezz: Same reason as the other comment, the `$args` is not used in the below `stub:` code, only in the `script:` block (which has its own variable definitions).

modules/nf-core/kallistobustools/ref/tests/main.nf.test
CarsonJM: I think we probably could, but I honestly didn't look into it too much haha. I think you're right so I'll see what I can do before merging

CarsonJM:
@gennadyFauna In short, the changes that @jasmezz made to the `stub` section were to remove unnecessary code. The stub section is basically to perform a "dry run" of a module, so only the outputs are created via `touch` but nothing is actually run.

The great benefit of this is that you can very quickly make sure that your inputs/outputs chain together correctly without actually running all of the tools.

modules/nf-core/kallistobustools/ref/tests/main.nf.test
jasmezz: Would make it slightly more robust. Checking for existence is my last resort when md5 or file content checks are impossible (e.g. a binary file with changing content due to timestamp or something).

CarsonJM:
@jasmezz The index is in some compressed format and I couldn't quickly find out which. Merging for now get this in for @gennadyFauna, but definitely something to look into in the future

gennadyFauna:
@jasmezz Thank you – I see, this is for the tests, then. I am just somewhat confused about the changes to the `ref` main.nf:

https://github.com/nf-core/modules/pull/4981/files#diff-a9f2f88b26c29c6e7b7c6d62c15b44ca381c6d125e8eb3fecaca26aff3ce347f

As far as I understand, this should not have any test materials. But now I do not really see where `kb ref` is called at all.

@CarsonJM that should be the kallisto index. But I'm not sure there is documentation about its specifics.

PR #4992, opened 2024-02-26
cat/fastq : Using modules_testdata_base_path instead of test_data in nf-test

asp8200:
Needed for getting nf-tests to run in Sarek.

### DIFF
--- a/modules/nf-core/cat/fastq/tests/main.nf.test
+++ b/modules/nf-core/cat/fastq/tests/main.nf.test
@@ -16,11 +16,11 @@ nextflow_process {
             }
             process {
                 """
-                input[0] = [
+                input[0] = Channel.of([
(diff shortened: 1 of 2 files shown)


modules/nf-core/cat/fastq/tests/main.nf.test
```diff
@@ -120,18 +124,19 @@ nextflow_process {
             }
             process {
                 """
-                input[0] = Channel.of([
+                input[0] = [
```
maxulysse, line 127: why do you remove all the `Channel.of`?

asp8200: Just checking to see if you are awake

asp8200:
Not needed anyway

PR #4995, opened 2024-02-26
gatk4/applybqsr : Using modules_testdata_base_path instead of test_data in nf-test

asp8200:
Needed for Sarek.


maxulysse:
you don't need to update the snapshots if you just change the path to the files

modules/nf-core/gatk4/applybqsr/tests/main.nf.test.snap
edmundmiller, line 1: As @maxulysse said, let's remove this change.

edmundmiller (dismissed):

maxulysse (approved):

PR #5017, opened 2024-03-01
Update GATK_UG: change from gzip to bgzip

TCLamnidis:
Changes the compression used for the output VCF from gzip to bgzip. The latter is compatible with `bcftools`, which is the standard tool for using VCFs.


TCLamnidis:
`bgzip` is not included in the environment for `gatk`. I have added it to the conda environment definition, but will need to create a multi-tool-container for it now.

TCLamnidis:
Waiting on https://github.com/BioContainers/multi-package-containers/pull/2992

TCLamnidis:
@jfy133 @ilight1542 
This should be ready now.

nvnieuwk (approved):
LGTM

TCLamnidis:
tests seemed to pass, but the merge queue did not think so. updated barnch and trying again once i get the green ticks

PR #5053, opened 2024-03-05
add MAGUS

lrauschning:
Adds two modules for MAGUS exposing the guidetree and alignment functionality, the latter implementing the compression strategy used in other MSA packages..


modules/nf-core/magus/guidetree/main.nf
```diff
@@ -0,0 +1,48 @@
...
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    magus \\
+        -np $task.cpus \\
+        -i $fasta \\
+        -o ${prefix}.tree \\
+        --onlyguidetree TRUE \\
```
nvnieuwk, line 28: is `--onlyguidetree TRUE` a required option? Is there no use for the module without that flag?

nvnieuwk (commented):
Some comments, but overall good job!

modules/nf-core/magus/align/main.nf
```diff
@@ -0,0 +1,58 @@
...
+    path "versions.yml"                 , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def loadtree = tree ? "-t $tree" : ''
+    def write_output = compress ? "--overwrite -o >(pigz -cp ${task.cpus} > ${prefix}.aln.gz)" : "-o ${prefix}.aln"
```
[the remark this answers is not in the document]
lrauschning: I agree, that would be the ideal way to do it – this solution is a compromise, because many downstream tools don't support compressed files.
There was a fairly long thread discussing this in the slack:
https://nfcore.slack.com/archives/CJRH30T6V/p1706010248794839

modules/nf-core/magus/guidetree/main.nf
lrauschning: The guidetree module is made specifically to compute a guidetree with the align module handling the alignment, so I think there would not be many uses.
In theory the user could pass `--onlyguidetree FALSE` in ext.args, overwriting this I guess, but it would make more sense to just call the right module in the first place.

lrauschning:
As a general note, this is failing because MAGUS as of the current released version 0.1.3 does not work in containers, because at runtime some permissions in the filesystem are changed.
I'm working on repackaging MAGUS to alleviate this, but it has been slow going as I am not very familiar with the codebase and also busy with some other stuff.

modules/nf-core/magus/align/main.nf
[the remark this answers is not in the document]
nvnieuwk: Allright, fine by me :p

modules/nf-core/magus/guidetree/main.nf
nvnieuwk: I'm not sure splitting this is the best option since they are the same tool (not even a different subtool)

lrauschning:
As the snapshots seem to be identical in different runs on the same machine, but not across different machines, I've changed the tests to check that all the samples we expect to see are in there instead of snapshotting the file..

modules/nf-core/magus/guidetree/main.nf
lrauschning: Ah sorry, missed this earlier.
Maybe having the same functionality using a single module and multiple optional outputs would be cleaner.
This was done to maintain consistency with how we are handling it for other MSA tools that can do both, e.g. [ClustalO](https://github.com/nf-core/modules/tree/master/modules/nf-core/clustalo) or [FAMSA](https://github.com/nf-core/modules/tree/master/modules/nf-core/famsa).

luisas: That is a fair point, yet the guidetree has different input channels compared to the align module, that is why in the other modules we kept it separate (famsa and clustalo)

luisas (approved):

modules/nf-core/magus/guidetree/main.nf
SPPearce: This seems reasonable to keep separate in this case.

modules/nf-core/magus/align/main.nf
```diff
@@ -0,0 +1,58 @@
...
+    // using >() is necessary to preserve the return value,
+    // so nextflow knows to display an error when it failed
+    // using --overwrite is necessary, as the file descriptor generated by the named file will alerady exist
```
SPPearce, line 29:
```suggestion
    // using --overwrite is necessary, as the file descriptor generated by the named file will already exist
```

modules/nf-core/magus/align/main.nf
```diff
@@ -0,0 +1,58 @@
...
+        MAGUS: \$(magus --version)
+        pigz: \$(echo \$(pigz --version 2>&1) | sed 's/^.*pigz\\w*//' ))
+    END_VERSIONS
+    """
+
+    stub:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    touch ${prefix}.aln${compress ? '.gz' : ''}
```
SPPearce, line 50: nf-test doesn't like .gz files that haven't actually come from gzip, so this would technically fail if you tried to test the stub with compress. We have been using `echo "" | gzip` instead of `touch` instead.

SPPearce (approved):

modules/nf-core/magus/align/main.nf
lrauschning: Commited!

PR #5108, opened 2024-03-13
add paths in output directive in cellranger cout module

fmalmeida:
Update relates to requirement for PR in `nf-core/scrnaseq`: https://github.com/nf-core/scrnaseq/pull/301


grst (approved):
Given your reasoning in https://github.com/nf-core/scrnaseq/pull/301#issuecomment-1994257781, I guess why not. I also doesn't break existing workflows because the `outs` channel is unmodified.

nvnieuwk (approved):

PR #5113, opened 2024-03-13
add cram/index outputs to bwa/mem

matthdsm:
- Add cram and index outputs to bwa/mem
- Update all dependent subworflows
- Move everything to nf-test

## PR checklist

Closes #XXX 

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:


modules/nf-core/bwa/mem/main.nf
```diff
@@ -24,6 +28,9 @@ process BWA_MEM {
     def args2 = task.ext.args2 ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
     def samtools_command = sort_bam ? 'sort' : 'view'
+    def extension = args2.contains("--output-fmt sam") ? "sam" :
+                    args2.contains("--output-fmt cram") ? "cram" :
```
nvnieuwk, line 32: Maybe also check for the short form? `-O`

modules/nf-core/bwa/mem/main.nf
```diff
@@ -45,6 +52,7 @@ process BWA_MEM {
     def prefix = task.ext.prefix ?: "${meta.id}"
     """
     touch ${prefix}.bam
+    touch ${prefix}.csi
```
nvnieuwk, line 55: Can you make the stub output BAM/CRAM dependent on the `arg` given to samtools?

nvnieuwk (commented):
Some comments, but I like these changes! (I only reviewed `bwa/mem` for now)

matthdsm:
This is going to be a hard one to make tests pass 😓

modules/nf-core/bwa/mem/main.nf
```diff
@@ -32,7 +39,7 @@ process BWA_MEM {
         -t $task.cpus \\
         \$INDEX \\
         $reads \\
-        | samtools $samtools_command $args2 --threads $task.cpus -o ${prefix}.bam -
+        | samtools $samtools_command $args2 --reference ${fasta} --threads $task.cpus -o ${prefix}.${extension} -
```
pinin4fjords, line 42: Can we condition the provision of `--reference` on a) the extension and b) the fasta actually being provided? I feel like a backwards-compatible usage without fastas is important to retain. Will need an error is someone requests a cram without providing fasta though.

matthdsm: Done!

modules/nf-core/bwa/mem/main.nf
matthdsm: Done!

modules/nf-core/bwa/mem/main.nf
```diff
@@ -24,6 +28,13 @@ process BWA_MEM {
     def args2 = task.ext.args2 ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
     def samtools_command = sort_bam ? 'sort' : 'view'
+    def extension = args2.contains("--output-fmt sam")   ? "sam" :
+                    args2.contains("--output-fmt cram")  ? "cram":
+                    sort_bam && args2.contains("-O cram")? "cram":
+                    !sort_bam && args2.contains("-C")    ? "cram":
+                    "bam"
```
nvnieuwk, line 35: What about `-O bam` and I think the extension can also be uppercase, right? (Sorry for all the nitpicking ehehe :grin:)

pinin4fjords (commented):
Looks pretty good to me- just need a test to make sure things still work without the reference fasta

modules/nf-core/bwa/mem/tests/main.nf.test
nvnieuwk, line 1: Please also add stub tests :)

modules/nf-core/bwa/mem/main.nf
matthdsm: Anything not cram defaults to bam, so no need to check

nvnieuwk: Isn't the default SAM?

matthdsm: samtools infers the bam format from the `-o ${prefix}.bam`

modules/nf-core/bwa/mem/tests/main.nf.test
matthdsm: done!

matthdsm:
> Looks pretty good to me- just need a test to make sure things still work without the reference fasta

@pinin4fjords , added a 'nofasta' test

modules/nf-core/bwa/mem/tests/main.nf.test
```diff
@@ -170,4 +174,103 @@ nextflow_process {
...
+                """
+                input[0] = [
+                    [ id:'test', single_end:false ], // meta map
+                    [
+                        file(params.modules_testdata_base_path + 'genomics/sarscov2/illumina/fastq/test_1.fastq.gz', checkIfExists: true),
+                        file(params.modules_testdata_base_path + 'genomics/sarscov2/illumina/fastq/test_2.fastq.gz', checkIfExists: true)
+                    ]
+                ]
+                input[1] = BWA_INDEX.out.index
+                input[2] = [[id: 'test'],[]]
```
pinin4fjords, line 205:
```suggestion
                input[2] = [[:],[]]
```

modules/nf-core/bwa/mem/main.nf
```diff
@@ -8,13 +8,17 @@ process BWA_MEM {
...
-    tuple val(meta), path(reads)
+    tuple val(meta) , path(reads)
     tuple val(meta2), path(index)
+    tuple val(meta3), path(fasta)
     val   sort_bam
 
     output:
-    tuple val(meta), path("*.bam"), emit: bam
-    path  "versions.yml"          , emit: versions
+    tuple val(meta), path("*.bam")  , emit: bam,    optional: true
```
pinin4fjords, line 17: Sorry, to be really annoying and add another comment, does it make sense to have channels for output bam AND cram? Only one or the other is produced, right?

Could we not just have one non-optionoal 'alignment' output, and one 'index', globbing for multiple extensions? Or will that create difficulties?

pinin4fjords (commented):
Sorry to be extra annoying, but since this is defining a pattern I think it's OK to ask!

modules/nf-core/bwa/mem/tests/main.nf.test
matthdsm: I can agree with this one. I don't think it really matter 🤷🏻

modules/nf-core/bwa/mem/main.nf
matthdsm: It's against the guidelines 😄 
I kind of like having more granular control over which outputs I'm using TBH, that way I can fail faster if something goes wrong with the module config.

e.g, the output will always be bam, unless I configure it otherwise. With the global glob I'll always get some form of output, but not necessarily the one I want.

matthdsm: https://nf-co.re/docs/contributing/modules#inputoutput-options

```
3. Named file extensions MUST be emitted for ALL output channels e.g. path "*.txt", emit: txt.
```

pinin4fjords: Ahh, fair enuf

modules/nf-core/bwa/mem/tests/main.nf.test
pinin4fjords: Yep, just wanted to make sure the meta content wasn't required for anything. I'll stop picking nits now

modules/nf-core/bwa/mem/main.nf
pinin4fjords: Still, this means that if I were to switch alignment format in config I'd have to rewire the workflow.

matthdsm: Yes it does

pinin4fjords (approved):
OK, like it, and @maxulysse points out that separate bam/ cram channels are convention elsewhere in nf-core so I'll stop over-thinking.

asp8200:
Looks good to me. Just waiting for tests to pass before approving.

modules/nf-core/bwa/mem/main.nf
muffato: My view is that the rule is there to prevent the `*` pattern. In any case it doesn't forbid bundling multiple file extensions.
It's not just samtools, I see a lot of tools transparently support BAM/CRAM, or fasta/fastq/fasta.gz/fastq.gz, etc, and we generally don't list every possible file extension as its own channel.

edmundmiller:
Will the `optional` blow up on us? What does the error look like if there's no bam?

I'm wondering if we could do something like 

```
  input:
    tuple val(meta) , path(reads)
    tuple val(meta2), path(index)
    tuple val(meta3), path(fasta)
    val   sort_bam
    val   output_type

    output:
    tuple val(meta), path("*.${output_type}")  , emit: aligned_reads
    tuple val(meta), path("*.${fancy logic to set bai/csi") , emit: index
    path  "versions.yml"            , emit: versions
```

modules/nf-core/bwa/mem/main.nf
edmundmiller: Will the `optional` blow up on us? What does the error look like if there's no bam?

I'm wondering if we could do something like 

```
  input:
    tuple val(meta) , path(reads)
    tuple val(meta2), path(index)
    tuple val(meta3), path(fasta)
    val   sort_bam
    val   output_type

    output:
    tuple val(meta), path("*.${output_type}")  , emit: aligned_reads
    tuple val(meta), path("*.${fancy logic to set bai/csi") , emit: index
    path  "versions.yml"            , emit: versions
```

matthdsm: How do you mean blow up? If there's no bam or cram it means the tool failed and we should get an exit code corresponding to that.

I've made modules in the past with the fancy logic you suggest, but I got pulled back because it was against the guidelines 🤷🏻

maxulysse (approved):

matthdsm:
Thanks for all the helpful comments and suggestions everyone! 😄  and special thanks to @pinin4fjords and @maxulysse for the approvals!

PR #5171, opened 2024-03-18
Add wittyer as module

### LINKED ISSUE #4739: new module: wittyer

famosab:
## PR checklist

Closes #4739 , relates to https://github.com/nf-core/variantbenchmarking/issues/10

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:
---
According to [slack discussion](https://nfcore.slack.com/archives/CE6SDBX2A/p1709543357582789), the docker image still needs to be added to the nf-core registry.


modules/nf-core/wittyer/main.nf
```diff
@@ -0,0 +1,94 @@
...
+    wittyer \\
+        --truthVcf=${truth_vcf} \\
+        --inputVcf=${query_vcf} \\
+        --outputDirectory=bench \\
+        ${regions} \\
+        ${args}
+
+    mv bench/Wittyer.Stats.json ${prefix}.Stats.json
+    mv bench/*.vcf.gz ${prefix}.eval.vcf.gz
+    mv bench/*.vcf.gz.tbi ${prefix}.eval.vcf.gz.tbi
```
nvnieuwk, line 69: Could you remove any custom extensions that are not file extensions? This should be done to make all filenames as customizable as possible :)

modules/nf-core/wittyer/main.nf
```diff
@@ -0,0 +1,94 @@
...
+
+    stub:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    // TODO nf-core: A stub section should mimic the execution of the original module as best as possible
+    //               Have a look at the following examples:
+    //               Simple example: https://github.com/nf-core/modules/blob/818474a292b4860ae8ff88e149fbcda68814114d/modules/nf-core/bcftools/annotate/main.nf#L47-L63
+    //               Complex example: https://github.com/nf-core/modules/blob/818474a292b4860ae8ff88e149fbcda68814114d/modules/nf-core/bedtools/split/main.nf#L38-L54
+    """
+    touch ${prefix}.bam
```
nvnieuwk, line 87: Please touch all output files here too :)

nvnieuwk (commented):
It's looking very promising already!

modules/nf-core/wittyer/main.nf
famosab: I think `eval` was added to show that this vcf is the result of the benchmarking comparison. Should that be captured via `prefix` then?

nvnieuwk: Yes that should be added via the `prefix`. It could also be a good idea to add a simple check to see if the output filename isn't the same as the input files (otherwise Nextflow will give some weird errors) => See https://nf-co.re/docs/contributing/modules#naming-conventions part 6 for an example

kubranarci: but the name is generated through wittyer, I think it makes sense to keep original extensions.

kubranarci: _eval_ is already exist in original output, only prefix is being added here to the name. I would prefer it in this way actually.

nvnieuwk: I understand, but you lose the flexibility to call the file whatever you want. You can still add `eval` to the output name yourself if you want that :) (I'm just reciting what's in the guidelines, so don't shoot the messenger :wink:)

modules/nf-core/wittyer/main.nf
```diff
@@ -0,0 +1,69 @@
+process WITTYER {
+    tag "$meta.id"
+    label 'process_single'
+
+    container "quay.io/nf-core/wittier:0.3.3"
```
famosab, line 5: Here I am unsure if that is the correct "address" to get the container.

modules/nf-core/wittyer/main.nf
```diff
@@ -0,0 +1,69 @@
...
+    """
+    script:
+    if ("$truth_vcf" == "${prefix}.vcf.gz")         error "Input and output names are the same, set prefix in module configuration to disambiguate!"
+    if ("$query_vcf" == "${prefix}.vcf.gz")         error "Input and output names are the same, set prefix in module configuration to disambiguate!"
+    if ("$truth_vcf_tbi" == "${prefix}.vcf.gz.tbi") error "Input and output names are the same, set prefix in module configuration to disambiguate!"
+    if ("$query_vcf_tbi" == "${prefix}.vcf.gz.tbi") error "Input and output names are the same, set prefix in module configuration to disambiguate!"
+    
+    mkdir bench
+
+    wittyer \\
```
famosab, line 37: When I try it locally I always have to run
```bash
docker run wittyer
```
but I do not really know how it works with nf-core. In the `bcl2fastq` module they just call the program.

modules/nf-core/wittyer/tests/main.nf.test
```diff
@@ -0,0 +1,57 @@
...
+                    file(params.test_data['homo_sapiens']['illumina']['simulated_sv'], checkIfExists: true)
+                    ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
```
famosab, line 27: Do we need more assertions here?

modules/nf-core/wittyer/environment.yml
```diff
@@ -0,0 +1,5 @@
+name: wittyer
+channels:
+  - conda-forge
+  - bioconda
+  - defaults
```
famosab, line 5: I removed the dependencies analogous to `bcl2fastq` but that leads to a linting error.

modules/nf-core/wittyer/main.nf
kubranarci: But I was not adding the extension by myself :D it is the standard output by the tool.

modules/nf-core/wittyer/meta.yml
```diff
@@ -0,0 +1,66 @@
...
+  - structural-variants
+  - benchmarking
+  - vcf
+tools:
+  - "wittyer":
+      description: "Illumina tool for large variant benchmarking"
+      homepage: "https://github.com/Illumina/witty.er"
+      documentation: "https://github.com/Illumina/witty.er"
+      tool_dev_url: "https://github.com/Illumina/witty.er"
+      licence: "['BSD-2']"
```
famosab, line 14: This is the licence that is indicated on the witty.er page also somehow throws a linting error.

modules/nf-core/wittyer/main.nf
kubranarci: This type of names are important for traceability with standard outputs of the tool. So I would really prefer to keep is as it is.

modules/nf-core/wittyer/main.nf
kubranarci: remove quay.io

modules/nf-core/wittyer/tests/main.nf.test
famosab, line 1: `nf-core modules test wittyer` fails at the moment, probably because the wittyer image is not public yet.

modules/nf-core/wittyer/main.nf
kubranarci: in your local you can run as:

> docker run  quay.io/nf-core/wittyer:0.3.3.0


but in the process it should be like:

> nf-core/wittyer:0.3.3.0

modules/nf-core/wittyer/meta.yml
kubranarci: licence: ['BSD-2']

modules/nf-core/wittyer/tests/main.nf.test
kubranarci: yes, per output

modules/nf-core/wittyer/main.nf
famosab: But the `eval` is only added when you move the file from the output directory?

kubranarci: no, I am copying the result here. eval is in standard output.

modules/nf-core/wittyer/tests/main.nf.test
famosab: Ok I always get
`error [nextflow.exception.ProcessUnrecoverableException]: Path value cannot be null`

modules/nf-core/wittyer/main.nf
```diff
@@ -0,0 +1,69 @@
...
+        --truthVcf=${truth_vcf} \\
+        --inputVcf=${query_vcf} \\
+        --outputDirectory=bench \\
+        ${regions} \\
+        ${args}
+
+    mv bench/Wittyer.Stats.json ${prefix}.json
+    mv bench/*.vcf.gz ${prefix}.vcf.gz
```
kubranarci, line 45: mv bench/*eval.vcf.gz ${prefix}.eval.vcf.gz
mv bench/*eval.vcf.gz.tbi ${prefix}.eval.vcf.gz.tbi

famosab:
@nf-core-bot fix linting

modules/nf-core/wittyer/environment.yml
famosab: I now added `tabix` as a dependency but there should be a better solution.

famosab:
@nf-core-bot fix linting

famosab:
For some reason the stub tests are always failing. I've already fixed the `.gz` issue but I still run into issues with the version thing?

modules/nf-core/wittyer/tests/main.nf.test
```diff
@@ -0,0 +1,65 @@
...
+                    []
+                    ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out.version).match("version") }
```
nvnieuwk, line 59:
```suggestion
                { assert snapshot(process.out.version).match("stub_version") }
```

This should fix that error. You had two snapshots with the same name, nf-test doesn't like that very much

modules/nf-core/wittyer/main.nf
```diff
@@ -0,0 +1,70 @@
...
+    container "nf-core/wittyer:0.3.3.0"
+
+    // Exit if running this module with -profile conda / -profile mamba
+    if (workflow.profile.tokenize(',').intersect(['conda', 'mamba']).size() >= 1) {
+        error "WITTYER module does not support Conda. Please use Docker / Singularity / Podman instead."
+    }
+
+    input:
+    // use stageAs to allow comparing the same vcfs against each other
+    tuple val(meta), path(query_vcf, stageAs: "query.vcf.gz"), path(query_vcf_index, stageAs: "query.vcf.gz.tbi"), path(truth_vcf, stageAs: "truth.vcf.gz"), path(truth_vcf_index, stageAs: "truth.vcf.gz.tbi"), path(bed)
```
nvnieuwk, line 14: Don't use `stageAs` here, it's easier to debug modules if they retain the original filenames :)

nvnieuwk (commented):
Almost there!

modules/nf-core/wittyer/main.nf
famosab: I wanted to test truth and query with the same files. Because I find it easier to check if my benchmarking workflow works when I can compare a file against itself. So if I remove it here it would need to give them different names?

nvnieuwk: Yeah but at the cost of making it harder on yourself when you want to actually use the process? Can't you make a copy of the query file and rename it before running the pipeline? :p

famosab: I'll change it  and adapt the test accordingly

nvnieuwk:
Can you also exclude the conda test in `.github/workflows/test.yml`? You'll find a list of modules that have their conda test excluded. Please add this module there so you won't merge any failing tests :)

nvnieuwk (approved):
LGTM!

PR #5215, opened 2024-03-19
Add nf-test for md5sum module

### LINKED ISSUE #5159: [FEATURE] nf-test for md5sum

port pytest to nf-test for md5sum module

_No response_

_No response_

### Additional context

_No response_

human9:
Checks that the correct filename and contents are created. A custom config file is used to test extra args.

## PR checklist

Closes #5159 

- [X] `nf-test test modules/nf-core/md5sum/tests/main.nf.test --profile docker`


modules/nf-core/md5sum/tests/main.nf.test
```diff
@@ -0,0 +1,73 @@
...
+    tag "modules"
+    tag "modules_nfcore"
+    tag "md5sum"
+
+    test("md5sum on hello.txt") {
+
+        when {
+            process {
+                """
+                input[0] = Channel.of([
```
vagkaratzas, line 16: is the Channel.of part needed here and on the next test? Never seen it in nf-tests

modules/nf-core/md5sum/tests/main.nf.test
```diff
@@ -0,0 +1,73 @@
...
+                    [ id: 'hello' ],
+                    [ file(params.modules_testdata_base_path + 'generic/txt/hello.txt', checkIfExists: true) ]
+                ])
+                """
+            }
+        }
+
+        then {
+            with(process.out) {
```
vagkaratzas, line 25: assertAll(
                { assert process.success },
            )
            
Should be working just fine for this module, no? (same for next test hopefully)
Also need a versions snapshot, and these two lines (if working properly) will cover that.

vagkaratzas (dismissed):

modules/nf-core/md5sum/tests/main.nf.test
human9: You're right, that's much simpler! This was my first attempt to use nf-test so I may have gotten ahead of myself...

human9:
Hmm, I don't understand why the conda test failed...

SPPearce (approved):

SPPearce:
> Hmm, I don't understand why the conda test failed...

The conda test was failing because conda and docker/singularity were using different versions.
I've downgraded the conda version to match the ubuntu base (by copying what was used in another module).

PR #5048, opened 2024-03-04
Fix bakta: amrfinder database issue

jasmezz:
This fixes a crash of bakta when using the automatically downloaded database with bakta (reported [here](https://github.com/nf-core/modules/pull/4847#discussion_r1499767614)).
It does not seem to be a conda-only issue. I could reproduce it with singularity as well, that's why I execute the `amrfinder_update` line regard of the profile.

(The memory requirement in the config file was decreased only because GHA complained.)

## PR checklist

- [x] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [ ] Remove all TODO statements.
- [ ] Emit the `versions.yml` file.
- [ ] Follow the naming conventions.
- [ ] Follow the parameters requirements.
- [ ] Follow the input/output options guidelines.
- [ ] Add a resource `label`
- [ ] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - For modules:
    - [ ] `nf-core modules test <MODULE> --profile docker`
    - [x] `nf-core modules test <MODULE> --profile singularity`
    - [x] `nf-core modules test <MODULE> --profile conda`
  - For subworkflows:
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile docker`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile singularity`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile conda`

PR #5232, opened 2024-03-19
Remove AMRFinderPlus DB update on each invocation

oschwengers:
I removed the AMRFinderPlus DB update from the Bakta `main.nf` which seems to be invoked on each execution of Bakta. AMRFinderPlus db should be downloaded and installed automatically and transparently during the Bakta DB download via `bakta_db download` -> https://github.com/oschwengers/bakta/blob/d6443639958750c3bece5822e84978271d1a4dc7/bakta/db.py#L247

Maybe I'm missing something, but if not, this would be a waste of bandwidth and significantly increase wall clock run times.


jfy133:
I think it was @jasmezz who added this in because the module wasn't working without it IIRC ?

jasmezz:
Yes, the amrfinder update line comes from [here](https://github.com/nf-core/modules/pull/5048) (see also the original bug report in the PR description).
In short: The user had bakta crashing using conda without the amrfinder update line, which I could reproduce. So I wanted to add it for conda-only executions to save resources for non-conda usage. For some reason, bakta started to crash then for the singularity/docker containers as well without the line. Thus, I added it for all.
And now in your PR it seems to work again without the line! I am really confused about this flakiness.
Did you test your changes for conda as well locally?

oschwengers:
This is really interesting/odd. For Bakta we often see issues with AMRFinderPlus that we cannot reproduce or don't know what is causing these.

No, I haven't tested it locally. I just recognized this line and wanted to throw in my 0.02$ as a tool developer that, in principle, this should work w/o this extra invocation of `amrfinder_update`.

jasmezz:
That's alright, thank you! I'd prefer to approve your PR of course. I am investigating the issue locally with conda. Stay tuned, this might take a while.

jasmezz:
I cannot find any errors with removing the line on any of conda/singularity/docker. So let's go ahead with this PR. :+1: 
I did not really find the source from the earlier reported issue which lead to including the line in the first place. Hopefully I get better clues if it ever happens again (which will of course not happen!).

jasmezz (approved):

jasmezz:
Go ahead and merge @oschwengers :rocket: I need the module asap, so will need to do merge myself if you won't ;)

oschwengers:
Thanks a lot @jasmezz! I'd love to merge this but, unfortunately, I do not have write permissions.

PR #5238, opened 2024-03-19
5236 update cellrangermulti output directives to match work in nf core scrnaseq

fmalmeida:
Work related to https://github.com/nf-core/scrnaseq/pull/276
Adding two new output directives so it can be seamlessly coupled in scrnaseq.

### DIFF
--- a/modules/nf-core/cellranger/multi/main.nf
+++ b/modules/nf-core/cellranger/multi/main.nf
@@ -27,9 +27,11 @@ process CELLRANGER_MULTI {
     val skip_renaming
 
     output:
-    tuple val(meta), path("cellranger_multi_config.csv"), emit: config
-    tuple val(meta), path("**/outs/**")                 , emit: outs
-    path "versions.yml"                                 , emit: versions
+    tuple val(meta), path("cellranger_multi_config.csv")            , emit: config
+    tuple val(meta), path("**/outs/**")                             , emit: outs
+    tuple val(meta), path("**/outs/**/filtered_feature_bc_matrix**"), emit: filtered, optional: true
+    tuple val(meta), path("**/outs/**/raw_feature_bc_matrix**")     , emit: raw
+    path "versions.yml"                                             , emit: versions
 
     when:
     task.ext.when == null || task.ext.when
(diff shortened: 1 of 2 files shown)


grst (approved):

pinin4fjords (commented):
I'm not convinced about ouputing overlapping outputs. IMO we should either output the directory as a single emission or ALL of the output separately- this half way point seems like a hack.

FriederikeHanssen:
@pinin4fjords good point removed from queue to discuss this first

fmalmeida:
The output `outs` gets all the outputs in a file manner as an array. The downstream module needs to use the complete folder structure of the filtered / raw results, separately.

If I just try to apply `.filter()` in the `outs` channels, I get the files and not the dirs.

Any suggestion?

pinin4fjords:
I see this has already been done to [count](https://github.com/nf-core/modules/blob/master/modules/nf-core/cellranger/count/main.nf)

Again, we should probably have output channels for all the outputs in `out/`, or use the existing single emission.

pinin4fjords:
> The output `outs` gets all the outputs in a file manner as an array. The downstream module needs to use the complete folder structure of the filtered / raw results, separately.
> 
> If I just try to apply `.filter()` in the `outs` channels, I get the files and not the dirs.
> 
> Any suggestion?

Maybe filter out the directory you want, collect with groupBy(), and then adapt the consuming process? You can use options on the input to collect the files into a dir: https://www.nextflow.io/docs/latest/process.html#multiple-input-files

fmalmeida:
Found a way of doing without it. Closing it.
Can remove later from cellranger/count.

fmalmeida:
Here the code in the pipeline:

```
//
        // Split channels of raw and filtered to avoid file collision problems when loading the inputs in conversion modules.
        //
        ch_matrices_raw =
        CELLRANGER_MULTI.out.outs.map { meta, mtx_files ->
            def desired_files = []
            mtx_files.each{
                if ( it.toString().contains("raw_feature_bc_matrix") ) { desired_files.add( it ) }
            }
            [ meta, desired_files ]
        }

        ch_matrices_filtered =
        CELLRANGER_MULTI.out.outs.map { meta, mtx_files ->
            def desired_files = []
            mtx_files.each{
                if ( it.toString().contains("filtered_feature_bc_matrix") ) { desired_files.add( it ) }
            }
            [ meta, desired_files ]
```

Also opened PR-reversions for the other ones.

PR #5245, opened 2024-03-19
Dp24 pretextmap update

### LINKED ISSUE #5123: [Update] PretextMap should take a fai file

PretextMap will index the incoming fasta file, in many cases this file will already be generated and just needs to be visible to the process (it cannot be declared to the software).

In the cases of large genomes (I have a 25Gb Newt!) this will speed up the mapping process and waste less resources as we are not indexing again.

Solution is proposed here https://github.com/sanger-tol/treeval/pull/283

This changes the fasta file path into a tuple containing the meta, fasta and fai.

_No response_

### Additional context

This will require the module to also be updated to use NF-Test.

The module will not be updated to use the sanger container shown in the linked issue, this is temporary until we can access the main repo to update it properly and includes only an extra flag for buffer_memory manipulation.

DLBPointon:
Adding nf-test tests and files

Adding the option to also include a fai file. If this is not provided then pretext will generate one automatically.
Closes #5123


modules/nf-core/pretextmap/meta.yml
```diff
@@ -20,6 +20,19 @@ input:
...
+        Groovy Map containing sample information
+        e.g. [ iid:'test2', single_end:false ]
+  - input:
+      type: file
+      description: FASTA file
+      pattern: "*.{fasta,fa}"
+  - input:
+      type: file
+      description: Fasta index
+      pattern: "*fai"
```
lrauschning, line 35: Think that should be .fai?

modules/nf-core/pretextmap/tests/main.nf.test
lrauschning, line 1: Think testing the behaviour when not passing a `.fai` file would also be good, to ensure there are no future regressions.
Can just be copied from one of the already existing tests while deleting the `.fai` input IMO.

modules/nf-core/pretextmap/meta.yml
DLBPointon: You are right, i'll change that now

modules/nf-core/pretextmap/tests/main.nf.test
DLBPointon: No, you are right, I should include a cram + no_fai test at least. Just me being lazy. Testing it now. Thanks for reviewing!

yumisims:
look good to me

modules/nf-core/pretextmap/meta.yml
```diff
@@ -20,6 +20,19 @@ input:
       type: file
       description: BAM/CRAM/SAM file or pairs formatted reads file
       pattern: "*.{bam,cram,sam,pairs.gz}"
+  - input:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. [ iid:'test2', single_end:false ]
```
marrip, line 27: typo?
```suggestion
        e.g. [ id:'test2', single_end:false ]
```

modules/nf-core/pretextmap/meta.yml
```diff
@@ -20,6 +20,19 @@ input:
...
+  - input:
+      type: file
+      description: FASTA file
+      pattern: "*.{fasta,fa}"
```
marrip, line 31: I also see `.fna` sometimes

marrip (approved):
nice work, LGTM! 👍

modules/nf-core/pretextmap/meta.yml
DLBPointon: Always at least 1, thanks.

modules/nf-core/pretextmap/meta.yml
DLBPointon: Adding now

PR #5277, opened 2024-03-19
Update kalign module to work with compressed files.

lrauschning:
Add the compression handling introduced to MSA modules in https://github.com/nf-core/modules/pull/4754 to kalign (forgot learnmsa and kalign it in that PR).


itrujnara (approved):
Looks fine, free to merge

luisas (changes requested):
we need to fix before the fact that the snapshot was not recreated for some reason

luisas (approved):

PR #5293, opened 2024-03-19
nf-test for pilon

thanhleviet:
replaced pytest with nf-test for pilon


telatin:
I think `pytest` should be removed now. Quoting J F Yates:
> Yes, if you add nf-test, remove pytest. The -m8grate-pytest` flag should do that for you if you do it via Nf-core tools

modules/nf-core/pilon/tests/main.nf.test
```diff
@@ -0,0 +1,38 @@
...
+    name "Test Process PILON"
+    script "../main.nf"
+    process "PILON"
+
+    tag "modules"
+    tag "modules_nfcore"
+    tag "pilon"
+
+
+    test("Test PILON") {
```
itrujnara, line 12: Change test name to something more specific, there is no standard (yet) but something like `pilon - bam - paired` should do

modules/nf-core/pilon/tests/main.nf.test
```diff
@@ -0,0 +1,38 @@
...
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
+            )
+        }
+
+    }
```
itrujnara, line 36: I see that the module has multiple modes, is it possible to add tests for the others? And a stub test as well, although I see the module currently lacks a stub

itrujnara (commented):
The test seems fine, it just needs a better name. More tests could be useful if possible.

modules/nf-core/pilon/tests/main.nf.test
```diff
@@ -0,0 +1,38 @@
...
+    tag "pilon"
+
+
+    test("Test PILON") {
+
+        when {
+            process {
+                """
+                input[0] = [ [ id:'test', single_end:false ], // meta map
+                            file(params.test_data['homo_sapiens']['genome']['genome_fasta'], checkIfExists: true)
```
itrujnara, line 18: Apparently the new guideline is to use URLs instead of the test_data object, so please change the tests accordingly

itrujnara (commented):
Added some new info on test files

modules/nf-core/pilon/tests/main.nf.test
telatin: Hi, I just started so can you point to the guidelines? I found some info (but not technical details) [here](https://nf-co.re/docs/contributing/test_data_guidelines)?

SPPearce: Changing to the test-data path is not required currently, just optional.

SPPearce:
Please add a stub and test for the stub as mentioned by @itrujnara

SPPearce (dismissed):
Please add a stub & test.

modules/nf-core/pilon/main.nf
```diff
@@ -40,4 +40,17 @@ process PILON {
     "${task.process}":
         pilon: \$(echo \$(pilon --version) | sed 's/^.*version //; s/ .*\$//' )
```
itrujnara, line 41: END_VERSIONS is missing, probably not breaking anything but it looks weird

modules/nf-core/pilon/main.nf
```diff
@@ -40,4 +40,17 @@ process PILON {
...
+    stub:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def valid_mode = ["frags", "jumps", "unpaired", "bam"]
+    if ( !valid_mode.contains(pilon_mode) )  { error "Unrecognised mode to run Pilon. Options: ${valid_mode.join(', ')}" }
+    """
+    touch ${prefix}.fasta
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        pilon: \$(echo \$(pilon --version) | sed 's/^.*version //; s/ .*\$//' )
```
itrujnara, line 53: As above

modules/nf-core/pilon/main.nf
```diff
@@ -40,4 +40,17 @@ process PILON {
     "${task.process}":
         pilon: \$(echo \$(pilon --version) | sed 's/^.*version //; s/ .*\$//' )
     """
+
+    stub:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def valid_mode = ["frags", "jumps", "unpaired", "bam"]
```
itrujnara, line 46: Would be great to align equal signs, same goes for the original module

itrujnara (commented):
Thanks for the fixes, it looks much better now. I have added some picky aesthetic comments. Sorry for the delay, I got stuck working on urgent stuff in my lab.

modules/nf-core/pilon/main.nf
SPPearce: I'd have expected it to fail at that point! Will fix it when I'm back on my computer

modules/nf-core/pilon/main.nf
```diff
@@ -39,5 +39,20 @@ process PILON {
...
         pilon: \$(echo \$(pilon --version) | sed 's/^.*version //; s/ .*\$//' )
+    END_VERSIONS
     """
+
+    stub:
+    def prefix     = task.ext.prefix ?: "${meta.id}"
+    def valid_mode = ["frags", "jumps", "unpaired", "bam"]
+    if ( !valid_mode.contains(pilon_mode) )  { error "Unrecognised mode to run Pilon. Options: ${valid_mode.join(', ')}" }
+    """
+    touch ${prefix}.fasta
```
jfy133, line 50: I would suggest emitting all files even if optional, this would allow easier chaining of modules in a subwf/pipeline stub run, if a downstream module wants one of the optional outputs

jfy133 (approved):
One optional change

PR #5316, opened 2024-03-20
Leviosam2 index

lgrochowalski:
## PR checklist

Closes partially #1582 

In this PR I added the index command of liviosam2, there is also lift that should be added.


mauro-saporita (approved):
looks good

Issue #5142, opened 2024-03-18
Update module: propr and propd to produce an optional output for GSEA/GREA


suzannejin:
Optionally, when one wants to do GSEA or GREA aferwards, produce the output needed for it.
So an adjacency matrix for GREA,
and a gene list for GSEA.

BirkirRey:
Hi Suzanne,

Could you provide some more information on GREA?
What kind of analysis is it?
What kind of adjacency matrix is required?

Best regards,
Birkir

suzannejin:
Hey @BirkirRey! So sorry, I did not see your comment. I will check your pull request tomorrow morning
Cheers

PR #5355, opened 2024-03-20
Adressing issue 5142: propr/propr and propr/propd optionally output to GSEA

BirkirRey:
main.nf files and template .R files updated with optional input parameter so that both modules[propr/propr and propr/propd] can output files for GEAS analysis [.gct and .cls]. The issue also specified an interest in generating an adjacency matrix for GREA analysis, but the details on that were unclear so I have not completed that.

## PR checklist

Closes [#5142 ](https://github.com/nf-core/modules/issues/5142) 

  - For modules:


modules/nf-core/propr/propd/main.nf
```diff
@@ -10,17 +10,21 @@ process PROPR_PROPD {
     input:
     tuple val(meta), path(count)
     tuple val(meta2), path(samplesheet)
+    val GSEA
```
suzannejin, line 13: I think in this case lower case is preferred.
Also, only the very needed input should stay here. This kind of optional input should stay in ext.args.
If you are not used to ext.args, check https://github.com/nf-core/differentialabundance/blob/master/conf/modules.config
You can also check how I call ext.args inside the test configs.

modules/nf-core/propr/propd/main.nf
```diff
@@ -10,17 +10,21 @@ process PROPR_PROPD {
...
 
     output:
     tuple val(meta), path("*.propd.rds"), emit: propd
     tuple val(meta), path("*.propd.tsv"), emit: results
     tuple val(meta), path("*.fdr.tsv")  , emit: fdr         , optional:true
+    tuple val(meta), path("*.gct")  , emit: gct         , optional:true
```
suzannejin, line 19: For readability, you have to have the different elements of different outputs at same columns.
If you check the emit in session_info and versions, it follows a lot of spaces to be in the same place as the emits for propd, results and fdr. The same for optional.

modules/nf-core/propr/propd/main.nf
```diff
@@ -10,17 +10,21 @@ process PROPR_PROPD {
...
+    tuple val(meta), path("*.cls")  , emit: cls         , optional:true
     path "*.R_sessionInfo.log"          , emit: session_info
     path "versions.yml"                 , emit: versions
 
     when:
     task.ext.when == null || task.ext.when
 
     script:
+    GSEAopt = GSEA.toString().toUpperCase() // R expects upper case
```
suzannejin, line 28: You can do this kind of small processings inside the template.
Or maybe you can modify the R code so that it will not expect upper case.

modules/nf-core/propr/propd/templates/propd.R
```diff
@@ -63,6 +63,98 @@ read_delim_flexible <- function(file, header = TRUE, row.names = 1, check.names
...
+  
+  mat <- read.delim(
+    file,
+    sep = separator,
+    header = header,
+    row.names = row.names,
+    check.names = check.names
+  )
+  
+  return(mat)
```
suzannejin, line 87: Hey, actually it might be great if the input matrix for this module is already formatted so that the first column has the gene ids, and then the rest of columns are just the samples.
In this way you can convert read_delim_flexible_raw into read_delim_flexible, and remove the original read_delim_flexible.
If you do so, you may have to do a bit of groovy inside the test `when` section, so that the second column of the test data is removed before running the module.

modules/nf-core/propr/propd/main.nf
suzannejin: Sorry I should have specify it in the issue. There is a CUSTOM_TABULARTOGSEACLS module already implemented in nf-core. So we don't need to produce cls file from propr module.
What it is still nice that the GCT file you produce will be based on the ranked differential proportionality values/ p-values /etc.

modules/nf-core/propr/propd/templates/propd.R
```diff
@@ -63,6 +63,98 @@ read_delim_flexible <- function(file, header = TRUE, row.names = 1, check.names
...
+    col.names = TRUE,
+    row.names = FALSE,
+    sep       = '\t',
+    quote     = FALSE,
+    append=TRUE
+  )
+  return(samples)
+}
+
+writeCLSFile <-  function(samples,sampleFile,clsFile){
```
suzannejin, line 133: see above, we don't need cls

modules/nf-core/propr/propd/templates/propd.R
```diff
@@ -82,7 +174,8 @@ opt <- list(
     cutoff_min      = NA,                   # minimun threshold to test
     cutoff_max      = NA,                   # maximun threshold to test
     cutoff_interval = NA,                   # interval between thresholds
-    ncores          = as.integer('$task.cpus')
+    ncores          = as.integer('$task.cpus'),
+    GSEA            = '$GSEAopt'
```
suzannejin, line 178: we can set gsea as false in default, and then leave it parse through ext.args below in the parameter overrride section

modules/nf-core/propr/propd/main.nf
BirkirRey: Fair enough, I will remove the optional boolean from the propd module and just leave it in the ext.args with False as default.

modules/nf-core/propr/propd/main.nf
BirkirRey: I have aligned the commas. Regarding the CLS files, I was not aware of the existing module, so I will just remove the writeCLS functions from the template

modules/nf-core/propr/propd/main.nf
BirkirRey: This line has been removed from the code, since this parameter is now defined in the ext.args.

modules/nf-core/propr/propd/templates/propd.R
BirkirRey: I can see your point, since read_delim_flexible_raw is highly similar to read_delim_flexible. But I don't think the benefit outweighs complicating the module code. Another approach would be conditional logic in the read_delim_flexible function to switch between the two behaviors. But I will not make this update.

modules/nf-core/propr/propd/templates/propd.R
BirkirRey: I have removed it from both template files, propr.R and propd.R

modules/nf-core/propr/propd/templates/propd.R
BirkirRey: I have done that now

BirkirRey:
I have now pushed more commits that address the comments of @suzannejin and some failed tests during the initial commit

suzannejin:
> I have now pushed more commits that address the comments of @suzannejin and some failed tests during the initial commit

I would recommend you to update branch first, so that github CLI will only tests on the modules you modified, instead of all modules.

modules/nf-core/propr/propr/main.nf
```diff
@@ -8,18 +8,23 @@ process PROPR_PROPR {
...
+    tuple val(meta), path("*.gct"),	  emit: gct	    , optional:true
+    tuple val(meta), path("*.cls"),	  emit: cls	    , optional:true
     path "*.R_sessionInfo.log",           emit: session_info
     path "versions.yml",                  emit: versions
 
     when:
     task.ext.when == null || task.ext.when
 
     script:
+    sample_file = (opt_file) ? "$opt_file" : 'EMPTY'
```
suzannejin, line 28: what is the sample_file for? If it only sets a condition, I still recommend you to have it only in 'ext.args' and not here in path(opt_file).
Also you should always only parse the optional arguments inside the template.

modules/nf-core/propr/propr/templates/propr.R
```diff
@@ -61,6 +61,73 @@ read_delim_flexible <- function(file, header = TRUE, row.names = 1, check.names
...
+    sampleN <- dimensions[1]
+    geneN <- dimensions[2]
+
+    fileConn<-file(gctFile)
+    writeLines(c("#1.2",paste(geneN,sampleN,sep='\t')), fileConn)
+    close(fileConn)
+
+    samples <- row.names(mat)
+
+    gctDF <- counts[c('gene_name',samples)]
```
suzannejin, line 114: 'gene_name' and 'gene_id' should not be hard coded.
It is possible that the user give a file where instead of 'gene_id' it says 'ensembl_id' or anything else.

suzannejin: for this you can check opt$features_id_col

modules/nf-core/propr/propr/main.nf
BirkirRey: I have removed it, it was a remnant of generating the CLS file and I was also basically using at a flag for GSEA output generation. Since the script does not generate CLS anymore it does not need a sample sheet so now both the propr.R and propd.R templates have a GSEA flag. I have removed any optional input from the main.nf file

modules/nf-core/propr/propr/templates/propr.R
BirkirRey: Done, I added an arguement to the template files called features_name_col to handle this.

suzannejin:
Hey @BirkirRey at the end we are making some changes to include the adjacency matrix. And since your tests are also not passing, we are gonna close the PR. Thank you anyways for your work.

PR #5400, opened 2024-03-26
ci(renovate): Allow channels in environment.ymls

edmundmiller:
No clue if this will work, it passed locally 🤷🏻 

Adds a regex to fix not finding the dependency if `::` and a channel is in the conda dependency.


edmundmiller:
The scope has changed several times is why, I first set out to add the comments to the rnaseq modules, then I realized we had added back in all the channels (which I think helps conda figure the environment out quicker and more consistently), so I fixed the regex, then I was debating adding everything with an automated script.

Maybe that just turns into an `nf-core modules renovate` command, and it adds renovate comments to the module you want 🤔

PR #5423, opened 2024-04-03
UPDATE: added barcoding option to ToulligQC module

Salome-Brunon:
Added a barcoding option to the  module ToulligQC 

- [X ] This comment contains a description of changes (with reason).
- [X ] If you've fixed a bug or added code that should be tested, add tests!


modules/nf-core/toulligqc/meta.yml
```diff
@@ -32,6 +32,9 @@ input:
       type: file
       description: BAM file (necessary if no sequencing summary file)
       pattern: "*.bam"
+  - barcodes:
+      type: string
+      description: Coma separated barcode list (with no spaces)
```
mashehu, line 37:
```suggestion
      description: Comma separated barcode list (with no spaces).
      pattern: "^\\S+$",
```
if you could give an example in the description, that would be extra helpful, I think. (the pattern prohibits whitespace characters)

modules/nf-core/toulligqc/tests/main.nf.test
```diff
@@ -18,6 +18,33 @@ nextflow_process {
...
+                    ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(file(process.out.report_data.get(0).get(1)).readLines()[11..74]).match() },
+                { assert process.out.report_html[0] ==~ ".*/report.html"}
```
mashehu, line 31: please also capture the version in the snapshot

mashehu (changes requested):

mashehu (approved):

modules/nf-core/toulligqc/main.nf
```diff
@@ -28,18 +29,20 @@ process TOULLIGQC {
     def seq_summary_input = seq_summary ? "--sequencing-summary-source ${seq_summary}" : ""
     def fastq_input = fastq ? "--fastq ${fastq}" : ""
     def bam_input = bam ? "--bam ${bam}" : ""
+    def barcode_input = barcodes ? "--barcoding --barcodes ${barcodes}" : ""
```
jfy133, line 32: As this is jsut a string can this not just come via `$args`? It's not file based input and presumable(?) it's not required for the module to run?

Salome-Brunon: I wasn't sure about the best way to implement the barcoding option. I asked on the modules slack channel but the answer was not really clear. 

should I just create a nextfow.config file with the following?
process {

    withName: TOULLIGQC {
        ext.args = '--barcoding --barcodes barcode01,barcode02'
    }

}

Salome-Brunon: But yes, you are right it is not file based input and it is not required to run!

jfy133: Your example is exactly what you would do.

The critical thing about non-file inptu channels is they should only be an input channel is if they are _required_ for the module to run, which I guess based on the original PR, it is not?

Salome-Brunon: Okay, that makes sense! That was my initial idea but then I got confused.

Now, in terms of testing my module. How do I add the barcoding option to only one test?
for example I want to test the barcodes with only the sequencing summary file and not fastq/bam

jfy133: To one test is a good question... I guess you would want to import the module into the `test` script with teh nf-test 'alias' definition.

Then in the `nextflow.config` alongside the test file you would specify with the `alias` name of the module

PR #5424, opened 2024-04-03
Make use of prefix and support gzipped input

gariem:
- Unzip the input file if it has .gz extension
- Use the unzipped file for prediction
- Provide output according to the prefix of the input tuple

### DIFF
--- a/modules/nf-core/iphop/predict/main.nf
+++ b/modules/nf-core/iphop/predict/main.nf
@@ -12,9 +12,9 @@ process IPHOP_PREDICT {
     path iphop_db
 
     output:
-    tuple val(meta), path("*.Host_prediction_to_genus_m*.csv"), emit: iphop_genus
-    tuple val(meta), path("*.Host_prediction_to_genome_m*.csv"), emit: iphop_genome
-    tuple val(meta), path("*.Detailed_output_by_tool.csv"), emit: iphop_detailed_output
+    tuple val(meta), path("*.Host_prediction_to_genus_m*.csv")  , emit: iphop_genus
+    tuple val(meta), path("*.Host_prediction_to_genome_m*.csv") , emit: iphop_genome
+    tuple val(meta), path("*.Detailed_output_by_tool.csv")      , emit: iphop_detailed_output
     path "versions.yml"                                         , emit: versions
 
     when:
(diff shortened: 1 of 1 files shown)


modules/nf-core/iphop/predict/main.nf
```diff
@@ -12,9 +12,9 @@ process IPHOP_PREDICT {
     path iphop_db
 
     output:
-    tuple val(meta), path("Host_prediction_to_genus_m*.csv")    , emit: iphop_genus
-    tuple val(meta), path("Host_prediction_to_genome_m*.csv")   , emit: iphop_genome
-    tuple val(meta), path("Detailed_output_by_tool.csv")        , emit: iphop_detailed_output
+    tuple val(meta), path("*.Host_prediction_to_genus_m*.csv"), emit: iphop_genus
+    tuple val(meta), path("*.Host_prediction_to_genome_m*.csv"), emit: iphop_genome
+    tuple val(meta), path("*.Detailed_output_by_tool.csv"), emit: iphop_detailed_output
```
SPPearce, line 17: Can you line these commas back up please

modules/nf-core/iphop/predict/main.nf
```diff
@@ -25,17 +25,28 @@ process IPHOP_PREDICT {
...
 
-    mv iphop_results/Host_prediction_to_genus_m*.csv .
-    mv iphop_results/Host_prediction_to_genome_m*.csv .
-    mv iphop_results/Detailed_output_by_tool.csv .
+    for FILE in iphop_results/Host_prediction_to_*.csv; do
+        mv \${FILE} ./${prefix}.\${FILE##*/}
+    done
+
+    mv iphop_results/Detailed_output_by_tool.csv ./${prefix}.Detailed_output_by_tool.csv
+    rm ${prefix}.fasta
```
SPPearce, line 49: This is only present if it was unzipped, so you'll get a failure if not.

SPPearce (changes requested):

SPPearce:
You should delete the unzipped file, just conditionally on whether you needed to unzip. Else it'll hang around and take up space

modules/nf-core/iphop/predict/main.nf
gariem: Do you think rm -f would be ok? So, it will delete the file if it exists and otherwise just pass.
I would prefer that instead of adding more lines to see whether the file was zipped/plain.

SPPearce: Hmm. That seems to return an exitcode of 0, so should be fine?

SPPearce (commented):
Could you add a zipped fasta file to the tests?

SPPearce:
@gariem , why did you close this?

PR #1167, opened 2021-12-14
Add helper script to find duplicate test YAML md5s

ewels:
Related to https://github.com/nf-core/tools/pull/1358 - quick script to make it easy to find instances of duplicate md5sums.

image

```console
❯ python .github/check_duplicate_md5s.py --help
Usage: check_duplicate_md5s.py [OPTIONS]

  Find duplicate file MD5 sums in test YAML files.

Options:
  --min_dups INTEGER  Minimum number of duplicates to report  [default: 5]
  --search_dir TEXT   Glob directory pattern used to find test YAML files
                      [default: .github/../tests/**/test.yml]

  --help              Show this message and exit.
```

This script is not currently run anywhere automatically (in any CI), so is only an undocumented helper script for maintainers.

## PR checklist

- [x] This comment contains a description of changes (with reason).

PR #5452, opened 2024-04-09
add ruff linting

mashehu:
we have now more and more python files for the CI, time to lint them like we do in nf-core/tools


mashehu:
@nf-core-bot fix linting

.github/check_duplicate_md5s.py
adamrtalbot, line 1: Should this file be in here?

adamrtalbot (commented):
Just a question about the check_duplicate_md5s.py file. It's never called so I'm not sure it should be there?

mashehu:
Was also wondering about that. @edmundmiller I think you created it, can we remove it?

edmundmiller:
Wasn't me 😬 
https://github.com/nf-core/modules/commit/4e5406c221e91ede93d9d8036110da62c86cca66
@ewels

ewels:
image

ewels:
https://github.com/nf-core/modules/pull/1167 and #1358

ewels:
Riiight, vaguely coming back to me now. It was something to do with pytest tests and not wanting to include duplicate empty files.

I guess that as we're moving away from pytest to nf-test then it's no longer needed. I'd wager money that basically no-one has run it since I merged that PR 😆

.github/check_duplicate_md5s.py
mashehu: removed

adamrtalbot (approved):
Not a massive fan of removing a file in an unrelated PR, but I want the prettier bit in!

PR #5470, opened 2024-04-11
migrating flash to nf-test

DavideBag:
## PR checklist

- [x] Migrating flash module to nf-test.
  - For modules:


modules/nf-core/flash/main.nf
```diff
@@ -19,10 +19,12 @@ process FLASH {
 
     script:
     def args = task.ext.args ?: ''
+    def args2 = task.ext.args ?: ''
```
SPPearce, line 22: What is this for?

SPPearce (commented):
I don't understand what the args2 is for here, when it is just next to args in the code

modules/nf-core/flash/main.nf
DavideBag: I inserted args2 to later add a --quiet function in the config file that I use in the pipeline.

DavideBag: Maybe I can put everything in args1?

SPPearce: Yes. args2 is used when multiple tools are being piped together (like bwa into samtools) so you can pass separate arguments to each tool ,or when tools have positional arguments (which isn't very common)

DavideBag: Thank you for the review and for the clarity, I'll make the changes.

SPPearce (commented):
Does this module take fastq files in and produce different ones? If so, can you include a check that the output files have different names

SPPearce:
> Does this module take fastq files in and produce different ones? If so, can you include a check that the output files have different names

Or you could give them a suffix to make it clear they have been through this tool.

lescai:
Hi @SPPearce and @DavideBag 
I understand that preference is always not to *hard-code* any prefix / suffix, and leave this to the config.
**however**, as Simon highlights one should include an explicit check whenever there's the risk of an overlap in input/output names. 
Best practice I'd suggest is the one adopted in the `samtools` module, since in sorting / slicing this is very common:
https://github.com/nf-core/modules/blob/master/modules/nf-core/samtools/sort/main.nf

SPPearce:
There are definitely many fastq tools that don't follow that convention though

SPPearce:
I'm happy to introduce breaking changes that make sense. One option is to
rename the output channel, then it will definitely fail and force people to
do something

lescai (approved):
LGTM now, quite thorough and now more explicit with flash documentation.

SPPearce (approved):

DavideBag:
Thank you to @SPPearce and @lescai for investing time in reviewing.

PR #5475, opened 2024-04-11
update rest of seqkit modules to 2.8.1

mashehu:
remove unnecessary  `{ assert snapshot(process.out.versions).match("versions") }` when there is already `{ assert snapshot(process.out).match() }`


SPPearce:
Are you planning on updating the remaining pytests to nf-test too?

mashehu:
ran out of time...

maxulysse:
Why closing it?

modules/nf-core/seqkit/grep/tests/main.nf.test
```diff
@@ -0,0 +1,80 @@
...
+
+    test("with_file") {
+
+        when {
+            process {
+                """
+                input[0] = [[ id:'test', single_end:false ], // meta map
+                            file(params.test_data['homo_sapiens']['genome']['genome_21_fasta'], checkIfExists: true)
+                            ]
+                input[1] = file(params.test_data['homo_sapiens']['genome']['genome_header'], checkIfExists: true)
```
maxulysse, line 20: Can you update the paths?

maxulysse: Good enough

modules/nf-core/seqkit/seq/tests/main.nf.test
```diff
@@ -26,7 +26,6 @@ nextflow_process {
             assertAll(
                 { assert process.success },
                 { assert snapshot(process.out).match() },
-                { assert snapshot(process.out.versions).match("versions") }
```
heuermh, line 29: Thanks!  This is what I was hoping for, no longer considering checksums on the versions file.

SPPearce: Well, the versions is being checked in the line above, as it is part of `process.out`

maxulysse (approved):

heuermh (approved):
Thanks!

maxulysse (approved):

PR #5477, opened 2024-04-12
EGGNOGMAPPER: Handled optional inputs

GallVp:
## PR checklist

Closes #XXX 

- [x] EGGNOGMAPPER: Handled optional inputs
- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


GallVp:
Thank you @vagkaratzas for the excellent module and nf-test dataset. Here I have added handling logic for cases when diamond db and database file are not provided.

modules/nf-core/eggnogmapper/main.nf
```diff
@@ -23,11 +23,13 @@ process EGGNOGMAPPER {
...
     script:
-    def args = task.ext.args ?: ''
-    def prefix = task.ext.prefix ?: "${meta.id}"
-    def is_compressed = fasta.name.endsWith(".gz")
-    def fasta_name = fasta.name.replace(".gz", "")
-    def dbmem = task.memory.toMega() > 40000 ? '--dbmem' : ''
+    def args            = task.ext.args                 ?: ''
+    def prefix          = task.ext.prefix               ?: "${meta.id}"
+    def is_compressed   = fasta.name.endsWith(".gz")
+    def fasta_name      = fasta.name.replace(".gz", "")
```
vagkaratzas, line 29: Since you are at it, can you replace the is_compressed check and assignment with this?:
def is_compressed = fasta.getExtension() == "gz" ? true : false
def fasta_name = is_compressed ? fasta.getBaseName() : fasta
    
   Similarly to the blast/blastp module for example; more robust.

vagkaratzas (changes requested):
Thanks! Good job about the versions snapshot update and also left a minor comment.
But, other than that, I am not sure we can have optional inputs for files. These:
path(eggnog_db)
path(eggnog_data_dir)
    
are still there and will still need to be defined even if you pass the new args. Correct me if I am wrong. Or do the checks cover for empty inputs such as []? If so, maybe add another nf-test that tests exactly for that case, to prevent errors/bugs.

modules/nf-core/eggnogmapper/main.nf
GallVp: Thank you @vagkaratzas 

You are right. The checks only cover empty inputs such as []. Now I have added a test to verify this case.

I have also made changes to `is_compressed` and `fasta_name`.

modules/nf-core/eggnogmapper/tests/main.nf.test
```diff
@@ -47,7 +47,60 @@ nextflow_process {
...
-                { assert process.out.versions }
+                { assert snapshot(process.out.versions).match("versions") }
+            )
+        }
+
+    }
+
+    test("Should search for protein annotations against the eggnogmapper db -- empty-params") {
+
+        setup {
```
vagkaratzas, line 58: This is the same setup step as the previous test case. I think it can be written once, before the tests start.

vagkaratzas (approved):
Hokay! Left a minor comment.

modules/nf-core/eggnogmapper/tests/main.nf.test
GallVp: Thank you! I have made this change.

PR #5490, opened 2024-04-16
Add BRACKEN/BUILD module

### LINKED ISSUE #2957: New module: bracken/build

jfy133:
## PR checklist

Closes #2957 

This is a weird one because all that the module does is inserted a couple of extra files into an otherwise 'standard' kraken2 database.  You then pass the same database to both kraken2 and bracken after, thus the 'includeInputs' definition.

It does mean we violate the 'prefix' rule, but I don't really see why you would necessarily want to change this given it should be the same database. But we could theoreitcally do this if necessary by adding a `mv kraken2db ${prefix}` command after, but I fear this may violate some caching rules and thus break `-resume`.

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:


modules/nf-core/bracken/build/environment.yml
```diff
@@ -0,0 +1,9 @@
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/environment-schema.json
```
CarsonJM, line 2:
```suggestion
```

modules/nf-core/bracken/build/main.nf
```diff
@@ -0,0 +1,47 @@
+process BRACKEN_BUILD {
+    tag "$meta.id"
+    label 'process_medium'
```
CarsonJM, line 3: Should this be high? I've only run this command once but seem to recall it requiring a lot of memory

modules/nf-core/bracken/build/meta.yml
```diff
@@ -0,0 +1,48 @@
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/meta-schema.json
```
CarsonJM, line 2:
```suggestion
```

modules/nf-core/bracken/build/main.nf
```diff
@@ -0,0 +1,47 @@
...
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/bracken:2.9--py38h2494328_0':
+        'biocontainers/bracken:2.9--py38h2494328_0' }"
+
+    input:
+    tuple val(meta), path(kraken2db)
+
+    output:
+    tuple val(meta), path(kraken2db, includeInputs: true), emit: db
```
CarsonJM, line 14: Do you think it would be worth adding other outputs to capture the "*mer" files that are created? Just to ensure it has run successfully

modules/nf-core/bracken/build/tests/main.nf.test
```diff
@@ -0,0 +1,69 @@
...
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(
+                        file("${process.out.db[0][1]}/database100mers.kmer_distrib").name,
+                        file("${process.out.db[0][1]}/database100mers.kraken").name,
+                        file("${process.out.db[0][1]}/database.kraken").name,
```
CarsonJM, line 40: Inconsistent hashes?

modules/nf-core/bracken/build/main.nf
```diff
@@ -0,0 +1,47 @@
...
+    label 'process_medium'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/bracken:2.9--py38h2494328_0':
+        'biocontainers/bracken:2.9--py38h2494328_0' }"
+
+    input:
+    tuple val(meta), path(kraken2db)
+
```
CarsonJM, line 12: Do you think the read length parameter should also be included as input since it affect the output file names?

CarsonJM (commented):
Tbh it looks pretty good to me as is. Just a few comments/questions to consider and maybe address

modules/nf-core/bracken/build/main.nf
jfy133: Could do, I couldn't remember from when I building some databases... Will bump for security

modules/nf-core/bracken/build/main.nf
jfy133: There is a default (100bp) so I otherwise leave that to the pipeline developer to inject that into `args` :), and the current output glob doesn't explicitly pick them up so shouldn't matter

modules/nf-core/bracken/build/main.nf
jfy133: As in an additional channel of just the *mer files that someone could take instead? Yes I guess,... Good idea!

modules/nf-core/bracken/build/tests/main.nf.test
jfy133: Yup :/

jfy133 (commented):
Thanks @CarsonJM ! Will make the changes when I have a moment :)

nvnieuwk (approved):
This LGTM, it indeed doesn't make any sense indeed to have a prefix in this module. :p

PR #5505, opened 2024-04-19
Add support for Kallisto h5 format to tximeta-tximport

nictru:
This change allows the usage of inferential replicates in Kallisto. Handling them was not previously possible, since each replicate is saved in a separate plaintext file, but there is a single `h5` file which contains all of them. 

The according bioconda package was updated via [this PR](https://github.com/bioconda/bioconda-recipes/pull/47286)

### DIFF
--- a/modules/nf-core/tximeta/tximport/environment.yml
+++ b/modules/nf-core/tximeta/tximport/environment.yml
@@ -6,4 +6,4 @@ channels:
   - bioconda
   - defaults
 dependencies:
-  - "bioconda::bioconductor-tximeta=1.20.1"
+  - "bioconda::bioconductor-tximeta=1.20.1=r43hdfd78af_1"
--- a/modules/nf-core/tximeta/tximport/environment.yml
+++ b/modules/nf-core/tximeta/tximport/environment.yml
@@ -6,4 +6,4 @@ channels:
   - bioconda
   - defaults
 dependencies:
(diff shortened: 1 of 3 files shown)


pinin4fjords (changes requested):
A couple of points: 

~~- Don't forget to change the environment.yml to keep the versions the same as the container.~~ (sorry, bad, it's just a build so you don't need to do that)
- Do you think you could add a test for .h5 files to make sure this works and stays working?

SPPearce:
Hi @nictru , do you need some help to get this finished off?

nictru:
So the story here is a bit complicated. TLDR: Even with these changes, one cannot do the thing I was hoping for

The reason for this is that when importing the `h5` file, it still creates a list of objects instead of a single, aggregated object (as described [here](https://bookdown.org/jean_souza/PreProcSEQ/construction-of-the-gene-expression-matrix.html#:~:text=tximport%20generates%20a%20list%20of,for%20the%20genes%20and%20samples.)). So actually what I need is `TXIMETA_TXIMETA` which I implemented as a local module in circrna so far [here](https://github.com/nf-core/circrna/tree/dev/modules/local/tximeta/tximeta). 

Don't know if it makes sense to merge this in this situation, but I should probably create a separate module for `TXIMETA_TXIMETA`

SPPearce:
I don't really know the context, so you do what seems most sensible ;)

nictru:
I guess there is no need for this as it does not offer any benefit over the existing implementation

PR #5510, opened 2024-04-20
Updateblastdb new module

toniher:
This addresses issue #5267


modules/nf-core/blast/updateblastdb/main.nf
```diff
@@ -0,0 +1,50 @@
...
+    tag "$meta.id"
+    label 'process_medium'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/blast:2.15.0--pl5321h6f7f691_1':
+        'biocontainers/blast:2.15.0--pl5321h6f7f691_1' }"
+
+    input:
+    tuple val(meta)
```
itrujnara, line 11: If meta is the only input passed, a tuple is not necessary. I am not sure if meta is even the correct name for this, as from what I understand it refers to database IDs rather than samples. Consider `val db_name` or something similar.

modules/nf-core/blast/updateblastdb/main.nf
```diff
@@ -0,0 +1,50 @@
...
+
+    output:
+    tuple val(meta), path("${meta.id}"), emit: db
+    path "versions.yml"                , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
```
itrujnara, line 21: Align equal signs

modules/nf-core/blast/updateblastdb/tests/main.nf.test
```diff
@@ -0,0 +1,33 @@
...
+                input[0] = [ [id:'mito'] ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert process.out.db },
+                { assert process.out.versions }
```
itrujnara, line 28: Snapshots would be good here, as the files should in theory be stable. Certainly for the versions, if feasible also for the database.

itrujnara (commented):
Nicely done. I have added some minor comments that need addressing before merge.

modules/nf-core/blast/updateblastdb/tests/main.nf.test
toniher: It is not possible. Since these are downloaded from NCBI https://ftp.ncbi.nlm.nih.gov/blast/db/ and neither the tool nor the FTP provide any version, it is not convenient nor practical to check it.

modules/nf-core/blast/updateblastdb/main.nf
toniher: Do you know any other similar module example by chance?

modules/nf-core/blast/updateblastdb/main.nf
toniher: Hi, do you mean the '=' of args and prefix vertically? If that, other modules don't look they are doing it.

modules/nf-core/blast/updateblastdb/tests/main.nf.test
SPPearce: Do any of the output files have a static name?

toniher: Yep. They have. I will look for a test checking that.

SPPearce: There are ways to check names, let me know if you need help

mashehu:
Sorry, just on my phone, but how did you figure out that the github runners somehow decompress the files? Did you check the md5sums? We haven't seen these with any other compressed files in tests.

toniher:
> Sorry, just on my phone, but how did you figure out that the github runners somehow decompress the files? Did you check the md5sums? We haven't seen these with any other compressed files in tests.

Because the tests are working locally and on gitpod (you can check in that latter case). Do you remember by chance any of these tests with compressed output archives?

SPPearce:
> Sorry, just on my phone, but how did you figure out that the github runners somehow decompress the files? Did you check the md5sums? We haven't seen these with any other compressed files in tests.

On the action, it lists the files that it was searching to try and find the expected file:
image, see https://github.com/nf-core/modules/actions/runs/8897581550/job/24432800737?pr=5510#step:14:56

PR #5542, opened 2024-04-30
add updates in nf-core/cellrangermulti module script

fmalmeida:
This work relates to requirements and bugs identified during  https://github.com/nf-core/scrnaseq/pull/276
It relates to allowing module to use FFPE data and to properly parse optional files for config generation.

### DIFF
--- a/modules/nf-core/cellranger/multi/main.nf
+++ b/modules/nf-core/cellranger/multi/main.nf
@@ -44,25 +44,23 @@ process CELLRANGER_MULTI {
 
     // if references + FASTQ are empty, then don't run corresponding analyses
     // get names of references, if they exist
-    // empty reference channels can stage as "[]" when skipped by the workflow, or get an 'EMPTY' file when empty
+    // empty reference channels (all under references/) can stage as "[]" when skipped by the workflow
     // empty FASTQ channels stage as "fastqs"
-    // empty files stage as the file name, we check against 'EMPTY'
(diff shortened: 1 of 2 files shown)


modules/nf-core/cellranger/multi/main.nf
```diff
@@ -44,23 +44,24 @@ process CELLRANGER_MULTI {
 
     // if references + FASTQ are empty, then don't run corresponding analyses
     // get names of references, if they exist
-    // empty reference channels stage as "references"
+    // empty reference channels can stage as "[]" when skipped by the workflow, or get an 'EMPTY' file when empty
```
grst, line 47: Any point in retaining the `EMPTY` if `[]` is now supported?

grst (approved):
I'm wondering about the `EMPTY`, but LGTM

modules/nf-core/cellranger/multi/main.nf
fmalmeida: No particular reason. I just did not try yet. So was retained as 'legacy' compatibility.

But this is a good point.
We can solve this and keep entirely as '[]' instead of the EMPTY file.

But it needs a proper testing to guarantee these are the only points it is used.

fmalmeida: Hi @grst ,
I have just updated it to not rely on `EMPTY` file anymore. It works entirely with `[]`.
You can see in the `nf-test` that I modified it to use `[]` and did not change anything in the snaps, being all the same. And still passing all tests.

adamrtalbot (approved):
Looks way cleaner!

PR #5544, opened 2024-04-30
Unicycler nftest

FloWuenne:
## PR checklist

Adds nf-test for unicycler module.


modules/nf-core/unicycler/tests/main.nf.test
```diff
@@ -0,0 +1,37 @@
...
+    tag "modules_nfcore"
+    tag "unicycler"
+
+    test("bacteroides_fragilis - illumina - single end") {
+
+        when {
+            process {
+                """
+                input[0] = [ [ id:'test', single_end:true ], // meta map
+                               file(params.modules_testdata_base_path + 'genomics/sarscov2/illumina/fastq/test_1.fastq.gz', checkIfExists: true)
```
maxulysse, line 19:
```suggestion
                               [file(params.modules_testdata_base_path + 'genomics/sarscov2/illumina/fastq/test_1.fastq.gz', checkIfExists: true)],
```

ngspp:
@maxulysse as fast as lightning there

FloWuenne:
Still need to finalize the nf-test output assertions. @maxulysse and I were troubleshotting that togther 😉 @ngspp

modules/nf-core/unicycler/tests/main.nf.test
```diff
@@ -0,0 +1,41 @@
+// TODO nf-core: Once you have added the required tests, please run the following command to build this file:
+// nf-core modules test unicycler
```
maxulysse, line 2:
```suggestion
```

FloWuenne:
There seems to be an issue with the conda env. I have replicated outside of nextflow as well and get the same error. Don't know how to fix it at this point. Any ideas @JoseEspinosa @d4straub ?

d4straub:
The problem with conda might be due to

> The most significant one is that Unicycler now enforces that SPAdes cannot be a version later than v3.13.0. This is because after that version, SPAdes stopped saving intermediate files that Unicycler needs. Unicycler users that had later versions of SPAdes were getting suboptimal assemblies.

fixed in https://github.com/rrwick/Unicycler/releases/tag/v0.4.9. If that is the case, either using 0.4.9 instead of 0.4.8 or fixing spades version in the environment.yml might fix it. (Sorry, didnt test)

FloWuenne:
@d4straub There actually doesn't seem to be a version 0.4.9 but there are for 0.5.0. Testing right now to see if that fixes it. Gotta update the container versions then as well of course!

FloWuenne:
So I tried running with version 0.5.0 both conda and Docker and actually both fail with test data!

```
=== ERRORs:
 * 0:00:02.379     6M / 136M  ERROR   General                 (kmer_coverage_model.cpp   : 187)   Invalid kmer coverage histogram, make sure that the coverage is indeed uniform
 * system call for: "['/workspace/modules/.nf-test/tests/ec5addf7ea2590db79a5c8d180299d3d/work/conda/unicycler-66b9a73317c352e54972e624abe20c82/bin/spades-core', '/workspace/modules/.nf-test/tests/ec5addf7ea2590db79a5c8d180299d3d/work/97/630766f4ff131b3da613cfc9a21f36/spades_assembly/K81/configs/config.info', '/workspace/modules/.nf-test/tests/ec5addf7ea2590db79a5c8d180299d3d/work/97/630766f4ff131b3da613cfc9a21f36/spades_assembly/K81/configs/isolate_mode.info']" finished abnormally, OS return value: 21
```

d4straub:
Sorry for the misleading idea. Assemblers and small test data are often problematic in my experience. Unfortunately I dont have a better idea how to solve that.

SPPearce:
Any plan to get round this?

FloWuenne:
I have to admit that I have no idea how to fix it at this point and unfortunately don't have time at the moment to dig deeper. If anybody has a good idea or lead how to fix this, would love some suggestions!

SPPearce:
@FloWuenne , I found some nanopore data for bacteroides_fragilis existed, works with just long reads only (short reads too seemed to take too long)

FloWuenne:
Amazing @SPPearce ! Looks like all tests pass now. I will merge this now! Thanks for helping out here 💪🏻 !

FloWuenne:
Hm, looks like it might have just taken too long to pull the singularity image in the CI? Testing now on Gitpod to see if it works there!

SPPearce:
I think it is a general connection issue yes, will probably sort itself out

FloWuenne:
@SPPearce Just tested and yes singularity did work in Gitpod. Also removed the obsolete snapshots from `main.nf.test.snap`!

PR #5609, opened 2024-05-15
[FIX] Make output compression in TCOFFEE/ALIGN more robust

lrauschning:
Came across an issue with how some bash shells handle anonymous named pipes that causes tcoffee to interpret it as a parameter and fail.
This commit should make it a bit more robust.


luisas:
I am integrating exactly this in #5753 :) 

I am closing this PR as is redundant now

( I work with @lrauschning so i am not just closing a random PR )

PR #5618, opened 2024-05-16
Iqtree version bump and output fix

### LINKED ISSUE #5607: iqtree: optional output files missing from module

### Description of the bug

IQ-Tree can output multiple optional files depending on the flags/input parameters used. The nf-core module currently does not capture all possible output files.

_No response_

### Relevant files

_No response_

### System information

_No response_

tstoeriko:
## PR checklist

Closes  #5607  

- bump version from 2.3.0 to 2.3.3
- include all optional output files (as documented in the IQ-Tree command reference)
  - For modules:


maxulysse:
looking good, but I feel you added some optional files, is there any possibility to test for it too?

tstoeriko:
> looking good, but I feel you added some optional files, is there any possibility to test for it too?

Yes, the module was basically missing all optional output files, likely because there is no centralized documentation of the possible output files. I'll have a look at the tests again and try to include as many of the optional files as possible.

tstoeriko:
This module has now been majorly reworked  due to missing inputs and outputs and addition of new tests. I would be grateful for some more feedback.

Major changes:
 - Included all optional input and output files. Though there is no centralized documentation of input and output file options, so I cannot guarantee I've caught them all.
 - Removed one of the previous inputs (fconst) which is not a file and should be fine to supply via ext.args.
 - Made the main alignment input file optional as there are modes which do not require an alignment.
 - Wrote tests for nearly all output files.
 
There are a couple of things I am struggling with:
 - The module now has 14 inputs (up from 2), I am a bit worried about usability. Will users just have to adapt?
 - I've tried to include relevant file extensions in the meta.yml, but I think IQ-Tree itself doesn't actually check for them. Should I keep the patterns?
 - I'd love some feedback on the variable names I've chosen. I think with this huge number of inputs and outputs it is important to make them as intuitive as possible.
 - I've spent quite some time on setting up appropriate tests, but I am still missing tests for three specific outputs.
   The input data currently available do not work for these tests, they are likely not complex enough.
   I am personally not familiar enough with the corresponding modes to fully troubleshoot the input data requirements. 
   I've tested some data locally, but have not been able to find a suitable small dataset to use.
   I'm happy to take some time in the future to find a good dataset, but maybe we could get this PR merged for now?

tstoeriko:
This module has now been majorly reworked due to missing inputs and outputs and addition of new tests. I would be grateful for some more feedback.

Major changes:
 - Included all optional input and output files. Though there is no centralized documentation of input and output file options, so I cannot guarantee I've caught them all.
 - Removed one of the previous inputs (fconst) which is not a file and should be fine to supply via ext.args.
 - Made the main alignment input file optional as there are modes which do not require an alignment.
 - Wrote tests for nearly all output files.
 
There are a couple of things I am struggling with:
 - The module now has 14 inputs (up from 2), I am a bit worried about usability. Will users just have to adapt?
 - I've tried to include relevant file extensions in the meta.yml, but I think IQ-Tree itself doesn't actually check for them. Should I keep the patterns?
 - I'd love some feedback on the variable names I've chosen. I think with this huge number of inputs and outputs it is important to make them as intuitive as possible.
 - I've spent quite some time on setting up appropriate tests, but I am still missing tests for three specific outputs.
   The input data currently available do not work for these tests, they are likely not complex enough.
   I am personally not familiar enough with the corresponding modes to fully troubleshoot the input data requirements. 
   I've tested some data locally, but have not been able to find a suitable small dataset to use.
   I'm happy to take some time in the future to find a good dataset, but maybe we could get this PR merged for now?

modules/nf-core/iqtree/main.nf
```diff
@@ -4,34 +4,88 @@ process IQTREE {
...
+    tuple val(meta), path(mdef)
+    tuple val(meta), path(partitions_equal)
+    tuple val(meta), path(partitions_proportional)
+    tuple val(meta), path(partitions_unlinked)
+    tuple val(meta), path(guide_tree)
+    tuple val(meta), path(sitefreq_in)
+    tuple val(meta), path(constraint_tree)
+    tuple val(meta), path(trees_z)
+    tuple val(meta), path(suptree)
+    tuple val(meta), path(trees_rf)
```
mahesh-panchal, line 24: I suggest removing all the meta's except the for the first one. Practically, files related to any given sample should be made into a tuple in some operation before it reaches this process, and will then be split into the channels with `map`/`multiMap`.

modules/nf-core/iqtree/tests/iqtree_sup.config
```diff
@@ -0,0 +1,5 @@
+process {
+    withName: "IQTREE" {
+        ext.args = "-t https://raw.githubusercontent.com/nf-core/test-datasets/phyloplace/testdata/cyanos_16s.newick -sup https://raw.githubusercontent.com/nf-core/test-datasets/phyloplace/testdata/cyanos_16s.newick"
```
mahesh-panchal, line 3: Files generally should never be supplied through ext.args as it by passes Nextflow's staging mechanism and can lead to multiple access issues

modules/nf-core/iqtree/tests/iqtree_basic.config
```diff
@@ -1,5 +1,5 @@
 process {
     withName: "IQTREE" {
-        ext.args = "-bb 1000 -wbt"
+        ext.args = "-m LG -wt"
```
mahesh-panchal, line 3: If any of these options are mandatory to make the tool run with the minimum number of parameters then these should all be supplied as `input:`

mahesh-panchal (commented):
Not really knowing anything about the tool, this is a really nice effort. There's a smattering of comments to address.

modules/nf-core/iqtree/tests/iqtree_basic.config
tstoeriko: Neither of them are mandatory. With the first option I'm supplying a model to skip a model selection process (which I reasoned would unnecessarily slow down tests).
The other one is a flag required for a specific output type which would not work in any of the other tests. I though it would be easiest to include it here, but I can just move it to a separate test and keep the basic test truly basic.

modules/nf-core/iqtree/tests/iqtree_sup.config
tstoeriko: Good catch, thank you! This is a leftover from testing which I missed. This config file actually shouldn't be necessary anymore, I'll just remove it.

modules/nf-core/iqtree/main.nf
tstoeriko: I realized the way I had implemented the meta information did not make a lot of sense, will come up with a more appropriate solution.

SPPearce:
You really don't have to test every possible input/output of the tool, so don't worry about missing a few files.
Are a lot of the input files mutually exclusive with each other?

tstoeriko:
> You really don't have to test every possible input/output of the tool, so don't worry about missing a few files. Are a lot of the input files mutually exclusive with each other?

I don't think they're mutually exclusive for the most part. Different combinations of input files will result in different modes being run though.

modules/nf-core/iqtree/main.nf
tstoeriko: Fully removing the meta from most inputs is an issue, because the first input is optional and not required.
I looked at the options again and I believe that at least one of the first two inputs _has_ to be supplied for the program to run. I just kept the meta for these two inputs; by default the first meta is used, but if it is not supplied, the second will be used instead.

modules/nf-core/iqtree/main.nf
```diff
@@ -4,34 +4,89 @@ process IQTREE {
...
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
-        'https://depot.galaxyproject.org/singularity/iqtree:2.3.0--h21ec9f0_0' :
-        'biocontainers/iqtree:2.3.0--h21ec9f0_0' }"
+        'https://depot.galaxyproject.org/singularity/iqtree:2.3.4--h21ec9f0_0' :
+        'biocontainers/iqtree:2.3.4--h21ec9f0_0' }"
 
     input:
     tuple val(meta), path(alignment)
-    val constant_sites
+    tuple val(meta2), path(tree)
```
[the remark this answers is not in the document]
tstoeriko: I get the idea, but wouldn't this be an issue when users want to supply both an alignment and a tree? The module would still need a channel for each of the two main inputs.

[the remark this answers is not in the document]
mahesh-panchal: Ah, then I would make the top tuple take both a tree and an alignment.

tstoeriko:
> Another thing one could add to the tests is that certain channels should be empty. Otherwise this is great work. Thank you.

Thanks for the suggestion and the thorough review in general, @mahesh-panchal ! I added some tests to confirm that certain optional output channels are empty. If there's nothing else missing, I'll go ahead and get this PR merged.

mahesh-panchal:
Sounds good. Great work.

PR #5631, opened 2024-05-18
Stub and nf-test for fastk/fastk

GallVp:
## PR checklist

Closes #XXX 

- [x] Stub and nf-test for fastk/fastk
- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


GallVp:
Strange!

I had to add

```bash
find . -name '*.ktab*' \
    | xargs chmod +r
```
to get it working on docker_self_hosted. Somehow, FASTK is messing with the file permissions on `ktab` files on docker self hosted. Does not happen on my MacBook/macOS/docker.

mahesh-panchal:
I don't know what's going on with the self hosted workers really, but Docker on Mac is running through a VM of sorts that corrects permissions even when it's run as root. I would check the user is supplied in the runner here. 

( and thanks for updating the tests )

GallVp:
nf-core/slack post: https://nfcore.slack.com/archives/CJRH30T6V/p1716244157631629

modules/nf-core/fastk/fastk/environment.yml
```diff
@@ -3,3 +3,5 @@ channels:
   - conda-forge
   - bioconda
   - defaults
+dependencies:
```
mahesh-panchal, line 6: Have things changed. I thought tools without conda could have this file removed?

modules/nf-core/fastk/fastk/main.nf
```diff
@@ -33,6 +33,39 @@ process FASTK_FASTK {
         -N${prefix}_fk \\
         $reads
 
+    find . -name '*.ktab*' \\
```
mahesh-panchal, line 36: I find this very strange having to add this here. To me, this is an issue with the runner not running as the user.

mahesh-panchal (approved):
While the chmod'ing shouldn't make a difference, I feel the solution to this should be something else. I don't have the capacity to investigate at the moment though.

modules/nf-core/fastk/fastk/environment.yml
GallVp: It fails nf-core/lint if this line and the next line are not added.

mahesh-panchal: That I understood. I meant for other modules this environment file was completely deleted.

modules/nf-core/fastk/fastk/main.nf
GallVp: But the docker runner should be using the nf-test.config

https://github.com/nf-core/modules/blob/5e0c5677ea33b3d4c3793244035a191bd03e6736/tests/config/nf-test.config#L33

mahesh-panchal: I think you just found the problem. I believe it should be using:
https://github.com/nf-core/modules/blob/5e0c5677ea33b3d4c3793244035a191bd03e6736/tests/config/nf-test.config#L38 for the self hosted runners.

mahesh-panchal: The question is why are there two different settings for the same thing

GallVp: But if we remove user emulation, the module will not work in nf-core templated pipelines which use `-u $(id -u):$(id -g)`?

modules/nf-core/fastk/fastk/environment.yml
GallVp: Let me try that!

modules/nf-core/fastk/fastk/main.nf
GallVp: Ah, sorry, I missed that. Yes, why? Good find.

mahesh-panchal: No, my point is that this should also use `-u $(id -u):$(id -g)` in the run options of the self hosted profile. And `fixownership` is something nextflow runs afterwards so shouldn't be used really. It's for user outputs

mahesh-panchal: `fixownership` is for when the container must run as `root`

PR #5679, opened 2024-05-23
Swap ngscheckmate/ncm to nftest

SPPearce:
Working on bringing NGSCheckMate modules to nf-test


modules/nf-core/ngscheckmate/ncm/tests/main.nf.test
```diff
@@ -0,0 +1,184 @@
...
+        }
+
+    }
+
+    test("sarscov2 - vcf") {
+        config "./vcf.config"
+        when {
+            process {
+                """
+                input[0] = BCFTOOLS_MPILEUP1.out.vcf.combine(BCFTOOLS_MPILEUP2.out.vcf.map{it[1]}).map{meta, one, two -> [meta, [one, two]]}.map{meta, stuff -> [meta, stuff.flatten()]}
```
SPPearce, line 104: This is horrific and could do with cleaning up...
I have two outputs that are both:
`[meta, vcf]`
and I want to make
`[new_meta, [vcf1, vcf2]]`

nvnieuwk (approved):
LGTM, don't worry about weird channel magic. It's fine as long as it works :p

modules/nf-core/ngscheckmate/ncm/tests/main.nf.test
```diff
@@ -0,0 +1,184 @@
...
+    tag "modules_nfcore"
+    tag "ngscheckmate"
+    tag "ngscheckmate/ncm"
+    tag "bedtools/makewindows"
+    tag "bcftools/mpileup"
+
+    setup {
+
+        run("BEDTOOLS_MAKEWINDOWS") {
+            script "../../../bedtools/makewindows/main.nf"
```
asp8200, line 18: @SPPearce @nvnieuwk : Will that work in a pipeline where bedtools isn't installed?  I'm having problems running this nf-test in Sarek where we don't have bedtools installed. The nf-test ran better after I installed bedtools in Sarek, but that doesn't seem like the right solution to me 🤔

SPPearce: Hmmm.
The alternative would be to add files to test_data instead, but there are quite a few modules that do this kind of setup block. This is essentially taking a bed file and splitting it into individual bases.
I guess BCFTOOLS_MPILEUP exists in sarek already, so you didn't have an issue with that part.

asp8200: BCFTOOLS_MPILEUP is installed in Sarek, so that didn't cause any issues. However, I guess the nf-test should be kind of self-contained, and not rely on what other modules might or might not be installed in the particular pipeline that is trying to use the module.

I also had to change all `params.test_data..` to `params.modules_testdata_base_path..` to get the test kind of work, that is, I managed to get 3 out of 4 "subtests" working:

https://github.com/maxulysse/nf-core_sarek/pull/7

PR #5683, opened 2024-05-23
Added nf-tests for modules/nf-core/racon module

LeonHornich:
Added nf-tests to ```modules/nf-core/racon``` module

Added:
- ```racon/tests``` subfolder containing ```main.nf.test```, ```main.nf.test.snap``` and ```tags.yml```

test runs successfully using ```nf-test test modules/nf-core/racon/ --profile docker --cleanSnapshot```using docker


Joon-Klaps (approved):
LGTM

PR #5702, opened 2024-05-27
nf-test pear module: Cannot get property 'baseName' on null object

LeonHornich:
I am trying to add nf-tests to the nf-core/pear module:
- added ```main.nf-test```, ```nextflow.config```, ```main.nf.test.snap``` and ```tags.yml``` file

After running the tests using ```nf-test test --tag pear --profile docker ``` i am getting a ```Cannot get property 'baseName' on null object``` error. 

I haven't removed any of the old pytest files and i also haven't yet excluded the module from ```pytest_modules.yml```.

[nextflow.log](https://github.com/nf-core/modules/files/15452982/nextflow.log)


SPPearce (approved):
Should be passing the linting now.

Issue #5693, opened 2024-05-24
Ultraplex output channel fastq includes no_match.fastq.gz


PikalaxALT:
### Have you checked the docs?

- [X] [nf-core website: troubleshooting](https://nf-co.re/usage/troubleshooting)
- [X] [nf-core modules documentation](https://nf-co.re/docs/contributing/modules)

### Description of the bug

My workflow attempts to associate the demultiplexed fastq output from ULTRAPLEX with the matched barcode sequence using a regex:
```groovy
ch = ULTRAPLEX.out.fastq.flatMap{meta, fastq ->
    fastq.collect{ filename ->
        barcode = (filename.name =~ /.*([ACGT]+)_(Fwd|Rev)\.fastq\.gz$/)[0][1]
        [meta.id, barcode, meta + ['barcode': barcode], filename
    }
}
```

This fails with an index error (out of range for 0..-1 (index = 0)). When dumping the channel, I noticed that the `no_match` FASTQ were included in this channel, despite the glob intended to exclude them at ultraplex/main.nf:16.

### Command used and terminal output

```console
Unable to share output due to company policy
```

### Relevant files

Unable to share relevant files due to company policy

### System information

Nextflow version 24.04.1-all
Hardware: HPC
Executor: UGE
Container engine: singularity 3.10.5
OS: centos7 x86_64

SPPearce:
Fixed in #5706

PR #5706, opened 2024-05-28
Convert ultraplex to nf-test

SPPearce:
Convert ultraplex to nf-test, and address the no_match files not being removed by the attempted regex, as reported in #5693.


modules/nf-core/ultraplex/main.nf
```diff
@@ -13,10 +13,10 @@ process ULTRAPLEX {
     val(adapter_seq)
 
     output:
-    tuple val(meta), path("*[!no_match].fastq.gz"),              emit: fastq
-    tuple val(meta), path("*no_match.fastq.gz"), optional: true, emit: no_match_fastq
-    path "*.log",                                                emit: report
-    path "versions.yml",                                         emit: versions
+    tuple val(meta), path("*matched.fastq.gz")   , emit: fastq
```
mahesh-panchal, line 16:
```suggestion
    tuple val(meta), path("*matched.fastq.gz") , emit: fastq
```

modules/nf-core/ultraplex/main.nf
```diff
@@ -47,6 +47,31 @@ process ULTRAPLEX {
 
     """
     ${ultraplex_command}
+
+    ## rename the matched files to be able to emit only the matched files
+    MATCHES=\$( find . -type f -name '*.fastq.gz' ! -name '*_no_match_*.fastq.gz' )
+    for MATCH in \$MATCHES; do
+        mv \$MATCH \${MATCH/.fastq.gz/_matched.fastq.gz}
```
mahesh-panchal, line 54: There's no prefix?

mahesh-panchal (approved):

SPPearce:
I should say thank you to @mahesh-panchal for suggesting the way to not capture the no_match files in the channel.

modules/nf-core/ultraplex/main.nf
SPPearce: Not at the moment. It makes files called e.g. ```ultraplex_demux_Sample1_matched.fastq.gz``` based on the barcodes in the input file. The ultraplex_demux could be replaced with a custom prefix, that might be useful

SPPearce: Apparently the tool has an `outputprefix` option which changes the "demux" (but not the `ultraplex_`), so I've done that to be the `prefix`.

SPPearce: @mahesh-panchal , if you can take a look again please

mahesh-panchal (commented):
Nice work. 
I notice the `adapter_seq` is an optional val. This could be removed but I don't know what this will break. 
I just checked the linting and the meta.yml needs updating.

SPPearce:
> Nice work. I notice the `adapter_seq` is an optional val. This could be removed but I don't know what this will break. I just checked the linting and the meta.yml needs updating.

Ok, I removed the adapter_seq value channel, that can be passed via args, and updated the meta.
I also rewrote the code for the actual command while I was at it.

CharlotteAnne:
Hi guysCould you hold fire until I have a chance to review please, this module is used exclusively in my pipelines for now I think.CheersCharlotteSent from my iPhoneOn 5 Jun 2024, at 19:55, Simon Pearce ***@***.***> wrote:﻿

Nice work. I notice the adapter_seq is an optional val. This could be removed but I don't know what this will break. I just checked the linting and the meta.yml needs updating.

Ok, I removed the adapter_seq value channel, that can be passed via args, and updated the meta.
I also rewrote the code for the actual command while I was at it.

SPPearce:
I'll pop it into draft to make sure it doesn't get merged before you look at it @CharlotteAnne

SPPearce:
@CharlotteAnne , are you able to take a look at the suggested changes?

CharlotteAnne:
Yes, I can.

SPPearce:
@CharlotteAnne , there are bulk updates incoming to the modules repo, so I would like to get this finished off this week.
Can you please take a look if you want to do so.

CharlotteAnne:
Hi Simon

I really appreciate the work on the module. I think the nf-core team need
to be a bit considerate that not all contributors are doing Nextflow
nf-core as their full-time job. I am a researcher juggling a lot of
projects and am currently on Summer holiday - given this is a module I am
pretty sure only I am using - why should it be a problem to wait a month to
push all the updates? You can of course go ahead without my testing, but
feels like it would just make more work in the long run. I think this is a
general issue and also comes back to a problem that I think is that the
nf-core team are too cavalier about making breaking changes in updates. It
creates a lot of tech debt for scientists who are not full-time Nextflow
developers (ie. most people using Nextflow/nf-core). I guess when I'm back
I'll turn up to the office hours and have a grumble there :P

Cheers,
Charlotte

SPPearce:
@CharlotteAnne , do you have some time to look at this now?
For the record, I'm not doing this as my job either, and there is at least one other person trying to use the module: https://github.com/nf-core/modules/issues/5693.

CharlotteAnne:
Hi SimonThanks for your patience and pointing out the issue, I’ll look through this tomorrow - cheers!CharlotteSent from my iPhoneOn 20 Aug 2024, at 17:06, Simon Pearce ***@***.***> wrote:﻿
@CharlotteAnne , do you have some time to look at this now?
For the record, I'm not doing this as my job either, and there is at least one other person trying to use the module: #5693.

SPPearce:
Bump for @CharlotteAnne

SPPearce:
@CharlotteAnne , are you going to take a look at this?

mahesh-panchal:
I should say I've brought this up in the maintainers meeting. We've agreed that 3 months is a reasonable time to ask a maintainer for a reasonable time to review. We're going to ask for a second review and once that's in, we'll move forward with it's inclusion.

matthdsm (approved):
Makes sense to me!

PR #5716, opened 2024-05-29
Move annotsv containers to seqera containers

nvnieuwk:
The old `biocontainers` were having some permission errors that were bugging me for a long time. It seems the permissions do get set correctly by migrating to seqera containers

### DIFF
--- a/modules/nf-core/annotsv/annotsv/environment.yml
+++ b/modules/nf-core/annotsv/annotsv/environment.yml
@@ -4,4 +4,4 @@ channels:
   - bioconda
   - defaults
 dependencies:
-  - bioconda::annotsv=3.4.1
+  - bioconda::annotsv=3.4.2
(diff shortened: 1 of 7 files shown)


JoseEspinosa (approved):
LGTM! 🚀

nvnieuwk:
> Is the `oras` protocol just working when singularity is used?

It's only used for singularity in this case, but I'm not sure what it does in case of docker (but I guess it will fail since this is pointing to a singularity container)

nvnieuwk:
(currently fixing the tests and updating the other AnnotSV module)

mahesh-panchal:
I think you can use that protocol for both docker and singularity

nvnieuwk:
I'll give it a go locally :p

nvnieuwk:
```
│   Command error:                                                                                                                                                        │
│     docker: invalid reference format.                                                                                                                                   │
│     See 'docker run --help'. 
```

mahesh-panchal:
Same, Oh well. It seems I misunderstood oras.

nvnieuwk:
Everyone okay with me merging this or should I wait for more reviews?

nvnieuwk:
Okay I'll merge this, we can still revert it if something comes up (I'm the only using this module as far as I know so it should be fine)

PR #5720, opened 2024-05-29
Handle empty files

### LINKED ISSUE #5612: new File breaks cloud execution in bcl_demultiplex

### Description of the bug

This bug relates to the bcl_demultiplex sub workflow.

`new File` will not work with cloud storage like s3, see https://github.com/nf-core/tools/pull/354 for reference, need to use  `file` instead. 

_No response_

### Relevant files

_No response_

### System information

_No response_

k1sauce:
This PR fixes #5612
This PR also handles empty FASTQ files by including a empty fastq emit channel


glichtenstein (approved):
Needs a way to avoid sending falco empty fastqs. Modifying falco module itself looked straighforward to me but it seems I was not aware that it will cause other troubles as stated by the reviewer: https://github.com/nf-core/modules/pull/5769#pullrequestreview-2106076051, I will close my PR then.

SPPearce:
Why is the read group parsing being removed?

k1sauce:
This is a mistake on my part. I thought the read group parsing would download a local copy of the FASTQ files. I see now that it is a buffered Input stream and the file is not downloaded. I will add the read group parsing back in.

k1sauce:
@matthdsm Can I get your review on this, thanks

matthdsm:
sorry to be a bother, but what's the reason you'd like to output the empty fastq entries at all?

I'd suggest a `.filter` approach to filter out the empty fastq's right after the `.transpose`. If you want to output the empty fastq's you could do a `.branch` and put the empty fastq's in a different output channel.

k1sauce:
The simple reason to output them is that they are valid outputs of demultiplexing;  bcl2fastq and bcl-convert both write empty gzip FASTQ files. I don't want to be opinionated on whether a user of this workflow may or may not need them. In fact, I can think of some use cases where a user may want to know if a barcode / umi has 0 reads assigned to it. Would you agree with that approach?

I will implement your suggestion to emit the empty FASTQs in a different output channel. Thanks!

matthdsm:
> The simple reason to output them is that they are valid outputs of demultiplexing; bcl2fastq and bcl-convert both write empty gzip FASTQ files. I don't want to be opinionated on whether a user of this workflow may or may not need them. In fact, I can think of some use cases where a user may want to know if a barcode / umi has 0 reads assigned to it. Would you agree with that approach?

Sure! Let's go with that 😄

k1sauce:
@matthdsm OK should be good now

SPPearce:
@matthdsm , can you take another look at this now?

subworkflows/nf-core/bcl_demultiplex/main.nf
```diff
@@ -67,106 +64,71 @@ workflow BCL_DEMULTIPLEX {
...
+                return [meta, fastq]
+            }
+            // Group by the meta id so that we can find mate pairs if they exist
+            .groupTuple(by: [0])
+            .map { meta, fastq ->
+                meta.single_end = fastq.size() == 1
+                return [meta, fastq.flatten()]
+            }
+            .branch {
+                fastq       : it[0].empty == true
```
matthdsm, line 123: shouldn't this be `false`?

k1sauce:
> Is it possible to also add a test with an empty fastq?

Will do

subworkflows/nf-core/bcl_demultiplex/main.nf
k1sauce: yep

glichtenstein:
@k1sauce @matthdsm I've used this test-dataset and found it useful in my case, maybe be of interest:

### Dataset used
- A public BCL from 10X Genomics. I will add it to nf-core tests datasets when possible.
```
mkdir -p iseq-DI; cd iseq-DI
wget https://cf.10xgenomics.com/supp/spatial-exp/demultiplexing/iseq-DI.tar.gz
wget https://cf.10xgenomics.com/supp/spatial-exp/demultiplexing/bcl_convert_samplesheet.csv
```
SampleSheet was modified to include bad barcodes.

**nf-core_samplesheet.csv**
```
id,samplesheet,lane,flowcell,per_flowcell_manifest
AA0BCDFGH,/data/scratch/iseq-DI/bcl_convert_samplesheet.csv,1,/data/scratch/iseq-DI/iseq-DI.tar.gz
```
**bcl_convert_samplesheet.csv**
```
[Header],,,
FileFormatVersion,2,,
,,,
[BCLConvert_Settings],,,
CreateFastqForIndexReads,0,,
,,,
[BCLConvert_Data],,,
Lane,Sample_ID,index,index2
1,iseq-DI,GTAACATGCG,AGGTAACACT
1,Bad-iseq-DI,TATACAGATA,GATACAGATA
```

k1sauce:
This should be ready to go now. I had trouble with the nf-test for BCL Convert. The problem was that the test data file permissions were too restrictive. Because the GitHub action is self hosted with rootless docker,  the test is executed as the host user and it could not read the test data when the volume is mounted. The fix was to [change the test data file permissions to 775](https://github.com/nf-core/test-datasets/pull/1253). It was an easy fix but it took me awhile to figure out what was going wrong.  I found this explanation helpful https://joeeey.com/blog/rootless-docker-avoiding-common-caveats/#storage-fixes.

SPPearce:
@matthdsm , @glichtenstein, are you happy with this?

matthdsm (approved):
👍🏻

PR #5727, opened 2024-05-31
Update LAST modules

charles-plessy:
This is an update of all the `last/*` modules to version 1542.  This is needed for the finishing of the `nf-core/pairgenomealign` pipeline.  It also solves the problem that the `last/lastal` module currently is at an older version from the other ones, which can cause some crashes when recently added options.

Additional minor changes:

 - Fixed the description of `last/lastal`
 - Changed file name suffix for trained parameters to match upstream documentation and examples.

Unfortulately I could not run `nf-core modules test` as it complained that `nf-test` was not available on my system, but I expect the GitHub CI to catch issues as the changes in this PR are quite minor.

## PR checklist

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:


charles-plessy:
I do not understand error in the failed tests:

```
Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run
```

SPPearce:
These modules don't have a nf-test yet, only the older pytest.
The failures seem to be related to the md5sums of some of the files having changes, e.g.
image
Note that the actual error is under the Run pytest workflow heading, not the lower ones.

nf-test has to be installed separately (https://github.com/askimed/nf-test?tab=readme-ov-file#installation), or you can use gitpod (https://nf-co.re/docs/tutorials/gitpod/git_in_gitpod), where all the required software is automatically installed.

charles-plessy:
I have converted all LAST modules to _nf-test_.  The conda test for `last/dotplot` fail probably because image files are not binary identical if not produced by identical libraries, and the MD5 fingerprints were made using the Singularity image, not conda.  What shall I do?  Relax the tests so that they only test for the existence of the output?

charles-plessy:
Like something like `{ assert process.out.png != null }` as in `modules/nf-core/busco/generateplot/tests/main.nf.test` ?

SPPearce:
Yeah, png files often give different md5sums on different platforms.
You can just check existence of the file, see https://nf-co.re/docs/contributing/nf-test/assertions
Something like `assert { process.out.png.get(0).get(1).endsWith("test.png") }`

modules/nf-core/last/dotplot/environment.yml
```diff
@@ -1,7 +1,9 @@
...
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/environment-schema.json
+name: "last_dotplot"
 channels:
   - conda-forge
   - bioconda
   - defaults
 dependencies:
-  - bioconda::last=1453
+  - "bioconda::last=1542"
```
SPPearce, line 9: Any reason for the quote marks here? Other modules don't have them.

modules/nf-core/last/lastal/meta.yml
```diff
@@ -1,41 +1,43 @@
...
   - align
   - fastq
   - fasta
 tools:
-  - last:
-      description: LAST finds & aligns related regions of sequences.
-      homepage: https://gitlab.com/mcfrith/last
-      documentation: https://gitlab.com/mcfrith/last/-/blob/main/doc/last-train.rst
-      tool_dev_url: https://gitlab.com/mcfrith/last
+  - "last":
```
SPPearce, line 11: Again, why do these have quote marks?

SPPearce:
I pushed a existence check on the png test.

modules/nf-core/last/dotplot/environment.yml
charles-plessy: To modernize the modules I deleted the files and ran `nf-core modules create` from scratch. This is where the quotes come from.

modules/nf-core/last/lastal/meta.yml
charles-plessy: Same reason here.  Please note that some other modules also have the quotes.

modules/nf-core/last/mafconvert/tests/main.nf.test.snap
SPPearce, line 1: This doesn't seem to have any content?

modules/nf-core/last/mafconvert/tests/main.nf.test
```diff
@@ -0,0 +1,35 @@
...
+    test("sarscov2 - bam") {
+
+        when {
+            process {
+                """
+                input[0] = [
+                    [ id:'contigs.genome' ], // meta map
+                    file(params.test_data['sarscov2']['genome']['contigs_genome_maf_gz'], checkIfExists: true)
+                ]
+                input[1] = [ 'psl' ]
```
SPPearce, line 21: You can just give strings like this without the square brackets btw.

SPPearce (commented):
I'm impressed by your dedication to updating all of them to nf-test.
Could you add a stub to each module too?

charles-plessy:
I could not generate stubs because their tests were always failing and I did not understand how to troubleshoot it.

SPPearce:
> I could not generate stubs because their tests were always failing and I did not understand how to troubleshoot it.

Did you add an actual stub to the main.nf?

charles-plessy:
Thank you.  I  will add the other stubs tomorrow.

modules/nf-core/last/mafconvert/tests/main.nf.test
SPPearce: Is it easy to catch that the program has failed in that way, and force the exitcode to not be 0?

charles-plessy:
I added stubs to all modules. I had difficulties with empty files created with `touch` with a name ending in .gz (which in principle should not exist since gzipping an empty file produces a non-empty file).  they seemt to cause `nf-core modules test` to always fail and strangely incorporate the full path to the file on our HPC cluster in the `snap` file. I circuumvented the problem by doing `echo stub | gzip --no-name` in these cases. Grepping the whole repo for the resulting MD5 sum shows that I was not the first to have this idea…

The tests pass on my side and some of the GitHub action failures appear to be related to transient network errors while downloading contents from raw.githubcontent, and some other errors I just do not understand them.

modules/nf-core/last/mafconvert/tests/main.nf.test
charles-plessy: I just tried `maf-convert` outside virtualisation and it exits with error status 1.  I do not know why it was 0 when run through Nextflow/SLURM/Singularity…

SPPearce:
> I added stubs to all modules. I had difficulties with empty files created with `touch` with a name ending in .gz (which in principle should not exist since gzipping an empty file produces a non-empty file). they seemt to cause `nf-core modules test` to always fail and strangely incorporate the full path to the file on our HPC cluster in the `snap` file. I circuumvented the problem by doing `echo stub | gzip --no-name` in these cases. Grepping the whole repo for the resulting MD5 sum shows that I was not the first to have this idea…
> 
> The tests pass on my side and some of the GitHub action failures appear to be related to transient network errors while downloading contents from raw.githubcontent, and some other errors I just do not understand them.

Ah yes, nf-test does not like fastq.gz files that aren't actually zipped.
Typically we do `echo "" | gzip > file.gz`, but your option looks fine too.

modules/nf-core/last/train/tests/main.nf.test.snap
```diff
@@ -0,0 +1,72 @@
...
+                "1": [
+                    "versions.yml:md5,c5578547acf9e77e1e8f6bf796e32ac2"
+                ],
+                "param_file": [
+                    [
+                        {
+                            "id": "contigs",
+                            "single_end": false
+                        },
+                        "contigs..des.train:md5,d41d8cd98f00b204e9800998ecf8427e"
```
SPPearce, line 23: This seems to be missing part of its name

SPPearce (commented):
Looking good, thank you for doing that too.
There seems to be an issue with `INDEX_NAME` in last train, if you can take a look at that please.

modules/nf-core/last/train/tests/main.nf.test.snap
charles-plessy: It is because the UNTAR module also returns a stub file and not the contents of the tarball, which does not cause failure because the files are not used by LAST_TRAIN when it runs as a stub. However, in that case, `    INDEX_NAME=\$(basename \$(ls $index/*.des) .des)` does not return the expected value, but `.des` instead, which is ugly but does not cause failure and is reproducible.  Shall we ignore it?  The index channel of LAST_TRAIN gets input stripped of its metadata because I was advised to do so when I created the module.

SPPearce: Oh, is it just in the stub? Didn't notice that.
You could just define explictly `INDEX_NAME=STUB` or something then

modules/nf-core/last/mafconvert/tests/main.nf.test
charles-plessy: Good catch!  I am all for setting pipefail, but isn't it already done at the fist line of `.command.sh`?
```
#!/bin/bash -euo pipefail
```

SPPearce: Yeah, I think it should be set anyway, but we could easily split this to be a separate line

SPPearce:
> Hi Simon I would like to object on the removal of the pipe on the `mafconvert` submodules. The files handled by this module can be gigabyte-scale after decompression, and IO and scratch storage space are often bottlenecks in my experience. I just tested to set `pipefail` and it works. How about I add it to all the LAST modules that pile their output to gzip?

Sure, that is fine, you're the one who actually knows what the tool does and how it is used, I'm just trying to help along the way.

charles-plessy:
OK, from my side I think that we are ready: I have no further changes to add, all discussions are resolved, the branch is up to date, and all CI tests are passing :tada:

SPPearce (approved):

SPPearce:
> OK, from my side I think that we are ready: I have no further changes to add, all discussions are resolved, the branch is up to date, and all CI tests are passing :tada: 

Great, I've pressed the merge button then. 
Thanks for your hard work on this.

charles-plessy:
Thanks to you too; it was a pleasure.

PR #5728, opened 2024-05-31
Adding CircularMapper/realignsamfile

jbv2:
Adds module for circularmapper realignsamfile.

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:


TCLamnidis:
Had to add entire BWA workflow in the tests >.< 
This SHOULD work, but nf-tests complains that the `.join` operation cannot happen as it only works on channels. Not sure why. 
I also tried applying a `map` to the `BWA_ALN.out.bam` channel, and add the fastq file in the middle of it, but that also complained.

modules/nf-core/circularmapper/realignsamfile/main.nf
```diff
@@ -0,0 +1,52 @@
...
+    output:
+    tuple val(meta), path("*_realigned.bam")    , emit: bam
+    path "versions.yml"                         , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}" // Module actually has a hard-coded output of {input}_realigned.bam
```
SPPearce, line 24: Can we mv the output file to include the prefix then.

modules/nf-core/circularmapper/realignsamfile/meta.yml
```diff
@@ -0,0 +1,58 @@
...
+      description: Realigned BAM file
+      pattern: "*.bam"
+  - versions:
+      type: file
+      description: File containing software versions
+      pattern: "versions.yml"
+authors:
+  - "@shyama-mama"
+  - "@jbv2"
+  - "TCLamnidis"
```
SPPearce, line 58: This is missing a @

tests/config/pytest_modules.yml
SPPearce, line 1: Don't want to readd the pytest for circulargenerator ;)

modules/nf-core/circularmapper/realignsamfile/main.nf
TCLamnidis: Added!

jfy133 (commented):
Looks like you still have pytest files still in there, you should remove the relevant files in 'tests/' and also the entry in the pytest_tags.yml (or whatever it is called, doing this from my phone)

Otherwise looks ok to me

modules/nf-core/circularmapper/realignsamfile/tests/nextflow.config
SPPearce, line 1: Do you need this config file anymore?

TCLamnidis: I guess strictly not, though it tested that the prefix was correctly picked up

SPPearce (approved):

SPPearce:
Looks like the renaming to use the prefix needs to be conditional on the name of the file

TCLamnidis:
indeed. good catch! Curious that it only got flagged on conda and not the other profiles 😅

SPPearce:
> indeed. good catch! Curious that it only got flagged on conda and not the other profiles 😅

Yeah, I don't know why exactly. It doesn't throw an error on my macbook terminal, so presumably something about the conda enviroment is set differently.

SPPearce (approved):

Issue #5735, opened 2024-06-01
[FEATURE] Add tracks and indicies to the input of igvreports


cmatKhan:
### Is your feature request related to a problem? Please describe

Add additional inputs to the first input channel to pass in tracks and track indicies.

https://github.com/nf-core/modules/blob/2a93f0d7e4d8d95a11a84172dcf5450d5263e702/modules/nf-core/igvreports/main.nf#L11

### Describe the solution you'd like

If the tracks list of paths is not null, it should be added to the create_report cmd as `--tracks file.bam refGene.txt.gz ...`.

If the user passes `--tracks ...` in the task.args, the argument created from the input tracks can replace `--track` so that the users parameterized tracks are retained.

### Describe alternatives you've considered

files may be passed in via task.args. But, that wouldn't allow passing in files that are generated by the pipeline in which this is used

### Additional context

the igvreports README says

"Although --tracks is optional, a typical report will include at least an alignment track (BAM or CRAM) file from which the variants were called."

cmatKhan:
resolved in #5736

PR #5736, opened 2024-06-01
igvreports: adding tracks and indicies to the input channel

cmatKhan:
See issue #5735

Tests and meta.yml are updated


modules/nf-core/igvreports/main.nf
```diff
@@ -8,7 +8,7 @@ process IGVREPORTS {
         'biocontainers/igv-reports:1.12.0--pyh7cba7a3_0' }"
 
     input:
-    tuple val(meta), path(sites)
+    tuple val(meta), path(sites), path(tracks), path(tracks_indicies)
```
Joon-Klaps, line 11: Why the tracks_indicies if you are not using it in the module? 
Does the tool assume it to be in the same dir?

cmatKhan: for BAM/CRAM at least, yes

modules/nf-core/igvreports/main.nf
```diff
@@ -22,11 +22,21 @@ process IGVREPORTS {
     def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
     def fasta = genomeFasta ? "--fasta ${genomeFasta}" : ""
+    // If tracks is not null, create a string of the track paths
+    def track_arg = tracks ? "--tracks "+ tracks.collect { it.toString() }.join(' ') : ""
+    // if "--tracks" is in the args, then add track_string immediately after it in
+    // the args string and set the trag_arg to ""
```
SPPearce, line 28:
```suggestion
    // the args string and set the track_arg to ""
```

modules/nf-core/igvreports/meta.yml
```diff
@@ -25,7 +25,19 @@ input:
...
       type: file
-      description: VCF, BED, MAF, BEDPE, or generic tab delimited file of genomic variant sites
+      description: |
+        VCF, BED, MAF, BEDPE, or generic tab delimited file of genomic variant sites
+  - tracks:
+      type: file
+      description: |
+        List of any set of files of the types that IGV can display,
+        eg BAM/CRAM, GTF/GFF, VCF, BED, etc
+  - tracks:
```
SPPearce, line 35:
```suggestion
  - tracks_indices:
```

modules/nf-core/igvreports/main.nf
```diff
@@ -8,7 +8,7 @@ process IGVREPORTS {
         'biocontainers/igv-reports:1.12.0--pyh7cba7a3_0' }"
 
     input:
-    tuple val(meta), path(sites)
+    tuple val(meta), path(sites), path(tracks), path(tracks_indicies)
     path genomeFasta //optional genome fasta file
```
SPPearce, line 12: Can we not use camelCase please, use snake_case.
I think this should just be fasta though, to be consistent with everything else (will need changing elsewhere, and the command changing to e.g. fasta_command)
```suggestion
    path fasta 
```

modules/nf-core/igvreports/tests/main.nf.test
```diff
@@ -38,6 +39,37 @@ nextflow_process {
     }
 
 
+    test("testvariantsWithFastaAndTracks - vcf") {
```
SPPearce, line 42:
```suggestion
    test("variants with fasta and tracks") {
```
(and the one above I can't comment on)

modules/nf-core/igvreports/tests/main.nf.test
```diff
@@ -38,6 +39,37 @@ nextflow_process {
...
+                        }
+                    }
+                }
+            )
+        }
+
+    }
+
 
     test("testvariantsWithGenome - vcf") {
```
SPPearce, line 74:
```suggestion
    test("vcf with genome") {
```

modules/nf-core/igvreports/tests/main.nf.test
cmatKhan: addressed in 56a451393

modules/nf-core/igvreports/tests/main.nf.test
cmatKhan: addressed in 56a451393

modules/nf-core/igvreports/tests/main.nf.test.snap
```diff
@@ -23,6 +35,18 @@
         },
         "timestamp": "2024-03-19T13:55:24.216860945"
     },
+    "testvariantsWithFastaAndTracks - vcf": {
```
SPPearce, line 38: This still has some of the old snapshots in it, can you regenerate it (with `nf-core modules test igvreports --update`)

SPPearce (approved):
I'm happy now, just please remove the obsolete snapshots before merging.

modules/nf-core/igvreports/tests/main.nf.test.snap
cmatKhan: whoops -- will do.

I am looking at the igvreports github readme again, and it says that the fasta should be indexed. I am going to add that, along with the metadata map to the `fasta` channel input, since that seems to be the current way of things.

SPPearce: Yeah, a meta map on the fasta channel is probably sensible, to be consistent with most other tools

PR #5738, opened 2024-06-02
Allow picard/collecthsmetrics to be feeded the same input bed file twice

rollf:
This PR allows to feed the same BED file twice as input to `picard/collecthsmetrics`. This is done by simply staging the file in two different sub directories.

## PR checklist

None of the check boxes apply.


SPPearce:
Would you be able to migrate this module to nf-test instead of pytest while you are changing it?

rollf:
@SPPearce I would do this. Can you just point me to an example of a pytest-based test?

SPPearce:
This is currently pytest, which means that the tests exist within `tests/modules/nf-core/picard/collecthsmetrics`. We are using nf-test now, where the tests would be in `modules/nf-core/picard/collecthsmetrics/tests`.
The workflow to migrate is:

1. Run `nf-core modules create picard/collecthsmetrics --migrate-pytest`, it will create the skeleton of all the files for you, as well as print the current tests to the terminal.
2. Edit `modules/nf-core/picard/collecthsmetrics/tests/main.nf.test`, adding in the required inputs for each test inside the blocks where the skeleton says 
```
            process {
                """
                // TODO nf-core: define inputs of the process here. Example:
                
                input[0] = file(params.modules_testdata_base_path + 'genomics/sarscov2/illumina/bam/test.single_end.bam', checkIfExists: true)
                """
            }
```
  This just directly uses `input[0], input[1]` etc for the multiple inputs for the module, so you don't need to have `input = ..., fasta = ...`, but is otherwise just copy & paste. 

3. Run `nf-core modules test picard/collecthsmetrics --update` to test the module, this will generate the snapshot files. If the output is not stable (which the pytests suggest is probably the case, [as they don't have md5sums in this file](https://github.com/nf-core/modules/blob/master/tests/modules/nf-core/picard/collecthsmetrics/test.yml) ), then you need to change the assertions to check e.g. contents, see https://nf-co.re/docs/contributing/nf-test/assertions.
4. Run `nf-core modules lint picard/collecthsmetrics` to lint the module.

Any issues please just shout.

rollf:
@SPPearce I have migrated the test to `nf-test`. I do get `nf-test` snapshot is not stable` critical error. I do not have experience with neither `picard` nor `nf-test`. Can you please have a look or point me to what I should do?

SPPearce:
The assertion needs to be changed then, to only check part of the file. Presumably it contains a timestamp or something.

SPPearce:
An example from a different module:
`                { assert path(process.out.log.get(0).get(1)).getText().contains("reads passed filter: 99") }`

rollf:
@SPPearce Thanks again for you input. I modified the tests and they are stable now. This PR is ready from my point of view.

modules/nf-core/picard/collecthsmetrics/tests/nextflow.config
SPPearce, line 1: This isn't required, given that it is empty.

SPPearce (approved):
Can you just remove the empty config (which isn't being loaded) and then we are good to go.

modules/nf-core/picard/collecthsmetrics/tests/nextflow.config
rollf: Done.

SPPearce:
Thanks for changing to nf-test as well as making the fix you were after!

PR #5754, opened 2024-06-05
Swap samtools collatefastq to nf-test

SPPearce:
Swapping samtools collatefastq to nf-test.
I also found that the pytests weren't actually using the files that they claimed to be, and that the interleaved function was actually producing a fastq file, not the fastq.gz file it claimed to be.


SPPearce:
Currently the interleaved option makes a fastq file, rather than a fastq.gz file.  
I have adjusted the name and the emit statement to match this, but probably wants to have a conditional gzip in the process to zip it up.

modules/nf-core/samtools/collatefastq/tests/nextflow.config
maxulysse, line 1: I think this file can be deleted

maxulysse:
minor comments

modules/nf-core/samtools/collatefastq/main.nf
```diff
@@ -52,4 +52,20 @@ process SAMTOOLS_COLLATEFASTQ {
         samtools: \$(echo \$(samtools --version 2>&1) | sed 's/^.*samtools //; s/Using.*\$//')
     END_VERSIONS
     """
+
+    stub:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def outputcommand =    (interleave && ! meta.single_end) ? "echo '' | gzip > ${prefix}_interleaved.fq.gz"                     :
+                    meta.single_end                   ? "echo '' | gzip > ${prefix}_1.fq.gz ; echo '' | gzip > ${prefix}_singleton.fq.gz" :
+                    "echo '' | gzip > ${prefix}_1.fq.gz ; echo '' | gzip > ${prefix}_2.fq.gz ; echo '' | gzip > ${prefix}_singleton.fq.gz"
```
nvnieuwk, line 60: Can you make this a bit more readable? :grin:

modules/nf-core/samtools/collatefastq/tests/main.nf.test.snap
```diff
@@ -0,0 +1,188 @@
...
+        "content": [
+            [
+                
+            ],
+            [
+                
+            ],
+            [
+                
+            ],
```
nvnieuwk, line 12: The stub doesn't output anything, this doesn't seem right?

SPPearce:
@nvnieuwk, I have cleaned up the stub if you want to take a look.

nvnieuwk (approved):
Much better!

PR #5758, opened 2024-06-05
Revert "nf-test bamaligncleaner + test out nft-bam"

k1sauce:
Reverts nf-core/modules#5755

I am reverting this PR because the nft-bam plugin is causing an error with the nf-test CI workflows. 

For example see https://github.com/nf-core/modules/actions/runs/9388939895/job/25855343012?pr=5720

image

The plugin is included in the nf-test config here https://github.com/nf-core/modules/blob/1cacaceabae75b0c3bc393dee52cb6a5020fcb5c/nf-test.config#L16C9-L16C29


nvnieuwk:
This PR should not be necessary once https://github.com/askimed/nf-test-plugins/pull/4 is merged

PR #5768, opened 2024-06-06
Renovate DRY

edmundmiller:
Removes all the stuff that's duplicated at https://github.com/nf-core/ops/blob/main/.github/renovate/default.json5

> Even if you automerge PRs, you are likely to still get notification noise - one when the PR is created and another when it is merged. For this reason we recommend you consider setting automergeType=branch which will mean:

>   - Renovate first creates a branch and no PR
>   - If tests pass, Renovate pushes a commit directly to the base branch without PR
>   - If tests fail, Renovate raises a PR for you to review

> Add the renovate/** branch to your testing workflow files, or Renovate will not work properly with the automergeType=branch setting.


mashehu:
We don't allow direct pushes to the dev branch, so not sure if that automerge here will work.

edmundmiller:
I think we can all renovate both to by pass that(according to their docs)

edmundmiller:
image
This is what we'd need to enable(didn't save the config)

SPPearce:
So are we planning on changing the settings?

edmundmiller:
Unless anyone states otherwise, but I'd need someone else to at-least approve the PR.

mirpedrol:
Where are we setting `automergeType=branch`?

mashehu:
I personally don't like an automerge for version updates. I would still prefer a PR per update even if we always accept them. But it allows a bit better oversight what has been changed.

SPPearce:
If it can stop spamming PRs like #5757 that'd be good

edmundmiller:
Okay, since we've finally got all of the things in place I'm starting to work through these one step at a time. 

I don't think automerge was enabled a few months ago when I made this PR. To clarify I've specifically turned them **off** for all of the module software version bumps. I think that's the best approach for now, and we can address it in the @nf-core/maintainers meeting after a few months when everything's rolled out and we're tired of clicking approve.

This should automerge minor version bumps for other versions like GitHub actions if the tests pass.

maxulysse (approved):

mashehu (approved):

PR #5777, opened 2024-06-07
Bump plugins

edmundmiller:
Waiting for #5768 to merge first.

Wondering if we can get away with just this or if I need to add the whole section to the renovate config...

https://docs.renovatebot.com/modules/manager/regex/#advanced-capture


nvnieuwk:
Is it possible to check the version here instead? https://github.com/askimed/nf-test-plugins/blob/main/plugins.json

The newer plugin version will only be available if it's been added to the plugin repository so it doesn't really make sense to it on a plugin release

edmundmiller:
We tried to parse it, it's technically possible. 

But if this is only running once a week the disconnect between the gh release and the json getting updated probably won't matter.

nvnieuwk:
Allright fine by me then :p

luisas:
@edmundmiller Bump :)

edmundmiller:
Closing — stale, depends on unmerged #5768, conflicts.

PR #5779, opened 2024-06-08
Update kmcp index to pass linting

SPPearce:
Not sure how, but kmcp index is failing linting because it doesn't have snapshot the versions file in one test.


jfy133 (approved):
I don't get that either...

PR #5780, opened 2024-06-08
Update nonpareil tests and linting

SPPearce:
Tests for nonpareil seem to be [failing currently]((https://github.com/nf-core/modules/actions/runs/9427296179/job/25971253841?pr=2703), not sure why they are running on that), as does the linting, so lets fix that up.


jfy133 (approved):
Wow I did a really bad job when I wrote those... Sorry about that....

Issue #1754, opened 2021-02-19
[FEATURE] Fix bioconda recipe for deepvariant


maxulysse:
Latest release is [`1.1.0`](https://github.com/google/deepvariant/releases/tag/v1.1.0).
Latest available on bioconda is: [`1.0.0`](https://bioconda.github.io/recipes/deepvariant/README.html)

SPPearce:
Anyone know if this is fixed yet?

famosab:
I would say yes: https://anaconda.org/bioconda/deepvariant

vagkaratzas:
I see [v1.9.0](https://bioconda.github.io/recipes/deepvariant/README.html#package-package%20&#x27;deepvariant&#x27;) now on bioconda (1.9.0--pyh697b589_0 for containers), so `deepvariant` modules can now be updated to work with conda, hopefully.

maxulysse:
That's an amazing news

PR #5801, opened 2024-06-12
build: Bump deepvariant to 1.6.1

edmundmiller:
[Slack thread](https://nfcore.slack.com/archives/CJRH30T6V/p1718189925184689)

https://github.com/nf-core/modules/issues/1754 is why we're still manually bumping this.


maxulysse (approved):

modules/nf-core/deepvariant/main.nf
```diff
@@ -38,7 +40,7 @@ process DEEPVARIANT {
         --output_gvcf=${prefix}.g.vcf.gz \\
         ${args} \\
         ${regions} \\
-        --intermediate_results_dir=. \\
+        --intermediate_results_dir=intermediate_results \\
```
maxulysse, line 43:
```suggestion
        --intermediate_results_dir=tmp \\
```

fellen31:
Don't merge this yet, the version says 1.6.0.

fellen31:
@edmundmiller So even though the code in this docker is from 1.6.1, the output will say 1.6.0. The developers don't seem very keen on building a new image themselves. Would you either be able to: 
```
# $ git clone https://github.com/google/deepvariant.git
# $ cd deepvariant
```
Manually update `ARG VERSION=1.6.0` to `ARG VERSION=1.6.1` in the Dockerfile before building (confirmed this is the only thing needed in the linked issue above)
```
# $ sudo docker build -t deepvariant .
```

Or use the docker at `docker.io/fellen31/deepvariant:1.6.1` where I have done the above?

edmundmiller:
I think we shouldn't modify the Docker image and update the version inside for "reproducibility issues". Wouldn't want people to get upset because the hashes don't match or something 🙃.

fellen31:
So just double checking, you are happy with having `nf-core/deepvariant:1.6.1` writing version 1.6.0 in the output VCF?

maxulysse:
I think, it's a known issue, so I'm happy enough, that's what the tool itself report, we could have add an env and report 1.6.1 instead to be fair.
@fellen31 would you rather we do that?

fellen31:
I’d argue that for the end user it might be more confusing to have a mismatch in versions between a multiqc report (1.6.1) and vcf file (1.6.0) using the 1.6.1 docker image (like now), than having 1.6.0 written in both multiqc and vcf, even though it’s really 1.6.1. I can always patch with my own docker, so I’ll leave it up to you or @ramprasadn. As long as you are aware if you update Sarek or raredisease.

maxulysse:
good point, I'll create a PR then

maxulysse:
It is already done actually: `def VERSION = '1.6.1'`, and that's what we report

fellen31:
> It is already done actually: `def VERSION = '1.6.1'`, and that's what we report

Yea, but the VCF header would still say 1.6.0. But maybe nobody looks 😄

maxulysse:
Oh I see.

Can we just edit the vcf.gz header and change that then?

maxulysse:
I agree with @edmundmiller I don't want to modify the container itself

edmundmiller:
Ah, I was afraid it might mess with the output 😞 

I think we should modify the container then. If DeepVariant is fine with confusing people so be it.

We could also just use `1.6.0` and skip `1.6.1` because they refuse to push a new container, and anyone with the sharding issue can manually update the container?

maxulysse:
I don't really have a preference there...
We cannot fix everything at our level

Issue #4234, opened 2023-10-25
Give conda environments a name


Emiller88:
> Not seen [this error](https://github.com/nf-core/modules/actions/runs/6635360127/job/18030179941?pr=4225) before:
  java.nio.channels.OverlappingFileLockException
is that related to the conda environment change?

I see this error appears when running tests for subworkflows. Could be that we need to add a name to `environment.yml`? If two modules in a subworkflow have exactly the same `environment.yml` I suspect the environment names may clash

_Originally posted by @mirpedrol in https://github.com/nf-core/modules/issues/4225#issuecomment-1778842964_

Emiller88:
We need to decide how we're going to name them.

There's a few with a name already:
```yaml
name: nf-core-vep-108.2
name: nf-core-snpeff-5.1
```

However, I'd vote we just keep it in line with the `name` field in the `meta.yml`. Easy to lint, no `/`'s or extra nf-core seperators.

PR #5802, opened 2024-06-12
build: Remove conda env name

edmundmiller:
Originally trying to reproduce the error in https://github.com/nf-core/modules/issues/4234#issue-1961683896

It did get produced. 

Figured the easiest path is actually to just stop testing conda and pytest-workflow together. I think it's a worthy sacrifice to move this forward, since we're phasing pytest-workflow out soon(™️)

This PR removes a few env names, and changes the linting to not require the name field, and removes the `defaults` channel (Take it up with @pditommaso if you liked the `defaults` channel)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -456,6 +456,7 @@ jobs:
       - uses: mamba-org/setup-micromamba@v1
         with:
           init-shell: bash
+          micromamba-binary-path: ${{ runner.temp }}/bin/micromamba
 
       # Set up secrets
       - name: Set up nextflow secrets
@@ -686,6 +687,7 @@ jobs:
       - uses: mamba-org/setup-micromamba@v1
         with:
           init-shell: bash
+          micromamba-binary-path: ${{ runner.temp }}/bin/micromamba
 
       # Set up secrets
       - name: Set up nextflow secrets
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -173,8 +173,6 @@ jobs:
 
       - name: Setup Nextflow
         uses: nf-core/setup-nextflow@v2
-        with:
-          version: latest-everything
 
       - name: Install nf-core tools development version
         run: python -m pip install --upgrade --force-reinstall git+https://github.com/nf-core/tools.git@dev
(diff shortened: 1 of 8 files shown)


edmundmiller:
https://github.com/nextflow-io/nextflow/issues/4464#issuecomment-2160348935 Might be relevant

edmundmiller:
Removed conda pytest-workflow tests from running. 

It creates a race condition with the environment creation that it appears nf-test doesn't run into.

Keeping it prevents us from removing the env name from the `environment.yml` files, which is preventing us from having consistent image naming with Seqera containers, which is preventing us from building our "mulled" containers with Wave and using the Seqera container registry.

mirpedrol (commented):
If I am following this correctly, we not only want `name` to not be required but we want to avoid having a `name` in `environment.yml`, is this right? In such case we should add a listing test for this.

Also, would it be better to remove the name progressively with PRs where we start using wave?

Apart from that, I agree with not using conda with pytest.

maxulysse:
> If I am following this correctly, we not only want `name` to not be required but we want to avoid having a `name` in `invironment.yml`, is this right? In such case we should add a listing test for this.
> 
> Also, would it be better to remove the name progressively with PRs where we start using wave?
> 
> Apart from that, I agree with not using conda with pytest.

so basically the opposite of the current linting?

mirpedrol:
> so basically the opposite of the current listing?

Yes! We can try to add a `not` in the JSON schema
```
"required": ["channels", "dependencies"],
"not": { "required": [ "name" ] }
```

ewels:
The main problem I see with this is that the error pytest-workflow is throwing isn't limited to _just_ pytest-workflow. It's also a real-life bug that can hit end users. So by removing `name` we potentially open up conda users to hitting the same error.

The workaround is not too difficult: use mamba. But it's not ideal. But I can't think of any way around it, given that we need to remove `name` for the Wave generated container name consistency..

edmundmiller:
So the person who's going to run into issues with this:
1. Has multiple users/pipelines going on the same machine
2. Is using Conda
3. Can't use mamba(mambaforge should work for just about any user space)

The other option is we convince the Seqera containers people to change their default so we can keep our environment names.

The issue isn't really with Nextflow either, it's a conda problem.

I think the point I'm trying to make is anyone who has those three requirements is pretty niche. I want nf-core to try to support everyone, but that's a pretty inflexible system, and we would have provided that person with multiple solutions.

Another thought, what if we changed the environment names that Nextflow autogenerated to match the wave default? Like a spike-in at runtime. That should work in theory if it works when we give it an environment name. Could be a backlog item and we see if our stubborn conda users with multiple people running it come out of the woodwork...

ewels:
Note that the conda error should be fine with micromamba. Currently some issues with that (see https://github.com/nextflow-io/nextflow/issues/4161 ) but hoping to get fixed soon.

So maybe we can continue testing with conda without this problem by switching to mamba, if needed.

PR #5428, opened 2024-04-04
New module: gnu/split

k1sauce:
## PR checklist

Closes #5415  

- [X] This comment contains a description of changes (with reason).
- [X] If you've fixed a bug or added code that should be tested, add tests!
- [X] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [X] Remove all TODO statements.
- [X] Emit the `versions.yml` file.
- [X] Follow the naming conventions.
- [X] Follow the parameters requirements.
- [X] Follow the input/output options guidelines.
- [X] Add a resource `label`
- [X] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - For modules:
    - [X] `nf-core modules test <MODULE> --profile docker`
    - [X] `nf-core modules test <MODULE> --profile singularity`
    - [X] `nf-core modules test <MODULE> --profile conda`
  - For subworkflows:
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile docker`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile singularity`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile conda`

PR #5805, opened 2024-06-14
Update gnu_sort dependency versions

nictru:
The primary goal is to fix the upcoming problems with old docker images not being compatible with current docker versions. I took inspiration from [gnu_split](https://github.com/nf-core/modules/tree/master/modules/nf-core/gnu/split), which has the same dependencies.

### DIFF
--- a/modules/nf-core/gnu/sort/environment.yml
+++ b/modules/nf-core/gnu/sort/environment.yml
@@ -4,4 +4,4 @@ channels:
   - bioconda
   - defaults
 dependencies:
-  - bioconda::coreutils=8.25
+  - bioconda::coreutils=9.3
--- a/modules/nf-core/gnu/sort/environment.yml
+++ b/modules/nf-core/gnu/sort/environment.yml
@@ -1,7 +1,9 @@
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/environment-schema.json
 name: gnu_sort
 channels:
   - conda-forge
   - bioconda
   - defaults
 dependencies:
-  - bioconda::coreutils=9.3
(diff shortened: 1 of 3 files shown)


modules/nf-core/gnu/sort/environment.yml
```diff
@@ -1,7 +1,9 @@
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/environment-schema.json
 name: gnu_sort
 channels:
   - conda-forge
   - bioconda
   - defaults
 dependencies:
-  - bioconda::coreutils=8.25
+  - conda-forge::coreutils=9.3
```
pinin4fjords, line 9: Why the conda forge one? It's better from a Conda perspective, but it takes things out of synch with Docker. I think it would be best to bump it but keep to the Bioconda one.

pinin4fjords (changes requested):
LGTM- nice tidy-up!

modules/nf-core/gnu/sort/environment.yml
nictru: The problem here is that the package is not in bioconda any more (does not exist in the repo). I don't know how the biocontainers received the update to 9.3, but the latest version in [bioconda](https://anaconda.org/bioconda/coreutils) is 8.25.

nictru: I guess they did this for the same reason in [gnu_split](https://github.com/nf-core/modules/tree/master/modules/nf-core/gnu/split), but it was not discussed in #5428

pinin4fjords: Ahh, as you were then :-)

pinin4fjords (approved):

PR #5827, opened 2024-06-18
Conda lock files

edmundmiller:
- **build: Add wave**
- **build: Set strategy to dockerfile, conda then container**
- **refactor: Remove container**
- **build: Add a repo to push to**
- **ci(wave): Add wave build**
- **ci(wave): Switch to all_changed_files**
- **ci(wave): Only look for envronment.ymls**
- **dummy: Change env**
- **ci(wave): Remove raw format**
- **ci(wave): Try a bunch of different things at once**
- **ci(wave): Remove redundant fromJson and wrap in an array**
- **ci(wave): I have no idea what I'm doing**
- **ci(wave): Wrap it**
- **ci(wave): Found an example**
- **ci(wave): Maybe quotes?**
- **ci(wave): That'll do it**
- **ci(wave): Fix wave install**
- **ci(wave): Hard code an image**
- **ci(wave): Add secrets**
- **feat: Try a different files structure**
- **ci(wave): First stab at building singularity images**
- **fixup! feat: Try a different files structure**
- **ci(wave): Add profile to matrix**
- **ci(wave): Give up on fancy substitution**
- **ci(wave): Add await**
- **ci(wave): Switch to quay**
- **test(wave): Add freeze and update build repo**
- **refactor(wave): What happens if I add a container?**
- **refactor(wave): Have both bowtie modules use the same env**
- **test: Cut out using wave on tests**
- **refactor: What happens if we use the singularity one?**
- **refactor: Keep container directives for offline download**
- **feat: Try new singularity OCI setting**
- **build: Update container name**
- **chore: Bump wave-cli version**
- **ci: Install runc**
- **ci: Switch to singularityhub action**
- **ci: Install new singularity manually**
- **ci: Install dependancies for singularity**
- **ci: runc => crun**
- **ci: Fix cgroup error**
- **ci: That'll do it**
- **ci: Remove Dockerfile**
- **ci: Update name**
- **ci: Push to the correct repos**
- **ci: Remove OCI stuff**
- **ci: Need a full URL**
- **ci: Fix // in container name**
- **ci: Remove push**
- **build: Add containers back**
- **ci: Add cache repos**
- **ci: Change registry name to use _**
- **build: / => _ in container name**
- **Try ociAutoPull**
- **chore: Add renovate comments to samtools**
- **test: Add ociAutoPull to nf-test**
- **ci: Bump wave version**
- **chore: Bump containers with new wave version**
- **build: Update to use commity.wave.seqera.io**
- **ci: Bump wave-cli to 1.4.1**
- **ci: Try apptainer**
- **ci: Remove build-repo to see what happens**
- **build: Bump Nextflow version requirement**
- **fix: Get rid of the environment name?**
- **ci: Bump action versions**
- **ci: Try name-strategy tagPrefix**
- **ci: Remove singularity build for now**
- **ci: Try imageSuffix**
- **ci: Try none**
- **ci: What is the bowtie container name**
- **ci: Remove --name-strategy**
- **style: Add back in container elvis operator**
- **ci: Remove cache repo**
- **Revert "build: Bump Nextflow version requirement"**
- **Revert "test: Add ociAutoPull to nf-test"**
- **ci(#2193): Add condalock file generation**


edmundmiller:
Status update:

This PR doesn't really do anything for us until we adopt wave. 

Main issue was the GitHub action can't commit the lock file back for whatever permissions issue.

pinin4fjords:
This is pretty frickin' awesome, lock files have been on my xmas list for ages. 

My only questions surround how these lock files will actually get used. As-is. the `conda` and `container` directives are unchanged for this module, so behaviour in use will be unchanged.

We spoke on Slack about how I thought (and others disagreed) that we should also change the container directive to use what Wave spits back. I also wonder how we could make it so that people deploying via Conda could get a static environment built off the lock file?

modules/nf-core/bowtie2/align/main.nf
```diff
@@ -2,7 +2,7 @@ process BOWTIE2_ALIGN {
     tag "$meta.id"
     label 'process_high'
 
-    conda "${moduleDir}/environment.yml"
+    conda "${moduleDir}/conda-lock.yml"
```
pinin4fjords, line 5: Nextflow can work with conda-lock outputs natively? I didn't even know.

pinin4fjords: Ahh, it can't- just checked the test.

pinin4fjords, line 5:
```suggestion
    conda "${moduleDir}/conda-lock.txt"
```

pinin4fjords:
@adamrtalbot raises the excellent point that lock files will be architecture-specific. So maybe what we need is actually a conda-lock-amd64.txt, conda-lock-arm64.txt? Then workflows would need to be run with architecture-specific configurations referencing these?

pinin4fjords:
> Can you embed multiple architectures in a lock file? The docs kinda say you can: https://github.com/conda/conda-lock?tab=readme-ov-file#platform-specification

YAAAAS! 

```
 conda-lock -f environment.yml -p osx-64 -p linux-64
Spec hash already locked for ['osx-arm64']. Skipping solve.
Locking dependencies for ['linux-64', 'osx-64']...
INFO:conda_lock.conda_solver:linux-64 using specs ['bioconda::multiqc=1.23']
...
```

Good man

.github/workflows/wave.yml
```diff
@@ -0,0 +1,87 @@
...
+        id: lockfile-name
+        with:
+          result-encoding: string
+          script: |
+            return '${{ matrix.files }}'.replace('/environment.yml', '/conda-lock.yml');
+
+      - name: Run conda-lock
+        run: |
+          rm --force "${{steps.lockfile-name.outputs.result}}"
+          conda-lock lock $MAMBA --file "${{ matrix.files }}" --kind lock --platform linux-64 --lockfile "${{steps.lockfile-name.outputs.result}}"
```
pinin4fjords, line 66: So, I tested why the lock file didn't work, and I think it's because of the `--platform` here. To get a lock file compatible with Nextflow (which just does [like this](https://github.com/nextflow-io/nextflow/blob/5eeed5bf096c05d77c79863a6d9e48d09cdd5fcc/modules/nextflow/src/main/groovy/nextflow/conda/CondaCache.groovy#L280-L297)), you have to render it to a platform specific one:

```
conda-lock render -p osx-64  
Rendering lockfile(s) for osx-64...
 - Install lock using : conda create --name YOURENV --file conda-osx-64.lock
```

So for nextflow as-is, we will need platform-specific lockfiles.

pinin4fjords:
> > @adamrtalbot raises the excellent point that lock files will be architecture-specific. So maybe what we need is actually a conda-lock-amd64.txt, conda-lock-arm64.txt? Then workflows would need to be run with architecture-specific configurations referencing these?
> 
> Can you embed multiple architectures in a lock file? The docs kinda say you can: https://github.com/conda/conda-lock?tab=readme-ov-file#platform-specification

@adamrtalbot unfortunately I just discovered that the reason @edmundmiller 's new conda declaration didn't work is probably because of the usage of `--platform`.

.github/workflows/wave.yml
pinin4fjords, line 66:
```suggestion
          conda-lock lock $MAMBA --file "${{ matrix.files }}" --kind explicit --platform linux-64 --lockfile "${{steps.lockfile-name.outputs.result}}"
          mv  conda-linux-64.lock conda-lock-linux-64.yml
```

Nextflow needs it to be a .yml, but use of `--kind explicit` means we get .lock files, so we'll need to do some file renaming.

Platforms should probably be set in the environment.ymls actually.

.github/workflows/wave.yml
```diff
@@ -0,0 +1,87 @@
...
+
+      - name: Commit & push changes
+        id: commit-and-push
+        run: |
+          git config user.email "core@nf-co.re"
+          git config user.name "nf-core-bot"
+          git config push.default upstream
+          git add .
+          git status
+          git commit -m "[automated] autogenerated conda-lock file"
```
pinin4fjords, line 79:
```suggestion
          git commit -m "[automated] autogenerated conda-lock files"
```

modules/nf-core/bowtie2/align/main.nf
pinin4fjords, line 5:
```suggestion
    conda "${moduleDir}/conda-lock-linux-64.yml"
```

.prettierignore
```diff
@@ -15,3 +15,4 @@ __pycache__
 *.pyo
 *.pyc
 .github/renovate.json5
+**/conda-lock.yml
```
pinin4fjords, line 18:
```suggestion
**/conda-lock*.yml
```

pinin4fjords (commented):
So basically I think we'll need to do like this

modules/nf-core/bowtie2/align/main.nf
pinin4fjords: (I'm obviously not proposing we actually change modules for platform specificity, just illustrating).

.github/workflows/wave.yml
edmundmiller: It should just be a single commit per lock file I think?

modules/nf-core/bowtie2/align/main.nf
edmundmiller: I think tracking them at this point might keep us sane though 😅

.github/workflows/wave.yml
edmundmiller: Doesn't look like that works 😞

pinin4fjords: The `explicit` thing is important. It makes you a lock file for each platform which is consumable by `conda env create`, and therefore by Nextflow.

But by doing `explicit` you lose control over the naming of the multiple outputs, and Nextflow needs `.yml`, hence the rename hack I suggested.

.github/workflows/wave.yml
pinin4fjords: One lock file per platform

modules/nf-core/bowtie2/align/main.nf
pinin4fjords: Because the lock files will be platform specific they can't go directly on the modules. I think we'd need platform-specific configs at the workflow level that referenced all the lockfiles.

luisas:
@edmundmiller Is this PR still relevant? :)

ewels:
TODO:

* Don't use lock files for the module `conda` directive, stick with `environment.yml`
* Save the conda lock files for each arch in `.conda` directory for the module
* Link to these files in the `meta.yml` file for the module

ewels:
Nextflow upstream should now be able to handle lock files with any extension (based on file contents instead). Not sure that it affects things here, but thought it good to mention.

edmundmiller:
Closing — 20 months old, conflicts, stale.

PR #5837, opened 2024-06-19
update JSON schema for modules meta.yml

mirpedrol:
Updating the JSON schema used to lint modules meta.yml files according to the changes which will be introduced by https://github.com/nf-core/tools/pull/3028


modules/meta-schema.json
```diff
@@ -182,5 +120,44 @@
...
+        "elementProperties": {
+            "type": "object",
+            "patternProperties": {
+                ".*": {
+                    "type": "object",
+                    "properties": {
+                        "type": {
+                            "type": "string",
+                            "description": "Type of the channel element",
+                            "enum": ["map", "file", "directory", "string", "integer", "float", "boolean", "list"]
```
mashehu, line 133: are these all nextflow types?

mirpedrol: Those are the types that we define, same as before :) The nextflow "type" I called it qualifier. 
Can be discussed if we need this or not, because it can be inferred from the `main.nf` or from the ontology (once we add it)

mashehu: I would prefer nextflow types, especially if we get type hinting in nextflow in the future. Or what do you think, @ewels ?

mirpedrol: I converted the PR to a draft because the tools PR is not ready anyway. And we can keep discussing this. 
I think @ewels was against both `type` and `qualifier` in the `meta.yml` 😄

mashehu: Hmm, at least one of them would make the move over to typed nextflow easier I think...

mirpedrol: Removed qualifier after [some discussion](https://github.com/nf-core/modules/pull/5867#discussion_r1656681444)

nvnieuwk (approved):
No blocking issues as far as I can see :). It would maybe be nice if there would be a native way of specifying which outputs are optional, but that can be in the description for now

JoseEspinosa (approved):
LGTM! though another pair of 👀 might be good

mirpedrol:
thank you!

PR #5867, opened 2024-06-21
POC of new meta.yml structure and ontologies

mirpedrol:
POC for https://github.com/nf-core/tools/pull/3028 and https://github.com/nf-core/tools/pull/3032

Run `nf-core modules lint <module_name> --fix` for modules: 
- `fastp`
- `samtools/sort`
- `samtools/index` 
- `atlas/splitmerge`
- `pear`
- `multiqc`
- `bwa/mem`

Note that linting won't pass until https://github.com/nf-core/modules/pull/5837 is merged

ℹ️ The bio.tools ID was not automatically found for `bwa`. 
ℹ️ The modules `samtools/index`, `atlas/splitmerge` and `bwa/mem` had wrong channels specified or missing inputs/outputs, thus the resulting modified `meta.yml` had to be fixed manually.

Ontology for input and output files were added for `bwa/mem` as an example.

### DIFF
--- a/modules/nf-core/atlas/splitmerge/meta.yml
+++ b/modules/nf-core/atlas/splitmerge/meta.yml
@@ -7,58 +7,71 @@ keywords:
   - read group
 tools:
   - "atlas":
-      description: "ATLAS, a suite of methods to accurately genotype and estimate genetic diversity"
+      description: "ATLAS, a suite of methods to accurately genotype and estimate genetic
+        diversity"
       homepage: "https://bitbucket.org/wegmannlab/atlas/wiki/Home"
       documentation: "https://bitbucket.org/wegmannlab/atlas/wiki/Home"
       tool_dev_url: "https://bitbucket.org/wegmannlab/atlas"
       doi: "10.1101/105346"
-      licence: "['GPL v3']"
+      licence: ["GPL v3"]
+      identifier: biotools:atlas_db
 input:
-  - meta:
-      type: map
-      description: |
-        Groovy Map containing sample information
-        e.g. [ id:'test', single_end:false ]
-  - bam:
-      type: file
-      description: Single input BAM file.
-      pattern: "*.bam"
-  - bai:
-      type: file
-      description: The BAI file for the input BAM file
-      pattern: "*.bai"
-  - read_group_setting:
-      type: file
-      description: |
-        TXT file containing the split and merge settings for
-        each readgroup. Each line consist of one readgroup,
-        single/double identifier and the maximum cycle number
-        of the sequencer. e.g. "RG1 single 100"
-      pattern: "*.txt"
-  - blacklist:
-      type: file
-      description: |
-        blacklist.txt (optional), A txt file with blacklisted read names
-        that should be ignored and just written to file, each on a new line
-      pattern: "*.txt"
+  - - meta:
+        qualifier: val
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. [ id:'test', single_end:false ]
+    - bam:
+        qualifier: path
+        type: file
+        description: Single input BAM file.
+        pattern: "*.bam"
+    - bai:
+        qualifier: path
+        type: file
+        description: The BAI file for the input BAM file
+        pattern: "*.bai"
+    - read_group_settings:
+        type: file
+        description: |
+          TXT file containing the split and merge settings for
+          each readgroup. Each line consist of one readgroup,
+          single/double identifier and the maximum cycle number
+          of the sequencer. e.g. "RG1 single 100"
+        pattern: "*.txt"
+        qualifier: path
+    - blacklist:
+        qualifier: path
+        type: file
+        description: |
+          blacklist.txt (optional), A txt file with blacklisted read names
+          that should be ignored and just written to file, each on a new line
+        pattern: "*.txt"
 output:
-  - meta:
-      type: map
-      description: |
-        Groovy Map containing sample information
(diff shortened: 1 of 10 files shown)


modules/nf-core/atlas/splitmerge/meta.yml
```diff
@@ -7,58 +7,71 @@ keywords:
...
-        Groovy Map containing sample information
-        e.g. [ id:'test', single_end:false ]
+  - data:
+      - meta:
+          qualifier: val
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. [ id:'test', single_end:false ]
+      - "*_mergedReads.bam":
```
SPPearce, line 59: The output channels seem to have lost their names.

mirpedrol: The name of the output channel is `data` in [this case](https://github.com/nf-core/modules/blob/master/modules/nf-core/atlas/splitmerge/main.nf#L14). And we specify each element with an array, here we have `meta` and two files. To make it as automated as possible, we are copying what's written in the `main.nf` as output, which is the pattern to the files.
I agree it's not the prettiest, specially in some cases where this regex can get more complicated, but I think it's the best way to do it, instead of having custom names which can't be read automatically from the `main.nf` file.

SPPearce: Ok, I see

modules/nf-core/atlas/splitmerge/meta.yml
```diff
@@ -7,58 +7,71 @@ keywords:
...
-      description: "ATLAS, a suite of methods to accurately genotype and estimate genetic diversity"
+      description: "ATLAS, a suite of methods to accurately genotype and estimate genetic
+        diversity"
       homepage: "https://bitbucket.org/wegmannlab/atlas/wiki/Home"
       documentation: "https://bitbucket.org/wegmannlab/atlas/wiki/Home"
       tool_dev_url: "https://bitbucket.org/wegmannlab/atlas"
       doi: "10.1101/105346"
-      licence: "['GPL v3']"
+      licence: ["GPL v3"]
+      identifier: biotools:atlas_db
```
ewels, line 17: Should just be `atlas`, not `atlas_db` in this case: https://bio.tools/atlas

modules/nf-core/atlas/splitmerge/meta.yml
```diff
@@ -7,58 +7,71 @@ keywords:
...
-      pattern: "*.txt"
-  - blacklist:
-      type: file
-      description: |
-        blacklist.txt (optional), A txt file with blacklisted read names
-        that should be ignored and just written to file, each on a new line
-      pattern: "*.txt"
+  - - meta:
+        qualifier: val
+        type: map
```
ewels, line 21: @bentsherman - does this look about right to you, in terms of what we call things here?

bentsherman: what's the use of adding the qualifier when you already have the type?

mirpedrol: The qualifier can be useful to know if two modules have the same inputs and outputs. 
For example, we could have the case where two different outputs have type string, but the qualifier is val and env.

bentsherman: You mean to check if the modules can be chained? In that case it doesn't matter if it a `val` or `env`, as long as they are both strings

mirpedrol: I am more thinking of a setting where modules can be interchangeable, for example for benchmarking of different tools. I would like to make sure that using aligner1 and using aligner2, both inputs and outputs will be exactly the same.
I know this can already be done with if/else and operating the channels if needed. But in the future, it would be cool if we can use this to automatically add any module which matches the requirements.

bentsherman: Even there it doesn't matter if the string input is a `val` or `env`, as long as it is a string they are interchangeable.

The only difference is that val says "provide this value as a variable" whereas env says "provide this value as an environment variable", but those details are internal to the process

mirpedrol: OK, I see what you mean! Would it be best to remove the `qualifier`?

bentsherman: yes I think so

modules/nf-core/atlas/splitmerge/meta.yml
```diff
@@ -7,58 +7,63 @@ keywords:
...
-        of the sequencer. e.g. "RG1 single 100"
-      pattern: "*.txt"
-  - blacklist:
-      type: file
-      description: |
-        blacklist.txt (optional), A txt file with blacklisted read names
-        that should be ignored and just written to file, each on a new line
-      pattern: "*.txt"
+  - - meta:
+        type: map
```
ewels, line 20: I think in the comments we said that we'd remove this, right? But it seems to still be in the diff.

ewels: Ah wait ignore me, it was a key called `qualifier`. Sorry.

ewels (approved):
Let's do it!

ewels:
Yeah, now you have merge conflicts 😓  Sorry @mirpedrol !

mirpedrol:
I have updated the branch and solved conflicts for all modules here https://github.com/nf-core/modules/pull/6015

PR #5870, opened 2024-06-23
Update main.nf

RaqManzano:
VEP html report is missing when publishing results because by default vep html file is `_summary.html` instead of `.summary.html`. This changes just adjust the module to the vep defaults to avoid the issue.


maxulysse (approved):

maxulysse:
Actually, since one can choose the output name, why not just keep the html extension?

RaqManzano:
> Actually, since one can choose the output name, why not just keep the html extension?

I have done that now, if you don't mind, could you take a quick look? If all test pass, shall we merge?

PR #5904, opened 2024-07-02
correct president stub

paulwolk:
I just fixed the path for the output files in the stub section. Nothing big, but important.

### DIFF
--- a/modules/nf-core/president/main.nf
+++ b/modules/nf-core/president/main.nf
@@ -9,37 +9,38 @@ process PRESIDENT {
 
     input:
     tuple val(meta), path(fasta)
-    path reference
+    tuple val(meta2), path(reference)
     val compress
 
     output:
-    tuple val(meta), path("output/*.fasta*"), emit: fasta
-    tuple val(meta), path("output/*.tsv")   , emit: report
-    tuple val(meta), path("output/*.log")   , emit: log
-    path "versions.yml"                     , emit: versions
+    tuple val(meta), path("${prefix}valid.fasta*")  , emit: valid_fasta
+    tuple val(meta), path("${prefix}invalid.fasta*"), emit: invalid_fasta
+    tuple val(meta), path("*.tsv")                   , emit: report
+    tuple val(meta), path("*.log")                   , emit: log
+    path "versions.yml"                              , emit: versions
 
     when:
     task.ext.when == null || task.ext.when
 
     script:
     def args = task.ext.args ?: ''
-    def prefix  = task.ext.prefix ?: ''
-    if (prefix != ''){
-        prefix = "--prefix " + prefix
-    }
-
+    prefix = task.ext.prefix ?: "${meta.id}_"
+    if ("${fasta}" == "${reference}") error "Input and reference names are the same!"
+    if ("${fasta}" == "${prefix}valid.fasta") error "Input and output file names are the same, use \"task.ext.prefix\" to disambiguate!"
+    if ("${fasta}" == "${prefix}invalid.fasta") error "Input and output file names are the same, use \"task.ext.prefix\" to disambiguate!"
+    if ("${fasta}" == "${prefix}valid.fasta.gz" && compress) error "Input and output file names are the same, use \"task.ext.prefix\" to disambiguate!"
+    if ("${fasta}" == "${prefix}invalid.fasta.gz" && compress) error "Input and output file names are the same, use \"task.ext.prefix\" to disambiguate!"
     """
-    mkdir output
     president \\
         --query $fasta \\
         --reference $reference \\
-        --path output \\
+        --path . \\
         --threads $task.cpus \\
-        $prefix \\
+        --prefix $prefix \\
         $args
 
     if [ "$compress" = true ] ; then
-        gzip output/*.fasta;
+        gzip ${prefix}*.fasta;
     fi
 
     cat <<-END_VERSIONS > versions.yml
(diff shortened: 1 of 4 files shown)


nvnieuwk (commented):
Can you change `output` to the `$prefix` instead? This allows bigger flexibility in the module

SPPearce:
Is there a reason to put the output into a subfolder, rather than just leave it in the top level?

paulwolk:
> Can you change `output` to the `$prefix` instead? This allows bigger flexibility in the module

@nvnieuwk I use the prefix just for the file prefix, the user of the module doesn't see anything of the output directory. Is it necessary? I would change it to ${prefix}/${prefix}report.tsv in that case. 


> Is there a reason to put the output into a subfolder, rather than just leave it in the top level?

@SPPearce I don't want the input fasta (based in root) clash with the output fasta (based in output), which it would do with path("*.fasta*") instead of path("output/*.fasta*"), right?

SPPearce:
Also do you really want the valid and invalid fasta files to go into the same output channel?

paulwolk:
> Also do you really want the valid and invalid fasta files to go into the same output channel?

This is something I would actually happily change. I read some conventions that files of the same filetype should go into one channel, but I would also prefer different channels. So, if that is in line with the conventions, I would do it!

nvnieuwk:
> @nvnieuwk I use the prefix just for the file prefix, the user of the module doesn't see anything of the output directory. Is it necessary? I would change it to ${prefix}/${prefix}report.tsv in that case.

This will emit a folder called `output` in the output directory which will be overwritten for each process so it's possible this causes some weird issues later on :)

SPPearce:
> > Can you change `output` to the `$prefix` instead? This allows bigger flexibility in the module
> 
> @nvnieuwk I use the prefix just for the file prefix, the user of the module doesn't see anything of the output directory. Is it necessary? I would change it to ${prefix}/${prefix}report.tsv in that case.
> 
> > Is there a reason to put the output into a subfolder, rather than just leave it in the top level?
> 
> @SPPearce I don't want the input fasta (based in root) clash with the output fasta (based in output), which it would do with path("*.fasta*") instead of path("output/*.fasta*"), right?

You typically fix that by including a disambiguation check to see if they would clash before the module runs, but that is only going to happen if the input file includes `valid` or `invalid` at the end.

SPPearce:
> > Also do you really want the valid and invalid fasta files to go into the same output channel?
> 
> This is something I would actually happily change. I read some conventions that files of the same filetype should go into one channel, but I would also prefer different channels. So, if that is in line with the conventions, I would do it!

Definitely fine to put the different files into different output channels.
Give me a minute and I'll push an updated version

SPPearce:
Ok, turned into a bit longer than a minute.

- Added meta to the reference fasta
- Swapped to outputting into the working directory rather than a subfolder
- Added a default prefix to the meta.id as standard for most modules
- Compressed the output files in the stub if you have compress set
- Updated the tests to just check the snapshot for the fasta files as they should be stable (it might be good to check that there is only one file in each; that can be seen in the snapshot but isn't being checked that stays that way).
- Changed the channels to have valid and invalid separately.

I ended up putting the underscore on the default prefix which was a bit annoying, it didn't seem to set it otherwise.

SPPearce:
Oh, the meta will need updating with the new channel names

nvnieuwk (approved):
LGTM!

SPPearce:
Note that the tool says: 
`2024-07-02 10:01:49,883 - president - INFO - Using the prefix: test__* to store results.`
and then completely ignores the second underscore.
I have moved the underscore into the command itself because it makes more sense to me to force the underscore.

paulwolk:
Thank you! The only issue I see now is that if you don't want a prefix, there is still the underscore in front of the name. Is there a way to avoid this? Or is the prefix always mandatory?

SPPearce:
> Thank you! The only issue I see now is that if you don't want a prefix, there is still the underscore in front of the name. Is there a way to avoid this? Or is the prefix always mandatory?

You mean that if you just want the files to be called valid.fasta and invalid.fasta? I guess that is true. 
It could go back into the definition of the prefix if you think that is preferable.

SPPearce:
> > Thank you! The only issue I see now is that if you don't want a prefix, there is still the underscore in front of the name. Is there a way to avoid this? Or is the prefix always mandatory?
> 
> You mean that if you just want the files to be called valid.fasta and invalid.fasta? I guess that is true. It could go back into the definition of the prefix if you think that is preferable.

Is this the kind of module that you'd only run once in a workflow? It looks to me like you'd run it repeatedly, at which point a prefix is a good idea to make them distinct if you want to publish the output.

paulwolk:
> > > Thank you! The only issue I see now is that if you don't want a prefix, there is still the underscore in front of the name. Is there a way to avoid this? Or is the prefix always mandatory?
> > 
> > 
> > You mean that if you just want the files to be called valid.fasta and invalid.fasta? I guess that is true. It could go back into the definition of the prefix if you think that is preferable.
> 
> Is this the kind of module that you'd only run once in a workflow? It looks to me like you'd run it repeatedly, at which point a prefix is a good idea to make them distinct if you want to publish the output.

Both is possible. The pipeline I use it in executes it only one time with a bulk fasta, but of course, you can process multiple fasta with it. Having a mandatory prefix is not a problem, but making it optional would make it more flexible.

paulwolk:
You know what? I think it's fine. A prefix doesn't hurt. Thank you for your work, @SPPearce !! 😃

PR #5905, opened 2024-07-02
ADD argNorm module

Vedanth-Ramji:
[argNorm](https://github.com/BigDataBiology/argNorm) is a tool to normalize antibiotic resistance genes (ARGs) by mapping them to the [antibiotic resistance ontology (ARO)](https://obofoundry.org/ontology/aro.html) created by the [CARD database](https://card.mcmaster.ca/).
 

argNorm also enhances antibiotic resistance gene annotations by providing drug categorization of the drugs that antibiotic resistance genes confer resistance to.

argNorm takes the outputs of the [hAMRonization](https://nf-co.re/modules/hamronization_abricate) module as well as ARG annotation tools (DeepARG, ARGs-OAP, ABRicate, ResFinder, and AMRFinderPlus) to normalize them to the antibiotic resistance ontology.

### DIFF
--- a/modules/nf-core/argnorm/main.nf
+++ b/modules/nf-core/argnorm/main.nf
@@ -44,4 +44,25 @@ process ARGNORM {
         argnorm: $VERSION
     END_VERSIONS
     """
+
+    stub:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def VERSION = '0.5.0' // WARN: Version information not provided by tool on CLI. Please update this string when bumping container versions.
+    if (!tool) {
+        error 'Tool not provided.'
+    }
+    if ((tool in ["abricate"]) && !db) {
+        error "$tool requires a database but <db> not provided."
+    }
+
+    """
+    touch ${prefix}
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        argnorm: $VERSION
+    END_VERSIONS
+    """
+
 }
--- a/modules/nf-core/argnorm/main.nf
+++ b/modules/nf-core/argnorm/main.nf
@@ -14,7 +14,7 @@ process ARGNORM {
 
     output:
     tuple val(meta), path("*.tsv"), emit: tsv
-    path "versions.yml", emit: versions
+    path "versions.yml"           , emit: versions
 
     when:
     task.ext.when == null || task.ext.when
@@ -23,7 +23,7 @@ process ARGNORM {
     def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
     def VERSION = '0.5.0' // WARN: Version information not provided by tool on CLI. Please update this string when bumping container versions.
-     if ((tool in ["abricate"]) && !db) {
+    if ((tool in ["abricate"]) && !db) {
         error "$tool requires a database but <db> not provided."
     }
     db = db ? ("--db " + db) : ""
--- a/modules/nf-core/argnorm/main.nf
+++ b/modules/nf-core/argnorm/main.nf
@@ -24,6 +24,9 @@ process ARGNORM {
     def prefix = task.ext.prefix ?: "${meta.id}"
     def VERSION = '0.5.0' // WARN: Version information not provided by tool on CLI. Please update this string when bumping container versions.
     def db_args = db ? "--db ${db}" : ""
+    if (!tool) {
+        error 'Tool not provided.'
+    }
     if ((tool in ["abricate"]) && !db) {
         error "$tool requires a database but <db> not provided."
     }
--- a/modules/nf-core/argnorm/main.nf
+++ b/modules/nf-core/argnorm/main.nf
@@ -1,0 +1,40 @@
+process ARGNORM {
+    tag "$meta.id"
+    label 'process_low'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/argnorm:0.5.0--pyhdfd78af_0':
+        'biocontainers/argnorm:0.5.0--pyhdfd78af_0' }"
+
+    input:
+    tuple val(meta), path(input_tsv)
+    val(db)
+    val(hamronized)
+    val(tool)
+
+    output:
+    tuple val(meta), path("*.tsv"), emit: tsv
+    path "versions.yml"           , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def VERSION = '0.5.0'
+    """
+    argnorm \\
+        $tool \\
+        -i $input_tsv \\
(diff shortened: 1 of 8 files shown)


modules/nf-core/argnorm/meta.yml
```diff
@@ -0,0 +1,65 @@
...
+
+  - db:
+      type: string
+      description: Database used for ARG annotation
+      pattern: "sarg|ncbi|resfinder|deeparg|megares|argannot|resfinderfg"
+
+  - hamronized:
+      type: string
+      description: Indicate ARG annotation output has been processed by hAMRonization
+      pattern: "|--hamronized"
```
SPPearce, line 40: This should be true/false, and the code change the command appropriately

Vedanth-Ramji: I've just updated the code now. Is it ok?

muffato: Hi @Vedanth-Ramji . Can you also update `main.nf` with something like https://github.com/nf-core/modules/blob/master/modules/nf-core/khmer/normalizebymedian/main.nf#L24 ?

muffato:
Hi @Vedanth-Ramji . The code looks fine but I have a question about the input channels you've defined for the module. The [guidelines](https://nf-co.re/docs/guidelines/components/modules#non-file-mandatory-command-arguments) say that:

- "Non-file mandatory arguments (...) SHOULD be provided as value channels". I think that's the `tool` parameter 👍🏼 
- "All _non-mandatory_ command-line tool _non-file_ arguments MUST be provided as a string via the `$task.ext.args` variable". I don't use argnorm myself but the [documentation](https://argnorm.readthedocs.io/en/latest/#basic-usage) seems to indicate that `--db` and `--hamronized` are non-mandatory, and therefore shouldn't be promoted as input channels. Can you confirm ?

Vedanth-Ramji:
Hi @muffato, yes, the `tool` parameter is mandatory. 

argNorm doesn't always require the `--db` and `--hamronized` options, however, in some cases they are mandatory. For example, [here](https://argnorm.readthedocs.io/en/latest/cli/#examples-for-each-tool:~:text=When%20using%20abricate%2C%20it%20is%20necessary%20to%20specify%20the%20database%20used%3A), argNorm will throw an error if the '--db' option isn't provided. Also, if [argNorm is taking inputs from the hAMRonization package](https://argnorm.readthedocs.io/en/latest/#:~:text=hamronized%3A%20use%20this%20if%20the%20input%20is%20hamronized%20by%20hAMRonization), the `--hamronized` option is mandatory.

modules/nf-core/argnorm/main.nf
```diff
@@ -0,0 +1,42 @@
...
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/argnorm:0.5.0--pyhdfd78af_0':
+        'biocontainers/argnorm:0.5.0--pyhdfd78af_0' }"
+
+    input:
+    tuple val(meta), path(input_tsv)
+    val db
+    val hamronized
+    val tool
```
muffato, line 14: I think this is a more natural order
```suggestion
    val tool
    val db
```

modules/nf-core/argnorm/main.nf
```diff
@@ -0,0 +1,42 @@
...
+    path "versions.yml", emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def VERSION = '0.5.0'
+    hamronized = hamronized ? "--hamronized" : ""
```
muffato, line 27:
```suggestion
    if ((tool in ["abricate"]) && !db) {
        error "$tool requires a database but <db> not provided."
    }   
    def db_args = db ? "--db " + db : ""
```

modules/nf-core/argnorm/main.nf
```diff
@@ -0,0 +1,42 @@
...
+
+    """
+    argnorm \\
+        $tool \\
+        -i $input_tsv \\
+        -o $prefix \\
+        --db $db \\
+        $hamronized
```
muffato, line 35: I just noticed that `$args` was not included at the end
```suggestion
        $db_args \\
        $args
```

modules/nf-core/argnorm/main.nf
```diff
@@ -0,0 +1,42 @@
...
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        argnorm: $VERSION
+    END_VERSIONS
+    """
+}
```
muffato, line 42: Need to add this trailing new line

modules/nf-core/argnorm/main.nf
```diff
@@ -0,0 +1,42 @@
...
+    tuple val(meta), path("*.tsv"), emit: tsv
+    path "versions.yml", emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def VERSION = '0.5.0'
```
muffato, line 26:
```suggestion
    def VERSION = '0.5.0' // WARN: Version information not provided by tool on CLI. Please update this string when bumping container versions.
```

modules/nf-core/argnorm/main.nf
```diff
@@ -0,0 +1,42 @@
+process ARGNORM {
+    tag "$meta.id"
+    label 'process_low'
+
```
muffato, line 4:
```suggestion

    // WARN: Version information not provided by tool on CLI. Please update version string below when bumping container versions.
```

muffato (commented):
Oki, thanks. In that case, I think hamronized should _not_ be a channel input here and should be passed via `task.ext.args`. It's no different from the `-p` parameter of `tabix` having to be carefully set according to the input, and it currently comes via `task.ext.args` (see <https://github.com/nf-core/modules/blob/master/modules/nf-core/tabix/tabix/tests/tabix_bed.config> or <https://github.com/nf-core/modules/blob/master/modules/nf-core/tabix/tabix/tests/tabix_vcf_tbi.config>)

The database can stay as a channel input with extra code (see my suggestion) to check its validity. That's exactly the guideline [1.2](https://nf-co.re/docs/guidelines/components/modules#non-file-mandatory-command-arguments). I've suggested some code to make it truly optional.

I've also added warnings that the command doesn't provide its version, as per [1.8](https://nf-co.re/docs/guidelines/components/modules#emission-of-versions).

modules/nf-core/argnorm/main.nf
muffato: I forgot one thing: `meta.yaml` will have to be updated accordingly

muffato (approved):
I fixed the linting, the stub mode, and added tests for `tool` and `db`

Vedanth-Ramji:
Thank you very much @muffato. I rebased the branch to be up to date with upstream. Is there anything else to be added before merging?

muffato:
Thanks. I've kicked off the auto-merge, which will do a last CI run before merging.

muffato:
All done 🚀

PR #5823, opened 2024-06-18
Using Seqera containers for Sentieon to avoid LD_LIBRARY errors

asp8200:
None of the sentieon-tests - except `sentieon/bwaindex`- will pass until the license-server is up and running again.

The test `sentieon/bwaindex` is passing since running bwaindex from the sentieon-package doesn't require a license.

With the test `sentieon/bwaindex` passing for both docker and singularity, there is a good chance the the other sentieon-tests will also pass for docker and singularity once the license-server is running. The changes in this PR shouldn't affect the sentieon-modules working with conda.

## PR checklist

Closes #XXX 

- [ ] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [ ] Remove all TODO statements.
- [ ] Emit the `versions.yml` file.
- [ ] Follow the naming conventions.
- [ ] Follow the parameters requirements.
- [ ] Follow the input/output options guidelines.
- [ ] Add a resource `label`
- [ ] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - For modules:
    - [ ] `nf-core modules test <MODULE> --profile docker`
    - [ ] `nf-core modules test <MODULE> --profile singularity`
    - [ ] `nf-core modules test <MODULE> --profile conda`
  - For subworkflows:
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile docker`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile singularity`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile conda`

PR #5945, opened 2024-07-09
Refactor Sentieon

edmundmiller:
Here's what I had hacked together so far!


asp8200:
@edmundmiller : I see you're removing `fix_ld_library_path` in the sentieon/bwamem-module, and I believe that is what is causing the CI-test of sentieon/bwamem to fail. If that is indeed the case, then I see two solutions

1) Keep `fix_ld_library_path`, or
2) use seqera-containers like I tested out in [this](https://github.com/nf-core/modules/pull/5823) draft PR.

We definitely want to switch to the seqera-containers at one point.

asp8200:
Hmmm ... It doesn't seem quite right. The tests fail with ` .command.sh: line 2: SENTIEON_LICENSE_BASE64: unbound variable`

edmundmiller:
Fixed it! Not sure if the license file works or not, I still need to test that locally.

asp8200:
Looks good to me. Happy to see the CI-tests pass. I'll just wait for the results of your "local test" of the setup for license-files, and then I'll approve.

modules/nf-core/sentieon/bwamem/main.nf
```diff
@@ -37,27 +33,13 @@ process SENTIEON_BWAMEM {
 
     def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
-    def sentieon_auth_mech_base64 = task.ext.sentieon_auth_mech_base64 ?: ''
-    def sentieon_auth_data_base64 = task.ext.sentieon_auth_data_base64 ?: ''
+    // nextflow secrets set SENTIEON_LICENSE_BASE64 \$(cat <sentieon_license_file.lic> | base64 -w 0)
+    def sentieonLicense = secrets.SENTIEON_LICENSE_BASE64 ?
+        "export SENTIEON_LICENSE=\$(mktemp);echo -e \"${secrets.SENTIEON_LICENSE_BASE64}\" | base64 -d > \$SENTIEON_LICENSE; " :
+        ""
```
adamrtalbot, line 39: This doesn't work with:
 - SENTIEON_LICENSE is an IP or DNS of a Sentieon license server (90% of users), e.g. `SENTIEON_LICENSE=123.456.789:8990`
 - Sentieon uses the enhanced security license method we were previously using which includes an authentication layer.

adamrtalbot (changes requested):

adamrtalbot (approved):

maxulysse:
CI is now passing <3

modules/nf-core/sentieon/bwamem/tests/nextflow.config
```diff
@@ -1,7 +1,11 @@
 env {
-    SENTIEON_LICSRVR_IP = "$SENTIEON_LICSRVR_IP"
+    // NOTE This is how nf-core/sarek users will use Sentieon in real world use
+    SENTIEON_LICENSE = "$SENTIEON_LICSRVR_IP"
+    // NOTE This should only happen in GitHub actions or nf-core MegaTests
     SENTIEON_AUTH_MECH = "$SENTIEON_AUTH_MECH"
```
adamrtalbot, line 5: Might throw error if SENTIEON_AUTH_MECH isn't populated?

maxulysse:
We either need to fix pytest tests, disable them or port them to nf-test cc @FriederikeHanssen

modules/nf-core/sentieon/bwamem/main.nf
edmundmiller: https://github.com/nf-core/modules/blob/b891a909e3031e41b410dfa447f5f527bf9781d4/modules/nf-core/sentieon/README.md
✨ Feel free to clear up anything in there! 

WIP Sarek PR for the actual end users: https://github.com/nf-core/sarek/pull/1579

edmundmiller:
I've just disabled them. As we update the modules, we'll port them to `nf-test`. Couldn't be bothered to duplicate the logic to the CI.

modules/nf-core/sentieon/bwamem/tests/nextflow.config
adamrtalbot: It's on the user to configure their `env`, but that's where the docs comes in (and maybe a warning!). Not certain about profiles, would want to see a working version first and test the UX.

modules/nf-core/sentieon/bwamem/main.nf
adamrtalbot: Could we port to the website or something? It's a bit buried here?

adamrtalbot: but this is fine for now.

DonFreed (approved):
Looks OK to me. Glad that the Sentieon license variables are greatly simplified.

modules/nf-core/sentieon/bwamem/main.nf
edmundmiller: We could link to it in the CI?

Honestly, half the battle is CI, and won't matter for the end-users.

The Sarek(and other pipeline) docs will be what anyone actually using the pipeline or module will care about.

PR #5993, opened 2024-07-18
Swap samtools dict to nftest

SPPearce:


modules/nf-core/samtools/dict/tests/main.nf.test
```diff
@@ -0,0 +1,59 @@
...
+    tag "samtools/dict"
+
+    test("sarscov2 - fasta") {
+
+        when {
+            process {
+                """
+                input[0] = [
+                    [ id:'test', single_end:false ], // meta map
+                    file(params.test_data['sarscov2']['genome']['genome_fasta'], checkIfExists: true)
```
maxulysse, line 19: sorry to bother you with this, but could you use the other path system?

modules/nf-core/samtools/dict/tests/main.nf.test
```diff
@@ -0,0 +1,59 @@
...
+                    ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out.versions).match() },
+                { assert path(process.out.dict[0][1]).readLines().first().contains("@HD\tVN:1.0\tSO:unsorted") }
```
nvnieuwk, line 29: Can you instead of this try to snapshot the stable lines of the file?

nvnieuwk (commented):
One small comment here

nvnieuwk (approved):

Issue #5832, opened 2024-06-19
Migrate to using Seqera Containers


ewels:

subwaystation:
It is still unclear to me, that images generated on demand should not harm reproducibility.

subwaystation:
Was this ever evaluated?

subwaystation:
This will also increase the used CPU hours, right?
Also, it feels a bit like Seqera will have a monopoly of containers. I am aware that Seqera did an awesome job developing Nextflow open source. But moving all container logic to Seqera (and I don't know the details here, maybe I am uninformed) gives a weird taste.

CharlotteAnne:
What’s the purpose of enforcing this? Biocontainers are automatically generated for every bioconda package and get automatically generated upon bioconda software version bump. I’m not seeing the reason to then manually create a seqera container?

ewels:
Hi both - thanks for your comments. You're right that this issue precedes some community discussion that we still need to have. That started with the recent two bytesize talks and resulting conversations on Slack, but we should still open it up to wider input.

To address your concerns:

---

> It is still unclear to me, that images generated on demand should not harm reproducibility.

_Wave_ generates images on demand, but _Seqera Containers_ is a registry that sits behind Wave. The intention here is that the images are generated on demand by the developer when a package is updated - but then they are cached in the Seqera Containers registry. The image URIs will then be hardcoded into pipelines and the exact same container images will always be fetched by all users - just the same as they are today. We're also going to introduce conda-lock files (see https://github.com/nf-core/modules/issues/5835) so reproducibility should be even better than it is today.

---

> This will also increase the used CPU hours, right?

No - Wave / Seqera Containers handles the build server side. As mentioned above, the generated images are stored in a registry and simply downloaded. So just as today, native images will be downloaded. No increase in CPU hours.

---

> Also, it feels a bit like Seqera will have a monopoly of containers. But moving all container logic to Seqera gives a weird taste.

This one is more subjective. We will not make it a requirement to use Seqera Containers, just as we don't make it a requirement to use BioContainers today, so for me it feels about the same. We will keep the vast majority of build logic (eg. conda env files, conda lock files) on the nf-core side and will be free to reverse the decision at any point should we wish.

---

> Biocontainers are automatically generated for every bioconda package and get automatically generated upon bioconda software version bump. I’m not seeing the reason to then manually create a seqera container?

[...]

PR #6063, opened 2024-07-30
Replace bedtools sort with unix sort in BEDTOOLS_GENOMECOV

adamrtalbot:
`bedtools sort` uses a large amount of CPUs and memory, but when using it here it doesn't require the  additional genome based features of `bedtools`. Replacing it should speed up the process and make it many times more efficient.

First discovered by @pabloaledo 

## PR checklist

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/bedtools/genomecov/main.nf
+++ b/modules/nf-core/bedtools/genomecov/main.nf
@@ -27,7 +27,7 @@ process BEDTOOLS_GENOMECOV {
     if (!args_list.contains('-bg') && (scale > 0 && scale != 1)) {
         args += " -bg"
     }
-    def sort_cmd = sort ? '| bedtools sort' : ''
+    def sort_cmd = sort ? '| sort -k1,1 -k2,2n' : ''
 
     def prefix = task.ext.prefix ?: "${meta.id}"
     if (intervals.name =~ /\.bam/) {
--- a/modules/nf-core/bedtools/genomecov/main.nf
+++ b/modules/nf-core/bedtools/genomecov/main.nf
@@ -28,7 +28,7 @@ process BEDTOOLS_GENOMECOV {
     if (!args_list.contains('-bg') && (scale > 0 && scale != 1)) {
         args += " -bg"
     }
-    def sort_cmd = sort ? '| sort $args2 -k1,1 -k2,2n' : ''
+    def sort_cmd = sort ? "| sort $args2 -k1,1 -k2,2n" : ''
 
     def prefix = task.ext.prefix ?: "${meta.id}"
     if (intervals.name =~ /\.bam/) {
--- a/modules/nf-core/bedtools/genomecov/main.nf
+++ b/modules/nf-core/bedtools/genomecov/main.nf
@@ -21,17 +21,16 @@ process BEDTOOLS_GENOMECOV {
     task.ext.when == null || task.ext.when
 
     script:
-    def args  = task.ext.args  ?: ''
-    def args2 = task.ext.args2 ?: ''
+    def args      = task.ext.args  ?: ''
     def args_list = args.tokenize()
-    def buffer = task.memory.toGiga().intdiv(2)
     args += (scale > 0 && scale != 1) ? " -scale $scale" : ""
     if (!args_list.contains('-bg') && (scale > 0 && scale != 1)) {
         args += " -bg"
     }
     // Sorts output file by chromosome and position using additional options for performance and consistency
     // See https://www.biostars.org/p/66927/ for further details
-    def sort_cmd = sort ? "| LC_ALL=C sort --parallel=$task.cpus --buffer-size=${buffer}G -k1,1 -k2,2n" : ''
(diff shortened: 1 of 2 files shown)


JoseEspinosa (commented):
We actually had issues with bedtools sort in [nf-core/atacseq](https://github.com/nf-core/atacseq), see [this](https://github.com/nf-core/atacseq/pull/369) and [this](https://github.com/nf-core/atacseq/pull/374/) PR. Actually, in our case, we switch to use gnu sort and be able to use the `--parallel` and `--buffer-size` arguments, see [here](https://github.com/JoseEspinosa/nf-core-atacseq/blob/dbab390e422bd455c2d9155bcb745276c51ed57a/modules/local/bedtools_genomecov.nf#L42-L44). The reason for these changes was that the previous implementation of the module was very memory gready when dealing with big files resulting for the merge of the files at the library level, making the pipeline to fail for many users.

JoseEspinosa:
> We actually had issues with bedtools sort in [nf-core/atacseq](https://github.com/nf-core/atacseq), see [this](https://github.com/nf-core/atacseq/pull/369) and [this](https://github.com/nf-core/atacseq/pull/374/) PR. Actually, in our case, we switch to use gnu sort and be able to use the `--parallel` and `--buffer-size` arguments, see [here](https://github.com/JoseEspinosa/nf-core-atacseq/blob/dbab390e422bd455c2d9155bcb745276c51ed57a/modules/local/bedtools_genomecov.nf#L42-L44). The reason for these changes was that the previous implementation of the module was very memory gready when dealing with big files resulting for the merge of the files at the library level, making the pipeline to fail for many users.

So I am totally for this change. Also, I would suggest to use gnu sort and add an `args2` to be able to use the above-mentioned arguments to tune `sort`.

JoseEspinosa (commented):
I am not wrong sort in `biocontainers/bedtools:2.31.1--hf5e1c6e_0` is not gnu sort and thus, the arguments `--parallel` and the `--buffer-size` won't work. Also, not sure if we should also include `LC_ALL=C sort` to be sure that the sort order remains the same across different systems. Finally for reference, our implementation in nf-core/atacseq was inspired by [this](https://github.com/nf-core/hicar/blob/d2d17a924e42d6f88640b79d48d8b332f33a953f/modules/local/atacreads/bedsort.nf#L23-L29) implementation on hicar of single sort process.

JoseEspinosa:
> > Rather than complicate things with `args2`, perhaps we should just add those features if they don't cause any harm?
> 
> the default biocontainer comes with a version of `sort` that does not support `--parallel`, so that won't work. The Seqera container (`community.wave.seqera.io/library/bedtools:2.31.1--8fd0e3802b0dc02e`) does, so we could switch to that but I'm not certain what the current thinking is regarding switching containers. I notice that @JoseEspinosa switched to using that container in the PRs above.

For me is OK to add them as default, this is what I actually did in the atacseq module.
For the container, I thought that for mulled containers it was ok to use `wave`, since if you want to use sort from conda-forge you will need to construct one mulled container.

JoseEspinosa (approved):
Wonderful! 🚀

JoseEspinosa (commented):
One last thought, should we control for `task.memory` being not `null` (not set) ?

SPPearce (approved):
Can you just add a comment as to why it has been changed please, so nobody decides to change it back in future ;)

ewels:
> > One last thought, should we control for `task.memory` being not `null` (not set) ?
> 
> Good point. It should always be set but I've just moved the logic to handle the case where it has been explicitly set to `null`.

Some HPC clusters refuse job submissions where memory has been set, so there is a valid use case for this being set to `null` deliberately.

SPPearce:
> We still have an open discussion around Seqera containers and private registries. Clearly, people are using them (great!) but support for private or offline registries isn't solved and it makes me wary of going too far down this route.

Are we actually having that discussion anywhere?

edmundmiller:
Bedtools recommends this method itself https://bedtools.readthedocs.io/en/latest/content/tools/sort.html

So method wise no controversy.

Seqera containers, @maxulysse has been copying them over to quay.io for the simplicity of having all the containers in one place (Why that makes it easier I still don't understand). So I think that's your quick fix there.

> Are we actually having that discussion anywhere?

#5832

FriederikeHanssen:
> Seqera containers, @maxulysse has been copying them over to quay.io for the simplicity of having all the containers in one place (Why that makes it easier I still don't understand). So I think that's your quick fix there.

if people set the `registry` because they have their own it's a lot easier because you don't need to overwrite random containers everywhere

FriederikeHanssen:
> Are we actually having that discussion anywhere?

@SPPearce several POCs here https://github.com/seqeralabs/nf-aggregate/pull/43 https://github.com/seqeralabs/nf-aggregate/pull/44 https://github.com/seqeralabs/nf-aggregate/pull/45 https://github.com/seqeralabs/nf-aggregate/pull/46 and here: https://github.com/nf-core/modules/issues/5832, https://github.com/nf-core/tools/issues/2408

PR #6072, opened 2024-07-31
Allow to use task.ext.prefix for naming chromap index file and bump version 0.2.6

JoseEspinosa:
Until now `CHROMAP_INDEX` output index could only be named after the input `fasta`, changed to be coherent with other similar modules as [BWA_INDEX](https://github.com/nf-core/modules/blob/409e81ae6cdc2f70e1cf179886059d15119702b6/modules/nf-core/bwa/index/main.nf#L21)

## PR checklist

  - For modules:


SPPearce (approved):

mahesh-panchal (changes requested):
Something is wrong here. The overall changes don't bump the version or change how `args` or `prefix` is used, even though there are commits that do that.

mahesh-panchal:
Ah, can this be closed since #6113 did the bumping? ( and I presume the changes disappeared because of the merging of master )

PR #6078, opened 2024-08-01
Sentieon/Haplotyper: Recal table is sample specific

FriederikeHanssen:
PR adds recalibration as part of the sample-specific input instead of a general input, since each sample has its own recalibration table generated upstream by sentieon/qualcal. 

Without the changes the recalibration table can't be matched to the sample.


adamrtalbot (approved):
Update the PR description please.

PR #6107, opened 2024-08-04
New module: build Kraken2 standard database

### LINKED ISSUE #6070: new module: kraken2/buildstandard

egreenberg7:
## PR checklist

Closes #6070

Adds a new module to build the Kraken2 standard database (as opposed to the existing Kraken2 Build module which only allows for creation of a Kraken database manually curated taxonomies/genomic libraries). This will allow for a simple Kraken2 database installation that can help to make pipelines using Kraken2 more self-contained.

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:


GallVp:
Looks like a good start. Another option is to add `db_type` as an input to the existing `kraken2/build` module: https://github.com/nf-core/modules/blob/34ac993e081b32d2170ab790d0386b74122f9d36/modules/nf-core/kraken2/build/main.nf and make the other inputs optional.

It will be easier to maintain a single  module instead of two? Is there a way to also test the module without `-stub`?

Shaun-Regenbaum (commented):
Everything seems to be good, it seems to pass in docker at least. I don't see the harm in approving?

Shaun-Regenbaum:
After talking with @egreenberg7, this may be redundant as the pre-built libraries are only 80Gb, while building takes 300Gb. I still think its good to have, especially for the future if there are hiccups with the provided pre-built.

jfy133:
> It could make sense to not have the separate modules for the sake of maintenance/ease of use, though the change in the input channels would likely affect a lot of existing met

In this case going into a single module may end up with extra maintenance burden.

There are two routes to getting a kraken database

download -> build
add -> build

These are two quite different

I ended up ignoring the `--standard` option in my original `build` module from the PoV, that in the first case, technically all it's doing downloading an archive - as @Shaun-Regenbaum says. You could just do this outside of a pipeline (As you would cache it anyway) and give it directly to Kraken, whereas my build module explicitly requires `kraken add` beforehand, as building 'custom' databases require more involved steps/decisions.

If we were to have a single module,Having to then choose is module downloading internally vs using an `add` prepared database would require more conditionals and possibly bash modifications around the command which is also more maintenance burder.  This is also reflected by having to add more 'optional' input channels as @egreenberg7 mentions, which makes the module more complicated to use (IMO!)

modules/nf-core/kraken2/buildstandard/main.nf
```diff
@@ -0,0 +1,44 @@
...
+    script:
+    def args = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "kraken2_standard_db"
+    """
+    kraken2-build \\
+        --standard \\
+        $args \\
+        --threads ${task.cpus} \\
+        --db ${prefix}
+    kraken2-build --clean --db ${prefix}
```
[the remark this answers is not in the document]
jfy133: Technically this should also have an `args2`, although I don't think `--clean` accepts any other parameters?

modules/nf-core/kraken2/buildstandard/meta.yml
```diff
@@ -0,0 +1,30 @@
+name: "kraken2_buildstandard"
+description: Builds Kraken2 standard database
+keywords:
+  - metagenomics
+  - db
+  - classification
+  - build
+  - kraken2
```
jfy133, line 8:
```suggestion
  - kraken2
  - standard
  - download
```

modules/nf-core/kraken2/buildstandard/meta.yml
```diff
@@ -0,0 +1,30 @@
...
+      homepage: https://ccb.jhu.edu/software/kraken2/
+      documentation: https://github.com/DerrickWood/kraken2/wiki/Manual
+      tool_dev_url: "https://github.com/DerrickWood/kraken2"
+      doi: 10.1186/s13059-019-1891-0
+      licence: ["MIT"]
+      args_id: "$args"
+output:
+  - db:
+      type: directory
+      description: contains the database that can be used to perform taxonomic classification
```
jfy133, line 21:
```suggestion
      description: Directory containing the Kraken2 'standard' database that can be used to perform taxonomic classification
```

jfy133 (changes requested):
Very nice, mostly minor things! I would make the cleanup optional though, could be someone wants to inspects all the downloaded files...

modules/nf-core/kraken2/buildstandard/main.nf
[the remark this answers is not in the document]
egreenberg7: I don't think --clean has any parameters either. The reason I got rid of the clean option was my assumption is that if people are using the build-standard (rather than manually building or downloading a prebuilt index from [the index website](https://benlangmead.github.io/aws-indexes/k2) then they probably are not too interested in the particulars of how Kraken works and therefore are less likely to inspect the inside. As a result, I just made use of this as barebones as possible with no input.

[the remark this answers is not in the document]
jfy133: Fair enough, I still generally prefer not to make assumptions of what users want to do here (humans are weird)..., but I won't die on that hill :)

egreenberg7:
Cleanup is now optional

modules/nf-core/kraken2/buildstandard/main.nf
```diff
@@ -0,0 +1,46 @@
...
+    prefix = task.ext.prefix ?: "kraken2_standard_db"
+    runclean = cleaning ? "kraken2-build --clean --db ${db}" : ""
+    """
+    kraken2-build \\
+        --standard \\
+        $args \\
+        --threads ${task.cpus} \\
+        --db ${prefix}
+    $runclean
+    cat <<-END_VERSIONS > versions.yml
```
jfy133, line 29:
```suggestion
        --db ${prefix}
        
    $runclean
    
    cat <<-END_VERSIONS > versions.yml
```

modules/nf-core/kraken2/buildstandard/meta.yml
```diff
@@ -0,0 +1,32 @@
...
+  - download
+tools:
+  - kraken2:
+      description: "Kraken2 is a system for assigning taxonomic labels to short DNA sequences, usually obtained through metagenomic studies."
+      homepage: https://ccb.jhu.edu/software/kraken2/
+      documentation: https://github.com/DerrickWood/kraken2/wiki/Manual
+      tool_dev_url: "https://github.com/DerrickWood/kraken2"
+      doi: 10.1186/s13059-019-1891-0
+      licence: ["MIT"]
+      args_id: "$args"
```
jfy133, line 19: Missing `cleaning` input: meta yaml

jfy133 (commented):
Almost there!

jfy133 (approved):
Add the `input:` metal.yml for cleaning and then You're good to go :) Thanks @egreenberg7 !

jfy133:
@nf-core-bot fix linting

modules/nf-core/kraken2/buildstandard/main.nf
```diff
@@ -0,0 +1,48 @@
...
+    def args = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "kraken2_standard_db"
+    runclean = cleaning ? "kraken2-build --clean --db ${db}" : ""
+    """
+    kraken2-build \\
+        --standard \\
+        $args \\
+        --threads ${task.cpus} \\
+        --db ${prefix}
+
```
jfy133, line 28:
```suggestion

```

modules/nf-core/kraken2/buildstandard/main.nf
```diff
@@ -0,0 +1,48 @@
...
+    runclean = cleaning ? "kraken2-build --clean --db ${db}" : ""
+    """
+    kraken2-build \\
+        --standard \\
+        $args \\
+        --threads ${task.cpus} \\
+        --db ${prefix}
+        
+    $runclean
+
```
jfy133, line 30:
```suggestion

```

jfy133:
> Add the `input:` metal.yml for cleaning and then You're good to go :) Thanks @egreenberg7 !

Just this, to go - I fixed the linting (which was my fault)

PR #6116, opened 2024-08-06
Changed prefix/prefix to dbs/prefix

nbtm-sh:
Added variable for DB prefix.

It is common to create a Database that is named different form the database files.

If the DB variable is not set, it will default to the prefix value.


SPPearce:
I don't really understand why you need this, can you elaborate please.
We don't generate depend on extra fields in the meta map, so I think this would need a convincing argument

luisas:
@nbtm-sh are you planning to work on this or do you think we could close the issue?

SPPearce:
Closing as stale, please reopen if you are still interested.

PR #6138, opened 2024-08-08
ADD NONPAREIL/NONPAREILCURVESR

jfy133:
Adds a new module from a utility script in the pacakge that produces multiqc compatible files

Also tweaks the descriptions of similar modules in the same package to distinguish between the two

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


modules/nf-core/bioawk/main.nf
```diff
@@ -20,7 +20,6 @@ process BIOAWK {
     script:
     def args  = task.ext.args ?: '' // args is used for the main arguments of the tool
     prefix = task.ext.prefix ?: "${meta.id}"
```
jfy133, line 22:
```suggestion
    prefix = task.ext.prefix ?: "${meta.id}"
    if ("${input}" == "${prefix}") error "Input and output names are the same, use \"task.ext.prefix\" to disambiguate!"
```

modules/nf-core/nonpareil/nonpareilcurvesr/main.nf
```diff
@@ -0,0 +1,57 @@
...
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+
+    """
+    NonpareilCurves.R \\
+        $args \\
+        --json ${prefix}.json \\
+        --tsv ${prefix}.tsv \\
+        --csv ${prefix}.csv \\
+        --pdf ${prefix}.pdf \\
```
LilyAnderssonLee, line 34: Nice! Is this the proper way to define the optional outputs instead of defining a function for output options?

jfy133: Oh good point... I set them optional initially as they technically are, but I realised they are all small so could just generate alll...

Do you have a suggestion how we could ask the user/pipeline dev which ones they want?

LilyAnderssonLee: Emm, I am not sure if I could come up a wiser way. How about something like this, only keep the `tsv` output? I guess we could also keep all of them since they are quite small anyway. 
```
    output:
    tuple val(meta), path("**.{json,tsv,csv,pdf}"), emit: json, optional: true

    """
    NonpareilCurves.R \\
        $args \\
        --tsv ${prefix}.tsv \\
        $npos
    """

```

jfy133: The problem there is it's more tricky to 'pick one' for downstream processing (e.g. multiqc), but still publish the others...

I don't want to force devs  to pass output flags via `args`... as this could become confusing (if you see what I mean)

JoseEspinosa: Something like [this](https://github.com/nf-core/modules/blob/16af6ca4aeb31dcbb2d652ada5f395d62793426e/modules/nf-core/blast/blastdbcmd/main.nf#L32), could not work?

jfy133: Mmm, issue there is that is a fixed thing, whereas in this case I have to specify the file name after... `--tsv <filename> --csv <filename>` etc...

jfy133: and I don't want pipeline devs having to specify `--tsv ${prefix}.tsv` in their modules config etc.

jfy133:
OK me triggering 200 tests on this PR (I still have no idea why this is happening) seems crazy - I'm going to make a new PR

SPPearce:
> OK me triggering 200 tests on this PR (I still have no idea why this is happening) seems crazy - I'm going to make a new PR

That is what happens when the branch behind, clicking update branch should fix it

jfy133:
Was still happening when I did that...

PR #6155, opened 2024-08-09
Add cellbender modules

nictru:
Adds cellbender for ambient RNA removal in single cell transcriptomics. Is supposed to be used in both #scrnaseq and #scdownstream.

## PR checklist

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/cellbender/merge/templates/merge.py
+++ b/modules/nf-core/cellbender/merge/templates/merge.py
@@ -31,7 +31,11 @@ adata_cellbender = load_anndata_from_input_and_output("${unfiltered}", "${cellbe
                                             analyzed_barcodes_only=False)
 
 adata_cellbender = adata_cellbender[adata.obs_names]
-adata.X = adata_cellbender.layers["cellbender"]
+
+if "${output_layer}" == "X":
+    adata.X = adata_cellbender.layers["cellbender"]
+else:
+    adata.layers["${output_layer}"] = adata_cellbender.layers["cellbender"]
 
 adata.write_h5ad("${prefix}.h5ad")
 
--- a/modules/nf-core/cellbender/merge/templates/merge.py
+++ b/modules/nf-core/cellbender/merge/templates/merge.py
@@ -1,0 +1,49 @@
+#!/usr/bin/env python3
+
+import platform
+import cellbender
+import anndata as ad
+from cellbender.remove_background.downstream import load_anndata_from_input_and_output
+
+def format_yaml_like(data: dict, indent: int = 0) -> str:
+    """Formats a dictionary to a YAML-like string.
+
+    Args:
+        data (dict): The dictionary to format.
+        indent (int): The current indentation level.
+
+    Returns:
+        str: A string formatted as YAML.
+    """
+    yaml_str = ""
+    for key, value in data.items():
+        spaces = "  " * indent
+        if isinstance(value, dict):
+            yaml_str += f"{spaces}{key}:\\n{format_yaml_like(value, indent + 1)}"
+        else:
+            yaml_str += f"{spaces}{key}: {value}\\n"
+    return yaml_str
+
+
+adata = ad.read_h5ad("${filtered}")
+
+adata_cellbender = load_anndata_from_input_and_output("${unfiltered}", "${cellbender_h5}",
+                                            analyzed_barcodes_only=False)
+
+adata_cellbender = adata_cellbender[adata.obs_names]
+adata.X = adata_cellbender.layers["cellbender"]
+
+adata.write_h5ad("${prefix}.h5ad")
+
+# Versions
+
+versions = {
+    "${task.process}": {
+        "python": platform.python_version(),
+        "cellbender": cellbender.__version__,
+        "anndata": ad.__version__
+    }
+}
+
+with open("versions.yml", "w") as f:
+    f.write(format_yaml_like(versions))
--- a/modules/nf-core/cellbender/removebackground/main.nf
+++ b/modules/nf-core/cellbender/removebackground/main.nf
@@ -45,6 +45,7 @@ process CELLBENDER_REMOVEBACKGROUND {
     """
 
     stub:
+    prefix = task.ext.prefix ?: "${meta.id}"
     """
     touch "${prefix}.h5"
     touch "${prefix}_filtered.h5"
--- a/modules/nf-core/cellbender/removebackground/main.nf
+++ b/modules/nf-core/cellbender/removebackground/main.nf
@@ -47,8 +47,14 @@ process CELLBENDER_REMOVEBACKGROUND {
     stub:
     """
     touch "${prefix}.h5"
+    touch "${prefix}_filtered.h5"
+    touch "${prefix}_posterior.h5"
     touch "${prefix}_cell_barcodes.csv"
     touch "${prefix}_metrics.csv"
+    touch "${prefix}_report.html"
+    touch "${prefix}.pdf"
+    touch "${prefix}.log"
+    touch "ckpt.tar.gz"
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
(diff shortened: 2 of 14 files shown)


nictru:
Still need to transfer the test datasets to the modules branch, but the implementation is ready for review

modules/nf-core/cellbender/merge/templates/merge.py
```diff
@@ -0,0 +1,49 @@
...
+        str: A string formatted as YAML.
+    """
+    yaml_str = ""
+    for key, value in data.items():
+        spaces = "  " * indent
+        if isinstance(value, dict):
+            yaml_str += f"{spaces}{key}:\\n{format_yaml_like(value, indent + 1)}"
+        else:
+            yaml_str += f"{spaces}{key}: {value}\\n"
+    return yaml_str
```
grst, line 25: Really no pyyaml in the cellbender container? Either way not that important, feel free to leave as is.

modules/nf-core/cellbender/merge/templates/merge.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+adata = ad.read_h5ad("${filtered}")
+
+adata_cellbender = load_anndata_from_input_and_output("${unfiltered}", "${cellbender_h5}",
+                                            analyzed_barcodes_only=False)
+
+adata_cellbender = adata_cellbender[adata.obs_names]
+adata.X = adata_cellbender.layers["cellbender"]
```
grst, line 34: Maybe better to store cellbender in a layer rather than X? Otherwise the original expression values get lost

modules/nf-core/cellbender/removebackground/main.nf
```diff
@@ -0,0 +1,49 @@
...
+        'oras://community.wave.seqera.io/library/cellbender:0.3.0--c4addb97ab2d83fe':
+        'community.wave.seqera.io/library/cellbender:0.3.0--41318a055fc3aacb' }"
+
+    input:
+    tuple val(meta), path(h5ad)
+
+    output:
+    tuple val(meta), path("${prefix}.h5")               , emit: h5
+    tuple val(meta), path("${prefix}_cell_barcodes.csv"), emit: barcodes
+    tuple val(meta), path("${prefix}_metrics.csv")      , emit: metrics
```
grst, line 16: What about the HTML and PDF reports? I think there's some value of publishing those to the output directory. 

Is the information which barcodes are "filtered" according to cellbender also contained elsewhere? Otherwise it might be worth additionally emitting the `${prefix}_filtered.h5` file.

modules/nf-core/cellbender/removebackground/main.nf
```diff
@@ -0,0 +1,49 @@
...
+    task.ext.when == null || task.ext.when
+
+    script:
+    prefix = task.ext.prefix ?: "${meta.id}"
+    args = task.ext.args ?: ""
+    """
+    TMPDIR=. cellbender remove-background \
+        ${args} \
+        --input ${h5ad} \
+        --low-count-threshold 2 \
```
grst, line 29: This seems a bit arbitrary, why did you include this? Maybe this should be passed via `args` at the subworkflow level instead?

modules/nf-core/cellbender/removebackground/main.nf
```diff
@@ -0,0 +1,49 @@
...
+    path "versions.yml"                                 , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    prefix = task.ext.prefix ?: "${meta.id}"
+    args = task.ext.args ?: ""
+    """
+    TMPDIR=. cellbender remove-background \
```
grst, line 26: Cellbender profits a lot from GPU support (`--cuda`) flag. Is there an nf-core way of handling accelerators (something like `label 'process_gpu`) or would I just add this flag via `args` in my custom config file (plus selecting the appropriate queue and container flags)

modules/nf-core/cellbender/removebackground/main.nf
grst: might be used for a multiqc module in the future: https://github.com/broadinstitute/CellBender/issues/257

modules/nf-core/cellbender/removebackground/main.nf
nictru: There is not an official convention yet, but I have started a discussion in the maintainers team and set up [this issue](https://github.com/nf-core/scdownstream/issues/69#issuecomment-2278072245) in scdownstream.

I added a resource label now, that allows utilizing GPU as described.

modules/nf-core/cellbender/removebackground/main.nf
nictru: Added output emissions for all output files via the latest commits

modules/nf-core/cellbender/merge/templates/merge.py
nictru: What do you think about this solution:

- Input layer is specified as `task.ext.input_layer`, defaulting to `X`
- Output layer specified as `tesk.ext.output_layer`, defaulting to `cellbender` (or maybe something more general, e.g. `denoised` or `ambient_removed`)

With a little python logic to treat `X` differently

grst: That would be the most generic solution... but I think you mean `output_layer` defaults to X? 

Could you elaborate why this process is needed in the first place? It doesn't seem to do a lot and requires reading/writing the anndata object from disk which is inefficient. 

I'm also not sure under which circumstances you'd like to have cellbender counts in `.X` rather than a layer.

nictru:
Somehow the `_filtered.h5` and `_posterior.h5` seam to be unstable while the unfiltered `.h5` produces a stable hash. Any ideas? Otherwise I would change the testing approach. Could not find any option for setting random seed in cellbender.

grst:
Allegedly, it sets the seed already: 
https://github.com/broadinstitute/CellBender/issues/283#issuecomment-1726299475

However according to my experience these deep learning frameworks depend a lot on CPU architechture as well, so running on different machines with different number of available cores etc. will produce numerically different results.

nictru:
The github action runners should be quite identical in these concerns but I guess making the tests less strict will be ok. Interesting that the unfiltered is stable while the others are not.

> That would be the most generic solution... but I think you mean output_layer defaults to X?

I guess this refers to my idea with `ext.input_layer` and `ext.output_layer`: I don't care too much about the default, in scdownstream I have the raw counts backed up in a separate layer so I don't lose them (and I explicitly use the raw counts layer for whatever tools that require them). But I see that not having `X` as the default will be better here.

> Could you elaborate why this process is needed in the first place? It doesn't seem to do a lot and requires reading/writing the anndata object from disk which is inefficient.

scdownstream supports 4 different methods for removing ambient RNA. Each of them has a different way of providing the output. The `cellbender_merge` process brings the data into a uniform shape, so that downstream tasks do not have to handle different methods differently.

> Wouldn't that be equivalent to (for `load_anndata_from_input_and_output`)

The filtered and unfiltered inputs are provided by the user, in the scrnaseq pipeline case it will be the cellranger output. Cellbender also performs an empty droplet detection, which does not necessarily deliver the same output as the cellranger method. By setting `analyzed_barcodes_only` and using the `unfiltered` AnnData as input, we basically ignore the cellbender doublet detection and subset then to the cells present in the user-defined (cellranger) filtered AnnData.
Also the [cellbender documentation](https://cellbender.readthedocs.io/en/latest/reference/index.html#cellbender.remove_background.downstream.load_anndata_from_input_and_output) says that the first parameter for `load_anndata_from_input_and_output` needs to be the unfiltered file, that also was used for running `cellbender removebackground`.

grst:
I see, thanks for explaining! I think the proposed solution with `input_layer` and `output_layer` is great then!

nictru:
The `input_layer` method does not work for cellbender, as there is [no such option in the CLI](https://cellbender.readthedocs.io/en/latest/reference/index.html). It expects raw counts to be in `.X` but I guess this will not be a problem.

grst (approved):
LGTM now

PR #6156, opened 2024-08-09
`Fastqc`: FIX memory over-allocation

sguizard:
Fastqc will allocate to the JVM [--memory * --threads RAM ](https://github.com/s-andrews/FastQC/blob/1faeea0412093224d7f6a07f777fad60a5650795/fastqc#L211-L222). 
This results in the job to be killed by scheduler for exceeding the amount of RAM requested. 

Dividing the  `task.memory` by `task.cpus` allow to stick to requested amount of RAM.


SPPearce:
Can you please add a comment in the code explaining why it is required.

sguizard:
Sure! 
This is done.

SPPearce (approved):

sguizard:
Thanks Simon!

PR #6195, opened 2024-08-15
Pilon: allow adjustment of allocated memory

### LINKED ISSUE #6191: [FEATURE] Add memory allocation option to pilon

The default pilon command currently implemented does not allow the user to change the allocated memory such that pilon easily runs out of memory on bigger genomes. 

Pilon has no flag to change the allocated memory, it can only be changed via the JVM as shown [here](https://github.com/broadinstitute/pilon/wiki/Requirements-&-Usage#detailed-usage); the script needs to be adapted accordingly.

_No response_

### Additional context

_No response_

tstoeriko:
Changes:
 - call pilon via JVM/jar file to allow non-default memory requirements
 - remove deprecated --threads option

## PR checklist

Closes #6191


SPPearce:
Does this really want to try and make a heap size with the full amount of memory passed to it, with no overhead? Most of the java modules reduce it a bit.
I'm sure I saw a discussion on this recently but can't see it.

tstoeriko:
> Does this really want to try and make a heap size with the full amount of memory passed to it, with no overhead? Most of the java modules reduce it a bit. I'm sure I saw a discussion on this recently but can't see it.

Fair enough, I'll have another look at the other modules to figure out a better way to do this, thanks!

SPPearce:
They tend to subtract a gig or two from the allocated memory, or use a proportion (subtraction is probably better)
Check out fgbio and gatk modules or just search for the Xmx flag

tstoeriko:
I set the default to 8 GB as that is what the Pilon user guide recommends. If that's too much, I can decrease it. I'll have a look at how much the wrapper script usually allocates.

modules/nf-core/pilon/main.nf
```diff
@@ -28,11 +28,19 @@ process PILON {
     def prefix     = task.ext.prefix ?: "${meta.id}"
     def valid_mode = ["frags", "jumps", "unpaired", "bam"]
     if ( !valid_mode.contains(pilon_mode) )  { error "Unrecognised mode to run Pilon. Options: ${valid_mode.join(', ')}" }
+    def mem_mb = 1024
+    if (!task.memory) {
+        log.info '[Pilon] Available memory not known - defaulting to 8GB. Specify process memory requirements to change this.'
```
SPPearce, line 33: Is it not currently defaulting to 1?

tstoeriko: Yes, thanks! I changed the default, but missed this line.

SPPearce (approved):

PR #6213, opened 2024-08-19
Cellrangermulti patch

eolaniru:
#### current situation
In the [libraries] section of the multi config CSV, setting feature_types to VDJ enables auto-detection of the chain type. 

#### problems with this approach
- It fails if you have both TCR and BCR enriched fastq datasets
Please see [10x website](https://www.10xgenomics.com/support/software/cell-ranger/latest/analysis/running-pipelines/cr-5p-multi#autodetection)

#### solution
- best to specify `VDJ-T` for TCR and `VDJ-B` for BCR datasets in the `feature_type` column. This PR achieves that.

### DIFF
--- a/modules/nf-core/cellranger/multi/tests/main.nf.test.snap
+++ b/modules/nf-core/cellranger/multi/tests/main.nf.test.snap
@@ -2,13 +2,15 @@
     "cellranger - multi - 10k - PBMC": {
         "content": [
             [
-                "all_contig.fasta:md5,494adf38d81f749fc87409c982a1dd2d"
+                "all_contig.fasta:md5,494adf38d81f749fc87409c982a1dd2d",
+                "all_contig.fasta:md5,d41d8cd98f00b204e9800998ecf8427e"
             ],
             [
                 "all_contig.fasta.fai:md5,95e967ffde85602135d22764664d9e3e"
             ],
             [
-                "all_contig.fastq:md5,e5b0cec096f4d57981916d2247a9f9b7"
+                "all_contig.fastq:md5,e5b0cec096f4d57981916d2247a9f9b7",
+                "all_contig.fastq:md5,d41d8cd98f00b204e9800998ecf8427e"
             ],
             [
                 "consensus.bam.bai:md5,125978e04496fe193a5c162603a7e6f6",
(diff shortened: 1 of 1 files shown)


eolaniru:
nf-test fails if you make lint happy. will probably make nf-test pass and solve lint issues later

eolaniru:
@grst any suggestions on how to make lint pass for empty file MD5 sum `d41d8cd98f00b204e9800998ecf8427e`. I guess the test data for t-cell is light and generates the empty file but the module works with actual dataset. I see that this is common in other modules [here](https://github.com/search?q=repo%3Anf-core%2Fmodules+d41d8cd98f00b204e9800998ecf8427e&type=code)

klkeys:
> @grst any suggestions on how to make lint pass for empty file MD5 sum `d41d8cd98f00b204e9800998ecf8427e`. I guess the test data for t-cell is light and generates the empty file but the module works with actual dataset. I see that this is common in other modules [here](https://github.com/search?q=repo%3Anf-core%2Fmodules+d41d8cd98f00b204e9800998ecf8427e&type=code)

in the past, I have simply offered no MD5 for empty files

this is common for log files or files with unsorted output ([example](https://github.com/nf-core/modules/blob/8f6ac983108c53cd84a9141549d1e833e86b03e8/tests/modules/nf-core/getorganelle/fromreads/test.yml#L45))

while it's worked in the past, I'll defer to @grst whether or not this approach is desirable

klkeys (commented):
I'm leery of forcing users to specify T-cell or B-cell experiments -- the autodetection features in `cellranger` are pretty nifty -- but this PR does seem to fix a subtle problem

worth noting here that the corresponding [subworkflow](https://github.com/nf-core/scrnaseq/blob/master/subworkflows/local/align_cellrangermulti.nf) in nf-core/scrnaseq would need to change too

grst:
> while it's worked in the past, I'll defer to @grst whether or not this approach is desirable

I had the same issue in the past, and it seems just checking for existence of such a file is the way to go. 
(I'm not entirely happy with this solution, as there seems to be no way to check that a file is actually empty if it is supposed to be empty, but I guess it's good enough)

famosab:
@eolaniru Do you still have plans to finish this? Or can the PR be closed?

eolaniru:
sorry been carried away by other things by I plan to finish this soon. Thanks

SPPearce:
@eolaniru , @grst , any plans to actually finish this off?

grst:
Not a priority for me to move this forward at the moment, sorry!

eolaniru:
After further testing, I believe this approach is only needed when TR or IG mapping fails to satisfy Cell Ranger's autodetection criteria. Specifically, to distinguish between TR and IG chains, the following conditions must all be met:
  - A minimum of 10000 total reads
  - A minimum of 5.0% of the total reads needs to map to TR or IG
  - The number of reads mapped to TR should be at least 3.0x compared to the number of reads mapped to IG or vice versa. 
  
Except there are specific use cases, i think it can go back to a draft.

PR #6217, opened 2024-08-20
Add nfschema plugin subwf

nvnieuwk:
This PR adds the `nf-schema` equivalent of `utils_nfvalidation_plugin` 

This PR cannot be merged before nf-schema v2.1.0 has been released. Which needs one more feature ATM (https://github.com/nextflow-io/nf-schema/issues/43)


nvnieuwk:
Tests are failing because it's using the unreleased `nf-schema` v2.1.0 (which I tested with a locally installed version of the plugin)

subworkflows/nf-core/utils_nfschema_plugin/tests/nextflow.config
```diff
@@ -0,0 +1,8 @@
+plugins {
+    id "nf-schema@2.1.0"
+}
+
+validation {
+    parametersSchema = "${projectDir}/subworkflows/nf-core/utils_nfschema_plugin/tests/nextflow_schema.json"
```
adamrtalbot, line 6: A discussion for nf-schema, but it makes more sense for the schema to be an input for `validateParameters()` and `paramsSummaryLog()`. This way, you could support multiple schema JSON files per pipeline. e.g., for viral recon we could have one for Illumina and one for Nanopore.

subworkflows/nf-core/utils_nfschema_plugin/main.nf
```diff
@@ -0,0 +1,32 @@
...
+    take:
+    validate_params // boolean: validate the parameters
+
+    main:
+
+    //
+    // Print parameter summary to stdout. This will display the parameters
+    // that differ from the default given in the JSON schema
+    //
+    log.info paramsSummaryLog(workflow)
```
adamrtalbot, line 19: Can workflow be in the take statement? Any way of being precise with it instead of scooping it out global scope?

subworkflows/nf-core/utils_nfschema_plugin/tests/nextflow.config
nvnieuwk: It's still possible to supply parameter schemas to these functions :) I'd advise against using multiple JSON schemas for one pipeline though since that complicates it a lot

maxulysse: and for a meta pipeline?

nvnieuwk: Wouldn't `workflow` be the object of the meta pipeline then? These subworkflows are also made to run at the start of a meta pipeline and not at the start of each part, right?

adamrtalbot: Never assume anything.

nvnieuwk: Good point, I'll give in and add it as an input 😁

nvnieuwk: Also seems like I mixed up both conversations here 😁

nvnieuwk:
@maxulysse and @adamrtalbot I added the workflow as a mandatory input to the subwf. I also added parameters schema as an optional input. This is optional because in normal pipeline I'd always advise to only set the schema using the `validation.parametersSchema` config value to keep it consistent for the whole pipeline. It would be wise though to use this input for meta pipelines.

nvnieuwk:
> What about the failing tests?

Seems like some other modules were failing, I updated the branch :p

maxulysse (approved):

Issue #4570, opened 2023-12-11
[FEATURE] Add `stub` support to every module


ewels:
Using command `stub` blocks in Nextflow is useful when developing syntax and quickly testing pipelines. However, its use within nf-core is limited because the vast majority of nf-core modules do not have `stub` blocks.

My suggestion is to require that every module should have a `stub` block. All 1095 of them :trollface: 

We can lint for the presence of this in linting and also potentially run with `-stub-run` in the module CI.

Although adding a stub block for one or two modules is not difficult, adding it for all will be a significant effort. This could be a nice project focus for a future hackathon(s).
```[tasklist]
### Tasks
```

GallVp:
Hi @ewels 

This is such a good idea and can be immensely helpful when implementing dataflows.

Here is something that I have been wondering about and I thought I'll share it here to get your feedback. With nf-test snapshot feature, it may be possible to partially automate `stub` generation and testing.

If there is a function in nf-test (`snapshot.matchStub("test-name")`) which  compares the structure of process outputs against a snapshot while ignoring md5sums, it may be able to suggest stub code such as:

```groovy
touch ${prefix}.gff3
touch ${prefix}.report.html

...versions.yml...
``` 

It won't be able to tackle complex situations but can provide starter code. Moreover, the `stub` testing will become more robust by expecting a complete match of output structure across `script` and `stub` for the same inputs and configuration. 

Here is a pull request where I have tried to implement something similar but manually: https://github.com/nf-core/modules/pull/4627

ewels:
That would be great 👌🏻 @mashehu / @mirpedrol - what do you think: could we make a #tools command to scaffold a `stub` from nf-test outputs / snapshots?

GallVp:
I think if `snapshot.matchStub("test-name")` is implemented by nf-test,  nf-core tools might not have to change at all. I am hesitant to commit time at this point but keen to contribute after March 2024. Probably someone will pick it up before that otherwise, I am keen to give it a go.

GallVp:
I have tried to implement something similar in logic to `snapshot.matchStub` for the fastp module here: https://github.com/nf-core/modules/pull/4637

Each test creates a sorted list of outputs which should be matched by the same test with the `-stub` option.

```groovy
{
    assert snapshot(
        (
            [process.out.reads[0][0].toString()] + // meta
            process.out.reads.collect { file(it[1]).getName() } +
            process.out.json.collect { file(it[1]).getName() } +
[...]

PR #6219, opened 2024-08-20
bcftools_conensus stub section and tests

DimitriTernovoj:
## Description

The PR adds a stub section along side some tests for the bcftools_consensus module and therefore contributes to the goal of all modules having a stub section (and tests) (https://github.com/nf-core/modules/issues/4570). Furthermore I added another component to the input, namely `mask`. This is an optional input file, but should also be included in the `input` channel definitions according to the module specifications.

## PR checklist


modules/nf-core/bcftools/consensus/main.nf
```diff
@@ -20,6 +20,7 @@ process BCFTOOLS_CONSENSUS {
     script:
     def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
+    def masking = mask ? "-m $mask" : ""
```
SPPearce, line 23: You are not using this in the command below.

DimitriTernovoj: you are correct, thanks for pointing out. I messed up my commits with this one :I. I'll correct that!

SPPearce:
You need to remove the pytest files from `tests/nf-core/bcftools/consensus`, as well as the entry in `tests/pytest-config.yml`

Issue #6226, opened 2024-08-21
[FEATURE] Port pytest with `--migrate-pytest-hard`


GallVp:
### Is your feature request related to a problem? Please describe

Use `nf-core -v modules create --migrate-pytest --migrate-pytest-hard <tool/subtool>` to migrate to nf-test.

I am experimenting with this command variant to automate pytest migration as much as possible. See fork: https://github.com/GallVp/nf-core-tools/tree/migpyt

mashehu:
Interesting! Could you open a PR against `dev` to see the changes you made?

GallVp:
> Interesting! Could you open a PR against `dev` to see the changes you made?

Thank you @mashehu 

I have created a draft PR: https://github.com/nf-core/tools/pull/3138

GallVp:
Since a language parser using Antlr has now been implemented for Nextflow under the [nextflow/language-server](https://github.com/nextflow-io/language-server) project, I am using that to create a more universal pytest to nf-test converter here: https://github.com/GallVp/pytest2nf-test

Will be able to resume conversion of remaining components soon.

PR #6257, opened 2024-08-22
Added nf-test for rapidnj

GallVp:
Added nf-test for rapidnj with #6226


modules/nf-core/rapidnj/tests/main.nf.test
```diff
@@ -0,0 +1,36 @@
...
+
+    name "Test Process RAPIDNJ"
+    script "../main.nf"
+    process "RAPIDNJ"
+
+    tag "modules"
+    tag "modules_nfcore"
+    tag "rapidnj"
+
+    test("test-rapidnj") {
```
Joon-Klaps, line 12:
```suggestion
    test("sarscov2-fasta") {
```

PR #6281, opened 2024-08-23
Added nf-test for purecn/normaldb

GallVp:
Added nf-test for purecn/normaldb with #6226


GallVp:
Conda is broken. The hope is that it will be fixed by applying the following updates:

- bioconda::bioconductor-org.hs.eg.db 3.16.0 -> 3.18.0  
- bioconda::bioconductor-purecn 2.4.0 -> 2.8.1 
- bioconda::bioconductor-txdb.hsapiens.ucsc.hg38.knowngene 3.16.0 -> 3.18.0

GallVp:
Thank you @SPPearce 

Now that we have updated the conda.yml, we also need to update the containers. The tests are passing because the version is hard coded. https://github.com/nf-core/modules/blob/7296cb782d08e5b80c7a2dc3470240db41d0a372/modules/nf-core/purecn/normaldb/main.nf#L31

SPPearce:
Hah, didn't notice that.
We should get the version from the R code instead. 
Seqera containers are the way to go for mulled containers

GallVp:
> Hah, didn't notice that. We should get the version from the R code instead. Seqera containers are the way to go for mulled containers

I'll do it then. I'll replace it with a wave container.

GallVp:
> > Hah, didn't notice that. We should get the version from the R code instead. Seqera containers are the way to go for mulled containers
> 
> I'll do it then. I'll replace it with a wave container.

@SPPearce 

I have added the wave containers. Thank you!

SPPearce (approved):

lbeltrame:
Whoops. I was on holiday and I didn't notice this. Thanks Simon for doing the review!

PR #6288, opened 2024-08-25
Orthofinder

### LINKED ISSUE #6287: Add "resume" (-b) option in orthofinder module

There is a useful resume function within the orthofinder tool itself (-b). Starting from the reciprocal blast part (which takes the longest). This way, you can add just one new species, and not start at the beginning.

Add optional input to orthofinder with a directory (WorkingDirectory for orthofinder)

The other option is to write a separate module, but seems wasteful

### Additional context

_No response_

chriswyatt1:
## PR checklist

Closes #6287

Added a new test and an optional input in the form of a directory of a previous orthofinder run.
- [x] Add a resource `label`
  - For modules:


chriswyatt1:
Hey, 

I added the resume function of orthofinder. The only annoying thing, I think, is that now you need to supply two inputs, but second can be [] null, if you don't want to continue from a previous WorkingFolder. 

I also messed around a little bit with the nf-tests, as I found that they didn't always repeat in the same way. So thats worth checking too. If this is too annoying a change, I could make it a separate module (maybe??)

modules/nf-core/orthofinder/tests/main.nf.test
```diff
@@ -52,7 +51,51 @@ nextflow_process {
...
+    test("sarscov2 - candidatus_portiera_aleyrodidarum - proteome - resume") {
+
+        when {
+            process {
+                """
+                file(params.modules_testdata_base_path + 'genomics/sarscov2/genome/proteome.fasta', checkIfExists: true)
+                    .copyTo("${workDir}/sarscov2.fasta")
+
+                def file_a = file("https://raw.githubusercontent.com/nf-core/test-datasets/proteinfold/testdata/sequences/H1065.fasta")
+                def file_c = file("s3://temporary-nfcore-module-data/test/WorkingDirectory")
```
chriswyatt1, line 71: This line will change to :

def file_c = file("https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/delete_me/orthofinder/WorkingDirectory")

ONce the PR is through on test datasets

modules/nf-core/orthofinder/main.nf
```diff
@@ -18,22 +19,28 @@ process ORTHOFINDER {
...
+        $include_command $args
+
+    if [ -e input/OrthoFinder/Results_$prefix ]; then
+        mv input/OrthoFinder/Results_$prefix $prefix
+    fi
 
-    mv \\
-        input/OrthoFinder/Results_$prefix \\
-        $prefix
+    if [ -e WorkingDirectory/OrthoFinder/Results_$prefix ]; then
```
GallVp, line 41: Thank you for these enhancements @chriswyatt1

Does `WorkingDirectory` here mean that to correctly use this module in the resume mode, the data from the previous run has to be placed in a folder named `WorkingDirectory`? Can we make the name dynamic by using the  `prior_run` variable here?

modules/nf-core/orthofinder/tests/main.nf.test
```diff
@@ -52,7 +51,51 @@ nextflow_process {
             def stable_files = all_files.findAll { it.name in stable_file_names }
 
             assert snapshot(
-                all_file_names,
+                stable_files,
```
GallVp, line 54: To avoid test fails, `stable_files` should be swapped with `stable_files.toSorted()`.

modules/nf-core/orthofinder/tests/main.nf.test.snap
```diff
@@ -1,322 +1,18 @@
...
-                "sarscov2.tsv",
-                "sarscov2.tsv",
-                "sarscov2__v__candidatus_portiera_aleyrodidarum.tsv",
-                "test_search_results.txt.gz"
-            ],
             [
                 "Statistics_PerSpecies.tsv:md5,984b5011a34d54527fe17896bfa36a2d",
-                "SpeciesTree_Gene_Duplications_0.5_Support.txt:md5,8b7a673e2e8b6d1aeb697f2bb88afa18",
-                "SpeciesTree_rooted.txt:md5,4d5ea525feebe479fca0c0768271ba81"
+                "SpeciesTree_rooted.txt:md5,4d5ea525feebe479fca0c0768271ba81",
```
GallVp, line 6: I think these will stop moving around if we used `stable_files.toSorted()`.

GallVp:
> Hey,
> 
> I added the resume function of orthofinder. The only annoying thing, I think, is that now you need to supply two inputs, but second can be [] null, if you don't want to continue from a previous WorkingFolder. If this is too annoying a change, I could make it a separate module (maybe??)
> 
> I also changed the nf-tests (slightly), as I found that they didn't always repeat in the same way. So that may need to be checked again.
> 
> Also, I need to put the input WorkingFolder onto nf-core module/datasets,, so the runner test will fail for now. Will do a PR

I prefer a single module as it will be easier to maintain and will allow us to pool our resources into improving this together.

modules/nf-core/orthofinder/tests/main.nf.test
chriswyatt1: Great, I see now

modules/nf-core/orthofinder/main.nf
chriswyatt1: Sorry, I was away these last few days. 
Yes, you are right:

WorkingDirectory is the automatic name that Orthofinder creates inside the results folder. 
So when you run without a previous "WorkingDirectory", currently the script writes to the folder  (-f input), then a folder called "OrthoFinder/" and then to a test folder  (-n $prefix). Complicated,... so it then moves this back to the current directory:
`mv input/OrthoFinder/Results_$prefix $prefix`

In the `Results_$prefix` folder, you find the `WorkingDirectory`.

When you run with the resume flag, you need to supply the WorkingDirectory  in (-b)`input/OrthoFinder/Results_$prefix/WorkingDirectory`, which is moved to `$prefix/WorkingDirectory`, in that mv statement above. This then saves the output to mv `WorkingDirectory/OrthoFinder/Results_$prefix` (when running resume). So it runs out of the previously run folder. 

In (long) summary, we can use: `if [ -e ${prior_run}/OrthoFinder/Results_$prefix ]; then`

modules/nf-core/orthofinder/tests/main.nf.test.snap
chriswyatt1: Great, yes, this solved it

modules/nf-core/orthofinder/main.nf
GallVp: Thank you for the detailed response.

chriswyatt1:
Will need to tar.gz the workingDirectory and then we can untar in within the test like:

https://github.com/nf-core/modules/blob/master/modules/nf-core/custom/tx2gene/tests/main.nf.test
[4:47](https://nextflow.slack.com/archives/C049MBCEW06/p1710866845949069?thread_ts=1710865783.934309&cid=C049MBCEW06)
the modules repo would be better than the rnaseq one

chriswyatt1:
Waiting for https://github.com/nf-core/test-datasets/pull/1310 to be approved, then I can adjust the input on the test to this dir

chriswyatt1:
conda is failing, but not sure why

modules/nf-core/orthofinder/main.nf
```diff
@@ -9,31 +9,39 @@ process ORTHOFINDER {
...
 
     orthofinder \\
-        $args \\
         -t $task.cpus \\
         -a $task.cpus \\
         -p temp_pickle \\
         -f input \\
-        -n $prefix
+        -n $prefix \\
+        $include_command $args
```
GallVp, line 36: $include_command \\
$args

GallVp:
> conda is failing, but not sure why

It might be related to https://github.com/davidemms/OrthoFinder/issues/873

Can you look into the docker or singularity container for the diamond version and then lock it in the `environment.yml` file. This trick often solves conda problems.

chriswyatt1:
Thanks Usman, I may try that, but was playing around with the environment.yml and I think it works now. 

I had to specify the bioconda channel for the orthofinder tool, and I think that was the issue:

`  - bioconda::orthofinder=2.3.3`

chriswyatt1:
@GallVp , yes the force to diamond latest version fixed the issue. Thanks!

GallVp (approved):

GallVp:
LGTM!

GallVp:
Thank you @chriswyatt1

PR #6291, opened 2024-08-25
Added nf-test for picard/scatterintervalsbyns

GallVp:
Added nf-test for picard/scatterintervalsbyns with #6226


modules/nf-core/picard/scatterintervalsbyns/tests/main.nf.test
nvnieuwk, line 1: Can you add a stub test too? (and a stub block if that's missing)

GallVp: Hi @nvnieuwk 

I have been able to add a bit more automation. Now, when a pytest is converted to nf-test, a stub test will be added if the `stub:` block is present in the module `main.nf`, even if the pytest does not include a stub test.

Cases where `stub:` block is missing need to wait though.

For this PR, I have added a stub test manually.

Thank you!

SPPearce: Amazing

SPPearce (approved):

PR #6302, opened 2024-08-25
Added nf-test for gatk4/splitintervals

GallVp:
Added nf-test for gatk4/splitintervals with #6226


modules/nf-core/gatk4/splitintervals/tests/main.nf.test
nvnieuwk, line 1: Can you also add a stub test here?

GallVp: Done. Thank you!

SPPearce (approved):

PR #6368, opened 2024-08-27
Added nf-test for snippy/run

GallVp:
Added nf-test for snippy/run with #6226


GallVp:
Tests are failing despite passing on my macOS.

SPPearce (approved):

GallVp:
> Previous error message was suggesting snippy was trying to assign 8gb of RAM for sorting; telling it we have task.memory available should fix the issue.

Thank you!

PR #6380, opened 2024-08-27
Added nf-test for pureclip

GallVp:
Added nf-test for pureclip with #6226


modules/nf-core/pureclip/main.nf
```diff
@@ -56,10 +56,12 @@ process PURECLIP {
...
     def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
+    crosslinks_output_name = "${prefix}_pureclip_crosslinks.bed"
+    peaks_output_name      = "${prefix}_pureclip_peaks.bed"
 
     """
-    touch ${prefix}_pureclip_crosslinks.bed
-    touch ${prefix}_pureclip_peaks.bed
+    touch $crosslinks_output_name
+    touch $peaks_output_name
```
SPPearce, line 64: Curious as to why you added a command for this?

SPPearce (approved):

modules/nf-core/pureclip/main.nf
GallVp: Actually the author had used the variable names in the output emit statements. To match with that I had to copy the variables and then I guess I also pasted them in the command to minimise chances of name mismatches now and in the future.

PR #6392, opened 2024-08-27
Added nf-test for cnvkit/genemetrics

GallVp:
Added nf-test for cnvkit/genemetrics with #6226


modules/nf-core/cnvkit/genemetrics/main.nf
marrip, line 1: These appear empty for some reason.

marrip:
Thank you for adding nf-test! The testing part looks fine, it is just that the module `main.nf` and `meta.yml` are marked as empty files.

modules/nf-core/cnvkit/genemetrics/main.nf
GallVp: Thank you for the review. You are right. They appear empty. Strange! They are not empty though as you can click on the three dots and look at the file. The tests are also passing.

GallVp: Actually, the empty files mean that nothing was changed. Only the permissions were changed from 755 to 644. I did not do it. It was part of the nf-test migration done by nf-core/tools.

marrip (approved):
lgtm!

PR #6404, opened 2024-08-28
Added nf-test for epang/place

GallVp:
Added nf-test for epang/place with #6226


erikrikarddaniel (commented):
Looks good, but the Singularity test fails on the `*.jplace.gz` md5 sum. I've had problems with stability of this one as there might be rounding differences in the likelihoods. Waiting with approval for that reason.

GallVp:
> Looks good, but the Singularity test fails on the `*.jplace.gz` md5 sum. I've had problems with stability of this one as there might be rounding differences in the likelihoods. Waiting with approval for that reason.

Thank you @erikrikarddaniel 

I have moved it to name only and have also added a `stub` block.

PR #6412, opened 2024-08-28
Added nf-test for untarfiles

GallVp:
Added nf-test for untarfiles with #6226


modules/nf-core/untarfiles/tests/main.nf.test
```diff
@@ -0,0 +1,104 @@
...
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
+            )
+        }
+    }
+
+    test("test-untarfiles-different-output-path") {
```
famosab, line 34: I think I dont get the naming of the test and the idea behind the test. Does this not do the same thing as test-untarfiles?

GallVp: Thank you @famosab 

I think the idea is to test the case where the tar has multiple subdirectories in it.

famosab: ok maybe the test can be renamed then to: 
```suggestion
    test("test-untarfiles-subdirectories") {
```

maxulysse:
I was more thinking about getting rid of this modules, untar can work it out just as nicely

GallVp:
> I was more thinking about getting rid of this modules, untar can work it out just as nicely

Good idea. Should we add a deprecation notice and merge?

modules/nf-core/untarfiles/tests/main.nf.test
GallVp: Good idea. That's done now.

SPPearce:
@maxulysse , @GallVp , are we adding a deprecation notice here?

GallVp:
> @maxulysse , @GallVp , are we adding a deprecation notice here?

Done Thank you!

SPPearce (approved):

PR #6415, opened 2024-08-28
Updated nf-test for strelka/germline

GallVp:
Deleted residual files


SPPearce (approved):
Did this one have pytest and nf-test still?

GallVp:
> Did this one have pytest and nf-test still?

Yes, someone those buggers were still hanging in there.

PR #6422, opened 2024-08-28
Added nf-test for qcat

GallVp:
Added nf-test for qcat with #6226


modules/nf-core/qcat/tests/main.nf.test
```diff
@@ -0,0 +1,64 @@
...
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(
+                    process.out.reads,
+                    process.out.versions,
+                    file(process.out.versions[0]).readLines()[1].trim() // trap to catch conda version mismatches
```
SPPearce, line 33: Why this bit?

GallVp: Thank you for the review @SPPearce 

This module was failing due to `versions.yml` md5 mismatch under the conda profile. On my Mac, I only run the docker profile. To debug the mismatch, I setup this trap. In the past, I used to remove the trap after debugging. But I think, the modules which fail often, we should leave this trap as is to help future developers debug the mismatches.

SPPearce: Yes, I had this thought after seeing you do that that we should just be shapshotting the versions content directly

SPPearce (approved):

modules/nf-core/qcat/tests/main.nf.test
GallVp: Yes, I agree. That will be helpful in debugging. It is also more declarative than the md5. I think the md5 should be the last resort for most things because it is a black box.

PR #6428, opened 2024-08-28
Revert "Add tbi index as an output to lofreq callparallel"

nevinwu:
Reverts nf-core/modules#6427

### DIFF
--- a/modules/nf-core/lofreq/callparallel/tests/main.nf.test
+++ b/modules/nf-core/lofreq/callparallel/tests/main.nf.test
@@ -32,10 +32,7 @@ nextflow_process {
         then {
             assertAll(
                 { assert process.success },
-                { assert snapshot(
-                    process.out.versions,
-                    process.out.tbi.collect { file(it[1]).getName() }
-                    ).match() },
+                { assert snapshot(process.out.versions).match() },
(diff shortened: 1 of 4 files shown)


modules/nf-core/lofreq/callparallel/tests/main.nf.test
```diff
@@ -32,10 +32,7 @@ nextflow_process {
         then {
             assertAll(
                 { assert process.success },
-                { assert snapshot(
-                    process.out.versions,
-                    process.out.tbi.collect { file(it[1]).getName() }
-                    ).match() },
+                { assert snapshot(process.out.versions).match() },
```
famosab, line 35: Now you are missing the test for the tbi file
```suggestion
                { assert snapshot(process.out.versions,
                process.out.tbi).match() },
```

modules/nf-core/lofreq/callparallel/tests/main.nf.test
```diff
@@ -65,10 +62,7 @@ nextflow_process {
...
```
famosab, line 65: Now you are missing the test for the tbi file
```suggestion
                { assert snapshot(process.out.versions,
                process.out.tbi).match() },
```

nevinwu:
Perfect, then I'm updating the test and doing another PR! :)

PR #6442, opened 2024-08-29
Added nf-test for scimap/mcmicro

GallVp:
Added nf-test for scimap/mcmicro with #6226


GallVp:
Just needs conda profile exclusion. Otherwise, all good! Was a tough one!

modules/nf-core/scimap/mcmicro/tests/main.nf.test
```diff
@@ -0,0 +1,63 @@
...
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(
+                    file(process.out.annotedDataCsv[0][1]).readLines()[0],  // unstable
+                    file(process.out.annotedDataH5ad[0][1]).name, // unstable
```
SPPearce, line 32: Hol' up, `annotedDataCsv` could do with changing for two reasons:

1. It is not in snake_case
2. It should presumably be annotated rather than annoted

GallVp: Thank you!

I have simplified the output channel names.

SPPearce (approved):
0.22.0 -> 2.1.3 is also quite the version change!
Thanks for indulging my dislike of the channel names, I know it isn't your aim here.

GallVp:
> 0.22.0 -> 2.1.3 is also quite the version change!
> 
> Thanks for indulging my dislike of the channel names, I know it isn't your aim here.

No worries at all. My pleasure. It's good to weed out these gremlins along the way.

Issue #5601, opened 2024-05-14
deprecate pbbam/pbmerge (maybe new module: pbtk/pbmerge)


famosab:
### Is there an existing module for this?

- [X] I have searched for the existing module

### Is there an open PR for this?

- [X] I have searched for existing PRs

### Is there an open issue for this?

- [X] I have searched for existing issues

### Are you going to work on this?

- [ ] If I'm planning to work on this module, I added myself to the `Assignees` to facilitate tracking who is working on the module

### Additional info
This relates to #4753:  

> It seems to me that [pbbam was archived](https://github.com/PacificBiosciences/pbbam) and the individual tools were transferred to [pbtk](https://github.com/PacificBiosciences/pbtk). Tackle the proper update on a separate branch with a separate PR. 

After this new module is created, `pbbam/pbmerge` needs to be removed.

sguizard:
pbmerged was used in the first version of the isoseq pipeline. After a modification of the analysis strategy, pbmerge has been removed and not used by any pipeline. 
I've kept maintained it along with the other Pacbio modules to keep available in the hypothesis where would like to use it. 
Unless someone planned to use it in a new pipeline, I don't think it's necessary to keep a module for pbmerge.

SPPearce:
Note that pbmerge should be deprecated following the guidelines in: https://nf-co.re/docs/contributing/modules#deprecating-a-module

famosab:
Closed by: #6447

PR #6447, opened 2024-08-29
Added nf-test for pbbam/pbmerge

GallVp:
Added nf-test for pbbam/pbmerge with #6226


famosab:
I think this module needs to be deprecated #5601

GallVp:
> I think this module needs to be deprecated #5601

Thank you @famosab 

I have added a deprecation message.

SPPearce (approved):

PR #2904, opened 2023-02-21
new module paragraph/idxdepth

nvnieuwk:
## PR checklist

Closes #2903 

- [ ] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [ ] Remove all TODO statements.
- [ ] Emit the `versions.yml` file.
- [ ] Follow the naming conventions.
- [ ] Follow the parameters requirements.
- [ ] Follow the input/output options guidelines.
- [ ] Add a resource `label`
- [ ] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - [ ] `PROFILE=docker pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [ ] `PROFILE=singularity pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [ ] `PROFILE=conda pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`

PR #6449, opened 2024-08-29
Added nf-test for paragraph/idxdepth

GallVp:
Added nf-test for paragraph/idxdepth with #6226


GallVp:
Everything seems to be working except the `cram` test on conda. Thank you!

famosab:
From the original creation of the module it just says that there was a faulty dependency? Maybe that leads to an issue: https://github.com/nf-core/modules/pull/2904#issuecomment-1439604755

PR #6462, opened 2024-08-29
Added nf-test for dshbio/filterbed

GallVp:
Added nf-test for dshbio/filterbed with #6226


modules/nf-core/dshbio/filterbed/tests/main.nf.test
```diff
@@ -0,0 +1,63 @@
...
+
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(
+                    file(process.out.bed[0][1]).name, // empty
```
famosab, line 32: Why do you only test for the file name? Is the bed not stable enough to check the md5sum?

heuermh: Don't have this checked out at the moment, but yes it probably is empty.  The filter looks to be `--range chr1:0-1000` which won't match anything from sarscov2 genome.

famosab: Can we change it to something that leads to a non-empty file which can then be checked with md5sums?

GallVp: Thank you @heuermh and @famosab 

I have updated the params so that the output is not empty.

heuermh (approved):
LGTM, thank you!

PR #6513, opened 2024-09-01
Added nf-test for cnvpytor/partition

GallVp:
Added nf-test for cnvpytor/partition with #6226


modules/nf-core/cnvpytor/callcnvs/main.nf
```diff
@@ -23,7 +23,7 @@ process CNVPYTOR_CALLCNVS {
     """
```
SPPearce, line 23: This has no args.

modules/nf-core/cnvpytor/partition/tests/main.nf.test
```diff
@@ -0,0 +1,60 @@
...
+    test("test-cnvpytor-partition") {
+
+        when {
+            process {
+                """
+                input[0] = [
+				    [ id:'test'], // meta map
+				    file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/pytor/test.pytor', checkIfExists: true)
+				]
+                input[1] = []
```
SPPearce, line 23: Think we need a test checking that this works, given that it didn't previously.

modules/nf-core/cnvpytor/partition/tests/nextflow.config
SPPearce, line 1: Don't think this is doing anything as args is missing, and suspect it should be in the value channel anyway.

modules/nf-core/cnvpytor/partition/main.nf
```diff
@@ -19,15 +19,15 @@ process CNVPYTOR_PARTITION {
     task.ext.when == null || task.ext.when
```
SPPearce, line 20: No args here either.

SPPearce (changes requested):
Think this module needs some looking at, not sure how functional it really is.
Be worth checking if we can get prefix working too, rather than relying on pytor.baseName.

GallVp:
Thank you @SPPearce 

This one needs some serious fixing. The test data produces empty outputs.

```bash
cnvpytor -root test.dup.pytor -partition 100 1000 10000
```

Currently the module is not creating any output. It is simply passing the input as output.

PR #6515, opened 2024-09-01
Added nf-test for crumble

GallVp:
Added nf-test for crumble with #6226


modules/nf-core/crumble/main.nf
```diff
@@ -47,4 +47,25 @@ process CRUMBLE {
...
     END_VERSIONS
     """
+
+    stub:
+    def args       = task.ext.args ?: ''
+    def prefix     = task.ext.prefix ?: "${meta.id}"
+    def extension  = args.contains("-O sam") ? "sam" :
+                    args.contains("-O bam") ? "bam" :
+                    args.contains("-O cram") ? "cram" :
+                    "sam"
```
SPPearce, line 57: Ewww, defaulting to sam? Can we change this to default to bam?

SPPearce (approved):

modules/nf-core/crumble/main.nf
GallVp: Good point. I have changed it to `bam` now.

PR #6517, opened 2024-09-02
Added nf-test for elprep/filter

GallVp:
Added nf-test for elprep/filter with #6226


modules/nf-core/elprep/filter/main.nf
```diff
@@ -65,6 +65,8 @@ process ELPREP_FILTER {
     def assembly_regions_cmd = get_assembly_regions ? " --assembly-regions ${prefix}.assembly_regions.igv": ""
 
     """
+    mkdir logs
```
SPPearce, line 68: Do the logs get emitted?

GallVp: Thank you!

No the logs are not emitted currently. Due to the permissions issue, the docker self hosted was failing. So I added a command to create a logs folder in the current directory. Fortunately, the tool has a `--log-path` parameter.

GallVp: Hi @SPPearce 

Should we emit the logs? Or is this good to go?

SPPearce: If they are informative at all, should probably emit them

GallVp: Good idea. That's done now.

SPPearce (approved):

PR #6521, opened 2024-09-02
Added nf-test for getorganelle/config

GallVp:
Added nf-test for getorganelle/config with #6226


modules/nf-core/getorganelle/config/main.nf
```diff
@@ -35,7 +35,7 @@ process GETORGANELLE_CONFIG {
     def args = task.ext.args ?: ''
     """
     mkdir -p getorganelle/{LabelDatabase,SeedDatabase}
-    touch getorganelle/{LabelDatabase,SeedDatabase}/${organnel_type}.fasta
```
SPPearce, line 38: Aah, this is why we test the stubs...

modules/nf-core/getorganelle/config/tests/main.nf.test
```diff
@@ -0,0 +1,73 @@
...
+                input[0] = 'all'
+
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                {
```
SPPearce, line 28: Why is this so complicated?

modules/nf-core/getorganelle/config/main.nf
GallVp: Yes, I have seen quite a few broken stubs now which were never tested.

modules/nf-core/getorganelle/config/tests/main.nf.test
GallVp: It is because of the nested folders. The emitted folder contains subfolders which contain files. Using the simple `files(xyz).listFiles()` is not an option.

SPPearce (approved):

PR #6532, opened 2024-09-02
Added nf-test for kat/hist

GallVp:
Added nf-test for kat/hist with #6226


mahesh-panchal (dismissed):

mahesh-panchal:
I'm wondering if this module should be deprecated now. I don't think Kat is maintained anymore and there's both Merqury and MerquryFK now.

GallVp:
> I'm wondering if this module should be deprecated now. I don't think Kat is maintained anymore and there's both Merqury and MerquryFK now.

Thanks @mahesh-panchal 

The conda is also failing. If we agree to deprecate it, I'll remove the conda, add the deprecation message and ping you again for another look.

mahesh-panchal:
Do you know why conda is failing? I can't see what the error message is from the PR logs, but running in Gitpod, it's to do with needing to use the classic conda solver. 

Bernardo hasn't made a release since 2018, and while the issues do seem active on the repository, I think the tool is in maintenance mode at best. The problem is, I'm also not sure how to do the deprecation note. Merqury is the maintained version of the tool, but only does a subset of what KAT does. MerquryFK does more, but isn't really maintained.

GallVp:
> Do you know why conda is failing? I can't see what the error message is from the PR logs, but running in Gitpod, it's to do with needing to use the classic conda solver.
> 
> Bernardo hasn't made a release since 2018, and while the issues do seem active on the repository, I think the tool is in maintenance mode at best. The problem is, I'm also not sure how to do the deprecation note. Merqury is the maintained version of the tool, but only does a subset of what KAT does. MerquryFK does more, but isn't really maintained.

It says, 

>   Plotting K-mer frequency distributions ... done.  Saved to: None
>
> corrupted double-linked list
> .command.sh: line 6:  9563 Aborted                 (core dumped) kat hist --threads 2 --output_prefix test.hist -d test2_1.fastq.gz

It might be a memory related issue. But then docker and singularity have the same amount of memory. I'll try to pin the dependency versions to see if that helps.

mahesh-panchal:
Yeah, I don't know how to solve that. I've never come across it before.

modules/nf-core/kat/hist/environment.yml
```diff
@@ -5,3 +5,7 @@ channels:
   - bioconda
 dependencies:
   - bioconda::kat=2.4.2
+  - conda-forge::python=3.8.5 # Locked with the container
+  - conda-forge::numpy=1.19.1 # Locked with the container
+  - conda-forge::scipy=1.5.2 # Locked with the container
+  - conda-forge::matplotlib=3.3.1 # Locked with the container
```
vagkaratzas, line 11: Do you need to specify these separately here? Are they not included in the `bioconda::kat` environment anyway?
What does `# Locked with the container` imply?

vagkaratzas (changes requested):
Just a minor comment. Other than the `java.lang.NullPointerException: Cannot invoke method getAt() on null object for Docker and Conda`, it looks good.

PR #6558, opened 2024-09-05
Added nf-test for purecn/intervalfile

GallVp:
Added nf-test for purecn/intervalfile with #6226


lbeltrame (commented):
LGTM. Waiting for feedback from others if need be.

SPPearce (approved):

PR #6562, opened 2024-09-05
Added nf-test for msisensor/scan

GallVp:
Added nf-test for msisensor/scan with #6226


chriswyatt1:
I think this relates to : https://github.com/ding-lab/msisensor/issues/1

There is a version 0.6 (instead of 0.5), but it is not on conda. Think that would fix the segmentation fault error. 

This package is archived, and a new one (https://github.com/xjtu-omics/msisensor-pro)  is available that does the same thing, so probably this module should be updated in pipelines that use it

GallVp:
> I think this relates to : [ding-lab/msisensor#1](https://github.com/ding-lab/msisensor/issues/1)
> 
> There is a version 0.6 (instead of 0.5), but it is not on conda. Think that would fix the segmentation fault error.
> 
> This package is archived, and a new one (https://github.com/xjtu-omics/msisensor-pro) is available that does the same thing, so probably this module should be updated in pipelines that use it

Thank you @chriswyatt1

PR #6598, opened 2024-09-05
Added nf-test for gubbins

GallVp:
Added nf-test for gubbins with #6226


modules/nf-core/gubbins/tests/main.nf.test
```diff
@@ -0,0 +1,63 @@
...
+
+    name "Test Process GUBBINS"
+    script "../main.nf"
+    process "GUBBINS"
+
+    tag "modules"
+    tag "modules_nfcore"
+    tag "gubbins"
+
+    test("test-gubbins") {
```
sateeshperi, line 12: The test name `test-gubbins` is obvious so, better to have the name of the test contain details of the test-data being used.

Thoughts ?

sateeshperi (approved):
rest lgtm

modules/nf-core/gubbins/tests/main.nf.test
GallVp: Good idea. I have updated the names now!

PR #6645, opened 2024-09-13
Update output to use type dir for compatibility with tes.

### LINKED ISSUE #6644: update module: sratools /prefetch

vsmalladi:
## PR checklist

Closes #6644 

### DIFF
--- a/modules/nf-core/sratools/prefetch/main.nf
+++ b/modules/nf-core/sratools/prefetch/main.nf
@@ -13,7 +13,7 @@ process SRATOOLS_PREFETCH {
     path certificate
 
     output:
-    tuple val(meta), path(id), emit: sra
+    tuple val(meta), path(id, type: 'dir'), emit: sra
     path 'versions.yml'      , emit: versions
 
     when:


SPPearce:
Why is this necessary?

vsmalladi:
For Ga4gh_tes plugin it’s necessary to explicitly call out directories

SPPearce (approved):

ewels:
We should probably try to do this systematically for all modules 🤔 In this case it was already annotated in the meta yaml:

https://github.com/nf-core/modules/blob/a25abe62ccf3fe3767d2449d8322a35ab62844e0/modules/nf-core/sratools/prefetch/meta.yml#L44

vsmalladi:
> We should probably try to do this systematically for all modules 🤔 In this case it was already annotated in the meta yaml:
> 
> https://github.com/nf-core/modules/blob/a25abe62ccf3fe3767d2449d8322a35ab62844e0/modules/nf-core/sratools/prefetch/meta.yml#L44

I agree we should look at doing this systematically. How should we go about doing this?

ewels:
> How should we go about doing this?

I think @mirpedrol might have some magic in nf-core/tools. See recent work about updating the meta map, this is all autogenerated from the Nextflow code. Should be possible to do this as well I guess?

see the above issue to track this specific issue.

mirpedrol:
I didn't take this into account but I can add it. 
The `meta.yml` is generated  from `main.nf` information, so we could detect this `type: 'dir'` in the input channel and add `type: directory` to the meta.
In the opposite situation where we already documented the `meta.yml` but we didn't add the information to `main.nf` I would do it with a separate script for now, we won't need it in the long term if we start using the type in the `main.nf`. Sounds good?

PR #6675, opened 2024-09-20
Display message when error 132 found

JoseEspinosa:
As discussed on [this](https://github.com/TimoLassmann/kalign/issues/46) issue `kalign` throws an error 132 when run in some CPUs types. Here, we catch it and show a message so that users are aware of the issue until is fixed.

## PR checklist

  - For modules:


modules/nf-core/kalign/align/main.nf
```diff
@@ -23,6 +23,16 @@ process KALIGN_ALIGN {
     def prefix = task.ext.prefix ?: "${meta.id}"
     def write_output = compress ? ">(pigz -cp ${task.cpus} > ${prefix}.aln.gz)" : "${prefix}.aln"
     """
+    error_handler() {
```
mahesh-panchal, line 26: You're missing the call to this function.

mahesh-panchal: I also don't think this works. The trap needs call the function.

JoseEspinosa: yes, sorry I tested directly in the pipeline and miss to copy the line calling the function, now is fixed: https://github.com/nf-core/modules/pull/6675/commits/c38febf4fd44030fa7a549dc6226b46641c5b996
I tested it and it worked in the pipeline (after disabling the `ignore` in the `errorStrategy`)

mahesh-panchal (approved):
🚀

Issue #6505, opened 2024-08-30
Snapshot content of version files instead of a hash


edmundmiller:
Currently we're snapshoting the versions.yml and getting a hash:

https://github.com/nf-core/modules/blob/17b8bedf740a8403b548e49853246f1d9a24caa3/modules/nf-core/ale/tests/main.nf.test.snap#L26-L28

```diff
                "versions": [
-                    "versions.yml:md5,949da9c6297b613b50e24c421576f3f1"
+                    "content": ["{ALE={ale=20180904}}"],
                ]

**Comment by edmundmiller** [comment]
@nvnieuwk Pointed out that topic channels won't be hashable

#4517

**Comment by famosab** [comment]
How exactly would we assert for the versions then? I tried something like this:
```
{ assert snapshot(
             process.out,
             file(process.out.versions[0]).readLines()
             ).match()
}
```

edmundmiller:
> We might need to adapt the linting or snapshot both md5 and the content. Otherwise this error occurs:

Luckily, we make the rules of the linter so we can ignore any errors like that and fix it right after we settle on a standard!

I like `file(process.out.versions[0]).readLines()` but let me see what the snapshot looks like.

Ideally I'd love for it to look like

```diff
"sarscov2 [fasta_gz] - paired-end sorted bam": {
        "content": [
            {
                "0": [
                    [
                        {
                            "id": "test",
                            "single_end": false
                        },
                        "test_ALEoutput.txt:md5,4abcbd60ae1dbf78138c97e5fed97f3e"
                    ]
                ],
                "1": [
                    "versions.yml:md5,949da9c6297b613b50e24c421576f3f1"
                ],
                "ale": [
                    [
                        {
                            "id": "test",
                            "single_end": false
                        },
                        "test_ALEoutput.txt:md5,4abcbd60ae1dbf78138c97e5fed97f3e"
                    ]
                ],
-                "versions": [
-                    "versions.yml:md5,949da9c6297b613b50e24c421576f3f1"
-                ]
+                "versions": {
+                    "ale": "20180904"
+                }
            }
        ],
        "meta": {
            "nf-test": "0.8.4",
            "nextflow": "23.10.1"
        },
        "timestamp": "2024-03-19T09:06:19.589167"
    },

edmundmiller:
> Maybe there is a more elegant way?

I think we could write a lib function for this, either in this repo or the nf-test/utils plugin.

Two other things to consider:
[...]

PR #6704, opened 2024-09-27
chore(#6505): Capture actual version yaml instead of hash

edmundmiller:
Specifically for https://github.com/nf-core/website/pull/2740


edmundmiller:
It got the point across.

PR #6714, opened 2024-09-30
update vcfeval inputs + nf-test conversion

nvnieuwk:
The current input for golden truth BED files were incorrectly being used as the analysis BED file. This PR fixes this and adds some documentation to make this more clear


kubranarci (commented):
Only the name of the files have changed?

kubranarci (approved):
Seems like an appropriate fix! Thank you.

nvnieuwk:
And the `truth_bed` is now passed to `--evaluation-regions` and the other way around for the other BED file

PR #6727, opened 2024-10-03
Support nft-compress in nf-test.config

Jorisvansteenbrugge:
I am working on an nf-core module that outputs compressed files. In order to properly write tests for this module I need a way to validate the contents of these outputs. https://github.com/lukfor/nft-compress fits these needs and I therefore propose to add this plugin to the `nf-test.config` file.

## PR checklist


nvnieuwk (approved):
LGTM

PR #6753, opened 2024-10-08
Limma mixed models feature

KamilMaliszArdigen:
This extends limma functionality with option to process rna-seq data in pairwise and mixed mode. It will be used in differentialabundance pipeline and in future can be used in other pipelines which are using limma. 

## PR checklist

Closes #XXX 

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:


modules/nf-core/limma/differential/main.nf
```diff
@@ -4,17 +4,19 @@ process LIMMA_DIFFERENTIAL {
 
     conda "${moduleDir}/environment.yml"
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
-        'https://depot.galaxyproject.org/singularity/bioconductor-limma:3.54.0--r42hc0cfd56_0' :
-        'biocontainers/bioconductor-limma:3.54.0--r42hc0cfd56_0' }"
+        'oras://community.wave.seqera.io/library/bioconductor-edger_bioconductor-ihw_bioconductor-limma_r-dplyr_r-readr:7fc48564d286c1c6' :
```
maxulysse, line 7: we shouldn't have any oras protocol in the modules, it doesn't work with NXF_SINGULARITY_CACHEDIR

modules/nf-core/limma/differential/environment.yml
```diff
@@ -1,5 +1,10 @@
+name: limma_differential
```
maxulysse, line 1: We got rid of the names as we don't need them anymore

modules/nf-core/limma/differential/main.nf
```diff
@@ -4,17 +4,19 @@ process LIMMA_DIFFERENTIAL {
...
+        'community.wave.seqera.io/library/bioconductor-edger_bioconductor-ihw_bioconductor-limma_r-dplyr_r-readr:edea0f9fbaeba3a0' }"
```
maxulysse, line 8:
```suggestion
        'nf-core/bioconductor-edger_bioconductor-ihw_bioconductor-limma_r-dplyr_r-readr:edea0f9fbaeba3a0' }"
```
I pulled and pushed to quay.io

So we can change the registry from all container in a simpler way.
We'll be using `community.wave.seqera.io` as registry when we do the switch

pinin4fjords (changes requested):
There are a few changes I'd like to see here.

I understand the attraction of readr etc, but until we have full Seqera Containers integration with nf-core tooling (which shouldn't be long), extra packages create challenges. 

The forking of logic is also a bit ugly. If the logic is sufficiently different to require two scripts it should be two modules- though I'd favour integration if it didn't create excess complexity in the R code.

modules/nf-core/limma/differential/main.nf
```diff
@@ -23,5 +25,10 @@ process LIMMA_DIFFERENTIAL {
     task.ext.when == null || task.ext.when
 
     script:
-    template 'limma_de.R'
+    if (type == 'rnaseq') {
+        template 'limma_de_rnaseq.R'
```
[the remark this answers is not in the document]
KamilMaliszArdigen: How you envision "proper" integration? As single template? I will refactor that if needed.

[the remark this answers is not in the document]
pinin4fjords: I mean, if the two scripts share enough logic, they should be one with a simple conditional. If the logic is quite divergent such that doing that adds too much complexity, these should be separate modules.

modules/nf-core/limma/differential/main.nf
pinin4fjords: I'm not keen on this @maxulysse, there's no way for someone without privilege to do this and maintain the module.

modules/nf-core/limma/differential/main.nf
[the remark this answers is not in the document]
KamilMaliszArdigen: Well I had similar doubts in the one hand this is still limma and differential abundance analysis and on the other site this one is combined with voom and there is lot of differences. So maybe we will create new module named limma/differential-voom? or simply limma/voom - what are your thoughts on the module naming?

modules/nf-core/limma/differential/environment.yml
```diff
@@ -1,5 +1,9 @@
 channels:
-  - conda-forge
-  - bioconda
+- conda-forge
+- bioconda
 dependencies:
-  - bioconda::bioconductor-limma=3.54.0
+- bioconda::bioconductor-edger=4.0.16
+- bioconda::bioconductor-ihw=1.28.0
+- bioconda::bioconductor-limma=3.58.1
```
pinin4fjords, line 7: Limma is a [dependency of edger](https://github.com/bioconda/bioconda-recipes/blob/a489320df2777dfed69d979db18af47bcaa8a2d5/recipes/bioconductor-edger/meta.yaml#L24), you don't need to include it here

modules/nf-core/limma/differential/templates/limma_de_rnaseq.R
```diff
@@ -0,0 +1,411 @@
...
+dim(DGE)
+
+## Filter genes
+keep <- which(apply(DGE\$counts, 1, max) >= opt\$geneFilter)
+filtered_counts <- DGE\$counts[keep, ]
+DGE\$counts <- filtered_counts
+dim(DGE)
+
+# Calculate normalization factors
+DGE <- calcNormFactors(DGE)
```
pinin4fjords, line 222: Please correct me if I'm wrong, but this and the DGEList are the only things that need edgeR here, right? voom() will do the normalisation things, and will just take a matrix. Then we don't need the edgeR dependency.

modules/nf-core/limma/differential/templates/limma_de_rnaseq.R
```diff
@@ -0,0 +1,411 @@
...
+} else {
+    stop(
+        paste("Bad '--analysis_type' parameter. It can be either 'paired' or 'mixedmodel'. ")
+    )
+}
+
+## Generate outputs
+# Differential expression table
+
+if (opt\$IHW_correction) {
```
pinin4fjords, line 322: IMO if we're writing 'limma' modules they should be thin-as-possible wrappers around Limma functions. We shouldn't we embedding lots of additional logic, aside from the boilerplate required to pass inputs to those functions correctly and write the outputs.

pinin4fjords: Extensive customisations should probably be in local modules in the workflow, and we'd want to discuss to determine if this was part of standard best practice.

modules/nf-core/limma/differential/main.nf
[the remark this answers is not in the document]
pinin4fjords: I think this should be limma/differential-voom. 

As in the comments below, I also think this should be, as much as possible, a thin wrapper around Limma functions only. Otherwise it's just your custom script, rather than a 'standard' Limma module.

modules/nf-core/limma/differential/environment.yml
pinin4fjords: We only pin the primary dependency in the modules repo, unless you have a really compelling reason to do otherwise. We will at some point have lock files that pin all dependencies.

modules/nf-core/limma/differential/main.nf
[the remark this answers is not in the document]
pinin4fjords: (as in other comments, I'm not sure we need edgeR here, so we'd only pin Limma)

modules/nf-core/limma/differential/environment.yml
pinin4fjords: (as in other comments, I'm not sure we need edgeR here, so we'd only pin Limma)

modules/nf-core/limma/differential/templates/limma_de_rnaseq.R
KamilMaliszArdigen: We use edgeR dependency to read raw matrix count to create DGEList object which is one of required ways to provide expression data to voom: "counts - a numeric matrix containing raw counts, or an ExpressionSet containing raw counts, or a DGEList object." : https://www.rdocumentation.org/packages/limma/versions/3.28.14/topics/voom
 
 So we either need BioBase or edgeR.

modules/nf-core/limma/differential/environment.yml
KamilMaliszArdigen: Well we noticed that results are slightly different depending on limma versions so we decided to pin it to achieve reproducibility. I will adopt your comment in standalone module

modules/nf-core/limma/differential/templates/limma_de_rnaseq.R
pinin4fjords: Yep, as I say, we don't need edgeR, and that will simplify the software dependencies

modules/nf-core/limma/differential/templates/limma_de_rnaseq.R
```diff
@@ -0,0 +1,411 @@
...
+## Filter genes
+keep <- which(apply(DGE\$counts, 1, max) >= opt\$geneFilter)
+filtered_counts <- DGE\$counts[keep, ]
+DGE\$counts <- filtered_counts
+dim(DGE)
+
+# Calculate normalization factors
+DGE <- calcNormFactors(DGE)
+
+if (opt\$analysis_type == "pairwise") {
```
pinin4fjords, line 224:
```suggestion
if (opt\$analysis_type == "pairwise") {
```

I don't think you even need this extra mode. Sample pairing can be handled as a blocking variable, no? Even if it was needed, you're duplicating a lot of code between the two modes.

pinin4fjords (changes requested):
I've just had some time to review this again, and I think the R part also needs work. There's no need to treat the pairing variable in a special way, it can be a blocking variable like any other. At that point, the main difference between the two modes is the use of duplicateCorrelation, for which there could just be a separate flag.

Fundamentally, this would integrate much better into the existing differentialabundance logic if started from the existing limma module, and then replaced internals to run Voom instead, plus any improvements that are necessary.

modules/nf-core/limma/differential/environment.yml
pinin4fjords: Actually, I see your point on this one, apologies. Limma is the primary dependency so it's pinned. It's edgeR we probably don't need to bother pinning (though we need to include it).

pinin4fjords:
Just to re-state the above, and after refreshing myself on the Voom logic, I actually don't think Voom needs to be its own module. I think this can be integrated into the existing module. 

The voom-specific component is a matter of a few lines. All the other changes are not Voom specific.

modules/nf-core/limma/differential/main.nf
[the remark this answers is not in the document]
pinin4fjords: updating this older comment after newer ones below: I don't believe additional logic required for Voom merits its own module. We can add the Voom part in a conditional, and other changes (e.g. duplicateCorrelation) etc, apply equally to non-Voom Limma.

KamilMaliszArdigen:
> Just to re-state the above, and after refreshing myself on the Voom logic, I actually don't think Voom needs to be its own module. I think this can be integrated into the existing module.
> 
> The voom-specific component is a matter of a few lines. All the other changes are not Voom specific.

I missed your earlier comment I will tak a closer look on this one. At least thanks to new module we have nf-test almost ready :D

modules/nf-core/limma/differential/templates/limma_de_rnaseq.R
KamilMaliszArdigen: Well we are duplicating some code. Unfortunately the mixed model integration have large amount of small additions which resulted in multiple if's example can be voom which is called twice. I will try to incorporate voom into original script. I'm not yet fulli convinced about mixedmodel mode.

KamilMaliszArdigen:
@pinin4fjords I've extended original limma module with voom and IHW correction. This is still a draft as my results are now significantly different and sitll this have to be sorted out. I wanted to ask you if you fill that this is a right direction. Also any suggestion regarding incorporation of mixed models logic will be welcome :) 

Thank you in advance.

modules/nf-core/limma/differential/templates/limma_de.R
```diff
@@ -287,11 +295,38 @@ colnames(design) <- sub(
...
+
+    # Normalize counts using TMM
+    dge <- calcNormFactors(dge, method = "TMM")
+
+    # Run voom to transform the data
+    voom_result <- voom(dge, design)
+    data_for_fit <- voom_result
+
+    # Write the normalized counts matrix to a TSV file
+    normalized_counts <- voom_result\$E
```
pinin4fjords, line 311: If we're adding export of normalised values (which I think is a good thing) we should do the same for the microarry (e.g. with `normalizeBetweenArrays`), and do the writing outside this conditional.

pinin4fjords (commented):
Thank you for the ongoing work- this is definitely going the right way.

But I still think the IHW code doesn't belong in a module wrapping standard Limma functions.

modules/nf-core/limma/differential/templates/limma_de.R
```diff
@@ -322,18 +357,54 @@ ebayes_args = c(
 
 fit2 <- do.call(eBayes, ebayes_args)
 
-# Run topTable() to generate a results data frame
+if ((!is.null(opt\$use_voom) && opt\$use_voom) && (!is.null(opt\$IHW_correction) && opt\$IHW_correction)) {
```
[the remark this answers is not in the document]
KamilMaliszArdigen: Well I don't se a point to make a new module just for IHW correction and I'm not sure how we can handle integration of this logic into main pipeline at this point. Should I do sth like this one:
https://github.com/nf-core/differentialabundance/blob/master/modules/local/filter_difftable.nf?

modules/nf-core/limma/differential/templates/limma_de.R
KamilMaliszArdigen: Sure will try to add. - If I will have some problems I will let you know

modules/nf-core/limma/differential/templates/limma_de.R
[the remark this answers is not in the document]
pinin4fjords: Just do it without for now, and we'll evaluate if and how this should be included in the pipeline afterwards- feel free to make a separate workflow issue.

modules/nf-core/limma/differential/templates/limma_de.R
mzenczak: If I understand correctly, the assumption in limma for microarray analysis is that the input intensity table to this module has already been normalized. Therefore, we just need to write the normalized intensity table which was the input, and there's no need to apply `normalizeBetweenArrays`.

pinin4fjords: Oh yes, you're right actually- my bad, apologies @KamilMaliszArdigen, you can disregard this comment. 

See https://github.com/nf-core/differentialabundance/blob/3dd360fed0dca1780db1bdf5dce85e5258fa2253/workflows/differentialabundance.nf#L307.

KamilMaliszArdigen: That actually makes more sense :)

pinin4fjords: To be doubly clear, I take back the comment- no normalised table required for the array mode :-). That output will be optional.

modules/nf-core/limma/differential/templates/limma_de.R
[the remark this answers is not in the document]
KamilMaliszArdigen: Will do

KamilMaliszArdigen:
I finally find out why I had different results. I either don't understand how limma works or usage of blocking as is already implemented will not work as was mentioned earlier bellow the story of model and designs and how they change results: 

different order makes different design and different results:
model definition: `paste('~ 0 +', contrast_variable, '+ pair_IDs')` -> **this is what we end with current implementation**
model: `"~ 0 + treatment + pair_IDs"`
design:
```
           treatmenthND6 treatmentmCherry pair_IDs
SRX8042381             0                1        1
SRX8042382             0                1        2
SRX8042383             0                1        3
SRX8042384             1                0        1
SRX8042385             1                0        2
SRX8042386             1                0        3
```

model definition: `paste('~ 0 + pair_IDs +', contrast_variable)` -> **this is how paired model should be defined according to documentation**
model: `"~ 0 + pair_IDs + treatment"`
design:
```
           pair_IDs treatmenthND6 treatmentmCherry
SRX8042381        1             0                1
SRX8042382        2             0                1
SRX8042383        3             0                1
SRX8042384        1             1                0
SRX8042385        2             1                0
SRX8042386        3             1                0
```

Also it looks like there might be some unexpected situation if we have pair data as text and ints - the one with ints makes correct design one with string is missing one column in design bellow example :( 


model definition: `paste('~ 0 + pair_IDs +', contrast_variable)`
model: `"~ 0 + pair_IDs + treatment"`
design:
```
           pair_IDs treatmenthND6 treatmentmCherry
SRX8042381        1             0                1
SRX8042382        2             0                1
SRX8042383        3             0                1
SRX8042384        1             1                0
SRX8042385        2             1                0
SRX8042386        3             1                0
```

model definition: `paste('~ 0 + sample_number +', contrast_variable)`
model: `"~ 0 + sample_number + treatment"`
design:
```
           sample_numbersample1 sample_numbersample2 sample_numbersample3
SRX8042381                    1                    0                    0
SRX8042382                    0                    1                    0
SRX8042383                    0                    0                    1
SRX8042384                    1                    0                    0
SRX8042385                    0                    1                    0
SRX8042386                    0                    0                    1
           treatmentmCherry
SRX8042381                1
SRX8042382                1
SRX8042383                1
SRX8042384                0
SRX8042385                0
SRX8042386                0
```


I am thinking now that it might be worth to add option to override model via opt as each of above makes results significantly different. This will make module significantly more flexible. What do you think about that @pinin4fjords ?

pinin4fjords:
Could you explain how the results differ please? The fact that the design matrix has a different order shouldn't matter too much, as long as we construct the contrasts correctly.

pinin4fjords:
> The fact that the design matrix has a different order shouldn't matter too much, as long as we construct the contrasts correctly.

OK, that might not be true. But let's see if we can fix things without adding too much complexity. Will suggest soon...

pinin4fjords:
OK, so I've mucked about and put some changes in https://github.com/nf-core/modules/pull/6795 by way of illustration in isolation from your other changes. 

Basically, what I'd like is to pass pairing and blocking variables together, if it makes sense, to avoid having to re-engineer differentialabundance to too great a degree.

I _think_ that we can include pairing along with the other blocking factors, as long as those variables occur first in the model. So my suggestion is that we move them all, rather than having a new input for the pairing. 

I **don't** think allowing model specification is a good plan because the way the model is specified has ramifications for how the rest of the code is constructed, and it would be very easy to break.

KamilMaliszArdigen:
I agree with you making model a parameter can be dangerous for end users and result in unexpected results. That draft is great I will do that over the weekend. 

Can you comment on the design difference when blocking column contains str vs ints? - For us it looks like some kind of bug. Input file is `https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/genomics/mus_musculus/rnaseq_expression/SRP254919.samplesheet.csv`

Details bellow:


> Also it looks like there might be some unexpected situation if we have pair data as text and ints - the one with ints makes correct design one with string is missing one column in design bellow example :(
> 
> model definition: `paste('~ 0 + pair_IDs +', contrast_variable)` model: `"~ 0 + pair_IDs + treatment"` design:
> 
> ```
>            pair_IDs treatmenthND6 treatmentmCherry
> SRX8042381        1             0                1
> SRX8042382        2             0                1
> SRX8042383        3             0                1
> SRX8042384        1             1                0
> SRX8042385        2             1                0
> SRX8042386        3             1                0
> ```
> 
> model definition: `paste('~ 0 + sample_number +', contrast_variable)` model: `"~ 0 + sample_number + treatment"` design:
> 
> ```
>            sample_numbersample1 sample_numbersample2 sample_numbersample3
> SRX8042381                    1                    0                    0
> SRX8042382                    0                    1                    0
> SRX8042383                    0                    0                    1
> SRX8042384                    1                    0                    0
> SRX8042385                    0                    1                    0
> SRX8042386                    0                    0                    1
>            treatmentmCherry
> SRX8042381                1
> SRX8042382                1
> SRX8042383                1
> SRX8042384                0
> SRX8042385                0
> SRX8042386                0
> ```

I'm still not sure how we should approach paired model where we need to handle multiple variable references and targets... 

I guess this is a problem of future us.

pinin4fjords:
Strings or ints, they should be converted to factors- think I cover than in my draft, but we should make that happen. 

On your last point, assuming you mean multiple contrasts, that's handled in workflow logic.

KamilMaliszArdigen:
So I added your changes and it mostly works as expected now with one hiccup that blocking_variables have to be INT's in submission sheet. Despite that indeed we have factor cast we still end up with different model than expected and this causes failure in makeContrasts function (column not found in design). 

Below relevant opt values needed to reproduce the results and input data:
```
    input = 'https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/genomics/mus_musculus/rnaseq_expression/SRP254919.samplesheet.csv'
    matrix = 'https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/genomics/mus_musculus/rnaseq_expression/SRP254919.salmon.merged.gene_counts.top1000cov.tsv'
    transcript_length_matrix = 'https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/genomics/mus_musculus/rnaseq_expression/SRP254919.spoofed_lengths.tsv'
    contrasts = 'https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/genomics/mus_musculus/rnaseq_expression/SRP254919.contrasts.csv'
```

```
[1] "treatment_mCherry_hND6_sample_number"

$count_file
[1] "SRP254919.filtered.tsv"

$sample_file
[1] "SRP254919.samplesheet.sample_metadata.tsv"

$contrast_variable
[1] "treatment"

$reference_level
[1] "mCherry"

$target_level
[1] "hND6"

$blocking_variables
[1] "sample_number"

$probe_id_col
[1] "gene_id"

$sample_id_col
[1] "sample"
```

design which we get:
```
           sample_numbersample1 sample_numbersample2 sample_numbersample3
SRX8042381                    1                    0                    0
SRX8042382                    0                    1                    0
SRX8042383                    0                    0                    1
SRX8042384                    1                    0                    0
SRX8042385                    0                    1                    0
SRX8042386                    0                    0                    1
           treatmentmCherry
SRX8042381                1
SRX8042382                1
SRX8042383                1
SRX8042384                0
SRX8042385                0
SRX8042386                0
```
expected design: 
```
           sample_number treatment.hND6 treatment.mCherry
SRX8042381        sample1              0                 1
SRX8042382        sample2              0                 1
SRX8042383        sample3              0                 1
SRX8042384        sample1              1                 0
SRX8042385        sample2              1                 0
SRX8042386        sample3              1                 0
```

KamilMaliszArdigen:
I would say that we can live with that and we should simply add information that blocking data have to be provided as Int's? 

or we can do conversion to int's like we do in our version. 
```
    sample.sheet\$pair_IDs <- as.integer(as.factor(sample.sheet[[opt\$blocking_variables]]))
    model <- paste('~ 0 + pair_IDs +', contrast_variable)
```

Unless you have better idea @pinin4fjords.

pinin4fjords:
I think you're confused with the ints thing. The expansion of categorical variables across columns like that is normal, and we should not be casting things to integers- these are not ordinal variables, they are categories. The design matrix you're seeing is appropriate, it's just that we're missing the first level of the treatment variable because it's the 'reference' when we put the contrast variable last this way. 

We just have to do something like `TreatmentTreatment - 0` when we run makeContrasts() to account for that.

Here's a basic example I had ChatGPT work up for me by way of illustration:

```
# Load the limma package
library(limma)

# Step 1: Simulate sample metadata
# We have 6 samples, 3 blocks (e.g., subjects), and 2 treatments

# Create a data frame with sample metadata
sample_info <- data.frame(
  SampleID = paste0("Sample", 1:6),
  Block = rep(paste0("Block", 1:3), each = 2),
  Treatment = rep(c("Control", "Treatment"), times = 3)
)

# Display the sample metadata
print("Sample Metadata:")
print(sample_info)

# Step 2: Simulate expression data
# Simulate expression data for 100 genes
set.seed(123)  # For reproducibility
expression_data <- matrix(
  rnorm(100 * nrow(sample_info), mean = 10, sd = 2),
  nrow = 100,
  ncol = nrow(sample_info)
)
rownames(expression_data) <- paste0("Gene", 1:100)
colnames(expression_data) <- sample_info$SampleID

# Step 3: Prepare the sample_info data frame
# Convert variables to factors
sample_info$Block <- as.factor(sample_info$Block)
sample_info$Treatment <- as.factor(sample_info$Treatment)

# Step 4: Build the design matrix with blocking variable first
# Model formula with Block (blocking variable) first, then Treatment
# Using cell-means coding by including '0 +'
model_formula <- "~ 0 + Block + Treatment"

# Create the design matrix
design <- model.matrix(as.formula(model_formula), data = sample_info)

# Display the design matrix
print("Design Matrix:")
print(design)

# Step 5: Examine the design matrix
# Note that the first level of Treatment (Control) is missing
# This is because of collinearity between the intercept (absorbed by Block levels) and Treatment levels
# With '0 +' and including all Block levels, the model is saturated, and limma omits one level of Treatment
print("Column names of the design matrix:")
print(colnames(design))

# Step 6: Fit the linear model
# Fit the model using lmFit from limma
fit <- lmFit(expression_data, design)

# Step 7: Define contrasts
# Since 'TreatmentControl' is missing from the design matrix, we need to adjust our contrasts
# The coefficient 'TreatmentTreatment' represents the difference between Treatment and Control
# So, to test the effect of Treatment vs Control, we test if 'TreatmentTreatment' is significantly different from zero

# Create a contrast matrix
contrast_matrix <- makeContrasts(
  Treatment_vs_Control = TreatmentTreatment - 0,  # Testing if TreatmentTreatment coefficient is different from zero
  levels = design
)

# Display the contrast matrix
print("Contrast Matrix:")
print(contrast_matrix)

# Step 8: Apply contrasts and compute statistics
# Apply the contrasts to the fitted model
fit2 <- contrasts.fit(fit, contrast_matrix)
fit2 <- eBayes(fit2)

# Step 9: Extract the top differentially expressed genes
top_genes <- topTable(fit2, coef = "Treatment_vs_Control", number = 10)

# Display the top genes
print("Top Differentially Expressed Genes:")
print(top_genes)

# Step 10: Interpretation
# The coefficient 'TreatmentTreatment' represents the effect of Treatment compared to Control
# Since 'TreatmentControl' is not in the design matrix, 'TreatmentTreatment' is effectively 'TreatmentTreatment - TreatmentControl'

# Step 11: Conclusion
# Including the blocking variable first in the model formula leads to the first level of the Treatment variable being omitted from the design matrix
# This is expected behavior due to collinearity
# To test the effect of Treatment, we compare 'TreatmentTreatment' to zero using contrasts
```

modules/nf-core/limma/differential/templates/limma_de.R
```diff
@@ -255,84 +262,173 @@ if ((! is.null(opt\$exclude_samples_col)) && (! is.null(opt\$exclude_samples_val
...
+
+if ((treatment_target %in% colnames(design)) && (treatment_reference %in% colnames(design)))  {
+    contrast_string <- paste0(treatment_target, "-", treatment_reference)
+} else if (!(treatment_target %in% colnames(design)) && (treatment_reference %in% colnames(design))) {
+    contrast_string <- paste0(treatment_reference, "-0")
+} else if ((treatment_target %in% colnames(design)) && !(treatment_reference %in% colnames(design)))  {
+    contrast_string <- paste0(treatment_target, "-0")
+} else {
+    stop(paste0(treatment_target, " and ", treatment_reference, " Not found in design array"))
+}
```
KamilMaliszArdigen, line 382: So I've added this logic and cleaned up PR. 
What is left is to figure out the environment stuff.

KamilMaliszArdigen:
> I think you're confused with the ints thing. The expansion of categorical variables across columns like that is normal, and we should not be casting things to integers- these are not ordinal variables, they are categories. The design matrix you're seeing is appropriate, it's just that we're missing the first level of the treatment variable because it's the 'reference' when we put the contrast variable last this way.

Now I finally got it. So I had a bug in my original code...

modules/nf-core/limma/differential/environment.yml
```diff
@@ -1,5 +1,10 @@
+name: limma_differential
 channels:
-  - conda-forge
-  - bioconda
+- conda-forge
+- bioconda
 dependencies:
-  - bioconda::bioconductor-limma=3.54.0
+- bioconda::bioconductor-edger=4.0.16
```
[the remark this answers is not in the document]
KamilMaliszArdigen: So what I should do about this one?

modules/nf-core/limma/differential/templates/limma_de.R
```diff
@@ -255,84 +261,173 @@ if ((! is.null(opt\$exclude_samples_col)) && (! is.null(opt\$exclude_samples_val
...
```
pinin4fjords, line 381:
```suggestion
# Construct the expected column names for the target and reference levels in the design matrix
treatment_target <- paste0(contrast_variable, ".", opt\$target_level)
treatment_reference <- paste0(contrast_variable, ".", opt\$reference_level)

# Determine how to construct the contrast string based on which levels are present in the design matrix
if ((treatment_target %in% colnames(design)) && (treatment_reference %in% colnames(design))) {
    # Both target and reference levels are present in the design matrix
    # We can directly compare the two levels
    contrast_string <- paste0(treatment_target, "-", treatment_reference)
} else if (treatment_target %in% colnames(design)) {
    # Only the target level is present in the design matrix
    # The reference level may have been omitted due to collinearity or being set as the baseline
    # We compare the target level to zero (implicit reference)
    contrast_string <- paste0(treatment_target, "- 0")
} else if (treatment_reference %in% colnames(design)) {
    # Only the reference level is present in the design matrix
    # The target level may have been omitted from the design matrix
    # We compare zero (implicit target) to the reference level
    contrast_string <- paste0("0 - ", treatment_reference)
} else {
    # Neither level is present in the design matrix
    # This indicates an error; the specified levels are not found
    stop(paste0(treatment_target, " and ", treatment_reference, " not found in design matrix"))
}
```

Just simplifying logic and adding comments for future-us

pinin4fjords (commented):
OK, getting there. I'll comment something for the software - give me a sec

modules/nf-core/limma/differential/environment.yml
```diff
@@ -2,4 +2,8 @@ channels:
   - conda-forge
   - bioconda
 dependencies:
-  - bioconda::bioconductor-limma=3.54.0
+  - bioconda::bioconductor-edger=4.0.16
+  - bioconda::bioconductor-ihw=1.28.0
+  - bioconda::bioconductor-limma=3.58.1
+  - conda-forge::r-dplyr=1.1.4
+  - conda-forge::r-readr=2.1.5
```
pinin4fjords, line 9:
```suggestion
  - bioconda::bioconductor-edger=4.0.16
  - bioconda::bioconductor-limma=3.58.1
```

pinin4fjords (commented):
Think these two deps should do it now, and I think we can use Seqera containers as long as we use the HTTPS for singularity.

modules/nf-core/limma/differential/main.nf
```diff
@@ -4,8 +4,8 @@ process LIMMA_DIFFERENTIAL {
 
     conda "${moduleDir}/environment.yml"
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
-        'https://depot.galaxyproject.org/singularity/bioconductor-limma:3.54.0--r42hc0cfd56_0' :
-        'biocontainers/bioconductor-limma:3.54.0--r42hc0cfd56_0' }"
+        'oras://community.wave.seqera.io/library/bioconductor-edger_bioconductor-ihw_bioconductor-limma_r-dplyr_r-readr:7fc48564d286c1c6' :
+        'community.wave.seqera.io/library/bioconductor-edger_bioconductor-ihw_bioconductor-limma_r-dplyr_r-readr:edea0f9fbaeba3a0' }"
```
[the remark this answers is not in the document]
KamilMaliszArdigen: I didn't know how to convert oras to https. Thank you

KamilMaliszArdigen:
Okej so last remaining is conda test. I have no clue how I should handle the error from CI. Can you point me @pinin4fjords  in the right direction?

pinin4fjords:
> Okej so last remaining is conda test. I have no clue how I should handle the error from CI. Can you point me @pinin4fjords in the right direction?

Unfortunately, pytest with conda seems broken like this right now. I think we might need to bit the bullet and port the testing over to nf-test.

maxulysse:
> > Okej so last remaining is conda test. I have no clue how I should handle the error from CI. Can you point me @pinin4fjords in the right direction?
> 
> Unfortunately, pytest with conda seems broken like this right now. I think we might need to bit the bullet and port the testing over to nf-test.

I you can, that'll be amazing

KamilMaliszArdigen:
> > > Okej so last remaining is conda test. I have no clue how I should handle the error from CI. Can you point me @pinin4fjords in the right direction?
> > 
> > 
> > Unfortunately, pytest with conda seems broken like this right now. I think we might need to bit the bullet and port the testing over to nf-test.
> 
> I you can, that'll be amazing

Done, I used different input data to avoid additional setup. Tested logic is the same. 
Can you review @pinin4fjords, @maxulysse?

pinin4fjords (changes requested):
Sorry, we need to replicate the old tests, with the same data, as well as anything new. By using different data you risk a regression, that it would not work with the old data.

If you don't have time, I'll try to get to it at some point, amongst my other summit prep.

KamilMaliszArdigen:
I will try. However honestly I can't imagine the case where this could happen. We test the same options (what is covered in updated tests) but with simpler approach. Old tests have been working with new script (with except for conda what was not related to the script). I would say that it is better. 

If you still think that old approach to tests is must let me know. 
I will commit draft with the results of my work later today.

pinin4fjords:
It's just not appropriate to change test data at the same time as the testing framework, sorry. 

We need to retain the original tests,  re-formulated into nf-test, and then add your new ones for Voom on top.

KamilMaliszArdigen:
> It's just not appropriate to change test data at the same time as the testing framework, sorry. Also we need to test the array and rnaseq functionalities separately, with appropriate input data.
> 
> We need to retain the original tests, re-formulated into nf-test, and then add your new ones for Voom on top.

I've modified tests as requested. we have exactly equal hashes and comparisons as in pytest. I hope that it will be all okey.

KamilMaliszArdigen:
@pinin4fjords I know that you are for sure super busy with the summit. I would really appreciate your input. I would like to move forward with the pipeline integration.​

modules/nf-core/limma/differential/tests/main.nf.test
```diff
@@ -0,0 +1,284 @@
...
+
+    tag "modules"
+    tag "modules_nfcore"
+    tag "limma"
+    tag "limma/differential"
+    tag "affy"
+    tag "affy/justrma"
+    tag "untar"
+
+
```
pinin4fjords, line 15:
```suggestion
```

modules/nf-core/limma/differential/tests/main.nf.test
```diff
@@ -0,0 +1,284 @@
...
+        config "./nextflow.voom_blocking.config"
+
+        when {
+            process {
+                """
+                input[0] = Channel.of(['id': 'test', 'variable': 'treatment', 'reference': 'hND6', 'target': 'mCherry'])
+                        .map{tuple(it, it.variable, it.reference, it.target)}
+                input[1] = Channel.of([[id:'test'],
+                                      file(params.modules_testdata_base_path + 'genomics/mus_musculus/rnaseq_expression/SRP254919.samplesheet.csv', checkIfExists: true),
+                                      file(params.modules_testdata_base_path + 'genomics/mus_musculus/rnaseq_expression/SRP254919.salmon.merged.gene_counts.top1000cov.tsv', checkIfExists: true)])
```
pinin4fjords, line 215:
```suggestion
                    file(params.modules_testdata_base_path + 'genomics/mus_musculus/rnaseq_expression/SRP254919.samplesheet.csv', checkIfExists: true),
                    file(params.modules_testdata_base_path + 'genomics/mus_musculus/rnaseq_expression/SRP254919.salmon.merged.gene_counts.top1000cov.tsv', checkIfExists: true)])
```

modules/nf-core/limma/differential/tests/main.nf.test
```diff
@@ -0,0 +1,284 @@
...
+        config "./nextflow.voom.config"
+
+        when {
+            process {
+                """
+                input[0] = Channel.of(['id': 'test', 'variable': 'treatment', 'reference': 'hND6', 'target': 'mCherry'])
+                        .map{tuple(it, it.variable, it.reference, it.target)}
+                input[1] = Channel.of([[id:'test'],
+                                      file(params.modules_testdata_base_path + 'genomics/mus_musculus/rnaseq_expression/SRP254919.samplesheet.csv', checkIfExists: true),
+                                      file(params.modules_testdata_base_path + 'genomics/mus_musculus/rnaseq_expression/SRP254919.salmon.merged.gene_counts.top1000cov.tsv', checkIfExists: true)])
```
pinin4fjords, line 244:
```suggestion
                    file(params.modules_testdata_base_path + 'genomics/mus_musculus/rnaseq_expression/SRP254919.samplesheet.csv', checkIfExists: true),
                    file(params.modules_testdata_base_path + 'genomics/mus_musculus/rnaseq_expression/SRP254919.salmon.merged.gene_counts.top1000cov.tsv', checkIfExists: true)])
```

modules/nf-core/limma/differential/tests/main.nf.test
```diff
@@ -0,0 +1,284 @@
...
+
+        options "-stub"
+
+        when {
+            process {                """
+                input[0] = Channel.of(['id': 'test', 'variable': 'treatment', 'reference': 'hND6', 'target': 'mCherry'])
+                        .map{tuple(it, it.variable, it.reference, it.target)}
+                input[1] = Channel.of([[id:'test'],
+                                      file(params.modules_testdata_base_path + 'genomics/mus_musculus/rnaseq_expression/SRP254919.samplesheet.csv', checkIfExists: true),
+                                      file(params.modules_testdata_base_path + 'genomics/mus_musculus/rnaseq_expression/SRP254919.salmon.merged.gene_counts.top1000cov.tsv', checkIfExists: true)])
```
pinin4fjords, line 272:
```suggestion
                    file(params.modules_testdata_base_path + 'genomics/mus_musculus/rnaseq_expression/SRP254919.samplesheet.csv', checkIfExists: true),
                    file(params.modules_testdata_base_path + 'genomics/mus_musculus/rnaseq_expression/SRP254919.salmon.merged.gene_counts.top1000cov.tsv', checkIfExists: true)])
```

modules/nf-core/limma/differential/tests/main.nf.test
```diff
@@ -0,0 +1,284 @@
...
+        config "./nextflow.config"
+
+        setup {
+            run("UNTAR") {
+                script "../../../untar/main.nf"
+                process {
+                    """
+                    input[0] = [
+                        [id: 'test'],
+                        file(params.test_data['homo_sapiens']['genome']['affy_array_celfiles_tar'], checkIfExists: true)
```
pinin4fjords, line 28: You need to convert these to the newer syntax you use below for consistency, using `params.modules_testdata_base_path`. You can find the map here: https://github.com/nf-core/modules/blob/master/tests/config/test_data.config

pinin4fjords (changes requested):
Thanks for that- almost there. You just need to convert the old style test data specification the newer one

modules/nf-core/limma/differential/tests/main.nf.test
KamilMaliszArdigen: Done :)

KamilMaliszArdigen:
I made requested changes. One test had a timeout so I restarted it.

KamilMaliszArdigen:
So I will merge and start updating main pipeline. 
Thank you once again @pinin4fjords and @maxulysse for your fast responses :)

PR #6768, opened 2024-10-10
Fixed linked site location for bbam tools

J-Moravec:
Fixed linked site location for bbam tools.

Some inconsistencies that could be fixed as well:

* most tools are pointing at general bbam documentation, some are however pointing at specific tool documentation (e.g., `clumpify`
* sometimes, biostar post is linked instead (e.g., in `filterbyname`
* sometimes the tool-specific documentation is linked using `tool_dev_url` instead (e.g., `bbnorm`)
* inconsistency in quoted or unquoted links

### Potential issue: Linting

Couldn't find instruction on number of symbols per line or whether breaking URL is allowed


maxulysse (approved):

maxulysse (approved):

J-Moravec:
The failing tests are caused by `bash: line 1: micromamba: command not found`.

But this issue was not caused by the changes and thus this branch will never get automered.

I also have no clue why is it the case that micromamba is failing. Some random PR that I checked had skipped the pytests.

famosab:
Seems like the pytest for bbmap/align is failing for both conda and singularity. Given that we want to swap to nf-test anyway: Could you maybe try and port `bbmap/align` to nf-test?

famosab:
@maxulysse I would suggest to just merge this as it only updates meta.yml files :) and I'll open an issue that bbmap/align needs to be ported to nf-test!

PR #6778, opened 2024-10-14
Fix tar to always follow symlinks due to nxf staging in work dirs

jfy133:
Forget this bit :sweat_smile:


mahesh-panchal (approved):
Good point. Does that mean all files are in the archive resolve to their full files when decompressed?

PR #6780, opened 2024-10-14
rm fasta from bwameth index folder to be handled separately

sateeshperi:


modules/nf-core/bwameth/align/main.nf
```diff
@@ -24,17 +25,16 @@ process BWAMETH_ALIGN {
     def prefix = task.ext.prefix ?: "${meta.id}"
     def read_group = meta.read_group ? "-R ${meta.read_group}" : ""
     """
-    INDEX=`find -L ${index} -name "*.bwameth.c2t" | sed 's/\\.bwameth.c2t\$//'`
-
     # Modify the timestamps so that bwameth doesn't complain about building the index
     # See https://github.com/nf-core/methylseq/pull/217
-    touch -c -- *
+    touch -c $index/*
+    mv $fasta $index/
```
maxulysse, line 31: just wondering if mv is the best solution here, but good enough for me

maxulysse (approved):

modules/nf-core/bwameth/align/main.nf
sateeshperi: ya im also not sure but, will see if any problems arise due to it.

PR #6805, opened 2024-10-18
Update Version of UMICollapse to 1.1.0

MatthiasZepper:
In this PR, the version of UMICollapse is bumped to the latest version 1.1.0. Since the original tool author was unresponsive, [the Bioconda version is now built](https://github.com/bioconda/bioconda-recipes/pull/51441) from [a fork](https://github.com/siddharthab/UMICollapse) maintained by Siddhartha Bagaria, who also contributed [the important bug fix](https://github.com/Daniel-Liu-c0deb0t/UMICollapse/pull/32) that stimulated the new release. 

## PR checklist


modules/nf-core/umicollapse/meta.yml
```diff
@@ -11,7 +11,7 @@ tools:
         (UMIs)/Random Molecular Tags (RMTs)."
       homepage: "https://github.com/Daniel-Liu-c0deb0t/UMICollapse"
       documentation: "https://github.com/Daniel-Liu-c0deb0t/UMICollapse"
-      tool_dev_url: "https://github.com/Daniel-Liu-c0deb0t/UMICollapse"
+      tool_dev_url: "https://github.com/siddharthab/UMICollapse"
```
siddharthab, line 14: I think we should keep this to still point to the original one for now. If we don't get a response for a year or so, then we can take over permanently.

MatthiasZepper: I prefer to leave it, though. I don't think that changing the URL gives the impression of taking development out of Daniel's hands, yet clarifies what changes constitute the version 1.1.0. After all, there are relevant improvements that are not yet reflected in the original repository - if we don't point to your for fork, I think it could be (even more) confusing.

siddharthab: Makes sense.

mashehu (approved):

PR #6818, opened 2024-10-22
pbsv

tanyasarkjain:
## PR checklist

Closes #XXX 

- [] Remove all TODO statements.
- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


tanyasarkjain:
@nf-core-bot fix linting

fellen31:
Should this perhaps be two modules, pbsv/discover and pbsv/call? Especially if pbsv discover is single-threaded while psbv call can use multiple threads. Do you have a sense for how long the discover step takes on real data?

tanyasarkjain:
I think typically discover and call are used together - I can't think of a scenario where you would just want to discover but not call

modules/nf-core/pbsv/tests/main.nf.test
```diff
@@ -0,0 +1,68 @@
...
+                ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out.versions).match("versions") },
+                { assert process.out.vcf != null },
```
sateeshperi, line 32: can you try out this [nft-vcf plugin](https://github.com/seppinho/nft-vcf) and see if you can make better assertions for vcf files with it

fellen31:
> I think typically discover and call are used together - I can't think of a scenario where you would just want to discover but not call, and for calling with pbsv you must discover first, so I think it makes sense to keep them together, both are needed for structural variant calling

To align reads with pbmm2 you first need to run `pbmm2 index` to create an index, and then `pbmm2 align` to align the reads. They should still be made as separate modules (same as [bwamem2](https://github.com/nf-core/modules/tree/master/modules/nf-core/bwamem2) or [minimap2](https://github.com/nf-core/modules/tree/master/modules/nf-core/minimap2)). Partly because indexing uses a few cores for a couple of minutes, while aligning can take several hours. The align command might fail, while indexing worked, and by having separate modules you could resume the pipeline without having to re-index. 

The [cnvpytor](https://github.com/nf-core/modules/tree/master/modules/nf-core/cnvpytor) modules are another example of where you would probably never run the subcommands alone, but they are still added as separate modules, and can still be run one after the other in a workflow.

tanyasarkjain:
Okay - I can work on making it into 2 modules

modules/nf-core/pbsv/tests/main.nf.test
tanyasarkjain: - I split the pbsv module now into two modules (discover and call) - so changed this PR to be for the discover functionality specifically, where svsig.gz files are produced - and so the test no longer checks for a VCF

modules/nf-core/pbsv/discover/main.nf
```diff
@@ -0,0 +1,44 @@
...
+        'https://depot.galaxyproject.org/singularity/pbsv:2.9.0--h9ee0642_0':
+        'biocontainers/pbsv:2.9.0--h9ee0642_0' }"
+
+    input:
+    tuple val(meta), path(bam)
+    tuple val(meta2), path(fasta)
+
+    output:
+    tuple val(meta), path("*.svsig.gz"), emit: svsig
+    path "versions.yml"           , emit: versions
```
fellen31, line 15: Could you align the emits?

modules/nf-core/pbsv/discover/main.nf
```diff
@@ -0,0 +1,44 @@
...
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    pbsv discover ${bam} ${prefix}.pbsv.svsig.gz
```
fellen31, line 24: Needs `$args`, and think I'd remove `pbsv` from the output filename, the user can add it using prefix if they like.  
```suggestion
    pbsv \\
        discover \\
        $args \\
        ${bam} \\
        ${prefix}.svsig.gz
```

modules/nf-core/pbsv/discover/main.nf
```diff
@@ -0,0 +1,44 @@
...
+    END_VERSIONS
+    """
+
+    stub:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    echo "test" > test.txt
+    gzip test.txt
+    mv test.txt.gz ${prefix}.pbsv.svsig.gz
```
fellen31, line 37: This should do it I think

```suggestion
    echo | gzip > ${prefix}.svsig.gz
```

modules/nf-core/pbsv/discover/meta.yml
```diff
@@ -0,0 +1,54 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/meta-schema.json
+name: "pbsv_discover"
+description: "pbsv - PacBio structural variant (SV) signature discovery tool"
+keywords:
+  - sort
+  - example
+  - genomics
```
fellen31, line 7: Keywords needs updating to match the module.

modules/nf-core/pbsv/discover/meta.yml
```diff
@@ -0,0 +1,54 @@
...
+          e.g. `[ id:'sample1', single_end:false ]`
+    - bam:
+        type: file
+        description: Sorted BAM/CRAM/SAM file
+        pattern: "*.{bam,cram,sam}"
+  - - meta2:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. `[ id:'sample1', single_end:false ]`
```
fellen31, line 30:
```suggestion
          Groovy Map containing reference information
          e.g. `[ id:'reference' ]`
```

modules/nf-core/pbsv/discover/tests/main.nf.test
```diff
@@ -0,0 +1,66 @@
...
+                ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out.versions).match("versions") },
+                { assert snapshot(process.out.svsig).match("svsig")}
```
fellen31, line 33: Just this should be fine as well:
```suggestion
                { assert snapshot(process.out).match() }
```

fellen31 (commented):
Nice! A few comments, but thanks for making the effort to split this into two modules! ⭐

tanyasarkjain:
> Nice! A few comments, but thanks for making the effort to split this into two modules! ⭐

Thanks for the review!! - I believe I have resolved all your comments now

PR #6823, opened 2024-10-22
pypgx/runngspipeline

### LINKED ISSUE #5987: new module: pypgx/runngspipeline

Jorisvansteenbrugge:
## PR checklist

Closes #5987
This PR adds a new module: pypgx/runngspipeline

- [x] Add a resource `label`
  - For modules:


famosab:
Can you run the test locally with the conda profile?
`nf-core modules test pypgx/runngspipeline --profile conda -u

Jorisvansteenbrugge:
> Can you run the test locally with the conda profile? `nf-core modules test pypgx/runngspipeline --profile conda -u`

@famosab Thank you for taking a look! I get the same error unfortunately.. I wonder if this is the same issue as #4234 opened by @edmundmiller . 
> I see this error appears when running tests for subworkflows. Could be that we need to add a name to environment.yml? If two modules in a subworkflow have exactly the same environment.yml I suspect the environment names may clash

While this is not a subworkflow, the nf-test for this module is chained with other modules that use the exact same `environment.yml`. I am however not sure what the solution is.

modules/nf-core/pypgx/runngspipeline/main.nf
```diff
@@ -0,0 +1,58 @@
...
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/pypgx:0.25.0--pyh7e72e81_0':
+        'biocontainers/pypgx:0.25.0--pyh7e72e81_0' }"
+
+    input:
+    tuple val(meta), path(vcf), path(tbi), path(coverage), path(control_stats), val(pgx_gene)
+    tuple val(meta2), path(resource_bundle)
+    val(assembly_version)
```
LouisLeNezet, line 13: Is this argument mandatory ?

modules/nf-core/pypgx/runngspipeline/main.nf
```diff
@@ -0,0 +1,58 @@
...
+    """
+    export MPLCONFIGDIR="/tmp/"
+    export PYPGX_BUNDLE=${resource_bundle}/
+
+    pypgx run-ngs-pipeline \\
+        --assembly ${assembly} \\
+        ${pgx_gene} \\
+        ${prefix}_pypgx_output/ \\
+        --variants ${vcf} \\
+        --depth-of-coverage ${coverage} \\
```
LouisLeNezet, line 35: It seems that the `depth-of-coverage` and the `control-statistics` are not mandatory.
Could you add a check in the script to output the command only if the path is provided ?
Such as

```
def depth_coverage_cmd = coverage ? "--depth-of-coverage ${coverage}" : ""
```

modules/nf-core/pypgx/runngspipeline/meta.yml
```diff
@@ -0,0 +1,77 @@
...
+input:
+  - - meta:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. `[ id:'sample1', single_end:false ]`.
+    - vcf:
+        type: file
+        description: BGZIP compressed VCF file with SNVs/indels. Output of pypgx/createinputvcf.
+        pattern: "*.{vcf.gz}"
```
LouisLeNezet, line 28: Can it not be `.bcf` files ?

modules/nf-core/pypgx/runngspipeline/meta.yml
```diff
@@ -0,0 +1,77 @@
...
+    - tbi:
+        type: file
+        description: VCF tabix index.
+        pattern: "*.{vcf.gz.tbi}"
```
LouisLeNezet, line 32: Can it be `.csi` ?

modules/nf-core/pypgx/runngspipeline/meta.yml
```diff
@@ -0,0 +1,77 @@
...
+    - coverage:
+        type: file
+        description: ZIP compressed file with depth of coverage information. Output
```
LouisLeNezet, line 35: Could you add that they are optional ?

modules/nf-core/pypgx/runngspipeline/meta.yml
```diff
@@ -0,0 +1,77 @@
...
+          of pypgx/preparedepthofcoverage.
+        pattern: "*.{zip}"
+    - control_stats:
+        type: file
+        description: ZIP compressed file with control statistics. Output of pypgx/computecontrolstatistics.
+    - pgx_gene:
+        type: string
+        description: Pharmacogene to genotype/phenotype.
```
LouisLeNezet, line 43: Could you expand a bit the description ?
What does the gene name should correspond to, is there a list of them from which to chose ?

modules/nf-core/pypgx/runngspipeline/main.nf
```diff
@@ -0,0 +1,58 @@
...
+        'https://depot.galaxyproject.org/singularity/pypgx:0.25.0--pyh7e72e81_0':
+        'biocontainers/pypgx:0.25.0--pyh7e72e81_0' }"
+
+    input:
+    tuple val(meta), path(vcf), path(tbi), path(coverage), path(control_stats), val(pgx_gene)
+    tuple val(meta2), path(resource_bundle)
+    val(assembly_version)
+
+    output:
+    tuple val(meta), val(pgx_gene), path("*pypgx_output"), emit: outdir
```
LouisLeNezet, line 16: I don't think that the `val(pgx_gene)` should be here.
What is the rational behind this choice ?

modules/nf-core/pypgx/runngspipeline/meta.yml
```diff
@@ -0,0 +1,77 @@
...
+  - outdir:
+      - meta:
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. `[ id:'sample1', single_end:false ]`
+      - pgx_gene:
+          type: string
+          description: Pharmacogene used to genotype/phenotype
+      - "*pypgx_output":
```
LouisLeNezet, line 65: Could you provide a list of what to expect from the output directory ?
A more nf-core way of writing this, would be to list in different channel the different output file depending on their extension.

modules/nf-core/pypgx/runngspipeline/tests/main.nf.test
```diff
@@ -0,0 +1,125 @@
...
+            def zipFile = path(outDir+"/results.zip").zip
+            def paths = zipFile.extractAll()
+            def metaFile = paths.find { it.getFileName().toString() == 'metadata.txt' }
+
+            assertAll(
+                { assert process.success },
+                { assert zipFile.isValid() },
+                { assert paths.size() == 2 },
+                { assert snapshot(
+                    metaFile,
```
LouisLeNezet, line 100: Is there only a `results.zip` inside the output ?
Could you snapshot all the files names present inside the zip file ?

LouisLeNezet (changes requested):
Thanks for the contribution.
I'm just a bit concerned on how the output is given back.
Is there a need for it to be in a zip format ?

modules/nf-core/pypgx/runngspipeline/main.nf
Jorisvansteenbrugge: Hi @LouisLeNezet Thank you for your thorough review, and sorry for replying late! Which argument are you referring to exactly?

modules/nf-core/pypgx/runngspipeline/meta.yml
Jorisvansteenbrugge: The variants are calculated in another step of the same program [pypgx/createinputvcf](https://github.com/nf-core/modules/blob/master/modules/nf-core/pypgx/createinputvcf/main.nf), which happens to be VCF format. The documentation also only mentions VCF files. I did not test if the tool works with BCF but it might not be relevant since the recommended approach is to use the creatinputvcf output.  What do you think?

modules/nf-core/pypgx/runngspipeline/main.nf
Jorisvansteenbrugge: Ah yes you are most likely right.  I added that so I could easily group the outputs (on pgx_gene) of this process later in my own pipeline, but it might not be nf-core style. If you agree then I will remove it.

modules/nf-core/pypgx/runngspipeline/tests/main.nf.test
Jorisvansteenbrugge: There are multiple output files (mostly ZIP) in the output directory. But `results.zip` is the main output, which always contains only `metadata.txt` and `data.tsv`. If those snapshots match then the output of the pipeline is the same between runs as well.

modules/nf-core/pypgx/runngspipeline/main.nf
LouisLeNezet: I was talking about the `assembly_version`

modules/nf-core/pypgx/runngspipeline/meta.yml
LouisLeNezet: That does make sense to me, I was just thinking if the tool was able to produce something else.
You can keep it like that 😉

modules/nf-core/pypgx/runngspipeline/meta.yml
LouisLeNezet: Okay, fine for me

modules/nf-core/pypgx/runngspipeline/main.nf
LouisLeNezet: Even in your pipeline, the recommended approach would be to add this parameter through the config file of the module.
This way the user can clearly see what's is added to each module.

modules/nf-core/pypgx/runngspipeline/tests/main.nf.test
LouisLeNezet: Okay, this is fine for me !

modules/nf-core/pypgx/runngspipeline/main.nf
Jorisvansteenbrugge: I have set a default in the process as
`def assembly = assembly_version ?: "GRCh38"` (https://github.com/nf-core/modules/blob/pypgx/runngspipeline/modules/nf-core/pypgx/runngspipeline/main.nf#L26)

This is in line with how I pass the assembly_version value channel to other processes of the same tool (which are already in master):
- https://github.com/nf-core/modules/blob/master/modules/nf-core/pypgx/createinputvcf/main.nf#L28
- https://github.com/nf-core/modules/blob/master/modules/nf-core/pypgx/computecontrolstatistics/main.nf#L26
- https://github.com/nf-core/modules/blob/master/modules/nf-core/pypgx/preparedepthofcoverage/main.nf#L26

Do you propose something else?

Jorisvansteenbrugge:
> Thanks for the contribution. I'm just a bit concerned on how the output is given back. Is there a need for it to be in a zip format ?

The zip files are simply what the tool outputs, it is not something I added myself

modules/nf-core/pypgx/runngspipeline/main.nf
LouisLeNezet: I'm not sure that `def assembly = assembly_version ?: "GRCh38"` is the best thing to do.
I was just wondering if the tool needed this string and if so you could just use: `assert assembly : "Assembly name should be provided"`. From my point of view the specie should not be set as default, but if your previous PR was accepted as such this is fine for me.

LouisLeNezet:
> > Thanks for the contribution. I'm just a bit concerned on how the output is given back. Is there a need for it to be in a zip format ?
> 
> @LouisLeNezet The zip files are simply what the tool outputs, it is not something I added myself

That's fine for me then.

LouisLeNezet (approved):
From looking at the other modules of this tool, this is ready to merge !
LGTM

modules/nf-core/pypgx/runngspipeline/main.nf
Jorisvansteenbrugge: I think I understand what you mean. I guess in hindsight it sounds indeed be better to retrieve the assembly_version from the config file instead. I will keep it in mind, perhaps I can do an update of all pypgx modules at a later time to streamline this.

LouisLeNezet: No worry, but this would indeed be a better approach as all modules try to be agnostic to the specie studied 😉

Issue #6877, opened 2024-10-28
pigz uncompress missing backslash


ljwharbers:
### Have you checked the docs?

- [x] [nf-core website: troubleshooting](https://nf-co.re/usage/troubleshooting)
- [x] [nf-core modules documentation](https://nf-co.re/docs/contributing/modules)

### Description of the bug

Missing backslash in the stub of pigz/uncompress module results in the following error/warning:
`unknown recognition error type: groovyjarjarantlr4.v4.runtime.LexerNoViableAltException`

Current:
```
    stub:
    def args = task.ext.args ?: ''
    uncompressed_filename = zip.toString() - '.gz'
    """
    touch ${zip.dropRight(3)}

    cat <<-END_VERSIONS > versions.yml
    "${task.process}":
        pigz: \$(echo \$(pigz --version 2>&1) | sed 's/^.*pigz\w*//' ))
    END_VERSIONS
    """
```

Should be:
```
    stub:
    def args = task.ext.args ?: ''
    uncompressed_filename = zip.toString() - '.gz'
    """
    touch ${zip.dropRight(3)}

    cat <<-END_VERSIONS > versions.yml
    "${task.process}":
        pigz: \$(echo \$(pigz --version 2>&1) | sed 's/^.*pigz\\w*//' ))
    END_VERSIONS
    """
```

PR #6878, opened 2024-10-28
Change sed regex for version reporting in pigz modules

ljwharbers:
Missing backslash in pigz uncompress stub results in `unknown recognition error type: groovyjarjarantlr4.v4.runtime.LexerNoViableAltException`

Simplified sed regex to circumvent the missing/added trailing backslash

## PR checklist

Closes [#6877 ](https://github.com/nf-core/modules/issues/6877)

- [ x ] This comment contains a description of changes (with reason).


fellen31 (commented):
Could you add a stub test to this module? 

Also, make sure to ask to be added to the nf-core GitHub organisation so you can run the GitHub actions :)

ljwharbers:
@fellen31 Thanks for the reply. 
Diving a bit deeper into the sed logic I believe that it wasn't giving the expected result. 

I'm still new to contributing to nf-core, but I'm more than happy to help with adding tests. Could you either point me to some examples of how to add tests for this?

fellen31:
> I'm still new to contributing to nf-core, but I'm more than happy to help with adding tests. Could you either point me to some examples of how to add tests for this?

Sure. You would be adding another test identical the current one but with `options "-stub"` (also add stub to the name of the test). Afterwards run e.g. `nf-test test modules/nf-core/pigz/uncompress/ --profile docker --update-snapshot` to generate a new snapshot. This will test the stub part of the module, and should catch errors like the one you reported.

For a simple example: https://github.com/nf-core/modules/blob/3abde3de0970781e8e9aeec3cfeed759008a6ef4/modules/nf-core/unzip/tests/main.nf.test

modules/nf-core/pigz/uncompress/tests/main.nf.test
```diff
@@ -29,5 +29,34 @@ nextflow_process {
...
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(
+                    file(process.out.file[0][1]).name,
+                    process.out.versions
+                    ).match()
+                }
```
fellen31, line 57:
```suggestion
                { assert snapshot(process.out).match() }
```

fellen31:
No you don't need to fix the others. When you update this branch with the latest master, the other test should not be checked.

There was a problem with GitHub Actions earlier today, maybe something is still not working correctly.

fellen31:
Or actually, in this case it's because `PIGZ_COMPRESS` is being used in toffee/irmsd tests. It looks like it should be a simple fix (updating meta.yml to match main.nf), so perhaps you could try to fix in this PR.

fellen31 (dismissed):
Thanks for adding the stub test, looks good to me! 👍

fellen31:
> Looking into tcoffee/irmsd, it needs the added types and descriptions in the yaml file. However, while I can 'guess' the types based on the nextflow code, I have no idea what the descriptions should be.
> 
> So perhaps this is better that someone adds it who actually uses this.

Perhaps guessing a bit is ok here, or maybe @luisas could help. I think the linting has been updated to include more checks since that module was added.

ljwharbers:
Alright, I've added the types and some general descriptions. Feel free to make adjustments @luisas

famosab:
@luisas @ljwharbers Are you still working on this? :)

luisas:
Oh I never saw this, apologies. I will look at it now!

ljwharbers:
Same, I completely forgot about this. I resolved the minor conflicts. I assume it will need a pull from master since it's been a while but let's see.

luisas (approved):
LGTM! :) 

I reverted the changes in the meta for irmsd - as there is no need to update that module now

PR #6889, opened 2024-10-29
Fix krakenuniq filename splitting into prefix

Midnighter:
Updating the module will address the following bug in the taxprofiler pipeline https://github.com/nf-core/taxprofiler/issues/533


jfy133:
@Midnighter does this main.nf file get through the new language server in the VSCode extension (assuming you're using that)? :eyes:

Midnighter:
I had to not use the auto formatting because of the backslashes issue. It would move the functions below the process, but otherwise no major warnings.

Midnighter:
I'm closing this after our talk and will open a new PR that changes the module to accept sample identifiers as another input and use those as a prefix.

PR #6891, opened 2024-10-29
New module: scvitools/solo single-cell doublet detection

LeonHafner:
## PR checklist
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [ ] Remove all TODO statements.
- [ ] Emit the `versions.yml` file.
- [ ] Follow the naming conventions.
- [ ] Follow the parameters requirements.
- [ ] Follow the input/output options guidelines.
- [ ] Add a resource `label`

PR #6897, opened 2024-10-29
New module: doubletdetection

LeonHafner:
## PR checklist

- [x] Add a resource `label`


LeonHafner:
This module does not support conda, as the bioconda `doubletdetection` build is not working properly.

modules/nf-core/doubletdetection/main.nf
```diff
@@ -0,0 +1,49 @@
+process DOUBLETDETECTION {
+    tag "$meta.id"
+    label 'process_medium'
+
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'oras://community.wave.seqera.io/library/anndata_louvain_pip_doubletdetection:42d2326cc250350b':
+        'community.wave.seqera.io/library/anndata_louvain_pip_doubletdetection:cbe92394c10372fa' }"
+
+    // Prevent /tmp mount for singularity and set the MATPLOTLIB TMPDIR
+    containerOptions "${workflow.containerEngine == 'singularity' ?
```
LeonHafner, line 10: @nictru we talked about the `containerOptions` and how to use them in this example and also other PRs (https://github.com/nf-core/modules/pull/6891)
I basically see the following options:

1. Move the definition of the variables inside the python script, s.t. it gets executed regardless of the executer. Then we should still keep the `--no-mount tmp` option (at least for singularity and apptainer) unless we are really sure nothing writes into the mounted `tmp` directory to avoid conflicts with other users using the same process at the same time. This could be done either on pipeline level or on process level (preferred).
2. Set the environment variables in the process itself (as implemented right now), but regardless of which container type is used (remove `if`-block). Docker, Singularity and Apptainer definitely support the `--env` flag, however this is not clear for all other container types (Podman,, Sarus, ...). The `--no-mount tmp` argument would probably be defined on pipeline level, as not all container types support it.
3. Leave it as it is implemented right now. This is not optimal, as apptainer and docker will have their temporary data stored at different locations then singularity.

nictru: Let's see what insights arise from [this thread](https://nfcore.slack.com/archives/CE6SDBX2A/p1730720790953729)

nictru (approved):
LGTM

edmundmiller:
Hey @LeonHafner how did it get built by Seqera containers then? What's was in the environment.yml?

LeonHafner:
Hey @edmundmiller,
the Seqera containers are built using the `doubletdetection` package via pip. However, nf-core linting doesn’t allow pip dependencies in the `environment.yml`. To address this, I added the package to Bioconda, but after some painful hours of debugging and trying to match the versions with the containers, it turned out the Bioconda version is broken and isn’t working as expected.

PR #6903, opened 2024-10-29
pbmm2/align

tanyasarkjain:
## PR checklist

Closes #XXX 

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:


modules/nf-core/pbmm2/align/main.nf
```diff
@@ -0,0 +1,45 @@
...
+    path "versions.yml"           , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    pbmm2 align $fasta $bam ${prefix}.pbmm2.bam --num-threads ${task.cpus}
```
fellen31, line 26: $args needs to be in there as well, and I would remove the `.pbmm2` in the file name. If someone wants to have it, they can add it through prefix. 

```suggestion
    pbmm2 \\
        align \\
        $args \\
        $fasta \\
        $bam \\
        ${prefix}.bam \\
        --num-threads ${task.cpus}
```

modules/nf-core/pbmm2/align/main.nf
```diff
@@ -0,0 +1,45 @@
...
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        pbmm2: \$(pbmm2 --version |& sed '1!d ; s/samtools //')
```
fellen31, line 30: Perhaps? 

```suggestion
        pbmm2: \$(pbmm2 --version |& sed '1!d ; s/pbmm2 //')
```

modules/nf-core/pbmm2/align/meta.yml
```diff
@@ -0,0 +1,61 @@
...
+        description: Sorted BAM/CRAM/SAM file
+        pattern: "*.{bam,cram,sam}"
+  - - meta:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. `[ id:'sample1', single_end:false ]`
+    - fasta:
+        type: file
+        description: Sorted BAM/CRAM/SAM file
```
fellen31, line 34: Description does not match

modules/nf-core/pbmm2/align/meta.yml
```diff
@@ -0,0 +1,61 @@
...
+            Groovy Map containing sample information
+            e.g. `[ id:'sample1', single_end:false ]`
+      - "*.bam":
+          type: file
+          description: Sorted BAM/CRAM/SAM file
+          pattern: "*.{bam,cram,sam}"
+          ontologies:
+            - edam: "http://edamontology.org/format_25722"
+            - edam: "http://edamontology.org/format_2573"
+            - edam: "http://edamontology.org/format_3462"
```
fellen31, line 50: What is this? Never seen before.

modules/nf-core/pbmm2/align/main.nf
```diff
@@ -0,0 +1,45 @@
...
+
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/pbmm2:1.14.99--h9ee0642_0':
+        'biocontainers/pbmm2:1.14.99--h9ee0642_0' }"
+
+    input:
+    tuple val(meta), path(bam)
+    tuple val(meta), path(fasta)
```
fellen31, line 13:
```suggestion
    tuple val(meta2), path(fasta)
```

modules/nf-core/pbmm2/align/main.nf
```diff
@@ -0,0 +1,45 @@
+process PBMM2_ALIGN {
+    tag "$meta.id"
+    label 'process_single'
```
fellen31, line 3: Alignment using a single core will take a long time :) 

```suggestion
    label 'process_large'
```

modules/nf-core/pbmm2/align/meta.yml
```diff
@@ -0,0 +1,61 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/meta-schema.json
+name: "pbmm2_align"
+description: Alignment with PacBio's minimap2 frontend
+keywords:
+  - sort
+  - example
+  - genomics
```
fellen31, line 7: These should be updated to match the process.

fellen31 (commented):
NIce! A few comments, but looks good overall!

modules/nf-core/pbmm2/align/meta.yml
fellen31: Never mind, I guess that this was added to the template when you create a new module? But the first link is broken, could you remove the last "2" in `http://edamontology.org/format_25722`?

tanyasarkjain: Yeah - it was added in the template, so I kept it - will do!

tanyasarkjain:
I have also made a module for pbsv/call - for whatever reason the conda portion for pbsv/call specifically is now failing

fellen31 (approved):
LGTM! 🚀

modules/nf-core/pbmm2/align/main.nf
```diff
@@ -0,0 +1,51 @@
...
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/pbmm2:1.14.99--h9ee0642_0':
+        'biocontainers/pbmm2:1.14.99--h9ee0642_0' }"
+
+    input:
+    tuple val(meta), path(bam)
+    tuple val(meta2), path(fasta)
+
+    output:
+    tuple val(meta), path("*.bam"), emit: bam
```
rollf, line 16: If the input `bam` was actually `input.bam` (and not `input.fastq.gz`), I believe this should cause the output `bam` to contain both files (input + aligned bam).

rollf: Assuming I am right, the input could simply be staged differently (i.e. `tuple val(meta), path(bam, stageAs: "input/*")`).

fellen31: I believe input files should not be included, https://www.nextflow.io/docs/latest/process.html#output-files-path

rollf: I see, thanks! The relevant documentation says "Input files are not included (unless includeInputs is true)".

rollf: On a second thought, having `${prefix}.bam` as input would be problematic, but not due to the asterisk but due to the fact that the input file and the output file both share the same name :thinking:

fellen31: Many modules have something similar to this: https://github.com/nf-core/modules/blob/dc71da5a29cee04ab99390d6f5f5d2fb9a8cb9ec/modules/nf-core/catpack/addnames/main.nf#L24-L26

You could add it to this module.

rollf: If `${prefix}` gets changed specifically for this reason _and_ the result is plugged into picard/collecthsmetrics _and_ that is in turn plugged into multiqc, you end up with:

image

(Picture from a friend ^^)

We might of course blame picard or multiqc. We may also circumvent the situation by using multiqc's `[extra_]fn_clean_exts` option (i.e. add "-aligned" in the example above). It does not feel "right" to change the module config for pbmm2/align because the input is a bam file and then also change the multiqc config to "revert" the sample name again.

So the `stageAs` approach feels better to me. If I were to open a PR, I would not go for the suggested error message.

PR #6910, opened 2024-10-30
325 mixed model in limma

KamilMaliszArdigen:
## PR checklist

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


modules/nf-core/limma/differential/templates/limma_de.R
```diff
@@ -317,6 +318,19 @@ if (!is.null(opt\$use_voom) && opt\$use_voom) {
...
+
+        corfit_args <- list(object = voom_result, design = design, block = sample.sheet[[opt\$block]])
+        corfit = do.call(duplicateCorrelation, corfit_args)
+
+        voom_args <- list(counts = dge, design = design, plot = FALSE, correlation = corfit\$consensus.correlation)
+        voom_result <- do.call(voom, voom_args)
+
+        corfit_args <- list(object = voom_result, design = design, block =  sample.sheet[[opt\$block]])
+        corfit = do.call(duplicateCorrelation, corfit_args)
+    }
```
pinin4fjords, line 332:
```suggestion
}

if (!is.null(opt\$block) {
    corfit = duplicateCorrelation, voom_result, design = design, block = sample.sheet[[opt\$block]])

    if (!is.null(opt\$use_voom) && opt\$use_voom) {
        data_for_fit <- voom(counts = dge, design = design, plot = FALSE, correlation = corfit\$consensus.correlation)
    }
}

if (!is.null(opt\$use_voom) && opt\$use_voom) {
```

Use of `block` and `duplicateCorrelation` is not voom-specific, so we should do something like this. It will require you to split the Voom conditional slightly. So the line above will also have to be changed from:

```
voom_result <- voom(dge, design)
```

to

```
data_for_fit <- voom(dge, design)
```

modules/nf-core/limma/differential/templates/limma_de.R
```diff
@@ -353,6 +367,8 @@ if (! is.null(opt\$block)){
 }
 if (! is.null(opt\$correlation)){
     lmfit_args[['correlation']] <- as.numeric(opt\$correlation)
+} else if (!is.null(opt\$mixed_model) && opt\$mixed_model) {
+    lmfit_args[['correlation']] <- corfit\$consensus.correlation
```
pinin4fjords, line 371:
```suggestion
    lmfit_args[['correlation']] <- corfit\$consensus.correlation
    lmfit_args[['block']] <- sample.sheet[[opt\$block]]
```

We should pass the `block` to lmFit in the same way as to duplicateCorrelation(), right? Also, maybe remove the existing `block` conditional above.

modules/nf-core/limma/differential/templates/limma_de.R
```diff
@@ -353,6 +367,8 @@ if (! is.null(opt\$block)){
 }
 if (! is.null(opt\$correlation)){
     lmfit_args[['correlation']] <- as.numeric(opt\$correlation)
+} else if (!is.null(opt\$mixed_model) && opt\$mixed_model) {
```
pinin4fjords, line 370:
```suggestion
} else if (! is.null(opt\$block)) {
```

pinin4fjords (changes requested):
I can't quite specify it correctly via comments, but hopefully you can see what I mean. 

Basically I think we can simplify and control this from the existing `block` parameter (which isn't properly implemented as-is).

I believe the logic needs to be rearranged so the following happens:

1. non-voom matrix initialisation (`data_for_fit <- as.matrix(intensities.table)`) OR the Voom DGE + first Voom run, saving to a common variable.
2. In either case, if user has specified `opt$block`, run `duplicateCorrelation`
    2 b) in the Zoom case, re-run with the correlation
3. Then to the normalised matrix handling conditional on Voom.

modules/nf-core/limma/differential/templates/limma_de.R
KamilMaliszArdigen: Well one can pass correlation value form via opt\$correlation what I think should have higher priority. With that in mind I don't think that we can merge it in a clean way and keep this flexibility.

modules/nf-core/limma/differential/templates/limma_de.R
```diff
@@ -78,6 +78,7 @@ opt <- list(
     exclude_samples_col = NULL,
     exclude_samples_values = NULL,
     use_voom = FALSE,
+    mixed_model = FALSE,
```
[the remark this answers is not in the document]
KamilMaliszArdigen: This will simplify workflow for sure.

KamilMaliszArdigen:
> I can't quite specify it correctly via comments, but hopefully you can see what I mean.
> 
> Basically I think we can simplify and control this from the existing `block` parameter (which isn't properly implemented as-is).
> 
> I believe the logic needs to be rearranged so the following happens:
> 
> 1. non-voom matrix initialisation (`data_for_fit <- as.matrix(intensities.table)`) OR the Voom DGE + first Voom run, saving to a common variable.
> 2. In either case, if user has specified `opt$block`, run `duplicateCorrelation`
>    2 b) in the Zoom case, re-run with the correlation
> 3. Then to the normalised matrix handling conditional on Voom.

I think that I understand what you mean. I updated the code based on your suggestions. There is only one difference which related to lmfit_args logic. Please tak a look on one unresolved comment and let me know what do you think about this logic.

pinin4fjords (changes requested):
I'd rather we had just the one duplicateCorrelation call, for ease of maintenance. I've made a PR against your feature branch for illustration: https://github.com/nf-core/modules/pull/6930

(I've also made a suggestion there about the fixed effects/ random effects thing.)

KamilMaliszArdigen:
Well I will not argue with the refactoring of the module. I personally find my version significantly easier to follow when one needs to understand what is the module logic. removal of single line duplication with addition of ifs in my opinion don't improves readability. With that mentioned this a module created by you so your suggestions have been merged.

KamilMaliszArdigen:
Also thank you for adding the warning for the fixed effects/ random effects thing.

pinin4fjords:
> Well I will not argue with the refactoring of the module. I personally find my version significantly easier to follow when one needs to understand what is the module logic. removal of single line duplication with addition of ifs in my opinion don't improves readability. With that mentioned this a module created by you so your suggestions have been merged.

I understand what you mean. But in my suggestion I'm sacrificing a bit of readability in the name of maintainability. i.e. it's harder for someone to come along and mistakenly update just one `duplicateCorrelation` call without the other.

pinin4fjords (approved):

KamilMaliszArdigen:
> > Well I will not argue with the refactoring of the module. I personally find my version significantly easier to follow when one needs to understand what is the module logic. removal of single line duplication with addition of ifs in my opinion don't improves readability. With that mentioned this a module created by you so your suggestions have been merged.
> 
> I understand what you mean. But in my suggestion I'm sacrificing a bit of readability in the name of maintainability. i.e. it's harder for someone to come along and mistakenly update just one `duplicateCorrelation` call without the other.

In the long term you are probably right.

KamilMaliszArdigen:
Thank you for your review

PR #6912, opened 2024-10-30
Expect input of prefixes to KrakenUniq

Midnighter:
Updating the module will address the following bug in the taxprofiler pipeline https://github.com/nf-core/taxprofiler/issues/533 but requires that the pipeline be refactored to supply the new input.

### DIFF
--- a/modules/nf-core/krakenuniq/preloadedkrakenuniq/main.nf
+++ b/modules/nf-core/krakenuniq/preloadedkrakenuniq/main.nf
@@ -8,7 +8,8 @@ process KRAKENUNIQ_PRELOADEDKRAKENUNIQ {
         'biocontainers/krakenuniq:1.0.4--pl5321h6dccd9a_2' }"
 
     input:
-    tuple val(meta), path(sequences)
+    // We stage sequencing files in a sub-directory so we don't accidentally gzip them later.
+    tuple val(meta), path(sequences, name: 'sequences/*'), val(prefixes)
     val sequence_type
     path db
     val ram_chunk_size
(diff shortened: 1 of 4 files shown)


modules/nf-core/krakenuniq/preloadedkrakenuniq/main.nf
```diff
@@ -38,25 +40,26 @@ process KRAKENUNIQ_PRELOADEDKRAKENUNIQ {
...
     if (meta.single_end) {
+        assert sequences.size() == prefixes.size()
+        command_inputs = [sequences, prefixes].transpose().collect { seq, prefix -> "${seq} ${prefix}" }
+
         """
+        cat <<-END_INPUTS > ${command_inputs_file}
+        ${command_inputs.join('\n        ')}
+        END_INPUTS
+
         krakenuniq \\
```
jfy133, line 55:
```suggestion
        // Preload the database into memory
        krakenuniq \\
```

jfy133: This confused me at first

modules/nf-core/krakenuniq/preloadedkrakenuniq/main.nf
```diff
@@ -38,25 +40,26 @@ process KRAKENUNIQ_PRELOADEDKRAKENUNIQ {
...
-        strip_suffix() {
-            local result=\$1
-            # Strip any file extensions.
-            echo "\${result%%.*}"
-        }
-
-        printf "%s\\n" ${sequences} | while read FASTQ; do \\
-            PREFIX="\$(strip_suffix "\${FASTQ}")"
-
+        while IFS=' ' read -r SEQ PREFIX; do
```
jfy133, line 62:
```suggestion
        // Execute the classification
        while IFS=' ' read -r SEQ PREFIX; do
```

modules/nf-core/krakenuniq/preloadedkrakenuniq/meta.yml
```diff
@@ -25,6 +25,11 @@ input:
         type: file
         description: List of input files containing sequences. All of them must be either
           in FASTA or FASTQ format.
+    - prefixes:
+        type: string
+        description: >
+          List of sample identifiers or filename prefixes. Must correspond in order and
```
jfy133, line 31: Maybe give a suggestio how to fix the order (maybe with `.sort`? Although I can imagine in some very rare cases someone might be weird and not name things such as pairs properly)

jfy133 (approved):

modules/nf-core/krakenuniq/preloadedkrakenuniq/meta.yml
Midnighter: I think that in the normal case, prefixes will be constructed from samples as they come out of a channel and thus they will be in the same order. I do not want to suggest something here that potentially makes the use more complicated than necessary and potentially leads to user errors.

I am in favor of creating a batching function and having that properly tested, but I don't know where we would provide that from?

PR #6931, opened 2024-11-04
nf-test for ashlar

kbestak:
Migrate pytest to nf-test for ASHLAR

## PR checklist

Closes https://github.com/nf-core/modules/issues/6895

- [x] Add a resource `label`
  - For modules:
    - [ ] `nf-core modules test <MODULE> --profile conda` --- CONDA fails tests - not reproducible between Docker - Conda?


modules/nf-core/ashlar/main.nf
```diff
@@ -49,4 +49,25 @@ process ASHLAR {
         ashlar: \$(ashlar --version | sed 's/^.*ashlar //' )
     END_VERSIONS
     """
+
+    stub:
+    def args          = task.ext.args           ?: ''
```
SPPearce, line 54:
```suggestion
```

SPPearce (approved):

kbestak:
Thanks for the review @SPPearce, I don't fully understand why the conda test is failing as the tool is on bioconda and deployed as a biocontainer. I would assume that's an issue. Should the conda test be skipped?

SPPearce:
There is presumably something inside the test_all.ome.tif file that is different between docker/singularity and conda.
This happens often for some file types, such as bam files, where the header of the file encodes a file path that is different between the two environments (i.e. `/opt/bin` instead of /`usr/bin` or something).
Generally at that point we test for existence of the file, or the presence of a string rather than the entire file.

kbestak:
The issue I'm seeing is that the conda output matched the md5sum in the previous test, even though the environment definition was the same.

kbestak:
I'm also noticing an issue with application of flat and dark field correction - should be looked into before merging.

kbestak:
A short update, I've just tested and a newly built Seqera container with ASHLAR from bioconda produces the same outputs as the conda version, but I haven't yet found the precise reason for the difference.

PR #6932, opened 2024-11-04
Set mkfastq reports and stats outputs to optional

alanmmobbs93:
## PR checklist

Closes [mkfastq fails on AWS Batch #281](https://github.com/nf-core/demultiplex/issues/281)

### Description of error reported:

`cellranger/mkfastq` process fails when running `nf-core/demultiplex` on AWS Batch with `-profile test_mkfastq,docker`. 
When running locally, the output folders are created (but without files) hence the process ends succesfully. When running on AWS Batch the process fails with the following error:

``` bash
Missing output file(s) `*_outs/outs/fastq_path/Reports` expected by process `NFCORE_DEMULTIPLEX:DEMULTIPLEX:MKFASTQ_DEMULTIPLEX:CELLRANGER_MKFASTQ (test_sample.1)`
```

Some explanation may come from the following [AWS doc](https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-folders.html) explaining that S3 is a flat system where files are grouped by common prefixes to simulate folders. It also explains that the S3 folders behave differently than local systems. So, uploading an empty folder to S3 may be the cause of the error reported by @KallyopeComp.

## Changes
* Set `reports` output to `optional: true`.
* Set `stats` output to `optional: true`.

Notes:
* Module snapshot was also tested.
* Module linting throws a warning related to the container version.

### DIFF
--- a/modules/nf-core/cellranger/mkfastq/main.nf
+++ b/modules/nf-core/cellranger/mkfastq/main.nf
@@ -10,8 +10,8 @@ process CELLRANGER_MKFASTQ {
     output:
     tuple val(meta), path("*_outs/outs/fastq_path/**/*.fastq.gz")          , emit: fastq
     tuple val(meta), path("*_outs/outs/fastq_path/Undetermined*.fastq.gz") , optional:true, emit: undetermined_fastq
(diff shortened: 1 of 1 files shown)


nschcolnicov (approved):
LGTM!

PR #6940, opened 2024-11-05
remove pre-commit config validator hook

mashehu:
I thought we did this already... it's just too noisy with updates and we don't need to update the config that often.


SPPearce (approved):

SPPearce:
I just merge the PR whenever I see it....

edmundmiller (approved):
The renovate zombie

PR #6943, opened 2024-11-06
Update quast container

nschan:
This is a more recent version of the quast container, which comes with bwa in the path. See also here https://github.com/bioconda/bioconda-recipes/pull/36236

## PR checklist


ewels (approved):
Assuming the conda version is unchanged, LGTM

PR #6945, opened 2024-11-07
Tidy up Python script to fetch Singularity HTTPS image URLs

ewels:
Tidy up script and make it more of a CLI tool.

Should be good to run within a GitHub Actions workflow now.

### DIFF
--- a/.github/scripts/wave_singularity.py
+++ b/.github/scripts/wave_singularity.py
@@ -2,7 +2,8 @@
 # /// script
 # requires-python = ">=3.10"
 # dependencies = [
-#     "httpx",
+#     "requests",
+#     "rich",
 # ]
 # ///
 
@@ -31,7 +32,7 @@ click.rich_click.SHOW_ARGUMENTS = True
 @click.command()
 @click.option(
     "--platform-pat",
-    envvar="TOWER_ACCESS_TOKEN",
+    envvar="SEQERA_ACCESS_TOKEN",
     show_envvar=True,
     help="Platform authentication token",
 )
--- a/.github/scripts/wave_singularity.py
+++ b/.github/scripts/wave_singularity.py
@@ -8,47 +8,79 @@
 
 import logging
 
-import httpx
+import requests
+import rich_click as click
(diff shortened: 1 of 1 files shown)


.github/scripts/wave_singularity.py
```diff
@@ -8,47 +8,79 @@
...
-    print(container_url)
-
-except httpx.RequestError as exc:
-    print(f"An error occurred while requesting {exc.request.url!r}.")
-    print("No singularity image for you")
+
+@click.command()
+@click.option(
+    "--platform-pat",
+    envvar="TOWER_ACCESS_TOKEN",
```
mashehu, line 34: is that really still the token name?

.github/scripts/wave_singularity.py
```diff
@@ -8,47 +8,79 @@
 
 import logging
 
-import httpx
+import requests
+import rich_click as click
+from rich.logging import RichHandler
```
mashehu, line 13: maybe we should add a requirements.txt or similar to make install these dependencies?

mashehu (approved):

.github/scripts/wave_singularity.py
ewels: For now, yes. Though we can set this to whatever we like as we're calling the API directly and not using the Wave CLI.

So probably makes sense to future proof ourselves 👍🏻

.github/scripts/wave_singularity.py
ewels: Depends if we expect anyone except GitHub Actions runner to run this. If we do, then we should probably stick it in nf-core/tools in fairness..

If it's just for the GHA runner I don't really care much

mashehu: I was just thinking it might be easier to keep track of things if one needs to add them to a requirements.txt compared to a step in an action, but maybe in the end its the same amount of work

mashehu: pfff, reading docs. totally overrated.

PR #6881, opened 2024-10-29
parabricks/applybqsr version updated and tests migrated to nf-test

gitanoqevaporelmundoentero:
Tools applybqsr is updated to latest version (4.3.2-1) and tests were migrated to new convention (from pytest to nf-test).

## PR checklist

Closes #6861 

- [X] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [X] Remove all TODO statements.
- [ ] Emit the `versions.yml` file.
- [ ] Follow the naming conventions.
- [ ] Follow the parameters requirements.
- [ ] Follow the input/output options guidelines.
- [ ] Add a resource `label`
- [ ] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - For modules:
    - [ ] `nf-core modules test <MODULE> --profile docker`
    - [ ] `nf-core modules test <MODULE> --profile singularity`
    - [ ] `nf-core modules test <MODULE> --profile conda`
  - For subworkflows:
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile docker`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile singularity`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile conda`

PR #6286, opened 2024-08-23
Use new nf-test features

edmundmiller:
- **ci: Attempt to split everything out**
- **ci: Add changed since, sharding, and ci**
- **ci: Add filter to try to get jobs split up**
- **ci: Switch to only-changed**
- **ci: See if follow-dependencies works without "related-tests"**

PR #6953, opened 2024-11-07
Add GPU-based tests

mashehu:
moved the nf-test logic into a separate actions.yml so I can reuse it between for non-gpu and gpu cases. Did some clean-up in the action while at it.


edmundmiller:
I love that we can do that. I'm sad because it'll mean I have to redo #6286

tests/config/nextflow.config
```diff
@@ -25,6 +25,10 @@ if ("$PROFILE" == "singularity") {
     podman.enabled = true
     podman.userEmulation = true
     podman.runOptions = "--runtime crun --platform linux/x86_64 --systemd=always"
+} else if ("$PROFILE" == "gpu") {
```
GallVp, line 28: Does `if else` mean that only one of the profiles will be active at a time? For a successful test with docker/gpu, we need both docker and gpu?

nf-test uses the nf-test.config from the project root directory (https://github.com/nf-core/modules/blob/473857e48869a078b64932241acb1cc742460530/nf-test.config) to locate the profiles defined in the [tests/config/nf-test.config](https://github.com/nf-core/modules/blob/473857e48869a078b64932241acb1cc742460530/tests/config/nf-test.config). Should we also define the gpu profile there?

.github/workflows/test.yml
```diff
@@ -666,108 +665,45 @@ jobs:
...
+    name: nf-test-gpu
+    needs: [nf-test-changes]
+    if: ( needs.nf-test-changes.outputs.paths != '[]' )
+    strategy:
+      fail-fast: false
+      matrix:
+        path: ["${{ fromJson(needs.nf-test-changes.outputs.paths) }}"]
+        profile: [conda, docker_self_hosted, singularity]
+        include:
+          - path: modules/nf-core/parabricks/applybqsr
```
GallVp, line 687: Will this trigger tests for these three modules whenever `if: ( needs.nf-test-changes.outputs.paths != '[]' )` is true.

GallVp:
Let's touch,

1. A non-gpu module
2. A gpu module
3. A non-gpu and a gpu module

to see how it goes in each case.

Cheers!

.github/workflows/test.yml
```diff
@@ -82,8 +82,7 @@ jobs:
         with:
           fetch-depth: 2 # To retrieve the preceding commit.
 
-      # TODO: change back to using dorny/paths-filter when https://github.com/dorny/paths-filter/pull/133 is implemented
-      - uses: mirpedrol/paths-filter@main
+      - uses: donrny/paths-filter@v3
```
mashehu, line 85:
```suggestion
      - uses: dorny/paths-filter@v3
```

tests/config/nextflow.config
```diff
@@ -25,6 +25,10 @@ if ("$PROFILE" == "singularity") {
...
+    docker.runOptions       = '-u $(id -u):$(id -g) --gpus all'
+    apptainer.runOptions    = '--nv'
+    singularity.runOptions  = '--nv'
```
adamrtalbot, line 31: I'd prefer this stuff to go in the configuration of the individual tools, so it can function as a reference implementation for users. Adding it here feels too "secret sauce" that will be hard for other people to follow.

Plus we should use `process.containerOptions` for process specific stuff.

Or we could add `label 'gpu'` to achieve a similar thing.

```
process {
    withLabel "gpu" {
        process.containerOptions = "--gpus all"
    }
}
```

tests/config/nextflow.config
```diff
@@ -25,6 +25,10 @@ if ("$PROFILE" == "singularity") {
     podman.enabled = true
     podman.userEmulation = true
     podman.runOptions = "--runtime crun --platform linux/x86_64 --systemd=always"
+} else if ("$PROFILE" == "gpu") {
+    docker.runOptions       = '-u $(id -u):$(id -g) --gpus all'
```
adamrtalbot, line 29:
```suggestion
    docker.runOptions       = '--platform=linux/amd64 -u $(id -u):$(id -g) --gpus all'
```

adamrtalbot (commented):
I think we should enable the GPU container flags a slightly different way.

tests/config/nextflow.config
mashehu: +1 for label

mashehu:
For reference, the gpu profile worked for the parabricks tests in https://github.com/nf-core/modules/pull/6881

tests/config/nextflow.config
GallVp: How do we handle the singularity/apptainer option `--nv` then?

sateeshperi (approved):

tests/config/nextflow.config
mashehu: added the missing profile

tests/config/nf-test.config
edmundmiller, line 1: @mashehu Which one is the right gpu profile? 🫠

mashehu: what do you mean?

edmundmiller: There's two `gpu` profiles in this config now

PR #6964, opened 2024-11-11
checkm/qa: Add tsv_output input

dialvarezs:
This PR adds the new input `tsv_output` to checkm/qa module.
Why? By default, `checkm qa` does not produce a TSV file, instead it produces a "pretty" table. To produce TSV output, the `--tab_table` option must be used. A TSV file is generally is better for integration with other tools and workflows, so explicitly allowing this option is a good idea.

In particular, this change addresses an issue I encountered while implementing tests for the GUNC + CheckM integration in the `gunc/mergecheckm` module, where the command expects the output in TSV format.

## PR checklist

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


jfy133 (changes requested):
As `--tab_table` is not mandaotry for the tool to run, this should not be added as an input parameter.

A pipeline developer could add `--tab_table` themselves in a pipeline via `ext.args`.

The question here is more whether `--tab_table` produces an extra output file e.g., with `.tsv` that should be included in the output channel stuff

(I have 30m to work toiday, and this caught my eye @dialvarezs sorry I can't look at the mag PR but I want to test taht properly)

dialvarezs:
Hi @jfy133, thanks for the review.
To be honest, I wasn’t entirely sure about this either, but I couldn’t find a cleaner way to address the issue with the `gunc/mergecheckm` test. The problem is that it specifically requires TSV output to function correctly, and with the default output, the command fails.

So, if this modification is not an option, I guess we could have the following options:
1. Is there some way to configure `ext.args` directly within the test itself?
2. Add an additional step in the test to transform the output into TSV format maybe?
3. Don't test `gunc/mergecheckm`

PS: Don't worry about the mag PR, take your time. These PRs in the modules are related since the gunc PR will solve the issue I told you in the mag pr.

jfy133:
Ahh I see!

Ok in that case, yes indeed no channel needed and yes you can specify ext.srgs in tests via a config file, but in fact an even better method where you specify it within the test file as in this PR which will be merged in soon: 

https://github.com/nf-core/website/pull/2832

You're welcome to try that out :)

dialvarezs:
That's great! Just what I've been looking for.

PR #6971, opened 2024-11-12
stub + nf-test

sateeshperi:


sateeshperi:
@nf-core-bot fix linting

modules/nf-core/nucmer/tests/main.nf.test
```diff
@@ -0,0 +1,61 @@
...
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(
+                    file(process.out.delta[0][1]).readLines()[2..-1],
+                    file(process.out.coords[0][1]).readLines()[2..-1],
```
maxulysse, line 29: Can we get the md5 of these lines instead of the lines themselves?

maxulysse (approved):
`file(process.out.delta[0][1]).readLines()[2..-1]`
Just wondering here if we can md5 this instead of having this content in the snapshot, but that's ok for me anyway.
:rocket:

sateeshperi:
not able to get md5 easily...lets go with this for now as it is still deterministic

maxulysse:
> not able to get md5 easily...lets go with this for now as it is still deterministic

I meant something like:
```
file(process.out.delta[0][1]).readLines()[2..-1].md5
```

sateeshperi:
ya i tried `file(process.out.delta[0][1]).readLines()[2..-1].md5()` and it didn't work

PR #6979, opened 2024-11-13
Fix/biscuit apps

sateeshperi:
* Update version and use seqera containers
* update align module to explicitly require fasta
* update tests and snap


modules/nf-core/biscuit/biscuitblaster/main.nf
```diff
@@ -24,23 +25,24 @@ process BISCUIT_BLASTER {
     def args = task.ext.args ?: ''
     def args2 = task.ext.args2 ?: ''
     def args3 = task.ext.args3 ?: ''
-    def biscuit_cpus = (int) Math.max(Math.floor(task.cpus*0.95),1)
-    def samtools_cpus = task.cpus-biscuit_cpus
+    def biscuit_cpus = [(task.cpus * 0.9) as int, 1].max()
+    def samtools_cpus = (task.cpus - biscuit_cpus < 1) ? biscuit_cpus : (task.cpus - biscuit_cpus)
     """
-    INDEX=`find -L ./ -name "*.bis.amb" | sed 's/\\.bis.amb\$//'`
+    ln -sf \$(readlink $fasta) $index/$fasta
```
matthdsm, line 31: wouldn't this be better as a `stageAs`?

sateeshperi: There was a edge case that recently popped up that prompted me to go this route. 

see more discussion about nested symlinks issue in #rubber-duck [here](https://nfcore.slack.com/archives/C04PPJLT1T3/p1728054131378659?thread_ts=1727963842.762909&cid=C04PPJLT1T3)

matthdsm (approved):

PR #6998, opened 2024-11-15
Fastq align dedup bismark

sateeshperi:


subworkflows/nf-core/fastq_align_dedup_bismark/main.nf
```diff
@@ -0,0 +1,154 @@
...
+    )
+    ch_bismark_summary = BISMARK_SUMMARY.out.summary
+    ch_versions        = ch_versions.mix(BISMARK_SUMMARY.out.versions)
+
+    /*
+     * MODULE: Run samtools sort on dedup bam
+     */
+    SAMTOOLS_SORT_DEDUPLICATED (
+        alignments,
+        [[:],[]] // Empty map and list as is optional input but required for nextflow
```
mahesh-panchal, line 116: My preference is usually to describe what the input is in the comment since we can already see it's an empty input.

subworkflows/nf-core/fastq_align_dedup_bismark/main.nf
```diff
@@ -0,0 +1,154 @@
...
+
+    /*
+     * Generate bismark summary report
+     */
+    BISMARK_SUMMARY (
+        BISMARK_ALIGN.out.bam.collect{ it[1].name }.ifEmpty([]),
+        alignment_reports.collect{ it[1] }.ifEmpty([]),
+        alignment_reports.collect{ it[2] }.ifEmpty([]),
+        BISMARK_METHYLATIONEXTRACTOR.out.report.collect{ it[1] }.ifEmpty([]),
+        BISMARK_METHYLATIONEXTRACTOR.out.mbias.collect{ it[1] }.ifEmpty([])
```
mahesh-panchal, line 106: It would be nice to be more descriptive of what the `it` variable is. 
Does `toList` work here? Should this process still run even if all channels are empty()?

subworkflows/nf-core/fastq_align_dedup_bismark/main.nf
```diff
@@ -0,0 +1,154 @@
...
+     */
+    SAMTOOLS_INDEX_DEDUPLICATED (
+        SAMTOOLS_SORT_DEDUPLICATED.out.bam
+    )
+    ch_versions = ch_versions.mix(SAMTOOLS_INDEX_DEDUPLICATED.out.versions)
+
+    /*
+     * Collect MultiQC inputs
+     */
+    ch_multiqc_files = BISMARK_SUMMARY.out.summary.ifEmpty([])
```
mahesh-panchal, line 131:
```suggestion
    ch_multiqc_files = BISMARK_SUMMARY.out.summary
```
There's no need to inject an empty list there.

subworkflows/nf-core/fastq_align_dedup_bismark/main.nf
sateeshperi: @mahesh-panchal could you elaborate on 
> It would be nice to be more descriptive of what the it variable is.

need a comment beside on what the it variable for that line represents ?

mahesh-panchal: Rename the `it` variable to something e.g., `.collect { meta, bam -> bam }`

sateeshperi: thanks for catching this.....the process is actually not working as expected....fixing now

subworkflows/nf-core/fastq_align_dedup_bismark/main.nf
```diff
@@ -0,0 +1,154 @@
...
+
+    /*
+     * Align with bismark
+     */
+    BISMARK_ALIGN (
+        ch_reads,
+        ch_fasta,
+        ch_bismark_index
+    )
+    ch_versions = ch_versions.mix(BISMARK_ALIGN.out.versions)
```
mahesh-panchal, line 41:
```suggestion
    ch_versions = ch_versions.mix(BISMARK_ALIGN.out.versions.first())
```
I'm not sure it matters as duplicates are filtered out in the end anyway so it's up to you if you want to use `.first()` on all the module version outputs.

mahesh-panchal (approved):
Looks good. 
It's my first time seeing an nf-test nextflow.config and I was surprised at the complexity. 
Would it not be more transparent to set what the actual flags are in the test without all the params?

subworkflows/nf-core/fastq_align_dedup_bismark/main.nf
sateeshperi: good point. will apply

sateeshperi: and you are right again that the test doesn't need that extensive config options. removed as applicable for the test

sateeshperi: going to leave it out for now. Will add back if convinced it has utility

mahesh-panchal:
@sateeshperi Some of those params are not included in the config. I think you need to make a patch

mahesh-panchal:
Never mind, I saw they're used as channel inputs.

PR #7000, opened 2024-11-15
fq lint module update: exit on failed validation

oligomyeggo:
I am proposing a change to the nf-core [fq lint module](https://nf-co.re/modules/fq_lint/), such that the process will exit if a FastQ file fails validation. This will allow us to add this module into nf-core pipelines like rnaseq (see relevant PR [here](https://github.com/nf-core/rnaseq/pull/1453)) and validate FastQ files early in the workflow, preventing pipelines from continuing and running more computationally-heavy steps on corrupted FastQ files.

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/fq/lint/main.nf
+++ b/modules/nf-core/fq/lint/main.nf
@@ -1,7 +1,6 @@
 process FQ_LINT {
     tag "$meta.id"
     label 'process_low'
-    errorStrategy 'terminate'
 
     conda "${moduleDir}/environment.yml"
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
@@ -30,10 +29,5 @@ process FQ_LINT {
     "${task.process}":
         fq: \$(echo \$(fq lint --version | sed 's/fq-lint //g'))
     END_VERSIONS
-
-    if ! tail -n 1 ${prefix}.fq_lint.txt | grep -q 'fq-lint end'; then
-        echo "ERROR: Linting failure detected for ${meta.id}. See ${prefix}.fq_lint.txt for details."
-        exit 1 
-    fi
     """
 }
--- a/modules/nf-core/fq/lint/main.nf
+++ b/modules/nf-core/fq/lint/main.nf
@@ -1,6 +1,7 @@
 process FQ_LINT {
     tag "$meta.id"
     label 'process_low'
+    errorStrategy 'terminate'
 
     conda "${moduleDir}/environment.yml"
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
@@ -29,5 +30,10 @@ process FQ_LINT {
     "${task.process}":
         fq: \$(echo \$(fq lint --version | sed 's/fq-lint //g'))
     END_VERSIONS
+
+    if ! tail -n 1 ${prefix}.fq_lint.txt | grep -q 'fq-lint end'; then
+        echo "ERROR: Linting failure detected for ${meta.id}. See ${prefix}.fq_lint.txt for details."
+        exit 1 
+    fi
     """
 }
--- a/subworkflows/nf-core/fastq_qc_trim_filter_setstrandedness/main.nf
+++ b/subworkflows/nf-core/fastq_qc_trim_filter_setstrandedness/main.nf
@@ -124,6 +124,7 @@ workflow FASTQ_QC_TRIM_FILTER_SETSTRANDEDNESS {
             ch_fastq_lint.map{ meta, fastqs -> [meta, fastqs.flatten()] }
         )
         ch_versions = ch_versions.mix(FQ_LINT.out.versions.first())
+        ch_lint_log = FQ_LINT.out.lint
     }
 
     ch_reads
@@ -324,6 +325,7 @@ workflow FASTQ_QC_TRIM_FILTER_SETSTRANDEDNESS {
 
     emit:
 
+    lint_log        = ch_lint_log
     reads           = ch_strand_inferred_fastq
     trim_read_count = ch_trim_read_count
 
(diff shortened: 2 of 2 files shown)


subworkflows/nf-core/fastq_qc_trim_filter_setstrandedness/main.nf
```diff
@@ -114,6 +116,16 @@ workflow FASTQ_QC_TRIM_FILTER_SETSTRANDEDNESS {
     ch_trim_read_count = Channel.empty()
     ch_multiqc_files   = Channel.empty()
 
+    //
+    // MODULE: Lint FastQ files
+    //
+    if(!skip_linting) {
+        FQ_LINT (
```
pinin4fjords, line 123: Could you add the output of the linting module as output of this subworkflow please?

modules/nf-core/fq/lint/main.nf
```diff
@@ -29,5 +30,10 @@ process FQ_LINT {
     "${task.process}":
         fq: \$(echo \$(fq lint --version | sed 's/fq-lint //g'))
     END_VERSIONS
+
+    if ! tail -n 1 ${prefix}.fq_lint.txt | grep -q 'fq-lint end'; then
```
pinin4fjords, line 34: We shouldn't change the module, generally we try to let the tools work as they do, without overlaying additional logic.

We can add some logic to the subworkflow + workflow that would filter out any libraries failing linting. You may already see logic in rnaseq that we use for trimming, strand failures etc, and we can use the same mechanism.

But my understanding is that fq_lint will return a non-zero error code on a failure? That should trigger a stop without this. Am I mistaken?

pinin4fjords (changes requested):

modules/nf-core/fq/lint/main.nf
oligomyeggo: @pinin4fjords Yeah, kind of. fq lint has different validators that are all assigned a [different code](https://github.com/stjude-rust-labs/fq#lint), and if a FastQ file fails the linting, you will see in the log what code is failed on. You can't grep for the codes themselves in the log, as the log will say upfront what validators are enabled by code. That's why I had to grep for "fq-lint end" and check if it **_doesn't_** exist. That's the list line you'll see in the linting log if the linting was successful, so failed FastQ files won't have that in their log.

So, fq lint doesn't actually put out an error code itself. If we left the current module alone as is, then that process will successfully complete on any FastQ file, even corrupt ones, which defeats the purpose of adding the linting into the pipeline.

I am definitely open to trying out some different logic to filter out samples that failed linting. Are you thinking those samples would just filtered out, and the pipeline would keep going with the good samples, or that the pipeline would exit if it finds a sample that failed linting? We find that our users often want the pipeline to stop so that they can try reuploading (or redownloading from source and then reuploading) the offending FastQ file, which has worked in a lot of cases. Maybe another conditional that could be set as to if a user wants the pipeline to exit if a sample fails linting or if they want the pipeline to continue with successful samples?

oligomyeggo:
@pinin4fjords - sorry, I accidentally clicked "Resolve conversation" on the topic of adding the module output to the subworkflow! I added that into my last commit, if that is what you meant/had in mind?

modules/nf-core/fq/lint/main.nf
pinin4fjords: What I mean is the following. Assume a bad fastq file, foo.fq, with content like:

```
@SEQ_ID_1
GATTTGGGGTTTTCCCAGTT
THIS SHOULD NOT BE HERE
IIIIIIIIIIIIIIIIIIII
@SEQ_ID_2
AGCTAGCTAGGCTAGCTAAG
+
JJJJJJJJJJJJJJJJJJJJ
```

The output of `fq lint foo.fq` is: 

```
2024-11-17T17:03:29.770623Z  INFO fq::commands::lint: fq-lint start
2024-11-17T17:03:29.770645Z  INFO fq::commands::lint: validating single end read
2024-11-17T17:03:29.770651Z  INFO fq::validators: disabled validators: []
2024-11-17T17:03:29.770658Z  INFO fq::validators: enabled single read validators: ["[S003] NameValidator", "[S004] CompleteValidator", "[S002] AlphabetValidator", "[S001] PlusLineValidator", "[S005] ConsistentSeqQualValidator", "[S006] QualityStringValidator"]
2024-11-17T17:03:29.770668Z  INFO fq::validators: enabled paired read validators: []
2024-11-17T17:03:29.770671Z  INFO fq::commands::lint: starting validation
/data/foo.fq:3:1: [S001] PlusLineValidator: missing + prefix
```

This produces a non-zero error code:

```
$ echo $?
1
```

Because Nextflow runs processes with `set -e` turned on, the task will fail when it encounters that issue. 

I see that `fq` has its own codes, but that wasn't what I meant.

pinin4fjords: For the specific rnaseq example, we'd just make sure to set the appropriate label so the process gets enough resource to start. 

But to the general point, you can use dynamic retry strategies to catch exit codes reflecting e.g. OOM:

```
errorStrategy { task.exitStatus in 137..140 ? 'retry' : 'terminate' }
```

pinin4fjords (approved):
Looks good- thanks!

modules/nf-core/fq/lint/main.nf
pinin4fjords: Actually you could PR against [this](https://github.com/nf-core/modules/blob/af84433c2e619cf28e026d4140b1b8763ae3690b/modules/nf-core/fq/lint/main.nf#L3), if you feel the current value isn't appropriate.

pinin4fjords:
@oligomyeggo sorry, I was pulled into other things for a couple of weeks, we can get this merged soon.

Before we do, [something came up on the workflow](https://github.com/nf-core/rnaseq/issues/1456): maybe we could run this linting after every FASTQ manipulation, to check that no problems were introduced? 

We'd just have to import the module multiple times, .e.g.:

```
include { FQ_LINT as FQ_LINT_AFTER_TRIMMING } from '../../../modules/nf-core/fq/lint/main'
```

Would you like to do that, or shall I?

pinin4fjords:
Never mind, this is actually a bit fiddly, I'm going to do it on a new PR

pinin4fjords:
Thanks for the work! Closing in favour of https://github.com/nf-core/modules/pull/7123

PR #7010, opened 2024-11-17
Fasta index bismark bwameth

sateeshperi:


maxulysse:
should you deprecate the old one?

sateeshperi:
@maxulysse which old one ?

maxulysse:
This is not to replace/update an existing subworklow ?

sateeshperi:
not in nf-core...it was a local subwf in methylseq that I am porting upstream

edmundmiller (approved):

PR #7011, opened 2024-11-18
update preseq/lcextrap errorStrategy to ignore

sateeshperi:


modules/nf-core/preseq/lcextrap/main.nf
```diff
@@ -1,7 +1,7 @@
 process PRESEQ_LCEXTRAP {
     tag "$meta.id"
     label 'process_single'
-    label 'error_retry'
+    label 'error_ignore'
```
edmundmiller, line 4: Can we remove the label and make a custom one?

Retry it once, then if it fails the second time we ignore it?

sateeshperi: `label 'error_retry_ignore'` ?

edmundmiller: No just define the error strategy at that point 😂

SPPearce: Retrying once and then failing sounds great, from what I recall this fails anytime the sample doesn't have enough depth.

PR #7013, opened 2024-11-18
update bcftools pluginsplit

LouisLeNezet:
## PR checklist

Here is a PR to update the `BCFTOOLS_PLUGINSPLIT` process.

The issue solved here was:
- The process didn't work in the case of a single sample present in the file with an equivalent output name -> solved by letting the file in the `$prefix` folder
- No possibility to add a suffix to the output files generated -> `ext.suffix` argument added

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


modules/nf-core/bcftools/pluginsplit/main.nf
```diff
@@ -15,17 +15,18 @@ process BCFTOOLS_PLUGINSPLIT {
     path(targets)
 
     output:
-    tuple val(meta), path("*.{vcf,vcf.gz,bcf,bcf.gz}"), emit: vcf
-    tuple val(meta), path("*.tbi")                    , emit: tbi, optional: true
-    tuple val(meta), path("*.csi")                    , emit: csi, optional: true
-    path "versions.yml"                               , emit: versions
+    tuple val(meta), path("*/*.{vcf,vcf.gz,bcf,bcf.gz}"), emit: vcf
```
fellen31, line 18: Could you stage the input file in a folder instead to fix the first issue? 

E.g.: `tuple val(meta), path(files, stageAs: "inputs/*")`

fellen31 (commented):
I don't fully understand the first issue, do you have an input file with the same name as one of the samples in the file? 


Generally https://nf-co.re/docs/guidelines/components/modules#types-of-meta-fields

> Modules MUST NOT use ‘custom’ hardcoded meta fields. The only accepted ‘standard’ meta fields are meta.id or meta.single_end. Proposals for other ‘standard’ fields for other disciplines must be discussed with the maintainers team.

Why do you need to have suffixes?

modules/nf-core/bcftools/pluginsplit/main.nf
```diff
@@ -42,7 +43,19 @@ process BCFTOOLS_PLUGINSPLIT {
...
+        for file in ${prefix}/*; do
+            # Extract the basename
+            base_name=\$(basename "\$file")
+            # Extract the part of the basename before the first dot
+            name_before_dot="\${base_name%%.*}"
+            # Extract the extension
+            extension="\${base_name#\${name_before_dot}}"
+            # Construct the new name
+            new_name="\${name_before_dot}${suffix}\${extension}"
+            mv "\$file" "${prefix}/\$new_name"
```
fellen31, line 56: What happens here if there are dots in the sample names?

LouisLeNezet:
> I don't fully understand the first issue, do you have an input file with the same name as one of the samples in the file?

This issue happens in our phaseimpute pipeline. In some case the user give a single vcf, that might be named as `sample_names.vcf.gz` therefore when this vcf is splitted the process as it was would crash as the input name was equivalent to the new one.

> Why do you need to have suffixes?

The suffix is used to correctly name the file for multiQC in our pipeline. It also helps to filter out and get back the meta information when you transpose the channel after the split operation.

modules/nf-core/bcftools/pluginsplit/main.nf
LouisLeNezet: Okay, so that's effectively a problem...
I will change the way the splitting is done to only check for (vcf, bcf, gz, tbi, csi) then append the suffix and the extension afterwards.

fellen31:
Okay, thanks for the clarification. 

Could you work with `prefix` instead of suffix? It seems easier to me and less error prone to add something to the beginning of the name instead of the middle, and `prefix` is not really used in this module anyway.

modules/nf-core/bcftools/pluginsplit/main.nf
fellen31: See my other comment :)

LouisLeNezet:
> Could you work with prefix instead of suffix? It seems easier to me and less error prone to add something to the beginning of the name instead of the middle, and prefix is not really used in this module anyway.

This might work in my use case, I will try.

LouisLeNezet:
>  > Could you work with prefix instead of suffix? It seems easier to me and less error prone to add something to the beginning of the name instead of the middle, and prefix is not really used in this module anyway.

> This might work in my use case, I will try.

Unfortunately it doesn't as I need the sample names to be first, then a suffix and the extension last...
Would the following works for you ?

```bash
if [ -n "${suffix}" ]; then
    for file in ${prefix}/*; do
        # Extract the basename
        base_name=\$(basename "\$file")
        # Extract the extension
        extension=""
        # Remove the extension if it exists
        if [[ "\$base_name" =~ \\.(vcf|bcf)(\\.gz)?(\\.tbi|\\.csi)?\$ ]]; then
            extension="\${BASH_REMATCH[0]}"
            base_name="\${base_name%\$extension}"
        fi
        # Construct the new name
        new_name="\${base_name}${suffix}\${extension}"
        mv "\$file" "${prefix}/\$new_name"
    done
fi
```

fellen31:
This seems a lot more complicated than

```
for file in output_dir/*; do
    mv \$file $prefix\$FILE
done
```

Would it not more clean to adapt the logic in phaseimpute? Perhaps the passing the suffix in the meta? Or adding a rename module, e.g. https://github.com/nf-core/raredisease/blob/master/modules/local/rename_align_files.nf?

LouisLeNezet:
Renaming the file would add a new step, but it would be cleaner for the nf-core repository.
I will do so.

LouisLeNezet:
@fellen31 thanks a lot for the help.
The module should now be cleaner and I will adapt the pipeline to it.
Is it good for you now ?

modules/nf-core/bcftools/pluginsplit/main.nf
```diff
@@ -15,17 +15,17 @@ process BCFTOOLS_PLUGINSPLIT {
     path(targets)
 
     output:
-    tuple val(meta), path("*.{vcf,vcf.gz,bcf,bcf.gz}"), emit: vcf
-    tuple val(meta), path("*.tbi")                    , emit: tbi, optional: true
-    tuple val(meta), path("*.csi")                    , emit: csi, optional: true
-    path "versions.yml"                               , emit: versions
+    tuple val(meta), path("*/*.{vcf,vcf.gz,bcf,bcf.gz}"), emit: vcf
+    tuple val(meta), path("*/*.tbi")                    , emit: tbi, optional: true
+    tuple val(meta), path("*/*.csi")                    , emit: csi, optional: true
```
fellen31, line 20:
```suggestion
    tuple val(meta), path("outputDir/*.{vcf,vcf.gz,bcf,bcf.gz}"), emit: vcf
    tuple val(meta), path("outputDir/*.tbi")                    , emit: tbi, optional: true
    tuple val(meta), path("outputDir/*.csi")                    , emit: csi, optional: true
```

modules/nf-core/bcftools/pluginsplit/main.nf
```diff
@@ -15,17 +15,17 @@ process BCFTOOLS_PLUGINSPLIT {
...
+    path "versions.yml"                                 , emit: versions
 
     when:
     task.ext.when == null || task.ext.when
 
     script:
     def args = task.ext.args ?: ''
-    def prefix = task.ext.prefix ?: "${meta.id}"
+    def prefix = task.ext.prefix ?: ""
```
fellen31, line 28: Nice

modules/nf-core/bcftools/pluginsplit/main.nf
```diff
@@ -65,11 +69,23 @@ process BCFTOOLS_PLUGINSPLIT {
...
+    mkdir -p outputDir
+
+    cut -f 3 ${determination_file} | sed -e 's/\$/.${extension}/' > files.txt
+    while IFS= read -r filename;
+    do ${create_cmd} "outputDir/\$filename";
+    if [ -n "${index}" ]; then
+        index_file=\$(sed -e 's/\$/.${index}/' <<< \$filename);
+        touch outputDir/\$index_file;
+    fi;
+    done < files.txt
```
fellen31, line 82: Could you indent the code within the while statement?

modules/nf-core/bcftools/pluginsplit/tests/main.nf.test.snap
```diff
@@ -1,4 +1,264 @@
 {
+    "homo_sapiens - [ vcf, tbi ], [], [], [], [], - tbi with suffix": {
```
fellen31, line 2: Could you remove the suffix snaps?

fellen31 (commented):
Nice, while renaming would be an extra step, I think perhaps it's worth it to keep this module a bit simpler.

modules/nf-core/bcftools/pluginsplit/main.nf
fellen31: Could you perhaps add a comment here to explain why task.ext.prefix is empty by default and not meta.id like (almost) all other modules?

modules/nf-core/bcftools/pluginsplit/tests/main.nf.test.snap
LouisLeNezet: Done

SPPearce (commented):
I don't understand. You are now making the files both inside an output directory and you are renaming them with a prefix?

To summarise, the module takes one vcf file and makes multiple, but sometimes the output file might have the same name as the input? At which point, why don't you stage the input file into a folder, then make all the output files in the work directory?

LouisLeNezet:
> I don't understand. You are now making the files both inside an output directory and you are renaming them with a prefix?

No they were staying in the outputDir.

> To summarise, the module takes one vcf file and makes multiple, but sometimes the output file might have the same name as the input? At which point, why don't you stage the input file into a folder, then make all the output files in the work directory?

This is an excellent suggestion ! I didn't though of it...

fellen31:
Do you still need to have a prefix/suffix to in the output files in this module? 

The prefix is usually `meta.id` which often corresponds to the sample  ID, but sometimes you want to change that. But in this module the file names will already have the sample names fixed in the output file names, so I don't really see why you would need a prefix if you can rename the files within your pipeline?

LouisLeNezet:
> Do you still need to have a prefix/suffix to in the output files in this module?

I don't think if we keep both input and output in separate folder.

modules/nf-core/bcftools/pluginsplit/main.nf
```diff
@@ -8,7 +8,7 @@ process BCFTOOLS_PLUGINSPLIT {
         'biocontainers/bcftools:1.20--h8b25389_0' }"
 
     input:
-    tuple val(meta), path(vcf), path(tbi)
+    tuple val(meta), path(vcf, stageAs: "?/*"), path(tbi, stageAs: "?/*")
```
SPPearce, line 11: Just to check, do these stage in the same directory?
I would have specified a folder to put them into (like `stageAs: "input/*"`), rather than using `?`.
Can't immediately find the table in the documents that explains this.

LouisLeNezet: It seems to stage everything in a folder name `1`.
I did as it was made in the other modules, but I would also prefer to specify the folder.

SPPearce (approved):
Ok, looks good to me then!

LouisLeNezet:
Thanks to both of you @SPPearce and @fellen31 !!

PR #7021, opened 2024-11-19
ci: Only run GPU tags and rely on "changed-since"

edmundmiller:
This should allow up to actually specify expected tests GitHub actions.


edmundmiller:
Closing in favor of #7024

PR #7023, opened 2024-11-19
Update KrakenUniq: Add SE/PE to suffixes as cannot control by typical ext.prefix

jfy133:
Due to the nature of the construction of the module (for efficiency loading the database then aligning all samples in one process in the same process), we found that as we cannot control the suffix in the normal way via `ext.args`, we often get file-name collisions or overwrites at a pipeline results level when the leading prefix is the same between a single-end and a paired-end library (/end longest sentence of the week).

This PR hardcodes `.se` and `.pe` to file output suffxes to ensure they do not overwrite each other on publishing by providing a unique name.

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


Midnighter:
I was wondering if this is not an artificial problem? Wouldn't under normal circumstances the run accession always be different if there was a paired-end and a single-end sequencing run of the same sample?

Midnighter:
Additional thoughts:

* You have not configured the language server with Harshil-alignment? 😱 
* I haven't seen any other modules or pipelines adding a distinction for the filenames based on PE or SE. I don't think this part is special to KrakenUniq and would be a problem for every module, no?

jfy133:
> I was wondering if this is not an artificial problem? Wouldn't under normal circumstances the run accession always be different if there was a paired-end and a single-end sequencing run of the same sample?

Yes it somewhat is, but I don't see a way around it. In the pipeline schema for taxprofiler, we can't unique sample IDs if they don't use run accessions because we can't conditionally check the uniquness based on the presnce of another column.

Regardless, the tests are failing so it needs to be fixed :sweat_smile: 

I just realised though I think I could actually add the `.se` `.pe` prefix in modules.conf, now I've looked through youre new code... I will try that first...

> Additional thoughts:
> 
>     * You have not configured the language server with Harshil-alignment? 😱
> 
>     * I haven't seen any other modules or pipelines adding a distinction for the filenames based on PE or SE. I don't think this part is special to KrakenUniq and would be a problem for every module, no?

- I forgot to turn back on the format-on-save since they fixed 
- No, becuase normally it's easy to distinguish between single- and paried- end in a run, but as we were trying to _infer_ a prefix I thought it workign work, but see my point above.

jfy133:
Had a huddle with @Midnighter - actually we can use the same logic as the typical nf-core `ext.prefix` - so we don't have to hardcode -the diffference is actually jkust that the the prefix comes from a n input channel rather than `ext.`, so I can close this.

If it comes up again, we can have a follow up PR to updateth the `meta.yaml` to give explicit guidance on this

PR #7025, opened 2024-11-19
Add pre filter and shard counting step

mashehu:


.github/actions/number-of-shards/action.yml
edmundmiller, line 1: I love that you found that we can do composite actions in the same repo ❤️

maxulysse (approved):

.github/actions/number-of-shards/action.yml
```diff
@@ -0,0 +1,36 @@
...
+    - name: Install nf-test
+      uses: nf-core/setup-nf-test@v1
+      with:
+        version: ${{ env.NFT_VER }}
+
+    - name: Get number of shards
+      shell: bash
+      run: |
+        nftest_output=$(nf-test test --dry-run --changed-since HEAD^ --filter process --follow-dependencies)
+        if nftest_output | grep -q 'No tests found'; then
```
mashehu, line 26:
```suggestion
        if $nftest_output | grep -q 'No tests found'; then
```

mashehu:
I wouldn't at special paths just all modules and subworkflows. Then the filter step will take care of only picking up GPU tests

sateeshperi:
why aren't some tests not launching ?

mashehu:
they are not stalled, they are just required, but never launched (because we don't have anything to test). so we need to remove them from the ruleset and add a step which confirms the passing of all tests (either that they ran successfully or that they didn't have to be run)

sateeshperi:
closing this as we got some of the logic in and will polish in another PR with working fixes from methylseq

Issue #7018, opened 2024-11-18
Address limitations of using nf-test sharding


edmundmiller:

edmundmiller:
On the conda failures, I see a few options going from short-term/making more maintence work to long-term
1. Implement some "exclude" type of mechanism in GitHub actions
2. Tag all of the tests that work with conda with a "conda" tag and we select for those using a `--tag conda_works` tag in the conda matrix profile
3. Address it at the root with working on https://github.com/askimed/nf-test/issues/260

We could skip the conda tests in 3 until we get that implemented. Not the end of the world IMO given the conda headaches, but I'm biased.

PR #7028, opened 2024-11-19
ci: Disable conda for the moment

edmundmiller:
See for explaination: https://github.com/nf-core/modules/issues/7018#issuecomment-2486119169

- [ ] Need to disable conda requires as well


maxulysse (approved):

mashehu:
No, I think we should still run tests but don't require conda to pass

mashehu:
removed conda tests from required status checks in branch protection ruleset for now

Issue #7016, opened 2024-11-18
Handle GitHub runners possibly running out of space


edmundmiller:
sorry this feels very draft status. move it into a another PR before merging this one

_Originally posted by @mashehu in https://github.com/nf-core/modules/pull/6286#discussion_r1846901512_

```yml
  # get-number-of-shards:
  #   runs-on: ubuntu-latest
  #   outputs:
  #     # Needs to be a json array
  #     shards: ${{ steps.shards.outputs.shards }}
  #     total_shards: ${{ steps.shards.outputs.total_shards }}
  #   steps:
  #     - name: Install nf-test
  #       uses: nf-core/setup-nf-test@v1
  #       with:
  #         version: ${{ env.NFT_VER }}

  #     - id: shards
  #       run: |
  #         nftest_output=$(nf-test test --dry-run --changed-since HEAD^ --filter process --follow-dependencies)
  #         number_of_shards=$(echo $nftest_output | grep -o 'Found [0-9]* related test' | tail -1 | awk '{print $2}')
  #         three_tests_per_shard=$(echo $(($number_of_shards / 3)) | awk '{print int($1+0.5)}')
  #         shards_array=$(for shard in $(seq 1 $number_of_shards); do echo $shard; done | tr ' ' '\n' | jq -R . | jq -s .)
  #         echo "shards=${shards_array}" >> $GITHUB_OUTPUT
  #         echo "total_shards=${number_of_shards}" >> $GITHUB_OUTPUT
```
WIP Code

edmundmiller:
Tested in #6716 with some examples. We'll see how many issues we run into with it.

edmundmiller:
Wondering if we could use [Fusion with S3 locally](https://www.nextflow.io/docs/latest/fusion.html#local-execution-with-aws-s3) to avoid this 🤔

nathanweeks:
To document this idea somewhere...

The GitHub-hosted runners have relatively little available space (18G on `ubuntu-latest`) on the root partition due to the size of the default runner image, but have a separate relatively-unused partition (66G available space on `ubuntu-latest`) mounted at /mnt; e.g.:

```
runner@runnervmg1sw1:~/work/modules/modules$ df -h
Filesystem      Size  Used Avail Use% Mounted on
/dev/root        72G   55G   18G  76% /
tmpfs           7.9G   84K  7.9G   1% /dev/shm
tmpfs           3.2G  1.1M  3.2G   1% /run
[...]

PR #7030, opened 2024-11-19
PoC: Use Fusion for test workdir

edmundmiller:
- **build(#7016): Add required wave config**
- **test(#7016): Can you even do that?**


edmundmiller:
This one still sparks joy, but we can close it and re-open it.

PR #7034, opened 2024-11-20
Update bclconvert

sateeshperi:
updating mostly in favor of use of descriptive collect statements rather than `file[it]`


matthdsm (approved):

modules/nf-core/bclconvert/tests/main.nf.test
```diff
@@ -4,11 +4,12 @@ nextflow_process {
...
+
     tag "bclconvert"
     tag "modules"
     tag "modules_nfcore"
 
-    test("homo sapiens illumina [bcl]") {
+    test("homo_sapiens illumina [bcl]") {
         when {
             process {
                 //TODO use new test dataset when available, see https://github.com/nf-core/test-datasets/issues/996
```
SPPearce, line 15: Looks like this has been implemented:
https://github.com/nf-core/test-datasets/pull/1082
(don't have to do it now, but in case you fancy doing it while modifying the module anyway)

modules/nf-core/bclconvert/tests/nextflow.config
SPPearce, line 1: This is unnecessarily complicated here, could just be a single string.

SPPearce (dismissed):

famosab:
Needs to be added to skip_conda.yaml

famosab:
@sateeshperi Can you finalize this? 🚀

PR #7055, opened 2024-11-22
use prefix for pilon ouput

nschan:
pilon did not use prefix in the script; instead used ${meta.id} as fixed output.

## PR checklist


sguizard (approved):
LGTM

PR #7056, opened 2024-11-22
Fix CI modules lint

mirpedrol:
Fix error when linting modules which uses the file names instead of module names
https://github.com/nf-core/modules/actions/runs/11963376901/job/33353699351?pr=7052


LouisLeNezet:
Shouldn't the `subworkflows_files` also needs to be changed ?

sateeshperi (approved):

maxulysse (approved):

edmundmiller:
Thought I avoided this with the `slice` 😞 Thanks for taking care of it!

PR #6921, opened 2024-11-01
fix read variable in for loop

MartinPippel:
- issue:
  - the for loop does always use the full list of given read files in each iteration and creates the same meryl database with different database names
 
- solution:
  - use read variable name in bash for loop 

## PR checklist

- [ ] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [ ] Remove all TODO statements.
- [ ] Emit the `versions.yml` file.
- [ ] Follow the naming conventions.
- [ ] Follow the parameters requirements.
- [ ] Follow the input/output options guidelines.
- [ ] Add a resource `label`
- [ ] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - For modules:
    - [ ] `nf-core modules test <MODULE> --profile docker`
    - [ ] `nf-core modules test <MODULE> --profile singularity`
    - [ ] `nf-core modules test <MODULE> --profile conda`
  - For subworkflows:
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile docker`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile singularity`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile conda`

PR #7074, opened 2024-11-24
reduce memory for meryl count to avoid crash

MartinPippel:
Avoid that meryl count crashes due to memory issues. It seems that meryl uses the complete buffer provided via `memory=SIZE` for the kmer counts, but meryl itself uses some memory too. When using meryl count within singularity containers and slurm,  then usually slurm is killing meryl due to that overhead. 

Solution, reduce the meryl memory kmer buffer `task.memory` by 10%. This is potentially a bit to much as usually the memory overhead was around 2-4Gb, but it solved all my issues on 3 three different compute clusters (Uppmax rackham, PDC dardel, and nac).


mahesh-panchal (approved):
This is fine, but need to get the tests working.

SPPearce:
Is the same issue happening with MERFIN here?

mahesh-panchal:
No, it's also things like version strings are different between meryl conda ( doesn't have one ) and meryl singularity ( 1.4.1 ).
I'm not sure what else is different atm.

SPPearce:
MERFIN_HIST is also failing, so might be as well as the other issues:
https://github.com/nf-core/modules/actions/runs/12006625811/job/33465567907?pr=7074#step:4:413

mahesh-panchal:
When #6921 is merged, this can be closed.

PR #7075, opened 2024-11-24
UMICollapse module: Drop external dependencies from UMICollapse module tests

MatthiasZepper:
Currently, the [UMICollapse module](https://nf-co.re/modules/umicollapse) tests run on data that is generated in the test `setup{}` with external dependencies (`umitools/extract`, `bwa/index`, `bwa/mem`, `samtools/index`). While this is no problem here in the modules' repository, those tests will fail on the pipeline level, if the additional modules are not installed.

While it may, on the long-term, be useful to devise a CI step detecting and installing missing modules for tests ([Leon Rauschning made a neat proposal for this](https://nfcore.slack.com/archives/C049MBCEW06/p1731664424635429?thread_ts=1731521149.600189&cid=C049MBCEW06)), I have now, due to time contraints, opted to add the required files to the test datasets ([test-datasets/1390](https://github.com/nf-core/test-datasets/pull/1390)) and change the tests accordingly (**this PR**).

I have also updated the `umi-tools dedup` tests, since they so far used inputs without UMIs.  

## PR checklist

  - For modules:
  - For subworkflows:


modules/nf-core/umicollapse/tests/main.nf.test
```diff
@@ -13,57 +13,16 @@ nextflow_process {
     tag "bwa/mem"
```
SPPearce, line 13: We can remove the tags then - can't comment on the lines above.

SPPearce (approved):
Looks much better.
Think you just need to swap to using the nft-bam plugin to test the bam files then should be good.

modules/nf-core/umicollapse/tests/main.nf.test
MatthiasZepper: Done. Thanks a lot for the eagle eye!

SPPearce: That'd be the point of reviewing ;)

PR #7081, opened 2024-11-25
remove soft-linking of fasta instead stage in dir

sateeshperi:
To avoid nested symlinks, the changes here mount the correct dir of fasta file onto container for original symlink to work

Thanks to @mahesh-panchal for this fix! ❤️ 

`+` 

enclosing all input variables with `${}`


modules/nf-core/bismark/align/main.nf
```diff
@@ -9,7 +9,7 @@ process BISMARK_ALIGN {
 
     input:
     tuple val(meta), path(reads)
-    tuple val(meta2), path(fasta)
+    tuple val(meta2), path(fasta, stageAs: 'tmp/*')
```
mahesh-panchal, line 12: I suggest adding a comment here to explain the situation, and why this is necessary.

modules/nf-core/bismark/coverage2cytosine/main.nf
```diff
@@ -9,7 +9,7 @@ process BISMARK_COVERAGE2CYTOSINE {
 
     input:
     tuple val(meta), path(coverage_file)
-    tuple val(meta2), path(fasta)
+    tuple val(meta2), path(fasta, stageAs: 'tmp/*')
```
mahesh-panchal, line 12: Same here. A comment would be helpful explaining why this is staged but not used directly.

mahesh-panchal (approved):

modules/nf-core/bismark/coverage2cytosine/main.nf
sateeshperi: done

PR #7084, opened 2024-11-25
ci: Just lint all the subworkflows

edmundmiller:
Takes less than 2 seconds to run locally.


edmundmiller:
Took 12s to run here https://github.com/nf-core/modules/actions/runs/12014943875/job/33491915903?pr=7084

PR #7085, opened 2024-11-25
style: Sort environment.yml dependencies

edmundmiller:
This fixes an issue with the Seqera containers naming.

The Seqera containers site doesn't care about the context of what module the
image is for. It just sorts the dependencies.

This avoids having multiple containers with a bunch of different names.

Should probably handle that on the wave-cli side.

### DIFF
--- a/.github/scripts/conda_env_sorter.py
+++ b/.github/scripts/conda_env_sorter.py
@@ -23,7 +23,25 @@ def main(argv: Optional[Sequence[str]] = None) -> None:
     parser.add_argument("paths", nargs="*", type=Path)
     args = parser.parse_args(argv)
     for path in args.paths:
-        doc = yaml.load(path)
+        with path.open() as f:
+            lines = f.readlines()
+
+        # Define the schema lines to be added if missing
+        schema_lines = [
+            "---\n",
+            "# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/environment-schema.json\n",
+        ]
+
+        # Check if the first two lines match the expected schema lines
+        if lines[:2] == schema_lines:
+            header = lines[:2]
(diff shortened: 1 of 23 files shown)


edmundmiller:
Not going to do the 1258 environment.yml updates here, just going to let those get added as people make changes to them. Like a river carving through a mountain.

modules/nf-core/basicpy/environment.yml
mirpedrol, line 1: why are we removing the channels?

.github/scripts/conda_env_sorter.py
```diff
@@ -0,0 +1,70 @@
...
+                if isinstance(value, list):
+                    value.sort(key=str)
+        dicts.sort(key=str)
+        doc["dependencies"].clear()
+        doc["dependencies"].extend(others)
+        doc["dependencies"].extend(dicts)
+
+        with path.open("w") as f:
+            f.writelines(header)
+            yaml.dump(doc, f)
```
mirpedrol, line 66: I think this function is overriding the file, so the header is never added

modules/nf-core/basicpy/environment.yml
maxulysse: same question

edmundmiller: It's an empty file that fails the validation because it has no dependencies. It's not used in the module.

.github/scripts/conda_env_sorter.py
edmundmiller: Good catch!

mirpedrol (approved):
looks good! Do you know why tests are failing? they don't show an error message

edmundmiller:
Merging as all of the pre-commit checks passed.

PR #7098, opened 2024-11-26
Check for new version snapshotting

### LINKED ISSUE #6505: Snapshot content of version files instead of a hash

edmundmiller:
- **style: Add sort-simple-yaml**
- **fix: This isn't working**
- **style: Copy conda env sorter script hook**
- **chore: Remove empty environment yamls**
- **chore: Add missing environment.yml**
- **style: Run conda-env-sorter, then prettier**
- **style: Check for yaml-language-server in environment.ymls**
- **style: Have conda_env_sorter use 2 spaces**
- **style: Check version snapshot**

Closes #6505

### DIFF
--- a/.github/scripts/check_version_snapshot.py
+++ b/.github/scripts/check_version_snapshot.py
@@ -1,0 +1,39 @@
+#!/usr/bin/env python3
+# /// script
+# requires-python = ">=3.10"
+# dependencies = []
+# ///
+
+"""Check that versions are captured not hashed."""
+
+import argparse
+import re
+from pathlib import Path
+from typing import Optional, Sequence
+
+
+def check_snapshotting(lines: list[str], path: Path) -> None:
+    """Check each line for proper snapshotting."""
+    good_pattern = re.compile(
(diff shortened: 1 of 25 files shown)


.github/scripts/check_version_snapshot.py
mashehu, line 1: this should be part of `nf-core modules lint` and not here, no?

edmundmiller: I think I just imagined a world where the linting was always up to date in modules because the code for the lints lived in modules.

But that wouldn't work for local modules.

Maybe for new rules that we're just starting to introduce while we're still catching the rough edges?

edmundmiller:
Closing in favor of https://github.com/nf-core/tools/pull/3676

PR #7101, opened 2024-11-26
ci: Let wave fail

edmundmiller:


.github/workflows/wave.yml
```diff
@@ -52,12 +52,12 @@ jobs:
     # if: github.repository == 'nf-core/modules'
     if: ${{ needs.generate-matrix.outputs.conda-matrix != '[]' }}
     needs: generate-matrix
-    name: Build ${{ matrix.profile }} ${{ matrix.platform }} Container
+    name: 🌊 ${{ matrix.files }} ${{ matrix.profile }} ${{ matrix.platform }}
```
mashehu, line 55: 🙅🏻 makes it difficult for newcomers to understand what is happening

.github/workflows/wave.yml
```diff
@@ -52,12 +52,12 @@ jobs:
...
     runs-on: ubuntu-latest
     timeout-minutes: 60
     strategy:
       fail-fast: false
-      max-parallel: 4
+      # max-parallel: 2
```
mashehu, line 60: why not 4? we have cores, so why not use them?

.github/workflows/wave.yml
mashehu, line 1: can we make this action fail gracefully for now, meaning that it exits with code 0 in any way? makes the launch of this less disruptive.

mashehu (dismissed):

.github/workflows/wave.yml
mirpedrol: I agree, we don't want all the CI to start failing. Maybe we could test locally for inspecting which containers are failing? and make this action always pass

.github/workflows/wave.yml
edmundmiller: Yeah, I can agree with that. 

I was trying to clean up the names because it's also hard for newcomers and me to figure out what's happening when the name is cut off 🙃 

Open to suggestions 🤷🏻‍♂️ 

image

.github/workflows/wave.yml
edmundmiller: 1. I turned it off for the purpose of running them all quickly
2. I don't know exactly how that will translate to AWS ec2 runners. A lot of these jobs will just ping wave and be done. So I didn't want to spin up 20 runners because it's getting ready for a lot of runs and we're going to need a lot of capacity, and then it's 20 of the same wave runs because it was GATK.

.github/workflows/wave.yml
edmundmiller: We've been doing that; this was @ewels' ask because he'd like a list of what modules are failing 🤷🏻‍♂️ 

Let's do a Slack hook or do that in an alternative script if we still want the wave jobs to pass.

ewels: Are we adding successful builds to `meta.yml`? That'd be nice. Then we could just not add to `meta.yml` on fail but not trigger a CI failure for now.

A slack hook on build fail would be nice if easy 🤔

edmundmiller:
Closing as they got updated anyways

PR #7106, opened 2024-11-27
ci: Update module_name filter step

edmundmiller:
Created a small test script to prevent re-running into edge-cases and to quickly add new test cases. If we don't hate the idea I think we can wire it up to the javascript file to avoid any copy-paste errors.

### DIFF
--- a/.github/scripts/get_module_path.js
+++ b/.github/scripts/get_module_path.js
@@ -50,8 +50,10 @@ function run({ github, context }) {
         const result = get_module_names(files);
         console.log('Processed result:', result);
 
-        // Return stringified result for GitHub Actions output
-        return JSON.stringify(result);
+        // Ensure we're returning a properly formatted JSON string
+        const jsonResult = JSON.stringify(result);
+        console.log('Final JSON output:', jsonResult);
+        return jsonResult;
     } catch (error) {
         console.error("Error processing module paths:", error);
         console.error("Error details:", {
@@ -59,7 +61,6 @@ function run({ github, context }) {
             errorMessage: error.message,
             errorStack: error.stack
         });
-        // Return empty array instead of throwing
         return "[]";
     }
 }
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -90,7 +90,8 @@ jobs:
       - name: debug
         run: |
           echo "Modules files: ${{ steps.filter.outputs.modules_files }}"
-          echo "Module names: ${{ steps.module_names.outputs.result }}"
+          echo "Module names: ${{ fromJson(steps.module_names.outputs.result) }}"
+          echo "Raw module names output: ${{ steps.module_names.outputs.result }}"
 
   nf-core-lint-modules:
     runs-on: ${{ github.event.inputs.runners  || 'self-hosted' }}


.github/scripts/get_module_path.js
```diff
@@ -0,0 +1,60 @@
...
+
+// https://github.com/nf-core/modules/pull/7075
+test_case_1 = [
+    "modules/nf-core/umicollapse/tests/main.nf.test",
+    "modules/nf-core/umicollapse/tests/main.nf.test.snap",
+    "modules/nf-core/umicollapse/tests/nextflow.config",
+    "modules/nf-core/umicollapse/tests/nextflow_PE.config",
+    "modules/nf-core/umicollapse/tests/nextflow_SE.config",
+    "modules/nf-core/umitools/dedup/tests/main.nf.test",
+    "modules/nf-core/umitools/dedup/tests/main.nf.test.snap",
```
mashehu, line 33:
```suggestion
    "modules/nf-core/umitools/dedup/tests/main.nf.test.snap",
    "modules/nf-core/umitools/dedup/main.nf",
```

mashehu (dismissed):

MatthiasZepper (dismissed):
I have now only tested the script in isolation, but the test cases pass for me locally as well. Thanks for the solid work!

MatthiasZepper:
Looking at the CI failure in this PR...can it be that it was only the missing bracket in the .map() function that caused all the issues?


```
      - name: Get module name
        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7
        id: module_names
        with:
          script: |
            return [...new Set(${{ steps.filter.outputs.modules_files }}
                .map(path => path
                .replace('tests/', '')
                .replace('modules/nf-core/', '')
                .split('/')
                .slice(0, 2)
                .filter(x => !x.startsWith('main.nf') && x !== 'tests' && x !== 'meta.yml' && x !== 'environment.yml')
                .join('/'))
              )
            ];
```

modules/nf-core/upd/tests/nextflow.config
```diff
@@ -1,3 +1,3 @@
 process {
-    ext.args = { "--proband earlycasualcaiman --father hugelymodelbat --mother slowlycivilbuck --af-tag AF regions"  }
+    ext.args = { "--proband earlycasualcaiman --father hugelymodelbat --mother slowlycivilbuck --af-tag AF regions" }
```
MatthiasZepper, line 2: Not sure how this ended up in the PR, but otherwise that looks good to me.

MatthiasZepper (approved):
I think, that will do. Nice work!

[Zizmor](https://woodruffw.github.io/zizmor) nags that _`steps.module_names.outputs.result` may expand into attacker-controllable code_, but I don't really see how that could be exploited, because the inputs must be a valid file path. And in the worst case, `nf-core modules lint` is just called with some invalid module.

kenibrewer:
Just flagging that this PR is needed to allow #8003 to be merged.

edmundmiller:
Closing — lint filter logic was rewritten on main with a simpler approach (filter by main.nf/snap suffix + regex replace). This 21-commit iteration is superseded.

PR #6679, opened 2024-09-20
update JSON schema for subworkflows meta.yml

mirpedrol:
Update the JSON schema to validate subworkflows meta.yml. This adds the structure field to describe the structure of input and output channels.

```yaml
input:
  - ch_fasta:
      description: A channel containing the reference fasta file.
      structure:
          - meta:
              type: map
              description: meta map
          - fasta:
              type: file
              description: fasta file
              path: "*.fasta"
  - ch_fai:
      description: channel containing fasta.fai file
      structure: 
          - meta:
              type: map
              description: meta map
          - fai:
              type: file
              description: fasta.fai file
output:
  - versions:
      description: channel containing software versions file
      structure:
          - versions.yml
                type: file
                description: file containing versions of the software used
  - mutect2_vcf:
      description: Channel containing a list of compressed vcf files
      structure:
          - meta:
              type: map
              description: meta map
          - vcf:
              type: file
              description: vcf file 
```

PR #7115, opened 2024-11-28
Facilitate iteration in differential subworkflow

pinin4fjords:
## Objective
To facilitate iterative plans proposed by community members for the differentialabundance pipeline.

## Key Changes

1. Iterative Processing
   - Introduce `ch_inputs` channel for iterative content
   - Contents: `meta`, abundance matrix, and analysis method, plus thresholds (since these might be dependent on method etc)

2. Static Inputs
   - Contrasts and sample sheet treated as once-per-pipeline inputs
   - Combined with `ch_inputs` using Nextflow's `combine` operator

3. Module Parameterization
   - DESeq2 and limma modules receive matrix-level meta
   - Enables custom parameters via closures in pipeline modules.conf

## Implementation
- Iterate through input matrices and parameter sets for all specified contrasts
- Use `combine` operation to merge static and iterative inputs

This is quite like the differential subworkflow proposal in https://github.com/nf-core/differentialabundance/pull/288/files, except:

- Single multiMap required
- No duplication of filtering subworkflow import 
- Reduced need to muck about with meta combinations in the subworkflow
- No assumptions about meta elements (method is part of the main channel content)

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


modules/nf-core/deseq2/differential/main.nf
```diff
@@ -14,24 +14,26 @@ process DESEQ2_DIFFERENTIAL {
...
+    tuple val(out_meta), path("*.vst.tsv")                         , optional: true, emit: vst_counts
+    tuple val(out_meta), path("*.deseq2.model.txt")                , emit: model
+    tuple val(out_meta), path("*.R_sessionInfo.log")               , emit: session_info
     path "versions.yml"                                        , emit: versions
 
     when:
     task.ext.when == null || task.ext.when
 
     script:
+    out_meta = meta + meta2
```
mirpedrol, line 32: what is the content of `meta` and `meta2`? It's not clear from the `meta.yml`

subworkflows/nf-core/abundance_differential_filter/main.nf
```diff
@@ -10,111 +10,81 @@ include { CUSTOM_FILTERDIFFERENTIALTABLE      } from '../../../modules/nf-core/c
...
-    ch_abundance          // [ meta_exp, counts ] with meta keys: method, args_diff
-    ch_transcript_lengths // [ meta_exp, transcript_lengths]
-    ch_control_features   // [ meta_exp, control_features]
-    ch_samplesheet        // [ meta_exp, samplesheet ]
-    ch_contrasts          // [ meta_contrast, contrast_variable, reference, target ]
-    differential_method   // limma, deseq2, propd
-    FC_threshold          // float
-    padj_threshold        // float
+    // Things we may need to iterate
+    ch_input                 // [[meta_input], counts, analysis method,]
```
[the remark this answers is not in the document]
suzannejin: How would you deal with the cases in which, a same method might be run with different parameter options?
Would you convert the current `analysis_method` into a `analysis_meta`?

So, to allow something like this:
```
    withName: "DESEQ2"{
        ext.args = { [
            (meta.args? "${meta.args}" : ''),
            "--p_adjust_method $params.deseq2_p_adjust_method",
            "--alpha $params.deseq2_alpha"
        ].join(' ').trim() }
    }
```

modules/nf-core/deseq2/differential/main.nf
suzannejin: so `meta` is the meta for contrast channel, that has information like the contrast id, and so on.
Then the `meta2` has the meta information about the count data, experiment id, etc.
Then `meta3` and `meta4` are in this case the same as `meta2`.
But I agree that it is not clear...

subworkflows/nf-core/abundance_differential_filter/main.nf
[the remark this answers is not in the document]
pinin4fjords: For the first question, just copying from Slack for completeness:

> As I mentioned earlier, it's frowned upon to bake meta assumptions into components. Required things should be explicit inputs. Since we need method as part of the iterating logic, I've added it that way. It can then be documented correctly in the meta.yml, for example (I need to do that).
The params are optional, and don't dictate anything about the workflow logic, and we can pass them from the meta via workflow configuration without baking the assumptions into the worklow logic. I'm imagining those params will be part of the meta_input.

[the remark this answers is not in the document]
pinin4fjords: For the second, I imagine that we'd have duplicate channel content, except for meta content. You'd also have to set publishing paths specific to the meta.

modules/nf-core/deseq2/differential/main.nf
pinin4fjords: So the meta.yml says for `meta`:

```
          Groovy Map containing contrast information. This can be used at the
          workflow level to pass optional parameters to the module, e.g.
          [ id:'contrast1', blocking:'patient' ] passed in as ext.args like:
          '--blocking_variable $meta.blocking'.
```

and for meta2:

```
          Groovy map containing study-wide metadata related to the sample sheet
          and matrix
```

I felt that that was pretty descriptive when I wrote it, but happy to revise if you have suggestions.

mirpedrol: Thanks for the explanation :) `meta` is more clear, but I was wondering the exact keys for `meta2`, for example if it also contains an `id` but with a different value, because then it's unclear which one will have preference if we combine them.

subworkflows/nf-core/abundance_differential_filter/main.nf
```diff
@@ -10,111 +10,81 @@ include { CUSTOM_FILTERDIFFERENTIALTABLE      } from '../../../modules/nf-core/c
...
-        )
-
-        ch_normalised_matrix = DESEQ2_NORM.out.normalised_counts
-        ch_variance_stabilised_matrix = DESEQ2_NORM.out.rlog_counts.concat(DESEQ2_NORM.out.vst_counts)
+    // Workflow-wide things, we don't need to iterate
+    ch_samplesheet           // [ meta_exp, samplesheet ]
+    ch_transcript_lengths    // [ meta_exp, transcript_lengths]
+    ch_control_features      // [meta_exp, control_features]
+    ch_contrasts             // [[ meta_contrast, contrast_variable, reference, target ]]
+    FC_threshold             // FC threshold (value)
```
maxulysse, line 21: I don't like the uppercase

modules/nf-core/deseq2/differential/main.nf
suzannejin: Does it mean that meta2.id = experiment id will override meta.id = contrast id?
Probably that is fine as far as we define more unique meta values (id, contrast_id, instead of id for both cases) at the pipeline level to avoid overwritting?

subworkflows/nf-core/abundance_differential_filter/main.nf
```diff
@@ -10,111 +10,81 @@ include { CUSTOM_FILTERDIFFERENTIALTABLE      } from '../../../modules/nf-core/c
...
+    padj_threshold           // padj threshold (value)
```
suzannejin, line 22: what do you think about providing these params `fc_threshold, padj_threshold` implicitly through meta_input, just like the rest of parameters?
In this way, different methods (deseq2, limma, propd) can use different thresholds (in the case of propd, the fc_threshold have a totally different meaning compared to the traditional pairwise approach, and hence difference value)...

subworkflows/nf-core/abundance_differential_filter/meta.yml
```diff
@@ -12,36 +12,31 @@ components:
   - limma/differential
   - custom/filterdifferentialtable
 input:
-  - ch_abundance:
-      type: file
+  - ch_input:
```
suzannejin, line 15: I think the new structure for meta.yml for subworkflows is this one: https://github.com/nf-core/modules/pull/6679

modules/nf-core/deseq2/differential/main.nf
pinin4fjords: Maybe we should create a composite ID of the abundance and contrast IDs

subworkflows/nf-core/abundance_differential_filter/main.nf
pinin4fjords: So, when I was prototyping I had these as part of the ch_input tuple, and then decided they would probably be static, so doing it this way would be simpler. 

But it sounds like I should actually do that, so I will put them back into ch_input. 

Once again, the meta shouldn't contain required inputs.

subworkflows/nf-core/abundance_differential_filter/meta.yml
pinin4fjords: Oh yep, nice :-)

modules/nf-core/deseq2/differential/main.nf
pinin4fjords: OK, changed my mind, I don't think we should merge meta in the modules, it makes 're-joining' results at the workflow level fiddly. 

So I've put some meta handling in the subworkflow instead (without imposing assumptions)

pinin4fjords:
Just trying to track down the cryptic timezone issues in the CI...

suzannejin (approved):

PR #7169, opened 2024-12-06
Add pigz to trimgalore

edmundmiller:
- **feat: Add pigz to trimgalore**
- **style: Run lsp format**

This one shouldn't be controversial. It's documented that using pigz speeds up trimgalore. In the DSL2 transition the requirement got dropped because it was in mega environment.yml in methylseq, and atac-seq.

Closes https://github.com/nf-core/atacseq/issues/394 

> Added multi-threading support with the new option -j/--cores INT; many thanks to Frankie James for initiating this. Multi-threading support works effectively if Cutadapt is run with Python 3, and if parallel gzip (pigz) is installed: 

https://github.com/FelixKrueger/TrimGalore/blob/master/CHANGELOG.md#version-060-release-on-1-mar-2019

### DIFF
--- a/modules/nf-core/trimgalore/main.nf
+++ b/modules/nf-core/trimgalore/main.nf
@@ -51,6 +51,7 @@ process TRIMGALORE {
         "${task.process}":
             trimgalore: \$(echo \$(trim_galore --version 2>&1) | sed 's/^.*version //; s/Last.*\$//')
             cutadapt: \$(cutadapt --version)
+            pigz: \$( pigz --version 2>&1 | sed 's/pigz //g' )
         END_VERSIONS
         """
     } else {
@@ -69,6 +70,7 @@ process TRIMGALORE {
         "${task.process}":
             trimgalore: \$(echo \$(trim_galore --version 2>&1) | sed 's/^.*version //; s/Last.*\$//')
             cutadapt: \$(cutadapt --version)
+            pigz: \$( pigz --version 2>&1 | sed 's/pigz //g' )
         END_VERSIONS
         """
     }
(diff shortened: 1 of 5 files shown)


edmundmiller:
Huh, I went back and added a check and it was in the image anyways but the bioconda packages don't include it.

modules/nf-core/trimgalore/main.nf
```diff
@@ -1,37 +1,43 @@
...
-    tuple val(meta), path("*report.txt")                        , emit: log     , optional: true
-    tuple val(meta), path("*unpaired*.fq.gz")                   , emit: unpaired, optional: true
-    tuple val(meta), path("*.html")                             , emit: html    , optional: true
-    tuple val(meta), path("*.zip")                              , emit: zip     , optional: true
-    path "versions.yml"                                         , emit: versions
+    tuple val(meta), path("*report.txt"), emit: log, optional: true
+    tuple val(meta), path("*unpaired*.fq.gz"), emit: unpaired, optional: true
+    tuple val(meta), path("*.html"), emit: html, optional: true
+    tuple val(meta), path("*.zip"), emit: zip, optional: true
+    path "versions.yml", emit: versions
```
SPPearce, line 19: Ewww, wobbly lines.

edmundmiller: https://github.com/nextflow-io/language-server/issues/71 😞

SPPearce, line 19:
```suggestion
    tuple val(meta), path("*report.txt")     , emit: log     , optional: true
    tuple val(meta), path("*unpaired*.fq.gz"), emit: unpaired, optional: true
    tuple val(meta), path("*.html")          , emit: html    , optional: true
    tuple val(meta), path("*.zip")           , emit: zip     , optional: true
    path "versions.yml"					     , emit: versions
```

SPPearce (approved):

PR #7173, opened 2024-12-06
Update Trimgalore CPUs

edmundmiller:
- **fix: Rework core logic**
- **feat: Add memory limit as well**
- **More logic**


SPPearce:
@edmundmiller , something is off on the `resourceLimits` directive here.

modules/nf-core/trimgalore/main.nf
```diff
@@ -1,6 +1,8 @@
 process TRIMGALORE {
     tag "${meta.id}"
     label 'process_high'
+    // ~3GB per actual core used (15 cores * 3GB + 3GB buffer)
+    resourceLimits [memory: 48.GB, cpus: 15]
```
bentsherman, line 5: I think this syntax will not work with Nextflow, though it does work with the language server.

In any case the "idiomatic" syntax would be:
```groovy
resourceLimits memory: 48.GB, cpus: 15
```

edmundmiller:
From @FelixKrueger  in Slack:

> We tend to launch Trim Galore with --threads 8, and the CPU usage is between 600 and 800%, the memory usage between 4 and 7.2GB.  So I think the thread number is more or less linear to the number of threads passed to Cutadapt.

> It’s been many years that I did any tests, and I believe this was before switching to using pigz for compression and igzip for decompression. Might be worth another shot?



> 48GB of RAM is almost certainly asking for 40GB too much memory
> This is a typical RNA-seq run with 8 cores:
Screenshot 2024-12-06 at 21 52 38
image (3)





> that is out of the box nf-core stuff without any tinkering






Test:

```
1	
2	22m 6.973s
4	11m 7.273s
8	5m 48.919s
16	3m 11.921s
reads: 	83823914
```

Cores




> still almost linear. I think there is no need to limit anything to be honest

- @FelixKrueger

PR #7176, opened 2024-12-07
Fix abritamr/run

sateeshperi:
* Fix language server issues
* add stub + test
* update version 
* update to seqera containers

### DIFF
--- a/modules/nf-core/abritamr/run/tests/main.nf.test
+++ b/modules/nf-core/abritamr/run/tests/main.nf.test
@@ -14,21 +14,57 @@ nextflow_process {
         when {
             process {
                 """
-                input[0] = [                                                                                                                                         
-                            [ id:'test', single_end:false ], // meta map                                                                                                  
-                            file(params.modules_testdata_base_path + 'genomics/prokaryotes/haemophilus_influenzae/genome/genome.fna.gz', checkIfExists: true)                                              
-                ] 
+                input[0] = [
+                            [ id:'test', single_end:false ], // meta map
+                            file(params.modules_testdata_base_path + 'genomics/prokaryotes/haemophilus_influenzae/genome/genome.fna.gz', checkIfExists: true)
+                ]
                 """
             }
         }
 
         then {
+            assert process.success
             assertAll(
-                { assert process.success },
-                { assert snapshot(process.out).match() }
+                { assert snapshot(
+                    process.out.matches,
+                    process.out.partials,
+                    process.out.virulence,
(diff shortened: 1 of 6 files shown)


modules/nf-core/abritamr/run/tests/main.nf.test
```diff
@@ -14,21 +14,59 @@ nextflow_process {
...
-                { assert process.success },
-                { assert snapshot(process.out).match() }
+                { assert snapshot(
+                    process.out.matches,
+                    process.out.partials,
+                    process.out.virulence,
+                    process.out.out,
+                    process.out.txt,
+                    process.out.versions,
+                    path(process.out.versions[0]).yaml
```
famosab, line 35: why can't you just assert 
`process.out.versions`?

modules/nf-core/abritamr/run/tests/main.nf.test
```diff
@@ -14,21 +14,59 @@ nextflow_process {
...
+            assert process.success
             assertAll(
-                { assert process.success },
-                { assert snapshot(process.out).match() }
+                { assert snapshot(
+                    process.out.matches,
+                    process.out.partials,
+                    process.out.virulence,
+                    process.out.out,
+                    process.out.txt,
```
famosab, line 33: I thought this file is optional? Is it always present in the test runs though?

modules/nf-core/abritamr/run/tests/main.nf.test
```diff
@@ -14,21 +14,59 @@ nextflow_process {
...
+            assertAll(
+                { assert snapshot(
+                    process.out.matches,
+                    process.out.partials,
+                    process.out.virulence,
+                    process.out.out,
+                    process.out.txt,
+                    process.out.versions,
+                    path(process.out.versions[0]).yaml
+                ).match() },
```
famosab, line 67: For the stub why not just do

```suggestion
{ assert snapshot(process.out).match() }
```

modules/nf-core/abritamr/run/tests/main.nf.test
SPPearce: This is an alternative option that shows in the nf-test the actual content, which is actually more useful than the md5sum.

famosab: ahh been looking for that for a while - we did want to use that in the stub section, right?

famosab:
@sateeshperi Bump :)

PR #7186, opened 2024-12-10
Fix ci confirm pass

sateeshperi:
* Update confirm-pass jobs to correctly identify failed & cancelled
* touch a gpu & cpu module to test

### DIFF
--- a/.github/workflows/gpu-tests.yml
+++ b/.github/workflows/gpu-tests.yml
@@ -108,12 +108,17 @@ jobs:
     needs: [nf-test-gpu]
     if: always()
     steps:
-      - name: All tests ok
-        if: ${{ success() || !contains(needs.*.result, 'failure') }}
-        run: exit 0
       - name: One or more tests failed
         if: ${{ contains(needs.*.result, 'failure') }}
         run: exit 1
+
+      - name: One or more tests cancelled
+        if: ${{ contains(needs.*.result, 'cancelled') }}
+        run: exit 1
+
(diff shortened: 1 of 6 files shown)


GallVp (approved):

mashehu (approved):

.github/workflows/gpu-tests.yml
```diff
@@ -108,13 +108,18 @@ jobs:
...
         if: ${{ contains(needs.*.result, 'failure') }}
         run: exit 1
 
+      - name: One or more tests cancelled
+        if: ${{ contains(needs.*.result, 'cancelled') }}
+        run: exit 1
+
+      - name: All tests ok
+        if: ${{ contains(needs.*.result, 'success') }}
+        run: exit 0
```
adamrtalbot, line 121: @sateeshperi This will now pass if any test passes with a success status. That's why it negated the presence of a 'failure' status before.

sateeshperi: negated the presence of which `failure` status ?

GallVp: .* is not for any single test. It is for all members of 'needs' which has only one element here so it just stands for 'nf'test-gpu'. 
Secondly, if any test fails, the fail branch will be triggered first as it is now placed above the pass branch. 
I created this new confirm-pass in methylseq and we tested it quite a bit with multiple runs where tests were passing, failing, being skipped, etc.

PR #7201, opened 2024-12-11
Add nftests custom modules

nschcolnicov:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


.github/workflows/pytest-workflow.yml
```diff
@@ -117,6 +117,12 @@ jobs:
             tags: controlfreec/makegraph2
           - profile: conda
             tags: coreograph
+          - profile: conda
```
nschcolnicov, line 120: Had to add the modules to the conda exclude from pytest. I was getting the error that micromamba is not installed when trying to execute the tests in github actions. Let me know if you have any thoughts about this, or if there is something that I'm missing.

modules/nf-core/custom/matrixfilter/tests/main.nf.test
```diff
@@ -0,0 +1,57 @@
...
+                """
+            }
+        }
+
+        then {
+            assertAll (
+                { assert process.success },
+                { assert snapshot(process.out.filtered).match("filtered_stub") },
+                { assert snapshot(process.out.tests).match("tests_stub") },
+                { assert snapshot(process.out.versions).match("versions_stub") }
```
famosab, line 53:
```suggestion
                { assert snapshot(process.out).match() }
```

modules/nf-core/custom/matrixfilter/tests/nextflow.config
famosab, line 1: for testing purpose you can simplify this file and do it as described here: https://nf-co.re/docs/guidelines/components/modules#configuration-of-extargs-in-tests

I would then write down the ext.args exactly as you expect them to be which is easier to understand if someone looks at your main.nf.test file :)

modules/nf-core/custom/tabulartogseacls/main.nf
```diff
@@ -38,7 +36,17 @@ process CUSTOM_TABULARTOGSEACLS {
...
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
-        bash: \$(echo \$(bash --version | grep -Eo 'version [[:alnum:].]+' | sed 's/version //'))
+        awk: \$(mawk -W version | head -n 1 | awk '{print  \$2}')
     END_VERSIONS
     """
+
+    stub:
+    """
+    touch mock.cls
```
famosab, line 45:
```suggestion
    stub:
    def prefix = task.ext.prefix ?: "${meta.id}"
    """
    touch ${prefix}.cls
```

modules/nf-core/custom/tabulartogseacls/tests/nextflow.config
famosab, line 1: see comment for other submodule

modules/nf-core/custom/tabulartogseagct/main.nf
```diff
@@ -34,7 +32,16 @@ process CUSTOM_TABULARTOGSEAGCT {
...
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
-        bash: \$(echo \$(bash --version | grep -Eo 'version [[:alnum:].]+' | sed 's/version //'))
+        awk: \$(mawk -W version | head -n 1 | awk '{print  \$2}')
+    END_VERSIONS
+    """
+
+    stub:
+    """
+    touch mock.gct
```
famosab, line 41:
```suggestion
    stub:
    def prefix = task.ext.prefix ?: "${meta.id}"
    """
    touch ${prefix}.gct
```

famosab (changes requested):

modules/nf-core/custom/matrixfilter/main.nf
```diff
@@ -27,4 +25,12 @@ process CUSTOM_MATRIXFILTER {
     // (new variables defined here don't seem to be available in templates, so
     // we have to access $task directly)
     template 'matrixfilter.R'
```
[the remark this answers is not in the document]
nschcolnicov: Ok, this is not something that I changed, but I'll reproduce the method that was used in the R script to bash and add it the stub

modules/nf-core/custom/matrixfilter/main.nf
```diff
@@ -27,4 +25,12 @@ process CUSTOM_MATRIXFILTER {
...
+    
+    stub:
+    """
+    touch mock.filtered.tsv
+    touch mock.tests.tsv
+    touch R_sessionInfo.log
```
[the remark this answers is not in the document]
nschcolnicov: The prefix is not used by the script, why do you suggest adding it to the stub?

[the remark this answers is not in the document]
famosab: Usually we enable custom filenames by adding the prefix to the output files. I am not sure how this behaves with the R script but in theory you should be able to pass the prefix there aswell? If you cannot add it to the sessionInfo its fine to leave it as is there.

modules/nf-core/custom/matrixfilter/tests/main.nf.test.snap
```diff
@@ -0,0 +1,95 @@
...
+            [
+                "versions.yml:md5,275e382be9ca5bdd30ff5d6fc4340616"
+            ]
+        ],
+        "meta": {
+            "nf-test": "0.9.2",
+            "nextflow": "24.10.2"
+        },
+        "timestamp": "2024-12-13T21:21:11.862199262"
+    },
```
famosab, line 29: From here it seems to be as if the R script takes the prefix (here the id = test) and uses this for the tsv files.

famosab:
You will also need to remove the modules from the pytest config file. Afterwards we should be good to go :)

https://github.com/nf-core/modules/blob/816ff3b787af7e12e75a7d95f76b631f474853ee/tests/config/pytest_modules.yml#L112C1-L120C53

nschcolnicov:
There is a linting error:
image
But running the command suggested "nf-core modules lint --fix" reveals that the changes that it is expecting are not correct, it tries to convert all of the output channels into map type:
image

What do you think about this? @famosab

nschcolnicov:
> Huh that is interesting can you commit the "fixed" yaml files and push them to this PR so I can have a look?

Take a look: https://github.com/nf-core/modules/pull/7201/commits/eb84fdeb225c7f6c7002aa2501112a6d93bb490c

modules/nf-core/custom/matrixfilter/meta.yml
```diff
@@ -42,29 +42,38 @@ output:
...
             e.g. [ id:'test' ]
-      - R_sessionInfo.log:
-          type: file
-          description: Log file containing R session information
+          pattern: "*.log"
+      - "*R_sessionInfo.log":
+          type: map
+          description: |
+            Groovy Map containing information on experiment.
+            e.g. [ id:'test' ]
```
famosab, line 76: I think the update was needed because we changed one of the outputs per my request. But why `--fix` now does these kind of replacements also seems off to me. @mashehu @mirpedrol what do you guys think? is that expected behavior and should we just manually revert these changes or is something wrong with tools?

modules/nf-core/custom/matrixfilter/meta.yml
```diff
@@ -42,29 +42,38 @@ output:
...
       - meta:
           type: map
           description: |
             Groovy Map containing information on experiment.
             e.g. [ id:'test' ]
-      - R_sessionInfo.log:
-          type: file
-          description: Log file containing R session information
+          pattern: "*.log"
+      - "*R_sessionInfo.log":
```
famosab, line 72: I guess this line is the only line that would need to be changed but not the rest.

modules/nf-core/custom/matrixfilter/tests/main.nf.test.snap
famosab: ok thats fine :)

modules/nf-core/custom/matrixfilter/tests/nextflow.config
famosab, line 1: there was a new convention recently, maybe you can update to adhere to that: https://nf-co.re/docs/guidelines/components/modules#configuration-of-extargs-in-tests

modules/nf-core/custom/tabulartogseacls/tests/nextflow.config
famosab, line 1: see comment above

famosab (commented):
We are almost done :)

modules/nf-core/custom/matrixfilter/meta.yml
mirpedrol: Changing the name of the output and using `*R_sessionInfo.log` is the right change but replacing the type and description is not! We saw the same with other modules, it must be a weird bug with the `--fix` command, you can revert this.
And the same for the other outputs where this happened, like `filtered.csv` and `tests.csv`

modules/nf-core/custom/matrixfilter/meta.yml
famosab: @nschcolnicov you can revert all the other changes to the meta.yml :)

famosab: I wrote you a message on slack for the bug report :)

modules/nf-core/custom/matrixfilter/tests/nextflow.config
nschcolnicov: Just to make sure, you are requesting me to pass the values that I'm providing via ext.args using params instead of the actual value? Didn't you correct me to do it this way on a previous comment? https://github.com/nf-core/modules/pull/7201#discussion_r1883981463 Want to make sure I'm doing it the exact way you are suggesting

famosab: If I did then I am sorry. But what is meant is that the params that you need to pass are written directly in the nf.test file which makes its directly understandable for other people that look at your tests and then they do not have to go through multiple nextflow.config files.

nschcolnicov: No, sorry, you are right, I was wrong both times, I just wasn't understanding how you wanted me to structure the config, I think I have now understood and addressed the issues, let me know what you think @famosab 😄

nschcolnicov:
@famosab Tests should be passing but the docker_self_hosted one is having an issue, I don't think it is related to my changes though. I see they all fail with this error:
`touch: cannot touch '.command.trace': Permission denied`
Also linting still fails due to the issue where it wants to have everything in the meta.yml converted to meta type.

famosab:
@nschcolnicov I think there are still issues with the CI-runners. Maybe you can check slack in the next few days to see if that gets resolved.

SPPearce:
The docker CI runners should be working now.

SPPearce:
Just a linting issue on one of the modules, then should be good to merge.

nschcolnicov:
> Just a linting issue on one of the modules, then should be good to merge.

Thank you for looking into this. The linting error seems to be a bug. After running the suggested command:
`nf-core modules lint --fix custom/matrixfilter`
We get that the meta.yml is updated with inaccurate values, output files get labeled as map, among other things:
image

famosab:
I have seen this error before, its already an issue on nf-core/tools. I would suggest to fix it manually:
```
 ╭─ [✗] 1 Module Test Failed ───────────────────────────────────────────────────╮
│                      ╷                          ╷                            │
│ Module name          │ File path                │ Test message               │
│╶─────────────────────┼──────────────────────────┼───────────────────────────╴│
│ custom/matrixfilter  │ modules/nf-core/custom/… │ Module meta.yml does not   │
│                      │                          │ match main.nf. Outputs     │
│                      │                          │ should contain:            │
│                      │                          │ {'filtered': ['meta',      │
│                      │                          │ '.filtered.tsv'], 'tests': │
│                      │                          │ ['meta', '.tests.tsv'],    │
│                      │                          │ 'session_info': ['meta',   │
│                      │                          │ '*R_sessionInfo.log'],     │
│                      │                          │ 'versions':                │
│                      │                          │ ['versions.yml']} Run      │
│                      │                          │ nf-core modules lint --fix │
│                      │                          │ to update the meta.yml     │
│                      │                          │ file.                      │
│                      ╵                          ╵                            │
╰──────────────────────────────────────────────────────────────────────────────╯
```

modules/nf-core/custom/matrixfilter/main.nf
```diff
@@ -13,7 +13,7 @@ process CUSTOM_MATRIXFILTER {
     output:
     tuple val(meta), path("*.filtered.tsv")             , emit: filtered
     tuple val(meta), path("*.tests.tsv")                , emit: tests
-    tuple val(meta), path("R_sessionInfo.log")          , emit: session_info
+    tuple val(meta), path("*R_sessionInfo.log")          , emit: session_info
```
famosab, line 16: You probably only need to adjust the naming to the changes here and the linting will pass :)
`modules/nf-core/custom/matrixfilter/meta.yml`

nschcolnicov: You were right! Thank you!

modules/nf-core/custom/matrixfilter/tests/main.nf.test
```diff
@@ -0,0 +1,92 @@
...
+                    "--minimum_abundance ${params.filtering_min_abundance}",
+                    (params.filtering_min_proportion ? "--minimum_proportion ${params.filtering_min_proportion}" : ''),
+                    (params.filtering_grouping_var ? "--grouping_variable \"${params.filtering_grouping_var}\"" : ''),
+                    (params.filtering_min_proportion_not_na ? "--minimum_proportion_not_na \"${params.filtering_min_proportion_not_na}\"" : ''),
+                    (params.filtering_min_samples_not_na ? "--minimum_samples_not_na \"${params.filtering_min_samples_not_na}\"" : '')
+                ].join(' ').trim()
+            }
+            process {
+                """
+                input[0] = [ [id:"test"], file("https://github.com/nf-core/test-datasets/raw/refs/heads/differentialabundance/modules_testdata/SRP254919.salmon.merged.gene_counts.top1000cov.assay.tsv",checkIfExists: true) ]
```
famosab, line 79: One more (small) thing - I would ask you to update your links so that they follow the convention and look like this:
```nextflow
file(params.modules_testdata_base_path + 'genomics/sarscov2/illumina/vcf/test.vcf.gz', checkIfExists: true)
```

famosab: That also means that all the test-datasets that you need must be present on the `modules` branch of the test-datasets repository. I know that there will be some restructuring around the test-datasets so maybe it makes sense to have them present there so they do not get lost and the tests stop working. I can review a PR over at the test-datasets repo when you add them.

SPPearce: @nschcolnicov , the file paths to the data in the modules directory you've just added should look like this (so it can be a local clone of the repository)

modules/nf-core/custom/matrixfilter/tests/nextflow.config
famosab: I think it looks good now! 👍

modules/nf-core/custom/tabulartogseacls/tests/main.nf.test
```diff
@@ -0,0 +1,63 @@
...
+            params {
+                module_args = [ "separator": "\t", "variable": "treatment" ]
+            }
+            process {
+                """
+                input[0] = [
+                    [id:'treatment_mCherry_hND6_sample_number', variable:'treatment', reference:'mCherry', target:'hND6', blocking:'sample_number'],
+                    file(params.modules_testdata_base_path + "genomics/mus_musculus/rnaseq_expression/SRP254919.samplesheet.sample_metadata.tsv", checkIfExists: true)
+                ]
+                """
```
famosab, line 25: why do you not reuse the args in the input? are the module_args used in any way by the module?

famosab: judging from the code below they are used but then the doubling in the input is confusing to me

famosab (commented):
Looks good to me mostly. Wondering why the tests fail though...

modules/nf-core/custom/tabulartogseacls/tests/main.nf.test
nschcolnicov: I'm not sure what you mean by this comment. The module args are being used by the module [here](https://github.com/nf-core/modules/blob/364398322af7abc19d2bb11011761bd715c2e885/modules/nf-core/custom/tabulartogseacls/main.nf#L21), and [here](https://github.com/nf-core/modules/blob/364398322af7abc19d2bb11011761bd715c2e885/modules/nf-core/custom/tabulartogseacls/main.nf#L23), and lastly [here](https://github.com/nf-core/modules/blob/364398322af7abc19d2bb11011761bd715c2e885/modules/nf-core/custom/tabulartogseacls/main.nf#L25).
The value used in the modules.config would typically be:
module_args = [ "separator": "\t", "variable": "$meta.variable" ], to make sure it matches the meta.variable value. In this case however, it is hardcoded since it is for an nf-test.
Could you elaborate a bit more?

SPPearce:
> Looks good to me mostly. Wondering why the tests fail though...

Think the runners are having a bad day again...

modules/nf-core/custom/tabulartogseacls/tests/main.nf.test
famosab: I meant using the meta.variable and meta.separator fields for the input of the module_args. I am unsure where this would need to happen (probably on the module level) but the duplication here seems odd to me.

famosab: For example variable
```
def variable = args.variable
```
is determined by accessing the args.variable. But if this is handed via the meta anyways (as in your test case) why cant this be accessed from meta.variable?

nschcolnicov: Ah I see what you mean, I'm not sure why it was done in that way. My guess is that they either wanted to be able to use different values from the ones they have in the meta coming from their samplesheet or maybe not all samplesheets provide these values in their meta, so they provide them externally. But since the scope of this ticket is to add nf-tests, I would prefer to not modify this, as it will likely mess with any pipeline currently using this module. In any case it would be better to open a new ticket to assess this.

SPPearce: I don't know the reasoning behind this either, but I agree with @nschcolnicov that this can be modified later rather than in this PR.
I think that current guidelines would have these as a value channel rather than on either args.variable or meta.variable.

atrigila (approved):
Looks good to me! I agree with opening a new ticket with the suggestion so this new nf-tests can move forward :)

SPPearce (approved):

famosab (approved):

PR #7207, opened 2024-12-12
use detailed profiles in tests/nf-test.config

sateeshperi:


.github/workflows/gpu-tests.yml
```diff
@@ -74,14 +74,14 @@ jobs:
...
     runs-on: "gpu"
-    name: "GPU | ${{ matrix.profile }} | ${{ matrix.shard }}"
+    name: "GPU Test | ${{ matrix.profile }} | ${{ matrix.shard }}"
     needs: nf-test-changes
     if: ${{ needs.nf-test-changes.outputs.modules != '[]' || needs.nf-test-changes.outputs.subworkflows != '[]' }}
     strategy:
       fail-fast: false
       matrix:
         shard: [1, 2]
-        profile: [docker_self_hosted, singularity] # conda?
```
mashehu, line 84: Please add conda. We should exclude the modules that don't use it afterwards, like we do in the normal tests

tests/config/nextflow.config
```diff
@@ -6,34 +6,75 @@ params {
...
-    memory = 3.GB
-    time = 2.h
+    resourceLimits = [
+        cpus: 2,
+        memory: '4.GB',
+        time: '1.h'
+    ]
 }
 
-if ("$PROFILE" == "singularity") {
```
mashehu, line 14: Will this still work for pytest?

sateeshperi: not sure...I can leave it in there but, I think it should work

sateeshperi: but, feel free to add changes and merge this

mashehu (approved):

PR #7220, opened 2024-12-13
Add propd to abundance_differential_filter subworkflow

suzannejin:
## PR checklist

Closes #XXX 

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:


subworkflows/nf-core/abundance_differential_filter/main.nf
```diff
@@ -56,45 +57,80 @@ workflow ABUNDANCE_DIFFERENTIAL_FILTER {
         .combine(ch_contrasts.first()) // Just taking the first contrast
         .multiMap(criteria)
 
-    // Perform normalization and differential analysis
+    // ----------------------------------------------------
+    // Perform differential analysis with limma
```
pinin4fjords, line 61:
```suggestion
    // Run Limma just once to generate a normalised matrix, configure so 
    // that we don't pass subsetting arguments anddisregard differential 
    // outputs 
```

subworkflows/nf-core/abundance_differential_filter/main.nf
```diff
@@ -56,45 +57,80 @@ workflow ABUNDANCE_DIFFERENTIAL_FILTER {
...
+        norm_inputs.samples_and_matrix.filter{it[0].method == 'limma'}
+    )
+
+    LIMMA_DIFFERENTIAL(
+        inputs.contrasts.filter{it[0].method == 'limma'},
+        inputs.samples_and_matrix.filter { it[0].method == 'limma' }
+    )
+
+    // ----------------------------------------------------
+    // Perform differential analysis with deseq2
```
pinin4fjords, line 75:
```suggestion
    // Run DESeq2 just once to generate a normalised matrix, configure so 
    // that we don't pass subsetting arguments and disregard differential 
    // outputs 
```

subworkflows/nf-core/abundance_differential_filter/main.nf
```diff
@@ -56,45 +57,80 @@ workflow ABUNDANCE_DIFFERENTIAL_FILTER {
...
-    // Perform normalization and differential analysis
+    // ----------------------------------------------------
+    // Perform differential analysis with limma
+    // ----------------------------------------------------
+
+    LIMMA_NORM(
+        norm_inputs.contrasts.filter{it[0].method == 'limma'}.first(),
+        norm_inputs.samples_and_matrix.filter{it[0].method == 'limma'}
+    )
+
```
pinin4fjords, line 68:
```suggestion
//
// Perform differential analysis with limma
//

```

subworkflows/nf-core/abundance_differential_filter/main.nf
```diff
@@ -56,45 +57,80 @@ workflow ABUNDANCE_DIFFERENTIAL_FILTER {
...
+        inputs.contrasts.filter{it[0].method == 'propd'},
+        inputs.samples_and_matrix.filter { it[0].method == 'propd' }
     )
 
-    // Combine results
-    ch_results = DESEQ2_DIFFERENTIAL.out.results.mix(LIMMA_DIFFERENTIAL.out.results)
-    ch_normalised_matrix = DESEQ2_NORM.out.normalised_counts.mix(LIMMA_NORM.out.normalised_counts)
-    ch_model = DESEQ2_DIFFERENTIAL.out.model.mix(LIMMA_DIFFERENTIAL.out.model)
+    // ----------------------------------------------------
+    // Recollect results
```
pinin4fjords, line 102:
```suggestion
    // Collect results
```

subworkflows/nf-core/abundance_differential_filter/main.nf
```diff
@@ -109,8 +145,14 @@ workflow ABUNDANCE_DIFFERENTIAL_FILTER {
     )
 
     emit:
-    results_genewise           = ch_results.map{meta, results -> [meta, results] }
```
pinin4fjords, line 112: good spot

subworkflows/nf-core/abundance_differential_filter/meta.yml
```diff
@@ -122,5 +132,8 @@ authors:
   - "@bjlang"
   - "@araiz2001"
   - "@suzannejin"
+  - "@roskamsh"
+  - "@syangaskhan"
+  - "@WackerO"
```
pinin4fjords, line 137: Maybe add yourself as a maintainer?

Did all these authors contribute to your version of the differential subworkflow?

subworkflows/nf-core/abundance_differential_filter/main.nf
pinin4fjords: The normalisations don't get passed to the differential modules, so they should sit separately (which is why they were positioned distinctly before). I'm fine with your arrangement, but if we're mixing them we should probably label them appropriately actually explain why it's done this way.

pinin4fjords (changes requested):
Good work, minor comments

subworkflows/nf-core/abundance_differential_filter/main.nf
suzannejin: Regarding this, are the normalized matrices in the NORM modules the exact matrices that are used in the DE modules? If so, maybe it would be nice at some point to rearrange the module so NORM -> goes to DE?

subworkflows/nf-core/abundance_differential_filter/meta.yml
suzannejin: The last three they actually helped during the hackathon to call the multi-channel logic on DESeq2 and limma.
I'm not exactly sure up to which point the contributions should be added as author here in the meta, but wondering that it is anyway nice to credit them

suzannejin:
@pinin4fjords Thank you for the review!
I applied your suggestions and added some more comments to make the intentions behind NORM clear.

I also changed PROPR_PROPD to PROPD_DIFFERENTIAL to remain consistent in the naming of DIFFERENTIAL modules inside the subworkflow. But I recall that it is usually not recommended to change modules names. What do you think about this?

subworkflows/nf-core/abundance_differential_filter/main.nf
pinin4fjords: So, DESeq2 and Limma Voom output normalised matrices (they take raw values), so they don't need to be passed normalised matrices.

So, originally on differentialabundance we just took the first output of DESEQ2_DIFFERENTIAL and used the normalised matrix from there. But it caused problems, since that first contrast might have options (the exclude samples ones) that lead to the matrix being subsetted, so the normalised matrix we got did not always contain all samples.

For that reason we developed this workaround, where the differential module is run via a separate alias, just once, not passing any subsetting options (see the config here: https://github.com/nf-core/differentialabundance/blob/3dd360fed0dca1780db1bdf5dce85e5258fa2253/conf/modules.config#L162), and we take the normalised output from there, disregarding the differential.

pinin4fjords:
> @pinin4fjords Thank you for the review! I applied your suggestions and added some more comments to make the intentions behind NORM clear.
> 
> I also changed PROPR_PROPD to PROPD_DIFFERENTIAL to remain consistent in the naming of DIFFERENTIAL modules inside the subworkflow. But I recall that it is usually not recommended to change modules names. What do you think about this?

DESEQ2_DIFFERENTIAL is the actual module name. It was named that way in anticipation of modules for other parts of DESeq2 (though that hasn't happened so far). No need to alias your module unless you're using it in more than one place.

subworkflows/nf-core/abundance_differential_filter/main.nf
suzannejin: Why for DIFF you do have cases where you remove samples, whereas for NORM you want the full set? Is it because you need them for GSEA? Or that the plotting need the entire set?

pinin4fjords (approved):

PR #7233, opened 2024-12-17
Add functionality for `--unknown-slide` in `spaceranger/count`

fasterius:
There is an option to for Space Ranger Count to run without knowing the exact slide, `--unknown-slide`, that has been requested by users (see https://github.com/nf-core/spatialvi/issues/99#issue-2728970112). This PR adds this to the `spaceranger/count` module, which uses the already existing `meta.slide` value as input.

This would work well for the `nf-core/spatialvi` pipeline, since no other changes would need to be done; rather than specifying a slide (e.g. `H1-VJQNJXM`) in the `slide` samplesheet column the user would specify `visium-[1,2,2-large,hd]` and leave the `area`  column empty. I'm sure there are other ways of doing it, though, and would be happy to change it if somebody thinks another solution would be better.

## PR checklist


fasterius:
While the Conda tests are expected to fail (is there a way to disable Conda for some modules?), I'm not sure why some of the other tests are failing. Some seem to have run out of memory, while others reach an unexpected EOF when pulling the Docker image. The linting is also failing in a weird manner, saying it can't connect to the Docker image URL, while it clearly works fine for some most tests. Any ideas?

modules/nf-core/spaceranger/count/main.nf
```diff
@@ -31,17 +31,23 @@ process SPACERANGER_COUNT {
     def cytaimage = cytaimage ? "--cytaimage=\"${cytaimage}\"" : ""
     def darkimage = darkimage ? "--darkimage=\"${darkimage}\"" : ""
     def colorizedimage = colorizedimage ? "--colorizedimage=\"${colorizedimage}\"" : ""
+    def slide = meta.slide ? meta.slide : ""
+    def area = meta.area ? meta.area : ""
+    if (slide.matches("visium-(.*)") && area == "" && slidefile == "") {
+        slide_and_area = "--unknown-slide=\"${slide}\""
+    } else {
+        slide_and_area = "--slide=\"${slide}\" --area=\"${area}\""
+    }
```
grst, line 40: Do we need this at all? We could also rely on the pipeline developers to specify `--area` etc. via `task.ext.args`, e.g. 

```nextflow
process {
  withName: SPACERANGER_COUNT {
     task.ext.args = { (meta.slide & meta.area) ? "--area=\"${meta.area}\" --slide=\"${meta.slide}\"" : "--unknown-slide" }
  }
```
}

grst:
> The linting is also failing in a weird manner, saying it can't connect to the Docker image URL, while it clearly works fine for some most tests. Any ideas?

just restarted the checks, let's see

modules/nf-core/spaceranger/count/main.nf
fasterius: Yes, I did think about this, but since this is _specifically_ part of the Space Ranger Count software as an explicit option, I thought it made more sense to have it in the module itself. It's also required to actually specify the version of Visium being run (_e.g._ `--unknown-slide visium-1`), which is also a demand of the software itself. It felt odd to me to write something so software-specific outside of the software module.

modules/nf-core/spaceranger/count/main.nf
```diff
@@ -31,17 +31,23 @@ process SPACERANGER_COUNT {
     def cytaimage = cytaimage ? "--cytaimage=\"${cytaimage}\"" : ""
     def darkimage = darkimage ? "--darkimage=\"${darkimage}\"" : ""
     def colorizedimage = colorizedimage ? "--colorizedimage=\"${colorizedimage}\"" : ""
+    def slide = meta.slide ? meta.slide : ""
+    def area = meta.area ? meta.area : ""
+    if (slide.matches("visium-(.*)") && area == "" && slidefile == "") {
+        slide_and_area = "--unknown-slide=\"${slide}\""
```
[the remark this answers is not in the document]
fasterius: It's not a boolean, it takes `visium-[1,2,2-large,hd]` as argument.

grst (approved):
Your comments make sense, I'm happy with the changes then.

fasterius:
Great! Any ideas regarding the failing tests/linting? I'm still seeing the same errors.

grst:
unfortunately not... best to ask around on slack I guess.

cavenel:
It looks like the failing linting is because of the quay.io container name (). It is hosted as `nf-core/modules/spaceranger` while linting expects just `nf-core/spaceranger`. Note that only three other modules in https://quay.io/organization/nf-core have the `/modules/` path.
Was there a reason to have it in the first place?

The other linting issue is that [main.nf.test.snap](https://github.com/nf-core/modules/blob/master/modules/nf-core/spaceranger/count/tests/main.nf.test.snap) should contain the string `versions` somewhere. I see that other modules just renamed their tests into `versions_testname` to fix this...

fasterius:
Okay, fixed the second of your points by cleaning the snapshots of the tests. 

Linting with the previous `nf-core/spaceranger:3.0.0` rather than `nf-core/modules/spaceranger:d71611e316a8614b` works as well, but obviously we don't want that image. Do I remember correctly if the way to update an nf-core image was to ask you to do it, @maxime?

fasterius:
Okay, linting is now fine, and the Docker image has moved location. Now I think that the problem is that the newer version of Spaceranger cannot be run on Apple Silicon chips anymore, so I can't properly update the snapshots. Looking for an alternative.

cavenel:
I now updated the snapshots. It still complains about a difference in `genes.gtf.gz` for `homo_sapiens - fasta - gtf` in `spaceranger/mkref`. Anybody would know why this differs on GitHub CI and on my local tests?

Also it is strange that it tries to run CI on `spaceranger` and `spaceranger/Dockerfile` as if they were test folders... Any idea why?

mahesh-panchal (changes requested):
Sorry, the guidelines don't allow custom meta fields in modules:
https://nf-co.re/docs/guidelines/components/modules#types-of-meta-fields

modules/nf-core/spaceranger/count/main.nf
fasterius: I'm not really sure how this would be written then, at least if following the guidelines of using `ext.args`, since there would then need to be some large rewrites of both module and pipeline code. Or maybe there's a simple fix I'm not seeing? We're reading entries in the supplied samplesheet and conditionally changing a specific set of spaceranger arguments based on that.

fasterius: I guess just adding `slide` and `area` as new input values? The module already has a lot 😅

mahesh-panchal: If I understand correctly, exactly how @grst proposed above. Although write it more like this:

https://community.seqera.io/t/composing-strings-based-on-multiple-conditions/1832

fasterius: But then this would become a feature of the `spatialvi` pipeline, rather than the `spaceranger` module, which is _where_ the feature actually exists. Wouldn't it make more sense to have all the features of a module exist inside the module?

FloWuenne:
@nf-core-bot fix linting

cavenel:
@FloWuenne any idea on how to proceed here?

FloWuenne:
@nf-core-bot fix linting

cavenel:
@FloWuenne I don't think that fixes the issue with files being run through linting even though they are not module folders (`spacerange` and `spaceranger/Dockerfile`). Still not sure what the best way to proceed here, and it would be good to merge this PR :)

mahesh-panchal:
@mashehu Do you know how to pass this?

mashehu:
On parental leave today. Cc @mirpedrol

grst:
That may have helped: https://github.com/nf-core/modules/pull/7233/commits/b664c066cb0468b08625acd43b11a7288ca3bfa2

The linting error in `bam_split_by_region` seems to be a genuine linting error (not sure why it is being checked though)

cavenel:
> The linting error in `bam_split_by_region`
I don't see that one, but I do see test failing on the snapshots of `SPACERANGER_COUNT (CytAssist_11mm_FFPE_Human_Glioblastoma_2)` now... So frustrating that snapshots change at every nextflow version, I don't really get it.

And the main folder `spaceranger` is still being sent to linting which I don't understand either...

grst:
https://github.com/nf-core/modules/pull/8191

I'd hope that this would fix the issue with picking up the spaceranger module. But not sure if it has unwanted side-effects. Need to wait for @nf-core/a-team to review. 

I'll anyway merge it into this PR for testing.

fasterius:
This was apparently auto-merged from when I turned that on two months ago - didn't expect this PR to have experimental things in it related to linting in general. Should we revert the merge?

grst:
I think it's ok as @mirpedrol approved #8191

FloWuenne:
> @FloWuenne I don't think that fixes the issue with files being run through linting even though they are not module folders (`spacerange` and `spaceranger/Dockerfile`). Still not sure what the best way to proceed here, and it would be good to find a way to merge this PR :)

Yeah sorry, kicked this off, then tried to check how to fix manually late evening yesterday but didn't find a solution and forgot to post here, sorry! 😅

PR #7244, opened 2024-12-18
READY FOR REVIEW: Adapt output declarations

simoxb:
When adapting nf-core code, I realized that some modules declare their output paths using single-quoted strings (e.g. `path('*.txt')`), while others use double-quoted strings (like `path("*.txt")`). When adapting the code, this leads to inconsistencies: 

Within single quoted outputs, process variables (such as `meta.id`) cannot be referenced using the usual mechanism `${meta.id}`. This makes the code significantly harder to adapt, since using this reference mechanism in single-quoted strings leads to a (not very helpful) "Output file not found" error. I also have not found any helpful information on how to debug this problem, so it took quite some time to fix. Changing output declarations to double-quoted strings everywhere would not bring any disadvantages I'm aware of, but just enable better adaptability because the usual reference mechanism for nextflow process variables with `$' could be used in all output declarations.
 
Especially the inconsistent use of both single- and double-quoted strings in the different modules makes no sense. I would therefore propose to change all output path declarations to double-quoted strings.

## PR checklist

- [ x] This comment contains a description of changes (with reason).
- [ x] Remove all TODO statements.
- [ x] Follow the naming conventions.
- [ x] Follow the input/output options guidelines.
- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/abacas/main.nf
+++ b/modules/nf-core/abacas/main.nf
@@ -12,7 +12,7 @@ process ABACAS {
     path  fasta
 
     output:
-    tuple val(meta), path('*.abacas*'), emit: results
+    tuple val(meta), path("*.abacas*"), emit: results
     path "versions.yml"               , emit: versions
 
     when:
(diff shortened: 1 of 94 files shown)


bentsherman:
Requiring double-quoted strings everywhere seems like overkill to me. The Nextflow syntax highlighting makes it clear whether the `${meta.id}` is being interpolated

simoxb:
> Requiring double-quoted strings everywhere seems like overkill to me. The Nextflow syntax highlighting makes it clear whether the `${meta.id}` is being interpolated

Fair enough, over-regulation is also a bad thing. At the same time, in nf-core a lot of things are standardized, which mostly aims to improve the user experience and simplify using or adapting the code. Just think it would not hurt too much to set this as standard in nf-core modules, since not everybody is writing Nextflow code using the syntax highlighting, and also is not aware of the different behaviour of different quotes in output path declarations. 

There has also been a bit of a discussion on the nf-core slack about this, btw.

PR #7258, opened 2024-12-20
modify custom/filterdifferentialtable to allow >=/<=

suzannejin:
This modification allows setting ">=" or "<=" comparisons through ext.args

## PR checklist

Closes #XXX 

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:


pinin4fjords (changes requested):
The interface is a little confusing here. But in any case, you're effectively making the cardinality a required input here, which means it should go in through a channel.

I suggest we change the inputs from:

```    
    tuple val(meta), path(input_file)
    val(logFC_column)
    val(FC_threshold)
    val(padj_column)
    val(padj_threshold)
```

To:

```
    tuple val(meta), path(input_file)
    val(logFC_column)
    val(FC_threshold)
    val(FC_cardinality)
    val(padj_column)
    val(padj_threshold)
    val(padj_cardinality)
```

(removing the args thing altogether)

OR, maybe we also want to bundle these like:

```
    tuple val(meta), path(input_file)
    tuple val(logFC_column), val(FC_threshold), val(FC_cardinality)
    tuple val(padj_column), val(padj_threshold), val(padj_cardinality)
```

There should then be a check for a valid cardinality ('-', '<', '>', '<=', or '>=').

suzannejin:
@pinin4fjords Hello!! Thank you for the suggestions! Now the cardinal comparison operators are explicitly given as input :)

subworkflows/nf-core/abundance_differential_filter/main.nf
```diff
@@ -142,24 +142,28 @@ workflow ABUNDANCE_DIFFERENTIAL_FILTER {
         .join(inputs.filter_params)
         .multiMap { meta, results, filter_meta ->
             def method_params = [
-                'deseq2': [fc_column: 'log2FoldChange', padj_column: 'padj'],
-                'limma' : [fc_column: 'logFC', padj_column: 'adj.P.Val'],
-                'propd' : [fc_column: 'lfc', padj_column: 'weighted_connectivity']
+                'deseq2': [fc_column: 'log2FoldChange', fc_cardinality: '>=', padj_column: 'padj', padj_cardinality: '<='],
+                'limma' : [fc_column: 'logFC', fc_cardinality: '>=', padj_column: 'adj.P.Val', padj_cardinality: '<='],
+                'propd' : [fc_column: 'lfc', fc_cardinality: '>=', padj_column: 'weighted_connectivity', padj_cardinality: '>=']
```
pinin4fjords, line 147: Is it intentional to have '>=' for the propd padj cardinality?

subworkflows/nf-core/abundance_differential_filter/tests/main.nf.test.snap
```diff
@@ -1108,9 +1108,9 @@
                         "blocking": "sample_number",
                         "method": "propd",
                         "fc_threshold": 1.5,
-                        "padj_threshold": 50
+                        "padj_threshold": 100
```
pinin4fjords, line 1111: I should have asked this before- but why isn't the p value threshold something less than 1?

subworkflows/nf-core/abundance_differential_filter/main.nf
suzannejin: yes, because we use a different estimate that is not p-value
Maybe we should call stat_column instead of padj_column? There are also some other DE methods that don't use p-value, isn't it.

subworkflows/nf-core/abundance_differential_filter/tests/main.nf.test.snap
pinin4fjords: OK, now I understand that this isn't a p value at all. Can we maybe leave it at the same value though, so we know the md5sum isn't affected by the code change?

subworkflows/nf-core/abundance_differential_filter/main.nf
pinin4fjords: Yes, that sounds sensible, or others will be as confused as me ;-)

subworkflows/nf-core/abundance_differential_filter/tests/main.nf.test.snap
suzannejin: since the operator is ">=" instead of "<=" now (as it should be for weighted connectivity), it will not keep the same md5sum anyway even keeping threshold = 50

pinin4fjords: Ahh, gotcha

pinin4fjords (approved):
Looking good, thanks for addressing feedback

PR #7261, opened 2024-12-23
PINTS Updates

edmundmiller:
- **feat(pints): Allow running on different chr numbers**
- **fix(pints): Only run on 1 bam at a time**
- **fix(pints): Lock down the python version**
- **fix(pints): Actually we shouldn't care about the python version**
- **chore(deps): Manually bump pypints**


edmundmiller:
Closing — 14 months old, conflicts, no engagement.

PR #7263, opened 2024-12-26
Add new module `decoupler`

atrigila:
This PR adds the module `decoupler/progeny` as required in https://github.com/nf-core/differentialabundance/issues/350

## PR checklist

Closes #XXX 

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:


atrigila:
Issues:
- omnipath is not being detected in singularity container. 
- docker: `cannot touch '.command.trace': Permission denied`

atrigila:
If we want to use the R bioconductor package, it also has compatibility problems with the available decoupler and omnipath versions (https://wave.seqera.io/view/builds/bd-d3c1e824e13fcf24_2)

modules/nf-core/decoupler/progeny/main.nf
```diff
@@ -0,0 +1,119 @@
...
+    # Force HOME to current working directory to avoid permission issues
+    os.environ["HOME"] = os.getcwd()
+
+    # Set cache and config dirs to a writable location
+    os.environ["OMNIPATH_CACHE_DIR"] = "./tmp"
+    os.environ["XDG_CONFIG_HOME"] = "./tmp"
+    os.makedirs("./tmp", exist_ok=True)
+
+    os.environ["MPLCONFIGDIR"] = "./tmp"
+    os.environ["NUMBA_CACHE_DIR"] = "./tmp"
```
maxulysse, line 42: I'm a bit unsure about this part

matthdsm: Don't really see an issue here. Very similar to what we need to do to get [gatk](https://github.com/nf-core/modules/blob/master/modules/nf-core/gatk4/postprocessgermlinecnvcalls/main.nf#L36) to work

matthdsm: or any other module that has an option to set a TMP_DIR

mahesh-panchal: Don't forget `input:` can also be type `env`. This could be a single tuple of `env` values 
```nextflow
workflow {
    TASK(['$PWD','./tmp']).view()
}

process TASK {
    input:
    tuple env('TASK_HOME'), env('TMP_DIR')

    script:
    """
    echo "TASK_HOME=\$TASK_HOME"
    echo "TMP_DIR=\$TMP_DIR"
    """

    output:
    stdout
}
```

mahesh-panchal: Another option is to have a config file that you source with a single command to set the environment variables

atrigila: Here are some other examples using the same approach:
https://github.com/nf-core/modules/blob/4d95826db394fc94c9e9ea11170bbe19e0d849ca/modules/nf-core/doubletdetection/templates/doubletdetection.py#L6-L7

https://github.com/nf-core/modules/blob/4d95826db394fc94c9e9ea11170bbe19e0d849ca/modules/nf-core/scimap/mcmicro/main.nf#L28-L29

https://github.com/nf-core/modules/blob/4d95826db394fc94c9e9ea11170bbe19e0d849ca/modules/nf-core/sourcepredict/main.nf#L28-L29

Would you all be agree with keeping it as-is? 

Additionally, I couldn't find any example in nf-core using the `env()` approach. I did attempt, but wasn't able to make it work. This is what I got when trying to use $PWD in the nf-test:
`groovy.lang.MissingPropertyException: No such property: PWD for class: com.askimed.nf.test.lang.process.ProcessContext`

mahesh-panchal: Can you post what you did please that generated the error? My example above works, but this error seems something to do with nf-test, but I don't understand why there would be an issue there.

Also I'm mostly posting alternatives. At the end of the day, the code needs to be somewhere, and setting the variable in the script section is probably the most contained way. Using `env` for something static I think reduces readability/increases debug time since you need to switch between files. 

My vote is keep it in the script block.

atrigila: Yes, your example worked perfectly so I followed it and I tried to replicate it by adding this:
```
    input:
    tuple val(meta), path(mat)
    tuple env('HOME'), env('OMNIPATH_CACHE_DIR')
```
to the `main.nf` and then added to the `main.nf.test`:
```
                input[0] = [
                    [ id: 'test' ],
                    file("https://github.com/nf-core/test-datasets/raw/modules/data/genomics/homo_sapiens/scrnaseq/h5ad/pbmc3k.h5ad", checkIfExists: true),
                    ['$PWD','./tmp']]
                ]
```

In the script, I temporarily commented the:
```
    #os.environ["HOME"] = os.getcwd()
    #os.environ["OMNIPATH_CACHE_DIR"] = "./tmp"
```
and I got the above error when running the test.

mahesh-panchal:
```
tuple env('HOME'), env('OMNIPATH_CACHE_DIR')
```
Is a second input channel so your test input should be:
```
input[0] = [
    [ id: 'test' ],
    file("https://github.com/nf-core/test-datasets/raw/modules/data/genomics/homo_sapiens/scrnaseq/h5ad/pbmc3k.h5ad", checkIfExists: true),
]
input[1] = ['$PWD','./tmp']
```

atrigila: oops, sorry yes you are right. still got the same error though: `groovy.lang.MissingPropertyException: No such property: PWD for class: com.askimed.nf.test.lang.process.ProcessContext`

mahesh-panchal: Try escaping the `$` since it's in a string
```
input[1] = ['\$PWD','./tmp']
```

mahesh-panchal: Interesting. Now I'm curios why it didn't have the intended behaviour, but I don't want to waste your time.

atrigila: Could anybody, if you all agree, give a review/approve and merge, then? Thank you :)

modules/nf-core/decoupler/progeny/meta.yml
```diff
@@ -0,0 +1,129 @@
...
+            Groovy Map containing sample information
+            e.g. [ id:‘test’, single_end ]
+          pattern: "*pvals__decoupler.tsv"
+      - progeny_mlm_pvals_decoupler.tsv:
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. [ id:‘test’, single_end ]
+          pattern: "*pvals__decoupler.tsv"
+      - s:
```
mahesh-panchal, line 63: What's this?

modules/nf-core/decoupler/progeny/meta.yml
```diff
@@ -0,0 +1,129 @@
...
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. [ id:‘test’, single_end ]
+          pattern: "*estimate__decoupler.tsv"
+      - progeny_mlm_estimate_decoupler.tsv:
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. [ id:‘test’, single_end ]
```
mahesh-panchal, line 48: Could you correct this please

modules/nf-core/decoupler/progeny/meta.yml
```diff
@@ -0,0 +1,129 @@
...
+          Should be in HDF5 format (`*.h5ad`).
+        pattern: "*.h5ad"
+output:
+  - dc_estimate:
+      - meta:
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. [ id:‘test’, single_end ]
+          pattern: "*estimate__decoupler.tsv"
```
mahesh-panchal, line 43: This needs removing
```suggestion
```

modules/nf-core/decoupler/progeny/tests/nextflow.config
```diff
@@ -0,0 +1,6 @@
+process {
+    withName: "DECOUPLER_PROGENY" {
+        ext.prefix = 'test'
+        ext.args = { """{"organism": "human", "top_genes": 1000}""" }
```
mahesh-panchal, line 4: Could you supply these through the test, as described in the [guidelines](https://nf-co.re/docs/tutorials/tests_and_test_data/nf-test_writing_tests#nf-test-guidelines-for-a-simple-un-chained-module) please?

mahesh-panchal (changes requested):
Can you update the meta.yml please, in particular the output descriptions.

modules/nf-core/decoupler/progeny/main.nf
```diff
@@ -0,0 +1,119 @@
...
+    tuple val(meta), path("matrixplot_decoupler.png")           , emit: dc_matrixplot
+    path("versions.yml"), emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ? new groovy.json.JsonSlurper().parseText(task.ext.args) : [:]
+    def organism = args.organism ?: 'human'
+    def top_genes = args.top_genes ?: 1000
```
mahesh-panchal, line 28: There are no guidelines on this. Does this come under the same realm as custom meta fields. i.e. should these be `input:` ?

modules/nf-core/decoupler/progeny/meta.yml
```diff
@@ -0,0 +1,112 @@
...
+      - figures/*violin.png:
+          type: file
+          pattern: "figures/*violin.png"
+  - dc_matrixplot:
+      - meta:
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. [ id:‘test’, single_end ]
+          pattern: "matrixplot_decoupler.png"
```
mahesh-panchal, line 98:
```suggestion
```

modules/nf-core/decoupler/progeny/meta.yml
```diff
@@ -0,0 +1,112 @@
...
+          type: file
+          description: |
+            UMAP plot
+          pattern: "figures/*umap.png"
+  - dc_violin:
+      - meta:
+          type: map
+          description: |
+            Violin plot
+          pattern: "figures/*violin.png"
```
mahesh-panchal, line 88:
```suggestion
```

modules/nf-core/decoupler/progeny/meta.yml
```diff
@@ -0,0 +1,112 @@
...
+          description: |
+            Activities in TSV format from the AnnData object
+          pattern: "acts_decoupler.tsv"
+  - dc_umap:
+      - meta:
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. [ id:‘test’, single_end ]
+          pattern: "figures/*umap.png"
```
mahesh-panchal, line 77:
```suggestion
```

modules/nf-core/decoupler/progeny/main.nf
atrigila: I did those as input in my first design :) but then I remembered this one: https://nf-co.re/docs/guidelines/components/modules#non-file-mandatory-command-arguments
and I changed it to `ext.args`.

atrigila: (I prefer them as `input:`, though. It's a lot cleaner)

modules/nf-core/decoupler/progeny/main.nf
```diff
@@ -0,0 +1,119 @@
...
+    """
+    #!/usr/bin/env python3
```
pinin4fjords, line 30: Guidelines currently say that non-trivial scripts should be templates rather than inline code:

mahesh-panchal: Oh, when did that get added?

pinin4fjords: Not that long ago. There was some discussion...

SPPearce:
nf-core tools linting for pip should be addressed in the hackathon: https://github.com/nf-core/tools/issues/3461

grst:
Hi @atrigila, 

I suggest to reduce the scope of the module a bit and make it database-agnostic. I.e. a pure "decoupler" module that can work with any resource rather than only progeny. 

Instead of querying omnipath internally (`dc.get_progeny()`) I suggest the module takes a reference database as a long-form TSV data frame. In differentialabundance, this resource could then be obtained by making a call to the [omnipath web api](https://pypath.omnipathdb.org/webservice.html) or by providing a custom TSV file. 

Sorry, I should have probably stated that a bit earlier.

atrigila:
@grst 
Could you give me an example of how you obtain the same structure TSV of the progeny data with the web api call as is obtained via `dc.get_progeny()`? 
Thank you.

grst:
https://omnipathdb.org/annotations?databases=PROGENy

atrigila:
Test run locally OK but fail in CI.

grst:
seems like unstable snapshots? 

You could try to disable multithreading of the underlying linalg libraries (only for the tests), this sometimes helps getting stable results across systems: 

```
export MKL_NUM_THREADS="1"
export OPENBLAS_NUM_THREADS="1"
export OMP_NUM_THREADS="1"
export NUMBA_NUM_THREADS="1"
```

SPPearce:
Don't know if relevant here, but changes in numerical precision between systems happen so you end up with different values at the last few decimal places if not rounded.

PR #7273, opened 2025-01-07
Add filter step to CI to exclude some tests for some profiles

mashehu:


sateeshperi (approved):

.github/skip_nf_test.yml
SPPearce, line 1: Can we sort this please

sateeshperi: people are going to be adding randomly even if we sort now....so you sure it will help ?

SPPearce: > people are going to be adding randomly even if we sort now....so you sure it will help ?

I mean, that depends on whether we require them to add it in order...

mashehu: fixed

maxulysse (approved):

mashehu:
> I propose we lock the workflow files down to code owners to prevent any wild modifications without review from infrastructure.

You mean all of your modifications? 😜

maxulysse:
> > I propose we lock the workflow files down to code owners to prevent any wild modifications without review from infrastructure.
> 
> You mean all of your modifications? 😜

:fire:

mashehu:
>I propose we lock the workflow files down to code owners

as far as I can see, that is already the case or I don't know about that specific setting

PR #7296, opened 2025-01-13
Swap transdecoder predict to nf-test

SPPearce:
Swapping transdecoder/predict to nf-test.
There is some funky behaviour here with the prefix.
When I tried using the `-O $prefix` as it was previously, the tool tries to look inside a folder called `prefix` to for the transdecoder/longorf output. So I removed that.


nvnieuwk (approved):

modules/nf-core/transdecoder/predict/main.nf
```diff
@@ -5,7 +5,7 @@ process TRANSDECODER_PREDICT {
     conda "${moduleDir}/environment.yml"
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
         'https://depot.galaxyproject.org/singularity/transdecoder:5.5.0--pl5262hdfd78af_4':
```
charles-plessy, line 7: Your Singularity image is still at version 5.5.0, causing the tests to fail.

SPPearce: Aah, thanks!

SPPearce, line 7:
```suggestion
        'https://depot.galaxyproject.org/singularity/transdecoder:5.7.1--pl5321hdfd78af_0':
```

PR #7302, opened 2025-01-14
update NORM meta in abundance_differential_filter

suzannejin:
Just a small change, to use a meta for `deseq2_norm` and `limma_norm` that does not contain contrast info (since they are irrelevant here).

## PR checklist

Closes #XXX 

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:


subworkflows/nf-core/abundance_differential_filter/main.nf
```diff
@@ -49,29 +52,17 @@ workflow ABUNDANCE_DIFFERENTIAL_FILTER {
...
-    // LIMMA_NORM directly. It internally runs normalization + DE analysis.
-
+    // NOTE that we only need a normalised matrix from one contrast. The reason
+    // we don't simply use the first output from DIFFERENTIAL modules is that
+    // depending on the contrast setting etc, these modules may subset matrices,
+    // hence not returning the full normalized matrix as NORM modules would do.
     LIMMA_NORM(
-        norm_inputs.contrasts.filter{it[0].method == 'limma'}.first(),
-        norm_inputs.samples_and_matrix.filter{it[0].method == 'limma'}
+        inputs.contrasts_for_norm.filter{it[0].method == 'limma'}.unique(),
```
pinin4fjords, line 64: Surely this will have LIMMA_NORM run for EVERY contrast?

subworkflows/nf-core/abundance_differential_filter/main.nf
```diff
@@ -49,29 +52,17 @@ workflow ABUNDANCE_DIFFERENTIAL_FILTER {
         .combine(ch_contrasts)
         .multiMap(criteria)
 
-    // We only need a normalised matrix from one contrast. The reason we don't
-    // simply use the first output from DIFFERENTIAL modules is that depending
-    // on the contrast setting etc, these modules may subset matrices, hence
-    // not returning the full normalized matrix as NORM modules would do.
-    norm_inputs = ch_input
```
pinin4fjords, line 56: The reason it was done this way is to apply only the first contrast to each combination of count matrix etc from `ch_input`. So unless I'm missing something you've broken it.

subworkflows/nf-core/abundance_differential_filter/main.nf
```diff
@@ -40,6 +40,9 @@ workflow ABUNDANCE_DIFFERENTIAL_FILTER {
         filter_params:
             meta_map = mergeMaps(meta_contrasts, meta_input) + [ 'method': analysis_method ]
             [meta_map, [ 'fc_threshold': fc_threshold, 'stat_threshold': stat_threshold ]]
+        contrasts_for_norm:
```
pinin4fjords, line 43: Just this change would be fine, used in norm_inputs

pinin4fjords (changes requested):
Unless I'm missing something (always possible), I think what you're doing here is making the _NORM modules run for EVERY combination of contrasts and matrix etc, rather than just the first contrast.

subworkflows/nf-core/abundance_differential_filter/main.nf
suzannejin: You are right! I put norm_inputs back, now it should be fine, it should only compute NORM as many times as unique meta_input have. 
The reason I use `unique()` instead of `first()` is to make sure that the ch_inputs coming with different meta (eg. args) will not be ignored.

pinin4fjords: Actually, I think we should just lose `.first()`, it was a hangover from before our iterations and I should have removed it before. Don't think we need `unique` either. The whole idea of the `.combine()s` is to generate all the combinations we need.

suzannejin: Now the code should be fine! I also added `def meta` inside the multimap criteria, because otherwise the two iterations sometimes mix up.

pinin4fjords (approved):
Looking good

PR #7318, opened 2025-01-16
Added 2FAST2Q module

afombravo:


modules/nf-core/fast2q/tests/test.nf
SPPearce, line 1: This is not a nf-test test file, did you use the nf-core tools to create the module?

SPPearce (commented):
Is there a reason it isn't called 2fastq? Did something fall over because of the number at the start?

modules/nf-core/fast2q/tests/test.nf
afombravo: I am remaking the test file according to regulations!

afombravo:
The reason the I am "filling" it as fast2q is because the PyPI module is called fast2q (numbers at the start are a no go naming wise), the same for nextflow processes i find out? So, despite the tool being published as 2FAST2Q, internaly I am keeping it as fast2q in line with numenclature.

afombravo:
This is apparently failing on the tests. @SPPearce you are more experienced. Could you provide some insight to what I need to change, please? Cheers!

modules/nf-core/fast2q/tests/main.nf.test
```diff
@@ -0,0 +1,71 @@
+// nf-core modules test 2FAST2Q
+
+config './nextflow.config'
+
```
SPPearce, line 4:
```suggestion
```

modules/nf-core/fast2q/tests/main.nf.test
```diff
@@ -0,0 +1,71 @@
...
+nextflow_process {
+
+    name "Test Process 2FAST2Q"
+    script "../main.nf"
+    process "FAST2Q"
+
+    tag "modules"
+    tag "modules_nfcore"
+    tag "fast2q"
+
```
SPPearce, line 14:
```suggestion
config './nextflow.config'

```

modules/nf-core/fast2q/tests/main.nf.test
```diff
@@ -0,0 +1,71 @@
...
+    test("FASTQ file with optional guides") {
+
+        when {
+
+            params {
+                module_args = '--g ' + params.modules_testdata_base_path + '/brunello_target_sequence.txt'
```
SPPearce, line 20: You can't do this, if the file is required (for any functionality) then it needs to be staged as an input to the module.

afombravo: it is not always required.

SPPearce: Even if not always required, you can't point to files that aren't in listed as input channels, it can be an optional channel.

SPPearce: (basically this path works on a local/hpc setup, but wouldn't work on a cloud provider)

afombravo:
@SPPearce The test errors I am getting now are due to some issue loading the right data. I am trying to use 2 files from here: https://github.com/nf-core/test-datasets/tree/crisprseq/testdata but the errors I am getting from the 2FAST2Q are related with the the files not being correct (either the wrong extention, or the wrong format). I know the files are right and the programs works with them "outside" nextflow, so, could it be some issue loading the right data from modules_testdata_base_path on nextflow.config?

modules/nf-core/fast2q/main.nf
```diff
@@ -0,0 +1,38 @@
...
+
+    script:
+    def args            = task.ext.args ?: ''
+    def input_file      = "--s ${fastq}" ?: ''
+    def library_file    = (library instanceof Path && library.exists()) ? "--g ${library}" : ''
+
+    """
+    mkdir -p ./2FAST2Q_output
+
+    2fast2q -c --o ./2FAST2Q_output/ $input_file $library_file $args
```
SPPearce, line 30:
```suggestion
    2fast2q 
        -c \\
        --o ./2FAST2Q_output/ \\
        $input_file \\
        $library_file \\
        $args
```

modules/nf-core/fast2q/main.nf
```diff
@@ -0,0 +1,38 @@
...
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args            = task.ext.args ?: ''
+    def input_file      = "--s ${fastq}" ?: ''
+    def library_file    = (library instanceof Path && library.exists()) ? "--g ${library}" : ''
+
+    """
+    mkdir -p ./2FAST2Q_output
```
SPPearce, line 28: Is there a reason to make a subdirectory, as opposed to just making the files inside the working directory?
What files does it output? 
Could it use `prefix` to name them?

modules/nf-core/fast2q/main.nf
```diff
@@ -0,0 +1,38 @@
...
+    output:
+    tuple val(meta), path("./2FAST2Q_output/*.*")   , emit: processed
+    path "versions.yml"                             , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args            = task.ext.args ?: ''
+    def input_file      = "--s ${fastq}" ?: ''
```
SPPearce, line 24: Is this always required? If so, don't need to define this here, just put directly in the code.

afombravo: you mean just avoid naming a new variable for this?

afombravo: i thought it would be cleaner like this. I saw all other modules following the same style

afombravo:
@SPPearce I added local test data under tests and modified the paths accordingly but it is failing to pick it up as the testing env seems to create some temp directory? It gives out the following mistake: ERROR ~ No such file or directory: /home/runner/tests/c183867ba37e5d8b3b8be8fa1ea94571/example.fastq.gz

afombravo:
I directly added the link you mentioned above ( https://raw.githubusercontent.com/nf-core/test-datasets/refs/heads/crisprseq/testdata/SRR8983579.small.fastq.gz ) and reduced the tests.

afombravo:
@SPPearce seems to be ready for one last review and approval. Sorry for having spammed too much!

modules/nf-core/fast2q/tests/nextflow.config
```diff
@@ -0,0 +1,10 @@
+params {
+    test_data_base = 'https://raw.githubusercontent.com/nf-core/test-datasets/modules'
```
SPPearce, line 2:
```suggestion
```

modules/nf-core/fast2q/tests/main.nf.test
```diff
@@ -0,0 +1,103 @@
...
+                    file(params.test_data_base + '/data/genomics/mus_musculus/mageck/yusa_library.csv', checkIfExists: true) // library file
+                ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert path(process.out.count_matrix).exists() } // Check processed count table exists
```
SPPearce, line 98: Is the count table not stable between runs? I'd have thought it should be a fixed output which could be checked with the md5sum.

modules/nf-core/fast2q/tests/main.nf.test
```diff
@@ -0,0 +1,103 @@
...
+                { assert path(process.out.versions).readLines().join("\n").contains("2FAST2Q version") } // Verify version output
```
SPPearce, line 99: The version should just contain the tool name, see [the guidelines](https://nf-co.re/docs/guidelines/components/modules#emission-of-versions).
You can capture the whole thing inside a snapshot using:
`path(process.out.versions[0]).yaml`

I would try for the assertion here (both checking the count_matrix and the versions):
{ assert snapshot(
            process.out,
).match() },
(with correct indentation)

modules/nf-core/fast2q/tests/main.nf.test
```diff
@@ -0,0 +1,103 @@
...
+
+    test("FASTQ file with optional library") {
+
+        when {
+
+            process {
+                """
+                input[0] = [
+                    [ id:'test1', single_end:true ], // meta map
+                    file(params.test_data_base + '/data/genomics/mus_musculus/mageck/ERR376998.small.fastq.gz', checkIfExists: true) // FASTQ file
```
SPPearce, line 85:
```suggestion
                    file(params.modules_testdata_base_path + '/genomics/mus_musculus/mageck/ERR376998.small.fastq.gz', checkIfExists: true) // FASTQ file
```

modules/nf-core/fast2q/tests/main.nf.test
```diff
@@ -0,0 +1,103 @@
...
+                ]
+                input[1] = [
+                    [ id:'test2', single_end:true ], // meta map for second input
+                    file(params.test_data_base + '/data/genomics/mus_musculus/mageck/yusa_library.csv', checkIfExists: true) // library file
```
SPPearce, line 89:
```suggestion
                    file(params.modules_testdata_base_path + '/genomics/mus_musculus/mageck/yusa_library.csv', checkIfExists: true) // library file
```

modules/nf-core/fast2q/tests/main.nf.test
```diff
@@ -0,0 +1,103 @@
...
+
+            params {
+            module_args = '-t'
+                }
+
+            process {
+                """
+                input[0] = [
+                    [ id:'test1', single_end:true ], // meta map
+                    file(params.test_data_base + '/data/genomics/mus_musculus/mageck/ERR376998.small.fastq.gz', checkIfExists: true) // FASTQ file
```
SPPearce, line 57:
```suggestion
                    file(params.modules_testdata_base_path + '/genomics/mus_musculus/mageck/ERR376998.small.fastq.gz', checkIfExists: true) // FASTQ file
```

modules/nf-core/fast2q/tests/main.nf.test
```diff
@@ -0,0 +1,103 @@
...
+                ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert path(process.out.count_matrix).exists() } // Check processed count table exists
+                { assert path(process.out.versions).readLines().join("\n").contains("2FAST2Q version") } // Verify version output
```
SPPearce, line 71: With corect indentation try this:
```suggestion
{ assert snapshot(
process.out,
path(process.out.versions[0]).yaml
).match() },
```

modules/nf-core/fast2q/tests/main.nf.test
afombravo: I included a md5sum check on the test("2FAST2Q self-test"). The other I think it is suffice to leave as your subsequent changes

afombravo:
@SPPearce I made all tests snapshot comparisons and they are working locally. Everything seems ok. Any more comments?

afombravo:
@SPPearce Do you think this is ready for clousure :) ?

SPPearce:
> @SPPearce Do you think this is ready for clousure :) ?

I made some changes (more than I could do easily via comments), see what you think.

afombravo:
> > @SPPearce Do you think this is ready for clousure :) ?
> 
> I made some changes (more than I could do easily via comments), see what you think.

wow you really dug into it... I appreciate it! I will fix the linting and then resubmit!

SPPearce:
> > > @SPPearce Do you think this is ready for clousure :) ?
> > 
> > 
> > I made some changes (more than I could do easily via comments), see what you think.
> 
> wow you really dug into it... I appreciate it! I will fix the linting and then resubmit!

Sorry, I didn't notice you saying that you'd fix the linting; I've just done that.
You might want to update the descriptions in the meta.yml; I didn't know what the figures actually are.

afombravo:
> > > > @SPPearce Do you think this is ready for clousure :) ?
> > > 
> > > 
> > > I made some changes (more than I could do easily via comments), see what you think.
> > 
> > 
> > wow you really dug into it... I appreciate it! I will fix the linting and then resubmit!
> 
> Sorry, I didn't notice you saying that you'd fix the linting; I've just done that. You might want to update the descriptions in the meta.yml; I didn't know what the figures actually are.

No problem. I will do that then!

afombravo:
@SPPearce Thank you so much! Your help was invaluable in this entire process.

SPPearce:
Need to get someone else to review it that isn't me now, asking on the request-review channel on the nf-core slack is generally a good way

modules/nf-core/fast2q/meta.yml
```diff
@@ -0,0 +1,98 @@
+name: fast2q
```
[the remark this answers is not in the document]
kubranarci: okay seen the previous conversation :)

kubranarci (approved):

kubranarci:
left small commends, otherwise LGTM!

modules/nf-core/fast2q/meta.yml
```diff
@@ -0,0 +1,98 @@
...
+      doi: 10.7717/peerj.14041
+      licence: ["GPL-3.0-or-later"]
+      identifier: ""
+
+input:
+  - - meta:
+        type: map
+        description: |
+          Groovy Map containing output name.
+          e.g. [ id:'test']
```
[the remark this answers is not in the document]
afombravo: single-paired ended knowledge is not required for operation so I am leaving it out. The tool will always read on a read-by-read basis to extract features. if paired-ended these need to be merged before input anyway.

PR #7336, opened 2025-01-20
update hifiasm module to support --ul with ONT reads

nschan:
This pull request adds support for `--ul`, the ultralong algorithm in hifiasm that performs assemblies of HiFi and ONT reads.
See discussion with @LouisLeNezet here https://github.com/nf-core/genomeassembler/pull/65#discussion_r1919926243

## PR checklist


modules/nf-core/hifiasm/main.nf
```diff
@@ -32,6 +32,8 @@ process HIFIASM {
     script:
     def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
+    def cmd_reads_ont = reads_ont ? "--ul ${reads_ont}" : ""
```
muffato, line 35: I don't see where `cmd_reads_ont` is used in the file 🤔

nschan: Good point, I used `$reads_ont`, I have updated the script

modules/nf-core/hifiasm/main.nf
```diff
@@ -8,7 +8,7 @@ process HIFIASM {
         'biocontainers/hifiasm:0.24.0--h5ca1c30_0' }"
 
     input:
-    tuple val(meta) , path(reads)
+    tuple val(meta) , path(reads_hifi)        , path(reads_ont)
```
fellen31, line 11: Since regular ONT reads can be used (with --ont, without --ul) since 0.21.0, could we name these something else? ul_reads for example.

nschan: Yes, that is a great point. I will rename them `ul_reads` and change from `cmd_reads_ont` to `ultralong` to make it clearer.

muffato (approved):

muffato:
Thanks @nschan !

Issue #7342, opened 2025-01-22
update module: liftoff


nschan:
I would like to add functionality to the current liftoff to check for gzipped input fastas and automatically gunzip those if necessary. The current liftoff module expects both fastas (reference and that to be lifted onto) in fasta format.
What do you think of this suggestion @GallVp ?

### Is there an existing module for this?

- [x] I have searched for the existing module

### Is there an open PR for this?

- [x] I have searched for existing PRs

### Is there an open issue for this?

- [x] I have searched for existing issues

### Are you going to work on this?

- [x] If I'm planning to work on this module, I added myself to the `Assignees` to facilitate tracking who is working on the module

GallVp:
Thank you @nschan for tagging me. That's a really good idea. Please feel free to tag me for a review.

PR #7351, opened 2025-01-23
liftoff: allow gzipped fasta inputs

nschan:
This PR implements #7342. I have updated the script and meta.yml

### DIFF
--- a/modules/nf-core/liftoff/main.nf
+++ b/modules/nf-core/liftoff/main.nf
@@ -9,7 +9,7 @@ process LIFTOFF {
 
     input:
     tuple val(meta), path(target_fa)
-    path ref_fa,
+    path ref_fa
     path ref_annotation
     path ref_db
 
--- a/modules/nf-core/liftoff/main.nf
+++ b/modules/nf-core/liftoff/main.nf
@@ -9,7 +9,7 @@ process LIFTOFF {
 
     input:
     tuple val(meta), path(target_fa)
-    path ref_fa, name: 'ref_assembly.fa'
+    path ref_fa,
     path ref_annotation
     path ref_db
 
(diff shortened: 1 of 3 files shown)


nschan:
The tests fail because both input and output are named `genome.fasta`. Previously, this was probably solved by staging in the target as `assembly.fasta`, but I think the target should be staged in with the original filename, since this is how I check if it is gzipped. I am not very into nf-test and I am not sure what the best way to solve this would be?

modules/nf-core/liftoff/main.nf
```diff
@@ -28,15 +28,31 @@ process LIFTOFF {
     def arg_db  = ref_db            ?   "-db $ref_db"           : ''
     prefix      = task.ext.prefix   ?:  "${meta.id}"
     """
+    if [[ ${target_fa} == *.gz ]]; then
+        zcat ${target_fa} > target.fasta
+    fi
+
+    if [[ ${target_fa} == *.fa || ${target_fa} == *.fasta ]]; then
```
GallVp, line 35: Instead of restricting the extensions, can we use something like `"$target_fa".endWith('.gz')` and  `! "$target_fa".endWith('.gz')`. `fsa` and `fas` are very common extension.

GallVp:
Thank you @nschan

modules/nf-core/liftoff/main.nf
nschan: Sure.

modules/nf-core/liftoff/main.nf
```diff
@@ -28,15 +28,31 @@ process LIFTOFF {
...
     liftoff \\
         $arg_g \\
         $arg_db \\
         -p $task.cpus \\
         -o "${prefix}.gff3" \\
         -u "${prefix}.unmapped.txt" \\
         $args \\
-        $target_fa \\
-        ref_assembly.fa
+        target.fasta \\
```
GallVp, line 54: Excellent!

PR #7370, opened 2025-01-26
Make erroring wave builds fail CI

edmundmiller:
So we can know when the arm builds are failing. Reporting hook to come later.

### DIFF
--- a/.github/workflows/wave.yml
+++ b/.github/workflows/wave.yml
@@ -57,7 +57,7 @@ jobs:
     timeout-minutes: 60
     strategy:
       fail-fast: false
-      max-parallel: 4
+      max-parallel: 2
       matrix:
         files: "${{ fromJson(needs.generate-matrix.outputs.conda-matrix) }}"
         profile: [docker, singularity]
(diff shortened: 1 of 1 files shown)


maxulysse (approved):

.github/workflows/wave.yml
```diff
@@ -57,7 +57,7 @@ jobs:
     timeout-minutes: 60
     strategy:
       fail-fast: false
-      max-parallel: 4
+      max-parallel: 2
```
mashehu, line 60: why not 4?

mashehu (approved):

ewels:
Looks like this logic is the wrong way around, meant that _only_ PRs coming from branches on the nf-core repo were being run:

https://github.com/nf-core/modules/blob/326bde891db0956f0148ed7e21938360678d3e75/.github/workflows/wave.yml#L25

@edmundmiller I'm not sure that we want to fail builds yet - are we confident enough that it won't block new modules?

edmundmiller:
> Looks like this logic is the wrong way around, meant that only PRs coming from branches on the nf-core repo were being run:

No that's the right way around because it fails if you try to make a PR from anywhere but the nf-core modules repo. So it should skip.

Ideally, the bumps will come only from the renovate commits. Users will still have to make the initial containers.

> @edmundmiller I'm not sure that we want to fail builds yet - are we confident enough that it won't block new modules?

@ewels This PR was because you asked for it 🤣 I think it's fine; it's not required, so it won't block the PR.

.github/workflows/wave.yml
edmundmiller: Why set a limit at all? Idk, I think it was to not soak up a ton of runners.

mashehu:
People get confused about failing tests, even if the tests are not required. Better to skip it/succeed with warning.

edmundmiller:
> People get confused about failing tests, even if the tests are not required. Better to skip it/succeed with warning.

Sounds good! Closing this then, and we'll link to it if anyone asks and can revisit it later.

PR #7375, opened 2025-01-27
Update cooler/makebins

SPPearce:
I noticed that the tag on this module had a spurious `}`, so I removed it.
Also added a stub + test at the same time.


charles-plessy (approved):
Good catch! Looks good to me.

PR #7376, opened 2025-01-27
catpack/contigs fix a typo in the option for the diamond table

ochkalova:
## PR checklist

Hello, 

Thank you for the [recent update for this module](https://github.com/nf-core/modules/commit/3486e0ae4ae7ff09d64e807e51d9581f782e605e) that added missing input options. It was really helpful for me. 

However I've noticed that there is a typo in a flag for diamond alignment file. 
The correct flag is 
```
  -a , --diamond_alignment
                        Path to alignment table. If supplied, the alignment
                        step is skipped and classification is carried out
                        directly. A predicted proteins fasta file should also
                        be supplied with argument [-p / --proteins].
```
The flag used is a flag for providing the database:
` def premade_table = diamond_table ? "-d ${diamond_table}" : ''`
Documentation for CAT_pack: https://github.com/MGXlab/CAT_pack?tab=readme-ov-file#taxonomic-annotation-of-contigs-or-mags-with-cat-and-bat

Thank you for your time.

### DIFF
--- a/modules/nf-core/catpack/contigs/main.nf
+++ b/modules/nf-core/catpack/contigs/main.nf
@@ -30,7 +30,7 @@ process CATPACK_CONTIGS {
     def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
     def premade_proteins = proteins ? "-p ${proteins}" : ''
-    def premade_table = diamond_table ? "-d ${diamond_table}" : ''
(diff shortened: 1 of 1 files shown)


mashehu (approved):

mashehu:
Would be even better to use the long parameter names, i.e. `--diamond_alignment` in this case. If you have a minute to do it for the whole command.

ochkalova:
@mashehu good idea, thanks!

PR #7365, opened 2025-01-25
Added nf-test for krona/ktimporttaxonomy

GallVp:
Added nf-test for krona/ktimporttaxonomy with [pytest2nf-test](https://github.com/GallVp/pytest2nf-test)

PR #7378, opened 2025-01-27
Pytest with nftest

SPPearce:
In #7365 @GallVp was converting pytest to nftest for `krona/ktimporttaxonomy`, but it turns out it already had pytest files.

So I spent some time to check issues with the pytest files and found:

- Entries in `pytest_modules.yml` that had no corresponding modules or subworkflow file at all (e.g. `gatk4/selectsamples`)
- Entries in `pytest_modules.yml` that no longer had any pytest files (e.g. `angsd/contamination`)
- Modules with pytest and nf-test files, sometimes present in `pytest_modules.yml` but sometimes not. 

In total this removes 18 entries from `pytest_modules.yml`, plus some more subfolders that weren't listed in there.


GallVp:
Thank you @SPPearce 

I was about to do something similar and then I saw the email notification for this PR.

GallVp (approved):

PR #7381, opened 2025-01-28
Diamond/blastp updates

### LINKED ISSUE #7350: update module: diamond/blastp

I have some issues with the current module, some easy to solve, some that need a discussion since they potentially change the interface of the module.

1. I need to use the `--include-lineage` parameter supported by Diamond 2.10 but not the current 2.8. Easy to fix of course.
2. Diamond supports reading gzipped files, so the unzipping of gzipped input seems unnecessary and a waste of space.
3. Diamond also supports gzipping output files via `--compress 1`. Seems like a good feature to support, but is made difficult by the current third parameter being the file suffix which in turn decides the output format. My suggestion is to change the third param to the numerical output format and set the suffix from this instead. If `$args` contains `--compress 1`, a `.gz` can be added to the file suffix (except for `.daa`). (Overall, I also find the use of a suffix to decide the output format a bit too implicit for my taste. I definitely consider the blast output format a `.tsv` and would use that if I didn't know about the taxonomy (102) format.)
4. In my experience, Diamond uses cpus quite efficiently and tends to require a lot of memory, so I'd suggest labelling the process `process_high` instead of medium.
5. You're using `find` to get the path of the input `.dmnd` but I don't understand why since this is param two to the module.
6. I think it would be good to include `${meta2.id}" in the default output file name as that's likely (it is in my case at least) the name of the database.

Update: I realize I need to match output channels from this module with the database channel (`meta2`) in my pipeline before calling the next module, i.e. join the output of this module with a channel that is indexed with database names (`meta2.id`). As I see this, I can either return both meta objects or add a field (e.g. `db`) containing `meta2.id` to the `meta` object before returning it. An alternative is perhaps to modify `meta.id` to contain the database name before calling this module.

I need a module like my description above for #metatdenovo, so I've made a local while we discuss. At the moment it looks like the below but hasn't been properly tested yet (i.e. WIP). I'm happy to contribute this to nf-core/modules if and when we agree.

```
    tag "$meta.id"
    label 'process_high'

        'https://depot.galaxyproject.org/singularity/diamond:2.1.10--h5ca1c30_3':
        'biocontainers/diamond:2.1.10--h5ca1c30_3' }

   input:
    tuple val(meta) , path(fasta)
    tuple val(meta2), path(db)
    val outfmt
    val blast_columns

    output:
    tuple val(meta), path('*.blast*'), optional: true, emit: blast
    tuple val(meta), path('*.xml*')  , optional: true, emit: xml 
    tuple val(meta), path('*.txt*')  , optional: true, emit: txt 
    tuple val(meta), path('*.daa')   , optional: true, emit: daa 
    tuple val(meta), path('*.sam*')  , optional: true, emit: sam 
    tuple val(meta), path('*.tsv*')  , optional: true, emit: tsv 
    tuple val(meta), path('*.paf*')  , optional: true, emit: paf 
    path "versions.yml"              , emit: versions

    when:

    script:
    def prefix = task.ext.prefix ?: "${meta.id}.${meta2.id}"
    def columns = blast_columns ? "${blast_columns}" : ''
    switch ( outfmt ) { 
        case 0:   out_ext = "blast"; break
        case 5:   out_ext = "xml";   break
        case 6:   out_ext = "txt";   break
        case 100: out_ext = "daa";   break
        case 101: out_ext = "sam";   break
        case 102: out_ext = "tsv";   break
        case 103: out_ext = "paf";   break
        default:
            outfmt = 6;
            out_ext = 'txt';
            log.warn("Unknown output file format provided (${out_ext}): selecting DIAMOND default of tabular BLAST output (txt)");
            break
    }   
    if ( args =~ /--compress\s+1/ ) out_ext += '.gz'
    """ 
    diamond \\
        blastp \\
        --threads ${task.cpus} \\
        --db $db \\
        --query ${fasta} \\
        --outfmt ${outfmt} ${columns} \\
        ${args} \\
        --out ${prefix}.${out_ext}

    "${task.process}":
        diamond: \$(diamond --version 2>&1 | tail -n 1 | sed 's/^diamond version //')
    END_VERSIONS
    """ 
```

vagkaratzas:
## PR checklist

Closes #7350 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


erikrikarddaniel (approved):
Looks great, except for our point of disagreement! ;-)

vagkaratzas:
@erikrikarddaniel Shall we merge this version? Or wait for a verdict on -outfmt format?

erikrikarddaniel:
I think it's up to you and me, at least if nobody is actually using it. (I'm for the time being using my own in #metatdenovo.) I'll let you decide, but to reiterate my arguments for having a numerical output specification rather than a suffix:

1. Users of the module are likely users of the tool, so they will be familiar with the different choices for `--outfmt`.
2. The output format is both a format, e.g. tsv, and content, and at least two output formats are actually the same format, i.e. both blast and taxonomy are tsv.
3. If the param is called `ext` or similar a user of the module might (I did) guess that one can specify compression by adding a `.gz` suffix to the format.
4. IMO, it's more explicit to ask for the tool's param value and get the module to return a file with a suffix it chose, than the reverse.
5. I know everything is (should be) specified in the documentation -- but who reads that?
6. I suppose there are other modules with similar problems and I don't know if there's a general recommendation.

vagkaratzas:
> I think it's up to you and me, at least if nobody is actually using it. (I'm for the time being using my own in #metatdenovo.) I'll let you decide, but to reiterate my arguments for having a numerical output specification rather than a suffix:
> 
> 1. Users of the module are likely users of the tool, so they will be familiar with the different choices for `--outfmt`.
> 2. The output format is both a format, e.g. tsv, and content, and at least two output formats are actually the same format, i.e. both blast and taxonomy are tsv.
> 3. If the param is called `ext` or similar a user of the module might (I did) guess that one can specify compression by adding a `.gz` suffix to the format.
> 4. IMO, it's more explicit to ask for the tool's param value and get the module to return a file with a suffix it chose, than the reverse.
> 5. I know everything is (should be) specified in the documentation -- but who reads that?
> 6. I suppose there are other modules with similar problems and I don't know if there's a general recommendation.

Ok, I'm sold. Have a look an feel free to merge if tests pass and you are happy with the changes!

modules/nf-core/diamond/blastp/main.nf
```diff
@@ -1,63 +1,66 @@
...
+    } else if (outfmt == 101) {
+        out_ext = "sam"
+    } else if (outfmt == 102) {
+        out_ext = "tsv"
+    } else if (outfmt == 103) {
+        out_ext = "paf"
+    } else {
+        outfmt = '6'
+        out_ext = 'txt'
+        log.warn("Unknown output file format provided (${out_ext}): selecting DIAMOND default of tabular BLAST output (txt)")
```
erikrikarddaniel, line 53:
```suggestion
    } else {
        log.warn("Unknown output file format provided (${outfmt}): selecting DIAMOND default of tabular BLAST output (txt)")
        outfmt = 6
        out_ext = 'txt'
```

modules/nf-core/diamond/blastp/main.nf
```diff
@@ -69,23 +72,32 @@ process DIAMOND_BLASTP {
...
```
erikrikarddaniel, line 96:
```suggestion
    } else {
        log.warn("Unknown output file format provided (${outfmt}): selecting DIAMOND default of tabular BLAST output (txt)")
        outfmt = 6
        out_ext = 'txt'
```

erikrikarddaniel (approved):
OK, I made two suggestions. I'll accept them and merge.

erikrikarddaniel:
@vagkaratzas I'm stuck with failing tests due to the runner problems recently. I'll just wait until I need to merge upstream again, and hope they pass then. (I don't know of another way of triggering them... :-S)

PR #7416, opened 2025-01-31
add another clean-up step to GHA

mirpedrol:


mashehu (commented):
I am not sure this will help, because there seems to be a problem with left over files from a different run, which might raise the same permission error

mashehu:
let's close for now. if it comes up again, we might reopen

PR #7427, opened 2025-02-04
Allow `last/maf-convert` to output multiple files in different formats.

charles-plessy:
The input channel conveying output format information becomes a list of strings instead of just a string.  The module loops on the formats and runs `maf-convert` on each of them.  This will simplify the structure of pipelines that need conversions to more than one format.

Silently, the module still accepts the input of single format as a simple string, for backwards compatibility.

## PR checklist

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/last/mafconvert/main.nf
+++ b/modules/nf-core/last/mafconvert/main.nf
@@ -9,7 +9,7 @@ process LAST_MAFCONVERT {
 
     input:
     tuple val(meta), path(maf)
-    val(format)
+    val(format_list)
     tuple val(meta2), path(fasta)
     tuple val(meta3), path(fai)
     tuple val(meta4), path(gzi)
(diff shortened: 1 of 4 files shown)


famosab:
I think technically this looks good. But I am wondering why the pipelines cannot use the usual channel logic of Nextflow so that each instance of this module processes one string. Do you know where I am going with this question?

charles-plessy:
I am not sure to understand, so maybe there is a part of Nextflow's channel logic that I do not know well enough.  But the channel on which the module processes input in parallel is the first one that contains pairwise genome alignments.  Is there an easy way to do a cartesian product on input alignments and output formats?

famosab:
I am not sure if I get the application exactly. But what I would do if I would use the module as is:
- add information to the meta map on which formats to convert to (like this [ id:test, gff:true, cram:true, html_gz:true]
- branch the channel based on the info given in the meta map (so every channel that has gff:true in the meta map goes into the channel that then gets fed into last/maf-convert to be converted to gff and so on)
This way the conversion happen in parallel and not after each other. And each conversion can be used independently in downstream processes. Does that help?
If I do not get the use case here correctly just let me know and maybe someone that actually uses the module in their pipeline can review :)

charles-plessy:
I see… So if the format information goes in the first input channel, I can have the alignment files appear multiple times in the queue with a different output format, and everything will be processed in parallel.  Thanks for the suggestion!  Would it be fine if the format information goes in a third member of the tuple instead of in the metadata map, like `tuple val(meta), path(maf), val(format)`?  That would simplify channel operations; I could do something like:
```
LAST_MAFCONVERT(
  LAST_ALIGN.out.maf.combine(['sam', 'bam']),
  ch_genomefile,
  ch_genomefile_index,
  []
)
```

famosab:
Sure! If that fits your implementation I think that works aswell. If people want to hand the value via the meta map they could still do that by remapping during the input stage. Having the format specified as input also makes it more transparent that this is a necessary information.

charles-plessy:
Thanks!  I will open an new PR when I am ready.

PR #7434, opened 2025-02-05
Add emu module with nf-test

### LINKED ISSUE #3020: new module: Emu

sofstam:
Closes #3020 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/emu/abundance/tests/main.nf.test
+++ b/modules/nf-core/emu/abundance/tests/main.nf.test
@@ -1,0 +1,72 @@
+nextflow_process {
+    name "Test Process EMU_ABUNDANCE"
+    script "../main.nf"
+    process "EMU_ABUNDANCE"
+    tag "modules"
+    tag "untar"
+    tag "modules_nfcore"
+    tag "emu"
+    tag "emu/abundance"
+
+    setup {
+        run("UNTAR") {
+            script "modules/nf-core/untar/main.nf"
+            process {
+                """
+                input[0] = Channel.of([
+                    [],
+                    file(
+                        params.modules_testdata_base_path + "genomics/prokaryotes/streptococcus_agalactiae/genome/emu.tar.gz",
+                        checkIfExists: true
+                    )
+                ])
+                """
+            }
+        }
+    }
+
+    test("sarscov2 nanopore [fastq_gz]") {
+        when {
+            process {
+                """
+                input[0] = [ [ id:'test' ], // meta map
+                            file(params.modules_testdata_base_path + 'genomics/prokaryotes/streptococcus_agalactiae/nanopore/fastq/test.fastq.gz', checkIfExists: true)
+                        ]
+                input[1] = UNTAR.out.untar.map{ it[1] }
+                """
+            }
+        }
+
+        then {
+            assert process.success
+            assert snapshot(
+                process.out.versions,
+                file(process.out.report[0][1]).readLines()[0]
+            ).match()
+        }
+    }
+
+    test("sarscov2 nanopore [fastq_gz] - stub") {
+        options "-stub"
+        when {
+            process {
+                """
+                input[0] = [ [ id:'test' ], // meta map
+                            file(params.modules_testdata_base_path + 'genomics/prokaryotes/streptococcus_agalactiae/nanopore/fastq/test.fastq.gz', checkIfExists: true)
+                        ]
+                input[1] = UNTAR.out.untar.map{ it[1] }
+                """
+            }
+        }
+
+        then {
+            assert process.success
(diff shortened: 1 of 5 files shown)


modules/nf-core/emu/abundance/tests/main.nf.test
```diff
@@ -0,0 +1,72 @@
...
+                input[1] = UNTAR.out.untar.map{ it[1] }
+                """
+            }
+        }
+
+        then {
+            assert process.success
+            assert snapshot(
+                process.out.versions,
+                file(process.out.report[0][1]).readLines()[0]
```
LeonHafner, line 44: Looks good overall, is there a reason why you cannot check the whole file for consistency with `assert snapshot(process.out).match()` but only the first line?

sofstam: No specific reason. I got  previously a suggestion that it would be nice to check for the contents of the file. 
Do you have a suggestion for those specific tests?

modules/nf-core/emu/abundance/tests/main.nf.test
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(
+                    process.out.report,
+                    process.out.versions
+                ).match() }
+            )
```
LeonHafner, line 47: Great, now the files are checked for the hashes. There are also other output channels, that are currently not checked for consistency: `assignment_report`, `samfile`, and `unclassified_fa`.

You can just check all output channels at once by replacing
```
{ assert snapshot(
    process.out.report,
    process.out.versions
).match() }
```
with
```
```

This checks all channels at once and makes sure we don't forget any.
Apart from that the module looks good :-)

sofstam: Since the outputs are optional, is there a way of testing those outputs? Or is it ok that the snapshot for those files are empty?

LeonHafner: Best case would be to have a test case where all outputs are produced, but there won't be any error if the snapshots of the files are empty.

sofstam: Ok, I will add a `nextflow.config` to produce those outputs.

sofstam: I updated the PR. I get `md5sum for empty file` found and the file is indeed empty. Do you know how I can possibly handle this?

LeonHafner: I didn’t know the linting disallows empty files—good to know! 👍
It seems the tool doesn’t produce any output for this file with the current test data.
If no test data produces valid output for all files, you can disable this test and define tests per output channel again.

But ideally, we should figure out why there’s no output with the current test data. 🙂

LeonHafner: Ah I see!
I think it might be too much effort to find a test case where we get unclassified reads as output, so the better option is to just check if the two files (`unclassified_mapped.fasta` and `unmapped.fasta`) exist.

You could do something like that for the `sarscov2 nanopore [fastq_gz]` test:
```groovy
assertAll(
    { assert process.success },
    { assert snapshot(
        process.out.report,
        process.out.assignment_report,
        process.out.samfile,
        process.out.versions
        ).match() },
    // Match files unclassified_mapped.fasta and unmapped.fasta as they are empty with this test case
    { assert file(process.out.unclassified_fa[0][1][0]).exists() },
    { assert file(process.out.unclassified_fa[0][1][1]).exists() }
)
```
We can keep the `stub` test as it is, since empty files are allowed there.

modules/nf-core/emu/abundance/tests/main.nf.test
```diff
@@ -0,0 +1,69 @@
...
+    tag "modules"
+    tag "untar"
+    tag "modules_nfcore"
+    tag "emu"
+    tag "emu/abundance"
+
+    setup {
+        run("UNTAR") {
+            script "modules/nf-core/untar/main.nf"
+            config "./nextflow.config"
```
LeonHafner, line 14: Can you load the config file in line 10 instead of within the setup section? I think both works, but the global version should be preferred if we you use the config for all tests.

sofstam: I moved the config to line 10 and I am resolving this.

LeonHafner (approved):
LGTM! Thank you for sticking with it for so long and incorporating all the changes! 🚀

PR #7455, opened 2025-02-07
fix tiny thing

luisas:
MAFFT adds some weird prefix, this line removes it. Without ^ it was removing also stuff that was not the prefix. 

## PR checklist


mashehu (approved):

SPPearce (changes requested):
You need to update the snapshots, but otherwise fine.

luisas:
> You need to update the snapshots, but otherwise fine.

Oh yes! done now :)

PR #7484, opened 2025-02-14
Swap graphtyper/genotype to nf-test

SPPearce:


modules/nf-core/graphtyper/genotype/main.nf
```diff
@@ -23,7 +23,6 @@ process GRAPHTYPER_GENOTYPE {
 
     script:
     def args = task.ext.args ?: ''
```
itrujnara, line 25: I know it's out of PR scope but can we get Harshil alignment on this section?

modules/nf-core/graphtyper/genotype/tests/main.nf.test
```diff
@@ -0,0 +1,92 @@
...
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
+            )
+        }
+
+    }
+
+    test("single - region_file - stub") {
```
itrujnara, line 66: Does it make sense to make a stud test for the other option? I know the outputs will be the same but technically the inputs are different

itrujnara (commented):
Nice work as always, I've left you some bits of me being picky to think about. Feel free to argue.

modules/nf-core/graphtyper/genotype/tests/main.nf.test
SPPearce: Mmm, the stub itself isn't really actually mimicking the module, as I don't understand what it is doing in terms of output.

itrujnara: Me neither so I guess it can stay this way

itrujnara (approved):
Thanks for the reply, good to go

SPPearce:
I've aligned the main.nf for you @itrujnara.
Also added an explicit sort that I think is the reason it was failing on the multiple bam test.

PR #7502, opened 2025-02-19
Allow cram output for parabricks/fq2bam module and remove fastq_align_parabricks

famosab:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


famosab:
```bash
java.lang.IllegalArgumentException: A reference must be supplied that includes the reference sequence for MT192765.1.
```

modules/nf-core/parabricks/fq2bam/main.nf
```diff
@@ -14,12 +14,14 @@ process PARABRICKS_FQ2BAM {
...
-    tuple val(meta), path("*.table"), emit: bqsr_table         , optional:true
+    tuple val(meta), path("*.bam")  , emit: bam              , optional:true
+    tuple val(meta), path("*.bai")  , emit: bai              , optional:true
+    tuple val(meta), path("*.cram") , emit: cram             , optional:true
+    tuple val(meta), path("*.crai") , emit: crai             , optional:true
+    tuple val(meta), path("*.table"), emit: bqsr_table       , optional:true
     path("versions.yml")            , emit: versions
-    path("qc_metrics")              , emit: qc_metrics          , optional:true
-    path("duplicate-metrics.txt")   , emit: duplicate_metrics  , optional:true
+    path("*")                       , emit: qc_metrics       , optional:true
```
maxulysse, line 23: this seems a bit too wide of a glob for my taste

modules/nf-core/parabricks/fq2bam/tests/main.nf.test
```diff
@@ -43,10 +57,11 @@ nextflow_process {
 
         when {
             params {
-                module_args = '--low-memory'
+                module_args = "--low-memory"
```
maxulysse, line 60: why double quotes for `--low-memory`, but single quotes for `--output-fmt bam`

subworkflows/nf-core/fastq_align_parabricks/main.nf
maxulysse, line 1: I think we can remove line 5 that include the call to the modules you removed in the code

modules/nf-core/parabricks/fq2bam/main.nf
famosab: its a directory so I was not sure how put it correctly? the name is `${prefix}_qc_metrics` at least in the stub (not sure how to do this correctly in the script block though :/

fellen31: I think if you remove the `def` in `def prefix = task.ext.prefix ?: "${meta.id}"` you are able to use `${prefix}_qc_metrics` instead of `*`.

SPPearce: And why is it not a `tuple`?

modules/nf-core/parabricks/fq2bam/main.nf
```diff
@@ -30,8 +32,10 @@ process PARABRICKS_FQ2BAM {
         error "Parabricks module does not support Conda. Please use Docker / Singularity / Podman instead."
     }
     def args = task.ext.args ?: ''
+    def args2 = task.ext.args2 ?: ''
```
SPPearce, line 35: Why is this args2 though?

SPPearce (commented):
Should we not just take a value input that controls whether to make a bam or a cram? Rather than using args2 for this.

modules/nf-core/parabricks/fq2bam/main.nf
famosab: Because these two metrics can be generated independently of each other if i got the docs correctly.

famosab:
I am wondering now: Do we even need a subworkflow if its just a wrapper around one module? 🤔 @maxulysse what do you think? Because for the sarek integration I think I need to chain this anyways with QC?

modules/nf-core/parabricks/fq2bam/main.nf
famosab: --out-duplicate-metrics OUT_DUPLICATE_METRICS
Path of duplicate metrics file after Marking Duplicates. (default: None)

--out-qc-metrics-dir OUT_QC_METRICS_DIR
Path of the directory where QC metrics will be generated. (default: None)

famosab: or do you mean that it needs a meta map as well?

famosab: tuple val(meta), path("${prefix}_qc_metrics")   , emit: qc_metrics       , optional:true

This is currectly just handled by ext.args (so if the user supplies --out-qc-metrics) so I am unsure if I can control the naming of the directory like that?

SPPearce:
> I am wondering now: Do we even need a subworkflow if its just a wrapper around one module? 🤔 @maxulysse what do you think? Because for the sarek integration I think I need to chain this anyways with QC?
> 
> Edit: Subworkflows should combine tools that make up a logical unit in an analysis step. A subworkflow must contain at least two modules. -> I will remove the sbwf

But it had two modules, aligning and applying the BQSR?

famosab:
> > I am wondering now: Do we even need a subworkflow if its just a wrapper around one module? 🤔 @maxulysse what do you think? Because for the sarek integration I think I need to chain this anyways with QC?
> > Edit: Subworkflows should combine tools that make up a logical unit in an analysis step. A subworkflow must contain at least two modules. -> I will remove the sbwf
> 
> But it had two modules, aligning and applying the BQSR?

Yes but I realized that fq2bam already does the BQSR and hence removed it and then its only fq2bam left in the subworkflow.

maxulysse:
In that particular, given that the subworkflow had no extra value, I think it was the good decision to remove it

PR #7717, opened 2025-02-28
add GHA to pass required tests when skipped

mirpedrol:
Replace https://github.com/nf-core/modules/pull/7528

### DIFF
--- a/.github/workflows/skipped-nf-test-gpu.yml
+++ b/.github/workflows/skipped-nf-test-gpu.yml
@@ -1,0 +1,22 @@
+name: Run when nf-test-gpu tests are skipped
+on:
+  push:
+    branches:
+      # https://docs.renovatebot.com/key-concepts/automerge/#branch-vs-pr-automerging
+      - "renovate/**" # branches Renovate creates
+    paths:
+      - "**/meta.yml"
+  pull_request:
+    branches: [master]
+    paths:
+      - "**/meta.yml"
+  merge_group:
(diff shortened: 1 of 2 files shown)


maxulysse (approved):

modules/nf-core/dragmap/hashtable/meta.yml
```diff
@@ -32,12 +35,16 @@ output:
       - dragmap:
           type: file
           description: DRAGMAP hash table
-          pattern: "*.{cmp,.bin,.txt}"
+          pattern: "*.{cmp,bin,txt}"
+          ontologies:
+            - edam: http://edamontology.org/format_2330 # TXT
```
maxulysse, line 40: this is genome index

mirpedrol:
Indeed, I see the tests are running because we also modified other files, not only meta.yml files, so this can't be tested in this PR

modules/nf-core/dragmap/hashtable/meta.yml
mirpedrol:
```suggestion
            - edam: http://edamontology.org/format_2330 # TXT
            - edam: http://edamontology.org/data_3210 # Genome index
```

maxulysse:
> Indeed, I see the tests are running because we also modified other files, not only meta.yml files, so this can't be tested in this PR

ok, then let's remove your changes and merge

mirpedrol:
Merging now :)

PR #5624, opened 2024-05-17
Bump fgbio versions and swap to nftest

SPPearce:
Starting the process of updating fgbio versions.
This is porting the modules from [fastquorum](https://github.com/nf-core/fastquorum), updating them to the most recent fgbio version and swapping to nf-test

PR #7718, opened 2025-02-28
Fix fastq_create_umi_consensus_fgbio and swap to nf-test

SPPearce:
The `fastq_create_umi_consensus_fgbio` subworkflow was broken last summer when the fgbio modules were updated in #5624, but the pytest didn't trigger at that point.

This PR:
- Fixes the subworkflow
- Swaps to nf-test
- Adds the meta map to the fasta, index and dict channels
- Removes the use of `samtools view` to filter the bam files with -f 1, but I don't think that the nf-core/fastquorum pipeline does this, so I suspect this is not required (pending confirmation from @nh13 or someone else from Fulcrum Genomics).
- Removes the `dump` commands


maxulysse:
no duplex?

SPPearce:
Sorry, I get you, I've updated the PR description to explain.

modules/nf-core/samtools/fastq/tests/main.nf.test
```diff
@@ -28,11 +28,14 @@ nextflow_process {
...
-                { assert snapshot(file(process.out.singleton[0][1]).name).match("bam_singleton") },
-                { assert snapshot(file(process.out.other[0][1]).name).match("bam_other") },
-                { assert snapshot(process.out.versions).match("bam_versions") }
+                { assert snapshot(
+                    process.out.fastq[0][1].collect { path(it).linesGzip[0..6] },
+                    process.out.interleaved,
+                    file(process.out.singleton[0][1]).name,
+                    file(process.out.other[0][1]).name,
+                    process.out.versions
+                    ).match()
```
nh13, line 37: nit: I am curious why this syntax is preferred?  While perhaps more concise, it now requires (1) me to go back and look at the `*.test.snap` to see what each item is matching, and (2) couples the _ordering_ in the `*.test.snap` with the _ordering_ here.  The latter to me is a footgun.  I'd prefer the explicit declaration.

modules/nf-core/samtools/fastq/tests/main.nf.test
```diff
@@ -56,12 +59,64 @@ nextflow_process {
...
-                { assert snapshot(process.out.singleton).match("bam_singinterleave_leton") },
-                { assert snapshot(file(process.out.other[0][1]).name).match("bam_interleave_other") },
-                { assert snapshot(process.out.versions).match("bam_verinterleave_sions") }
+                { assert snapshot(
+                    process.out.fastq,
+                    path(process.out.interleaved[0][1]).readLines()[0..6],
+                    process.out.singleton,
+                    file(process.out.other[0][1]).name,
+                    process.out.versions
+                    ).match()
```
nh13, line 68: ditto

subworkflows/nf-core/fastq_create_umi_consensus_fgbio/tests/main.nf.test
```diff
@@ -0,0 +1,147 @@
...
+
+        then {
+            assertAll(
+                { assert workflow.success},
+                { assert snapshot(workflow.out).match()}
+            )
+        }
+    }
+
+
```
nh13, line 146: nit
```suggestion
```

tests/config/pytest_modules.yml
```diff
@@ -441,9 +441,6 @@ subworkflows/fastq_align_chromap:
 subworkflows/fastq_contam_seqtk_kraken:
   - subworkflows/nf-core/fastq_contam_seqtk_kraken/**
   - tests/subworkflows/nf-core/fastq_contam_seqtk_kraken/**
-subworkflows/fastq_create_umi_consensus_fgbio:
-  - subworkflows/nf-core/fastq_create_umi_consensus_fgbio/**
-  - tests/subworkflows/nf-core/fastq_create_umi_consensus_fgbio/**
```
nh13, line 446: excuse my ignorance, but I think this is done since tests are moved to the sub-workflow src directory.  Is that right?

nh13 (approved):
@SPPearce this looks really good, thank-you for figuring this out and fixing it.  Take or leave my comments.

modules/nf-core/samtools/fastq/tests/main.nf.test
SPPearce: Hmm. I like it because it is more concise, and it doesn't need you to name each individual snapshot (nf-test allows you to leave the `match()` unnamed if you have only one per test context).

The ordering is fixed between the test and the snapshot, I never alter the snapshot manually, so not sure when that might be a problem.

PR #7728, opened 2025-03-03
nf-test for bam_ngscheckmate

SPPearce:
Swaps bam_ngscheckmate to nf_test.
I've also removed the dependency on bedtools_makewindows in the tests for both this subworkflow and the bam_ngscheckmate module.


SPPearce:
@maxulysse, this will remove the bedtools_makewindows requirement for the nf-test for ngscheckmate/ncm. It still needs bcftools_mpileup, but at least sarek already uses that.

maxulysse (approved):

PR #7747, opened 2025-03-05
ci: Run tests on ARM and update RNA-seq modules

edmundmiller:
Trying to use the new ARM-based runners to test the arm bioinformatics efforts.


edmundmiller:
Getting an error about apptainer on arm :(
```
The following packages have unmet dependencies:
 apptainer:amd64 : Depends: libc6:amd64 (>= 2.28) but it is not installable
                   Depends: libfuse3-3:amd64 (>= 3.2.3) but it is not installable
                   Depends: libseccomp2:amd64 (>= 2.5.0) but it is not installable
                   Depends: libuuid1:amd64 (>= 2.16) but it is not installable
                   Depends: zlib1g:amd64 (>= 1:1.1.4) but it is not installable
                   Depends: squashfs-tools:amd64 but it is not installable
```

famosab:
@edmundmiller Do you still need this PR? :)

delagoya:
Something is happening with the container `community.wave.seqera.io/library/grep_sed_tar:4b2523b1c76a5494` while Seqera Containers shows this is an arm64 build, docker sees it as `amd64`:

```
docker inspect  community.wave.seqera.io/library/grep_sed_tar:4f61255d5eff93dc | grep "Architecture"
        "Architecture": "amd64",
```

.github/actions/nf-test-action/action.yml
```diff
@@ -38,27 +38,20 @@ runs:
...
-      if: contains(inputs.profile, 'singularity')
-      shell: bash
-      run: |
-        sudo add-apt-repository universe
-        sudo apt-get update
-        sudo apt-get install -y uidmap squashfs-tools
-
-    - name: Setup apptainer
-      if: contains(inputs.profile, 'singularity')
-      uses: eWaterCycle/setup-apptainer@3f706d898c9db585b1d741b4692e66755f3a1b40 #v2
```
maxulysse, line 51: why removing singularity?

mirpedrol: I think we still need this for the `4cpu-linux-x64` runner

.github/workflows/nf-test.yml
```diff
@@ -101,16 +91,18 @@ jobs:
   nf-test:
     runs-on:
       - runs-on=${{ github.run_id }}
-      - runner=4cpu-linux-x64
-      - image=ubuntu22-full-x64
```
mirpedrol, line 105: don't we need to specify an image?

edmundmiller: According to the docs: 
>By default, the image defaults to ubuntu22-full-x64. If you want to use arm64 family instance types, and don’t specify a default runner name as base (e.g. runner=2cpu-linux-arm64) you will also need to override the image label.

.github/actions/nf-test-action/action.yml
mirpedrol: I meant installing `uidmap` and `squashfs-tools` and the `Setup apptainer` step

.github/workflows/nf-test.yml
mirpedrol: 🤔 maybe because we do specify a runner at [line 94 ](https://github.com/nf-core/modules/pull/7747/files#diff-a6e08ed61fd370dba7712991afc0f04ab727f2c501c3ca2e0da629142c4e7bbaR94)👍

mashehu: yep, image is not needed. I just always add this line, because I will probably switch to ubuntu24 soon to mirror github

.github/actions/nf-test-action/action.yml
mashehu: I think it should work now without it.

.github/scripts/conda_env_sorter.py
```diff
@@ -8,7 +8,7 @@
 
 """Sort dependencies in conda environment files."""
 # Test with
-# uv run --with pytest -- pytest -v .github/scripts/conda_env_sorter.py
+# uv run --with pytest --with "ruamel.yaml" -- pytest -v .github/scripts/conda_env_sorter.py
```
mashehu, line 11: I don't like that this has uv as a dependency.

.github/scripts/conda_env_sorter.py
mashehu, line 1: I think tools should do that instead of a script here.

mashehu: Actually modules lint should already do most of it.

.github/scripts/conda_env_sorter.py
edmundmiller: It doesn't, it's just an ease of use option 🤷🏻‍♂️ 

You can still run it with `python .github/scripts/conda_env_sorter.py`

.github/scripts/conda_env_sorter.py
edmundmiller: Agreed, but this is all I have time for right now. The logic can be moved to tools. 

I also think it's a worthwhile trade-off for now. Usually, these CI changes run into edge cases we can't predict.

mashehu: Nope, sorry. I don't think we should use modules CI for experiments again. This goes to tools, with tests etc and then it can be used here (we use the tools dev branch here anyway so it is a bit faster)

edmundmiller: Alright, I'll just remove the script for now, then. I'll add it to tools when I get to it!

edmundmiller:
> Do you think switching to ubuntu 24 in the runner would solve the singularity setup problem?

Tried it, no dice 😞

PR #7761, opened 2025-03-09
allow --Reference to be .fa as well as .fasta

mbeavitt:
Another small change that checks for both extensions (.fa or .fasta) in the input reference file.

No checklist/tests as this is a minor change.


SPPearce:
Why is the module checking the ending of the input file at all, and not just passing it to the tool if one is provided? Same with the vcf input.

SPPearce:
(using an appropriate ?: operator to check if a file was given)

mbeavitt:
Good question. I suppose checking filetype can be omitted.

mbeavitt:
Should the whole module be refactored to match?

mbeavitt:
I have a couple more things to add actually (including use of $prefix as output file prefix) so I will close this PR in favour of a more comprehensive refactor with tests, as this is no longer a small change.

Thanks for the feedback!

PR #7772, opened 2025-03-11
basicpy nf-test

kbestak:
Attempt 3:
New PR opened due to conflicts.
In Gitpod, tests pass when running `nf-core modules test BASICPY --profile docker` and `nf-core modules test BASICPY --profile singularity`

## PR checklist

Closes #[7539](https://github.com/nf-core/modules/issues/7539) 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


modules/nf-core/basicpy/tests/main.nf.test
```diff
@@ -0,0 +1,57 @@
...
+                                file(params.modules_testdata_base_path + 'imaging/ome-tiff/cycif-tonsil-cycle1.ome.tif', checkIfExists: true),
+                ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
```
famosab, line 51:
```suggestion
                { assert snapshot(process.out,
                path(process.out.versions[0]).yaml).match() }
```
Can you add this and then we check how the versions string looks?

famosab: ahh but I just saw its a manual version thing so this makes no sense, sorry

modules/nf-core/basicpy/tests/main.nf.test
```diff
@@ -0,0 +1,57 @@
...
```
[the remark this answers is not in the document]
kbestak: Thanks for the tip, I've updated the output specification to:
```
    output:
    tuple val(meta), path("*-dfp.tiff"), path("*-ffp.tiff"), emit: profiles
```
The md5sum is still exactly the same as before - I don't see how this will affect the difference between the files produced by the github runners and in the gitpod environment. Is there potentially any way to check the output file from the action runners so I could run diff to see the difference?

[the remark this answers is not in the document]
famosab: You will need to do two output channels so
```
    output:
    tuple val(meta), path("*-dfp.tiff"), emit: dfp
    tuple val(meta), path("*-ffp.tiff"), emit: ffp
```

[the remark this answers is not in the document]
famosab: I'll check in later again though. I have some ideas.

kbestak:
As with some other modules, I could just check for the presence of output files, but I think it's essential to understand where the difference in the "cycif-tonsil-cycle1.ome-ffp.tiff" md5sums between github runners and other environments comes from. Especially since both Docker and Singularity produce reproducible results when run on the same platform.

famosab:
Hey I think we should try this plugin: https://github.com/nf-core/nft-tiff

kbestak:
I think it would be easier to understand the source of the difference if I could access or reproduce the outputs from runners here, would that be possible somehow?

famosab:
MAybe with codespaces?

kbestak:
Thanks for highlighting Codespaces! I actually got a third md5sum value there (again reproducible across Docker and Singularity) but these files I could now access.

The difference seems to come from floating point precision issues

kbestak:
I've updated the test to check for existence and size of output files - since this is a floating point error, is this appropriate?

Additionally, testing in the mcmicro workflow adds a groovy script: https://github.com/nf-core/mcmicro/blob/dev/tests/lib/ImageUtils.groovy to extract the metadata which gets validated. Could something similar be added to module testing?

SPPearce:
> I've updated the test to check for existence and size of output files - since this is a floating point error, is this appropriate?

Yes, many modules only test for existence of files such as pdf or png or files that have varying precision on the outputs.

I'm not sure how best to add more functions, could be via a nf-test plugin? Probably ask that one on slack.

famosab:
I think we can leave it like that for now. And maybe ping the nft-tiff developers that we would be interested when they finalize their plugin to use it?

modules/nf-core/basicpy/main.nf
```diff
@@ -8,8 +8,9 @@ process BASICPY {
     tuple val(meta), path(image)
 
     output:
-    tuple val(meta), path("*.tiff"), emit: fields
-    path "versions.yml"            , emit: versions
+    tuple val(meta), path("*-dfp.tiff"), emit: dfp
+    tuple val(meta), path("*-ffp.tiff"), emit: ffp
```
famosab, line 12: if this is impractical downstream you can also revert this change :)

modules/nf-core/basicpy/tests/main.nf.test
```diff
@@ -0,0 +1,60 @@
...
+                                [ id:'test' ],
+                                file(params.modules_testdata_base_path + 'imaging/ome-tiff/cycif-tonsil-cycle1.ome.tif', checkIfExists: true),
+                ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
```
famosab, line 26:
```suggestion
                { assert process.success },
                { assert snapshot(
                     process.out.versions,
                     file(process.out.ffp.get(0).get(1)).exists(),
                     file(process.out.dfp.get(0).get(1)).exists(),
                     file(process.out.ffp.get(0).get(1)).size(),
                     file(process.out.dfp.get(0).get(1)).size()
                     ).match() 
                     },
```
maybe this can also go into the snapshot? and also even the size

modules/nf-core/basicpy/main.nf
kbestak: Perfect! In that case I would revert it to the profiles output `tuple val(meta), path("*-dfp.tiff"), path("*-ffp.tiff"), emit: profiles

SPPearce (commented):
One more thing.
The stub is currently making files like:
test.-dfp.tiff
using a `${prefix}.-ffp.tiff`
But the actual tool script doesn't use prefix, so what does the actual tool do?
Is it worth including a renaming step from the tool output, to actually use prefix (assuming the tool doesn't have any way of setting the output file names)

kbestak:
As it stands, the tool names the outputs based on the input file name (strips `tiff` in the end, adds `-dfp.tiff` or `-ffp.tiff`, respectively for the calculated dark, or flat field profiles. The tool is planned to be updated, would it be okay to leave it like this for now, and then make sure the `prefix` is used when naming outputs?

kbestak:
Is there anything else needed for the review, or can this be approved for merging?

SPPearce (commented):
Can you just delete the pytest test files then you can merge.

kbestak:
Thanks for noticing, I removed them now!

SPPearce (approved):

PR #7777, opened 2025-03-11
Add estimated depth output to bbmap/sendsketch

zachary-foster:
This allows for pipelines to treat samples differently depending on depth estimates. I  added the output of this channel to the current tests, but it seems that this value is no always output by sendsketch. I checked and for this data it is not present, but the exact same code in my pipeline almost always returns a value. Perhaps it is because of how small the test data is.


famosab:
Can you paste how you would the output expect to look?

fellen31:
This looks like something that should be done in the pipeline rather than the module to me.

zachary-foster:
The output looks like this with a `.view()`:

```
[[id:LR1], 38.039]
[[id:LF4], 12.682]
[[id:OT3], 2.677]
[[id:LR2], 16.347]
[[id:LF2], 4.113]
[[id:PF3], 5.553]
[[id:PF2], 5.818]
[[id:VC2], 10.701]
```

> This looks like something that should be done in the pipeline rather than the module to me.

Yea, thats a good point. I can do that too. When I first put this in the module I was newer to nextflow and didn't realize you can read files in channels. Either way.

zachary-foster:
The equivalent channel manipulation code to get this info is:

```
sendsketch_depth = BBMAP_SENDSKETCH.out.hits
    .splitText(limit: 2, by: 2)
    .map { sample_meta, header ->
        def match = header =~ /Depth: ([0-9.]+)/
        [sample_meta, match[0][1]]
    }
```
Not too hard now that I have some experience with nextflow, but a beginner would have a hard time I think. Anyway, I have adapted our pipeline to use this code so we can use the current version of `bbmap/sendsketch`, so I leave it up to you all as to whether this is useful in the module. Feel free to close if it is not useful.

PR #7788, opened 2025-03-12
New module: `busco/download`

zachary-foster:
Add new modules `BUSO_DOWNLOAD` to download databases for use with BUSCO tools.


modules/nf-core/busco/download/main.nf
```diff
@@ -0,0 +1,48 @@
...
+        busco: \$( busco --version 2>&1 | sed 's/^BUSCO //' )
+    END_VERSIONS
+    """
+
+    stub:
+    def args = task.ext.args ?: ''
+    """
+    echo "busco \\
+        --download $lineage \\
+        $args"
```
atrigila, line 39: I don't think this is necessary. You can simply keep the mkdir below.

zachary-foster: Ok, I removed the code.

zachary-foster:
Thanks for reviewing! I think I have addressed the issue.

atrigila (approved):

PR #7793, opened 2025-03-13
Update bwa/index: Name bwa/index output path relative to input

zachary-foster:
This will make is so the published output directory has a better structure. Each reference will have its own folder instead of them all being mixed into a redundant `bwa` directory underneath a `bwa_index` directory.


SPPearce:
If this is implemented, it'll be via #9947

PR #7795, opened 2025-03-14
Update PIRATE

zachary-foster:
This does a few things:

* Renames .gff3 files (like those produced by `bakta` to .gff so `PIRATE` uses them
* Updates conda/docker/singularity versions
* Names output directories by the prefix, so multiple runs dont mix their results in the same published directory


luisas (approved):
LGTM :)

PR #7804, opened 2025-03-14
Update `graphtyper/genotype` to allow for compressed reference

zachary-foster:
* Allows for compressed reference files
* Updates conda/docker repos
* Handles cases when outputs were not in the expected folder structure


famosab:
Somehow there is a lot of commits that you want to merge in? Maybe it makes sense to start from a cleaner state? :)

zachary-foster:
ok, I redid the changes here: 

https://github.com/nf-core/modules/pull/8387

All the commits were trying to make the tests pass here even though they pass locally with `nf-test`

PR #7806, opened 2025-03-14
New module:  `khmer/trimlowabund`

zachary-foster:
Adds a new module for `trim-low-abund.py` from the khmer tool kit. This removes low abundance k-mers from reads, which is useful as a preprocessing step when making sketches of reads using tools like `sourmash sketch`.


modules/nf-core/khmer/trimlowabund/main.nf
```diff
@@ -0,0 +1,43 @@
...
+    path "versions.yml"                 , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    """
+    trim-low-abund.py \\
+        -M ${task.memory.toGiga()}e9 \\
```
SPPearce, line 24: There is some cases when task.memory might not be set, so you need to check for that. See the gatk or fgbio modules for instance.

zachary-foster: Ok, I have added this check in. Thanks!

modules/nf-core/khmer/trimlowabund/main.nf
```diff
@@ -0,0 +1,49 @@
...
+    path "versions.yml"                 , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    if (!task.memory) {
+        log.info '[KHMER_TRIMLOWABUND] Available memory not known - defaulting to 16GB. Specify process memory requirements to change this.'
+        avail_mem = 16
```
SPPearce, line 24: 16 is too much for a process that is set to process_low in the label

SPPearce (commented):
Few questions.
What resources does the tool need? You've set process_low but then defaulting to 16G.
Is there a way to name the output file with prefix rather than the filename?

modules/nf-core/khmer/trimlowabund/main.nf
zachary-foster: This one is a bit odd. It is single threaded, but requires at least 16Gb in most cases.

zachary-foster: OK, increased it to `process_medium`

modules/nf-core/khmer/trimlowabund/main.nf
```diff
@@ -0,0 +1,63 @@
...
+    def args = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "${meta.id}"
+    if (!task.memory) {
+        log.info '[KHMER_TRIMLOWABUND] Available memory not known - defaulting to 16GB. Specify process memory requirements to change this.'
+        avail_mem = 16
+    } else {
+        avail_mem = task.memory.toGiga()
+    }
+    file_ext = seq_file.name - ~/\.gz$/ - ~/^[^\.]*\./
+    output_path = "${prefix}_trimmed.${file_ext}"
```
SPPearce, line 30:
```suggestion
    output_path = "${prefix}.${file_ext}"
```

modules/nf-core/khmer/trimlowabund/main.nf
```diff
@@ -0,0 +1,63 @@
...
+    output:
+    tuple val(meta), path("${output_path}"), emit: trimmed
+    path "versions.yml"                    , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "${meta.id}"
```
SPPearce, line 22:
```suggestion
    prefix = task.ext.prefix ?: "${meta.id}_trimmed"
```

SPPearce (commented):
Can you add the check that the output file isn't going to be the same as the input file, then I think it is good.

zachary-foster:
Ok, I made those changes. Good idea having the `_trimmed` only applied when `prefix` is not defined.

modules/nf-core/khmer/trimlowabund/meta.yml
```diff
@@ -0,0 +1,51 @@
...
+input:
+  - - meta:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. `[ id:'sample1' ]`
+    - seq_file:
+        type: file
+        description: fasta/fastq file
+        pattern: "*.{fa,fasta,fas,fq,fastq}"
```
SPPearce, line 29:
```suggestion
        pattern: "*.{fa,fasta,fas,fq,fastq}"
        ontologies:
          - edam: "http://edamontology.org/data_2044" # Sequence
          - edam: "http://edamontology.org/format_1929" # FASTA
          - edam: "http://edamontology.org/format_1930" # FASTQ
```

modules/nf-core/khmer/trimlowabund/meta.yml
```diff
@@ -0,0 +1,51 @@
...
+  - trimmed:
+    - meta:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. `[ id:'sample1' ]`
+    - '${output_path}':
+        type: file
+        description: fasta/fastq file with low abundance k-mers removed.
+        pattern: "*.{fa,fasta,fas,fq,fastq}"
```
SPPearce, line 41:
```suggestion
        pattern: "*.{fa,fasta,fas,fq,fastq}"
        ontologies:
          - edam: "http://edamontology.org/data_2044" # Sequence
          - edam: "http://edamontology.org/format_1929" # FASTA
          - edam: "http://edamontology.org/format_1930" # FASTQ
```

tests/modules/nf-core/khmer/uniquekmers/nextflow.config
SPPearce, line 1: This file shouldn't be being changed here (but if you could volunteer to migrate khmer/uniquekmers to nf-test that'd be great!)

SPPearce (approved):
Ok, think that is good.
I've added some ontologies (if you can add these to modules going forward that'd be great)

PR #7807, opened 2025-03-14
New module:  `seqkit/head`

zachary-foster:
Add new module `seqkit/head` that subsets FASTA/FASTQ files to some number of sequences. This is useful for limiting the size of read inputs. To make it easier to deal with paired end reads, it accepts a list of FASTAs/FASTQs, but also works on a single one. It also accepts the number of sequences to subset as a channel input instead of using `args` because this might be different for each input subset (e.g. subsetting to a proportion of the reads based on a depth estimate).


modules/nf-core/seqkit/head/main.nf
```diff
@@ -0,0 +1,51 @@
...
+    script:
+    def args = task.ext.args ?: ''
+    """
+    for f in ${fastqs.join(' ')}
+    do
+        seqkit head \\
+            ${args} \\
+            --threads $task.cpus \\
+            -n ${seq_count} \\
+            -o "\$(basename \$f .fastq.gz)_subset.fastq.gz" \\
```
luisas, line 29: Here I would instead use $prefix, so that it is tunable by the configs. e.g. https://github.com/nf-core/modules/blob/adb94eac45898f7ddbccc5d52101805d5fa55437/modules/nf-core/famsa/align/main.nf#L27

luisas:
Also, I saw you left the modules with the right tag but if you want a faster review you can post it on the #request-review channel in the nf-core slack and people are veeeery fast to review usually :)

zachary-foster:
Thanks! I have added prefix to the output paths.

PR #7808, opened 2025-03-15
Update `sourmash/compare`: Name output by prefix and update conda/docker

### LINKED ISSUE #7698: nf-test migration: sourmash/gather

### LINKED ISSUE #7699: nf-test migration: sourmash/taxannotate

zachary-foster:
* Name output files by the prefix so that multiple runs of  `sourmash compare` do not overwrite each other when published
* Updates the conda/docker dependencies for all sourmash tools
* Adds stub and associated test for `sourmash/sketch` to fix broken stub test for `sourmash/compare`
* Migrates `sourmash/gather` to nf-test (resolves #7698)
* Migrates `sourmash/taxannotate` to nf-test (resolves #7699)


SPPearce:
Oh, looks like the other subtools of sourmash are still on pytest, so probably best not updating in this PR anyway.

SPPearce:
> It looks to me like sourmash compare is already on the latest version? I can update the other sourmash tools, but it sounds like we should not do that right now?

If you could update and swap the other sourmash tools to nf-test at some point in the near future that would be much appreciated, we are trying to finish off all the pytest -> nf-test conversions :)

zachary-foster:
Ok, I will update the docker/conda versions for all the sourmash tools and do the pytest -> nf-test conversions for `gather` and `taxannotate`.

zachary-foster:
Thanks! No problem! I migrated the tests in this PR since  I had already started before I saw you last message.

PR #7824, opened 2025-03-17
Added braker3

### LINKED ISSUE #6034: new module: BRAKER3

GallVp:
Closes #6034
Added braker3


GallVp:
TODO,

Add conda to exclude list. It is not supported.

GallVp:
Hi @OlivierCoen 

I don't think it runs with Conda. Yes, the container runs with singularity. That's what we have at our HPC. Singularity/Apptainer can auto-convert this container in the background.

OlivierCoen:
Have you tried yourself if it can run with conda? there's an up-to-date braker3 conda package so normally it should (but perhaps I'm wrong):
https://anaconda.org/bioconda/braker3

For modules, the best practice is to write:
```
        '<singularity image>':
        '<docker image>' }"
```

GallVp:
Thank you @OlivierCoen 

From the BRAKER team,

> Warning: installing GeneMark-ETP for BRAKER in conda environments has lead to multiple problems reported by users (Issues!). We can not offer support for conda installations. Please use the singularity image instead.

right above https://github.com/Gaius-Augustus/BRAKER#augustus


Yes, that's the best-practice but of bioconda is not available, we can just provide docker. Please see: https://github.com/nf-core/modules/blob/81fd83c9f3eb7cbb55aa97b7f1fdf0a81ec98233/modules/nf-core/parabricks/minimap2/main.nf#L7

OlivierCoen:
@GallVp you're right! My bad :) Braker is really not stable, I don't understand how packages like this can still be so hard so install and configure...
I tried myself to setup a conda env, but nf-test tests fail for obscure reasons...
Could you please add a comment in main.nf explaining why you just provided a container statement (and no conda)? Thanks a lot :)

Also, I would remove the single_end in the meta map (in tests/main.nf.test), as it is useless here:
https://nfcore.slack.com/archives/CJRH30T6V/p1738835521052779

GallVp:
Thank you @OlivierCoen 

I have made the suggested changes.

PR #7830, opened 2025-03-17
Added tirlearner

### LINKED ISSUE #5552: new module: TIRLEARNER

GallVp:
Closes #5552

Added tirlearner

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,13 +1,4 @@
 repos:
-  - repo: local
-    hooks:
-      - id: conda-env-sorter
-        name: Sort dependencies in conda environment files.
-        entry: ./.github/scripts/conda_env_sorter.py
-        language: python
-        files: environment
-        types: [yaml]
-        additional_dependencies: ["ruamel.yaml"]
   - repo: https://github.com/pre-commit/mirrors-prettier
     rev: "v4.0.0-alpha.8"
     hooks:
--- a/modules/nf-core/tirlearner/main.nf
+++ b/modules/nf-core/tirlearner/main.nf
@@ -1,6 +1,6 @@
 process TIRLEARNER {
     tag "$meta.id"
-    label 'process_high'
+    label 'process_medium'
 
     conda "${moduleDir}/environment.yml"
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
--- a/modules/nf-core/tirlearner/main.nf
+++ b/modules/nf-core/tirlearner/main.nf
@@ -4,8 +4,8 @@ process TIRLEARNER {
 
     conda "${moduleDir}/environment.yml"
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
-        'https://depot.galaxyproject.org/singularity/tir-learner:3.0.3--hdfd78af_0':
-        'biocontainers/tir-learner:3.0.3--hdfd78af_0' }"
+        'https://depot.galaxyproject.org/singularity/tir-learner:3.0.7--hdfd78af_0':
+        'biocontainers/tir-learner:3.0.7--hdfd78af_0' }"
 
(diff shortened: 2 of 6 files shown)


modules/nf-core/tirlearner/main.nf
```diff
@@ -0,0 +1,74 @@
+process TIRLEARNER {
+    tag "$meta.id"
+    label 'process_high'
```
SPPearce, line 3: Does this need process_high?

modules/nf-core/tirlearner/main.nf
```diff
@@ -0,0 +1,74 @@
...
+    $unzip_fasta
+
+    TIR-Learner \\
+        -f $input_name \\
+        -s $species \\
+        -t $task.cpus \\
+        -o $prefix \\
+        $args \\
+        &> >(tee "${prefix}.log" 2>&1) \\
+        || echo "TIR-Learner failed! See ${prefix}.log"
```
SPPearce, line 43: Do we not just want to fail the module at this point?

SPPearce (commented):
Generally fine, just a question on the resource requirements and whether we definitely want to not fail if nothing is detected.
I guess the argument is that the tool doesn't handle finding nothing gracefully, but will this silently fail if e.g. the arguments aren't passed correctly?

modules/nf-core/tirlearner/main.nf
GallVp: Good question.

Generally yes. But for some situations, the pipeline can use alternatives if a module fails. This module was bing used in such a situation.

We do have a capability built into Nextflow where we can choose the level at which to handle the errors. For example, we can set the error strategy to `ignore` for a particular process and then use partial join to route the inputs to a secondary module which we want to run in place of the original module. However, not all tools map the error codes correctly. For example, error code `1` is used for all types of error. Thus, we have to rely on error log to decide whether to ignore the error or not. 

The limitation of the current solution is that error handling at the pipeline level is not fully robust. For example, if the tool fails with error code `1` and the log does not contain the magic words such as `This file does not contain TIRs`, what should the pipeline do? If we do a System.exit(1), that -- if my memory or a recent Nextflow release do not fail me -- leads to a pipeline crash rather than a pipeline `finish` even if the `errorStrategy` is set to finish.

I feel this needs further input from a wider group. Maybe, I should link a Nextflow issue? Or, should I post to some channel?

modules/nf-core/tirlearner/main.nf
GallVp: Not sure. Demoted to `process_medium`. I have also updated the ontologies.

Thank you for the review.

modules/nf-core/tirlearner/main.nf
SPPearce: I would ask this on the maintainers team Slack channel I think

mahesh-panchal: Can you use `grep -c "STRING"` after the `||` like documented here: https://nf-co.re/docs/guidelines/components/modules#capturing-exit-codes

GallVp: Thank you @mahesh-panchal for the very useful suggestion. I'll use it elsewhere.

Here after bumping the version, I found out that the tool author has changed the behaviour and now the tool exits with `0` when no TIR is found. I have removed the second (`||`) command.

.pre-commit-config.yaml
```diff
@@ -1,13 +1,4 @@
 repos:
-  - repo: local
-    hooks:
-      - id: conda-env-sorter
-        name: Sort dependencies in conda environment files.
-        entry: ./.github/scripts/conda_env_sorter.py
```
GallVp, line 6: @mashehu 
This script was removed by https://github.com/nf-core/modules/pull/7747 but the hook was retained.

atrigila:
Any plans to finish this PR @GallVp or can we close it as unplanned for the moment?

PR #7846, opened 2025-03-19
`last/mafconvert`: add read group IDs and sort SAM/BAM/CRAM files.

charles-plessy:
Just after I started to use this module to produce BAM files, my users asked if I can provide thes BAM files already sorted.  Sorting is necessary for indexing, which is necessary for random access and display on browsers.

I can't think of a workflow where the performance or the ease of use would be worsened when SAM/BAM/CRAM files are sorted by default. Therefore I think that it is best to sort these files on the fly in this module.

In addition, this MR also adds a read group with ID and sample name the same as `meta.id` for all converted alignments.  This makes it easy to concatenate files where different queries are aligned on the same target.

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:


famosab:
This module emit multiple files in one output channel (in line 22) I think we usually do not allow that in nf-core. The rest of the PR seems fine to me though. I am a bit confused why there are so many changes in the nf.test file.

famosab:
Maybe we can generally swap the nf-test to md5sums which is easier to read
                    process.out.bam.collect { bam(it[1]).getSamLines() },
                                        process.out.bam.collect { bam(it[1]).getReadsMD5() },

charles-plessy:
Hi @famosab, thank you for your comments! I have replaced `getSamLines` with `getReadsMD5` as you suggested.

Regarding `tuple val(meta), path("*.cram"), path(fasta)` it is not part of the PR (the module already has it). I would be very happy to move `path(fasta)` to its own output channel or just drop it altogether, but I need the FASTA file in the test `process.out.cram.collect { cram(it[1], it[2]).getReadsMD5() }`, and more than a hour of trial and error was not enough to find how to replace ` it[2]` with an alternative way to pass that file, which is defined a few lines above.

```
input[1] = [
                    [ id:'contigs.genome' ], // meta map
                    file(params.modules_testdata_base_path + 'genomics/sarscov2/genome/genome.fasta',     checkIfExists: true)
```

In my other new commits after your first review, I have also simplfied the command generating the CRAM files, added a small fix about BED support and samtools version number to the version output channel.  I hope you don't mind.

famosab:
It seems like for cram files people usually only assert the HeaderMD5() which does not need the fasta. So maybe we can use that?

```
bam(process.out.cram[0][1]).getHeaderMD5(),
```

charles-plessy:
Unfortunately `getHeaderMD5` is not reproducible because the CRAM file is made in a different temporary folder at each run and the absolute path to the folder is part of the path of the reference genome that is recorded in the CRAM header…

famosab:
Ok then lets just leave it at that

famosab (approved):

PR #7867, opened 2025-03-20
Make arg parsing regex in seaqtk sample more specific avoiding spurious matches

samuell:
While reviewing a PR pulling down the seqtk sample module, we noticed a regex in some of the argument parsing logic that could cause troubles by being too inspecific.

It is on [this line](https://github.com/nf-core/modules/blob/7f88aae/modules/nf-core/seqtk/sample/main.nf#L23)

I have tested the current pattern against a number of imaginary contents of the arg variable in the `nextflow console`, and one can see that the pattern:

1. Misses cases where there is a space between `-s` and the number.
2. Erroneously matches against filenames containing `-s[0-9]` as part of the filename.

image

With the fixed regex, that adds a space before the `-s` flag, and an optional space after it, both of these are fixed:

image

## PR checklist

  - For modules:

### DIFF
--- a/modules/nf-core/seqtk/sample/main.nf
+++ b/modules/nf-core/seqtk/sample/main.nf
@@ -20,7 +20,7 @@ process SEQTK_SAMPLE {
     script:
     def args   = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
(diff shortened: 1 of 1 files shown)


charles-plessy (approved):
Good catch!  The changes look good to me.

PR #7940, opened 2025-03-24
Parameterising args for RTGTOOLS_FORMAT

steven-rusk:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


SPPearce:
Hi, 
What is the reason for this PR? [Modules should generally](https://nf-co.re/docs/guidelines/components/modules#optional-command-arguments) just use `task.ext.args` to pass optional arguments directly to the tool.

steven-rusk:
> Hi, What is the reason for this PR? [Modules should generally](https://nf-co.re/docs/guidelines/components/modules#optional-command-arguments) just use `task.ext.args` to pass optional arguments directly to the tool.

The reason for this PR is to allow for some type-checking and input validation for those optional parameters. $task.ext.args seems to just be a string - do you know if there is an established pattern for input validation of this single string variable?

steven-rusk:
> Yes, `task.ext.args` is typically just a string (there is one module I can think of which passes it as a list, but that is non-standard). The type checking is normally just left to the tool itself to complain about. What do you mean by "input validation"? Anything that requires a file needs to be treated separately as an input channel.

@SPPearce Thanks for clarifying. Several of the possible optional args for this module *do* require files, so good to know that those should always be handled as inputs - I can at least make that change. If leaving the type-checking for non-file optional args up to the tool is the established norm, then that's what I'll do. Thanks for your feedback.

SPPearce:
Ok, thanks.
Yes, anything that can be a file input must be staged via the input channels, as Nextflow expects all its inputs to be inside the work directory. On a HPC or locally you can get away with pointing to a random file in your system (although bad practice), but when executing on a cloud infrastructure that doesn't work and only the files that are being input will actually be present.

SPPearce:
Also I note that one of the arguments is called seed. If that is controlling randomness, does the default produce reproducible results? If not, then I'd be keen for that one to be checked for. (For example)[https://github.com/nf-core/modules/blob/bbaee8e515e10d623cf255f04897c5d3696085da/modules/nf-core/quilt/quilt/main.nf] or (here)[https://github.com/nf-core/modules/blob/bbaee8e515e10d623cf255f04897c5d3696085da/modules/nf-core/umitools/dedup/main.nf#L32]

SPPearce:
Closing this now, as it hasn't had any action for a year. If you (or anyone else) wish to resume at a later point that is fine.

PR #7959, opened 2025-03-24
ultra/pipeline nf-test conversion

### LINKED ISSUE #7615: nf-test migration: ultra/pipeline

SPPearce:
Will close #7615


SPPearce:
Not currently working as the test gives an error :(

awgymer:
So I ran the old pytest and nf-test locally and I found that the "sorted" gtf is different across the two invocations.

awgymer:
Ok, working now. Note that the md5sum for the SAM output matches in the snapshot and the old pytest yml

SPPearce:
Thanks @awgymer !

Issue #7900, opened 2025-03-23
deepvariant


jen-reeve:

scorreard:
Iit seems that empty .gz files generated by simply touch <filename>.gz break the snapshot functionality of nf-test by causing an EOFException because the files are not actually compressed.

A fix is to replace the touch command with `echo "" | gzip > <filename>.gz`

PR #7960, opened 2025-03-24
Updated deevariant to avoid failing nf-test

scorreard:
## PR checklist

Closes [#7900](https://github.com/nf-core/modules/issues/7900)

It seems that empty .gz files generated by simply touch .gz break the snapshot functionality of nf-test by causing an EOFException because the files are not actually compressed.

A fix is to replace the touch command with echo "" | gzip > <filename>.gz


scorreard:
Waiting for the checks to run before merfing. If checks are validated, merge the PR

maxulysse (approved):

PR #7965, opened 2025-03-24
Swap muscle to nf-test

### LINKED ISSUE #7630: nf-test migration: muscle

SPPearce:
Closes #7630


modules/nf-core/muscle/tests/main.nf.test
```diff
@@ -0,0 +1,98 @@
...
+
+    test("test_muscle_maketree") {
+
+        setup {
+            run("MUSCLE", alias: "MUSCLE_SETUP") {
+                script "../main.nf"
+                params {
+                    module_args = '-fasta -verbose -phys -phyi -maxiters 2'
+                }
+
```
awgymer, line 47: As in `ultra/pipleine` I don't think this belongs here. 

You need to set these in the `when` block, probably naming them more like `muscle_setup_args` and then use that in the config. 

You will most likely need to set the args in the config using the alias name too as otherwise it will clash with the config for the actual `muscle` process under test

SPPearce: I generally both of these using Usman's automatic conversion, maybe something needs looking at in that then

awgymer: Do you mean the `migrate-pytest` in tools?

LouisLeNezet: > Do you mean the migrate-pytest in tools?

No it's using `pytest2nf-test` we could effectively change how it's written in the tool.

SPPearce:
Thanks for fixing it @awgymer !

PR #7969, opened 2025-03-24
Updated gatk4mergevcfs to pass nf-core test

### LINKED ISSUE #7902: gatk4/mergevcfs

scorreard:
## PR checklist

Closes #7902

It seems that empty .gz files generated by simply touch .gz break the snapshot functionality of nf-test by causing an EOFException because the files are not actually compressed.

A fix is to replace the touch command with echo "" | gzip > <filename>.gz


SPPearce (approved):
I don't think this is actually being tested, is there a stub test?

famosab:
@SPPearce yes there is a stub test now :)

famosab:
Failure related to #8241

PR #7983, opened 2025-03-24
Migration `cnvnator/cnvnator` to nf-test 

### LINKED ISSUE #7549: nf-test migration: cnvnator/cnvnator

max-laurent:
## PR checklist

Closes #7549

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


SPPearce (changes requested):
I think the tests are missing the config arguments they had as pytest.

modules/nf-core/cnvnator/cnvnator/tests/main.nf.test
```diff
@@ -0,0 +1,124 @@
...
+
+    tag "modules"
+    tag "modules_nfcore"
+    tag "cnvnator"
+    tag "cnvnator/cnvnator"
+
+    setup {
+        run("CNVNATOR_CNVNATOR", alias: "CNVNATOR_RD") {
+            script "../main.nf"
+
```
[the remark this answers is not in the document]
LouisLeNezet: The warning is normal, it is due to the way nf-test assert the parameters during compilation.

LouisLeNezet (approved):
Looking good to me !
LGTM 🎉

modules/nf-core/cnvnator/cnvnator/tests/main.nf.test.snap
```diff
@@ -0,0 +1,86 @@
+{
+    "test_cnvnator": {
+        "content": [
+            [
+                [
+                    {
+                        "id": "test",
+                        "single_end": false
+                    },
+                    "test.tab:md5,d41d8cd98f00b204e9800998ecf8427e"
```
LouisLeNezet, line 10: Do you know why the test.tab is empty ?

LouisLeNezet: It seems that it was the case in the pytest...
Maybe the datasets is not enough.

max-laurent: Yes the input files are the same as in the pytest so I don't really understand if this is the normal behavior or not. I will try with other .bam files as input.

LouisLeNezet: Maybe increasing to reducing the bin size would help ?
But maybe there is just no CNV in the bam file it's located at chr22:16570000-16610000

max-laurent: I thought about that too. I tested other input files that did not yield any cnv but then I found an input file similar to the one they used in their documentation on the nf-core test datasets (NA12878.chr22.bam). This runs and yields 3 cnvs now. I had to add a parameter to the CNVNATOR_RD process otherwise the process was terminated after the memory allocation.

LouisLeNezet: In this case you can modify the nf-test to only check for the tab file name (same as root) and add a comment to explain that no CNV are present in this location.

max-laurent: The output was not consistent across profiles resulting in an error for the Conda test. This was due to a change in the p-values obtained, I saw that the container version had changed on biocontainer (https://biocontainers.pro/tools/cnvnator) resulting I think in different environment per profile that could explain the differences in p-values. I therefore updated the flags for singularity and docker images and now the output is consistent across the different profiles.

LouisLeNezet: You can also impose the image on conda by adding `- bioconda::cnvnator=0.4.1=py310h2dce045_7`

max-laurent: Ok it is done, I had to use '=' instead of '--'. 
Should I change the test data back to what was in the pytest? I think that it is interesting to have data that generates CNVs.

LouisLeNezet: > I had to use '=' instead of '--'

I never remenber 😞 

> Should I change the test data back to what was in the pytest? I think that it is interesting to have data that generates CNVs.

I totally agree. How it is now is perfect, you can go ahead and merge !!

SPPearce (approved):
Well done for finding an input file that actually has some CNV and fixing the issues.
We often test modules on inappropriate input; it is best if some meaningful output is generated but it isn't essential.

PR #8008, opened 2025-03-25
nf-test migration: cellrangerarc/count

### LINKED ISSUE #7543: nf-test migration: cellrangerarc/count

ignatiusm:
## PR checklist

Closes #7543

This is a migration from pytest to nf-test for `cellrangerarc/count` module (initially started during the hackathon.

I used pytest2nftest for the migration, and edited the `tests/main.nf.test` file to reflect the fact that several of the outputs do no produce stable snapshots.

I also removed conda CI tests as the `cellrangerarc/count` module does not work with conda (I also removed it from the pytest-related ci actions).

There is no stub test, as the setup for these tests involve the `unzip` module. The unzip module stub is non-functional: it creates a directory, but it stays empty and no archive contents are extracted into it. This caused the downstream tests to fail as a `.fa` file was expected from the input `.fa.gz`.

Tests passing locally and in CI.


sateeshperi (approved):

sateeshperi:
there might be some inconsistent md5sums...plz check again

PR #8026, opened 2025-03-25
[New module] Infernal cmsearch

prototaxites:
Adds a module for infernal's cmsearch tool. Pretty much a drop-in replacement for hmmsearch - I've basically cloned the module in its entirety.

[DONE] Once https://github.com/nf-core/test-datasets/pull/1542 is in I'll change the path to the model file in the tests.

## PR checklist

Closes #XXX 

- [x] Add a resource `label`
  - For modules:


modules/nf-core/infernal/cmsearch/meta.yml
```diff
@@ -0,0 +1,85 @@
...
+      doi: "10.1093/bioinformatics/btt509"
+      licence: ["BSD-3-clause"]
+      identifier: biotools:infernal
+
+input:
+  - - meta:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. [ id:'test', single_end:false ]
```
famosab, line 24:
```suggestion
          e.g. [ id:'test' ]
```
probably not applicable

modules/nf-core/infernal/cmsearch/main.nf
```diff
@@ -0,0 +1,73 @@
...
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args        = task.ext.args   ?: ''
+    def prefix      = task.ext.prefix ?: "${meta.id}"
+    output          = "${prefix}.txt"
+    alignment       = write_align     ? "-A ${prefix}.sto" : ''
+    target_summary  = write_target    ? "--tblout ${prefix}.tbl" : ''
```
famosab, line 29: Not sure if that is our best practise (see comment below) but I see why. I would however leave this to the pipeline developer and do it in the config since these are just tool flags :) but keep the optional outputs

famosab (commented):
Few comments on the optional outputs and setting of tool flags :)

prototaxites:
Thanks for the review!

Because the tool is by the same authors of (and has the same interface as) hmmer/hmmsearch, I copied the module code itself almost verbatim from https://github.com/nf-core/modules/blob/master/modules/nf-core/hmmer/hmmsearch/main.nf. 

I would tend to agree about leaving the optional flags to the args, but left it as-is to keep the same interface as the other module. Happy to drop it though, but in which case, do you think the module still try and gzip optional outputs if they exist? In large cases I imagine they could become quite sizable.

modules/nf-core/infernal/cmsearch/main.nf
```diff
@@ -0,0 +1,73 @@
...
+    stub:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    touch "${prefix}.txt"
+    ${write_align ? "touch ${prefix}.sto" : ''} \\
+    ${write_target ? "touch ${prefix}.tbl" : ''}
+
+    gzip --no-name *.txt \\
+        ${write_align ? '*.sto' : ''} \\
+        ${write_target ? '*.tbl' : ''}
```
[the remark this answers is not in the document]
prototaxites: Can simplify - this is also verbatim from the other module

prototaxites:
OK, have looked into removing these val inputs - having taken some input on the #modules Slack (https://nfcore.slack.com/archives/CJRH30T6V/p1742982305507179) - I think in this case they're better to leave in, as there is a lot of ambiguity in the possible file names a user could specify in `ext.args` that wouldn't match the output block specifications. This way we can enforce a schema on file extensions, which also makes the gzip compression element more stable too.

modules/nf-core/infernal/cmsearch/tests/main.nf.test.snap
```diff
@@ -0,0 +1,119 @@
+{
+    "rfam rrna - ecoli - stub": {
+        "content": [
+            {
+                "0": [
+                    [
+                        {
+                            "id": "test",
+                            "single_end": false
```
famosab, line 9: where does this come from in the snapshot 🤔

prototaxites: I think I didn’t regenerate it…

famosab: Then maybe do that :)

prototaxites: 🤞

modules/nf-core/infernal/cmsearch/meta.yml
famosab: then remove it :)

modules/nf-core/infernal/cmsearch/meta.yml
```diff
@@ -0,0 +1,97 @@
...
+          Flag to save optional per target summary in a tabular
+          format. Specify with 'true' to save.
+
+output:
+  - output:
+      - meta:
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. [ id:'test', single_end:false ]
```
famosab, line 55:
```suggestion
            e.g. [ id:'test' ]
```

famosab (approved):
Looks good now :)

PR #8077, opened 2025-03-26
Remove environment.yml from imc2mc

SPPearce:
The `mcstaging/imc2mc` module uses a docker file and doesn't work on conda, yet it had a conda environment.yml file.
In particular, this file had `anaconda` as a channel, so I've just deleted the environment.yml


LouisLeNezet (approved):
Seems fine for me

PR #8125, opened 2025-03-26
New module: imagesegmentationmetrics

BioinfoTongLI:
## PR checklist

This is a module that compares two (or more) segmentation label images and compute a list of metrics which describes the differences between them. It leverages functions from the MONAI(https://docs.monai.io/en/stable/metrics.html) and seg-metrics(https://pypi.org/project/seg-metrics/) pakcages.

- [x] Add a resource `label`
  - For modules:


BioinfoTongLI:
there are 2 points that you might have different opinnions
1. we're using our own buckets for s3 test data hosting (https://github.com/nf-core/modules/pull/8125/files#diff-95d638d619883dceb2a6efe036da2ef4801ed547df695ec0943e1f452c4e9e38R3)
2. we're putting the python script as an executable in the `resource/usr/bin` folder and enabled moduleBinaries=true ([modules/nf-core/imagesegmentationmetrics/resources/usr/bin/segmentation_metrics.py](https://github.com/nf-core/modules/pull/8125/files#diff-8a1b84656e4dc617f27a4f1a419ca59f439e0eca31d00f15cbaa5ad1a4a582bf))

SPPearce (commented):
Hmm, this hasn't been reviewed for a year :( 
@BioinfoTongLI , are you still interested in this?

BioinfoTongLI:
Hi @SPPearce, yes if some body can review this? I don't think there's more to do on our side. But I might be wrong, since it's so long ago :)

SPPearce (commented):
Well, the immediate things I see:

- We can't use the resources/usr/bin folder for the python script in nf-core/modules, because that requires wave to work on cloud computing. So this would need to be a `template` instead.
- I think this would need to be in the custom folder rather than a seperate tool name (e.g. custom/imagesegmentationmetrics).

BioinfoTongLI:
thanks for the quick answers! I will amend those.
PS : is there any docs why resource/usr/bin is not allowed? Is `template` the general recommendations for all bespoke scripts now?

SPPearce:
https://nf-co.re/docs/specifications/components/modules/general#script-inclusion, doesn't explicitly say why.
The reason is: https://docs.seqera.io/wave/nextflow/bundle-scripts#module-bin-directory
image
and we don't want to require wave a dependency.
Which is a shame, as module binaries are much nicer than templates in my opinion.

BioinfoTongLI:
Thanks for the pointers! 
hmm...ok, that's indeed a shame.  I will have a look if there's much to change and update here.

SPPearce:
The swap to template isn't too bad, you can still have the same separate script, you just refer to the parameters in a slightly different way.

SPPearce (commented):
Note that you'd also need to swap the version extraction to use topics rather than the END_VERSIONS, [see the docs](https://nf-co.re/docs/specifications/components/modules/general#topic-qualifiers-should-be-used-for-emitting-versions) or basically all the other modules now. But if you look at other python modules you can see how they generate that inside the python script.

PR #8171, opened 2025-03-31
Support formula in `abundance_differential_filter` and associated modules

atrigila:
This PR is a work in progress.

The changes in this PR allow to support formula as input in `abundance_differential_filter`. 
As of now, formula is null for all the modules, except `VARIANCEPARTITION_DREAM`. The default behaviour of the other modules was not changed.

While the main goal of this PR is to be able to incorporate `VARIANCEPARTITION_DREAM` into the subworkflow, naturally this requires modifying some other components.

Some notes to discuss: 
- `PROPR_PROPD` does not seem to handle formula, so I have to think how to work around this. At the moment, it has formula as part of the inputs, but this will be modified as it is not needed. The module should remain as it previously was.
- In the case of modules that allow formulas:
  - If formula is provided, should we prioritize that, rather than building the model with the contrast variables? This is the case of `limma` and `deseq2`. 

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/propr/propd/meta.yml
+++ b/modules/nf-core/propr/propd/meta.yml
@@ -39,9 +39,6 @@ input:
         description: |
           The value within the contrast_variable column of the sample sheet that
           should be used to derive the target samples
-    - formula:
-        type: string
-        description: (Optional) R formula string used for modeling, e.g. '~ treatment + (1 | sample_number)'.
   - - meta2:
         type: map
         description: |
--- a/modules/nf-core/propr/propd/meta.yml
+++ b/modules/nf-core/propr/propd/meta.yml
@@ -39,6 +39,9 @@ input:
         description: |
           The value within the contrast_variable column of the sample sheet that
           should be used to derive the target samples
+    - formula:
+        type: string
+        description: (Optional) R formula string used for modeling, e.g. '~ treatment + (1 | sample_number)'.
   - - meta2:
         type: map
         description: |
@@ -58,13 +61,13 @@ output:
       - meta:
           type: file
           description: |
-            TSV-format table of genes associated with differential expression 
+            TSV-format table of genes associated with differential expression
             information as compiled from the propd results
           pattern: "*.propd.genewise.tsv"
       - "*.propd.genewise.tsv":
           type: file
           description: |
-            TSV-format table of genes associated with differential expression 
+            TSV-format table of genes associated with differential expression
             information as compiled from the propd results
           pattern: "*.propd.genewise.tsv"
   - genewise_plot:
@@ -97,15 +100,15 @@ output:
       - meta:
           type: file
           description: |
-            (Optional) TSV-format table of the native propd pairwise results. This 
-            table contains the differential proportionality values associated to 
+            (Optional) TSV-format table of the native propd pairwise results. This
+            table contains the differential proportionality values associated to
             each pair of genes.
           pattern: "*.propd.pairwise.tsv"
       - "*.propd.pairwise.tsv":
           type: file
           description: |
-            (Optional) TSV-format table of the native propd pairwise results. This 
-            table contains the differential proportionality values associated to 
+            (Optional) TSV-format table of the native propd pairwise results. This
+            table contains the differential proportionality values associated to
             each pair of genes.
           pattern: "*.propd.pairwise.tsv"
   - results_pairwise_filtered:
@@ -113,28 +116,28 @@ output:
           type: file
           description: |
             (Optional) TSV-format table of the filtered propd pairwise results. This
-            table contains the pairs of genes with significant differential 
+            table contains the pairs of genes with significant differential
             proportionality values.
           pattern: "*.propd.pairwise_filtered.tsv"
       - "*.propd.pairwise_filtered.tsv":
           type: file
           description: |
             (Optional) TSV-format table of the filtered propd pairwise results. This
-            table contains the pairs of genes with significant differential 
+            table contains the pairs of genes with significant differential
             proportionality values.
           pattern: "*.propd.pairwise_filtered.tsv"
   - adjacency:
       - meta:
           type: file
           description: |
-            (Optional) CSV-format table of the adjacency matrix defining a graph, with 
+            (Optional) CSV-format table of the adjacency matrix defining a graph, with
             edges (1) associated to pairs of genes that are significantly differentially
             proportional.
           pattern: "*.propd.adjacency.csv"
       - "*.propd.adjacency.csv":
           type: file
           description: |
-            (Optional) CSV-format table of the adjacency matrix defining a graph, with 
+            (Optional) CSV-format table of the adjacency matrix defining a graph, with
             edges (1) associated to pairs of genes that are significantly differentially
             proportional.
           pattern: "*.propd.adjacency.csv"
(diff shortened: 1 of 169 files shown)


grst:
> In the case of modules that allow formulas:
If formula is provided, should we prioritize that, rather than building the model with the contrast variables? This is the case of limma and deseq2.

In my opinion, all modules that support formulas should *only* support formulas as this is the most general way to specify a model. For compatibility with the previous system, formulas can be built off the (column, blocking_var) tuples outside the modules, or even outside this workflow. This avoids redundant code that would be shared between the modules. 

> PROPR_PROPD does not seem to handle formula, so I have to think how to work around this. At the moment, it has formula as part of the inputs, but this will be modified as it is not needed. The module should remain as it previously was.

@suzannejin can you clarify if PROPD supports any complex design at all? From the vignette I could only find ANOVA-like comparisons between two or more groups, but it doesn't even seem to support blocking variables. 

In that case we could either parse a formula and raise an error if it doesn't match `^~\s+(\w)$`, i.e. a single factor. Alternatively, we can support different ways of specifying models per method and document properly which method requires which type of model specification.

subworkflows/nf-core/abundance_differential_filter/main.nf
```diff
@@ -26,25 +26,24 @@ workflow ABUNDANCE_DIFFERENTIAL_FILTER {
...
     // Set up how the channels crossed below will be used to generate channels for processing
-    def criteria = multiMapCriteria { meta_input, abundance, analysis_method, fc_threshold, stat_threshold, meta_exp, samplesheet, meta_contrasts, variable, reference, target ->
+    def criteria = multiMapCriteria { meta_input, abundance, analysis_method, fc_threshold, stat_threshold, meta_exp, samplesheet, meta_contrasts, variable, reference, target, formula ->
         def meta_for_diff = mergeMaps(meta_contrasts, meta_input) + [ 'method_differential': analysis_method ]
         def meta_input_new = meta_input + [ 'method_differential': analysis_method ]
         samples_and_matrix:
             [ meta_input_new, samplesheet, abundance ]
         contrasts_for_diff:
-            [ meta_for_diff, variable, reference, target ]
+            [ meta_for_diff, variable, reference, target, formula ]
```
pinin4fjords, line 42:
```suggestion
            [ meta_for_diff, variable, reference, target ]
        contrasts_for_diff_with_formula:
            [ meta_for_diff, variable, reference, target, formula ]
```

subworkflows/nf-core/abundance_differential_filter/main.nf
```diff
@@ -26,25 +26,24 @@ workflow ABUNDANCE_DIFFERENTIAL_FILTER {
...
         filter_params:
             [ meta_for_diff, [ 'fc_threshold': fc_threshold, 'stat_threshold': stat_threshold ]]
         contrasts_for_norm:
-            [ meta_input_new, variable, reference, target ]
+            [ meta_input_new, variable, reference, target, formula ]
```
pinin4fjords, line 46:
```suggestion
            [ meta_input_new, variable, reference, target ]
```

subworkflows/nf-core/abundance_differential_filter/main.nf
```diff
@@ -131,19 +130,41 @@ workflow ABUNDANCE_DIFFERENTIAL_FILTER {
...
+    // ----------------------------------------------------
+
+    // DREAM only runs with formula
+    ch_contrasts_with_formula = inputs.contrasts_for_diff
+        .filter{ it[0].method_differential == 'dream' }
+        .branch {
+            valid: it[0].formula != null
+            invalid: it[0].formula == null
+        }
+        .valid
```
pinin4fjords, line 144:
```suggestion
    dream_inputs = inputs.contrasts_for_diff_with_formula
        .filter { meta_for_diff, variable, reference, target, formula ->
            meta.method_differential == 'dream' && formula != null
        }
```

subworkflows/nf-core/abundance_differential_filter/main.nf
```diff
@@ -131,19 +130,41 @@ workflow ABUNDANCE_DIFFERENTIAL_FILTER {
...
+
+    VARIANCEPARTITION_DREAM(
+        ch_contrasts_with_formula,
```
pinin4fjords, line 147:
```suggestion
        dream_inputs,
```

pinin4fjords (changes requested):
Suggest keeping it simple for now. We're never going to formula passable to everything, so it makes sense to generate contrasts channels both with and without the formula in the multimap, so you have both types available.

You don't need most of the other changes here for now, the other modules can be made formula-accepting progressively (where possible).

modules/nf-core/propr/propd/meta.yml
```diff
@@ -39,6 +39,9 @@ input:
         description: |
           The value within the contrast_variable column of the sample sheet that
           should be used to derive the target samples
+    - formula:
```
grst, line 42: propd doesn't support formulas. Do we still want it to accept a formula field for consistency that is ignored, or is it better to remove it?

modules/nf-core/variancepartition/dream/meta.yml
```diff
@@ -21,22 +21,44 @@ input:
...
           workflow level to pass optional parameters to the module, e.g.
           [
-            contrast_id: 'treatment_mCherry_hND6',
+            id: 'treatment_mCherry_hND6',
             contrast_variable: 'treatment',
-            contrast_reference: 'mCherry',
-            contrast_target: 'hND6',
+            reference: 'mCherry',
+            target: 'hND6',
             formula: '~ treatment + (1 | sample_number)'
```
grst, line 28: This has been removed from meta, hasn't it?

subworkflows/nf-core/abundance_differential_filter/main.nf
grst, line 144: Can we fail or at least warn on "invalid" inputs?

grst (changes requested):

grst:
> Suggest keeping it simple for now. We might never have formulas passable to everything, so it makes sense to generate contrasts channels both with and without the formula in the multimap, so you have both types available.

We certainly won't have formulas for everything, there will always be some methods that don't support it. You can always build a formula from a contrast vector + blocking factor, but not necessarily vice-versa. 

For me the question is at what level to address that. Either each module can take both and do the conversion internally, or each module takes only whatever is the most general specification it supports and the conversion is done outside. For me the latter makes more sense because it avoids reimplementing the same thing across modules. But is not the hill I am going to die on if you prefer otherwise. 

> You don't need most of the other changes here for now, the other modules can be made formula-accepting progressively (where possible).

Agree, let's first get DREAM right, and follow up with the others separately. Maybe let's even think again about (complex) contrast specification (https://github.com/nf-core/differentialabundance/issues/377) first before updating the other methods.

atrigila:
Alright, then next steps are:  
- We’ll add an additional contrast channel with formula, used only for `dream` for now.  
- We’ll remove `val(formula)` from `deseq2`, `limma`, and `propd` for the time being.  

I also tested `val(formula)` with `limma` and it produces correct results, so we could consider supporting it there later if it makes sense.

Let’s keep the discussion going around how we want to handle complex contrast specifications, and whether supporting formulas in other models makes sense longer term.

pinin4fjords (approved):
Much better- thanks. Tidy, and flexible.

atrigila:
@grst since you requested changes we'll need your approval to be able to merge

grst (approved):
LGTM now

PR #8177, opened 2025-04-01
Fix: Quote shell parameters

zeehio:
https://github.com/nf-core/differentialabundance/pull/444

I ran a differential abundance pipeline with a study name that had spaces in it, and the pipeline crashed because the `make_app_from_files.R` caller script was not escaping parameters as expected. I believe the patch included in this PR is pretty easy to follow: The parameters were expanded by the `$` and the second word of the study name was interpreted as another argument.

Making this change made the pipeline succeed.

If you believe this contribution is of your interest but you require of me to add tests I can try to find time to learn how to write tests and comply with the rest of the checklist. It may take me a while though to find the time, and the patch is really minor so I would appreciate your consideration.

You may want to consider security ramifications of being able to pass a study_name like `; arbitrary_command` and inject code that would be executed by the pipeline. I am not familiar enough with nextflow to understand whether that should be considered a vulnerability risk or not, but if you think this is an issue, then this PR would fix it.

In any case, thanks for your time and for all the work you do. 

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/shinyngs/app/main.nf
+++ b/modules/nf-core/shinyngs/app/main.nf
@@ -38,13 +38,13 @@ process SHINYNGS_APP {
 
     """
     make_app_from_files.R \\
-        --sample_metadata $sample \\
-        --feature_metadata $feature_meta \\
-        --assay_files ${assay_files.join(',')} \\
-        --contrast_file $contrasts \\
-        --contrast_stats_assay $contrast_stats_assay \\
-        --differential_results ${differential_results.join(',')} \\
-        --output_dir $prefix \\
+        --sample_metadata "$sample" \\
+        --feature_metadata "$feature_meta" \\
+        --assay_files "${assay_files.join(',')}" \\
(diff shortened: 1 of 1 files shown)


mashehu:
maybe we should add one test where we cover this situation, just to avoid running into the same thing in the future.

zeehio:
Sorry I have been out of the office for the last few weeks.

Instead of having to write tests for these kind of small issues I believe it would be better to be able to make the CI run `shellcheck` on the shell scripts that are part of the nextflow files, so these issues are automatically reported. I would find it a challenge to extract all the shell scripts from the nf files though, although nextflow is already doing that so it should be feasible to implement.

I would have to set up the development environment and learn to write tests. I can do that, but it may take me a long while (another month?) to find the time. Is the test case necessary?

grst:
@nf-core-bot fix linting

PR #8198, opened 2025-04-03
ragtag/scaffold and ragtag/patch modules and tests

nschan:
Here are two modules for [`ragtag`](https://github.com/malonge/RagTag), one for assembly scaffolding to a reference, and one for assembly patching based on a target sequence. These were local modules in [`genomeassembler`](https://github.com/nf-core/genomeassembler/pull/136) which I am [beginning to migrate](https://github.com/nf-core/genomeassembler/issues/110) to nf-core/modules.


modules/nf-core/ragtag/patch/main.nf
```diff
@@ -0,0 +1,79 @@
...
+        mv ${query} query.fa
+    fi
+
+    ragtag.py patch target.fa query.fa \\
+        -o "${prefix}" \\
+        -t ${task.cpus} \\
+        ${args}
+
+    mv ${prefix}/ragtag.patch.agp ${prefix}.patched.agp
+    mv ${prefix}/ragtag.patch.fasta ${prefix}.patched.fasta
```
jfy133, line 50:
```suggestion
    mv ${prefix}/ragtag.patch.agp ${prefix}.patch.agp
    mv ${prefix}/ragtag.patch.fasta ${prefix}.patch.fasta
```

modules/nf-core/ragtag/patch/main.nf
```diff
@@ -0,0 +1,79 @@
...
+    container "${workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container
+        ? 'https://depot.galaxyproject.org/singularity/ragtag:2.1.0--pyhb7b1952_0'
+        : 'biocontainers/ragtag:2.1.0--pyhb7b1952_0'}"
+
+    input:
+    tuple val(meta), path(target), path(query)
+
+    output:
+    tuple val(meta), path("*.patched.fasta"), emit: patched_fasta
+    tuple val(meta), path("*.patched.agp"),   emit: patched_agp
```
jfy133, line 15:
```suggestion
    tuple val(meta), path("*.patch.fasta"), emit: patch_fasta
    tuple val(meta), path("*.patch.agp"),   emit: patch_agp
```

modules/nf-core/ragtag/patch/main.nf
```diff
@@ -0,0 +1,79 @@
...
+    mv ${prefix}/ragtag.patch.fasta ${prefix}.patched.fasta
+    mv ${prefix}/ragtag.patch.comps.fasta ${prefix}.comps.fasta
+    mv ${prefix}/ragtag.patch.ctg.agp ${prefix}.ctg.agp
+    mv ${prefix}/ragtag.patch.ctg.fasta ${prefix}.ctg.fasta
+    mv ${prefix}/ragtag.patch.rename.agp ${prefix}.rename.agp
+    mv ${prefix}/ragtag.patch.rename.fasta ${prefix}.rename.fasta
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        RagTag: \$(echo \$(ragtag.py -v | sed 's/v//'))
```
jfy133, line 59:
```suggestion
        ragtag: \$(echo \$(ragtag.py -v | sed 's/v//'))
```

modules/nf-core/ragtag/patch/main.nf
```diff
@@ -0,0 +1,79 @@
...
+    END_VERSIONS
+    """
+
+    stub:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def args   = task.ext.args ?: ""
+    """
+    touch ${prefix}.patched.agp
+    touch ${prefix}.patched.fasta
```
jfy133, line 68:
```suggestion
    touch ${prefix}.patch.agp
    touch ${prefix}.patch.fasta
```

modules/nf-core/ragtag/patch/main.nf
```diff
@@ -0,0 +1,79 @@
...
+    touch ${prefix}.comps.fasta
+    touch ${prefix}.ctg.agp
+    touch ${prefix}.ctg.fasta
+    touch ${prefix}.rename.agp
+    touch ${prefix}.rename.fasta
```
jfy133, line 73:
```suggestion
    touch ${prefix}.rename.fasta
    
```

modules/nf-core/ragtag/patch/main.nf
```diff
@@ -0,0 +1,79 @@
...
+        -t ${task.cpus} \\
+        ${args}
+
+    mv ${prefix}/ragtag.patch.agp ${prefix}.patched.agp
+    mv ${prefix}/ragtag.patch.fasta ${prefix}.patched.fasta
+    mv ${prefix}/ragtag.patch.comps.fasta ${prefix}.comps.fasta
+    mv ${prefix}/ragtag.patch.ctg.agp ${prefix}.ctg.agp
+    mv ${prefix}/ragtag.patch.ctg.fasta ${prefix}.ctg.fasta
+    mv ${prefix}/ragtag.patch.rename.agp ${prefix}.rename.agp
+    mv ${prefix}/ragtag.patch.rename.fasta ${prefix}.rename.fasta
```
jfy133, line 55: Are these all the files that are produced in the `${prefix}/` directory? Are there anything else that a user may want to use?

modules/nf-core/ragtag/patch/main.nf
```diff
@@ -0,0 +1,79 @@
...
+    else
+        mv ${target} target.fa
+    fi
+
+    if [[ ${query} == *.gz ]]
+    then
+        zcat ${query} > query.fa
+    else
+        mv ${query} query.fa
+    fi
```
jfy133, line 42: Is there a reason for the renaming and not just a straight decompression if `.gzipped`?

modules/nf-core/ragtag/patch/main.nf
```diff
@@ -0,0 +1,79 @@
...
+    touch ${prefix}.patched.agp
+    touch ${prefix}.patched.fasta
+    touch ${prefix}.comps.fasta
+    touch ${prefix}.ctg.agp
+    touch ${prefix}.ctg.fasta
+    touch ${prefix}.rename.agp
+    touch ${prefix}.rename.fasta
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        RagTag: \$(echo \$(ragtag.py -v | sed 's/v//'))
```
jfy133, line 76:
```suggestion
        ragtag: \$(echo \$(ragtag.py -v | sed 's/v//'))
```

modules/nf-core/ragtag/patch/main.nf
```diff
@@ -0,0 +1,79 @@
...
+    fi
+
+    if [[ ${query} == *.gz ]]
+    then
+        zcat ${query} > query.fa
+    else
+        mv ${query} query.fa
+    fi
+
+    ragtag.py patch target.fa query.fa \\
```
jfy133, line 44: Does the module produce any logging info? [^1]

[^1]: https://nf-co.re/docs/guidelines/components/modules#capturing-stdout-and-stderr

modules/nf-core/ragtag/patch/meta.yml
```diff
@@ -0,0 +1,110 @@
+name: "ragtag_patch"
+description: Patches a target assembly using a query assembly
+keywords:
+  - assembly
+  - consensus
+  - ragtag
```
jfy133, line 6:
```suggestion
  - ragtag
  - patch
```

modules/nf-core/ragtag/patch/meta.yml
```diff
@@ -0,0 +1,110 @@
...
+    - target:
+        type: file
+        description: Target assembly
+        pattern: "*.{fasta,fasta.gz}"
+    - query:
+        type: file
+        description: Query assembly
+        pattern: "*.{fasta,fasta.gz}"
+output:
+  - patched_fasta:
```
jfy133, line 31:
```suggestion
  - patch_fasta:
```

modules/nf-core/ragtag/patch/meta.yml
```diff
@@ -0,0 +1,110 @@
...
+      - meta:
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. [ id:'test' ]
+      - "*.patched.fasta":
+          type: file
+          description: FASTA file containing the patched assembly
+          pattern: "*.patched.fasta"
+  - patched_agp:
```
jfy133, line 41:
```suggestion
  - patch_agp:
```

modules/nf-core/ragtag/patch/meta.yml
```diff
@@ -0,0 +1,110 @@
+name: "ragtag_patch"
+description: Patches a target assembly using a query assembly
```
jfy133, line 2: To be a 'naive' user: what does 'patching' mean here? Changing what for what? exactly

modules/nf-core/ragtag/patch/tests/main.nf.test
```diff
@@ -0,0 +1,61 @@
+// TODO nf-core: Once you have added the required tests, please run the following command to build this file:
+// nf-core modules test ragtag/patch
```
jfy133, line 2:
```suggestion
```

modules/nf-core/ragtag/patch/tests/main.nf.test
```diff
@@ -0,0 +1,61 @@
...
+    name "Test Process RAGTAG_PATCH"
+    script "../main.nf"
+    process "RAGTAG_PATCH"
+
+    tag "modules"
+    tag "modules_nfcore"
+    tag "ragtag"
+    tag "ragtag/patch"
+
+    // TODO nf-core: Change the test name preferably indicating the test-data and file-format used
```
jfy133, line 14:
```suggestion
```

modules/nf-core/ragtag/patch/tests/main.nf.test
```diff
@@ -0,0 +1,61 @@
...
+test("A. thaliana Col-0 test data - ragtag - patch") {
+
+        when {
+            process {
+                """
+                input[0] = [
+                    [ id:'Col-0_2MB' ], // meta map
+                    file('https://raw.githubusercontent.com/nf-core/test-datasets/genomeassembler/A_thaliana_Col-0_2mb/ONT-Col-0_flye_assembly.fasta.gz', checkIfExists: true),
+                    file('https://raw.githubusercontent.com/nf-core/test-datasets/genomeassembler/A_thaliana_Col-0_2mb/Col-CEN_v1.2.Chr1_5MB-7MB.fasta.gz', checkIfExists: true)
```
jfy133, line 23: You should'nt use pipeline test data in modules, as you could theoretically move this - the modules branch should be more stable (use the exsiting FASTA files in test-data, or copy these over if necessary)

jfy133: (e.g. just use the SARSCov2 genome twice) - note the output does not need to make biological sense - we just need to keep the tool happily executing without an error.

modules/nf-core/ragtag/patch/meta.yml
jfy133: From the tool's help:

Homology-based assembly patching: Make continuous joins and fill gaps in 'target.fa' using sequences from 'query.fa'

modules/nf-core/ragtag/scaffold/main.nf
```diff
@@ -0,0 +1,68 @@
...
+    tag "$meta.id"
+    label 'process_medium'
+
+    conda "${moduleDir}/environment.yml"
+    container "${workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container
+        ? 'https://depot.galaxyproject.org/singularity/ragtag:2.1.0--pyhb7b1952_0'
+        : 'biocontainers/ragtag:2.1.0--pyhb7b1952_0'}"
+
+    input:
+    tuple val(meta), path(assembly), path(reference)
```
jfy133, line 11: Same two missing files as above, as well as an extra:

```
  -e <exclude.txt>     list of reference sequences to ignore [null]
  -j <skip.txt>        list of query sequences to leave unplaced [null]
  -J <hard-skip.txt>   list of query headers to leave unplaced and exclude from 'chr0' ('-C') [null]
```

modules/nf-core/ragtag/patch/main.nf
jfy133: As in, we need to accomodate all possible output files (that a user _could_ use downstream)

modules/nf-core/ragtag/scaffold/main.nf
```diff
@@ -0,0 +1,68 @@
...
+        -C \\
+        $args
+
+    mv ${prefix}/ragtag.scaffold.fasta ${prefix}.fasta
+    mv ${prefix}/ragtag.scaffold.agp ${prefix}.agp
+    mv ${prefix}/ragtag.scaffold.stats ${prefix}.stats
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        RagTag: \$(echo \$(ragtag.py -v | sed 's/v//'))
```
jfy133, line 52:
```suggestion
        ragtag: \$(echo \$(ragtag.py -v | sed 's/v//'))
```

modules/nf-core/ragtag/scaffold/main.nf
```diff
@@ -0,0 +1,68 @@
...
+    """
+
+    stub:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def args = task.ext.args ?: ''
+    """
+    touch ${prefix}.fasta
+    touch ${prefix}.agp
+    touch ${prefix}.stats
+    cat <<-END_VERSIONS > versions.yml
```
jfy133, line 63:
```suggestion

    cat <<-END_VERSIONS > versions.yml
```

modules/nf-core/ragtag/scaffold/main.nf
```diff
@@ -0,0 +1,68 @@
...
+    "${task.process}":
+        RagTag: \$(echo \$(ragtag.py -v | sed 's/v//'))
```
jfy133, line 65:
```suggestion
        ragtag: \$(echo \$(ragtag.py -v | sed 's/v//'))
```

modules/nf-core/ragtag/scaffold/tests/main.nf.test
```diff
@@ -0,0 +1,57 @@
...
+
+    test("A. thaliana Col-0 test data - ragtag - scaffold") {
+
+        when {
+            process {
+                """
+                input[0] = [
+                    [ id:'Col-0_2MB' ], // meta map
+                    file('https://raw.githubusercontent.com/nf-core/test-datasets/genomeassembler/A_thaliana_Col-0_2mb/ONT-Col-0_flye_assembly.fasta.gz', checkIfExists: true),
+                    file('https://raw.githubusercontent.com/nf-core/test-datasets/genomeassembler/A_thaliana_Col-0_2mb/Col-CEN_v1.2.Chr1_5MB-7MB.fasta.gz', checkIfExists: true)
```
jfy133, line 20: save as above

jfy133 (commented):
Overall pretty good, just some optmisation suggetions to bring in line with the nf-core specifications: https://nf-co.re/docs/guidelines/components/modules#capturing-stdout-and-stderr

modules/nf-core/ragtag/patch/main.nf
nschan: Not really.

modules/nf-core/ragtag/patch/main.nf
nschan: Yes!

modules/nf-core/ragtag/patch/main.nf
nschan: There are some more, I have added them.

modules/nf-core/ragtag/patch/main.nf
jfy133: I would remove it then, just gunzip and pass the same `${target}` file but without the `.gz` suffix to the ragtag command itself

nschan: I would prefer to leave this as is, mainly because I have switched the test to both target and query being the same, so I need to stage stuff a bit more carefully...

jfy133: You can add a check for that: https://nf-co.re/docs/guidelines/components/modules#command-file-output-naming

And then you can quickly rename the file itself in the test with 

https://nfcore.slack.com/archives/CJRH30T6V/p1632732509092300?thread_ts=1632485648.017300&cid=CJRH30T6V

nschan: Ok, I am fine with changing that, but what if someone wants to actually use two files that have the same name? I don't think it is an awesome idea, but I can easily imagine a person that uses `genome.fasta.gz` from different folders as reference/target and query?

jfy133: They just can't, thus the error :)

The module will fail anyway because nextflow will stay it can't stage two files with the same name

nschan: If my tests would not take ages I would have pushed my recent changes; but I changed it to:
`tuple val(meta), path(target, name: 'target/*'), path(query, name: 'query/*'),`

jfy133:
@nschan how much data does the module produce?

```
    >   tail: 'test/ragtag.patch.err' has become inaccessible: No such file or directory
```

And what does the tail of this file look like locally?

modules/nf-core/ragtag/patch/main.nf
jfy133: That doesn't really address the issue here, what is the reasoning behind that here?

modules/nf-core/ragtag/patch/main.nf
```diff
@@ -0,0 +1,101 @@
...
+
+    if [[ ${query} == *.gz ]]
+    then
+        zcat ${query} > query.fa
+    else
+        cp ${query} query.fa
+    fi
+    mkdir ${prefix} && touch ${prefix}/ragtag.patch.err
+    tail -F ${prefix}/ragtag.patch.err >&2 &
+    ragtag.py patch target.fa query.fa \\
```
jfy133, line 49: I would rather go for the `tee` method unless you had a specific reason for using the `tail` method, given you seem to only be saving the STDERR stream here.

modules/nf-core/ragtag/patch/tests/main.nf.test
```diff
@@ -0,0 +1,71 @@
...
+
+test("A. thaliana Col-0 test data - ragtag - patch") {
+
+        when {
+            process {
+                """
+                input[0] = [
+                    [ id:'test' ], // meta map
+                    file('https://raw.githubusercontent.com/nf-core/test-datasets/refs/heads/modules/data/genomics/sarscov2/genome/genome.fasta.gz', checkIfExists: true),
+                    file('https://raw.githubusercontent.com/nf-core/test-datasets/refs/heads/modules/data/genomics/sarscov2/genome/genome.fasta.gz', checkIfExists: true),
```
jfy133, line 21: Please use the the `modules_testdata_Base_path` pattern: https://nf-co.re/docs/guidelines/components/modules#input-data

(sorry just realised I didn't give that link in my original comment)

modules/nf-core/ragtag/scaffold/main.nf
```diff
@@ -0,0 +1,77 @@
...
+    tag "${meta.id}"
+    label 'process_medium'
+
+    conda "${moduleDir}/environment.yml"
+    container "${workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container
+        ? 'https://depot.galaxyproject.org/singularity/ragtag:2.1.0--pyhb7b1952_0'
+        : 'biocontainers/ragtag:2.1.0--pyhb7b1952_0'}"
+
+    input:
+    tuple val(meta), path(assembly, name: 'assembly/*'), path(reference, name: 'reference/*'), path(exclude), path(skip), path(hard_skip)
```
jfy133, line 11: Please add `metas` for the other files in separate input channels too rather than a single tuple.

jfy133 (commented):
Almost there (barring the timing out conda test)

modules/nf-core/ragtag/patch/main.nf
nschan: The reasoning is that this way you can give the same filename twice, since they one will go into `work/target/` and the other one will go into `work/query/`, and then they are unzipped / moved to `work/` with the name `target.fa` or `query.fa` such that their names no longer collide. Maybe I misunderstand what the issue you are referring to is though

modules/nf-core/ragtag/scaffold/main.nf
nschan: OK, I can do that. Can you clarify for me what the advantage of having more channels with their own meta is in this case? To me it makes sense to have the exclude file somehow linked to the assembly, since it probably needs to be crafted for each one individually, and I prefer (personally) to have fewer input channels, but maybe I am missing an advantage of having multiple input channels?

nschan: And to clarify more: should i do:

```
    input:
    tuple val(meta), path(assembly, name: 'assembly/*'), path(reference, name: 'reference/*')
    tuple val(meta2), path(exclude), path(skip), path(hard_skip)
```

or

```
    tuple val(meta), path(assembly, name: 'assembly/*'), path(reference, name: 'reference/*')
    tuple val(meta2), path(exclude), 
    tuple val(meta3), path(skip), path(hard_skip)
```

or something different?

jfy133: The latter, or but also spliut the `assembly` and `refernece` too

jfy133:
My feeling

```
    >   tail: 'test/ragtag.patch.err' has become inaccessible: No such file or directory
    ```
    
    It's to do with the funky `tail -F` thing, I would suggest trying to redirect with `>` at the end instead and see what happens (or try removing the logging for now and see if thehn it runs through fine)

File: modules/nf-core/ragtag/patch/main.nf (hunk 22 above)
Comment (jfy133) [reply modules/nf-core/ragtag/patch/main.nf:52]: I'm not following, you're creating a file with that name on line 50, no?

I guess my point is to try playing around with teh different methods (I don't actually know where this `tail -F` pattern has come from, I've not seen before :smile_sweat:

Comment (nschan) [reply modules/nf-core/ragtag/patch/main.nf]: I have to create that file because it does not exist before I start ragtag, and then `tail -F` complains that the file does not exist. Tbh, I don't really get how this is supposed to work with `tail -F` being called before the actual tool..

I do not think that this is causing the timeout problem, I also encountered it when I tried to test ragtag/patch locally with conda before adding the error redirection part.

**Comment by mashehu** [comment]
looks like you were right, @jfy133!

**Comment by jfy133** [comment]
Interesting... maybe that pattern isn't so robust then... and unfortuantely we have no blame history of that file because we did that big restructure :facepalm: 

I'm going to take a random guess.. @mahesh-panchal were you by any chance the author of: https://nf-co.re/docs/guidelines/components/modules#capturing-stdout-and-stderr ?

**Comment by mahesh-panchal** [comment]
> Interesting... maybe that pattern isn't so robust then... and unfortuantely we have no blame history of that file because we did that big restructure 🤦
> 
> I'm going to take a random guess.. @mahesh-panchal were you by any chance the author of: https://nf-co.re/docs/guidelines/components/modules#capturing-stdout-and-stderr ?

Yes, that was me: https://github.com/nf-core/website/pull/2136.

**Comment by nschan** [comment]
So, just to clarify `patch.err` is created by ragtag, I had to touch the file first to prevent `tail -F` from complaining; it is also emitted.

**Comment by mahesh-panchal** [comment]
Wait. Not the `tail -F` part.

**Comment by mahesh-panchal** [comment]
@pontus Any insight on why the `tail -F` was causing issues?

**Comment by pontus** [comment]
> @pontus Any insight on why the `tail -F` was causing issues?

Do you have a link to the check where this happened readily available? (I would expect the setup shown to give some ugly things in logs but expect them to be red herrings typically.)

**Comment by nschan** [comment]
Here is one @pontus https://github.com/nf-core/modules/actions/runs/14303829077/job/40083278515?pr=8198

File: modules/nf-core/ragtag/patch/main.nf [hunk 25]
```diff
@@ -0,0 +1,106 @@
...
+    else
+        cp ${target} target.fa
+    fi
+
+    if [[ ${query} == *.gz ]]
+    then
+        zcat ${query} > query.fa
+    else
+        cp ${query} query.fa
+    fi
```
jfy133, line 49: I still don't really follow the logic here, why load them into a directory as defined in `target` if you then just use the list of input fastas directly ? And again, didn't fully undersstand why you need to rename the files - is really trying to put in two files as `genome.fa` going to be so common? I would still rather just decompress with `gunzip`, and give them direcly to the command.

I would also not `cp` the reference file as for large reference genomes that's a lot of HDD space to use, maybe just symlink to get the correct name?

However this is not a blocker here.

modules/nf-core/ragtag/patch/tests/main.nf.test
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+    }
+    test("A. thaliana Col-0 test data - ragtag - patch - stub") {
+
+        options "-stub"
+
+        when {
+            process {
+"""
```
jfy133, line 60:
```suggestion
               """
```

jfy133 (approved):
One last comment but is nto a blocker, just optimisation suggestion

pontus:
@nschan Thanks, unfortunately, that doesn't make sense to me. 

My guess is it's failing because the nextflow process somehow doesn't end as expected when it reaches its end but rather hangs around waiting for the `tail` to go to away. If that's correct I think all backgrounding of persistent processes are likely to cause issues spuriously.

It would be interesting to see what happens if one kills the `tail` after it has served its purpose (e.g. as below).

```
tail -F something >&2 &
tailpid=$!
do stuff
...
kill -TERM "$tailpid"
```

pontus:
(I'm not sure what is best practice, this module seems to rely fairly heavily on the nf-core default options for `process.shell` to do error detection, should that be expected to be mentioned somewhere?)

modules/nf-core/ragtag/patch/main.nf
nschan: > I still don't really follow the logic here, why load them into a directory as defined in target if you then just use the list of input fastas directly ? And again, didn't fully undersstand why you need to rename the files - is really trying to put in two files as genome.fa going to be so common? I would still rather just decompress with gunzip, and give them direcly to the command.

I do not know how common it is that someone uses two files with the same name for both target and reference, but it doesn't hurt to support that I guess? Loading them into a different directory is purely to avoid having a file with the same name twice; not sure if there is something I am not getting about what you are asking?

> I would also not cp the reference file as for large reference genomes that's a lot of HDD space to use, maybe just symlink to get the correct name?

I have changed to symlinking.

nschan:
So it seems that killing `tail` when the tool ran, as suggested by @pontus, solves the issue. 

> (I'm not sure what is best practice, this module seems to rely fairly heavily on the nf-core default options for process.shell to do error detection, should that be expected to be mentioned somewhere?)

To be honest, I do not really understand; I was under the impression that these modules are in line with what most nf-core modules do?

PR #8230, opened 2025-04-07
Allow file-system image as GTDB-Tk database.

muniheart:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

In conjunction with [785](https://github.com/nf-core/mag/pull/785), this PR allows the GTDB-Tk database to be presented as a squash-fs image.  This reduces the inode requirements of the database from >200K inodes to 1 inode.  The dependent PR, [785](https://github.com/nf-core/mag/pull/785), is compatible with `singularity` and `apptainer` container engines.


jfy133:
@nf-core/maintainers any thoughts on this one?

Context:

- https://github.com/nf-core/mag/issues/779
- https://github.com/nf-core/mag/pull/785

TL;DR: the database is huge (>100 GB, with hundreds of files). 
This can cause issues on some systems, so the proposal is to also support a `squashfs` image instead of the database directory itself.
This, however, requires mounting the image to the container.

mahesh-panchal:
I've never encountered this before, so my naive question is the folder just a path that needs to be bind mounted into the container? How does this differ to providing a folder path, and that getting symlinked into the work directory?

jfy133:
> I've never encountered this before, so my naive question is the folder just a path that needs to be bind mounted into the container? How does this differ to providing a folder path, and that getting symlinked into the work directory?

If I understand correctly, it means that the squashfs image itself only has an inode size of 1 (a single file), rather than when unpacking the `tar.gz` results in thousands inodes

https://github.com/nf-core/mag/issues/779#issue-2947467733

mahesh-panchal:
> For this, I'd recommend rather making it something that's supported that the user/site does (e.g. just like you have references that can be provided locally for other pipelines).

I'm leaning towards this too. My initial thought was that this should be a institutional pipeline specific config.

jfy133:
Hah, funnily enough @prototaxites suggested a similar thing: https://github.com/nf-core/mag/pull/785

@muniheart what do you think?

Basically just means a config with:

```
process {
    withName: GTDBTK_CLASSIFYWF {
        containerOptions: "foo"
    }
}
```

The remaining issue would be to somehow change the database loading logic (gz vs directory) to be more permissive in mag, correct?

muniheart:
@mahesh-panchal, the image is a file-system within a file.  Mounting the file-system as a directory outside a container would require permission to mount a loop device.  Mounting inside a container does not require elevated permissions.

@jfy133, I have tried setting `containerOptions` in a configuration file, but I find that the delayed evaluation of environment variables, e.g. `NXF_TASK_WORKDIR` does not work as desired.   Also, the level-1 path within the db image would also have to be passed as an environment variable, I suspect.

muniheart:
Though the following works,

```
process {
        withName: foo {
                containerOptions = "-B $params.gtdb_db:\$NXF_TASK_WORKDIR/database:image-src=\$image_root"
        }
}

process foo {
        input:
            env image_root
}
```
it requires an input to specify the top-level directory inside image fs.  This top-level directory could be removed when the image is built.

muniheart:
I have decided to withdraw this pull request.  The container options can be passed through config as [@jfy133 ](https://github.com/nf-core/modules/pull/8230#issuecomment-2785790251) and [@prototaxites](https://github.com/nf-core/mag/pull/785#discussion_r2032673548) have suggested.

mahesh-panchal:
> @mahesh-panchal, the image is a file-system within a file. Mounting the file-system as a directory outside a container would require permission to mount a loop device. Mounting inside a container does not require elevated permissions.

Sorry, just for completion for me:

- So this is not compatible with using conda, but it's not a deal breaker in that respect. The module code could check for profile against the extra `input:` channel.
- and one would need to pass a string into the process for the database directory to be found. This means though the module would not be standalone in that the `containerOptions` would need to be supplied in the pipeline config. The meta.yml documentation might be enough though.

pontus:
I don't know how mag (which I think was the pipeline it popped up in) works with this database, but generally, I don't think anything should need to know about exactly how things are provided (e.g. is a directory bound in from a network file system or is it a mounted squash image) -  that should be up to the site.

For the pipelines/modules, I think it makes sense to support appropriate uses of (reference) data. I think someone mentioned `tar.gz` at some point (which seems kind of odd), but if that's the current way of accepting some database and then unpacking (appropriate bits?), it would probably make sense to make sure you could use a already unpacked version provided.

mahesh-panchal:
I agree. The only sticking point is the technical detail of how it's supplied as an input parameter to the process. Normally databases are File objects (type `path`) and then a symlink is created in the working directory. In the case of the mounted image, it should be passed as a String object (whether that be `env` or `val` type).

PR #8231, opened 2025-04-07
feat: add literal string to make contrasts in `variancepartition/dream`

atrigila:
POC of https://github.com/nf-core/differentialabundance/issues/386

Current implementation is:
- If `comparison` is provided, then it is passed as a literal string to `limma::makeContrasts`.
- Otherwise, the previous design remains.
- The `comparison` is only used in `inputs.contrasts_for_diff_with_formula`, which is the input to `VARIANCEPARTITION_DREAM` in `ABUNDANCE_DIFFERENTIAL_FILTER` for now. 

@grst , when you have a chance, could you take a look and let me know if this implementation makes sense? If anything needs updating, I’m happy to make the changes.

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/variancepartition/dream/main.nf
+++ b/modules/nf-core/variancepartition/dream/main.nf
@@ -8,7 +8,7 @@ process VARIANCEPARTITION_DREAM {
         'community.wave.seqera.io/library/bioconductor-edger_bioconductor-variancepartition_r-optparse:ba778938d72f30c5' }"
 
     input:
-    tuple val(meta), val(contrast_variable), val(reference), val(target), val(formula)
+    tuple val(meta), val(contrast_variable), val(reference), val(target), val(formula), val(comparison)
     tuple val(meta2), path(samplesheet), path(counts)
 
     output:
--- a/modules/nf-core/variancepartition/dream/templates/dream.R
+++ b/modules/nf-core/variancepartition/dream/templates/dream.R
@@ -53,6 +53,7 @@ opt <- list(
     contrast_variable  = "$contrast_variable",      # Variable for contrast (e.g., "treatment")
     contrast_reference = "$reference",              # Reference level for the contrast
     contrast_target    = "$target",                 # Target level for the contrast (e.g., "mCherry")
+    contrast_string    = "$comparison",             # Optional full (complex) contrast expression comparison
     sample_id_col      = "sample",                  # Column name for sample IDs
     threads            = "$task.cpus",              # Number of threads for multithreading
     subset_to_contrast_samples = FALSE,            # Whether to subset to contrast samples
@@ -85,6 +86,11 @@ for (ao in names(args_opt)) {
 # If there is no formula, convert string "null" to NULL
 if (!is.null(opt\$formula) && tolower(opt\$formula) == "null") {
     opt\$formula <- NULL
+}
+
+# If there is no contrast string, convert string "null" to NULL
+if (!is.null(opt\$contrast_string) && tolower(opt\$contrast_string) == "null") {
+    opt\$contrast_string <- NULL
 }
 
 # Load metadata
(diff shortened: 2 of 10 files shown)


modules/nf-core/variancepartition/dream/templates/dream.R
```diff
@@ -147,13 +153,28 @@ fitmm <- eBayes(fitmm, proportion = opt\$proportion,
 head(fitmm\$design, 3)
 print(colnames(fitmm\$design))
 
-# Write coefficient name from contrast variable and target level (e.g., "treatmentmCherry")
-coef_name <- paste0(opt\$contrast_variable, opt\$contrast_target)
+# If contrast_string is provided, use that for makeContrast
+if (!is.null(opt\$contrast_string) && opt\$contrast_string != "") {
```
grst, line 157: Maybe move the check for `""` up and also set it to NULL in that case?

modules/nf-core/variancepartition/dream/templates/dream.R
```diff
@@ -147,13 +153,28 @@ fitmm <- eBayes(fitmm, proportion = opt\$proportion,
...
-results <- topTable(fitmm, coef = coef_name,
-                    adjust.method = opt\$adjust.method, p.value = opt\$p.value,
-                    lfc = opt\$lfc, confint = opt\$confint)
+} else {
+    coef_name <- paste0(opt\$contrast_variable, opt\$contrast_target)
+    cat("Using default contrast matrix:", coef_name, "\n")
+
+    results <- topTable(fitmm, coef = coef_name,
+                        adjust.method = opt\$adjust.method, p.value = opt\$p.value,
+                        lfc = opt\$lfc, confint = opt\$confint)
```
grst, line 176: Is there a reason why you specify `number` explicitly in one case but not the other?

subworkflows/nf-core/abundance_differential_filter/main.nf
```diff
@@ -138,7 +138,7 @@ workflow ABUNDANCE_DIFFERENTIAL_FILTER {
 
     // DREAM only runs with formula
     dream_inputs = inputs.contrasts_for_diff_with_formula
-        .filter { meta, variable, reference, target, formula ->
+        .filter { meta, _variable, _reference, _target, formula, _comparison ->
```
grst, line 141: The module supports both, doesn't it? 
(You wouldn't be able to specify mixed effects, but in that case DREAM just behaves like limma)

grst (commented):
Few, minor things

subworkflows/nf-core/abundance_differential_filter/main.nf
atrigila: Yes, the module can support both. However for this version of the subworkflow, I would prefer to focus only on supporting the formula.

modules/nf-core/variancepartition/dream/main.nf
```diff
@@ -8,7 +8,7 @@ process VARIANCEPARTITION_DREAM {
         'community.wave.seqera.io/library/bioconductor-edger_bioconductor-variancepartition_r-optparse:ba778938d72f30c5' }"
 
     input:
-    tuple val(meta), val(contrast_variable), val(reference), val(target), val(formula)
+    tuple val(meta), val(contrast_variable), val(reference), val(target), val(formula), val(comparison)
```
pinin4fjords, line 11: Convention in nf-core is for non-mandatory non-file things to go in via ext.args. Otherwise people end up having to supply '[]' or '' or other messy things.

modules/nf-core/variancepartition/dream/templates/dream.R
```diff
@@ -147,13 +153,29 @@ fitmm <- eBayes(fitmm, proportion = opt\$proportion,
 head(fitmm\$design, 3)
 print(colnames(fitmm\$design))
 
-# Write coefficient name from contrast variable and target level (e.g., "treatmentmCherry")
-coef_name <- paste0(opt\$contrast_variable, opt\$contrast_target)
+# If contrast_string is provided, use that for makeContrast
+if (!is.null(opt\$contrast_string)) {
```
pinin4fjords, line 157: Just to check, if `$comparison` is NOT supplied, does `opt$contrast_string` really end up as a NULL?

modules/nf-core/variancepartition/dream/main.nf
pinin4fjords: Doing that would also mean you don't need to change the workflow logic, you can pass it from the meta when it's available.

pinin4fjords (changes requested):
Making optional things mandatory in module and subworkflow inputs complicates workflow logic. Suggest this new thing goes via `ext.args`, passed from `meta` at the workflow level.

This is how blocking variables are handled in deseq2, for example.

modules/nf-core/variancepartition/dream/main.nf
pinin4fjords: Yes, I suppose that's true. But it might be simpler to leave those in place for now and resolve separately. I'm mostly concerned with not adding additional subworkflow complexity for this.

modules/nf-core/variancepartition/dream/templates/dream.R
atrigila: The idea would be that if `make_contrast_str` is not supplied it is indeed null, because that's taken care of in the main workflow: 
https://github.com/nf-core/differentialabundance/blob/80c3ba1ddce8d7f843d3d4b6e8aeaa718f416135/workflows/differentialabundance.nf#L380
In case it is not supplied it will use the classic approach.

modules/nf-core/variancepartition/dream/main.nf
pinin4fjords: @atrigila @grst I chatted with the maintainers, and the consensus was that in a situation where e.g. A + B OR C + D were required, a sensible thing might be to have two separate input channels. But this is not the case here, since these can be somewhat overlapping (right)?

Let's keep your version for now, I hadn't considered this point of Gregor.

modules/nf-core/variancepartition/dream/templates/dream.R
pinin4fjords: OK, let's assume that works, it's been a while since I've tinkered with this args stuff

pinin4fjords (approved):
I think we may end up adjusting this stuff to more cleanly handle the different ways of specifying contrasts. But I think it's OK to run with this for now and see how you go

modules/nf-core/variancepartition/dream/main.nf
grst: In a future version they'll be mutually exclusive. But sounds good to keep it as is for the current iteration.

PR #8312, opened 2025-04-15
longstitch module, tests

nschan:
This is a module for the scaffolding tool [`longstitch`](https://github.com/bcgsc/longstitch), part of the [genomeassembler local module migration](https://github.com/nf-core/genomeassembler/issues/110).

I was unable to locate any description of the expected outputs, which also depend on the `command` used, so this is a best effort based on a bunch of tests and guessing. The output names are presumably also dependent on parameters such as `w`, `k_ntLink ` and `z`, making it difficult to come up with good patterns to catch them.


modules/nf-core/longstitch/main.nf
```diff
@@ -0,0 +1,153 @@
...
+    tuple val(meta), path("*k*.w*.z*.stitch.path"),                                         emit: stitch_path,                          optional: true
+    tuple val(meta), path("*k*.w*.z*.trimmed_scafs.agp"),                                   emit: trimmed_scaffolds_agp,                optional: true
+    tuple val(meta), path("*k*.w*.z*.trimmed_scafs.fa"),                                    emit: trimmed_scaffolds_fasta,              optional: true
+    tuple val(meta), path("*k*.w*.z*.trimmed_scafs.path"),                                  emit: trimmed_scaffolds_path,               optional: true
+    tuple val(meta), path("*k*.w*.z*.trimmed_scafs.tsv"),                                   emit: trimmed_scaffolds_tsv,                optional: true
+    tuple val(meta), path("*k*.w*.z*.verbose_mapping.tsv"),                                 emit: verbose_mapping_tsv,                  optional: true
+    tuple val(meta), path("*.k*.w???.tsv"),                                                 emit: tsv,                                  optional: true
+    path "versions.yml",                                                                    emit: versions
```
ggabernet, line 43: That's a lot of outputs, would it make sense maybe to just output the whole archive containing all these files? Or do you need specific files for specific processes only, and the archive is too big?

ggabernet (commented):
The conda tests seem to be failing, but otherwise looks good to me!

modules/nf-core/longstitch/main.nf
nschan: Personally I think most of these files are not very useful, but as I understand nf-core modules should (must?) emit all outputs produced by the tool.

ggabernet: it's true that it's recommended to output them all to ensure that everyone else can use the module. I also personally think in this case it's a bit of an overkill but I'm also happy to go with this as this is the recommendation.

ggabernet (approved):
Looks good to me!

nschan:
The conda test fails because `versions.yml` has a different hash. That is kind of unexpected.

modules/nf-core/longstitch/main.nf
SPPearce: I would agree with @ggabernet here, it isn't essential to output all these files individually, especially if you don't think that they have value (or that you'd always want all of them together). You could emit a global pattern like `"*"` and call it `files`, and then seperately the few files that are the typically wanted output. Or something like `*.ntLink.scaffolds.*` and call it `scaffolds` for instance.

Someone else can always update the module later if they really want different output.

PR #8337, opened 2025-04-21
Add jellyfish count

vagkaratzas:
## PR checklist

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/jellyfish/count/meta.yml
+++ b/modules/nf-core/jellyfish/count/meta.yml
@@ -1,0 +1,57 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/meta-schema.json
+name: "jellyfish_count"
+description: Efficiently counts k-mers from DNA sequencing reads using a fast, memory-efficient,
+  parallelized algorithm
+keywords:
+  - k-mer
+  - DNA
+  - substrings
+tools:
+  - "jellyfish":
+      description: "Jellyfish is a tool for fast, memory-efficient counting of k-mers\
+        \ in DNA. A k-mer is a substring of length k, and counting the occurrences of\
+        \ all such substrings is a central step in many analyses of DNA sequence"
+      homepage: "https://github.com/gmarcais/Jellyfish"
+      documentation: "https://github.com/gmarcais/Jellyfish/blob/master/doc/Readme.md"
+      tool_dev_url: "https://github.com/gmarcais/Jellyfish"
+      doi: "10.1093/bioinformatics/btr011"
+      licence: ["GPL v3"]
+      identifier: biotools:jellyfish
+input:
+  - - meta:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. `[ id:'sample1', single_end:false ]`
+    - fasta:
+        type: file
+        description: Nucleotide sequences in FASTA format
+        pattern: "*.{fasta,fa,fna,faa}"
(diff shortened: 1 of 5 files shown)


modules/nf-core/jellyfish/count/meta.yml
```diff
@@ -0,0 +1,57 @@
...
+        description: Nucleotide sequences in FASTA format
+        pattern: "*.{fasta,fa,fna,faa}"
+        ontologies:
+          - edam: http://edamontology.org/format_1929 # FASTA
+  - - kmer_length:
+        type: integer
+        description: k-mer size to use
+  - - size:
+        type: string
+        description: Specifies the memory hash size (in bytes) to allocate for the k-mer counting hash table
```
SPPearce, line 37: I don't really understand this, what does this mean and why do I have to specify it rather than it being set automatically?

vagkaratzas: The `-s` option tells Jellyfish how big the internal hash table should be; that's the data structure it uses to store k-mers. It is a mandatory value that the user might want to set themselves, depending on their k-mer length and fasta size. Since it's mandatory, I had to make it a val input (instead of setting it through args). More documentation on it [here](https://github.com/gmarcais/Jellyfish/blob/master/doc/Readme.md#counting-k-mers-in-sequencing-reads).

SPPearce: If it just controls the memory, is there a reason not to just set it to be proportional to the memory available to the task?

vagkaratzas: I get your point! I am just wrapping their original functionality in an nf-core module. Since they implemented it this way, I guess there could be special scenarios where the user might need to handpick that. There is no formula on their wiki on how to calculate this number.

nschan: This is a bit of a weird input, I agree. I wonder if there is a reasonable way to set a default, if the user also does not know what to provide here? The documentation states that it should be roughly `genome size + (genome size * coverage * kmer-length * error rate)`, maybe it would be worth adding that here? Since jellyfish grows the hashtable if it is too small, setting this correctly does not seem extremely important.

modules/nf-core/jellyfish/count/meta.yml
```diff
@@ -0,0 +1,57 @@
...
+      tool_dev_url: "https://github.com/gmarcais/Jellyfish"
+      doi: "10.1093/bioinformatics/btr011"
+      licence: ["GPL v3"]
+      identifier: biotools:jellyfish
+input:
+  - - meta:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. `[ id:'sample1', single_end:false ]`
```
nschan, line 25: I think `single_end:false` should only be part of the example meta if it actually used for something

nschan:
I have approved the changes, but I would appreciate if you could address the minor comments on meta.yml before merging

modules/nf-core/jellyfish/count/meta.yml
vagkaratzas: Alright, I will add this formula to the meta.yml file as well

modules/nf-core/jellyfish/count/meta.yml
vagkaratzas: Defintely not used in the test here, so removing it

PR #8340, opened 2025-04-21
Swap more sentieon modules to nf-test

SPPearce:
More sentieon modules swapped to nf-test


modules/nf-core/sentieon/applyvarcal/main.nf
```diff
@@ -22,19 +22,24 @@ process SENTIEON_APPLYVARCAL {
     task.ext.when == null || task.ext.when
 
     script:
-    def args = task.ext.args ?: ''
-    def prefix = task.ext.prefix ?: "${meta.id}"
+    def args   = task.ext.args   ?: ''
+    def args2   = task.ext.args   ?: ''
```
atrigila, line 26: `args` and `args2` are both initialized from the same source (task.ext.args) and used twice in the script block. Is this correct?

atrigila: I think it should be `task.ext.args2`?

atrigila (commented):
Minor comment

modules/nf-core/sentieon/applyvarcal/main.nf
SPPearce: Thanks for spotting

SPPearce:
```suggestion
    def args   = task.ext.args   ?: ''
    def args2  = task.ext.args2  ?: ''
```

atrigila (approved):

PR #8349, opened 2025-04-22
updated for 9.5, remove gfastats from meta, fixed version output

DLBPointon:
- Updates coreutils to 9.5.
- Removes gfastats link from meta.yml
- Corrected version output for sort.
- Updated snapshots

### DIFF
--- a/modules/nf-core/gnu/sort/meta.yml
+++ b/modules/nf-core/gnu/sort/meta.yml
@@ -3,13 +3,18 @@ description: |
   Writes a sorted concatenation of file/s
 keywords:
   - GNU
+  - coreutils
   - sort
   - merge compare
 tools:
-  - sort:
-      description: "Writes a sorted concatenation of file/s"
-      homepage: "https://github.com/vgl-hub/gfastats"
-      documentation: "https://www.gnu.org/software/coreutils/manual/html_node/sort-invocation.html"
+  - "gnu":
+      description: "The GNU Core Utilities are the basic file, shell and text manipulation
+        utilities of the GNU operating system. These are the core utilities which are
+        expected to exist on every operating system."
+      homepage: "https://www.gnu.org/software/coreutils/"
+      documentation: "https://www.gnu.org/software/coreutils/manual/html_node/index.html"
+      tool_dev_url: "https://git.savannah.gnu.org/cgit/coreutils.git"
+      doi: "10.5281/zenodo.581670"
       licence: ["GPL"]
       identifier: ""
 input:
--- a/modules/nf-core/gnu/split/main.nf
+++ b/modules/nf-core/gnu/split/main.nf
@@ -4,8 +4,8 @@ process GNU_SPLIT {
 
     conda "${moduleDir}/environment.yml"
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
-        'https://depot.galaxyproject.org/singularity/coreutils:9.3':
-        'biocontainers/coreutils:9.3' }"
+        'https://depot.galaxyproject.org/singularity/coreutils:9.5':
+        'biocontainers/coreutils:9.5' }"
 
     input:
     tuple val(meta), path(input)
--- a/modules/nf-core/gnu/sort/environment.yml
+++ b/modules/nf-core/gnu/sort/environment.yml
@@ -4,4 +4,4 @@ channels:
   - conda-forge
   - bioconda
 dependencies:
-  - conda-forge::coreutils=9.3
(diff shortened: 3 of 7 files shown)


mahesh-panchal:
For gnu/split, update the glob to use `prefix` instead of `*`.

modules/nf-core/gnu/split/meta.yml
```diff
@@ -33,15 +33,15 @@ output:
...
+          pattern: "${outfile_prefix}.*"
   - versions:
       - versions.yml:
           type: file
-          description: The split files
+          description: File containing software versions
+          pattern: "versions.yml"
 authors:
   - "@k1sauce"
-maintainers:
```
mirpedrol, line 46: Was `maintainers` deleted on purpose? If so, you should delete the github handle, it is duplicated now

modules/nf-core/gnu/sort/meta.yml
```diff
@@ -29,6 +34,13 @@ output:
           description: |
             Groovy Map containing sample information
             e.g. [ id:'test', single_end:false ]
+          pattern: "${output_file}"
+      - ' "${output_file}" ':
```
mirpedrol, line 38:
```suggestion
      - '${output_file}':
```

modules/nf-core/gnu/split/meta.yml
```diff
@@ -33,15 +33,15 @@ output:
           description: |
             Groovy Map containing sample information
             e.g. `[ id:'sample1', single_end:false ]`
-      - ' "*split*" ':
+      - ' "${outfile_prefix}.*" ':
```
mirpedrol, line 36:
```suggestion
      - '${outfile_prefix}.*':
```

DLBPointon: Doing that will fail the nf-core lint

╭─ [✗] 1 Module Test Failed ─────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│              ╷                                    ╷                                                                                    │
│ Module name  │ File path                          │ Test message                                                                       │
│╶─────────────┼────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────╴│
│ gnu/split    │ modules/nf-core/gnu/split/meta.yml │ Module meta.yml does not match main.nf. Outputs should contain: {'split': ['meta', │
│              │                                    │ ' "${outfile_prefix}.*" '], 'versions': ['versions.yml']} Run nf-core modules lint │
│              │                                    │ --fix to update the meta.yml file.                                                 │
│              ╵                                    ╵                                                                                    │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

DLBPointon:
Ok, so now it's failing because ULTRA/PIPELINE isn't generating the output that is expected.

DLBPointon:
I can confirm the CI

The conda run with ULTRA_PIPELINE is failing, although it is failing consistently. Which tells me that is more likely an issue with ULTRA?

The fact that none of the GNU specific stuff is failing gives me confidence.

What do people think would be the best course of action?

modules/nf-core/gnu/split/meta.yml
DLBPointon: No problem, adding it back in now. 

Only issue now is ULTRA_PIPELINE

maxulysse (approved):

DLBPointon:
Hi all,

Apologies for letting this get forgotten about.
@maxulysse could you give your stamp of approval again (if it is worthy), as things have changed and I wouldn't want to presume the previous one still stands. 

The 1 error is still from the ULTRA module that uses  GNU_SORT.

Thanks.

DLBPointon:
Nice, thanks for updating this and ULTRA @SPPearce

modules/nf-core/gnu/split/main.nf
```diff
@@ -4,28 +4,29 @@ process GNU_SPLIT {
...
     task.ext.when == null || task.ext.when
 
     script:
-    def args   = task.ext.args   ?: ''
-    def prefix = task.ext.prefix ?: "${meta.id}"
-    def suffix = input.extension
+    def args        = task.ext.args   ?: ''
+    def prefix      = task.ext.prefix ?: "${meta.id}"
+    def suffix      = input.extension
+    outfile_prefix  = "${prefix}.split"
```
SPPearce: Could this not just be:
```suggestion
    prefix      = task.ext.prefix ?: "${meta.id}.split"
    def suffix      = input.extension
```

DLBPointon: It can be indeed, just updated it. I was trying not to change much of the original module is all.

PR #8359, opened 2025-04-23
Swap sequencetools/pileupcaller to nf-test

SPPearce:


modules/nf-core/sequencetools/pileupcaller/main.nf
```diff
@@ -30,6 +30,7 @@ process SEQUENCETOOLS_PILEUPCALLER {
     freqsum_output = "-e" in args_list || "--eigenstratOut" in args_list || "-p" in args_list || "--plinkOut" in args_list ? '' : "| gzip -c > ${prefix}.freqsum.gz"
 
     """
+    echo $args
```
LouisLeNezet, line 33:
```suggestion
```

LouisLeNezet: Was it only for debugging ?

modules/nf-core/sequencetools/pileupcaller/main.nf
```diff
@@ -42,4 +43,23 @@ process SEQUENCETOOLS_PILEUPCALLER {
         sequencetools: \$(echo \$(pileupCaller --version 2>&1) )
     END_VERSIONS
     """
+
+    stub:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def args_list = args.tokenize()
```
LouisLeNezet, line 50: We should definitely have some kind of standard in nf-core to check all the parameters.
As for now everyone has its own way but we should find a way to harmonize this.
I really like `args.tokenize()`

LouisLeNezet (approved):
Just a small deletion maybe.

modules/nf-core/sequencetools/pileupcaller/tests/main.nf.test
```diff
@@ -0,0 +1,123 @@
...
+    name "Test Process SEQUENCETOOLS_PILEUPCALLER"
+    config "./nextflow.config"
+    script "../main.nf"
+    process "SEQUENCETOOLS_PILEUPCALLER"
+   
+    tag "modules"
+    tag "modules_nfcore"
+    tag "sequencetools"
+    tag "sequencetools/pileupcaller"
+    config './nextflow.config'
```
SPPearce, line 12:
```suggestion
```

SPPearce:
Thanks Louis, yes, I was debugging with the ext.args echo.

PR #8362, opened 2025-04-24
Update mapAD to v0.45.0

jch-13:
Bump mapAD version from `v0.44.1` to `v0.45.0`

### DIFF
--- a/subworkflows/nf-core/fastq_align_mapad/tests/main.nf.test.snap
+++ b/subworkflows/nf-core/fastq_align_mapad/tests/main.nf.test.snap
@@ -9,7 +9,7 @@
                         "key_read_ref": "test_genome",
                         "id_index": "genome"
                     },
-                    "test.sorted.bam.bai:md5,d7ed154ff09b8d06024af523a45f3ed2"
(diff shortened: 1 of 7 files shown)


subworkflows/nf-core/fastq_align_mapad/tests/main.nf.test.snap
```diff
@@ -9,7 +9,7 @@
                         "key_read_ref": "test_genome",
                         "id_index": "genome"
                     },
-                    "test.sorted.bam.bai:md5,d7ed154ff09b8d06024af523a45f3ed2"
+                    "test.sorted.bam.bai:md5,a14a162090589149cfa5813dee5a6fd2"
```
famosab, line 12: do you know why the bai changes its md5sum for the updated version? 🤔

famosab (approved):
LGTM (only the bai is changed which is sensible I think)

subworkflows/nf-core/fastq_align_mapad/tests/main.nf.test.snap
jch-13: We don't directly test output BAM files for equality as they are sometimes flaky. With the updated mapAD version, its output may change slightly, which may be enough to change the BAI index.

PR #8365, opened 2025-04-25
add new module rtgtools/bndeval

kubranarci:
## PR checklist

Closes #XXX 

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:


modules/nf-core/rtgtools/bndeval/main.nf
```diff
@@ -0,0 +1,79 @@
...
+    def args = task.ext.args ?: ""
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def bed_regions = regions_bed ? "--bed-regions=${regions_bed}" : ""
+    def truth_index = truth_vcf_tbi ? "" : "rtg index ${truth_vcf}"
+    def query_index = query_vcf_tbi ? "" : "rtg index ${query_vcf}"
+    def avail_mem = task.memory.toGiga() + "G"
+
+    """
+    ${truth_index}
+    ${query_index}
```
ggabernet, line 39: I was wondering whether this should be put into an independent module instead. Would there be a reason to add the index building as part of the same module?

ggabernet (commented):
I just had a comment but otherwise looks good to me

modules/nf-core/rtgtools/bndeval/main.nf
kubranarci: This was the case for vcfeval, but running it in the same process would make the process faster.  Also, index can also be given as input so this is optional anyways.

modules/nf-core/rtgtools/bndeval/main.nf
```diff
@@ -0,0 +1,79 @@
...
+    tag "$meta.id"
+    label 'process_medium'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/rtg-tools:3.12.1--hdfd78af_0':
+        'biocontainers/rtg-tools:3.12.1--hdfd78af_0' }"
+
+    input:
+    tuple val(meta), path(query_vcf), path(query_vcf_tbi), path(truth_vcf), path(truth_vcf_tbi), path(regions_bed)
```
jfy133, line 11: **question:** Are the query and truth VCFs truly mutually inclusive [^1]? Otherwise I would split into two or three channels: one for query vcf, one for truth vcf and one for the regions bed (all with meta)

[^1]: https://nf-co.re/docs/guidelines/components/modules#required-path-channel-inputs

modules/nf-core/rtgtools/bndeval/main.nf
jfy133: I agree with @ggabernet that the indexing should go into a different module as you're not piping here [^1].

Even if it's faster it's not really how we do things - particularly if it's optional. Presumably you would reuse the index anyway (so would generate it upfront), and maybe you want the index without running the whole `bndeval` module.

[^1]: https://nf-co.re/docs/guidelines/components/modules#use-of-multi-command-piping

modules/nf-core/rtgtools/bndeval/main.nf
```diff
@@ -0,0 +1,79 @@
...
+    """
+    ${truth_index}
+    ${query_index}
+
+    rtg RTG_MEM=$avail_mem bndeval \\
+        ${args} \\
+        --baseline=${truth_vcf} \\
+        ${bed_regions} \\
+        --calls=${query_vcf} \\
+        --output=output
```
jfy133, line 46: **suggestion:** 
```suggestion
        --output=${prefix}
```

Unless the use of the `output` string is mandatory?

modules/nf-core/rtgtools/bndeval/meta.yml
```diff
@@ -0,0 +1,185 @@
+name: "rtgtools_bndeval"
+description: The bndeval tool of RTG tools. It is used to evaluate called BND type of variants
+  for agreement with a BND baseline variant set
+keywords:
+  - benchmarking
+  - vcf
+  - rtg-tools
```
jfy133, line 7: **todo:** please add more useful keywords, e.g evaluation, bnd-type, etc. From the current ones I would not understand what the module is used for [^1]

[^1]: https://nf-co.re/docs/guidelines/components/modules#number-of-keywords

modules/nf-core/rtgtools/bndeval/meta.yml
```diff
@@ -0,0 +1,185 @@
...
+tools:
+  - "rtgtools":
+      description: "RealTimeGenomics Tools -- Utilities for accurate VCF comparison
+        and manipulation"
+      homepage: "https://www.realtimegenomics.com/products/rtg-tools"
+      documentation: "https://cdn.jsdelivr.net/gh/RealTimeGenomics/rtg-tools@master/installer/resources/tools/RTGOperationsManual.pdf"
+      tool_dev_url: "https://github.com/RealTimeGenomics/rtg-tools"
+      licence: ["BSD"]
+      identifier: ""
```
jfy133, line 16: **suggestion:** I guess can be removed?
```suggestion

```

modules/nf-core/rtgtools/bndeval/main.nf
kubranarci: It depends on the usage, but now all the benchmarking tools we have works in this logic. Truth files has to match query (must originate from the same sample). 

If we seperate those, then we cannot run multiple truths with multiple queries basically.

modules/nf-core/rtgtools/bndeval/main.nf
kubranarci: But indexing the output is recommended? Here, I am not sure since I never used myself but, indexing with rtgtools might be required for specific callers. https://github.com/nf-core/modules/blob/7b5d8ecb065e4eba0e3dd59653f6011b9470b258/modules/nf-core/rtgtools/vcfeval/main.nf does the same. @nvnieuwk is there any particular reason you made the implementation like that?

modules/nf-core/rtgtools/bndeval/main.nf
jfy133: You can still pass them so the come together using a `multiMap`. Which other benchmarking tols are you referring to?

kubranarci: truvari, svbenchmark, wittyer, happy, sompy .. basically all the becnhmarking tools my pipeline uses

modules/nf-core/rtgtools/bndeval/main.nf
kubranarci: It is indexing the input, but it is using **rtgtools index** not tabix - so not a different tool.

modules/nf-core/rtgtools/bndeval/main.nf
jfy133: I see... OK looking through the modules that is indeed the case...

So you can keep the single tuple for now, but I will complain at the previous authors/reviewers (@nvnieuwk @FriederikeHanssen @SPPearce _shakesfist_),

modules/nf-core/rtgtools/bndeval/main.nf
jfy133: It's not about the tool but rather keeping to the strict structure we aim to have of what goes in a module.

Please remove it as indexing, if not done on the output of an upstream module, should be executed as a standalone module. This means you keep consistency with the majority of other nf-core modules.

 You can ignore @nvnieuwk's 'precedence', he should go back and correct the vcfeval module ;)

nvnieuwk: But not a hill I am willing to die on

modules/nf-core/rtgtools/bndeval/main.nf
nvnieuwk: In my specific use case I want to validate multiple samples in one pipeline. Each sample has it's own query VCF and truth VCF so they cannot be on separate input lines for my use case to work :)

nvnieuwk: The validation region also depends on the ROI given to the specific sample

modules/nf-core/rtgtools/bndeval/main.nf
kubranarci: As the main user of those tools, I would argue that too. Yet, I will change it.

jfy133 (approved):
Thanks for bearing with my nitpicking @kubranarci !

modules/nf-core/rtgtools/bndeval/main.nf
nvnieuwk: IMO the nextflow modules should be as flexible as possible to allow for most use cases to work but that's a whole other discussion :grin:

ggabernet:
@kubranarci @jfy133 @nvnieuwk  sorry for the late reply, and thanks @jfy133 for taking over the review. I also kinda agree that if it makes it easier for the user it could be combined inside the same module, but I guess then the modules guidelines should be adapted for that. I also agree that this also saves computational time as there is an overhead when running this in the cluster and requesting resources twice for processing the same sample with one tool.

PR #8366, opened 2025-04-25
change links test-data to sarscov

nschan:
Following [some discussion on slack](https://nfcore.slack.com/archives/C02L5UB4Y9G/p1745578136387789), `scaffolding/` should be removed from `test-datasets/data/genomics/`. This module now uses sarscov2 data for testing.

### DIFF
--- a/modules/nf-core/links/tests/main.nf.test.snap
+++ b/modules/nf-core/links/tests/main.nf.test.snap
@@ -1,4 +1,66 @@
 {
+    "LINKS - sarscov2 test data - scaffolds": {
+        "content": [
+            "test.log",
+            [
+                [
+                    {
+                        "id": "test"
+                    },
+                    "test.pairing_issues:md5,d41d8cd98f00b204e9800998ecf8427e"
+                ]
+            ],
+            [
+                [
+                    {
+                        "id": "test"
+                    },
+                    "test.scaffolds:md5,095cc323b3af3a7873c8b80cf3736a1f"
+                ]
+            ],
+            [
+                [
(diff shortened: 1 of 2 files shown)


modules/nf-core/links/tests/main.nf.test.snap
```diff
@@ -1,4 +1,66 @@
...
+                    },
+                    "test.assembly_correspondence.tsv:md5,a65d30663dce705d382df52ab87ca8a4"
+                ]
+            ],
+            [
+                [
+                    {
+                        "id": "test"
+                    },
+                    "test.tigpair_checkpoint.tsv:md5,d41d8cd98f00b204e9800998ecf8427e"
```
jfy133, line 51: **question (non-blocking):** is this meant to be empty?

nschan: I do not know if it is _meant_ to be, but for this test it is, for the other one it contains two lines. It seems that during this test not much happens (scaffolds get renamed, that seems to be it?), so presumably there is no checkpointing done. 
I however realized something odd, the log contains this:
```
Scaffolding mid-point files:
test.tigpair_checkpoint.tsv
test.simplepair_checkpoint.tsv
detected; LINKS will skip contig pairing and use info from files instead
```

I had assumed that the workdirectory would be cleared before running the test, is that not the case?

jfy133: I thought it should be cleared too...

jfy133 (approved):

PR #8379, opened 2025-04-29
Allow iqtree to work with multiple input alignments

zachary-foster:
I also updated the iqtree version

### DIFF
--- a/modules/nf-core/iqtree/main.nf
+++ b/modules/nf-core/iqtree/main.nf
@@ -4,8 +4,8 @@ process IQTREE {
 
     conda "${moduleDir}/environment.yml"
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
-        'https://depot.galaxyproject.org/singularity/iqtree:2.3.4--h21ec9f0_0' :
-        'biocontainers/iqtree:2.3.4--h21ec9f0_0' }"
+        'https://depot.galaxyproject.org/singularity/iqtree:2.4.0--h503566f_0' :
+        'biocontainers/iqtree:2.4.0--h503566f_0' }"
 
     input:
     tuple val(meta), path(alignment), path(tree)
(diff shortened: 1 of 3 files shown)


modules/nf-core/iqtree/main.nf
atrigila, line 1: Ensure any changes to inputs (like alignments can now be a list) are reflected in the documentation (meta.yml).

modules/nf-core/iqtree/tests/main.nf.test.snap
```diff
@@ -17,21 +17,21 @@
                 ""
             ],
             "5",
-            "IQ-TREE 2.3.4 COVID-edition built May 23 2024",
-            "IQ-TREE multicore version 2.3.4 COVID-edition for Linux x86 64-bit built May 23 2024",
+            "IQ-TREE 2.4.0 built Feb 12 2025",
+            "IQ-TREE multicore version 2.4.0 for Linux x86 64-bit built Feb 12 2025",
```
atrigila, line 21: https://nf-co.re/docs/guidelines/components/modules#emission-of-versions

All reported versions must start with a numeric character. So I would remove everything except for: "2.4.0".

modules/nf-core/iqtree/main.nf
```diff
@@ -52,7 +52,7 @@ process IQTREE {
 
     script:
     def args                        = task.ext.args           ?: ''
-    def alignment_arg               = alignment               ? "-s $alignment"                 : ''
+    def alignment_arg               = alignment               ? "-s ${alignment.join(',')}"     : ''
```
atrigila, line 55: Nice to have (but not a blocker): a test case with multiple input alignments :)

atrigila (changes requested):

modules/nf-core/iqtree/main.nf
zachary-foster: Oh yea, good idea. Done.

modules/nf-core/iqtree/tests/main.nf.test.snap
zachary-foster: Are you talking about whats in the `versions.yml`? That is already just `2.4.0`. I'm not sure where that info in the snap is coming from, perhaps standard out/error.

modules/nf-core/iqtree/main.nf
zachary-foster: Yes, I looked into that but could not find test data that would work already in `test-datasets`. We would need two alignments with different names but the same headers.

modules/nf-core/iqtree/tests/main.nf.test.snap
atrigila: Indeed, you are right!

atrigila (approved):
Looks good!

zachary-foster:
Thanks for reviewing!

PR #8389, opened 2025-05-02
New module: gtdbtk/gtdbtoncbimajorityvote

prototaxites:
Adds a new module: gtdbtk/gtdbtoncbimajorityvote

This requires a couple of changes to the gtdbtk/classifywf module:
- The new module takes as input the files from classifywf, and requires that the structure be as-is. So I've removed the code that re-arranges and tidies up the classifywf outputs
- I've added a new output that outputs everything from gtdbtk/classifywf
- Harshill alignment ☺️
- Bump version to 2.4.1

For the new module, one of the inputs is the prefix used in the gtdbtk/classifywf run - I'm currently just inferring this from the input files. But it could be an explicit input if anyone thinks that's better

Note: It would be good to move the mockup database to somewhere in the nf-core test data sphere, as the download speed is in the range of a handful of kbps. 

## PR checklist

Closes #XXX 

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:


prototaxites:
@jfy133 Since you added the tests the other day - better to leave the empty file in the classifywf test despite causing a lint failure?

modules/nf-core/gtdbtk/gtdbtoncbimajorityvote/main.nf
```diff
@@ -0,0 +1,54 @@
...
+    bac120            = bac120_metadata ? "--bac120_metadata_file ${bac120_metadata}" : ""
+    ar53              = ar53_metadata   ? "--ar53_metadata_file ${ar53_metadata}"     : ""
+    def gtdbtk_prefix = gtdbtk_output.find { it.name =~ /summary\.tsv/}.getName() - ~/\.(bac120|ar53)\.summary\.tsv/
+    """
+    gtdb_to_ncbi_majority_vote.py \\
+        --gtdbtk_output_dir . \\
+        ${bac120} \\
+        ${ar53} \\
+        --gtdbtk_prefix ${gtdbtk_prefix} \\
+        --output_file gtdbtk.${prefix}.ncbi.tsv
```
jfy133, line 36: **todo:** Missing `args` (even if not required)

modules/nf-core/gtdbtk/gtdbtoncbimajorityvote/meta.yml
```diff
@@ -0,0 +1,94 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/meta-schema.json
+name: "gtdbtk_gtdbtoncbimajorityvote"
+description: Converts the output classifications of GTDB-TK from GTDB taxonomy to
+  NCBI taxonomy
+keywords:
+  - GTDB taxonomy
+  - NCBI taxonomy
+  - taxonomic classification
```
jfy133, line 8: **suggestion:** 
```suggestion
  - taxonomic classification
  - conversion
  - taxonomy
```

modules/nf-core/gtdbtk/gtdbtoncbimajorityvote/tests/main.nf.test
```diff
@@ -0,0 +1,88 @@
...
+        }
+    }
+
+    test("E. coli - genome fasta") {
+
+        when {
+            process {
+                """
+                input[0] = GTDBTK_CLASSIFYWF.out.gtdb_files
+                input[1] = [[id: "ar53"], file("https://data.ace.uq.edu.au/public/gtdb/data/releases/release226/226.0/ar53_metadata_r226.tsv.gz")]
```
jfy133, line 52: **suggestion:** 
```suggestion
                input[1] = [[id: "ar53"], file("https://data.ace.uq.edu.au/public/gtdb/data/releases/release226/226.0/ar53_metadata_r226.tsv.gz, checkIfExists: true")]
```

Same as the below.

How big are these files?

modules/nf-core/gtdbtk/gtdbtoncbimajorityvote/tests/nextflow.config
```diff
@@ -0,0 +1,5 @@
+process {
+    withName: GTDBTK_CLASSIFYWF {
+        ext.args = "--extension fa --skip_ani_screen"
```
jfy133, line 3: **suggestion:** I recommend using teh `modules_args` system https://nf-co.re/docs/guidelines/components/modules#configuration-of-extargs-in-tests

jfy133 (commented):
First pass!

In regards to the moving, I also wasn't sure what the purpose of that was either tbh - was it for renaming purposes (to insert `${prefix}` into the names)?

For the prefix question - the input/output meta should be the same for CLASSIF_WF no? So I think it would be fine to have an explicit input as it would just be `meta.id`, right?

Although probably inferring it is fine as the file name _should_ be stable

modules/nf-core/gtdbtk/gtdbtoncbimajorityvote/tests/nextflow.config
prototaxites: Does that work when the args you are trying to specify are for a process in the setup block? The params block is specified later in the when block

modules/nf-core/gtdbtk/gtdbtoncbimajorityvote/tests/main.nf.test
prototaxites: One is about 200mb, and one is 50mb

prototaxites:
> In regards to the moving, I also wasn't sure what the purpose of that was either tbh - was it for renaming purposes (to insert ${prefix} into the names)?

I thought it was maybe to simplify the output structure? But the prefix is already specified as an argument to the gtdb classifywf call, so the files don’t need to be renamed.

> For the prefix question - the input/output meta should be the same for CLASSIF_WF no? So I think it would be fine to have an explicit input as it would just be meta.id, right?

The files are prefixed with whatever the prefix was set to in the gtdb classifywf module - so it’s not guaranteed to be meta.id. So either we need to pass it as an input, let the user specify it with ext.prefix, or just infer it from the files.

modules/nf-core/gtdbtk/gtdbtoncbimajorityvote/tests/main.nf.test
jfy133: Is the download speed OK? Or not so good (I'm assuming not)... maybe another case to put on AWS somehwere...

modules/nf-core/gtdbtk/gtdbtoncbimajorityvote/tests/nextflow.config
jfy133: Ah good point... maybe @mahesh-panchal knows? He was the one who came up with that system.

jfy133:
> > In regards to the moving, I also wasn't sure what the purpose of that was either tbh - was it for renaming purposes (to insert ${prefix} into the names)?
> 
> I thought it was maybe to simplify the output structure? But the prefix is already specified as an argument to the gtdb classifywf call, so the files don’t need to be renamed.

If it is just that (I never looked too deeply tbh), then I think fine to leave as is!

> > For the prefix question - the input/output meta should be the same for CLASSIF_WF no? So I think it would be fine to have an explicit input as it would just be meta.id, right?
> 
> The files are prefixed with whatever the prefix was set to in the gtdb classifywf module - so it’s not guaranteed to be meta.id. So either we need to pass it as an input, let the user specify it with ext.prefix, or just infer it from the files.

Hrm I see. I think I would rather pass it as an input channel for explicitness then. If it's a mandatory argument for the tool it's fine to be an input channel.

jfy133 (commented):
Proper new module review now!

modules/nf-core/gtdbtk/gtdbtoncbimajorityvote/tests/nextflow.config
mahesh-panchal: I think so. This is what you mean right?
https://github.com/nf-core/modules/blob/master/modules/nf-core/cnvnator/cnvnator/tests/main.nf.test

jfy133: No, we mean do you know if you can use this system in `setup` block processes too:

nf-test
```
config './nextflow.config'
 
when {
  params {
    module_args = '--extra_opt1 --extra_opt2'
  }
  process {
    """
    input[0] = [
      [ id:'test1', single_end:false ], // meta map
      file(params.modules_testdata_base_path + 'genomics/prokaryotes/bacteroides_fragilis/genome/genome.fna.gz', checkIfExists: true)
    ]
    """
  }
}
```
 config
```
process {
  withName: 'MODULE' {
    ext.args = params.module_args
  }
}
```

modules/nf-core/gtdbtk/gtdbtoncbimajorityvote/tests/main.nf.test
prototaxites: Download speed is fine for these IME - I think it's the mock db specifically that seems throttled

modules/nf-core/gtdbtk/gtdbtoncbimajorityvote/tests/nextflow.config
prototaxites: I will switch back to this in this case 👍

modules/nf-core/gtdbtk/gtdbtoncbimajorityvote/main.nf
```diff
@@ -0,0 +1,54 @@
...
+    if(!bac120_metadata && !ar53_metadata) {
+        log.error("ERROR: Neither the ar53 or bac120 metadata files were provided to gtdbtk/gtdbtoncbimajorityvote!")
+    }
+    def prefix        = task.ext.prefix ?: "${meta.id}"
+    bac120            = bac120_metadata ? "--bac120_metadata_file ${bac120_metadata}" : ""
+    ar53              = ar53_metadata   ? "--ar53_metadata_file ${ar53_metadata}"     : ""
+    def gtdbtk_prefix = gtdbtk_output.find { it.name =~ /summary\.tsv/}.getName() - ~/\.(bac120|ar53)\.summary\.tsv/
+    """
+    gtdb_to_ncbi_majority_vote.py \\
+        --gtdbtk_output_dir . \\
```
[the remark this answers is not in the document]
prototaxites: It's a little bit complicated - the gtdbtk/classifywf module has an output dir that is hard-fixed to `.` - so to pass the files, I've added an output `*` channel that captures all the files so we can stage them in to the work directory - and then we need to specify this as `.`.

Possible fixes:
- Change GTDB-Tk classifywf module to output to a named folder, pass that as input instead of a list of all output files
- Leave the classifywf module as is, stageas these files to some directory

It depends whether we think anyone is going to use this module in a way where it's not interacting with the output of the current classifywf module.

modules/nf-core/gtdbtk/gtdbtoncbimajorityvote/main.nf
```diff
@@ -0,0 +1,54 @@
...
+    task.ext.when == null || task.ext.when
+
+    script:
+    if(!bac120_metadata && !ar53_metadata) {
+        log.error("ERROR: Neither the ar53 or bac120 metadata files were provided to gtdbtk/gtdbtoncbimajorityvote!")
+    }
+    def prefix        = task.ext.prefix ?: "${meta.id}"
+    bac120            = bac120_metadata ? "--bac120_metadata_file ${bac120_metadata}" : ""
+    ar53              = ar53_metadata   ? "--ar53_metadata_file ${ar53_metadata}"     : ""
+    def gtdbtk_prefix = gtdbtk_output.find { it.name =~ /summary\.tsv/}.getName() - ~/\.(bac120|ar53)\.summary\.tsv/
```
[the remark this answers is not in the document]
prototaxites: I think it's only optional in the sense that there's a default - if the GTDB/classifywf call has a prefix set (the nf-core module does) then this does need to be specified explicitly.

modules/nf-core/gtdbtk/gtdbtoncbimajorityvote/meta.yml
```diff
@@ -0,0 +1,94 @@
...
+          description: |
+            Groovy Map containing sample information
+            e.g. `[ id:'sample1' ]`
+          pattern: "*.ncbi.tsv"
+          ontologies:
+            - edam: http://edamontology.org/format_3475
+
+      - "*.ncbi.tsv":
+          type: map
+          description: TSV map for input genomes, mapping GTDB classification taxonomy
```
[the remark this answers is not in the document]
prototaxites: Actually (and annoyingly) doesn't give taxids - just NCBI lineages. I've changed that to make it more explicit

prototaxites:
@jfy133 What about the failing lint check due to the empty file? IMO am tempted to remove but I saw in your PR where you added the test that you explicitly wanted it there?

modules/nf-core/gtdbtk/gtdbtoncbimajorityvote/main.nf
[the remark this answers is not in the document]
jfy133: Ooof, that's a bit nasty hardcoding with the `.`... I would rather go with the former to update it  so that it points to `${prefix}/` (the original module was pre- specs, which is maybe why).

I think for the latter, we can always add additional output channels to 'pluck' specific ones?

prototaxites:
OK!

classifywf:
- Updated the classifywf module so it now puts all its files into a named directory called ${prefix}
- Removed the "gtdbtk." from the specified prefix argument so that it's easier to know what the prefix should be
-  Uses the new mag testdata hosted mock db
- Removes empty file from test

new module:
- takes the gtdbtk dir as input
- Adds explicit input for prefix, but it's optional

Notes for pipelines using this version (just mag?) - this might require some finagling with the publishDir to make sure all the required files are still output

modules/nf-core/gtdbtk/classifywf/meta.yml
```diff
@@ -47,15 +49,31 @@ input:
         description: The local copy of the Mash sketch database used by GTDB-tk if `ani_screen`
           mode is used (optional)
         pattern: "*.msh"
+
```
jfy133, line 52: **suggestion:** 
```suggestion
```

modules/nf-core/gtdbtk/classifywf/meta.yml
```diff
@@ -47,15 +49,31 @@ input:
...
+      - meta:
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. [ id:'test', single_end:false,  assembler:'spades' ]
+          pattern: "*"
+      - "${prefix}":
+          type: map
+          description: All files output by GTDB-Tk
+
```
jfy133, line 66: **suggestion:** 
```suggestion
```

modules/nf-core/gtdbtk/classifywf/meta.yml
```diff
@@ -136,15 +161,17 @@ output:
...
           pattern: "*.{warnings.log}"
   - versions:
       - versions.yml:
           type: file
           description: File containing software versions
           pattern: "versions.yml"
+          ontologies:
+            - edam: http://edamontology.org/format_3750 # YAML
 authors:
   - "@skrakau"
```
jfy133, line 176: **suggestion:** you should probabvly add yourself here now :sweat_smile:

modules/nf-core/gtdbtk/classifywf/tests/main.nf.test
```diff
@@ -61,16 +61,13 @@ nextflow_process {
                         process.out.msa,
                         process.out.user_msa,
                         process.out.filtered,
-                        process.out.failed,
```
jfy133, line 64: **question:** Reason why these are removed? They are still mandatory output channels

modules/nf-core/gtdbtk/gtdbtoncbimajorityvote/main.nf
```diff
@@ -0,0 +1,56 @@
...
+    output:
+    tuple val(meta), path("*.ncbi.tsv"), emit: tsv
+    path "versions.yml"                , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    if(!bac120_metadata && !ar53_metadata) {
+        log.error("ERROR: Neither the ar53 or bac120 metadata files were provided to gtdbtk/gtdbtoncbimajorityvote!")
```
jfy133, line 24: **nitpick:** I would put the module name in the all caps/underscore scheme, as this error would come up in the pipelien running context

modules/nf-core/gtdbtk/gtdbtoncbimajorityvote/tests/nextflow.config
```diff
@@ -0,0 +1,5 @@
+process {
+    withName: GTDBTK_CLASSIFYWF {
```
jfy133, line 2: **question:** Does this work then with the `setup` block? I assume yes as test passed but wanted t ocheck

jfy133 (approved):
Remaining points are nitpicks, but otherwise looks good!

modules/nf-core/gtdbtk/classifywf/tests/main.nf.test
prototaxites: They output empty files, which cause a linting failure (honestly not sure why it's not OK to expect/test empty file output anywhere outside the stub tests, but that's a separate discussion)

modules/nf-core/gtdbtk/gtdbtoncbimajorityvote/main.nf
prototaxites: good call

modules/nf-core/gtdbtk/gtdbtoncbimajorityvote/tests/nextflow.config
prototaxites: It does!

modules/nf-core/gtdbtk/classifywf/meta.yml
prototaxites, line 176:
```suggestion
  - "@skrakau"
  - "@prototaxites"
```

modules/nf-core/gtdbtk/gtdbtoncbimajorityvote/main.nf
prototaxites, line 24:
```suggestion
        log.error("ERROR: Neither the ar53 or bac120 metadata files were provided to GTDBTK_GTDBTONCBIMAJORITYVOTE!")
```

modules/nf-core/gtdbtk/classifywf/meta.yml
prototaxites: I'm never sure when it's appropriate to put my name down in these kinds of things... or if I necessarily want the responsibility 🫢

PR #8391, opened 2025-05-04
Run Nextflow lint in pre-commit

edmundmiller:
- **style: Add Nextflow linting support to pre-commit configuration**
- **ci: -o concise works for some reason**
- **style: Run nextflow lint on everything**


modules/nf-core/amulety/antiberty/tests/nextflow.config
```diff
@@ -1,3 +1,3 @@
 process {
-    cpus = 1 // Needed for deterministic results
```
robsyme, line 2: This seems like a shame to loose the comments herel

ewels: I think this is a current limitation in the formatter, it can't handle inline comments like this yet.

.pre-commit-config.yaml
```diff
@@ -8,6 +8,15 @@ repos:
         files: environment
         types: [yaml]
         additional_dependencies: ["ruamel.yaml"]
+      - id: nextflow-lint
+        name: Lint Nextflow scripts and fix formatting
+        entry: nextflow lint
+        language: system
+        files: '\.nf$|nextflow\.config$'
+        args: ["-format", "-sort-declarations", "-spaces", "4", "-harshil-alignment", "-output", "concise"]
```
ewels, line 16: I'd much prefer a config file 😬 https://github.com/nextflow-io/nextflow/issues/5934

Though I guess we can always add that later.

bentsherman:
I agree with Maxime. I would turn off the formatting and only fix linter errors in this PR. In a separate PR I would run the formatter as a once-off and undo any comment removals or controversial formatting changes. We should improve the formatter to handle those cases, not the other way around

.pre-commit-config.yaml
edmundmiller: I would as well, but perfect is the enemy of good 😉

modules/nf-core/amulety/antiberty/tests/nextflow.config
edmundmiller: That's the point of pre-commit though 😞

edmundmiller:
> I agree with Maxime. I would turn off the formatting and only fix linter errors in this PR. In a separate PR I would run the formatter as a once-off and undo any comment removals or controversial formatting changes. We should improve the formatter to handle those cases, not the other way around

Yeah it's a bit much. I'll chat with @nf-core/maintainers about it.

I'm thinking we
1. Turn on the pre-commit formatting(for the CI check and ease)
2. Report any issues we have to https://github.com/nextflow-io/language-server(Or Nextflow proper?)

ewels:
Works for me 👍🏻  I think a slow roll-out at first until we're confident is a good idea. We can check each module one by one as they are updated to make sure that the changes are sanitised.

This will fail if the locally installed Nextflow version isn't recent enough, right? Any way that we can deal with that gracefully?

edmundmiller:
> This will fail if the locally installed Nextflow version isn't recent enough, right? Any way that we can deal with that gracefully?

How complicated do you want to get?

1. We can use `script` instead and add a version check that way. 

2. We can use the `fail` directive
```yml
-   repo: local
    hooks:
    -   id: changelogs-rst
        name: changelogs must be rst
        entry: changelog filenames must end in .rst
        language: fail
        files: 'changelog/.*(?<!\.rst)$'
```

3. We can use `docker_image` or `conda` to force the right version

ewels:
> How complicated do you want to get?

As little as possible 😅

maxulysse:
@adamrtalbot played with this too in https://github.com/adamrtalbot/nf-lint-pre-commit

maxulysse:
Ran it manually for some modules: https://github.com/nf-core/modules/pull/8891
It is still slightly messy

LouisLeNezet:
Hi,
I'm digging up this old PR, as we would like to add nextflow lint to the Github action and therefore it would be nice to also have it to the pre-commit !

SPPearce:
@bentsherman, @ewels , do you have any progress on improving the formatter?

maxulysse:

mashehu:
we need to add setup-nextflow to the linting.yml and fix-litning.yml for this step to work

ewels:
> @bentsherman, @ewels , do you have any progress on improving the formatter?

No updates yet - we know it's a high priority issue and will get to it ASAP, I think it's currently scheduled to be done once the new Records stuff is in place. Hopefully done before the 26.04 release, but might be another month or so until we get to it.

Issues being tracked [here](https://github.com/nextflow-io/language-server/issues?q=sort%3Aupdated-desc+state%3Aopen+label%3Aformatting), if anything is missing please add more issues.

SPPearce:
I think that we could move any comments that are going to be affected and apply the auto formatting anyway.

SPPearce:
Can we also set vscode to do the same on save?

mashehu:
> Can we also set vscode to do the same on save?

this is just the linter without the formatter atm

jonasscheid:
> Can we also set vscode to do the same on save?

I would suggest only doing that if vscode is not in auto-save mode

mashehu:
My plan for this PR is waiting for @LouisLeNezet to be done with his heroic linting fix efforts and then merge just the `pre-commit-config.yml` with the nextflow lint hook.

.pre-commit-config.yaml
```diff
@@ -60,3 +60,11 @@ repos:
     hooks:
       - id: hadolint-docker
         args: ["--failure-threshold", "error"]
+
+  - repo: https://github.com/seqeralabs/nf-lint-pre-commit
+    rev: v0.2.0
+    hooks:
+      - id: nextflow-lint
+        files: '\.nf$|nextflow\.config$'
```
adamrtalbot, line 68: This is the almost the same as the default for the pre-commit hook: https://github.com/seqeralabs/nf-lint-pre-commit/blob/58b2431ebcdf99a0c024c9c7f695bfa30c01555c/.pre-commit-hooks.yaml#L6

Shall we add `*.nextflow` files upstream? I've never seen one in the wild!

mashehu:
Closing in favor of #10241

PR #8412, opened 2025-05-07
Swap paraclu to nf-test

SPPearce:


SPPearce (commented):
This one is slightly problematic.
The input dataset is not great, and leads to an empty final output file.
paraclu-cut filters out isolated clusters, even if you give it `-d1` (hence adding the args/args2), leading to nothing left.

modules/nf-core/paraclu/tests/main.nf.test
```diff
@@ -0,0 +1,61 @@
...
+                    ]
+                input[1] = 1
+                """
+            }
+        }
+        then {
+            assert process.success
+            assertAll(
+                { assert snapshot(
+                    file(process.out.bed[0][1]).name,
```
LouisLeNezet, line 28: Could you add a comment to explain why you check only the name ?
`// Check only the name as it generate an empty file`

LouisLeNezet (approved):

modules/nf-core/paraclu/tests/main.nf.test
SPPearce, line 28:
```suggestion
                    file(process.out.bed[0][1]).name, // Check name only as file is empty
```

PR #8421, opened 2025-05-08
Update Sentieon build

SPPearce:
All the sentieon modules are currently non-functional as the container needs a patch.


SPPearce:
Superceded by #8427

PR #8426, opened 2025-05-09
Update semibin/singleeasybin

SPPearce:
Swapping `semibin/singleeasybin` to nf-test, including significant updates to the tool.
The tool has been updated quite a lot, including a new SemiBin2 command instead of the previous SemiBin1 version. This includes no longer producing `output_prerecluster_bins` and `output_recluster_bins`, only `output_bins`. 
I don't know if the .h5 model file is still produced for some options, but I kept it in just in case.
I've also changed `args3` to `args2`, because there was no `args2`, and exported the output in a folder called ${prefix}.

Oh, and swapped to a contigs fasta file (instead of a fastq file?) and associated bam file, instead of requiring bowtie to run to align first.


modules/nf-core/semibin/singleeasybin/main.nf
```diff
@@ -4,43 +4,55 @@ process SEMIBIN_SINGLEEASYBIN {
...
+    touch ${prefix}/{contig_bins,recluster_bins_info}.tsv
+    touch ${prefix}/{data,data_split}.csv
+    mkdir ${prefix}/output_bins
+    touch ${prefix}/output_bins/SemiBin_{0,1,2,3}.fa
+    gzip  ${prefix}/output_bins/SemiBin*
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
-        SemiBin: \$( SemiBin --version )
+        SemiBin: \$( SemiBin2 --version )
```
LouisLeNezet, line 55: I don't know if in nf-core we should allow different version of a tool to coexist as glimpse1 and 2.
But this is outside the scope of this PR

LouisLeNezet, line 55:
```suggestion
        SemiBin2: \$( SemiBin2 --version )
```

modules/nf-core/semibin/singleeasybin/main.nf
```diff
@@ -4,43 +4,55 @@ process SEMIBIN_SINGLEEASYBIN {
...
+        --input-bam ${bam} \\
+        --output ${prefix} \\
         -t $task.cpus \\
-        $args3
+        $args2
 
-    mv $prefix/* .
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        SemiBin: \$( SemiBin2 --version )
```
LouisLeNezet, line 40:
```suggestion
        SemiBin2: \$( SemiBin2 --version )
```

LouisLeNezet: This will break the nf-test...

LouisLeNezet (approved):

modules/nf-core/semibin/singleeasybin/main.nf
SPPearce: The tool is called "SemiBin" on its github page, its just that they changed the executable name to be SemiBin2 when they made breaking changes. I think it wants to still be called SemiBin in the report.

SPPearce: The tool is called "SemiBin" on its github page, its just that they changed the executable name to be SemiBin2 when they made breaking changes. I think it wants to still be called SemiBin in the report.

LouisLeNezet: That's fine for me !

PR #8427, opened 2025-05-09
Fix sentieon modules

SPPearce:
Update all sentieon modules to the most recent version (essential to make them functional again, as there is a required patch to be able to talk to the license server).


FriederikeHanssen (approved):

SPPearce:
Thanks @FriederikeHanssen

PR #8446, opened 2025-05-13
Swap pneumocat to nf-test

SPPearce:
Note this only had a stub test before, and still does. The tool gives an error if the input doesn't contain S. pneumoniae, and we don't have any suitable test data at the moment.


LouisLeNezet (approved):
Looks good to me !

PR #8450, opened 2025-05-13
Swap phyloflash to nf-test

SPPearce:
Swap phyloflash to nf-test. 
This module only currently has a stub test, as it requires a largish database.


modules/nf-core/phyloflash/tests/main.nf.test
```diff
@@ -0,0 +1,38 @@
...
+                    ]
+                input[1] = []
+                input[2] = []
+                """
+            }
+        }
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
```
LouisLeNezet, line 33:
```suggestion
                { assert snapshot(
                    process.out,
                    path(process.out.versions[0]).yaml
                ).match() }
```

LouisLeNezet (approved):
Just a difference in the version generated by conda and singularity to fix !

PR #8484, opened 2025-05-15
Swap ncbitools/vecscreen to nf-test

SPPearce:


modules/nf-core/ncbitools/vecscreen/tests/main.nf.test.snap
```diff
@@ -0,0 +1,78 @@
+{
+    "homo sapiens": {
+        "content": [
+            {
+                "0": [
+                    [
+                        {
+                            "id": "test"
+                        },
+                        "test.vecscreen.out:md5,d41d8cd98f00b204e9800998ecf8427e"
```
LouisLeNezet, line 10: Do you know why it is empty ?

modules/nf-core/ncbitools/vecscreen/tests/nextflow.config
```diff
@@ -0,0 +1,9 @@
+process {
+    withName: "BLAST_MAKEBLASTDB" {
+        ext.args = '-dbtype nucl -blastdb_version 4'
```
LouisLeNezet, line 3: It could be set through a `params.module_blast_makeblastdb_args`

LouisLeNezet (changes requested):
`tuple val(meta), path("${meta.id}.vecscreen.out")    , emit: vecscreen_output` should be set to
`tuple val(meta), path("${prefix}.vecscreen.out")    , emit: vecscreen_output`

and all `prefix` declaration without `def`

Why `val(adapters_database_meta)` and not `val(meta2)` ?

Could you set the `VERSION` variable with the regular message ?

modules/nf-core/ncbitools/vecscreen/tests/main.nf.test
```diff
@@ -0,0 +1,81 @@
...
+                input[0] = [[id:'test'],
+                    file(params.modules_testdata_base_path + 'genomics/homo_sapiens/pacbio/fasta/primers.fasta', checkIfExists: true)
+                ]
+                """
+            }
+        }
+    }
+
+    test("homo sapiens") {
+        options '-stub'
```
SPPearce, line 30: @LouisLeNezet , this is why the file is empty.

LouisLeNezet: I didn't catch that. But why do you test twice the stub ?

modules/nf-core/ncbitools/vecscreen/tests/nextflow.config
LouisLeNezet: Make sense

modules/nf-core/ncbitools/vecscreen/tests/main.nf.test
SPPearce: I didn't mean to!

LouisLeNezet (approved):
Good to go for me !

PR #8492, opened 2025-05-16
Add nextdenovo to nf-core modules.

elmedjadjirayane:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/nextdenovo/environment.yml
+++ b/modules/nf-core/nextdenovo/environment.yml
@@ -1,0 +1,5 @@
+channels:
+  - conda-forge
+  - bioconda
+dependencies:
+  - bioconda::nextdenovo=2.5.2
--- a/modules/nf-core/nextdenovo/environment.yml
+++ b/modules/nf-core/nextdenovo/environment.yml
@@ -3,3 +3,4 @@ channels:
   - bioconda
 dependencies:
   - bioconda::nextdenovo=2.5.2
+  - conda-forge::paralleltask=0.1.1
--- a/modules/nf-core/nextdenovo/main.nf
+++ b/modules/nf-core/nextdenovo/main.nf
@@ -1,0 +1,52 @@
+process NEXTDENOVO {
+    tag "$meta.id"
+    label 'process_high'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/fa/fa1c1e961de38d24cf36c424a8f4a9920ddd07b63fdb4cfa51c9e3a593c3c979/data' :
+        'community.wave.seqera.io/library/flye:2.9.5--d577924c8416ccd8' }"
+
+    input:
+    tuple val(meta), path(reads)
+    path config
+
+    output:
+    tuple val(meta), path("*.fasta.gz"), emit: fasta
+    tuple val(meta), path("*.stat")     , emit: stat
+    path "versions.yml"                , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    echo ${reads} > input.fofn
+    nextDenovo \\
+        $config \\
+        input.fofn \\
+
+    gzip -c ./03.ctg_graph/nd.asm.fasta > ${prefix}.assembly.fasta.gz
+  
+    mv ./03.ctg_graph/nd.asm.fasta.stat ${prefix}.assembly_info.stat
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        \$( nextDenovo --version )
+    END_VERSIONS
+    """
+
+    stub:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    echo stub | gzip -c > ${prefix}.assembly.fasta.gz
+    echo contig_1 > ${prefix}.assembly_info.stat
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+    \$( nextDenovo --version )
+    END_VERSIONS
+    """
+}
--- a/modules/nf-core/nextdenovo/main.nf
+++ b/modules/nf-core/nextdenovo/main.nf
@@ -24,6 +24,7 @@ process NEXTDENOVO {
     def prefix = task.ext.prefix ?: "${meta.id}"
     """
     echo ${reads} > input.fofn
+    cat input.fofn
     nextDenovo \\
         $config \\
         input.fofn \\
--- a/modules/nf-core/nextdenovo/meta.yml
+++ b/modules/nf-core/nextdenovo/meta.yml
@@ -1,0 +1,68 @@
+name: "nextdenovo"
+description: NextDenovo is a string graph-based de novo assembler for long reads (CLR,
+  HiFi and ONT). It uses a “correct-then-assemble” strategy similar to canu (no correction
+  step for PacBio HiFi reads), but requires significantly less computing resources
+  and storages
+keywords:
+  - assembly
+  - genome
+  - de novo
+  - genome assembler
+  - single molecule
+tools:
+  - "nextdenovo":
+      description: "NextDenovo is a string graph-based de novo assembler for long reads
+        (CLR, HiFi and ONT). It uses a “correct-then-assemble” strategy similar to canu
+        (no correction step for PacBio HiFi reads), but requires significantly less
+        computing resources and storages"
+      homepage: "https://github.com/fenderglass/Flye"
+      documentation: "https://nextdenovo.readthedocs.io/en/latest/"
+      tool_dev_url: "https://github.com/Nextomics/NextDenovo"
+      doi: "10.1186/s13059-024-03252-4"
+      licence: ["GPL-3.0-license"]
+      identifier: biotools:nextdenovo
+input:
+  - - meta:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. [ id:'test' ]
+    - reads:
+        type: file
+        description: Input reads from Oxford Nanopore or PacBio data in FASTA/FASTQ
+          format.
+        pattern: "*.{fasta,fastq,fasta.gz,fastq.gz,fa,fq,fa.gz,fq.gz}"
+  - - config:
+        type: file
+        description: Input config file for nextDenovo
+        pattern: "*"
+output:
+  - fasta:
+      - meta:
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. [ id:'test' ]
+      - "*.fasta.gz":
+          type: file
+          description: Assembled FASTA file
+          pattern: "*.fasta.gz"
+  - stat:
+      - meta:
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. [ id:'test' ]
+      - "*.stat":
+          type: file
+          description: Extra information and statistics about resulting contigs
+          pattern: "*.stat"
+  - versions:
+      - versions.yml:
+          type: file
+          description: File containing software versions
+          pattern: "versions.yml"
+authors:
+  - "@elmedjadjirayane"
+maintainers:
+  - "@elmedjadjirayane"
--- a/modules/nf-core/nextdenovo/tests/main.nf.test
+++ b/modules/nf-core/nextdenovo/tests/main.nf.test
@@ -11,7 +11,7 @@ nextflow_process {
     tag "nextdenovo"
 
     // TODO nf-core: Change the test name preferably indicating the test-data and file-format used
-    test("sarscov2 - bam") {
+    test("nextdenovo_ont") {
 
         // TODO nf-core: If you are created a test for a chained module
         // (the module requires running more than one process to generate the required output)
@@ -24,9 +24,10 @@ nextflow_process {
                 // TODO nf-core: define inputs of the process here. Example:
                 
                 input[0] = [
-                    [ id:'test', single_end:false ], // meta map
-                    file(params.modules_testdata_base_path + 'genomics/sarscov2/illumina/bam/test.paired_end.sorted.bam', checkIfExists: true),
+                    [ id:'test' ], // meta map
+                    file(params.modules_testdata_base_path + '/genomics/sarscov2/nanopore/fastq/test.fastq.gz', checkIfExists: true),
                 ]
+                input[1] = file(params.modules_testdata_base_path + 'generic/txt/hello.txt')
                 """
             }
         }
@@ -42,32 +43,6 @@ nextflow_process {
 
     }
 
-    // TODO nf-core: Change the test name preferably indicating the test-data and file-format used but keep the " - stub" suffix.
-    test("sarscov2 - bam - stub") {
-
-        options "-stub"
-
-        when {
-            process {
-                """
-                // TODO nf-core: define inputs of the process here. Example:
-                
-                input[0] = [
-                    [ id:'test', single_end:false ], // meta map
-                    file(params.modules_testdata_base_path + 'genomics/sarscov2/illumina/bam/test.paired_end.sorted.bam', checkIfExists: true),
-                ]
-                """
-            }
-        }
-
-        then {
-            assertAll(
-                { assert process.success },
-                { assert snapshot(process.out).match() }
-                //TODO nf-core: Add all required assertions to verify the test output.
-            )
-        }
-
-    }
+    
 
 }
--- a/modules/nf-core/nextdenovo/tests/main.nf.test
+++ b/modules/nf-core/nextdenovo/tests/main.nf.test
@@ -1,0 +1,73 @@
+// TODO nf-core: Once you have added the required tests, please run the following command to build this file:
+// nf-core modules test nextdenovo
+nextflow_process {
+
+    name "Test Process NEXTDENOVO"
+    script "../main.nf"
+    process "NEXTDENOVO"
+
+    tag "modules"
+    tag "modules_nfcore"
+    tag "nextdenovo"
+
+    // TODO nf-core: Change the test name preferably indicating the test-data and file-format used
+    test("sarscov2 - bam") {
+
+        // TODO nf-core: If you are created a test for a chained module
+        // (the module requires running more than one process to generate the required output)
+        // add the 'setup' method here.
+        // You can find more information about how to use a 'setup' method in the docs (https://nf-co.re/docs/contributing/modules#steps-for-creating-nf-test-for-chained-modules).
+
+        when {
+            process {
+                """
+                // TODO nf-core: define inputs of the process here. Example:
+                
+                input[0] = [
+                    [ id:'test', single_end:false ], // meta map
+                    file(params.modules_testdata_base_path + 'genomics/sarscov2/illumina/bam/test.paired_end.sorted.bam', checkIfExists: true),
+                ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
+                //TODO nf-core: Add all required assertions to verify the test output.
+                // See https://nf-co.re/docs/contributing/tutorials/nf-test_assertions for more information and examples.
+            )
+        }
+
+    }
+
+    // TODO nf-core: Change the test name preferably indicating the test-data and file-format used but keep the " - stub" suffix.
+    test("sarscov2 - bam - stub") {
+
+        options "-stub"
+
+        when {
+            process {
+                """
+                // TODO nf-core: define inputs of the process here. Example:
+                
+                input[0] = [
+                    [ id:'test', single_end:false ], // meta map
+                    file(params.modules_testdata_base_path + 'genomics/sarscov2/illumina/bam/test.paired_end.sorted.bam', checkIfExists: true),
+                ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
+                //TODO nf-core: Add all required assertions to verify the test output.
+            )
+        }
+
+    }
+
+}
--- a/modules/nf-core/nextdenovo/tests/main.nf.test.snap
+++ b/modules/nf-core/nextdenovo/tests/main.nf.test.snap
@@ -1,73 +1,16 @@
-// TODO nf-core: Once you have added the required tests, please run the following command to build this file:
-// nf-core modules test nextdenovo
-nextflow_process {
-
-    name "Test Process NEXTDENOVO"
-    script "../main.nf"
-    process "NEXTDENOVO"
-
-    tag "modules"
-    tag "modules_nfcore"
-    tag "nextdenovo"
-
-    // TODO nf-core: Change the test name preferably indicating the test-data and file-format used
-    test("sarscov2 - bam") {
-
-        // TODO nf-core: If you are created a test for a chained module
-        // (the module requires running more than one process to generate the required output)
-        // add the 'setup' method here.
-        // You can find more information about how to use a 'setup' method in the docs (https://nf-co.re/docs/contributing/modules#steps-for-creating-nf-test-for-chained-modules).
-
-        when {
-            process {
-                """
-                // TODO nf-core: define inputs of the process here. Example:
-                
-                input[0] = [
-                    [ id:'test', single_end:false ], // meta map
-                    file(params.modules_testdata_base_path + 'genomics/sarscov2/illumina/bam/test.paired_end.sorted.bam', checkIfExists: true),
-                ]
-                """
-            }
-        }
-
-        then {
-            assertAll(
-                { assert process.success },
-                { assert snapshot(process.out).match() }
-                //TODO nf-core: Add all required assertions to verify the test output.
-                // See https://nf-co.re/docs/contributing/tutorials/nf-test_assertions for more information and examples.
-            )
-        }
-
+{
+    "nextdenovo_ont": {
+        "content": [
+            "798439cbd7fd81cbcc5078022dc5479d",
+            [
+                "versions.yml:md5,ed7a2cfe2387b964c06b6c6847f46996"
+            ]
+        ],
+        "meta": {
+            "nf-test": "0.8.4",
+            "nextflow": "24.04.4"
+        },
+        "timestamp": "2024-08-30T12:42:15.405989"
     }
-
-    // TODO nf-core: Change the test name preferably indicating the test-data and file-format used but keep the " - stub" suffix.
-    test("sarscov2 - bam - stub") {
-
-        options "-stub"
-
-        when {
-            process {
-                """
-                // TODO nf-core: define inputs of the process here. Example:
-                
-                input[0] = [
-                    [ id:'test', single_end:false ], // meta map
-                    file(params.modules_testdata_base_path + 'genomics/sarscov2/illumina/bam/test.paired_end.sorted.bam', checkIfExists: true),
-                ]
-                """
-            }
-        }
-
-        then {
-            assertAll(
-                { assert process.success },
-                { assert snapshot(process.out).match() }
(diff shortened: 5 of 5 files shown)


modules/nf-core/nextdenovo/main.nf
```diff
@@ -0,0 +1,53 @@
...
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    echo ${reads} > input.fofn
+    cat input.fofn
```
nschan, line 27: Not sure what the point of this `cat` is?

modules/nf-core/nextdenovo/environment.yml
```diff
@@ -0,0 +1,6 @@
+channels:
+  - conda-forge
+  - bioconda
+dependencies:
+  - bioconda::nextdenovo=2.5.2
+  - conda-forge::paralleltask=0.1.1
```
nschan, line 6: I think ymls need to end with an empty line

modules/nf-core/nextdenovo/tests/config.cfg
```diff
@@ -0,0 +1,25 @@
...
+[correct_option]
+read_cutoff = 500
+genome_size = 100k # estimated genome size
+
+minimap2_options_raw = -t 8
+pa_correction = 3 # number of corrected tasks used to run in parallel, each corrected task requires ~TOTAL_INPUT_BASES/4 bytes of memory usage.
+correction_options = -p 15
+
+[assemble_option]
+minimap2_options_cns = -t 8
```
nschan, line 22: Ideally the number of threads should be dynamic on `${task.cpus}`

elmedjadjirayane: @nschan whatever is inside this config file is defined by the person running the pipeline, all what is inside must be specific to a run. This one is only a test file, the user must provide this config file to run assembly with nextdenovo. Do you prefer that this file be created dynamically with params?

nschan: I think at least the threading option has to be generated dynamically for the commonly used resource scaling for failed jobs to work.

modules/nf-core/nextdenovo/main.nf
```diff
@@ -0,0 +1,54 @@
...
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/nextdenovo:2.5.2--py310h0ceaa1d_6' :
+        'biocontainers/nextdenovo:2.5.2--py310h0ceaa1d_6' }"
+
+    input:
+    tuple val(meta), path(reads)
+    path config
+
+    output:
```
nschan, line 14: Are these all outputs that are created by nextdenovo? 
All outputs should be emitted in case someone would like to use one of the other outputs in their pipeline.

modules/nf-core/nextdenovo/tests/config.cfg
```diff
@@ -0,0 +1,25 @@
+[General]
+job_type = local # local, slurm, sge, pbs, lsf
+job_prefix = nextDenovo
+task = all # all, correct, assemble
+rewrite = yes # yes/no
+deltmp = yes 
+parallel_jobs = 20 # number of tasks used to run in parallel
```
nschan, line 7: This should probably also be set based on `${task.cpus}`?

modules/nf-core/nextdenovo/tests/config.cfg
```diff
@@ -0,0 +1,25 @@
...
+input_type = raw # raw, corrected
+read_type = ont # clr, ont, hifi
```
nschan, line 9: How can this be switched easily? If integrated somewhere in a pipeline, do you expect users to craft a config specific for nextdenovo, or could this be created in a separate process  e.g `NEXTDENOVO_CREATE_CONFIG` based on pipeline params and then passed to this?

modules/nf-core/nextdenovo/tests/config.cfg
```diff
@@ -0,0 +1,25 @@
...
+input_fofn = input.fofn
+workdir = .
+
+[correct_option]
+read_cutoff = 500
+genome_size = 100k # estimated genome size
```
nschan, line 15: This is another important parameter that may need to be set dynamically?

nschan (changes requested):

modules/nf-core/nextdenovo/tests/config.cfg
elmedjadjirayane: I am sure there is a commit where I set this dynamically.

modules/nf-core/nextdenovo/main.nf
elmedjadjirayane: nextdenovo outputs, a fasta file containing the assembly and a stat file, of course there are other files but they mostly related to the process itself and not results of the assembly.

modules/nf-core/nextdenovo/tests/config.cfg
elmedjadjirayane: Yes, I see. So the module still expects to be given a config file as input but when using that module in a pipeline, it will be created by a simple helper module that would take params and just fills the config file that we would give to nextdenovo. It means that this part won't be handled by nextdenovo's module?

nschan: I guess that is a decision that should be discussed. I would think if such a module is created anyway, it should be easily findable for others that would like to use your nextdenovo module, so having a `nextdenovo/create_config` alongside `nextdenovo/nextdenovo` would make sense to me.

modules/nf-core/nextdenovo/tests/main.nf.test
```diff
@@ -0,0 +1,45 @@
...
+        when {
+            process {
+                """
+               
+                
+                input[0] = [
+                    [ id:'test' ], // meta map
+                    file(params.modules_testdata_base_path + 'genomics/homo_sapiens/pacbio/fastq/test_hifi.fastq.gz', checkIfExists: true),
+                ]
+                input[1] = file("${moduleDir}/tests/config.cfg")
```
nschan, line 25: This should point to the config in test-data i think?

modules/nf-core/nextdenovo/main.nf
nschan: Could any of those files still be useful for someone?

maxulysse:
I'm concerned that this test is failing on docker only

nschan:
Maybe it is failing in docker because docker is somehow limiting memory for the parallel tasks?

nschan:
Hello @elmedjadjirayane, this seems to be somewhat stale, are there things blocking you that you need help resolving?

elmedjadjirayane:
> Hello @elmedjadjirayane, this seems to be somewhat stale, are there things blocking you that you need help resolving?

Hello, I was away these last days. For now there is the issue with docker that I don’t understand because when I test locally, the test passes but here it fails, not sure how to debug this to be honest.

modules/nf-core/nextdenovo/tests/config.cfg
nschan: This seems to be the config for the test. Maybe set it to 1 to see if that makes the docker run pass?

nschan:
I would suggest to modify the test config to run less tasks in parallel and see if that helps with the OOM error in docker tests.

elmedjadjirayane:
@nschan thanks! I think it worked?

nschan:
Nice, that seems to have solved the test problem.

nschan:
It has been a while since that started, and I am not sure what we decided on in the end; I think the we had agreed that configs for the module are to be provided by the pipeline / person using the module is that correct?

elmedjadjirayane:
> It has been a while since that started, and I am not sure what we decided on in the end; I think the we had agreed that configs for the module are to be provided by the pipeline / person using the module is that correct?

Yes, I think we had agreed to that!

nschan:
Ok, maybe it would be good to provide some example config, to me it seems like generating a config (e.g. putting in the correct number of cpu) correctly is some effort and if you have solved that already maybe it would be good to have that  included somewhere?

elmedjadjirayane:
Where do we need to provide that? Would it be the test config we are using or maybe you mean something else?

nschan:
I am also not completely sure what would be the best way, there are other tools with configs, for example https://github.com/nf-core/modules/tree/master/modules/nf-core/getorganelle seems to have process to generate configs. My main concern is, but maybe I am wrong, for something like `${task.cpus}` to actually have the correct value the file will need to be echo'ed or something? The other concern is to avoid making every person using that module go hunt for the default config to copy somewhere into the folder. I think you could also use `$moduleDir` to place a config inside the module dir, or some subdirectory of it.

nschan:
It was suggested by @mahesh-panchal to essentially dump a yaml from parameters, which seems like an easy and also transparent solution to me (see here https://github.com/nf-core/modules/blob/master/modules/nf-core/quartonotebook/main.nf). Let me know if that makes sense to you.

elmedjadjirayane:
Do you mean that the entity using the module will not have to provide that file? I don't understand how we would generate it automatically. I think the idea behind this file is to avoid typing a lot of parameters in a command line and regroup them in one file that we can modify. I am sorry maybe I am not seeing it from the right angle.

nschan:
The issue that I see is: How can resources be adjusted dynamically upon retries, if the file is provided? If the file contains `${task.cpus}` i dont think that this will be subtituted with the variable value. What you could do (I am not very good at groovy there is probably a smarter way) is something like this:

```groovy
script:
    nextdonovo_parameters = [
        job_type: "local",
        job_prefix: "nextDenovo",
        task: "all",
        rewrite: "yes",
        deltmp: "yes",
        parallel_jobs: "${task.cpus}",
        input_type: "raw",
        read_type: "ont",
        input_fofn: "input.fofn",
        workdir: "${meta.id}_work",
        read_cutoff: 1000,
        genome_size: "3g",
        sort_options: "-m 50g -t 30",
        minimap2_options_raw: "-t 8",
        pa_correction: 5,
        correction_options: "-p 30",
        minimap2_options_cns: "-t 8",
        nextgraph_options: "-a 1",
    ]
    def yamlBuilder = new groovy.yaml.YamlBuilder()
    yamlBuilder(nextdonovo_parameters)
    def yaml_content = yamlBuilder.toString().tokenize('\n').join("\n    ").replace(":", " =").replace('"','')
    """
    cat <<- END_YAML_PARAMS > nextdenovo.yml
    ${yaml_content}
    END_YAML_PARAMS
    """
```

this creates a file with this content (`meta.id` was "test"):
```
---
job_type = local
job_prefix = nextDenovo
task = all
rewrite = yes
deltmp = yes
parallel_jobs = 1
input_type = raw
read_type = ont
input_fofn = input.fofn
workdir = test_work
read_cutoff = 1000
genome_size = 3g
sort_options = -m 50g -t 30
minimap2_options_raw = -t 8
pa_correction = 5
correction_options = -p 30
minimap2_options_cns = -t 8
nextgraph_options = -a 1
```

which would allow to simply edit the config parameters in the script. 

I think this is still not optimal, ideally all of these would go into `params` for this module, so they can be edited without having to modify the module file? If this is part of the script file, everyone would need to patch main.nf for this script in their pipeline. 
I think this could go into `nextflow.config` in the module dir:

```
params {
   nextDenovo_job_tyoe = local
   nextDenovo_job_prefix = nextDenovo
   ...
}
```
and the script would become:

```groovy
script:
    nextdonovo_parameters = [
        job_type: params.nextDenovo_job_type,
        job_prefix: params.nextDenovo_job_prefix,
   .
    ]
    def yamlBuilder = new groovy.yaml.YamlBuilder()
    yamlBuilder(nextdonovo_parameters)
    def yaml_content = yamlBuilder.toString().tokenize('\n').join("\n    ").replace(":", " =").replace('"','')
    """
    cat <<- END_YAML_PARAMS > nextdenovo.yml
    ${yaml_content}
    END_YAML_PARAMS
    """
```

Does this make sense?

elmedjadjirayane:
Thank you for those clarifications. I see what you mean. So we would have a config in the module dir and if the user wants to change on the parameters they will have to change that in the module dir? instead of giving it as an input each time they run it ?

nschan:
If we want to be able to dynamically adjust resources for failed tasks, the user is unable to provide a new config file between attempts and the setting of the number of tasks has to be done somewhat dynamically based on `task.cpus` in the script.

nschan:
So, what works would be this in a file `nextdenovo.config` in the module dir:
```groovy
params {
        nextDenovo_job_type = "local"
        nextDenovo_job_prefix = "nextDenovo"
        nextDenovo_task = "all"
        nextDenovo_rewrite = "yes"
        nextDenovo_deltmp = "yes"
        nextDenovo_input_type = "raw"
        nextDenovo_read_type = "ont"
        nextDenovo_input_fofn = "input.fofn"
        nextDenovo_read_cutoff = 1000
        nextDenovo_genome_size = "3g"
        nextDenovo_sort_options = "-m 50g -t 30"
        nextDenovo_minimap2_options_raw = "-t 8"
        nextDenovo_pa_correction = 5
        nextDenovo_correction_options = "-p 30"
        nextDenovo_minimap2_options_cns = "-t 8"
        nextDenovo_nextgraph_options = "-a 1"
}
```

this in the pipeline `nextflow.config`:
```
includeConfig 'modules/nf-core/nextdenovo/nextdenovo.config'
```

and this in the script:
```groovy
nextdonovo_parameters = [
        job_type: params.nextDenovo_job_type,
        job_prefix: params.nextDenovo_job_prefix,
        task: params.nextDenovo_task,
        rewrite: params.nextDenovo_rewrite,
        deltmp: params.nextDenovo_deltmp,
        parallel_jobs: "${task.cpus}",
        input_type: params.nextDenovo_input_type,
        read_type: params.nextDenovo_read_type,
        input_fofn: params.nextDenovo_input_fofn,
        workdir: "${meta.id}_nxtd_work",
        read_cutoff: params.nextDenovo_read_cutoff,
        genome_size: params.nextDenovo_genome_size,
        sort_options: params.nextDenovo_sort_options,
        minimap2_options_raw: params.nextDenovo_minimap2_options_raw,
        pa_correction: params.nextDenovo_pa_correction,
        correction_options: params.nextDenovo_correction_options,
        minimap2_options_cns: params.nextDenovo_minimap2_options_cns,
        nextgraph_options: params.nextDenovo_nextgraph_options
    ]
    def yamlBuilder = new groovy.yaml.YamlBuilder()
    yamlBuilder(nextdonovo_parameters)
    def yaml_content = yamlBuilder.toString().tokenize('\n').join("\n    ").replace(":", " =").replace('"','')
    """
    cat <<- END_YAML_PARAMS > nextdenovo.yml
    ${yaml_content}
    END_YAML_PARAMS
    """
```

things like `${task.cpus}` cannot be stuffed into params that way I think because they get assigned later, when params are already written.

elmedjadjirayane:
> If we want to be able to dynamically adjust resources for failed tasks, the user is unable to provide a new config file between attempts and the setting of the number of tasks has to be done somewhat dynamically based on `task.cpus` in the script.

Ah yes you are right, the issue is with the parallel jobs parameters then? which we set here `echo "parallel_jobs = ${task.cpus}" >> conf.cfg`. I think we can come up with a solution that only include this parameter if it is the only causing issues, the file can have a lot of optional parameters and I am not sure whether it's right to include all of them in the script?

nschan:
I think there are more parameters that need to be set during the pipeline run: genome_size seems important, read_type too?
If nextdenovo uses some defaults when there is no config file it might be ok to only echo some into the config, but I think with assemblers there are many parameters that may need to be adjusted somehow. How would you adjust those during a pipeline run (with heterogenous samples) if you can provide 1 config file?

elmedjadjirayane:
I see, parameters are specific to a sample. Almost all parameters have default values but those default values aren't always suitable.
I am not really experienced with nextflow so I think I will take your advice on this haha.

nschan:
I am not sure if my advice is the best strategy, please take a look at some modules in this repo that have complex configurations and think about which is the best way to go for this particular tool. The modules should be written in such a way that they can be used as broadly as possible. In many modules this is done via command line args, and can become quite complex.

mahesh-panchal:
Groovy has a TomlBuilder class btw: https://groovy-lang.org/processing-toml.html

You could have it such that some defaults could be supplied from a file or map ( param ). The content is converted to a map outside the process, which can then be augmented `meta` map outside the process if necessary and within the module for the resource parameters. The map is then written out using the TomlBuilder.

So the process input expects a `val` object which should be a map of settings. In the process, the cpus, etc get added/overwritten and then it's written to a toml.

nschan:
Thanks @mahesh-panchal, using the correct builder is a much better approach than trying to use replace to make the yaml into a toml, however I am unable to use `groovy.toml.TomlBuilder()` in nextflow (`cause: unable to resolve class groovy.toml.TomlBuilder`) have you done that before?

mahesh-panchal:
You're right. It doesn't appear to be available. It seems it's still an incubating feature of Groovy 4.

The tidiest solution might be to write out a yaml and use something like `dasel` to convert it to toml. That way the files are still human readable and debuggable.

nschan:
That would add a dependency for dasel I guess.
Based on your suggestion of using a `meta` map, here is what I came up with:
```groovy
script:
    nextdonovo_parameters = [
        job_type: meta.nextDenovo.job_type ?: "local",
        job_prefix: meta.nextDenovo.job_prefix ?: "nextDenovo",
        task: meta.nextDenovo.task ?: "all",
        rewrite: meta.nextDenovo.rewrite ?: "yes",
        deltmp: meta.nextDenovo.deltmp ?: "yes",
        parallel_jobs: meta.nextDenovo.parallel_jobs ?: "${task.cpus}",
        input_type: meta.nextDenovo.input_type ?: "raw",
        read_type: meta.nextDenovo.read_type ?: error('Please provide the read type for nextDenovo'),
        input_fofn: meta.nextDenovo.input_fofn ?: "input.fofn",
        workdir: meta.nextDenovo.workdir ?: "${meta.id}_nxtd_work",
        read_cutoff: meta.nextDenovo.read_cutoff ?: "1k",
        genome_size: meta.nextDenovo.genome_size ?: "1g",
        sort_options: meta.nextDenovo.sort_options ?: "-m ${task.memory.toGiga()}g -t ${task.cpus}",
        minimap2_options_raw: meta.nextDenovo.minimap2_options_raw ?: "-t ${task.cpus}",
        pa_correction: meta.nextDenovo.pa_correction ?: "3",
        correction_options: meta.nextDenovo.correction_options ?: "-p 15",
        minimap2_options_cns: meta.nextDenovo.minimap2_options_cns ?: "-t ${task.cpus}",
        nextgraph_options: meta.nextDenovo.nextgraph_options ?: "-a 1"
    ]
    def yamlBuilder = new groovy.yaml.YamlBuilder()
    yamlBuilder(nextdonovo_parameters)
    def yaml_content = yamlBuilder.toString().tokenize('\n').join("\n    ").replace(":", " =").replace('"','').replace("---\n    ", "")
    """
    cat <<- END_YAML_PARAMS > ${meta.id}_nextdenovo.yml
    ${yaml_content}
    END_YAML_PARAMS
    """
```
This means that a `nextDenovo` map in `meta` is required (even if empty). As it currently it is, `meta.nextDenovo` also needs to contain `read_type`. I tested this with this channel: 
```
Channel.of([[id: "test_1", nextDenovo:[read_type: "ont"]], "test_1"], [[id: "test_2", nextDenovo: [read_type: "hifi"]], "test_2"])
``` 
and that works, filling in defaults for everything. It errors if `nextDenovo.read_type` is not provided, which I think it should.

What do you think @elmedjadjirayane ?

Note: I dont kow if nextDenovo is strict about the toml sections / headers, if it is that would need to be added I guess.

mahesh-panchal:
I would avoid that giant map of meta parameters. Ideally, set some minimal defaults like Quartonotebook does. Then add the input map to the default map.

```
    input:
    ...
    val cfg_params // Map:

    script:
    def nextdonovo_parameters = [
        job_type: "local",
        job_prefix: "nextDenovo",
        task: "all",
        rewrite: "yes",
        deltmp: "yes",
        parallel_jobs: "${task.cpus}",
        input_type: "raw",
        read_type: "hifi",
        input_fofn: "input.fofn",
        workdir: "${meta.id}_nxtd_work",
        read_cutoff: "1k",
        genome_size: "1g",
        sort_options: "-m ${task.memory.toGiga()}g -t ${task.cpus}",
        minimap2_options_raw: "-t ${task.cpus}",
        pa_correction: "3",
        correction_options: "-p 15",
        minimap2_options_cns: "-t ${task.cpus}",
        nextgraph_options: "-a 1"
    ] + cfg_params
```

Anything that's duplicated in `cfg_params` will overwrite what's in the defaults. 

Also it's against guidelines to define any `meta` fields in modules other than `id`, and `single_end` to allow developers flexibility.

mahesh-panchal:
Also if something is optional, then leave it out from the defaults, but make an assertion for fields that are required in some combination. It should then come in through the `input` map.

nschan:
Thats likely a better approach, I guess I misunderstood regarding the `meta` map for those params. It makes sense to supply them as a separate channel. What about something that is _required_ (read type maybe?)

mahesh-panchal:
You can either add it as a mandatory input, and/or test it's defined in the script:

```
input:
...
val read_type

script:
...
    read_type: read_type 
...
if( !nextdonovo_parameters.read_type ){
    error ...
}
```

That'll make it explicit, but it could be part of the cfg_params and described in the meta.yml.

modules/nf-core/nextdenovo/main.nf
```diff
@@ -0,0 +1,54 @@
...
+    label 'process_high'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/nextdenovo:2.5.2--py310h0ceaa1d_6' :
+        'biocontainers/nextdenovo:2.5.2--py310h0ceaa1d_6' }"
+
+    input:
+    tuple val(meta), path(reads)
+    path config
```
nschan, line 12:
```suggestion
    val nextdenovo_parameters
```

modules/nf-core/nextdenovo/main.nf
```diff
@@ -0,0 +1,54 @@
...
+    tuple val(meta), path("*.fasta.gz"), emit: fasta
+    tuple val(meta), path("*.stat")     , emit: stat
+    path "versions.yml"                , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
```
nschan, line 24:
```suggestion
    def prefix = task.ext.prefix ?: "${meta.id}"
    nextdonovo_config = [
        job_type: "local",
        job_prefix: "nextDenovo",
        task: "all",
        rewrite: "yes",
        deltmp: "yes",
        parallel_jobs: "${task.cpus}",
        input_type: "raw",
        read_type: null,
        input_fofn: "input.fofn",
        workdir: "${meta.id}_nxtd_work",
        read_cutoff: "1k",
        genome_size: "1g",
        sort_options: "-m ${task.memory.toGiga()}g -t ${task.cpus}",
        minimap2_options_raw: "-t ${task.cpus}",
        pa_correction: "3",
        correction_options: "-p 15",
        minimap2_options_cns: "-t ${task.cpus}",
        nextgraph_options: "-a 1"
    ] + nextdenovo_parameters
    if ( !nextdonovo_config.read_type ) error('Please provide the read type for nextDenovo')
    def yamlBuilder = new groovy.yaml.YamlBuilder()
    yamlBuilder(nextdonovo_config)
    def yaml_content = yamlBuilder.toString().tokenize('\n').join("\n    ").replace(":", " =").replace('"','').replace("---\n    ", "")
```

modules/nf-core/nextdenovo/main.nf
```diff
@@ -0,0 +1,54 @@
...
+    """
+    echo "parallel_jobs = ${task.cpus}" >> conf.cfg
+    cat $config >> conf.cfg
```
nschan, line 27:
```suggestion
    cat <<- END_YAML_PARAMS > ${meta.id}_nextdenovo.cfg
    ${yaml_content}
    END_YAML_PARAMS
```

modules/nf-core/nextdenovo/main.nf
```diff
@@ -0,0 +1,54 @@
...
+    echo ${reads} > input.fofn
+    nextDenovo \\
+        conf.cfg \\
```
nschan, line 30:
```suggestion
        ${meta.id}_nextdenovo.cfg \\
```

modules/nf-core/nextdenovo/main.nf
```diff
@@ -0,0 +1,54 @@
...
+        input.fofn \\
+
```
nschan, line 32:
```suggestion
        $args
```

modules/nf-core/nextdenovo/main.nf
```diff
@@ -0,0 +1,54 @@
...
+    gzip -c ./03.ctg_graph/nd.asm.fasta > ${prefix}.assembly.fasta.gz
```
nschan, line 33: Is this path affected by `workdir` in the config? If so this probably should be something like:
```suggestion
    gzip -c ${nextdenovo_config.work_dir}/03.ctg_graph/nd.asm.fasta > ${prefix}.assembly.fasta.gz
```

modules/nf-core/nextdenovo/main.nf
```diff
@@ -0,0 +1,54 @@
...
+  
+    mv ./03.ctg_graph/nd.asm.fasta.stat ${prefix}.assembly_info.stat
```
nschan, line 35: Similar to the above:
```suggestion
    mv ${nextdenovo_config.work_dir}/03.ctg_graph/nd.asm.fasta.stat ${prefix}.assembly_info.stat
```

modules/nf-core/nextdenovo/meta.yml
```diff
@@ -0,0 +1,68 @@
...
+          e.g. [ id:'test' ]
+    - reads:
+        type: file
+        description: Input reads from Oxford Nanopore or PacBio data in FASTA/FASTQ
+          format.
+        pattern: "*.{fasta,fastq,fasta.gz,fastq.gz,fa,fq,fa.gz,fq.gz}"
+  - - config:
+        type: file
+        description: Input config file for nextDenovo
+        pattern: "*"
```
nschan, line 38: This should be type map and list the required, and optional configuration parameters.

modules/nf-core/nextdenovo/tests/main.nf.test
elmedjadjirayane: Yes, to the test-data, I must edit that test file too in the repo to put 1 for parallel jobs.

modules/nf-core/nextdenovo/meta.yml
elmedjadjirayane: Do I list those in the description?

modules/nf-core/nextdenovo/main.nf
elmedjadjirayane: Yes, if in the config the work dir is set to something other than the pwd (.) then it won't work, thanks for noticing this.

modules/nf-core/nextdenovo/meta.yml
nschan: At minimum provide a link to the docs, but I think listing them would be better instead of making people click-through.

mahesh-panchal:
Off topic, just FYI.
How to use the TomlBuilder : https://nextflow.slack.com/archives/C02T98A23U7/p1750855387940709?thread_ts=1750842869.120739&cid=C02T98A23U7
The TomlBuilder class also only writes dotted key syntax and not nested table syntax.
Ben might include `groovy-toml` into Nextflow at some point.

modules/nf-core/nextdenovo/tests/main.nf.test
nschan: Or you make use of the proposed `nextdenovo_parameters` input map in your test to set those parameters.

PR #8531, opened 2025-05-21
Added compatibility with micromamba

d-jch:
micromamba and conda check channels in different ways

### DIFF
--- a/subworkflows/nf-core/utils_nextflow_pipeline/main.nf
+++ b/subworkflows/nf-core/utils_nextflow_pipeline/main.nf
@@ -87,17 +87,32 @@ def dumpParametersToJSON(outdir) {
 def checkCondaChannels() {
     def parser = new org.yaml.snakeyaml.Yaml()
     def channels = []
-    try {
-        def config = parser.load("conda config --show channels".execute().text)
-        channels = config.channels
-    }
-    catch (NullPointerException e) {
-        log.debug(e)
-        log.warn("Could not verify conda channel configuration.")
+
+    // Helper function: execute shell command and parse channel output
+    def getChannels = { String cmd ->
+        try {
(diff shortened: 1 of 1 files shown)


SPPearce:
@nf-core/maintainers , anyone have thoughts on this?

jfy133:
My instinct is this should be a nextflow level thing...?

mahesh-panchal:
Quick thought but I would probably change it such that, it did

```
def config = parser.load("conda config --show channels || micromamba config list channels".execute().text)
```

which tries to run `conda` first and if that fails, it runs `micromamba`. I'm not sure if it works though ( ~~for example if the `conda` binary is not there, what happens?~~ It reports binary is missing and executes 2nd part of the `||`, but only the command output is included in the stdout )

Ideally one would need to check the config, but I'm not sure how to check if `micromamba` was enabled from a script context off the top of my head.

nathanweeks:
> ```
> def config = parser.load("conda config --show channels || micromamba config list channels".execute().text)
> ```

Just tested this (`nextflow run nf-core/demo -profile test,conda --outdir outdir`) with micromamba & observed the following error:
```
ERROR ~ Invalid method invocation `debug` with arguments: java.io.IOException: Cannot run program "conda": error=2, No such file or directory (java.io.IOException) on Logger type
```
The following modification worked:
```
def config = parser.load(["sh", "-c", "conda config --show channels || micromamba config list channels"].execute().text)
```

Possible edge case: if both conda and micromamba commands are available in the same environment, `conda.useMicromamba = "true"`, and conda is configured with a different set of channels than micromamba, then an unexpected set of channels might be used? But if so, it would still be preferable to have micromamba working with nf-core and live with such an edge case.

PR #8209, opened 2025-04-04
New module metacache/build

Gullumluvl:
## PR checklist

Closes #7967 

- [ ] This comment contains a description of changes (with reason).
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [x] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [x] Remove all TODO statements.
- [x] Emit the `versions.yml` file.
- [x] Follow the naming conventions.
- [ ] Follow the parameters requirements.
- [x] Follow the input/output options guidelines.
- [x] Add a resource `label`
- [x] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - For modules:
    - [ ] `nf-core modules test <MODULE> --profile docker`
    - [x] `nf-core modules test <MODULE> --profile singularity`
    - [x] `nf-core modules test <MODULE> --profile conda`
  - For subworkflows:
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile docker`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile singularity`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile conda`

PR #8537, opened 2025-05-22
Add metacache query module

sofstam:
## PR checklist

This PR closes: #7852 

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/metacache/query/main.nf
+++ b/modules/nf-core/metacache/query/main.nf
@@ -1,0 +1,54 @@
+process METACACHE_QUERY {
+    tag "$meta.id"
+    label 'process_medium'
+
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/metacache:2.5.0--h077b44d_0':
+        'biocontainers/metacache:2.5.0--h077b44d_0' }"
+
+    input:
+    tuple val(meta), path(reads)
+    path db
+
+    output:
+    tuple val(meta), path("*.txt"), emit: mapping_results
+    path "versions.yml"           , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def input_file = meta.single_end ? reads : "${reads[0]} ${reads[1]} -pairfiles"
+    """
+    dbmeta=`find -L ${db}/ -name "*.meta" | head -n 1`
+    metacache \\
+        query \\
+        \$dbmeta \\
+        ${input_file} \\
+        $args \\
+        -out ${prefix}.txt
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        metacache: \$(metacache info |& sed -n 's/^MetaCache version \\+\\([0-9.]\\+\\).*\$/\\1/p')
+    END_VERSIONS
+    """
+
+    stub:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def input_file = meta.single_end ? reads : "${reads[0]} ${reads[1]} -pairfiles"
+
+    """
+    touch ${prefix}.txt
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+         metacache: \$(metacache info |& sed -n 's/^MetaCache version \\+\\([0-9.]\\+\\).*\$/\\1/p')
+    END_VERSIONS
+    """
+}
--- a/modules/nf-core/metacache/query/meta.yml
+++ b/modules/nf-core/metacache/query/meta.yml
@@ -1,0 +1,53 @@
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/meta-schema.json
+name: "metacache_query"
+description: Metacache query command for taxonomic classification
+keywords:
+  - metagenomics
+  - classification
+  - metacache
+tools:
+  - "metacache":
+      description: MetaCache is a classification system for mapping genomic sequences (short reads, long reads, contigs, ...) from metagenomic samples to their most likely taxon of origin.
+      homepage: https://github.com/muellan/metacache
+      documentation: https://muellan.github.io/metacache/
+      tool_dev_url: https://github.com/muellan/metacache
+      doi: 10.1093/bioinformatics/btx520
+      licence: ["GPL-3.0 license"]
+
+input:
+  - - meta:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. `[ id:'sample1', single_end:false ]`
+    - reads:
+        type: file
+        description: FASTA or FASTQ files
+        pattern: "*.{fasta,fa,fastq,fq}"
+  - - db:
+        type: directory
+        description:  A MetaCache database contains taxonomic information and min-hash signatures of reference sequences (complete genomes, scaffolds, contigs, ...).
+
+output:
+  - mapping_results:
+    - meta:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. `[ id:'sample1', single_end:false ]`
+    - "*.txt":
+        type: file
+        description: Output file containing mapping results
+        pattern: "*.txt"
+
+  - versions:
+    - "versions.yml":
+        type: file
+        description: File containing software versions
+        pattern: "versions.yml"
+
+authors:
+  - "@sofstam"
+maintainers:
+  - "@sofstam"
--- a/modules/nf-core/metacache/query/tests/main.nf.test
+++ b/modules/nf-core/metacache/query/tests/main.nf.test
@@ -1,0 +1,85 @@
+nextflow_process {
+
+    name "Test Process METACACHE_QUERY"
+    script "../main.nf"
+    process "METACACHE_QUERY"
+    tag "modules"
+    tag "modules_nfcore"
+    tag "metacache"
+    tag "metacache/query"
+
+    test("sarscov2 nanopore [fastq_gz]") {
+
+        setup {
+        run ("UNTAR") {
+            script "../../../untar/main.nf"
+            process {
+                """
+                input[0] =  [ [], file('https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/delete_me/test-db-metacache.tar.gz', checkIfExists: true) ]
+                """
+            }
+        }
+    }
+
+        when {
+            process {
+                """
+                input[0] = [
+                    [ id:'test', single_end:true ], // meta map
+                    file(params.modules_testdata_base_path + 'genomics/sarscov2/nanopore/fastq/test_2.fastq.gz', checkIfExists: true),
+                ]
+                input[1] = UNTAR.out.untar.map{it[1]}
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
+            )
+        }
+
+    }
+
+    test("sarscov2 nanopore [fastq_gz] stub") {
+
+       options "-stub"
+
+        setup {
+        run ("UNTAR") {
+            script "../../../untar/main.nf"
+            process {
+                """
+                input[0] =  [ [], file('https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/delete_me/test-db-metacache.tar.gz', checkIfExists: true) ]
+                """
+            }
+        }
+    }
+
+        when {
+            process {
+                """
+
+                input[0] = [
+                    [ id:'test', single_end:true ], // meta map
+                    file(params.modules_testdata_base_path + 'genomics/sarscov2/nanopore/fastq/test_2.fastq.gz', checkIfExists: true),
+                ]
+                input[1] = UNTAR.out.untar.map{it[1]}
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
+            )
+        }
+
+    }
+
+
+
+
+}
--- a/modules/nf-core/metacache/query/tests/main.nf.test.snap
+++ b/modules/nf-core/metacache/query/tests/main.nf.test.snap
@@ -1,0 +1,72 @@
+{
+    "sarscov2 nanopore [fastq_gz] stub": {
+        "content": [
+            {
+                "0": [
+                    [
+                        {
+                            "id": "test",
+                            "single_end": true
+                        },
+                        "test.txt:md5,d41d8cd98f00b204e9800998ecf8427e"
+                    ]
+                ],
+                "1": [
+                    "versions.yml:md5,803de66366edd7630306fc5b6a2b9cf3"
+                ],
+                "mapping_results": [
+                    [
+                        {
+                            "id": "test",
+                            "single_end": true
+                        },
+                        "test.txt:md5,d41d8cd98f00b204e9800998ecf8427e"
+                    ]
+                ],
+                "versions": [
+                    "versions.yml:md5,803de66366edd7630306fc5b6a2b9cf3"
+                ]
+            }
+        ],
+        "meta": {
+            "nf-test": "0.9.2",
+            "nextflow": "25.04.2"
+        },
+        "timestamp": "2025-05-22T09:08:14.789268824"
+    },
+    "sarscov2 nanopore [fastq_gz]": {
+        "content": [
+            {
+                "0": [
+                    [
+                        {
+                            "id": "test",
+                            "single_end": true
+                        },
+                        "test.txt:md5,2d4ed7a156235a484926649cff4e64ba"
+                    ]
+                ],
+                "1": [
+                    "versions.yml:md5,669526a447b62c30b8bce10f24d149b0"
+                ],
+                "mapping_results": [
+                    [
+                        {
+                            "id": "test",
+                            "single_end": true
+                        },
+                        "test.txt:md5,2d4ed7a156235a484926649cff4e64ba"
+                    ]
+                ],
+                "versions": [
+                    "versions.yml:md5,669526a447b62c30b8bce10f24d149b0"
+                ]
+            }
+        ],
+        "meta": {
+            "nf-test": "0.9.2",
+            "nextflow": "25.04.2"
+        },
+        "timestamp": "2025-05-22T09:07:59.44548077"
+    }
+}
--- a/modules/nf-core/metacache/query/environment.yml
+++ b/modules/nf-core/metacache/query/environment.yml
@@ -1,0 +1,7 @@
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/environment-schema.json
+channels:
+  - conda-forge
+  - bioconda
+dependencies:
+  - "bioconda::metacache=2.5.0"


modules/nf-core/metacache/query/main.nf
```diff
@@ -0,0 +1,54 @@
...
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def input_file = meta.single_end ? reads : "${reads[0]} ${reads[1]} -pairfiles"
+    """
+    dbmeta=`find -L ${db}/ -name "*.meta" | head -n 1`
```
Gullumluvl, line 27: How about groovy code here, instead:

```groovy
def dbmeta = db.find{ f -> f.name =~ $/\.meta$$/$ }
```

Gullumluvl: Ah sorry, did not understand that input `db` is a directory. Doesn't it make easier to provide the list of files? Because the output of `METACACHE_BUILD` is (currently) not inside a directory. (Then we would have to replace `UNTAR` with `UNTARFILES` in the tests).

modules/nf-core/metacache/query/main.nf
```diff
@@ -0,0 +1,54 @@
...
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/metacache:2.5.0--h077b44d_0':
+        'biocontainers/metacache:2.5.0--h077b44d_0' }"
+
+    input:
+    tuple val(meta), path(reads)
+    path db
+
+    output:
+    tuple val(meta), path("*.txt"), emit: mapping_results
```
Gullumluvl, line 16: To be exhaustive (maybe for later), we could add the optional output generated with the option `-abundances`.

```groovy
    path "abundances.table", emit: abundances, optional: true
```

Could be triggered by adding a boolean input `val abundances`.

Gullumluvl (commented):
My main question is about the `db` input being a directory. Should I change the `METACACHE_BUILD` module (https://github.com/nf-core/modules/pull/8209) to emit a directory instead?

sofstam:
> My main question is about the `db` input being a directory. Should I change the `METACACHE_BUILD` module (#8209) to emit a directory instead?

When I use the `metacache build` command, I get a directory. For example:

`metacache build testdb-metacache minimal_databases/`. 

Do you use the command differently?

sofstam:
> > ```
> > metacache build testdb-metacache minimal_databases/.
> > ```
> 
> I tried this command, and it created the files `testdb-metacache.meta` and `testdb-metacache.cache0` in the working directory.

Then it sounds that it should omit a directory?

modules/nf-core/metacache/query/main.nf
```diff
@@ -0,0 +1,54 @@
...
+
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/metacache:2.5.0--h077b44d_0':
+        'biocontainers/metacache:2.5.0--h077b44d_0' }"
+
+    input:
+    tuple val(meta), path(reads)
+    path db
```
Gullumluvl, line 13: I think I would change this to accept files, for example like:

```groovy
    path db, stageAs: "db/*"
```

Gullumluvl (commented):
I did not quite get your question, does this answer it? (see the comment in this review)

modules/nf-core/metacache/query/main.nf
sofstam: Thanks for the suggestion, I will update the PR.

modules/nf-core/metacache/query/main.nf
sofstam: @Gullumluvl Is the file produced when using `-abundances` as flag? It is a tsv file in this case.

Gullumluvl: It's a weird pipe + space -delimited format (https://github.com/muellan/metacache/blob/master/docs/output.md#abundance-summary):

```
# query summary: number of queries mapped per taxon
# rank:name                     | taxid   | reads | abundance
domain:Bacteria                 | 2       | 30    |  0.3%
phylum:Proteobacteria           | 1224    | 16    |  0.16%
phylum:Firmicutes               | 1239    | 1     |  0.01%
```

I would use the `.txt` extension?

modules/nf-core/metacache/query/main.nf
```diff
@@ -0,0 +1,55 @@
...
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/metacache:2.5.0--h077b44d_0':
+        'biocontainers/metacache:2.5.0--h077b44d_0' }"
+
+    input:
+    tuple val(meta), path(reads)
+    path (db, stageAs: 'db/*')
+
```
Gullumluvl, line 15:
```suggestion
    val(do_abundances)  // empty list or Boolean: flag to produce abundances.txt

```

I think the option to output the `abundances.txt` file should be given as input, rather than as `task.ext.args`, because it's a choice of the pipeline developer rather than a config option for the user.

modules/nf-core/metacache/query/main.nf
```diff
@@ -0,0 +1,55 @@
...
+    output:
+    tuple val(meta), path("*.txt")          , emit: mapping_results
```
Gullumluvl:
```suggestion
    tuple val(meta), path("*mapping.txt")   , emit: mapping_results
```
Avoids including "*abundances.txt"

modules/nf-core/metacache/query/main.nf
```diff
@@ -0,0 +1,55 @@
...
+    tuple val(meta), path("*abundances.txt"), emit: abundances, optional: true
+    path "versions.yml"                     , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def input_file = meta.single_end ? reads : "${reads[0]} ${reads[1]} -pairfiles"
```
Gullumluvl, line 27:
```suggestion
    def input_file = meta.single_end ? reads : "${reads[0]} ${reads[1]} -pairfiles"
    def abundance_opt = do_abundances ? "-abundances ${prefix}.abundances.txt" : ''
```

modules/nf-core/metacache/query/main.nf
```diff
@@ -0,0 +1,55 @@
...
+    """
+    dbmeta=`find -L ${db}/ -name "*.meta" | head -n 1`
+    metacache \\
+        query \\
+        \$dbmeta \\
+        ${input_file} \\
+        $args \\
+        -out ${prefix}.txt
```
Gullumluvl:
```suggestion
        $abundance_opt \\
        $args \\
        -out ${prefix}.mapping.txt
```

modules/nf-core/metacache/query/main.nf
```diff
@@ -0,0 +1,55 @@
...
+    "${task.process}":
+        metacache: \$(metacache info |& sed -n 's/^MetaCache version \\+\\([0-9.]\\+\\).*\$/\\1/p')
+    END_VERSIONS
+    """
+
+    stub:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def input_file = meta.single_end ? reads : "${reads[0]} ${reads[1]} -pairfiles"
+
```
Gullumluvl:
```suggestion
    def abundance_opt = do_abundances ? "-abundances ${prefix}.abundances.txt" : ''
    def dbmeta = db.find{ p -> p.name =~ $/\.meta$$/$ }
    assert dbmeta
```

Same as the `script` block.

modules/nf-core/metacache/query/main.nf
```diff
@@ -0,0 +1,55 @@
...
+    """
+    touch ${prefix}.txt
```
Gullumluvl:
```suggestion
    touch ${prefix}.mapping.txt

    [[ -n '$abundance_opt' ]] && touch '${prefix}.abundances.txt'
```

modules/nf-core/metacache/query/meta.yml
```diff
@@ -0,0 +1,65 @@
...
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. `[ id:'sample1', single_end:false ]`
+    - reads:
+        type: file
+        description: FASTA or FASTQ files
+        pattern: "*.{fasta,fa,fastq,fq}"
+  - - db:
+        type: directory
```
Gullumluvl, line 29:
```suggestion
        type: file
        pattern: "**.{meta,cache*}"
```

modules/nf-core/metacache/query/meta.yml
```diff
@@ -0,0 +1,65 @@
...
+        description:  A MetaCache database contains taxonomic information and min-hash signatures of reference sequences (complete genomes, scaffolds, contigs, ...).
+
```
Gullumluvl:
```suggestion
  - - do_abundances:
        type: boolean
        description: flag indicating whether to produce the abundance table.

```

modules/nf-core/metacache/query/meta.yml
```diff
@@ -0,0 +1,65 @@
...
+  - mapping_results:
+    - meta:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. `[ id:'sample1', single_end:false ]`
+    - "*.txt":
+        type: file
+        description: Output file containing mapping results
+        pattern: "*.txt"
```
Gullumluvl:
```suggestion
    - "*.mapping.txt":
        type: file
        description: Output file containing mapping results
        pattern: "*mapping.txt"
```

modules/nf-core/metacache/query/meta.yml
```diff
@@ -0,0 +1,65 @@
...
+  - abundances:
+     - meta:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. `[ id:'sample1', single_end:false ]`
+     - "*abundances.txt":
+        type: file
+        description: Output file showing absolute and relative abundance of each taxon
+        pattern: "*.txt"
```
Gullumluvl:
```suggestion
        pattern: "*abundances.txt"
```

modules/nf-core/metacache/query/tests/main.nf.test
```diff
@@ -0,0 +1,88 @@
...
+        run ("UNTAR") {
+            script "../../../untar/main.nf"
+            process {
+                """
+                input[0] =  [ [], file('https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/delete_me/test-db-metacache.tar.gz', checkIfExists: true) ]
+                """
+            }
+        }
+    }
+
```
Gullumluvl:
```suggestion
```

I couldn't get `UNTAR` to output files with `-stub`, but it apparently works to make a channel of files that don't exist (see next suggestion), so we can remove the `setup` in stub runs.

modules/nf-core/metacache/query/tests/main.nf.test
```diff
@@ -0,0 +1,88 @@
...
+        when {
+            process {
+                """
+
+                input[0] = [
+                    [ id:'test', single_end:true ], // meta map
+                    file(params.modules_testdata_base_path + 'genomics/sarscov2/nanopore/fastq/test_2.fastq.gz', checkIfExists: true),
+                ]
+                input[1] = UNTAR.out.untar.map{it[1]}
```
Gullumluvl:
```suggestion
                input[1] = Channel.fromPath(["metacache/testdb.cache0", "metacache/testdb.meta"], glob: false, checkIfExists: false)
                           | toList()
                input[2] = [false]

```

modules/nf-core/metacache/query/tests/main.nf.test
```diff
@@ -0,0 +1,88 @@
...
+            assert snapshot(
+                process.out.versions,
+                file(process.out.mapping_results[0][1]).readLines()[0]
+            ).match()
+        }
+
+    }
+
+
+
```
Gullumluvl, line 109: Adding tests for the abundances output:

```suggestion
    test("sarscov2 nanopore [fastq_gz] abundances") {

        setup {
        run ("UNTAR") {
            script "../../../untar/main.nf"
            process {
                """
                input[0] =  [ [], file('https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/delete_me/test-db-metacache.tar.gz', checkIfExists: true) ]
                """
            }
        }
    }

        when {
            process {
                """
                input[0] = [
                    [ id:'test', single_end:true ], // meta map
                    file(params.modules_testdata_base_path + 'genomics/sarscov2/nanopore/fastq/test_2.fastq.gz', checkIfExists: true),
                ]
                input[1] = UNTAR.out.untar.map{ tuple(it[1].listFiles()) }
                input[2] = [true]
                """
            }
        }

        then {
            assert process.success
            assert snapshot(
                process.out.versions,
                file(process.out.mapping_results[0][1]).readLines()[0],
                file(process.out.abundances[0][1]).readLines()[0]
            ).match()
        }

    }

    test("sarscov2 nanopore [fastq_gz] abundances stub") {

        options "-stub"

        when {
            process {
                '''
                input[0] = [
                    [ id:'test', single_end:true ], // meta map
                    file(params.modules_testdata_base_path + 'genomics/sarscov2/nanopore/fastq/test_2.fastq.gz', checkIfExists: true),
                ]
                input[1] = Channel.fromPath(["metacache/testdb.cache0", "metacache/testdb.meta"], glob: false, checkIfExists: false)
                           | toList()
                input[2] = [true]
                '''
            }
        }

        then {
            assert process.success
            assert snapshot(
                process.out.versions,
                file(process.out.mapping_results[0][1]).readLines()[0],
                file(process.out.abundances[0][1]).readLines()[0]
            ).match()
        }
    }
```

modules/nf-core/metacache/query/tests/main.nf.test.snap
Gullumluvl, line 1: (to update)

modules/nf-core/metacache/query/main.nf
```diff
@@ -0,0 +1,55 @@
...
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def input_file = meta.single_end ? reads : "${reads[0]} ${reads[1]} -pairfiles"
+    """
+    dbmeta=`find -L ${db}/ -name "*.meta" | head -n 1`
+    metacache \\
+        query \\
+        \$dbmeta \\
```
Gullumluvl:
```suggestion
    def dbmeta = db.find{ p -> p.name =~ $/\.meta$$/$ }
    assert dbmeta
    """
    metacache \\
        query \\
        $dbmeta \\
```

One less subshell + IO call ?

modules/nf-core/metacache/query/main.nf
sofstam: I am resolving this.

modules/nf-core/metacache/query/meta.yml
sofstam: The database is still defined as a directory having those two files, or?

Gullumluvl: Not if you accept this change: https://github.com/nf-core/modules/pull/8537#discussion_r2123829645

I also made suggestions in the tests with list of files as input instead of directory.

modules/nf-core/metacache/query/tests/main.nf.test
sofstam: I do not understand why untar is being removed. I think it would be nice to be able to untar a zipped folder. There is no issue with the stub when I run locally. What error are you encountering?

Gullumluvl: I just removed it in the 'stub' test case, to be clear. In my case, the output of `UNTAR` for a stub run is an empty list (not sure why, from reading the code I assumed it would touch the right filenames), so my process was failing because of the `assert dbmeta` line in the stub block.

So instead of untar, I directly produce the expected file names.

sofstam: You are right about the stub, saw it now that it produces null. Good catch!

sofstam: Maybe, it is not so strange that it returns null since the stub code is creating two files that are empty. I updated the nf-test, so maybe we do not need to remove the untar setup. By using the proposed suggestion, does it result in a non-empty file?

Gullumluvl: I don't think we are talking about the same thing?

I was saying that this *input* is a Channel containing an empty list: `UNTAR.out.untar.map{it[1]}`, so it actually provides zero file to the `METACACHE_QUERY` process. It's not a problem if we don't check them (but my suggestions for stub block did), maybe just not very realistic.

So, if `UNTAR` fails to provide input files to this test and we can create them without it, why keep it?

modules/nf-core/metacache/query/main.nf
```diff
@@ -0,0 +1,59 @@
...
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def input_file = meta.single_end ? reads : "${reads[0]} ${reads[1]} -pairfiles"
+    def abundance_opt = do_abundances ? "-abundances ${prefix}.abundances.txt" : ''
+    """
+    dbmeta=`find -L ${db}/ -name "*.meta" | head -n 1`
```
Gullumluvl:
```suggestion
    dbmeta=`find -L db/ -name "*.meta" | head -n 1`
    [ -n "\$dbmeta" ] || { echo 'Database file "*.meta" not found!' >&2 ; exit 1 ; }
```

1. First line because of `saveAs: "db/*"`
2. Second line in case it's empty

modules/nf-core/metacache/query/main.nf
```diff
@@ -0,0 +1,59 @@
...
+    """
+
+    stub:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def input_file = meta.single_end ? reads : "${reads[0]} ${reads[1]} -pairfiles"
+    def abundance_opt = do_abundances ? "-abundances ${prefix}.abundances.txt" : ''
+    """
+    touch ${prefix}.mapping.txt
+    touch ${prefix}.abundances.txt
```
Gullumluvl:
```suggestion
    [ -n "$abundance_opt" ] && touch ${prefix}.abundances.txt
```
Only create the file if requested.

modules/nf-core/metacache/query/meta.yml
```diff
@@ -0,0 +1,67 @@
...
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. `[ id:'sample1', single_end:false ]`
+    - reads:
+        type: file
+        description: FASTA or FASTQ files
+        pattern: "*.{fasta,fa,fastq,fq}"
+  - - db:
+        type: directory
```
Gullumluvl, line 29: With the current implementation, it can be `directory` OR `file`, isn't it beautiful? :smile:

Would be neat if we could write (but we can't):
```yaml
        type: directory or file
```

modules/nf-core/metacache/query/tests/main.nf.test
```diff
@@ -0,0 +1,77 @@
...
+    setup {
+        run ("UNTAR") {
+            script "../../../untar/main.nf"
+            process {
+                """
+                input[0] =  [ [], file('https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/delete_me/test-db-metacache.tar.gz', checkIfExists: true) ]
+                """
+            }
+        }
+    }
```
Gullumluvl: Just a side note, but nothing to change right now: maybe a setup that uses `metacache/build` would be interesting?

modules/nf-core/metacache/query/tests/main.nf.test
```diff
@@ -0,0 +1,77 @@
...
+                    file(params.modules_testdata_base_path + 'genomics/sarscov2/nanopore/fastq/test_2.fastq.gz', checkIfExists: true),
+                ]
+                input[1] = UNTAR.out.untar.map{it[1]}
+                input[2] = false
+                """
+            }
+        }
+
+        then {
+            assert process.success
```
Gullumluvl, line 47:
```suggestion
            assert process.success
            assert file(process.out.mapping_results[1][0]).size()
```

modules/nf-core/metacache/query/tests/main.nf.test
```diff
@@ -0,0 +1,77 @@
...
+            assert snapshot(
+                process.out.versions,
+                file(process.out.mapping_results[0][1]).readLines()[0]
+            ).match()
```
Gullumluvl, line 51: Just snapshot the full output instead?

```suggestion
            assert snapshot(process.out).match()
```

modules/nf-core/metacache/query/tests/main.nf.test
Gullumluvl: How about these tests for the abundance output?

Gullumluvl (approved):
Minor suggestions

modules/nf-core/metacache/query/tests/main.nf.test
sofstam: That is correct! I will update this test together with the abundance ones.

modules/nf-core/metacache/query/tests/main.nf.test
sofstam: Added the test for abundance.

PR #8586, opened 2025-06-02
ci: Turn arm64 tests off

edmundmiller:
Until we can work out a better "skip" if the build is failing mechanism.

Currently, all of the tests that fail are resulting in a large number of "failed" build requests on the wave side, instead of just passing an already made container.

### DIFF
--- a/.github/workflows/nf-test.yml
+++ b/.github/workflows/nf-test.yml
@@ -103,7 +103,8 @@ jobs:
         profile: [conda, docker, singularity]
         arch:
           - x64
-          - arm64
+          # FIXME
+          # - arm64
     env:
       NXF_ANSI_LOG: false
       TOTAL_SHARDS: ${{ needs.nf-test-changes.outputs.total_shards }}


mashehu:
What is the problem with the current setup?

edmundmiller:
> What is the problem with the current setup?

Currently, all of the tests that fail result in a large number of "failed" build requests on the wave side instead of just passing an already-made container.

So it's just wasting a bunch of compute trying to build containers that we know are failing.

mashehu (approved):

ewels:
To quote an @edmundmiller Slack classic:

> Like we can build out "Hey this package failed to build, if `arm-failure.txt` exists in the module directory, don't run arm wave request, and don't run arm CI" until we remove the txt

I think this is nice solution ☝🏻  Otherwise yeah, we're just running builds on a bunch of containers for every unrelated commit for basically no reason. Adding a file there would prevent that and hopefully annoy people a bit more and actually make folks more likely to try to fix stuff 😅

PR #8589, opened 2025-06-03
8541 update glnexus

### LINKED ISSUE #8541: update module: GLNEXUS

awgymer:
This PR updates the GLNexus process to add two pieces of functionality. 

- Allow outputting VCF format instead of BCF by piping through `bcftools convert`
- Allow usage of a custom config file for non-standard gVCF formats 

## PR checklist

Closes #8541  

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/glnexus/meta.yml
+++ b/modules/nf-core/glnexus/meta.yml
@@ -38,7 +38,7 @@ input:
         pattern: "*.bed"
   - - vcf_output:
         type: boolean
-        description: Specify that output should be VCF format
+        description: Specify to output to VCF format
 output:
   - bcf:
       - meta:


modules/nf-core/glnexus/meta.yml
```diff
@@ -32,15 +36,29 @@ input:
         type: list
         description: Input BED file
         pattern: "*.bed"
+  - - vcf_output:
+        type: boolean
+        description: Specify that output should be VCF format
```
jfy133: [suggestion(non-blocking)](https://pullpo.io/cc?l=suggestion&d=non-blocking) 
```suggestion
        description: Specify to output to VCF format
```

I found the phrasing was a bit ambiguous in how to interact with the channel (i.e., I couldn't tell immediately that it was a true/false action)

modules/nf-core/glnexus/main.nf
```diff
@@ -4,16 +4,18 @@ process GLNEXUS {
...
-    tuple val(meta), path(gvcfs)
+    tuple val(meta), path(gvcfs), path(custom_config)
     tuple val(meta2), path(bed)
+    val vcf_output
 
     output:
-    tuple val(meta), path("*.bcf"), emit: bcf
-    path "versions.yml"           , emit: versions
+    tuple val(meta), path("*.bcf")   , emit: bcf, optional:true
+    tuple val(meta), path("*.vcf.gz"), emit: vcf, optional:true
```
jfy133, line 17: [thought(non-blocking)](https://pullpo.io/cc?l=thought&d=non-blocking) 
I sort of wonder if these actually need to be separate given they are mutually exclusive, and they are actually the same contents just in a compressed format. Separating across the two would require a pipeline developer to do more channel manipulation after for simply because the format type is a bit different.

jfy133 (approved):

modules/nf-core/glnexus/main.nf
awgymer: I believe this is the way other optional-output-format modules handle this, and I think it's more explicit as well as preserving the status-quo of the module.  

It's frustrating to me that (like bam/cram) more downstream tools don't necessarily support both formats, but it's also worth noting that I don't think it's compressed vs uncompressed. 

In this case I _think_ the difference is *uncompressed binary* vs *compressed text*

jfy133: > I believe this is the way other optional-output-format modules handle this, and I think it's more explicit as well as preserving the status-quo of the module.

Mmm, some do but not all.

E.g. the guidelines allows it: https://nf-co.re/docs/guidelines/components/modules#one-output-channel-per-output-file-type

> It's frustrating to me that (like bam/cram) more downstream tools don't necessarily support both formats, but it's also worth noting that I don't think it's compressed vs uncompressed.

> In this case I _think_ the difference is _uncompressed binary_ vs _compressed text_

But regardless, fine to keep it as is for consistency if that is what the other modules in this area of the ecosystem are doing

awgymer: There's actually not *loads* of modules that emit bcf/vcf but I took my lead from `bcftools/convert` which has them separate. `delly` appears to be an outlier in this area as it emits vcf/bcf but in a channel always called `bcf` (this is confusing imo). 

Anyway I will merge as-is based on this discussion

PR #8618, opened 2025-06-09
feat: support no reference tumor only cnvkit analysis

znorgaard:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


modules/nf-core/cnvkit/batch/main.nf
```diff
@@ -78,6 +78,13 @@ process CNVKIT_BATCH {
         tumor_out = ""
     }
 
+    // tumor_only mode and no reference
+    // generate a "flat" reference which assumes equal coverage
+    // by passing '--normal' without any files
+    if (!reference_exists){
+        normal_args = normal_args ?: "--normal"
```
SPPearce, line 78: This will override the previous `normal_args`?

znorgaard: Yes and no.

If `normal_args` is empty it will be overwritten with the new value, otherwise it will be overwritten with the existing value.

Example of each case:
```nextflow
workflow {

    def a_args = "some a arg"
    a_args = a_args ?: "different a arg"
    println(a_args) // will print "some a arg"

    def b_args = ""
    b_args  = b_args ?: "different b arg"
    println(b_args) // will print "different b arg"

}
```

We're trying to solve for the case where reference does not exist and no normal args were supplied. If that's the case `normal_args` will be empty at this point so we'll update it, otherwise it's left as is.

I've seen this pattern used a few places, especially when pulling in `task.ext.args`. Example: https://github.com/nf-core/modules/blob/48c86d92fe761cb3faafdb38e4d8ca3bf6d1c657/modules/nf-core/fqtk/main.nf#L27

SPPearce: Yes, you are quite right, just wasn't thinking hard enough when reading the code on my phone. Apologies

SPPearce (changes requested):
The stub test is failing because there is no stub here, can you please add one.

znorgaard:
> The stub test is failing because there is no stub here, can you please add one.

That's not actually why it's failing. [In the absence of a stub the script block is executed](https://www.nextflow.io/docs/latest/process.html#stub). This is specifically a bug in nextflow=24.10.2, https://github.com/nextflow-io/nextflow/issues/5456. You can prove this to yourself locally.

1. Environment set up with mamba/conda

```console
mamba create -n nf-core-nextflow25 nf-core nextflow=25.04.2
mamba create -n nf-core-nextflow24 nf-core nextflow=24.10.2
```

2. Get to the master modules branch

```console
cd modules
git checkout master
git pull
```

3. Run tests with nextflow 24.20.2 and 25.04.2

```console
mamba activate nf-core-nextflow25
nf-test test modules/nf-core/cnvkit/batch/tests --profile docker
```

`SUCCESS: Executed 9 tests in 118.892s`

```console
mamba deactivate
mamba activate nf-core-nextflow24
nf-test test modules/nf-core/cnvkit/batch/tests --profile docker
```

`FAILURE: Executed 9 tests in 114.519s (1 failed)`

If you look at the failed actions they're specifically the ones for nextflow=24.20.2. The actions using latest everything pass.

That said, I'm happy to add a stub for this.

SPPearce:
> > The stub test is failing because there is no stub here, can you please add one.
> 
> That's not actually why it's failing. [In the absence of a stub the script block is executed](https://www.nextflow.io/docs/latest/process.html#stub). This is specifically a bug in nextflow=24.10.2, [nextflow-io/nextflow#5456](https://github.com/nextflow-io/nextflow/issues/5456). You can prove this to yourself locally.
> 
> 1. Environment set up with mamba/conda
> 
> ```
> mamba create -n nf-core-nextflow25 nf-core nextflow=25.04.2
> mamba create -n nf-core-nextflow24 nf-core nextflow=24.10.2
> ```
> 
> 2. Get to the master modules branch
> 
> ```
> cd modules
> git checkout master
> git pull
> ```
> 
> 3. Run tests with nextflow 24.20.2 and 25.04.2
> 
> ```
> mamba activate nf-core-nextflow25
> nf-test test modules/nf-core/cnvkit/batch/tests --profile docker
> ```
> 
> `SUCCESS: Executed 9 tests in 118.892s`
> 
> ```
> mamba deactivate
> mamba activate nf-core-nextflow24
> nf-test test modules/nf-core/cnvkit/batch/tests --profile docker
> ```
> 
> `FAILURE: Executed 9 tests in 114.519s (1 failed)`
> 
> If you look at the failed actions they're specifically the ones for nextflow=24.20.2. The actions using latest everything pass.
> 
> That said, I'm happy to add a stub for this.

Yes, that might be the case, but it will be fixed if you add a stub.

SPPearce:
@znorgaard , please note that the module version has been updated in the meantime.

modules/nf-core/cnvkit/batch/main.nf
```diff
@@ -107,4 +114,40 @@ process CNVKIT_BATCH {
...
+    stub:
+    def tumor_exists = tumor ? true : false
+    def normal_exists = normal ? true : false
+    def reference_exists = reference ? true : false
+
+    // identify BED naming pattern
+    def bed_prefix = reference_exists ? reference.BaseName : targets ? targets.BaseName : ""
+    def bed_suffix = reference_exists ? "-tmp.bed" : ".bed"
+
+    // execute samtools only when cram files are input, cnvkit runs natively on bam but is prohibitively slow
```
SPPearce:
```suggestion
    // execute samtools only when cram files are input, cnvkit runs natively on cram but is prohibitively slow
```
Should this be cram not bam?

modules/nf-core/cnvkit/batch/tests/main.nf.test
```diff
@@ -160,6 +160,36 @@ nextflow_process {
...
+                input[4] = [[:],file(params.modules_testdata_base_path + 'genomics/homo_sapiens/genome/chr21/sequence/reference_chr21.cnn', checkIfExists: true)]
+                input[5] = false
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out.versions).match() }
```
SPPearce: Are we only checking that the versions file is generated? No check that the output files are there at all?

modules/nf-core/cnvkit/batch/main.nf
znorgaard: Good catch. I just copied from the script block. Fixed in both places.

modules/nf-core/cnvkit/batch/tests/main.nf.test
znorgaard: Ah, I copied the checks done for the other tests. I expanded this test to check additional expected files.

modules/nf-core/cnvkit/batch/tests/main.nf.test
```diff
@@ -160,6 +160,45 @@ nextflow_process {
...
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                {
+                    assert snapshot(
+                        file(process.out.bed.flatten()[1]).name, // empty file
+                        file(process.out.bed.flatten()[2]).name, // empty file
```
znorgaard, line 185: I'm sure there's a better way to do this if anyone has recommendations. I just gave up fighting with groovy.

SPPearce, line 185: This would be one way:
```suggestion
                        file(process.out.bed[0][1]).name, // empty file
                        file(process.out.bed[0][2]).name, // empty file
```
Although it is somewhat odd that the channel emits two bed files?

famosab: it does not :) this first one gets you the meta map

famosab, line 185:
```suggestion
                        file(process.out.bed[0][1]).name, // empty file
```
This should suffice

famosab: ah it does not ...

modules/nf-core/cnvkit/batch/main.nf
famosab, line 1: we might need to add 

```
tuple val(meta), path("*.target.bed"), emit: targetbed
tuple val(meta), path("*.antitarget.bed"), emit: antitargetbed
```

instead of

```
tuple val(meta), path("*.bed"), emit: bed
```

znorgaard: That would be more correct. The emit statements in this module are a bit of a mess and I was doing my best to avoid rewriting them entirely which may break current uses of the module.

There may be many `.cnn` files generated for each run. A `reference.cnn`, `[tumor].antitargetcoverage.cnn`, `[tumor].targetcoverage.cnn`, `[normal].targetcoverage.cnn`, `[normal].antitargetcoverage.cnn`.

Multiple `.cns` files, `[tumor].call.cns`, `[tumor].bintest.cns`, `[tumor].cns`. 

Maybe we need an issue to improve the channel outputs for this module.

modules/nf-core/cnvkit/batch/tests/main.nf.test
znorgaard: This doesn't work for some reason. I get this error:

```
 groovy.lang.MissingMethodException: No signature of method: static com.askimed.nf.test.lang.extensions.GlobalMethods.file() is applicable for argument types: (ArrayList) values: [[/Users/znorgaard/software/modules/.nf-test/tests/4766b1fea7064081ec3ac81f59a61816/work/39/f8c2b9cbb9d107c5a7b82bbd1547fc/genome.antitarget.bed, ...]]
```

znorgaard: Ah, it needs to be 

```nextflow
file(process.out.bed[0][1][0]).name, // empty file
file(process.out.bed[0][1][1]).name, // empty file
```
First item in the channel, second element of the emitted item (the array of beds), first and second bed files.

I think I prefer the flatten.

modules/nf-core/cnvkit/batch/main.nf
SPPearce: Feel free to make the changes in this PR, or you can leave it to a later issue ;)

znorgaard: Made a new issue: https://github.com/nf-core/modules/issues/8820. I'll add that one to my list of future todo items!

modules/nf-core/cnvkit/batch/tests/main.nf.test
famosab: yeah but we have to seperate these two bed files into their own channel :) standard output channel is meta-map + file

PR #8622, opened 2025-06-11
Update cellbender modules

nictru:
This PR updates some dependency versions and switches to test data on the modules branch.

I added new test datasets to the modules branch, and switched to using them here. The `h5` input file for `CELLBENDER_MERGE` is still in the scdownstream branch, as I am not sure if an intermediate file like this should be placed in the modules branch.
One could also generate it in the test with `setup`, but the file is not stable and thus we could then not snapshot in the `CELLBENDER_MERGE` test.


mashehu:
hmm, not 100% what is going on with the conda cache there. could you maybe update the minimum nextflow version to 24.10.8 (`nf-core pipelines bump-version --nextlfow 24.10.8`) just in case it was fixed there?

modules/nf-core/cellbender/removebackground/tests/gpu.nf.test
```diff
@@ -0,0 +1,59 @@
...
+            assertAll(
+                {assert process.success},
+                {assert file(process.out.h5.get(0).get(1)).exists()},
+                {assert file(process.out.filtered_h5.get(0).get(1)).exists()},
+                {assert file(process.out.posterior_h5.get(0).get(1)).exists()},
+                {assert file(process.out.barcodes.get(0).get(1)).exists()},
+                {assert file(process.out.metrics.get(0).get(1)).exists()},
+                {assert file(process.out.report.get(0).get(1)).exists()},
+                {assert file(process.out.pdf.get(0).get(1)).exists()},
+                {assert file(process.out.log.get(0).get(1)).exists()},
```
maxulysse: instead of `exists()`, can you try snapshoting the filename?

nictru: Yes🫡

maxulysse (approved):

PR #8625, opened 2025-06-11
remove implicit params use in sub-wf

sateeshperi:
Replaces & Closes #8616


subworkflows/nf-core/fasta_index_bismark_bwameth/main.nf
```diff
@@ -81,10 +83,10 @@ workflow FASTA_INDEX_BISMARK_BWAMETH {
     /*
     * Generate fasta index if not supplied for bwameth workflow or picard collecthsmetrics tool
     */
-    if (params.aligner == 'bwameth' | params.run_picard_collecthsmetrics) {
+    if (aligner == 'bwameth' | collecthsmetrics) {
```
SPPearce, line 101: Is there a reason why this only works for bwameth?

subworkflows/nf-core/fasta_index_bismark_bwameth/main.nf
```diff
@@ -81,10 +83,10 @@ workflow FASTA_INDEX_BISMARK_BWAMETH {
...
         // already exising fasta index
         if (fasta_index) {
-            ch_fasta_index = Channel.value(file(fasta_index, checkIfExists: true))
+            ch_fasta_index = Channel.value(file(fasta_index, checkIfExists: true)).map{ index -> [ [:], index ]}
```
SPPearce: Why is this making a channel, it should already be an channel from the input?

subworkflows/nf-core/fasta_index_bismark_bwameth/main.nf
sateeshperi: i ll let @dcarrillox answer that

subworkflows/nf-core/fasta_index_bismark_bwameth/main.nf
sateeshperi: but, we want to cast the index as a value channel. u suggesting use to use .first() or something ?

sateeshperi: but we need to cast as value channel for reuse right? its being done above for other indices as well

subworkflows/nf-core/fasta_index_bismark_bwameth/main.nf
sateeshperi: yes. because it needs the output of picard markdup which is run only in the bwameth workflow

subworkflows/nf-core/fasta_index_bismark_bwameth/main.nf
SPPearce: but it should already be one?
Or if it happens to be a queue channel you can use `.first()` or `.collect()` to convert it.

sateeshperi: @SPPearce I tried to move the value channel creation out of this sub-wf to the main but, wasn't getting an easy and elegant solution. If you can let this pass this time, I will refactor the connections in a next release. lmk ur thoughts

sateeshperi: @SPPearce let me thoughts on above plz.

subworkflows/nf-core/fasta_index_bismark_bwameth/main.nf
SPPearce: It is currently an "or", so presumably should be "and"?

SPPearce (approved):

sateeshperi:
yes thank you

sateeshperi:
it is an or as we want the fasta index to be processed with either or

PR #8635, opened 2025-06-13
Add force-tile-extent module

Felix-Kummer:
Earth science module that computes a list of allowed FORCE tiles for a given datacube and geometry.

## PR checklist

- [x] Add a resource `label`
  - For modules:

### DIFF
--- a/.github/skip_nf_test.json
+++ b/.github/skip_nf_test.json
@@ -43,6 +43,7 @@
         "modules/nf-core/fcs/fcsadaptor",
         "modules/nf-core/fcs/fcsgx",
         "modules/nf-core/force/cube",
+        "modules/nf-core/force/tileextent",
         "modules/nf-core/ganon/buildcustom",
         "modules/nf-core/ganon/classify",
         "modules/nf-core/ganon/report",
--- a/modules/nf-core/force/tileextent/Dockerfile
+++ b/modules/nf-core/force/tileextent/Dockerfile
@@ -1,1 +1,1 @@
-FROM quay.io/nfcore/force:3.8.01
+FROM davidfrantz/force:3.8.01
--- a/modules/nf-core/force/tileextent/Dockerfile
+++ b/modules/nf-core/force/tileextent/Dockerfile
@@ -1,0 +1,1 @@
+FROM quay.io/nfcore/force:3.8.01
--- a/modules/nf-core/force/tileextent/main.nf
+++ b/modules/nf-core/force/tileextent/main.nf
@@ -1,5 +1,5 @@
 process FORCE_TILEEXTENT {
-    tag { aoi.simpleName }
+    tag "${aoi.simpleName}"
     label 'process_single'
 
     container "nf-core/force:3.8.01"
--- a/modules/nf-core/force/tileextent/main.nf
+++ b/modules/nf-core/force/tileextent/main.nf
@@ -1,6 +1,7 @@
 process FORCE_TILEEXTENT {
     tag "${aoi.simpleName}"
     label 'process_single'
+    stageInMode 'copy' // needed by the module to work properly when aoi is a shapefile
 
     container "nf-core/force:3.8.01"
 
--- a/modules/nf-core/force/tileextent/main.nf
+++ b/modules/nf-core/force/tileextent/main.nf
@@ -6,7 +6,7 @@ process FORCE_TILEEXTENT {
 
     input:
     path aoi
-    path "tmp/datacube-definition.prj"
+    path(datacube_definition, stageAs: "tmp/datacube-definition.prj")
     path shapefile_dbf
     path shapefile_prj
     path shapefile_shx
--- a/modules/nf-core/force/tileextent/main.nf
+++ b/modules/nf-core/force/tileextent/main.nf
@@ -1,0 +1,43 @@
+process FORCE_TILEEXTENT {
+    tag { aoi.simpleName }
+    label 'process_single'
+
+    container "nf-core/force:3.8.01"
+
+    input:
+    path aoi
+    path "tmp/datacube-definition.prj"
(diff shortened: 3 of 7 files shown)


.github/skip_nf_test.json
```diff
@@ -43,6 +43,7 @@
         "modules/nf-core/fcs/fcsadaptor",
         "modules/nf-core/fcs/fcsgx",
         "modules/nf-core/force/cube",
+        "modules/nf-core/force/tileextent",
```
olgabot, line 46: Why does this need to be added to `skip_nf_test.json`? It seems like you have `nf-tests` setup

modules/nf-core/force/tileextent/main.nf
```diff
@@ -0,0 +1,43 @@
+process FORCE_TILEEXTENT {
+    tag { aoi.simpleName }
```
olgabot:
```suggestion
    tag "${aoi.simpleName}"
```

Are the braces necessary?

olgabot (commented):
This looks good to me, but I feel too much power/responsibility for approving something into nf-core/modules right now, so I defer to @maxulysse and @edmundmiller!

.github/skip_nf_test.json
Felix-Kummer: This is within the `conda` scope. The tool in this module is not available on conda. Thus, we skip the tests with the `conda` profile.

modules/nf-core/force/tileextent/main.nf
```diff
@@ -0,0 +1,43 @@
+process FORCE_TILEEXTENT {
+    tag "${aoi.simpleName}"
+    label 'process_single'
+
+    container "nf-core/force:3.8.01"
+
+    input:
+    path aoi
+    path "tmp/datacube-definition.prj"
```
SPPearce: What is this?
If it is a file that you wish to stage into this location, can I suggest using 
```suggestion
    path(datacube-definition, stageAs: "tmp/datacube-definition.prj")
```

modules/nf-core/force/tileextent/tests/nextflow.shp.config
```diff
@@ -0,0 +1,4 @@
+process {
+    // this is required because the module needs all shapefile files in the same directory
+    stageInMode = 'copy'
```
SPPearce: I think this should be being set in the input staging instead.

modules/nf-core/force/tileextent/main.nf
```diff
@@ -0,0 +1,43 @@
...
+    path shapefile_dbf
```
SPPearce, line 11:
```suggestion
    path(shapefile_dbf, stageInMode: 'copy')
```
possibly this is what you want to do.

SPPearce (changes requested):
I have no idea on your workflow here, but you have no meta map of any kind, but always emit a file called `tile_allow`. Do you only ever run this process once?

modules/nf-core/force/tileextent/main.nf
Felix-Kummer: The problem is that the tool requires a file with that exact name. (Likely a remnant of the initial monolithic nature of the tool).

modules/nf-core/force/tileextent/main.nf
Felix-Kummer: This option does not exist on the level of individual inputs. Its also not in the Nextflow docs and Nextflow says:
` ERROR ~ No such variable: stageInMode`

Felix-Kummer:
@SPPearce The module runs once in the beginning of a pipeline to restrict the analysis extent for downstream tasks. A meta map does not make sense here as this is not specific to a single image but rather the entire extent.

modules/nf-core/force/tileextent/tests/nextflow.shp.config
Felix-Kummer: Not sure where to specify this, see comment below.

modules/nf-core/force/tileextent/main.nf
SPPearce: Yep, that's fine, so you stage it as exactly that required filename.

modules/nf-core/force/tileextent/main.nf
SPPearce: Ok, so can't be set at the input level.
You could set it in the process level though, by setting `stageInMode 'copy'` underneath the tag, [several other modules do this](https://github.com/search?q=repo%3Anf-core%2Fmodules%20stageInMode&type=code).

Felix-Kummer: That's a good point

modules/nf-core/force/tileextent/Dockerfile
SPPearce, line 1: I don't understand why this is here?

Felix-Kummer: nf-core modules guidelines: https://nf-co.re/docs/guidelines/components/modules#software-not-on-bioconda

Felix-Kummer: (the tool is not on bioconad and can't get there as it is not a bioinformatics tool)

SPPearce: But this Dockerfile contains nothing except building from the build that it should be creating?

Felix-Kummer: yes, maybe I misinterpret the guidelines here.

Felix-Kummer: @SPPearce we had [this conversation](https://nfcore.slack.com/archives/CJRH30T6V/p1750332391291719) on Slack regarding the Dockerfile. Do you have a preference for resolving this?

PR #8637, opened 2025-06-14
HTODemux

LuisHeinzlmeier:
- `HTODemux` is the demultiplexing module of `Seurat`, which demultiplex samples based on data from cell hashing
- This modules was created to be included in the `hadge` pipeline

### DIFF
--- a/modules/nf-core/htodemux/main.nf
+++ b/modules/nf-core/htodemux/main.nf
@@ -11,9 +11,9 @@ process HTODEMUX {
     tuple val(meta), path(seurat_object), val(assay)
 
     output:
-    tuple val(meta), path("*params_htodemux.csv")              , emit: params
-    tuple val(meta), path("*assignment_htodemux.csv")          , emit: assignment
-    tuple val(meta), path("*classification_htodemux.csv")      , emit: classification
+    tuple val(meta), path("*_params_htodemux.csv")             , emit: params
+    tuple val(meta), path("*_assignment_htodemux.csv")         , emit: assignment
+    tuple val(meta), path("*_classification_htodemux.csv")     , emit: classification
     tuple val(meta), path("*_htodemux.rds")                    , emit: rds
     path "versions.yml"                                        , emit: versions
 
--- a/modules/nf-core/htodemux/main.nf
+++ b/modules/nf-core/htodemux/main.nf
@@ -7,9 +7,6 @@ process HTODEMUX {
         'oras://community.wave.seqera.io/library/r-seurat_r-seuratobject:4c5a804804327d29':
         'community.wave.seqera.io/library/r-seurat_r-seuratobject:b11306d1bdc82827' }"
 
-
-
-    // community.wave.seqera.io/library/r-seurat_r-seuratobject:b11306d1bdc82827
     input:
     tuple val(meta), path(seurat_object)
 
@@ -36,13 +33,10 @@ process HTODEMUX {
     stub:
     def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
-    // TODO nf-core: A stub section should mimic the execution of the original module as best as possible
-    //               Have a look at the following examples:
-    //               Simple example: https://github.com/nf-core/modules/blob/818474a292b4860ae8ff88e149fbcda68814114d/modules/nf-core/bcftools/annotate/main.nf#L47-L63
-    //               Complex example: https://github.com/nf-core/modules/blob/818474a292b4860ae8ff88e149fbcda68814114d/modules/nf-core/bedtools/split/main.nf#L38-L54
     """
-
-    touch ${prefix}.bam
+    touch ${prefix}_assignment_htodemux.csv
+    touch ${prefix}_classification_htodemux.csv
+    touch ${prefix}_htodemux.rds
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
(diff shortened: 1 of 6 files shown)


modules/nf-core/htodemux/main.nf
```diff
@@ -0,0 +1,48 @@
...
+    kfunc = task.ext.kfunc ?: "clara"
+    nsamples = task.ext.nsamples ?: "100"
+    seed = task.ext.seed ?: '42'
+    verbose = task.ext.verbose ?: 'TRUE'
+    prefix = task.ext.prefix ?: "${meta.id}"
+
+    template 'HTODemux.R'
+
+    stub:
+    def args = task.ext.args ?: ''
```
nictru:
```suggestion
```

Not needed here

modules/nf-core/htodemux/main.nf
```diff
@@ -0,0 +1,48 @@
...
+    output:
+    tuple val(meta), path("*.csv")          , emit: csv
+    tuple val(meta), path("*.rds")          , emit: rds
+    path "versions.yml"                     , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    assay = task.ext.assay ?: "HTO"
```
nictru: This should rather be an input

modules/nf-core/htodemux/meta.yml
```diff
@@ -0,0 +1,61 @@
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/meta-schema.json
+name: "htodemux"
+description: Demultiplex samples based on data from cell hashing.
+keywords:
+  - demultiplexing
+  - hashing-based deconvoltion
```
nictru:
```suggestion
  - hashing-based deconvolution
```

modules/nf-core/htodemux/meta.yml
```diff
@@ -0,0 +1,61 @@
...
+input:
+  - - meta:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. `[ id:'sample1' ]`
+    - seurat_object:
+        type: file
+        description: |
+          A `.rds` file containing the seurat object. Assumes that the hash tag oligo (HTO) data has been added and normalized.
```
nictru, line 28: Add a note on the `assay` input here

modules/nf-core/htodemux/meta.yml
```diff
@@ -0,0 +1,61 @@
...
+  - csv:
+      - meta:
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. `[ id:'sample1' ]`
+      - "*.csv":
+          type: file
+          description: CSV's for the classification, assignment and parameters.
+          pattern: "*.csv"
```
nictru: If this is a constant number of CSV files with different purposes, there should be a separate output channel for each type of CSV file. Otherwise it's hard extract the necessary files for downstream steps

modules/nf-core/htodemux/meta.yml
nictru: If we expect that a `HTO` assay exists in the input RDS object, we should mention this here. Also we should explain what users can do, if the assay is stored with a different name.

nictru (changes requested):

modules/nf-core/htodemux/meta.yml
LuisHeinzlmeier: Adopted!

modules/nf-core/htodemux/main.nf
LuisHeinzlmeier: Removed it!

modules/nf-core/htodemux/main.nf
LuisHeinzlmeier: Added `assay` as an input and improved/reduced the R script.

modules/nf-core/htodemux/meta.yml
LuisHeinzlmeier: Added an own output stream for every file!

modules/nf-core/htodemux/meta.yml
LuisHeinzlmeier: Added `assay` as input and described what to do if it is not named `HTO`.

modules/nf-core/htodemux/templates/HTODemux.R
```diff
@@ -0,0 +1,100 @@
...
+################################################
+## Main Process                               ##
+################################################
+################################################
+
+# Loading Seurat object
+hashtag <- readRDS(seuratObj)
+
+# Demultiplex cells based on HTO enrichment
+# zu einem int, boolean casten
```
nictru:
```suggestion
```

modules/nf-core/htodemux/templates/HTODemux.R
```diff
@@ -0,0 +1,100 @@
...
+if (kfunc == "clara") {
+    hashtag <- HTODemux(hashtag, assay = assay, positive.quantile = quantile, init = init, nstarts = nstarts, kfunc = "clara", seed = seed, verbose = verbose)
+} else {
+    hashtag <- HTODemux(hashtag, assay = assay, positive.quantile = quantile, init = init, nstarts = nstarts, kfunc = "kmeans", seed = seed, verbose = verbose)
+}
```
nictru: `kfunc` will always be `clara`, check line 19

nictru: Also is it really necessary to have the if/else if only the `kfunc` is different?

LuisHeinzlmeier: removed if else statement and used nextflow parameter at the start

modules/nf-core/htodemux/templates/HTODemux.R
LuisHeinzlmeier: Removed comment

PR #8643, opened 2025-06-16
New module: vclust/align

Joon-Klaps:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/vclust/align/tests/main.nf.test
+++ b/modules/nf-core/vclust/align/tests/main.nf.test
@@ -1,0 +1,100 @@
+nextflow_process {
+
+    name "Test Process VCLUST_ALIGN"
+    script "../main.nf"
+    process "VCLUST_ALIGN"
+
+    tag "modules"
+    tag "modules_nfcore"
+    tag "vclust"
+    tag "vclust/align"
+    tag "vclust/prefilter"
+
+    test("sarscov2 - contigs") {
+
+        when {
+            process {
+                """
+
+                input[0] = [
+                    [ id:'test', single_end:false ], // meta map
+                    file(params.modules_testdata_base_path + 'genomics/sarscov2/illumina/fasta/contigs.fasta', checkIfExists: true),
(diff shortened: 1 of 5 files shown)


LeonHafner:
Why not generate the alignment file by default? Do you think it's too much overhead? I'm not sure how large these files can get.
Generating it by default would remove the need for an extra input option and the associated logic.
The file would always be written to the output channel, and users could use it as needed.

If that's too costly, I'm fine sticking with the current setup.

Joon-Klaps:
Since its not generated by default, i kept that logic. I'm not entirely sure how vclust filters the alignment results but they can get pretty big when you have a lot of genomes that have high ANI.

modules/nf-core/vclust/align/tests/main.nf.test
```diff
@@ -0,0 +1,133 @@
...
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(
+                    process.out.txt,
+                    process.out.ids,
+                    file(process.out.aln[0][1]).readLines()[0],
```
LeonHafner: That's probably because the alignments aren't sorted, right?

Joon-Klaps: Yes, tried sorting the table with the nf-tes csv plugin but that also didn't result in a stable result. some rows still weren't matching, the table is also pretty big for a snapshot as it will include all rows (42) and columns. Making the snapshot difficult to follow. Hence I snapshoted jus the number of rows and the column names.

LeonHafner (approved):
LGTM!

Joon-Klaps:
@LeonHafner thanks for the review! I noticed at the very end that there was an output file missing in the snapshot, i fixed it with the final commit.

LeonHafner:
Good catch, feel free to proceed with the merge.

PR #8646, opened 2025-06-16
Make sure to only skip nf-test for meta.yml files

mashehu:
The current setup also skips if additional files are changed. this checks first which files are changed.


mirpedrol (approved):
LGTM

PR #7727, opened 2025-03-02
`ASCAT` migration to nf-test

LouisLeNezet:
## PR checklist

Closes #7533 #5170
Should also close PR #5314

- [ ] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [ ] Remove all TODO statements.
- [ ] Emit the `versions.yml` file.
- [ ] Follow the naming conventions.
- [ ] Follow the parameters requirements.
- [ ] Follow the input/output options guidelines.
- [ ] Add a resource `label`
- [ ] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - For modules:
    - [ ] `nf-core modules test <MODULE> --profile docker`
    - [ ] `nf-core modules test <MODULE> --profile singularity`
    - [ ] `nf-core modules test <MODULE> --profile conda`
  - For subworkflows:
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile docker`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile singularity`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile conda`

PR #8695, opened 2025-06-23
Fix ASCAT for both folder or file input

SPPearce:
In #7727 in order to get the code to work with a test input of a single file, a change was made that means that the module doesn't work for a folder input (as is the expected typical use).
This adds some code to enable using either a single file or a folder of files.


Joon-Klaps (commented):
Wouldn't it be good to add a test for this? I don't think I saw them in #7727

SPPearce:
Yes, it would be good to have a test like this, but it requires a folder input rather than a single file
We don't have a test that uses multiple files at the moment.

SPPearce:
Thanks Maxime!
We do need to improve the tests for ASCAT here I think, as well as incorporating some non-stub tests into sarek

PR #8698, opened 2025-06-24
New module: sigprofiler

kdavydzenka:
## PR checklist

Closes #XXX 

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/sigprofiler/tests/main.nf.test
+++ b/modules/nf-core/sigprofiler/tests/main.nf.test
@@ -94,9 +94,7 @@ nextflow_process {
                         }
                         
                     } 
-      
-                    // Snapshot the full status map
-                    assert snapshot(filesStatus).match("outputs")  
+        
                 }
             )
         }  
--- a/modules/nf-core/sigprofiler/tests/main.nf.test.snap
+++ b/modules/nf-core/sigprofiler/tests/main.nf.test.snap
@@ -1,30 +1,4 @@
 {
-    "outputs": {
-        "content": [
-            {
-                "SBS96": {
-                    "SBS96/SBS96/Samples.txt": "missing",
-                    "SBS96/SBS96/Suggested_Solution/COSMIC_SBS96_Decomposed_Solution/Signatures/COSMIC_SBS96_Signatures.txt": "missing",
-                    "output/SBS/test.SBS96.all": "present"
-                },
-                "DBS78": {
-                    "DBS78/DBS78/Samples.txt": "present",
-                    "DBS78/DBS78/Suggested_Solution/COSMIC_DBS78_Decomposed_Solution/Signatures/COSMIC_DBS78_Signatures.txt": "present",
-                    "output/DBS/test.DBS78.all": "present"
-                },
-                "ID83": {
-                    "ID83/ID83/Samples.txt": "present",
-                    "ID83/ID83/Suggested_Solution/COSMIC_ID83_Decomposed_Solution/Signatures/COSMIC_ID83_Signatures.txt": "present",
-                    "output/ID/test.ID83.all": "present"
-                }
-            }
-        ],
-        "meta": {
-            "nf-test": "0.9.2",
-            "nextflow": "24.10.3"
-        },
-        "timestamp": "2025-08-18T22:25:05.413683317"
-    },
     "versions": {
         "content": [
             [


modules/nf-core/sigprofiler/main.nf
```diff
@@ -0,0 +1,38 @@
...
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    template "main_script.py"
+   
+
+    stub:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
```
LouisLeNezet:
```suggestion
    def args   = task.ext.args   ?: ''
    def prefix = task.ext.prefix ?: "${meta.id}"
```

modules/nf-core/sigprofiler/main.nf
```diff
@@ -0,0 +1,38 @@
...
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/mulled-v2-b6141a7b8f0674ac604d90eb1306a731da24a734:daf9213409b038023df2a741f058d1bfd66b3c4c-0':
+        'biocontainers/mulled-v2-b6141a7b8f0674ac604d90eb1306a731da24a734:daf9213409b038023df2a741f058d1bfd66b3c4c-0' }"
+
+    input:
+    tuple val(meta), path(tsv_list, stageAs: '*.tsv')
+    
+    output:
+    tuple val(meta), path("results/*"),    emit: sigprofiler_results
+    path "versions.yml",                   emit: versions
```
LouisLeNezet:
```suggestion
    tuple val(meta), path("results/*"), emit: sigprofiler_results
    path "versions.yml",                emit: versions
```

modules/nf-core/sigprofiler/main.nf
```diff
@@ -0,0 +1,38 @@
...
+    template "main_script.py"
+   
+
+    stub:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+
+    """
+    mkdir -p results
+    echo "stub output" > results/dummy.txt
```
LouisLeNezet: Instead of a dummy.txt, could you create file with the same names that would be created ?
```suggestion
    touch results/dummy.txt
```

modules/nf-core/sigprofiler/templates/main_script.py
```diff
@@ -0,0 +1,177 @@
...
+            continue
+        key, val = item.split(':', 1)
+        parsed_args[key.strip()] = val.strip().strip("'").strip('"')
+    return parsed_args
+
+# Default options
+
+opt = dict()
+opt["prefix"] = "${task.ext.prefix ?: meta.id}"
+opt["dataset"] = "${task.ext.dataset ?: meta.dataset}"
```
LouisLeNezet: If the dataset is always expected to be provided, can you add it as a value channel and add it to the meta.yml ?

modules/nf-core/sigprofiler/templates/main_script.py
```diff
@@ -0,0 +1,177 @@
...
+    "min_nmf_iterations": 10000,
+    "max_nmf_iterations": 100000,
+    "nmf_test_conv": 10000,
+    "seeds": "random",
+    "volume": "./",
+    "get_all_signature_matrices": True,
+    "make_decomposition_plots": True,
+    "download_genome_sigprofiler": True,
+    "genome_installed_path": ""
+    })
```
LouisLeNezet:
```suggestion
})
```

modules/nf-core/sigprofiler/tests/main.nf.test
```diff
@@ -0,0 +1,76 @@
...
+            }
+        }
+
+        then {
+            def (meta, resultDirs) = process.out.sigprofiler_results[0]
+            
+            assertAll( 
+                { assert process.success },
+                { assert snapshot(
+                    process.out.versions
```
LouisLeNezet: Could you also snapshot the interior of the result directory (even if only the names) ?
Could you also add `path(process.out.versions[0]).yaml` to ensure a correct version yml file ?

modules/nf-core/sigprofiler/tests/main.nf.test
```diff
@@ -0,0 +1,76 @@
...
+                    file(params.modules_testdata_base_path + 'delete_me/sigprofiler/test_GRCh38/HCC1395_WXS_HCC1395_joint_table.tsv', checkIfExists: true),
+              ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
```
LouisLeNezet:
```suggestion
                { assert snapshot(
                    process.out,
                    path(process.out.versions[0]).yaml
                ).match() }
```

tests/config/nf-test.config
```diff
@@ -47,6 +47,7 @@ profiles {
     singularity {
         singularity.enabled     = true
         singularity.autoMounts  = true
+        singularity.runOptions  = '--writable-tmpfs'
```
LouisLeNezet: I don't think this should be modify in this PR.
Why did you add this ?

modules/nf-core/sigprofiler/templates/main_script.py
```diff
@@ -0,0 +1,177 @@
...
+    genome = opt["genome"]
+    input_type = opt["input_type"]
+    context_types = opt["context_type"].split(",")
+    tsv_list = "${tsv_list.join(' ')}"
+
+    input_path = os.path.join(dataset_id)
+
+    if not os.path.exists(input_path):
+            os.mkdir(input_path)
+
```
LouisLeNezet:
```suggestion
```

modules/nf-core/sigprofiler/templates/main_script.py
```diff
@@ -0,0 +1,177 @@
...
+    "genome_installed_path": ""
+    })
+
+# Parse extra args
+args_opt = parse_args("${task.ext.args ?: ''}")
+for ao_k, ao_v in args_opt.items():
+    if ao_k in {"minimum_signatures", "maximum_signatures", "nmf_replicates", "min_nmf_iterations", "max_nmf_iterations", "nmf_test_conv"}:
+        opt[ao_k] = int(ao_v)
+    elif ao_k in {"get_all_signature_matrices", "make_decomposition_plots", "download_genome_sigprofiler"}:
+        opt[ao_k] = ao_v.lower() == "true"
```
LouisLeNezet: Could you add an error check if the ao_v.lower is different from false or true ?

modules/nf-core/sigprofiler/templates/main_script.py
```diff
@@ -0,0 +1,177 @@
...
+        f"SigProfilerMatrixGenerator matrix_generator "
+        f"-v {opt['volume']} "
+        f"{dataset_id} {genome} {input_path}"
+    )  
+    subprocess.run(generate_count_matrix, shell=True)
+
+    matrix_files = {
+            "SBS96": os.path.join("output", "SBS", f"{dataset_id}.SBS96.all"),
+            "DBS78": os.path.join("output", "DBS", f"{dataset_id}.DBS78.all"),
+            "ID83": os.path.join("output", "ID", f"{dataset_id}.ID83.all")
```
LouisLeNezet: Is it expected to be this files for each genome ?

modules/nf-core/sigprofiler/tests/main.nf.test
```diff
@@ -0,0 +1,76 @@
+nextflow_process {
+
+    name "Test Process SIGPROFILER"
+    script "../main.nf"
+    process "SIGPROFILER"
+
+    tag "modules"
+    tag "modules_nfcore"
+    tag "sigprofiler"
+
```
LouisLeNezet, line 10: Could you add a test with genome_installed_path and download_genome_sigprofiler: False ?

LouisLeNezet (changes requested):
Hi,
I'm not yet quite familiar on python modules but here are some comment !

tests/config/nf-test.config
kdavydzenka: Thank you for pointing that out. Initially, I added `singularity.runOptions = '--writable-tmpfs'` because I encountered an issue when testing the` sigprofiler `tool, specifically when` make_decomposition_plots = true`. That option required a writable` /tmp` inside the container. However, I've adjusted the test setup to use `make_decomposition_plots = false`, which avoids the need for a writable `/tmp.` So this change is no longer necessary, and I removed it from this PR

modules/nf-core/sigprofiler/templates/main_script.py
kdavydzenka: Yes, these files are expected to be generated once per dataset for each genome.

modules/nf-core/sigprofiler/templates/main_script.py
kdavydzenka: Added a check to ensure `ao_v.lower()` is either `true` or `false`, and raise a` ValueError `if it's not.

modules/nf-core/sigprofiler/main.nf
```diff
@@ -0,0 +1,38 @@
...
+    output:
+    tuple val(meta), path("results/*"), emit: sigprofiler_results
+    path "versions.yml",                emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    template "main_script.py"
+
```
LouisLeNezet:
```suggestion
```

tests/config/nf-test.config
LouisLeNezet: Could you check the PR, it seems that your change wasn't taken into account.

modules/nf-core/sigprofiler/tests/main.nf.test
kdavydzenka: Unfortunately, it's not feasible to add this test. This is because the required genome data is specific to the tool and too large (~3 GB) to be published in the `nf-core/test-datasets`, even in a subsampled form (the tool requires all genome .txt files to be fully installed), which makes a minimal test case impractical.
I was able to run this test successfully in my local environment to verify functionality.

modules/nf-core/sigprofiler/templates/main_script.py
kdavydzenka: The dataset is always expected to be provided, and it has already been included in the meta.yml via the `meta` map.

LouisLeNezet: After looking a bit more how dataset is used. It seems to be redundant with `prefix`.
From what I understood, `dataset` is only used to name the output file and folder.

Could you change it to use prefix therefore ?

itrujnara:
Hi @kdavydzenka, thanks for contributing to nf-core. Modules are designed to be interoperable and easily integrated into pipelines. For this reason, we have regulations on module structure. Here are some relevant parts of the [module guidelines](https://nf-co.re/docs/guidelines/components/modules):
>Modules MUST NOT use ‘custom’ hardcoded meta fields. This means both that they should not be referred to within the module as expected input, nor generate new fields as output. The only accepted ‘standard’ meta fields are `meta.id` or `meta.single_end`.

>Input channel `val` declarations SHOULD be defined for all mandatory non-file inputs that are essential for the functioning of the tool (e.g. parameters, flags etc).

As `dataset` is not conventional, and highly context-dependent, it is better to pass it through a channel. If you believe there is a major reason to break the guidelines, please provide a deeper explanation. We appreciate your cooperation on this matter.

SPPearce:
It looks to me like this `dataset` is only being used in the code to set the name of the output files (and a column inside them) - this is what we generally use `prefix` for, setting by default to be `meta.id`.

kdavydzenka:
> Hi @kdavydzenka, thanks for contributing to nf-core. Modules are designed to be interoperable and easily integrated into pipelines. For this reason, we have regulations on module structure. Here are some relevant parts of the [module guidelines](https://nf-co.re/docs/guidelines/components/modules):
> 
> > Modules MUST NOT use ‘custom’ hardcoded meta fields. This means both that they should not be referred to within the module as expected input, nor generate new fields as output. The only accepted ‘standard’ meta fields are `meta.id` or `meta.single_end`.
> 
> > Input channel `val` declarations SHOULD be defined for all mandatory non-file inputs that are essential for the functioning of the tool (e.g. parameters, flags etc).
> 
> As `dataset` is not conventional, and highly context-dependent, it is better to pass it through a channel. If you believe there is a major reason to break the guidelines, please provide a deeper explanation. We appreciate your cooperation on this matter.

Hi @itrujnara,
Thanks for pointing out to module structure guidelines.
I've addressed the issue by updating the module to pass the `dataset` as an input channel, using a `prefix `instead of a hardcoded meta field.

modules/nf-core/sigprofiler/templates/main_script.py
kdavydzenka: I've addressed the issue by updating the module to pass the `dataset` as an input channel, using a `prefix `instead of a hardcoded meta field.

kdavydzenka:
> Could you set the dataset as an input value channel ?
> 
> Feel free to ask me for a review when you are ready !

Hi @LouisLeNezet,
I've addressed the issue by updating the module to pass the `dataset` as an input channel, using a `prefix `instead of a hardcoded meta field. 
At the moment, I'm experiencing some issues running the `nf-core tests` on my local cluster. Once that's resolved, I’ll run the full test suite and fix any remaining bugs in the module.

modules/nf-core/sigprofiler/main.nf
```diff
@@ -0,0 +1,38 @@
...
+        'https://depot.galaxyproject.org/singularity/mulled-v2-b6141a7b8f0674ac604d90eb1306a731da24a734:daf9213409b038023df2a741f058d1bfd66b3c4c-0':
+        'biocontainers/mulled-v2-b6141a7b8f0674ac604d90eb1306a731da24a734:daf9213409b038023df2a741f058d1bfd66b3c4c-0' }"
+
+    input:
+    tuple val(meta), path(tsv_list, stageAs: '*.tsv')
+    path(genome_installed_path)  //optional
+    
+    output:
+    tuple val(meta), path("results/*"), emit: sigprofiler_results
+    path "versions.yml",                emit: versions
```
itrujnara: Align the commas correctly

modules/nf-core/sigprofiler/main.nf
```diff
@@ -0,0 +1,38 @@
...
+    def args   = task.ext.args   ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+
+    """
+    mkdir -p results/SBS96/SBS96/
+    touch results/SBS96/SBS96/Samples.txt
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        sigprofilermatrixGenerator: \$(python3 -c "import SigProfilerMatrixGenerator as matGen; print(matGen.__version__)")
```
itrujnara, line 58: You should also output Python version

modules/nf-core/sigprofiler/meta.yml
```diff
@@ -0,0 +1,60 @@
...
+          e.g. `[ id:'sample1' ]`
+    - tsv_list:
+        type: file
+        description: joint tsv file with validated mutations by CNAqc
+        pattern: "*.{tsv}"
+  - - genome_installed_path: 
+        type: file
+        description: path to installed reference genome (optional)
+output:
+  - sigprofiler_results:
```
itrujnara: `output` should be an object, not a list (i.e., no hyphens)

modules/nf-core/sigprofiler/tests/main.nf.test
```diff
@@ -0,0 +1,80 @@
...
+        options "-stub"
+
+        when {
+            process {
+                """ 
+                input[0] = [
+                    [ id:'test' ], 
+                    file(params.modules_testdata_base_path + 'delete_me/sigprofiler/test_GRCh38/HCC1395_WXS_HCC1395_joint_table.tsv', checkIfExists: true),
+                ] 
+                input[1] = []
```
itrujnara: Can we get a test where this option is used?

itrujnara (commented):
Hi, the prefix handling looks good now, thanks! I've left some comments on other issues to address

modules/nf-core/sigprofiler/main.nf
itrujnara: This also applies to the versions from the Python template script

modules/nf-core/sigprofiler/tests/main.nf.test
kdavydzenka: > Can we get a test where this option is used?

Unfortunately, it's not feasible to add this test: `genome_installed_path and download_genome_sigprofiler: False `. This is because the required genome data is specific to the tool and too large (~3 GB) to be published in the `nf-core/test-datasets`, even in a subsampled form (the tool requires all genome .txt files to be fully installed), which makes a minimal test case impractical.
I was able to run this test successfully in my local environment to verify functionality.

kdavydzenka:
Dear @LouisLeNezet and @itrujnara,
I’ve addressed all your latest comments and made some minor updates to the module. 
Please let me know if anything else needs attention.

modules/nf-core/sigprofiler/tests/main.nf.test
```diff
@@ -0,0 +1,80 @@
...
+                    process.out.versions,
+                    path(process.out.versions[0]).yaml
+                ).match() },
+                {
+                    // Directory existence check
+                    resultDirs.each { dir ->
+                        def file = new File(dir.toString())
+                        assert file.exists() 
+                        
+                    }
```
LouisLeNezet: Better than just checking if the files exist could you add a recursive file(it).name in the snapshot ?
This would help to diagnose if the prefix are correctly use and if the file are properly named.
Using this snapshot you will also be able to complete the stub test to be more representative of a real processing

modules/nf-core/sigprofiler/templates/main_script.py
```diff
@@ -0,0 +1,186 @@
...
+    df = data.assign(**new_columns)
+    df['chr'] = df['chr'].astype(str).str[3:]
+    df = df.rename(columns={'Indiv': 'Sample', 'chr': 'chrom', 'from': 'pos_start', 'to': 'pos_end'})
+    df["ID"] = df["Sample"]
+    return df.loc[:, ['Project', 'Sample', 'ID', 'Genome', 'mut_type', 'chrom', 'pos_start', 'pos_end', 'ref', 'alt', 'Type']]
+    
+if __name__ == '__main__':
+
+    prefix = opt["prefix"]
+    genome = opt["genome"]
```
LouisLeNezet, line 86: What happen if no genome is provided ?
If this value is mandatory I would add it as a value channel and document it in the meta.yml.
Same for the context_types.

LouisLeNezet (commented):
Thanks for adressing the different comment.
I'm not quite familiar with python processes yet.

I have a few comment left.

Could you also add tests that would throw the different errors ?

modules/nf-core/sigprofiler/tests/main.nf.test
kdavydzenka: I’ve updated the test so that it checks for the existence of a small, representative set of files for each detected context type. The tool produces a large number of subfolders and files in the `results/* `directory, and the exact structure depends on which mutational contexts are present in the input dataset. Now test dynamically detects the contexts produced and verifies the presence of representative files for each one.

modules/nf-core/sigprofiler/templates/main_script.py
kdavydzenka: @LouisLeNezet thank you for the suggestion. The `genome `argument is indeed mandatory for this process to run correctly. I have added` genome` as a value channel in the input and documented it in `meta.yml` as well, ensuring it is always provided. 
Regarding `context_types`, these are optional and provided as default arguments to the tool. If not explicitly specified, the tool automatically runs on all mutational contexts detected in the input data.

kdavydzenka:
> Thanks for adressing the different comment. I'm not quite familiar with python processes yet.
> 
> I have a few comment left.
> 
> Could you also add tests that would throw the different errors ?

Hi @LouisLeNezet,
I’ve addressed your latest suggestions and added a couple of tests that trigger the different error cases as well. However, I wasn’t able to find clear examples in the nf-core documentation on the best way to write asserts for these error scenarios.

modules/nf-core/sigprofiler/templates/main_script.py
```diff
@@ -0,0 +1,181 @@
...
+    input_type = opt["input_type"]
+    context_types = opt["context_type"].split(",")
+    tsv_list = "${tsv_list.join(' ')}"
+
+    if not os.path.exists(prefix):
+           os.mkdir(prefix)
+
+    data = process_tsv_join(tsv_list)
+    processed = input_processing(data, prefix, genome)
+    processed.to_csv(f"{prefix}/input_data.txt", sep="\t", index=False)
```
LouisLeNezet, line 96: Would it be possible to add this file to the stub as it would always be created in a normal run ?

modules/nf-core/sigprofiler/tests/main.nf.test
```diff
@@ -0,0 +1,160 @@
...
+                    [ id:'test' ],
+                    file(params.modules_testdata_base_path + 'delete_me/sigprofiler/test_GRCh38/HCC1395_WXS_HCC1395_joint_table.tsv', checkIfExists: true),
+                ]
+                input[1] = "GRCh38"
+                input[2] = []
+                """
+            }
+        }
+        then {
+             assert !process.success
```
LouisLeNezet: This could be simplified to `assert process.failed` 😉
You could also add a check of the message in the output such as:
```
assertAll(
    { assert process.failed },
    { assert process.errorReport.contains("download_genome_sigprofiler is False but no genome_installed_path was provided.") }
)
```

modules/nf-core/sigprofiler/main.nf
```diff
@@ -0,0 +1,49 @@
...
+    touch results/SBS96/SBS96/Samples.txt
+    touch results/SBS96/SBS96/Suggested_Solution/COSMIC_SBS96_Decomposed_Solution/Signatures/COSMIC_SBS96_Signatures.txt
+    
+    mkdir -p results/ID83/ID83/Suggested_Solution/COSMIC_ID83_Decomposed_Solution/Signatures/
+    touch results/ID83/ID83/Samples.txt
+    touch results/ID83/ID83/Suggested_Solution/COSMIC_ID83_Decomposed_Solution/Signatures/COSMIC_ID83_Signatures.txt
+    
+    mkdir -p results/DBS78/DBS78/Suggested_Solution/COSMIC_DBS78_Decomposed_Solution/Signatures/
+    touch results/DBS78/DBS78/Samples.txt
+    touch results/DBS78/DBS78/Suggested_Solution/COSMIC_DBS78_Decomposed_Solution/Signatures/COSMIC_DBS78_Signatures.txt
```
maxulysse: In my opinion, the path and filenames should be reconstructed from the input data, and not hardcoded.

SPPearce: This `SBS96` etc are coming from the tool (https://cancer.sanger.ac.uk/signatures/sbs/sbs96/).
Although why are they repeated (SBS96/SBS96) and nested so deeply into subfolders?

kdavydzenka: Hi @maxulysse and @SPPearce.
Thanks for pointing this out! 
I’ve reworked the stub section so that paths and filenames are derived dynamically from the input `context_type `parameter. 
Regarding the repeated/nested paths, the directory and filenames are dictated by [`SigProfiler`](https://osf.io/t6j7u/wiki/6.%20Output%20-%20Suggested%20Solution/) tool itself as well as output results nested structure

PR #8703, opened 2025-06-24
Add scanpy/highlyvariablegenes

nictru:
This is the third of 13 scanpy modules

### DIFF
--- a/modules/nf-core/scanpy/highlyvariablegenes/main.nf
+++ b/modules/nf-core/scanpy/highlyvariablegenes/main.nf
@@ -1,0 +1,50 @@
+process SCANPY_HIGHLYVARIABLEGENES {
+    tag "${meta.id}"
+    label 'process_medium'
+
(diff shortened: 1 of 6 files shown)


modules/nf-core/scanpy/highlyvariablegenes/main.nf
```diff
@@ -0,0 +1,50 @@
...
+    tuple val(meta), path("*.h5ad"), emit: h5ad
+    tuple val(meta), path("*.pkl") , emit: var
+    path "versions.yml"            , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    prefix = task.ext.prefix ?: "${meta.id}_hvgs"
+    batch_key = task.ext.batch_key ?: ""
```
SPPearce, line 24: Please don't add a new task.ext, should be an input channel I suspect

SPPearce (changes requested):

PR #8719, opened 2025-06-26
Add module: STARE

LeonHafner:
Transferring module from: https://github.com/nf-core/tfactivity/tree/dev/modules/local/stare

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/stare/main.nf
+++ b/modules/nf-core/stare/main.nf
@@ -56,13 +56,6 @@ process STARE {
     """
 
     stub:
-    def args = task.ext.args ?: ''
-
-    def path_bed_file       = bed_file       ? "-b ${bed_file}"       : ""
-    def path_exclude_bed    = exclude_bed    ? "-x ${exclude_bed}"    : ""
-    def path_genes          = genes          ? "-u ${genes}"          : ""
-    def path_contact_folder = contact_folder ? "-f ${contact_folder}" : ""
-    def path_existing_abc   = existing_abc   ? "-r ${existing_abc}"   : ""
     
     """
     mkdir -p ${meta.id}/Gene_TF_matrices
--- a/modules/nf-core/stare/main.nf
+++ b/modules/nf-core/stare/main.nf
@@ -18,36 +18,36 @@ process STARE {
     tuple val(meta8), path(existing_abc)    
 
     output:
-    tuple val(meta), path("${prefix}/Gene_TF_matrices/${prefix}_TF_Gene_Affinities.txt") , emit: affinities
-    path "versions.yml"                                                                  , emit: versions
+    tuple val(meta), path("${meta.id}/Gene_TF_matrices/${meta.id}_TF_Gene_Affinities.txt") , emit: affinities
+    path "versions.yml"                                                                    , emit: versions
 
     when:
     task.ext.when == null || task.ext.when
 
     script:
     def args = task.ext.args ?: ''
-    def prefix = task.ext.prefix ?: "${meta.id}"
 
-    def path_bed_file       = bed_file       ? "--bed_file ${bed_file}"             : ""
-    def path_exclude_bed    = exclude_bed    ? "--exclude_bed ${exclude_bed}"       : ""
-    def path_genes          = genes          ? "--genes ${genes}"                   : ""
-    def path_contact_folder = contact_folder ? "--contact_folder ${contact_folder}" : ""
-    def path_existing_abc   = existing_abc   ? "--existing_abc ${existing_abc}"     : ""
+    def path_bed_file       = bed_file       ? "-b ${bed_file}"       : ""
+    def path_exclude_bed    = exclude_bed    ? "-x ${exclude_bed}"    : ""
+    def path_genes          = genes          ? "-u ${genes}"          : ""
+    def path_contact_folder = contact_folder ? "-f ${contact_folder}" : ""
+    def path_existing_abc   = existing_abc   ? "-r ${existing_abc}"   : ""
 
     """
     STARE.sh \\
         ${args} \\
-        --annotation ${annotation} \\
-        --genome ${genome} \\
-        --psem ${psem} \\
-        --output ${prefix} \\
-        --cores ${task.cpus} \\
+        -a ${annotation} \\
+        -g ${genome} \\
+        -p ${psem} \\
+        -o ${meta.id} \\
+        -c ${task.cpus} \\
         ${path_bed_file} \\
         ${path_exclude_bed} \\
         ${path_genes} \\
         ${path_contact_folder} \\
         ${path_existing_abc}
 
+    gunzip -f ${meta.id}/Gene_TF_matrices/${meta.id}_TF_Gene_Affinities.txt.gz
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
@@ -57,7 +57,6 @@ process STARE {
 
     stub:
     def args = task.ext.args ?: ''
-    def prefix = task.ext.prefix ?: "${meta.id}"
 
     def path_bed_file       = bed_file       ? "--bed_file ${bed_file}"             : ""
     def path_exclude_bed    = exclude_bed    ? "--exclude_bed ${exclude_bed}"       : ""
@@ -66,7 +65,7 @@ process STARE {
     def path_existing_abc   = existing_abc   ? "--existing_abc ${existing_abc}"     : ""
     
     """
-    touch ${prefix}.bam
+    touch ${meta.id}/Gene_TF_matrices/${meta.id}_TF_Gene_Affinities.txt
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
--- a/modules/nf-core/stare/main.nf
+++ b/modules/nf-core/stare/main.nf
@@ -54,7 +54,6 @@ process STARE {
     """
 
     stub:
-    
     """
     mkdir -p ${meta.id}/Gene_TF_matrices
     touch ${meta.id}/Gene_TF_matrices/${meta.id}_TF_Gene_Affinities.txt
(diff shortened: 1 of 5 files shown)


LeonHafner:
For the inputs, we have two options:
```nextflow
input:
     tuple val(meta), path(annotation)
     tuple val(meta2), path(genome)
     tuple val(meta3), path(psem)
     tuple val(meta4), path(bed_file)
     tuple val(meta5), path(exclude_bed)
     tuple val(meta6), path(genes)
     tuple val(meta7), path(contact_folder)
     tuple val(meta8), path(existing_abc)
```

```nextflow
input:
     tuple val(meta), path(annotation), path(genome), path(psem)
     path(bed_file)
     path(exclude_bed)
     path(genes)
     path(contact_folder)
     path(existing_abc)
```
This is probably personal preference. The first approach matches how it's implemented in [nf-core/tfactivity](https://github.com/nf-core/tfactivity). The second might make upfront grouping simpler. The `meta{2-8}` objects aren't really needed since they aren’t passed on anyway.

modules/nf-core/stare/main.nf
```diff
@@ -0,0 +1,76 @@
...
+    """
+
+    stub:
+    def args = task.ext.args ?: ''
+
+    def path_bed_file       = bed_file       ? "-b ${bed_file}"       : ""
+    def path_exclude_bed    = exclude_bed    ? "-x ${exclude_bed}"    : ""
+    def path_genes          = genes          ? "-u ${genes}"          : ""
+    def path_contact_folder = contact_folder ? "-f ${contact_folder}" : ""
+    def path_existing_abc   = existing_abc   ? "-r ${existing_abc}"   : ""
```
nictru:
```suggestion
```
These are not used in the stub

modules/nf-core/stare/tests/main.nf.test
```diff
@@ -0,0 +1,117 @@
...
+
+    test("testing stare") {
+
+        setup {
+            run("GUNZIP", alias: "GUNZIP_ANNOTATION") {
+                script "../../gunzip/main.nf"
+                process {
+                    """
+                    input[0] = [
+                        [ id:'test', single_end:false ],
```
nictru:
```suggestion
                        [ id:'test' ],
```
The single_end is not relevant in this module, also remove in the occurences below

modules/nf-core/stare/meta.yml
```diff
@@ -0,0 +1,138 @@
...
+          - edam: http://edamontology.org/format_3475 # TSV
+  - - meta7:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. `[ id:'sample1' ]`
+    - contact_folder:
+        type: directory
+        description: "Directory of gzipped chromatin contact files per chromosome; set
+          to false to use distance-based estimate"
```
nictru: Is "set to false to use distance-based estimate" correct here? Would it not rather be "If not provided, a distance-based estimate will be used"?

modules/nf-core/stare/meta.yml
```diff
@@ -0,0 +1,138 @@
...
+        description: |
+          Groovy Map containing sample information
+          e.g. `[ id:'sample1' ]`
+    - bed_file:
+        type: file
+        description: "BED file with non-overlapping candidate regions; optional—if omitted
+          runs in promoter-mode around TSS"
+        pattern: "*.bed"
+        optional: true
+        ontologies: []
```
nictru: Is there no ontology for bed format?

modules/nf-core/stare/main.nf
```diff
@@ -0,0 +1,76 @@
...
+
+    input:
+    tuple val(meta), path(annotation)
+    tuple val(meta2), path(genome)
+    tuple val(meta3), path(psem)
+    tuple val(meta4), path(bed_file)
+    tuple val(meta5), path(exclude_bed)
+    tuple val(meta6), path(genes)
+    tuple val(meta7), path(contact_folder)
+    tuple val(meta8), path(existing_abc)
```
nictru: I think the `annotation` will be the genome GTF file, won't it? In this case, the order of inputs would not make a lot of sense here, as this is usually not sample-specific. It would make more sense to have the `bed_file` input first, because it is most sample-specific.

This is important because the `meta` from the first input channel will be used for the output.

LeonHafner: I reordered to match the order `STARE` uses in their documentation. The `bed_file` is an optional input, so makes no sense to put that first. I think leaving it like that makes it easier for users to compare to the documentation of `STARE` and maybe read more about the inputs there.

LeonHafner: I see your point with the `meta`, maybe we should use the second option as shown above.

modules/nf-core/stare/meta.yml
LeonHafner: How does an ontology for a file look like? I checked some random modules and none of them used it.

nictru: You used an ontology in line 89 of this file for TSV

nictru: Also for GZIP in line 113 and YAML in line 134

modules/nf-core/stare/main.nf
nictru: Also the second option only has reference genome inputs in the same tuple with the meta. Both your current implementation and the second option require some channel operations before the module will actually be usable.

In pipelines, we usually have a `ch_fasta`, a `ch_gtf` etc which are value channels with a meta and the corresponding file And will _always_ be the same for each sample. So if users have multiple samples, they will then have to `combine` their sample-specific bed files with at least the GTF and then multimap it, before it can be passed into the module.

The following things will be the same across all samples in most use cases:
- `annotation`
- `genome` (Fasta)
- `psem` (TF binding motifs)
- `exclude_bed` (genome-specific blacklist)
- Maybe `genes`, but this could also be sample-specific

And the following things will be sample-specific if provided:
- `bed_file`
- `contact_folder`
- `existing_abc`


So for me the most logical way of putting things would be:
```
tuple val(meta), path(bed_file), path(contact_folder), path(existing_abc)
tuple val(meta2), path(annotation)
tuple val(meta3), path(genome)
tuple val(meta4), path(psem)
tuple val(meta5), path(exclude_bed)
tuple val(meta6), path(genes)
```

Now we have the special situation that all of the files in the first input tuple are optional. I would still put it first because of the following reasons:

- Independently of the order, developers need to provide empty placeholders (`[]`) anyway
- If the pipeline developer actually wants to use the [promotor mode](https://stare.readthedocs.io/en/latest/Main.html#required-input:~:text=will%20run%20in-,promoter%2Dmode,-instead%20and%20construct), they just need to provide the first line as `[[id: 'whatever'], [], [], []]`
- It makes it eliminates the need for complicated channel operations in multi-sample setups

Generally I would say the `promoter mode` is not the primary use case of STARE. With the initially suggested input structures, promoter mode will be easily usable, but the "standard" mode will be complicated to use. My suggested solution might look a bit weird in promoter mode, but in my opinion clearly provides the better trade-off.

modules/nf-core/stare/meta.yml
LeonHafner: The friendly LLM from next door did it. I might have to double-check...

nictru: Have to confess I also used it for these xD Gets them surprisingly often right in my experience

modules/nf-core/stare/main.nf
LeonHafner: I agree, we can do it like that, even though I don't really like the idea of the first argument being optional...

nictru: I would say it's the smaller poison in this case

modules/nf-core/stare/tests/main.nf.test
```diff
@@ -0,0 +1,113 @@
+nextflow_process {
+
+    name "Test Process STARE"
+    script "../main.nf"
+    process "STARE"
+
+    tag "modules"
+    tag "modules_nfcore"
+    tag "stare"
+    tag "gunzip"
```
nictru, line 10:
```suggestion
    tag "tf-affinities"
```

modules/nf-core/stare/main.nf
```diff
@@ -0,0 +1,67 @@
...
+    gunzip -f ${meta.id}/Gene_TF_matrices/${meta.id}_TF_Gene_Affinities.txt.gz
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        stare: \$(STARE.sh --version | cut -f3 -d" ")
+    END_VERSIONS
+    """
+
+    stub:
+
```
nictru:
```suggestion
```

modules/nf-core/stare/tests/main.nf.test
```diff
@@ -0,0 +1,113 @@
...
+                process {
+                    """
+                    input[0] = [
+                        [ id:'test' ],
+                        file(params.modules_testdata_base_path + 'genomics/mus_musculus/genome/chr19.fa.gz', checkIfExists: true),
+                    ]
+                    """
+                }
+            }
+        }
```
nictru, line 37: Side note: you can also write the setup block outside of individual tests, then it is accessible in all tests

nictru (approved):

modules/nf-core/stare/tests/main.nf.test
```diff
@@ -0,0 +1,113 @@
+nextflow_process {
+
+    name "Test Process STARE"
+    script "../main.nf"
+    process "STARE"
+
+    tag "modules"
+    tag "modules_nfcore"
+    tag "stare"
+    tag "tf-affinities"
```
nictru:
```suggestion
    tag "gunzip"
```

PR #8726, opened 2025-06-27
Fix logic for gVCF vs. VCF

lvclark:
I noticed that the output file was set to be a VCF instead of a gVCF when I set `ext.args = "--gvcf --tmp-dir tmp"`.  It seems that the string and pattern were in the wrong order for the `=~` operator.

### DIFF
--- a/modules/nf-core/parabricks/deepvariant/main.nf
+++ b/modules/nf-core/parabricks/deepvariant/main.nf
@@ -25,7 +25,7 @@ process PARABRICKS_DEEPVARIANT {
     }
     def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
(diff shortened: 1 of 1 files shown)


famosab:
I think something is off with the gvcf related test then. We should fix that here too: https://github.com/nf-core/modules/blob/master/modules/nf-core/parabricks/deepvariant/tests/main.nf.test

famosab:
Maybe we can update this section to:
```
                { assert snapshot(
                    file(process.out.gvcf[0][1]).name,
                    path(process.out.gvcf[0][1]).vcf.getVariantsAsStrings(100),
                    process.out.versions,
                    ).match()
                }
```
To check that the filename is correctly generated.

famosab:
Fixed in #9078

famosab:
Closing this as it should be fixed by #9078 - feel free to reopen if something is still not working as expected :)

PR #8727, opened 2025-06-29
Update data types and versions

LuisHeinzlmeier:
- added correct version of used packages 
- updated correct datatypes that we can use it in the hadge pipeline


nictru:
Superseded by #8767

PR #8728, opened 2025-06-29
Update data types and alignment

LuisHeinzlmeier:
- updated to Harshil Alignment™️ format
- updated data types for the used parameters that we can use it in the hadge pipeline


nictru:
Superseded by #8767

PR #8751, opened 2025-07-04
New module: bascule

### LINKED ISSUE #8750: new module: bascule

### Further Information

_No response_

elena-buscaroli:
## PR checklist

Closes #8750 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


modules/nf-core/bascule/environment.yml
```diff
@@ -0,0 +1,14 @@
...
+  - conda-forge
+  - bioconda
+dependencies:
+  - bioconda::r-bascule=1.0.1
+  - conda-forge::compilers=1.9.0
+  - conda-forge::python=3.8.20
+  - conda-forge::pip=24.2
+  - conda-forge::r-ggplot2=3.5.2
+  - conda-forge::r-reticulate=1.42.0
+  - conda-forge::r-tidyverse=2.0.0
```
akaviaLab, line 12: You include the tidyverse, which is pretty big. Do you actually need the entire tidyverse, or can you pick only the pckages you need and get a smaller image?

modules/nf-core/bascule/templates/main_script.R
```diff
@@ -0,0 +1,125 @@
+#!/usr/bin/env Rscript
+
+parse_args = function(x) {
```
akaviaLab, line 3: This is very different from other parse_args R functions that I've seen in nf-core, see for example https://github.com/nf-core/modules/blob/efc7b2793391bb02bef37fb55d52e41b0b8b8839/modules/nf-core/gprofiler2/gost/templates/gprofiler2_gost.R#L41. 
I'm assuming that the standard one didn't work. Are the arguments to bascule very different from other R using modules?

modules/nf-core/bascule/templates/main_script.R
```diff
@@ -0,0 +1,125 @@
...
+  })
+
+  # Ensure the option vectors are length 2 (key/ value) to catch empty ones
+  args_vals = lapply(args_vals, function(z){ length(z) = 2; z})
+
+  parsed_args = structure(lapply(args_vals, function(x) x[2]), names = lapply(args_vals, function(x) x[1]))
+  parsed_args[! is.na(parsed_args)]
+}
+
+opt = list(
```
akaviaLab, line 22: There is more than one argument used in the script, so please set reasonable defaults in opt. Also, are the parameters set by args? Is it possible to move some of them into the task definition, allowing users to change it without messing around in configs?
I'm not familiar with bascule - to me the question is how likely are users going to want to change any parameter.
If very likely - I think it should be in the nf file.
If very unlikely, it can stay only in args.

akaviaLab: If this is a situation where (some) parameters should be changed together, I think there are best practices for doing so.

modules/nf-core/bascule/templates/main_script.R
```diff
@@ -0,0 +1,125 @@
...
+  tibble(counts_id=i, samples=rownames(counts_all[[i]]))
+}) %>% bind_rows() %>%
+  group_by(samples) %>%
+  summarise(n_types=n()) %>%
+  filter(n_types == !!n_types) %>%
+  pull(samples) %>% as.character()
+
+counts = lapply(counts_all, function(count_i) count_i[samples_list,])
+
+reference_cat = eval(parse(text=opt[["reference_cat"]]))
```
akaviaLab, line 56: This is one of the options that wasn't given a default in opt.

modules/nf-core/bascule/templates/main_script.R
```diff
@@ -0,0 +1,125 @@
...
+}
+
+x = convert_dn_names(x, cutoff=as.numeric(opt[["cutoff"]]))
+
+# plots
+pl_signatures = plot_signatures(x)
+pl_exposures = plot_exposures(x)
+pl_centroids = plot_centroids(x)
+
+if (clustering) {
```
akaviaLab, line 96: So line 71 to 94 are the python code? Have I understood this correctly?

modules/nf-core/bascule/templates/main_script.R
```diff
@@ -0,0 +1,125 @@
...
+                                 pl_centroids + theme(legend.position="bottom"),
+                                 design="aaaa\nbbbc")
+} else {
+  figure = patchwork::wrap_plots(pl_signatures,
+                                 pl_exposures + theme(legend.position="bottom"),
+                                 ncol=1)
+}
+
+# save objects
+saveRDS(x, file=paste0(opt[["prefix"]], "_fit_bascule.rds"))
```
akaviaLab, line 108: I would suggest putting prefix into the main.nf, with a reasonable default. You can incorporate it by using $, see an example in https://github.com/nf-core/modules/blob/efc7b2793391bb02bef37fb55d52e41b0b8b8839/modules/nf-core/gprofiler2/gost/templates/gprofiler2_gost.R#L137

modules/nf-core/bascule/templates/main_script.R
```diff
@@ -0,0 +1,125 @@
...
+saveRDS(figure, paste0(opt[["prefix"]], "_plots_bascule.rds"))
+ggplot2::ggsave(plot=figure, filename=paste0(opt[["prefix"]], "_plots_bascule.pdf"), height=210, width=210, units="mm")
```
akaviaLab, line 110: I very much like the style of using PACKAGE::FUNCTION. However, I'm a bit nitpicky, and I think that if you use this style, you should use it consistently.

modules/nf-core/bascule/templates/main_script.R
```diff
@@ -0,0 +1,125 @@
...
+# set up reticulate
+## define micromamba binary as main conda binary
+options(reticulate.conda_binary=system("which micromamba", intern=TRUE))
+## define the python binary - not necessary if RETICULATE_PYTHON is defined
+# use_python(system("which python", intern=TRUE), required=TRUE)
+## load pybascule package
+py = import("pybascule")
+
+x = fit(counts=counts,
+        k_list=eval(parse(text=opt[["k_list"]])),
```
akaviaLab, line 74: Using eval is problematic. Can you tuink of an alternative way to do so?

akaviaLab (commented):
In general, the code seems well written. You will have to make sure it passes linting and nf-test, but you probably know that. All the tests are failing due to lack of space on the device, so I think you need to update the resources required in the process section https://www.nextflow.io/docs/latest/reference/process.html#disk

I have one big problem, which I don't fully understand. It looks like R is used to 1) untar the matrices, 2) apply some transformations, and pass it on to pybascule. Later, R is used for 3) visualization.

Running python via reticualte is relatively slow and has overhead. Why not write a simple python script that will untar (via https://docs.python.org/3/library/tarfile.html), transform (via pandas), and then if necessary, do the visualization via ggplot?
You can do the visualization via seaborn (for example), directly in python.

If ggplot is better for visualization, I would write a python script that processes the data, outputs some kind of text version, and then runs (via sys calls) RScript visualize-bascule.R --output $prefix.

Another issue is the fact that you seem to be working on a newer better version of bascule, which Google already points to. I hope you get your paper published soon! Wouldn't it make sense to write the bascule module to use the newer version? Are you planning to modify the Nextflow code to the R version in the future?

modules/nf-core/bascule/templates/main_script.R
akaviaLab: Sorry, I meant this link https://nf-co.re/docs/guidelines/components/modules#optional-command-arguments

elena-buscaroli:
Hi @akaviaLab! Thank you very much for this detailed review and sorry for the late reply.

The bascule package is in R but it uses some models developed in Python through the reticulate package. I've had issues with reticulate as you also suggest, and I've also discussed the errors I've had on the slack channel, coming to the conclusion that the way in which reticulate manages environment and Nextflow are likely impossibile to work together. Therefore, as you suggested, I've been working on using Python scripts to run the model and maybe integrate some R visualizations.

Regarding the new version of bascule you mention, in this module I'm already using this newer version (which is in the main branch of the main repository)! Thanks for pointing this out

I was thinking to update this PR as soon as I have a working module version using Python (taking into account all your other comments as well), let me know if this is ok. Thanks

akaviaLab:
Hi @elena-buscaroli 
In general, it sounds good to me if you're going to create a mostly python version. I'm going to close this PR for now, since it doesn't seem like you're going to continue working directly on it.
I would suggest you open another PR with the pure-python script (or re-open this one). My advice is to spend some time thinking on the organization of the data transferring from one process to another, and save the files as CSV. That would also have the benefit of keeping intermediate files.

I'd be happy to review a new PR, or a re-opened version of this when it is ready, so feel free to tag me then.

elena-buscaroli:
Great, thank you again!

PR #8762, opened 2025-07-08
Init Proseg

derrik-gratz:
## PR checklist

- [x] Add a resource `label`
  - For modules:


modules/nf-core/proseg/proseg_to_baysor/main.nf
```diff
@@ -0,0 +1,50 @@
...
+    proseg-to-baysor \\
+        ${transcript_metadata} \\
+        ${cell_polygons} \\
+        --output-transcript-metadata ${prefix}-baysor-transcript-metadata.csv \\
+        --output-cell-polygons ${prefix}-baysor-cell-polygons.geojson \\
+        ${args}
+    
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        proseg: \$(proseg --version | sed 's/proseg //')
```
Felix-Kummer, line 35: is this sufficient? Will `proseg-to-baysor` always have the same version as `proseg` ?

modules/nf-core/proseg/proseg/main.nf
```diff
@@ -0,0 +1,57 @@
...
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/38/38ebf0dd1e071eb5a99fd220459c09625c1465c5491a3e2dab392bfbe8acb45f/data':
+        'community.wave.seqera.io/library/rust-proseg:2.0.5--dde937bdc1cf4715' }"
+
+    input:
+    tuple val(meta), path(transcripts)
+    val mode
+
+    output: 
+    tuple val(meta), path("transcript-metadata.csv.gz")     , emit: transcript_metadata
```
Felix-Kummer: Does this tool always write files with these specific names or can this be changed e.g. through `args`?

modules/nf-core/proseg/proseg/tests/main.nf.test
```diff
@@ -0,0 +1,66 @@
...
+            def lines = path(process.out.transcript_metadata[0][1]).linesGzip.size()
+            assertAll(
+                { assert process.success },
+                { assert snapshot(
+                    lines,
+                    process.out.versions,
+                    path(process.out.versions[0]).yaml,
+                    path(process.out.transcript_metadata[0][1]).linesGzip[0],
+                    path(process.out.cell_polygons[0][1]).linesGzip[0]
+                ).match() }
```
Felix-Kummer, line 42: any reason to exclude the other output channels here?

modules/nf-core/proseg/proseg_to_baysor/main.nf
derrik-gratz: yes proseg-to-baysor does not have a version command

modules/nf-core/proseg/proseg/main.nf
derrik-gratz: yes, see the changes allowing filenames with prefix and the specific invocation of the filenames in the script (so the user couldn't override it through args)

modules/nf-core/proseg/proseg/tests/main.nf.test
derrik-gratz: laziness, fixed

modules/nf-core/proseg/proseg/main.nf
```diff
@@ -0,0 +1,65 @@
...
+    """
+    mkdir outputs
+    proseg \\
+        ${preset} \\
+        ${args} \\
+        --output-transcript-metadata ${prefix}transcript-metadata.csv.gz \\
+        --output-cell-metadata ${prefix}cell-metadata.csv.gz \\
+        --output-cell-polygons ${prefix}cell-polygons.geojson.gz \\
+        --output-cell-polygon-layers ${prefix}cell-polygons-layers.geojson.gz \\
+        --output-expected-counts ${prefix}expected-counts.csv.gz \\
```
Felix-Kummer: are all of these options required to run the command? Otherwise we should let the user set them through `args` if they need them.

modules/nf-core/proseg/proseg/main.nf
```diff
@@ -0,0 +1,65 @@
...
+    val mode
+
+    output: 
+    tuple val(meta), path("outputs/*transcript-metadata.csv.gz", arity: '1')     , emit: transcript_metadata
+    tuple val(meta), path("outputs/*cell-metadata.csv.gz", arity: '1')           , emit: cell_metadata
+    tuple val(meta), path("outputs/*cell-polygons.geojson.gz", arity: '1')       , emit: cell_polygons
+    tuple val(meta), path("outputs/*cell-polygons-layers.geojson.gz", arity: '1'), emit: cell_polygons_layers
+    tuple val(meta), path("outputs/*expected-counts.csv.gz", arity: '1')         , emit: expected_counts
+    // Proseg has lots of optional file outputs, this is a catch-all for additional files specified in args
+    tuple val(meta), path("outputs/*")                                           , emit: other_outputs, optional: true
```
Felix-Kummer: I count 9 output options in the [proseq Readme](https://github.com/dcjones/proseg/blob/main/README.md). Maybe I'm missing something but that would only be 4 more optional output channels.

modules/nf-core/proseg/proseg/main.nf
```diff
@@ -0,0 +1,65 @@
...
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def preset = mode ? "--${mode}" : ''
+    def args = task.ext.args ?: ''
+    def prefix =  task.ext.prefix ? "outputs/${task.ext.prefix}-" : "outputs/"
+
+    """
+    mkdir outputs
```
famosab: I would argue that its better to have each file simply as a file and not nested in a folder. That makes subsequent processing of single files easier. (Except if you never need a single file but always the whole folder for subsequent tools then just ignore my comment)

derrik-gratz: Would the preferred option then be to create output channels for all possible files, designating some as optional, then using glob file patterns in the output path declaration, then relying on a user who wants the optional files to pass the flags in `args` with names that are compatible with the output pattern?

derrik-gratz: I think I count 11 here, but also with options to change the output format, so 33 possible output types 😵 
image

derrik-gratz: See the changes now. Only one file/flag is specified in the process, as the default naming convention overlaps with another file, making glob capture difficult. I couldn't find a way to use pattern exclusion in the output channel definition, e.g. `!*(union-)cell-polygons.geojson.gz`

famosab: I see its a little complicated but I would argue that its worth the work because otherwise using the outputs downstream is gonna be quite the pain.

modules/nf-core/proseg/proseg/main.nf
```diff
@@ -0,0 +1,68 @@
...
+    script:
+    def preset = mode ? "--${mode}" : ''
+    def args = task.ext.args ?: ''
+    def prefix =  task.ext.prefix ? "--output-path ${task.ext.prefix}" : ""
+
+    """
+    proseg \\
+        ${preset} \\
+        ${args} \\
+        --output-union-cell-polygons cell-polygons-union.geojson.gz \\
```
famosab: Can this output also be modified using prefix?

modules/nf-core/proseg/proseg/main.nf
```diff
@@ -0,0 +1,68 @@
...
+
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def preset = mode ? "--${mode}" : ''
+    def args = task.ext.args ?: ''
+    def prefix =  task.ext.prefix ? "--output-path ${task.ext.prefix}" : ""
+
```
famosab:
```suggestion
    def prefix =  task.ext.prefix ? : "{meta.id}"
```

modules/nf-core/proseg/proseg/main.nf
```diff
@@ -0,0 +1,68 @@
...
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        proseg: \$(proseg --version | sed "s/proseg //")
+    END_VERSIONS
+    """
+
+    stub:
+    """
+    touch cell-metadata.csv.gz
```
famosab:
```suggestion
    echo | gz > ${prefix}cell-metadata.csv.gz
```
You need to do this for all gzipped outputs :) and the names should not be hardcoded but rather take the prefix value :)

modules/nf-core/proseg/proseg/meta.yml
famosab, line 1: Can you add ontologies to this file as well? Running `nf-core modules lint proseg/proseg --fix` should help with that.

modules/nf-core/proseg/proseg/main.nf
famosab: > See the changes now. Only one file/flag is specified in the process, as the default naming convention overlaps with another file, making glob capture difficult. I couldn't find a way to use pattern exclusion in the output channel definition, e.g. !*(union-)cell-polygons.geojson.gz

But this would just mean that the user is not allowed to set union as prefix right? I think that can be captured somewhere!

modules/nf-core/proseg/proseg/meta.yml
derrik-gratz: The csv.gz outputs can be parquet, csv, or csv.gz outputs. Should I put all of them as ontologies? I'm also correcting the output path glob to allow different file extensions.

derrik-gratz: There seems to be a tradeoff between enforcing the `prefix` appended to the start of outputs and losing the ability to specify the file extension, e.g. `--output-cell-metadata ${prefix}cell-metadata.csv.gz` An alternative is to allow the process to take an `output format` as an input variable, then specify the formats for all outputs using that, but that loses out on the ability to granularly control each file individually. Is it sufficient to just not enforce the file-extension prefixes for this process? We could leave it as the default for the one file that needs to be manually specified to avoid collisions with output glob detection, then allow the user to specify it for other files with task.ext if they wanted

famosab: I would say, lets take this to slack and see what the others think. That way we can make a community based decision :) I'll write a quick message.

derrik-gratz:
See discussion [here](https://nfcore.slack.com/archives/CJRH30T6V/p1752736993252959). Per my comment there, I'd prefer to defer the prefix and file extension specification of optional files to task.ext.args to avoid cluttering up the main process.

modules/nf-core/proseg/proseg/main.nf
```diff
@@ -0,0 +1,74 @@
...
+
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def preset = mode ? "--${mode}" : ''
+    def args = task.ext.args ?: ''
+    def prefix =  task.ext.prefix ?: "${meta.id}-"
+
```
famosab, line 39:
```suggestion
```

Felix-Kummer:
I'd be ready to approve this now. @famosab are your points resolved?

famosab (approved):
LGTM :)

derrik-gratz:
thank you!

PR #8763, opened 2025-07-08
Update module/tabix environment.yml

yz533cb:
## PR checklist

Tabix and BGzip binaries are now part of the HTSlib project. And tabix-1.11 in Bioconda doesn't contain anything.
We can keep dependencies: bioconda::htslib=1.21 only (remove the tabix-1.11) to avoid potential errors in setting up local dependencies?

### DIFF
--- a/modules/nf-core/tabix/bgzip/environment.yml
+++ b/modules/nf-core/tabix/bgzip/environment.yml
@@ -6,4 +6,3 @@ channels:
 
 dependencies:
   - bioconda::htslib=1.21
-  - bioconda::tabix=1.11
--- a/modules/nf-core/tabix/bgziptabix/environment.yml
+++ b/modules/nf-core/tabix/bgziptabix/environment.yml
@@ -6,4 +6,3 @@ channels:
 
 dependencies:
   - bioconda::htslib=1.21
-  - bioconda::tabix=1.11
--- a/modules/nf-core/tabix/tabix/environment.yml
+++ b/modules/nf-core/tabix/tabix/environment.yml
@@ -6,4 +6,3 @@ channels:
 
 dependencies:
   - bioconda::htslib=1.21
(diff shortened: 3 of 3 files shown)


mashehu:
How come the snapshots don't need to be updated? Are they not capturing the tabix version?

yz533cb:
> How come the snapshots don't need to be updated? Are they not capturing the tabix version?

Sorry I didn't do local nf-test (thought GitHub CI tests can do the work). 
Today ran a local nf-test on bgzip module and it took ~1 hour.. plan to continue the test tomorrow :)

SPPearce:
> > How come the snapshots don't need to be updated? Are they not capturing the tabix version?
> 
> Sorry I didn't do local nf-test (thought GitHub CI tests can do the work). Today ran a local nf-test on bgzip module and it took ~1 hour.. plan to continue the test tomorrow :)

All the tests are still passing, so the snapshot won't update.
They are capturing the tabix version correctly as far as I can see, it is just it was being claimed from two channels rather than one.

yz533cb:
> > > How come the snapshots don't need to be updated? Are they not capturing the tabix version?
> > 
> > 
> > Sorry I didn't do local nf-test (thought GitHub CI tests can do the work). Today ran a local nf-test on bgzip module and it took ~1 hour.. plan to continue the test tomorrow :)
> 
> All the tests are still passing, so the snapshot won't update. They are capturing the tabix version correctly as far as I can see, it is just it was being claimed from two channels rather than one.

Yes exactly. The test passes and snapshot is not updated.
`bgzip` and `tabix` seem not claimed from the `tabix=1.1` channel previously, because it's empty. So what I did is just removed the unused `tabix=1.1` in the module `environment.yml`.

maxulysse (approved):

PR #8765, opened 2025-07-10
Update blastn module

AnnaNoren:
This pr adds taxid input (to decrease run time) and updates the test+snapshot

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/blast/blastn/meta.yml
+++ b/modules/nf-core/blast/blastn/meta.yml
@@ -35,15 +35,15 @@ input:
         type: directory
         description: Directory containing the blast database
         pattern: "*"
-  - - taxidlist:
+  - taxidlist:
         type: file
         description: File containing NCBI taxon ids, one per line, for filtering the database.
         pattern: "*.{txt}"
-  - - taxids:
+  - taxids:
         type: string
         description: Comma-delimited list of NCBI taxon ids for filtering the database.
         pattern: "^[0-9]+(,[0-9]+)*$"
-  - - negative_tax:
+  - negative_tax:
         type: boolean
         description: Use negative filtering (true) to exclude taxon ids or normal filtering (false).
         pattern: "true or false"


SPPearce (commented):
Can you add a test for the new input?

modules/nf-core/blast/blastn/main.nf
```diff
@@ -23,12 +24,15 @@ process BLAST_BLASTN {
...
     def is_compressed = fasta.getExtension() == "gz" ? true : false
     def fasta_name = is_compressed ? fasta.getBaseName() : fasta
+    def taxidlist_cmd = taxidlist ? "-taxidlist ${taxidlist}" : ""
 
     """
     if [ "${is_compressed}" == "true" ]; then
         gzip -c -d ${fasta} > ${fasta_name}
     fi
 
+    export BLASTDB=${db}
```
muffato: What is this addition for ?

muffato (commented):
There are multiple taxon filters:

- `-taxids` and `-taxidlist`
- `-negative_taxids` and `-negative_taxidlist`

Can we find a way of supporting all four cases ?

modules/nf-core/blast/blastn/main.nf
AnnaNoren: I see now how it is unnecessary in this context.

AnnaNoren:
@muffato @SPPearce 
I've updated the pr to include `-taxidlist -negative_taxidlist -taxids -negative_taxids`. To not overdo the inputs, it relies on `ext.taxid_mode` in nextflow.config.

A problem I've had is that the nf-test relies on the makeblastdb module which does not create the database with a taxid map (vital for taxid filtering to work) and if I were to update that module as well one problem still stands, which is the addition of another two vital files from ncbi (`taxdb.bti` (18MB) and `taxdb.btd` (170MB)) which I don't know how to implement into the database through nf-core makeblastdb module. I've tested this blastn module pr with nf-test and a local database (including ncbi files) and succeeded however.

I'm not sure how to proceed. Do you have any inputs of the possibility of this?

SPPearce:
Please don't add extra `ext.foo` bits, we'd rather just add another input to swap the mode.
Is that the minimum size of the database? Doesn't have to contain everything, just to check the functionality works.
If you can't test it in the CI then it isn't the end of the world, many modules don't have complete tests.

AnnaNoren:
@SPPearce Okay, I see. That would mean four inputs instead of two. So as for the minimum size of the database, created with only one reference genome and taxid mapping and including those vital ncbi files it constitutes 189MB (locally from what I can see). In order to actually run blastn with taxid flags the database needs to be created through taxid mapping and include those files so I don't think there is a way around that. I will update the pr with these latest changes including nf-test with empty taxid inputs.

LilyAnderssonLee:
I’m not sure about including an input channel for each of the arguments: `--taxidlist`,` --negative_taxidlist`, `--taxids`, and `--negative_taxids`, since they are not compatible with each other. In other words, only one of these four arguments can be used at a time.

muffato:
> In other words, only one of these four arguments can be used at a time.

That can be handled with `if` statements that raise errors if incompatible options are passed together, like in the [hifiasm](https://github.com/nf-core/modules/blob/master/modules/nf-core/hifiasm/main.nf#L40-L60) module.

To limit the number of input channels of the pipeline, an idea is to have 3 arguments:

- `taxids` as a list of _values_
- `taxidlist` as a _path_
- `negative_tax` as a boolean

The only thing to check is that `taxid` and `taxidlist` are not passed simultaneously.
Then, the name of the command-line argument is built from the presence/absence of the input channels.

Another idea would be to not have an input channel named `negative_tax` but instead expect the user to add `--negative_tax` to `ext.args` and have the code manipulate `ext.args` to turn `--negative_tax` into `--negative_taxids 1234`  or `--negative_taxidlist /path/to/...`.

muffato:
> Another idea would be to not have an input channel named `negative_tax` but instead expect the user to add `--negative_tax` to `ext.args` and have the code manipulate `ext.args` to turn `--negative_tax` into `--negative_taxids 1234` or `--negative_taxidlist /path/to/...`.

There's an example of that in the vcftools module: https://github.com/nf-core/modules/blob/master/modules/nf-core/vcftools/main.nf#L103-L108

modules/nf-core/blast/blastn/main.nf
```diff
@@ -23,6 +26,12 @@ process BLAST_BLASTN {
     def prefix = task.ext.prefix ?: "${meta.id}"
     def is_compressed = fasta.getExtension() == "gz" ? true : false
     def fasta_name = is_compressed ? fasta.getBaseName() : fasta
+    def negative_tax_cmd = negative_tax ? "negative_" : ""
+    def taxidlist_cmd = taxidlist ? "-${negative_tax_cmd}taxidlist ${taxidlist}" : ""
+    def taxids_cmd = taxids ? "-${negative_tax_cmd}taxids ${taxids}" : ""
+    if (taxidlist_cmd.any() && taxids_cmd.any()) {
+        log.error("ERROR: taxidlist and taxids can not be used at the same time, choose only one argument to use for tax id filtering.")
```
AnnaNoren, line 33: @muffato I added an if statement per your suggestion, similar to in the [hifiasm](https://github.com/nf-core/modules/blob/master/modules/nf-core/hifiasm/main.nf#L40-L60) module. However, It feels somewhat unnecessary because of how blast already outputs quite clear errors when using the two incompatible flags simultaneously. Hence, both the if statement will tell you, and blast errors will tell you. Whats your input, should the if statement stay or is it repetitive?

AnnaNoren:
@muffato I've changed the inputs to include the three proposed ones:
* `taxidlist` as path
* `taxids` as values
* `negative_tax` as value (boolean)
I refrained from the alternative of using `ext.args` because of @SPPearce earlier comment asking to stay clear of that.

modules/nf-core/blast/blastn/main.nf
```diff
@@ -23,6 +26,12 @@ process BLAST_BLASTN {
     def prefix = task.ext.prefix ?: "${meta.id}"
     def is_compressed = fasta.getExtension() == "gz" ? true : false
     def fasta_name = is_compressed ? fasta.getBaseName() : fasta
+    def negative_tax_cmd = negative_tax ? "negative_" : ""
+    def taxidlist_cmd = taxidlist ? "-${negative_tax_cmd}taxidlist ${taxidlist}" : ""
+    def taxids_cmd = taxids ? "-${negative_tax_cmd}taxids ${taxids}" : ""
```
muffato, line 31:
```suggestion
    def taxidlist_cmd = taxidlist ? "-${negative_tax_cmd}taxidlist ${taxidlist}" : "-taxidlist ${taxidlist}"
    def taxids_cmd = taxids ? "-${negative_tax_cmd}taxids ${taxids}" : "-taxids ${taxids}"
```

modules/nf-core/blast/blastn/main.nf
muffato: There's still an advantage in having the `if` here: it gets executed by Nextflow itself when it prepares the job. This means that Nextflow can exit immediately if there's a problem.
Otherwise Nextflow will submit the job, which needs to wait to be scheduled and start, to then immediately fail. And after that Nextflow can find out about the error in `.command.log`.

It's more straightforward to have the test here.

modules/nf-core/blast/blastn/main.nf
AnnaNoren: This change would make it so that if the input `taxidlist` or `taxids` are falsey, that is null or empty, blast will still run **with** the option (e.g., `-taxidlist`) enabled but **without** belonging argument (e.g., `taxidlist.txt`) and yield errors. We want to keep the alternative of not using one or any tax id filtering option when providing empty input.

muffato: 🤦🏼‍♂️ of course you're right ! Ignore my suggestion

AnnaNoren:
@muffato meta.yml is now updated. When it comes to adding a test, the problem stills stands with the usage of the `makeblastdb` module for testing the `blastn` module. To run blastn with taxid filtering, the database must be created with option `-taxid_map` and include two files (taxdb.bt[di], ~190MB) from ncbi (manually added). @SPPearce mentioned in earlier comment: "If you can't test it in the CI then it isn't the end of the world, many modules don't have complete tests."

Just to be clear, I've tested it against a local database with success, but the manual addition of the two pretty large files from ncbi poses challenges in integrating it into the nf-core system.

modules/nf-core/blast/blastn/meta.yml
```diff
@@ -34,6 +35,18 @@ input:
         type: directory
         description: Directory containing the blast database
         pattern: "*"
+  - - taxidlist:
```
muffato:
```suggestion
  - taxidlist:
```

modules/nf-core/blast/blastn/meta.yml
```diff
@@ -34,6 +35,18 @@ input:
...
+        type: file
+        description: File containing NCBI taxon ids, one per line, for filtering the database.
+        pattern: "*.{txt}"
+  - - taxids:
```
muffato:
```suggestion
  - taxids:
```

modules/nf-core/blast/blastn/meta.yml
```diff
@@ -34,6 +35,18 @@ input:
...
+        type: string
+        description: Comma-delimited list of NCBI taxon ids for filtering the database.
+        pattern: "^[0-9]+(,[0-9]+)*$"
+  - - negative_tax:
```
muffato:
```suggestion
  - negative_tax:
```

PR #8770, opened 2025-07-14
Update bclconvert

SPPearce:
Updating bclconvert to add stub and another test.
I'm also explicitly copying the InterOp folder, because when using fusion filesystem the `includeInputs` doesn't work, and so the module will currently crash.


matthdsm (approved):
I approve, but just make sure you don't overwrite any `*.bin` files in the InterOp directory that were generated by bcl-convert. (not sure if that might happen)

SPPearce:
> I approve, but just make sure you don't overwrite any `*.bin` files in the InterOp directory that were generated by bcl-convert. (not sure if that might happen)

I can add `-n` to ensure it crashes in that case?

matthdsm:
That'd be great 👍🏻

SPPearce:
@matthdsm , are you happy for me to merge this?

matthdsm:
Go for it!

PR #8793, opened 2025-07-21
Bumping container versions

mbeavitt:
Closes https://github.com/nf-core/methylseq/issues/550

bwa-mem2 is supported by bwameth, but the current module container does not have the executable installed. The most recent container on biocontainers adds support for this.

### DIFF
--- a/modules/nf-core/bwameth/align/main.nf
+++ b/modules/nf-core/bwameth/align/main.nf
@@ -4,8 +4,8 @@ process BWAMETH_ALIGN {
 
     conda "${moduleDir}/environment.yml"
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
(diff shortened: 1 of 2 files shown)


mashehu (approved):

mashehu:
how come no test caught that?

mbeavitt:
Good question. The pipeline runs with bwa-mem by default, only using bwa-mem2 when explicitly specified or when a bwa-mem2 index is used. I don't believe [methylseq](https://github.com/nf-core/methylseq) has a test suite that uses bwa-mem2, and perhaps the PR adding support for it flew under the radar. Relevant: https://github.com/brentp/bwa-meth/pull/71

To be fair, bioconda only bwa-mem2 to the recipe 2 months ago!

mashehu:
I meant more a module level test here. Could you maybe add one?

mbeavitt:
I think we need to update the bwameth index module with a bwa-mem2 mode first - this will make the testing more straightforward.

PR #8797, opened 2025-07-21
add viber module

### LINKED ISSUE #8712: new module: VIBER

### Further Information

R package: https://caravagnalab.github.io/VIBER/

giorgiagandolfi:
## PR checklist

Closes #8712 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/viber/main.nf
+++ b/modules/nf-core/viber/main.nf
@@ -4,8 +4,8 @@ process VIBER {
 
     conda "${moduleDir}/environment.yml"
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
-        'oras://community.wave.seqera.io/library/r-cnaqc_r-viber:18267cea84b40084':
-        'community.wave.seqera.io/library/r-cnaqc_r-viber:421cfc5931b930b5'}"
+        'oras://community.wave.seqera.io/library/r-cnaqc_r-viber:014077a3164189d5':
+        'community.wave.seqera.io/library/r-cnaqc_r-viber:2314592f7d2f9abe'}"
 
     input:
     tuple val(meta), path(rds_join), val(tumour_samples)
--- a/modules/nf-core/viber/environment.yml
+++ b/modules/nf-core/viber/environment.yml
@@ -3,7 +3,7 @@ channels:
   - bioconda
 
 dependencies:
-  - bioconda::r-cnaqc=1.1.3
+  - bioconda::r-cnaqc=1.1.2 
   - bioconda::r-viber=1.0.0
   - conda-forge::r-cli=3.6.5
   - conda-forge::r-dplyr=1.1.4
--- a/modules/nf-core/viber/tests/main.nf.test.snap
+++ b/modules/nf-core/viber/tests/main.nf.test.snap
@@ -2,14 +2,14 @@
     "viber - csv": {
         "content": [
             [
-                "versions.yml:md5,7a6511ad7ad5dee601605192dc2f8ffc"
+                "versions.yml:md5,d27db9579ed0b91c5131d11f037ec121"
             ]
         ],
         "meta": {
             "nf-test": "0.9.3",
             "nextflow": "25.04.7"
         },
-        "timestamp": "2025-10-07T10:52:11.5139"
+        "timestamp": "2025-10-10T17:31:00.066038"
     },
     "viber - csv - stub": {
         "content": [
@@ -85,10 +85,10 @@
                     ]
                 ],
                 "7": [
-                    "versions.yml:md5,b10fa01ce120532bb26faddb0a855edf"
+                    "versions.yml:md5,eddeda8789b0134c822050d0935c1de1"
                 ],
                 "versions": [
-                    "versions.yml:md5,b10fa01ce120532bb26faddb0a855edf"
+                    "versions.yml:md5,eddeda8789b0134c822050d0935c1de1"
                 ],
                 "viber_heuristic_plots_rds": [
                     [
@@ -166,19 +166,19 @@
             "nf-test": "0.9.3",
             "nextflow": "25.04.7"
         },
-        "timestamp": "2025-10-07T10:53:24.629036"
+        "timestamp": "2025-10-10T17:32:15.602115"
     },
     "viber - rds": {
         "content": [
             [
-                "versions.yml:md5,7a6511ad7ad5dee601605192dc2f8ffc"
+                "versions.yml:md5,d27db9579ed0b91c5131d11f037ec121"
             ]
         ],
         "meta": {
             "nf-test": "0.9.3",
             "nextflow": "25.04.7"
         },
-        "timestamp": "2025-10-07T10:48:26.216394"
+        "timestamp": "2025-10-10T17:27:18.630026"
     },
     "viber - rds - stub": {
         "content": [
@@ -254,10 +254,10 @@
                     ]
                 ],
                 "7": [
-                    "versions.yml:md5,b10fa01ce120532bb26faddb0a855edf"
+                    "versions.yml:md5,eddeda8789b0134c822050d0935c1de1"
                 ],
                 "versions": [
-                    "versions.yml:md5,b10fa01ce120532bb26faddb0a855edf"
+                    "versions.yml:md5,eddeda8789b0134c822050d0935c1de1"
                 ],
                 "viber_heuristic_plots_rds": [
                     [
@@ -335,6 +335,6 @@
             "nf-test": "0.9.3",
             "nextflow": "25.04.7"
         },
-        "timestamp": "2025-10-07T10:49:40.893699"
+        "timestamp": "2025-10-10T17:28:34.670871"
     }
 }


modules/nf-core/viber/templates/main_script.R
```diff
@@ -0,0 +1,175 @@
...
+)
+args_opt = parse_args('$task.ext.args')
+for ( ao in names(args_opt)) opt[[ao]] = args_opt[[ao]]
+print(opt)
+
+# Script #####
+
+library(VIBER)
+library(dplyr)
+library(tidyverse)
```
akaviaLab: Your code calls tidyverse, but your conda library doesn't. If you actually need all of tidyverse, I think you need to add it to the conda environment yml.

modules/nf-core/viber/templates/main_script.R
```diff
@@ -0,0 +1,175 @@
...
+    # args_vals = lapply(args_list, function(x) strsplit(x, split=":")[[1]])
+    args_vals = lapply(args_list, function(x) {
+        x_splt = strsplit(x, split=":")[[1]]
+        c(x_splt[1],  paste(x_splt[2:length(x_splt)], collapse=":"))
+    })
+
+    # Ensure the option vectors are length 2 (key/ value) to catch empty ones
+    args_vals = lapply(args_vals, function(z){ length(z) = 2; z})
+
+    parsed_args = structure(lapply(args_vals, function(x) x[2]), names = lapply(args_vals, function(x) x[1]))
```
akaviaLab, line 18: This line seems an odd bit of code. Can you clarify what it does and why it is needed?

modules/nf-core/viber/main.nf
```diff
@@ -0,0 +1,67 @@
...
+        'oras://community.wave.seqera.io/library/r-cnaqc_r-viber:014077a3164189d5':
+        'community.wave.seqera.io/library/r-cnaqc_r-viber:2314592f7d2f9abe'}"
+
+    input:
+        tuple val(meta), path(rds_join), val(tumour_samples) //rds from either JOIN_CNAQC or JOIN_FIT, should be always grouped
+
+    output:
+        tuple val(meta), path("*_viber_best_st_fit.rds"), emit: viber_rds
+        tuple val(meta), path("*_viber_best_st_heuristic_fit.rds"), emit: viber_heuristic_rds
+        tuple val(meta), path("*_${plot1}"), emit: viber_plots_rds
```
akaviaLab: I find this a bit weird that there are two options for outptut, but neither is actually optional. 
Could you  just use the common pattern as the defintion, like  plot2 would be something like "*heuristic*.rds"

Why do you call them plots when they are rds (not plots)?

Don't plot1 and plo2 overlap with viber_rds and viber_heuristic_rds if you have more than one sample? If these are obligatory and always outputted, just make the other version optional.

modules/nf-core/viber/templates/main_script.R
```diff
@@ -0,0 +1,175 @@
...
+print("$meta.patient")
+print("$tumour_samples")
+print("$rds_join")
+print(samples)
+
+if ( grepl(".rds\$", tolower("$rds_join")) ) {
+    input_obj = readRDS("$rds_join")
+    if (class(input_obj) == "m_cnaqc") {
+        shared = input_obj %>% get_sample(sample=samples, which_obj="shared")
+        joint_table = lapply(names(shared),
```
akaviaLab, line 49: I find the mixture of lapply and dplyr grating, and would prefer using purrr:: and dplyr::, but that's a personal taste.

modules/nf-core/viber/templates/main_script.R
```diff
@@ -0,0 +1,175 @@
...
+
+## Convert the input table into longer format
+reads_data = input_tab %>%
+dplyr::select(chr, from, ref, alt, NV, DP, VAF, sample_id) %>%
+tidyr::pivot_wider(names_from="sample_id",
+                    values_from=c("NV","DP","VAF"), names_sep=".")
+
+## Extract DP (depth)
+dp = reads_data %>%
+# dplyr::filter(mutation_id %in% non_tail) %>% ## this step should be managed before by other module
```
akaviaLab: If these commented lines are not necessary, please just remove them.

modules/nf-core/viber/templates/main_script.R
```diff
@@ -0,0 +1,175 @@
...
+dplyr::select(dplyr::starts_with("DP")) %>%
+dplyr::mutate(dplyr::across(.cols=dplyr::everything(),
```
akaviaLab: Wouldn't line 75 imply there is only one DP column? Then why eveything? Also, since the same code repeats, why not just use [tidyr::replace_na() ](https://tidyr.tidyverse.org/reference/replace_na.html)for all the values you want to replace in one go?

modules/nf-core/viber/templates/main_script.R
```diff
@@ -0,0 +1,175 @@
...
+
+# Save report plot
+n_samples = ncol(best_fit[["x"]]) - 1
+marginals = multivariate = ggplot()
+
+try(expr = {marginals <<- VIBER::plot_1D(best_fit)} )
+#try(expr = {multivariate = plot(best_fit) %>% patchwork::wrap_plots()} )
+#top_p = patchwork::wrap_plots(marginals, multivariate, design=ifelse(n_samples>2, "ABB", "AAB"))
+
+try(expr = {multivariate = plot(best_fit)})
```
akaviaLab, line 136: There's no catch, so is the point only to continue if this fails? Or is the error standard? Do you need encapsulated in try? If not, just have the inside code and let it fail.

modules/nf-core/viber/templates/main_script.R
```diff
@@ -0,0 +1,175 @@
...
+try(expr = {multivariate = ggpubr::ggarrange(plotlist = multivariate)})
+top_p = ggpubr::ggarrange(plotlist = list(marginals, multivariate), widths=ifelse(n_samples>2, c(1,2), c(2,1)))
```
akaviaLab, line 138: Please try running a code fromatter, because these kind of lines are hard to read. RStudio has a good code formatter.

modules/nf-core/viber/templates/main_script.R
```diff
@@ -0,0 +1,175 @@
+#!/usr/bin/env Rscript
```
akaviaLab, line 1: I would prefer that the script be called something more informtive like
viber_main_script.R
or viber_template.R
or run_viber.R

Calling it main_script.R is too generic to be really useful.

modules/nf-core/viber/templates/main_script.R
giorgiagandolfi: Yes they are not really necessary, I will remove them

modules/nf-core/viber/templates/main_script.R
giorgiagandolfi: Ok, originally I did not include it since it is a dependency for other packages, like VIBER and CNAqc, but I can explicitly write it. Thanks

modules/nf-core/viber/templates/main_script.R
giorgiagandolfi: I used these lines of code in order to properly parse the command line arguments that are specified by the `taks.ext.args`. I used this way since other templates in R were written in this way (see [deseq2 module](https://github.com/nf-core/modules/blob/9b696bde5b422831751df1a4f2ba728567ec51fd/modules/nf-core/deseq2/differential/templates/deseq2_differential.R)).
Hope to have been clear.

modules/nf-core/viber/templates/main_script.R
giorgiagandolfi: I think maybe by adding the name of the module to the template it can become a little bit redundant. But if it is strictly necessary I can change it

modules/nf-core/viber/templates/main_script.R
giorgiagandolfi: Thanks for the suggestions. I will go for it

modules/nf-core/viber/templates/main_script.R
giorgiagandolfi: Thanks for the suggestion, but if is not a problem I will stay with lapply and dplyr.

modules/nf-core/viber/templates/main_script.R
akaviaLab: Okay. I honestly think that R templates need a review of this, but that is beyond the scope of this PR.

modules/nf-core/viber/templates/main_script.R
akaviaLab: Okay.

modules/nf-core/viber/templates/main_script.R
giorgiagandolfi: The original input data `input_tab` contains a single `DP` column but just after line 75 we perfom a pivot wider across samples since it is the required input format for `VIBER` (see [documentation](https://caravagnalab.github.io/VIBER/articles/VIBER.html)). We decided to run the same code both for `DP` and `NV` column since VIBER requires two different dataframes of successes (NV) and trials (DP).

akaviaLab: I understand what you did, but I don't think it is good code practice.

How about something like
```
input_tab = joint_table %>%
dplyr::mutate(VAF=replace(VAF, VAF==0, 1e-7))

## Convert the input table into longer format
reads_data = input_tab %>%
dplyr::select(chr, from, ref, alt, NV, DP, VAF, sample_id) %>%
tidyr::replace_na(list(DP=0, NV=0)) %>%
tidyr::pivot_wider(names_from="sample_id",
                    values_from=c("NV","DP","VAF"), names_sep=".")

## Extract DP (depth)
dp = reads_data %>%
dplyr::select(DP) 

## Extract NV (alt_counts)
nv = reads_data %>%
dplyr::select(NV) 
```

Is there a reason you want to have reads_data NOT to have NA replaced?

modules/nf-core/viber/main.nf
giorgiagandolfi: The problem is related to the number of samples in the joint table. If I have more than one sample, I can run the VIBER function `plot(fit)`, otherwise I cannot run such function, since at least two samples are required. Indeed the VIBER function `plot(fit)` is an S3 method plot that runs `plot_2D` (here the [source code](https://github.com/caravagnalab/VIBER/blob/master/R/S3_methods_plot.R)). So for the case of a single sample we need to run a different function. Anyway, I agree that is better to have the same output file name for the plots, so I will change them. 

We prefer to have single plots save in `rds` format since we are planning to write a new subworkflow (in the new version of the pipeline) which requires plots in such format in order to properly assembly them. 

In case multiple samples are present, plot1 and plot2 does not overlap with `viber_rds` and `viber_heuristic_rds`: plot1 is `_viber_best_st_fit_plots.rds` and it is related to the fit in `*viber_rds` while plot2 is `*_viber_best_st_heuristic_fit_plots.rds` and it is related to the fit in `*viber_heuristic_rds`.

modules/nf-core/viber/templates/main_script.R
giorgiagandolfi: Yes, by doing in the way you suggested me, the VIBER function `variational_fit` will drop an errror:
```
Error in var_check_input(x = x, y = y, K = K, data = data, alpha_0 = alpha_0,  : 
  all(x <= y) is not TRUE
```

modules/nf-core/viber/main.nf
```diff
@@ -0,0 +1,52 @@
...
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'oras://community.wave.seqera.io/library/r-cnaqc_r-viber:014077a3164189d5':
+        'community.wave.seqera.io/library/r-cnaqc_r-viber:2314592f7d2f9abe'}"
+
+    input:
+        tuple val(meta), path(rds_join), val(tumour_samples) //rds from either JOIN_CNAQC or JOIN_FIT, should be always grouped
+
+    output:
+        tuple val(meta), path("*_viber_best_st_fit.rds"), emit: viber_rds
```
akaviaLab: I think in general, nf-core prefers Harshill alignment https://nf-co.re/docs/contributing/code_editors_and_styling/harshil_alignment
Some editor plugins might make this happen (the editor plugin for VScode might help, but doesn't include all the rules).

modules/nf-core/viber/templates/main_script.R
akaviaLab: Still not changed, still important.

modules/nf-core/viber/templates/main_script.R
akaviaLab: Please remove unnecessary commented lines from the script.

modules/nf-core/viber/templates/main_script.R
akaviaLab: Okay. If you use tidyr::replace_na() instead of dplyr::mutate(dplyr::across.... (line 75), does it work? Because that will look neater. If even this doesn't work, then just leave it as is.

akaviaLab (changes requested):
Hi.
Thanks for correcting, but there are some corrections still needed. Mostly just some tidying of the R script and renaming it. I still don't understand the try without catch (line 151).
@nf-core/maintainers  - can you help with the alignment please? I'm not sure I got that comment right.

modules/nf-core/viber/templates/main_script.R
SPPearce: `pivot_wider` has an option `values_fill` that lets you specify what to use instead of entries that would be NA.

modules/nf-core/viber/main.nf
```diff
@@ -0,0 +1,52 @@
...
+    tag "$meta.id"
+    label "process_single"
+    label "error_ignore"
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'oras://community.wave.seqera.io/library/r-cnaqc_r-viber:014077a3164189d5':
+        'community.wave.seqera.io/library/r-cnaqc_r-viber:2314592f7d2f9abe'}"
+
+    input:
+        tuple val(meta), path(rds_join), val(tumour_samples) //rds from either JOIN_CNAQC or JOIN_FIT, should be always grouped
```
SPPearce:
```suggestion
    tuple val(meta), path(rds_join), val(tumour_samples) //rds from either JOIN_CNAQC or JOIN_FIT, should be always grouped
```

modules/nf-core/viber/main.nf
```diff
@@ -0,0 +1,52 @@
+process VIBER {
+    tag "$meta.id"
+    label "process_single"
+    label "error_ignore"
+    conda "${moduleDir}/environment.yml"
```
SPPearce, line 5:
```suggestion

    conda "${moduleDir}/environment.yml"
```

modules/nf-core/viber/main.nf
```diff
@@ -0,0 +1,52 @@
...
+
+    output:
+        tuple val(meta), path("*_viber_best_st_fit.rds"), emit: viber_rds
+        tuple val(meta), path("*_viber_best_st_heuristic_fit.rds"), emit: viber_heuristic_rds
+        tuple val(meta), path("*_viber_best_st_fit_plots.rds"), emit: viber_plots_rds
+        tuple val(meta), path("*_viber_best_st_heuristic_fit_plots.rds"), emit: viber_heuristic_plots_rds
+        tuple val(meta), path("*_REPORT_plots_viber.rds"), emit: viber_report_rds
+        tuple val(meta), path("*_REPORT_plots_viber.pdf"), emit: viber_report_pdf
+        tuple val(meta), path("*_REPORT_plots_viber.png"), emit: viber_report_png
+        path "versions.yml", emit: versions
```
SPPearce: Yes, please align all the `, emit`s here. It is a pain to do on github suggestions
```suggestion
    tuple val(meta), path("*_viber_best_st_fit.rds"), emit: viber_rds
    tuple val(meta), path("*_viber_best_st_heuristic_fit.rds"), emit: viber_heuristic_rds
    tuple val(meta), path("*_viber_best_st_fit_plots.rds"), emit: viber_plots_rds
    tuple val(meta), path("*_viber_best_st_heuristic_fit_plots.rds"), emit: viber_heuristic_plots_rds
    tuple val(meta), path("*_REPORT_plots_viber.rds"), emit: viber_report_rds
    tuple val(meta), path("*_REPORT_plots_viber.pdf"), emit: viber_report_pdf
    tuple val(meta), path("*_REPORT_plots_viber.png"), emit: viber_report_png
    path "versions.yml", emit: versions
```

modules/nf-core/viber/main.nf
```diff
@@ -0,0 +1,52 @@
...
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'oras://community.wave.seqera.io/library/r-cnaqc_r-viber:014077a3164189d5':
+        'community.wave.seqera.io/library/r-cnaqc_r-viber:2314592f7d2f9abe'}"
+
+    input:
+        tuple val(meta), path(rds_join), val(tumour_samples) //rds from either JOIN_CNAQC or JOIN_FIT, should be always grouped
+
+    output:
+        tuple val(meta), path("*_viber_best_st_fit.rds"), emit: viber_rds
+        tuple val(meta), path("*_viber_best_st_heuristic_fit.rds"), emit: viber_heuristic_rds
```
SPPearce: I don't think these need "viber" on the channel name.
```suggestion
        tuple val(meta), path("*_viber_best_st_heuristic_fit.rds"), emit: heuristic_rds
```

modules/nf-core/viber/main.nf
```diff
@@ -0,0 +1,52 @@
...
+        tuple val(meta), path("*_viber_best_st_heuristic_fit_plots.rds"), emit: viber_heuristic_plots_rds
+        tuple val(meta), path("*_REPORT_plots_viber.rds"), emit: viber_report_rds
+        tuple val(meta), path("*_REPORT_plots_viber.pdf"), emit: viber_report_pdf
+        tuple val(meta), path("*_REPORT_plots_viber.png"), emit: viber_report_png
+        path "versions.yml", emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+
```
SPPearce:
```suggestion
```

modules/nf-core/viber/main.nf
```diff
@@ -0,0 +1,52 @@
...
+    script:
+
```
SPPearce:
```suggestion
```

modules/nf-core/viber/main.nf
```diff
@@ -0,0 +1,52 @@
...
+    template "viber_main_script.R"
+
+    stub:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+
+
```
SPPearce:
```suggestion
```

modules/nf-core/viber/templates/viber_main_script.R
```diff
@@ -0,0 +1,174 @@
...
+
+mix_p = VIBER::plot_mixing_proportions(best_fit)
+binom = VIBER::plot_peaks(best_fit)
+elbo = VIBER::plot_ELBO(best_fit)
+bottom_p = ggpubr::ggarrange(plotlist = list(mix_p, binom, elbo), widths = c(1,2,1), nrow = 1)
+
+report_fig = ggpubr::ggarrange(top_p, bottom_p, nrow = 2, heights = ifelse(n_samples>2, c(3,1), c(2,1)))
+saveRDS(report_fig, file=paste0(opt[["prefix"]], "_REPORT_plots_viber.rds"))
+ggplot2::ggsave(plot=report_fig, filename=paste0(opt[["prefix"]], "_REPORT_plots_viber.pdf"), height=210, width=210, units="mm", dpi = 200)
+ggplot2::ggsave(plot=report_fig, filename=paste0(opt[["prefix"]], "_REPORT_plots_viber.png"), height=210, width=210, units="mm", dpi = 200)
```
SPPearce: why do these filenames shout `REPORT`?

modules/nf-core/viber/templates/viber_main_script.R
```diff
@@ -0,0 +1,174 @@
...
+  dplyr::rename_all(function(x) stringr::str_remove_all(x,"DP."))
+
+## Extract NV (alt_counts)
+nv = reads_data %>%
+  dplyr::select(dplyr::starts_with("NV")) %>%
+  dplyr::mutate(dplyr::across(.cols=dplyr::everything(),
+                              .fns=function(x) replace(x, is.na(x), 0))) %>%
+  dplyr::rename_all(function(x) stringr::str_remove_all(x,"NV."))
+
+
```
SPPearce:
```suggestion
```

modules/nf-core/viber/templates/viber_main_script.R
```diff
@@ -0,0 +1,174 @@
...
+# Standard fit
+viber_K = as.integer(opt[["K"]])
+
```
SPPearce, line 90:
```suggestion
```

modules/nf-core/viber/templates/viber_main_script.R
```diff
@@ -0,0 +1,174 @@
...
+        } else {
+          cli::cli_alert_warning("Object of class {class(input_obj)} not supported.")
+          return()
+        }
+} else {
+  joint_table = read.csv("$rds_join")
+}
+
+print("Subset joint done")
+
```
SPPearce, line 65:
```suggestion
```

modules/nf-core/viber/templates/viber_main_script.R
```diff
@@ -0,0 +1,174 @@
...
+binom = VIBER::plot_peaks(best_fit)
+elbo = VIBER::plot_ELBO(best_fit)
+bottom_p = ggpubr::ggarrange(plotlist = list(mix_p, binom, elbo), widths = c(1,2,1), nrow = 1)
+
+report_fig = ggpubr::ggarrange(top_p, bottom_p, nrow = 2, heights = ifelse(n_samples>2, c(3,1), c(2,1)))
+saveRDS(report_fig, file=paste0(opt[["prefix"]], "_REPORT_plots_viber.rds"))
+ggplot2::ggsave(plot=report_fig, filename=paste0(opt[["prefix"]], "_REPORT_plots_viber.pdf"), height=210, width=210, units="mm", dpi = 200)
+ggplot2::ggsave(plot=report_fig, filename=paste0(opt[["prefix"]], "_REPORT_plots_viber.png"), height=210, width=210, units="mm", dpi = 200)
+
+
```
SPPearce, line 149:
```suggestion
```

modules/nf-core/viber/tests/nextflow.config
```diff
@@ -0,0 +1,18 @@
...
+        ]
+     }
+}
+
+process {
+    withLabel:error_ignore {
+        maxRetries  = 3
+        errorStrategy = {task.attempt <= 3 ? 'retry' : 'ignore'}
+    }
+}
```
SPPearce: Does it fail quite often?

SPPearce (commented):
Does this module need a seed setting, to ensure that the output is reproducible?

modules/nf-core/viber/main.nf
giorgiagandolfi: Do I need to remove `viber_` suffix to all of the output channel?

modules/nf-core/viber/templates/viber_main_script.R
giorgiagandolfi: This was the original output but I agree it is not required to highlight it. I will change it. Thanks

modules/nf-core/viber/main.nf
```diff
@@ -0,0 +1,50 @@
+process VIBER {
+    tag "$meta.id"
+    label "process_single"
+    //label "error_ignore"
```
SPPearce: Can we remove this rather than leave it commented in.

giorgiagandolfi: yes sure.

giorgiagandolfi:
The module is probabilistic, but it is not necessary to set a specific seed. The default seed of the r session is enough to ensure reproducibility.

SPPearce:
> The module is probabilistic, but it is not necessary to set a specific seed. The default seed of the r session is enough to ensure reproducibility.

I don't understand this - either it is deterministic or it requires the seed setting? The default seed will change every time.

giorgiagandolfi:
I noticed that my setting different seed with `set.seed()` it returns the same output regarding mutations to cluster assignment. What changes is the number of iterations to reach the ELBO. So, since the `seed` is not a parameter of the function, I think that we can avoid setting a seed.

modules/nf-core/viber/templates/viber_main_script.R
```diff
@@ -0,0 +1,167 @@
+#!/usr/bin/env Rscript
+
+parse_args = function(x) {
+    x = gsub("\\\\[","",x)
+    x = gsub("\\\\]","",x)
+    # giving errors when we have lists like c(xxx, xxx) since it will separate it
+    # args_list = unlist(strsplit(x, ', ')[[1]])
```
akaviaLab, line 7: These two commented code examples seem to indicate what didn't work. It would be a lot tidier to have one comment that says something like
"Can't use unlist(strsplit(x, ',')) because it will give an error when we have lists like c(xxx, xxx)" and please remove the commented code for args_list and args_val that shouldn't be used.

Issue #8798, opened 2025-07-22
[FEATURE] Add `ext.args2` directive to `cellranger` module


fmalmeida:
### Is your feature request related to a problem? Please describe

It was flagged by a user [here](https://github.com/nf-core/scrnaseq/issues/475) that the cellranger modules, when used in the `scrnaseq` pipeline, are not straightforward to customise with additional tool parameters since the pipeline itself uses the `ext.args` to setup some basic parameters handling.

### Describe the solution you'd like

Thus, it would be of interest to add `ext.args2` to the modules, so users can append extra parameters in a more straightforward manner.

fmalmeida:
Cancelling in favour of updating pipeline documentation.

PR #8799, opened 2025-07-22
add ext.args2 directive to cellranger modules

fmalmeida:
dding `ext.args2` to `cellranger` modules, to allow that `ext.args` are used in pipelines, like in `nf-core/scrnaseq` while allowing straightforward customisation for users.

Relates to:

* https://github.com/nf-core/modules/issues/8798
* https://github.com/nf-core/scrnaseq/issues/475

# Question for maintainers

Is this conceptually fine to perform, or should we find a different way of doing it in the pipeline itself?

### DIFF
--- a/modules/nf-core/cellranger/count/main.nf
+++ b/modules/nf-core/cellranger/count/main.nf
@@ -21,6 +21,7 @@ process CELLRANGER_COUNT {
         error "CELLRANGER_COUNT module does not support Conda. Please use Docker / Singularity / Podman instead."
     }
     args = task.ext.args ?: ''
+    args2 = task.ext.args2 ?: ''
     prefix = task.ext.prefix ?: "${meta.id}"
     template "cellranger_count.py"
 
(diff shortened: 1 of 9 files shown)


grst (changes requested):
Why would we use `args2` if it can already be passed via `args`? 

I thought `args2` is only used if multiple tools are used within a single module.

fmalmeida:
Hi @grst ,
The question is really on point and part of the reason behind my question in the first "PR comment" [1](https://github.com/nf-core/modules/pull/8799#issue-3251767041).

Essentially, it is because in the `scrnaseq` pipeline, we make use of the `ext.args` to define some parameters using the `modules.config`. The idea came from https://github.com/nf-core/scrnaseq/issues/475 where to pass an additional parameter, the user had to make sure to copy the pre-defined `ext.args`.

I thought the `ext.args2` could be used like that to just extend it. But, if this is not the intended way to use it, we can totally close this PR and issue, and document there in the mother ticket, that ensuring all is copied is indeed the way to do it.

:)

fmalmeida:
Cancelling in favour of updating pipeline documentation.

Issue #8800, opened 2025-07-22
new module: biomformat/convert


jcbioinformatics:
### Is there an existing module for this?

- [x] I have searched for the existing module

### Is there an open PR for this?

- [x] I have searched for existing PRs

### Is there an open issue for this?

- [x] I have searched for existing issues

### Further Information

[biom-format's convert tool](http://biom-format.org/documentation/biom_conversion.html) facilitates converting files in biom format, e.g. json biom to hdf5 biom. 

### Are you going to work on this?

- [ ] If I'm planning to work on this module, I added myself to the `Assignees` to facilitate tracking who is working on the module

Issue #8236, opened 2025-04-08
[FEATURE] nf-test assertions - use md5 sums for readLines


famosab:
### Is your feature request related to a problem? Please describe

Sometimes parts of output files have timestamps or similar stuff in them which makes the total md5sum of the file itself instable. Thus we usally go for asserting the lines with readLines() or even just the name. 

### Describe the solution you'd like

I suggest using
```
file.readLines()[2..-1].join('\n').md5()
```
to create md5sums for the stable parts. This makes the snapshots smaller :)

vagkaratzas:
Hmm is this getting adopted? For small lines like `cafe` I'd prefer to see the lines themselves to be fair.

famosab:
Its a possibility if you just need to exclude one line (which produces a timestamp for example) from a bigger output. Its not needed to do this when you only want to see one line.

PR #8801, opened 2025-07-22
Add biomformat/convert module

jcbioinformatics:
## PR checklist

Closes [biomformat/convert#8800](https://github.com/nf-core/modules/issues/8800) 

- [x] Add a resource `label`
  - For modules:

### DIFF
--- a/modules/nf-core/biomformat/convert/main.nf
+++ b/modules/nf-core/biomformat/convert/main.nf
@@ -1,0 +1,49 @@
+process BIOMFORMAT_CONVERT {
+    tag "$meta.id"
+    label 'process_low'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/biom-format:2.1.15':
+        'biocontainers/biom-format:2.1.15' }"
+
+    input:
+    tuple val(meta), path(biom)
+
+    output:
+    tuple val(meta), path("*.biom")        , optional: true, emit: biom
+    tuple val(meta), path("*.txt")         , optional: true, emit: txt
+    path "versions.yml"                    , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def output = task.ext.args.contains("--to-tsv") ? "-o ${prefix}.txt" : "-o ${prefix}.biom" 
+    """
+    biom convert \\
+        -i $biom \\
+        $output \\
+        $args 
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        biom-format: \$(biom --version | cut -f 3 -d ' ')
+    END_VERSIONS
+    """
+
+    stub:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def extension = task.ext.args.contains("--to-tsv") ? ".txt" : ".biom" 
+    """
+    touch $output
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        biom-format: \$(biom --version | cut -f 3 -d ' ')
+    END_VERSIONS
+    """
+}
--- a/modules/nf-core/biomformat/convert/meta.yml
+++ b/modules/nf-core/biomformat/convert/meta.yml
@@ -1,0 +1,62 @@
+name: "biomformat_convert"
+description: Convert biom table to different format
+keywords:
+  - biom
+  - feature table
+  - formatting
+tools:
+  - "biomformat":
+      description: "Biological Observation Matrix (BIOM) format"
+      homepage: "http://biom-format.org/index.html"
+      documentation: "http://biom-format.org/documentation/index.html"
+      tool_dev_url: "https://github.com/biocore/biom-format"
+      doi: "10.1186/2047-217X-1-7"
+      licence: ['BSD License']
+      identifier: biotools:biomformat
+
+input:
+  - - meta:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. `[ id:'sample1', single_end:false ]`
+    - biom:
+        type: file
+        description: Biom formatted feature table file
+        pattern: "*.{biom,txt,tsv}"
+output:
+  biom:
+    - - meta:
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. [ id:'test' ]
+      - "*.biom":
+          type: file
+          description: Converted biom file
+          pattern: "*.biom"
+          ontologies:
+            - edam: http://edamontology.org/format_3746 # BIOM format
+  txt:
+    - - meta:
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. [ id:'test' ]
+      - "*.txt":
+          type: file
+          description: Converted biom file
+          pattern: "*.txt"
+          ontologies:
+            - edam: http://edamontology.org/format_3746 # BIOM format
+  versions:
+  - "versions.yml":
+      type: file
+      description: File containing software versions
+      pattern: "versions.yml"
+      ontologies:
+        - edam: http://edamontology.org/format_3750 # YAML
+authors:
+  - "@jcbioinformatics"
+maintainers:
+  - "@jcbioinformatics"
--- a/modules/nf-core/biomformat/convert/tests/main.nf.test
+++ b/modules/nf-core/biomformat/convert/tests/main.nf.test
@@ -1,0 +1,113 @@
+nextflow_process {
+
+    name "Test Process BIOMFORMAT_CONVERT"
+    script "../main.nf"
+    process "BIOMFORMAT_CONVERT"
+    config "./nextflow.config"
+
+    tag "modules"
+    tag "modules_nfcore"
+    tag "vsearch/cluster"
+    tag "gunzip"
+    tag "biomformat"
+    tag "biomformat/convert"
+
+    test("convert - to text") {
+
+        config "./to_text.config"
+
+        setup {
+            run ("VSEARCH_CLUSTER") {
+
+                script "../../../vsearch/cluster/main.nf"
+                process {
+                    """
+                    input[0] = [
+                        [ id:'test', single_end:false ], // meta map
+                        file(params.modules_testdata_base_path + 'genomics/sarscov2/illumina/fastq/test_1.fastq.gz', checkIfExists: true)
+                    ]
+                    """
+                }
+            }
+
+            run ("GUNZIP") {
+
+                script "../../../gunzip/main.nf"
+                process {
+                    """
+                    input[0] = VSEARCH_CLUSTER.out.biom
+                    """
+                }
+            }
+
+        }
+
+        when {
+            process {
+                """
+                input[0] = GUNZIP.out.gunzip
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
+            )
+        }
+
+    }
+
+    test("convert - to json biom") {
+
+        config "./to_json_biom.config"
+
+        setup {
+            run ("VSEARCH_CLUSTER") {
+
+                script "../../../vsearch/cluster/main.nf"
+                process {
+                    """
+                    input[0] = [
+                        [ id:'test', single_end:false ], // meta map
+                        file(params.modules_testdata_base_path + 'genomics/sarscov2/illumina/fastq/test_1.fastq.gz', checkIfExists: true)
+                    ]
+                    """
+                }
+            }
+
+            run ("GUNZIP") {
+
+                script "../../../gunzip/main.nf"
+                process {
+                    """
+                    input[0] = VSEARCH_CLUSTER.out.otu
+                    """
+                }
+            }
+
+        }
+
+        when {
+            process {
+                """
+                input[0] = GUNZIP.out.gunzip
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(
+                    path(process.out.biom[0][1]).readLines().last().contains('Biological Observation Matrix'),
+                    process.out.versions
+                    ).match() 
+                }
+            )
+        }
+
+    }
+
+}
--- a/modules/nf-core/biomformat/convert/tests/nextflow.config
+++ b/modules/nf-core/biomformat/convert/tests/nextflow.config
@@ -1,0 +1,9 @@
+process {
+
+    withName: VSEARCH_CLUSTER {
+        ext.args = '--id 0.8'
+        ext.args2 = "--cluster_fast"
+        ext.args3 = "--centroids"
+    }
+
+}
--- a/modules/nf-core/biomformat/convert/environment.yml
+++ b/modules/nf-core/biomformat/convert/environment.yml
@@ -1,0 +1,7 @@
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/environment-schema.json
+channels:
+  - conda-forge
+  - bioconda
+dependencies:
+  - "conda-forge::biom-format=2.1.15"
--- a/modules/nf-core/biomformat/convert/tests/main.nf.test.snap
+++ b/modules/nf-core/biomformat/convert/tests/main.nf.test.snap
@@ -1,0 +1,56 @@
+{
+    "convert - to text": {
+        "content": [
+            {
+                "0": [
+                    
+                ],
+                "1": [
+                    [
+                        {
+                            "id": "test",
+                            "single_end": false
+                        },
+                        "test.txt:md5,c8d78c31c4874f2d9c59c4bf16e68665"
+                    ]
+                ],
+                "2": [
+                    "versions.yml:md5,4002d31c7b16830c8606b63af75e1870"
+                ],
+                "biom": [
+                    
+                ],
+                "txt": [
+                    [
+                        {
+                            "id": "test",
+                            "single_end": false
+                        },
+                        "test.txt:md5,c8d78c31c4874f2d9c59c4bf16e68665"
+                    ]
+                ],
+                "versions": [
+                    "versions.yml:md5,4002d31c7b16830c8606b63af75e1870"
+                ]
+            }
+        ],
+        "meta": {
+            "nf-test": "0.9.2",
+            "nextflow": "24.10.4"
+        },
+        "timestamp": "2025-07-22T08:35:21.975747196"
+    },
+    "convert - to json biom": {
+        "content": [
+            true,
+            [
+                "versions.yml:md5,4002d31c7b16830c8606b63af75e1870"
+            ]
+        ],
+        "meta": {
+            "nf-test": "0.9.2",
+            "nextflow": "24.10.4"
+        },
+        "timestamp": "2025-07-22T08:35:39.162772345"
+    }
+}
--- a/modules/nf-core/biomformat/convert/tests/to_json_biom.config
+++ b/modules/nf-core/biomformat/convert/tests/to_json_biom.config
@@ -1,0 +1,13 @@
+process {
+
+    withName: VSEARCH_CLUSTER {
+        ext.args = '--id 0.8'
+        ext.args2 = "--cluster_fast"
+        ext.args3 = "--otutabout"
+    }
+
+    withName: BIOMFORMAT_CONVERT {
+        ext.args = '--table-type="OTU table" --to-json'
+    }
+
+}
--- a/modules/nf-core/biomformat/convert/tests/to_text.config
+++ b/modules/nf-core/biomformat/convert/tests/to_text.config
@@ -1,0 +1,13 @@
+process {
+
+    withName: VSEARCH_CLUSTER {
+        ext.args = '--id 0.8'
+        ext.args2 = "--cluster_fast"
+        ext.args3 = "--biomout"
+    }
+
+    withName: BIOMFORMAT_CONVERT {
+        ext.args = "--to-tsv"
+    }
+
+}


famosab:
You can ask in the nf-core slack in #github-invitations to be added to organization so that the CI tests are being run for you :)

modules/nf-core/biomformat/convert/meta.yml
```diff
@@ -0,0 +1,62 @@
...
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. [ id:'test' ]
+      - "*.txt":
+          type: file
+          description: Converted biom file
+          pattern: "*.txt"
+          ontologies:
+            - edam: http://edamontology.org/format_3746 # BIOM format
```
famosab: Should this not be txt then? :)

modules/nf-core/biomformat/convert/tests/main.nf.test
```diff
@@ -0,0 +1,113 @@
...
+
+                script "../../../gunzip/main.nf"
+                process {
+                    """
+                    input[0] = VSEARCH_CLUSTER.out.biom
+                    """
+                }
+            }
+
+        }
```
famosab: You can move this whole setup block before the test block and reuse it later to avoid code-duplication :)

modules/nf-core/biomformat/convert/tests/nextflow.config
famosab, line 1: Can you change these files to adhere to the guidelines (https://nf-co.re/docs/guidelines/components/modules) - section 8.9 is related to configuration :) You should end up with only one nextflow.config file

modules/nf-core/biomformat/convert/tests/main.nf.test
```diff
@@ -0,0 +1,113 @@
...
+                input[0] = GUNZIP.out.gunzip
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(
+                    path(process.out.biom[0][1]).readLines().last().contains('Biological Observation Matrix'),
```
famosab: Is the file that unstable that you can only assert the last line? :)

modules/nf-core/biomformat/convert/main.nf
```diff
@@ -0,0 +1,49 @@
...
+    tuple val(meta), path("*.biom")        , optional: true, emit: biom
+    tuple val(meta), path("*.txt")         , optional: true, emit: txt
+    path "versions.yml"                    , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
```
famosab, line 23: Maybe we need to capture here somewhere that input & output files cannot have the same name.

modules/nf-core/biomformat/convert/main.nf
```diff
@@ -0,0 +1,49 @@
+process BIOMFORMAT_CONVERT {
+    tag "$meta.id"
+    label 'process_low'
```
famosab: Does your tool allow multi-threading? Then add that otherwise I would change to process_single

famosab (changes requested):
Looks good already just a few minor tweaks :)

jcbioinformatics:
> Looks good already just a few minor tweaks :)

Thanks!

modules/nf-core/biomformat/convert/main.nf
jcbioinformatics: I doublechecked their documentation and the help information, and as best I can tell, it does not. Changed to process_single now, thanks!

modules/nf-core/biomformat/convert/tests/main.nf.test
jcbioinformatics: Yes, I tried running it twice myself just to check, and the biom output yielded different md5sums.

So, I based this check on https://github.com/nf-core/modules/blob/80db11d5d10e7319937c674afabc563bb14e3d93/modules/nf-core/metaphlan/metaphlan/tests/main.nf.test#L47

modules/nf-core/biomformat/convert/tests/main.nf.test
jcbioinformatics: I think I'd still need both sets of VSEARCH and GUNZIP modules even with moving it to global.

I tried running the VSEARCH_CLUSTER module to output both txt and biom formats at the same time, but it didn't work for me. 

Looking at the code for that, the way it handles the output extension seems to make it so that you can only specify one type of output at a time

https://github.com/nf-core/modules/blob/8b6514c107bac9520a69aac3471f33e2ab07f269/modules/nf-core/vsearch/cluster/main.nf#L40-L52

jcbioinformatics:
> You can ask in the nf-core slack in #github-invitations to be added to organization so that the CI tests are being run for you :)

Ah, that's what I was missing, oops. I just did that now, thanks!

modules/nf-core/biomformat/convert/tests/main.nf.test
famosab: Can you maybe output the file once completely and then check whether its only for example the first line that differs? Then you could go for something like #8236

modules/nf-core/biomformat/convert/tests/main.nf.test
famosab: You can do aliases for the setup block if you need a specific process multiple times but its ok if we leave it like this :)

jcbioinformatics: I realized I needed aliases to facilitate switching to just using nextflow.config, so I ended up making all four setup modules global. That's in commit e82c1f9fb90df03f4fa9e3b717533947d51f866a

You can probably tell I'm a bit new to using GitHub for anything other than my own projects, so I forgot to check that I actually had my username set appropriately prior to pushing, oops. That should be fixed moving forward (I switched from local to HPC to run the tests faster)

modules/nf-core/biomformat/convert/tests/main.nf.test
jcbioinformatics: Huh, I never thought of creating a md5sum like that, that's really cool!

Unfortunately, the biom output is just one really long line

image

I'll try to remember that suggestion for future modules though :)

modules/nf-core/biomformat/convert/tests/nextflow.config
jcbioinformatics: Thanks!

Heh, I didn't do a great job there the first time around, but now I know. Speaking of, I realized I probably should have been commenting with specific commits before marking comments as resolved. 

Sorry, the commit addressing this comment is db8df9cac5331d17b1256200e7f3c4cf5c28b02d

modules/nf-core/biomformat/convert/main.nf
jcbioinformatics: Ooh, yes, good thinking!

I took a stab at that, pretty simple, just checking that the input biom channel and derived output variable are not equal (moved "-o" flag outside of variable).

See commit 3843ff940a32293f08a5316d1fece7e6d687f37b

I'm still not 100% certain I got the CLI setup alright on my cluster account, but at least I got my username to matchup alright, baby steps :)

Anyway, what do you think? This commit should do the trick, right?

modules/nf-core/biomformat/convert/tests/main.nf.test
jcbioinformatics: Sorry, I realized you can't actually tell that it's just one line from that screenshot, oops.

image

That's what I get for doing this close to bed

famosab: The output looks like a json file to me 🤔 maybe it can be read as a json somehow?

modules/nf-core/biomformat/convert/main.nf
famosab: maybe we can do it like here: https://github.com/nf-core/modules/blob/c5c5a32c091113b9eccf11390f0091439ba85baa/modules/nf-core/lima/main.nf#L34

jcbioinformatics: Thanks! I switched the check to match what was in the lima module's code.

See commit b7f6097f54ef60e7292249ac5ed183cbde3425d4

I didn't see a test for that logic in their main.nf.test file, so I left mine the same for now, still just checking that the process fails

modules/nf-core/biomformat/convert/tests/main.nf.test
jcbioinformatics: You're right, and I think I found the issue.

It turns out it has a time stamp.

image

Hopefully, that's the only part that's changing between runs, and if so, I think I can ignore the "date" key (hopefully just by using JsonSlurper), and then do something similar to your initial suggestion, with converting the matching text to a md5sum.

I'm not sure if I'll have time to tackle that this morning, or what the weekend will hold, but hopefully within the next few days, I'll be able to come up with something that works there, and then I'll re-request review once that's done.

Thanks so much for all your help!

famosab: Thats perfect! Happy to help :)

famosab: Maybe something like that:
```
path(process.out.biom[0][1]).json
```
and then we try to exclude the date key somehow

famosab:
```
                    process.out.biom.collect{[
                        file(it[1]).name,
                        path(it[1]).json[0].findAll{it.key != "date"}.toString().md5()
                    ]},
```

modules/nf-core/biomformat/convert/tests/main.nf.test
```diff
@@ -0,0 +1,136 @@
...
+                input[0] = GUNZIP_TXTOUT.out.gunzip
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(
+                    path(process.out.biom[0][1]).readLines().last().contains('Biological Observation Matrix'),
```
famosab:
```suggestion
                    path(process.out.biom[0][1]).json.findAll{it.key != "date"}.toString()
```

modules/nf-core/biomformat/convert/tests/main.nf.test
```diff
@@ -0,0 +1,136 @@
...
+                input[0] = GUNZIP_TXTOUT.out.gunzip
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(
+                    path(process.out.biom[0][1]).json.findAll{it.key != "date"}.toString()
```
famosab:
```suggestion
                    path(process.out.biom[0][1]).json.findAll{it.key != "date"}.toString(),
```

modules/nf-core/biomformat/convert/tests/main.nf.test
```diff
@@ -0,0 +1,136 @@
...
+                input[0] = GUNZIP_TXTOUT.out.gunzip
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(
+                    path(process.out.biom[0][1]).json.findAll{it.key != "date"}.toString(),
```
famosab:
```suggestion
                    path(process.out.biom[0][1]).json.findAll{it.key != "date"}.toString().md5(),
```

modules/nf-core/biomformat/convert/tests/nextflow.config
```diff
@@ -0,0 +1,19 @@
...
+        ext.args = '--id 0.8'
+        ext.args2 = "--cluster_fast"
+        ext.args3 = "--biomout"
+    }
+
+    withName: VSEARCH_CLUSTER_TXTOUT {
+        ext.args = '--id 0.8'
+        ext.args2 = "--cluster_fast"
+        ext.args3 = "--otutabout"
+    }
```
famosab: Can you move these to the nf.test file as well? :)

ext.args = params.biomout_args
ext.args2 = params.biomout_args2

etc.

famosab:
(Sorry I got too exited about the json thingy and had to try it out)

jcbioinformatics:
No problem, thanks!

Your method of parsing the json is a lot cleaner than what I was planning on trying. 

I think the config change should be pretty straightforward, so I'll take a stab at that either later today or over the weekend. So, fingers-crossed, I'll be ready for you to look everything over by Monday

modules/nf-core/biomformat/convert/tests/nextflow.config
jcbioinformatics: Could you point me at a module with a test that is structured with global setup modules and arguments specified with params for them?

I tried the below with and without wrapping it in "when" inside the "run" block, and also without wrapping it in "when" just within the setup block.

            params {
                txtout_args  = '--id 0.8'
                txtout_args2 = '--cluster_fast'
                txtout_args3 = '--otutabout'
            }  

All those attempts gave the same error

```ERROR ~ Unknown config attribute `process.withName:VSEARCH_CLUSTER_TXTOUT.params.txtout_args` -- check config file: /net/clusterhn.cluster.com/home/see/Wright_Labs/GitHub/modules/modules/nf-core/biomformat/convert/tests/./nextflow.config```

The config for that process looks like this

```
    withName: VSEARCH_CLUSTER_TXTOUT {
        ext.args  = params.txtout_args
        ext.args2 = params.txtout_args2
        ext.args3 = params.txtout_args3
    }
```

I think it would work if I moved the setup modules back to their respective tests and specified the params in their when blocks. But, that would mean duplicating the setup for the json test in the test for the module failing when input matches output.

I feel like I'm just missing something very obvious, but I'm kinda stuck on things to try right now

**Comment by jcbioinformatics** [comment]
Well, the downside is my hubris got the better of me, and my attempts to make the additional config changes haven't gone great.

But, on the brighter side, we now have stub tests :)

File: modules/nf-core/biomformat/convert/main.nf [hunk 10]
```diff
@@ -0,0 +1,51 @@
...
+        'https://depot.galaxyproject.org/singularity/biom-format:2.1.15':
+        'biocontainers/biom-format:2.1.15' }"
+
+    input:
+    tuple val(meta), path(biom)
+
+    output:
+    tuple val(meta), path("*.biom")        , optional: true, emit: biom
+    tuple val(meta), path("*.txt")         , optional: true, emit: txt
+    path "versions.yml"                    , emit: versions
```
famosab:
```suggestion
    tuple val(meta), path("*.biom"), optional: true, emit: biom
    tuple val(meta), path("*.txt") , optional: true, emit: txt
    path "versions.yml"                            , emit: versions
```

modules/nf-core/biomformat/convert/tests/nextflow.config
famosab: You can either do it like this:
https://github.com/nf-core/modules/blob/6f835e2153334430b0519fd5bb53d146b26cf1a4/modules/nf-core/muscle/tests/main.nf.test#L53

or like this:
https://github.com/nf-core/modules/blob/6f835e2153334430b0519fd5bb53d146b26cf1a4/modules/nf-core/syri/tests/main.nf.test#L13

(I prefer the second one).

jcbioinformatics: Thanks!

I'm trying to match the second one, but I'm having trouble figuring out where I'm going wrong.

Changing the setup for `VSEARCH_CLUSTER_BIOMOUT` to the below

```
        run ("VSEARCH_CLUSTER", alias: "VSEARCH_CLUSTER_BIOMOUT") {
            script "../../../vsearch/cluster/main.nf"

            params {
                biomout_args  = '--id 0.8'
                biomout_args2 = "--cluster_fast"
                biomout_args3 = "--biomout"
            }

            process {
                """
                input[0] = [
                    file(params.modules_testdata_base_path + 'genomics/sarscov2/illumina/fastq/test_1.fastq.gz', checkIfExists: true)
                ]
                """
            }
        }
```

With the config being

```
    withName: VSEARCH_CLUSTER_BIOMOUT {
        ext.args  = params.biomout_args
        ext.args2 = params.biomout_args2
        ext.args3 = params.biomout_args3
    }
```

Yields
```
ERROR ~ Unknown config attribute `process.withName:VSEARCH_CLUSTER_BIOMOUT.params.biomout_args` -- check config file: /net/clusterhn.cluster.com/home/see/Wright_Labs/GitHub/modules/modules/nf-core/biomformat/convert/tests/./nextflow.config
```

I was able to get it to work with a variant of the first option. I needed to add all params to all modules, but it worked after that (commit 30dd47a0d46d5abe088306414ae61daa6dbf1b20). I feel like that shouldn't be necessary though. 

Am I just missing something, or is this kinda a weird edge-case due to having multiple setup modules with different params?

famosab: yeah the params have to be added to every run unfortunately - but you can also add them and leave them empty with
biomout_args = '' 
so you solved it correctly!

modules/nf-core/biomformat/convert/tests/main.nf.test
```diff
@@ -0,0 +1,217 @@
...
+    test("convert - to text - stub") {
+
+        options "-stub"
+
+        when {
+            params {
+                module_args = '--to-tsv'
+                biomout_args  = '--id 0.8'
+                biomout_args2 = "--cluster_fast"
+                biomout_args3 = "--biomout"
```
famosab:
```suggestion
                biomout_args  = ""
                biomout_args2 = ""
                biomout_args3 = ""
```

modules/nf-core/biomformat/convert/tests/main.nf.test
```diff
@@ -0,0 +1,217 @@
...
+
+        when {
+            params {
+                module_args = '--table-type="OTU table" --to-json'
+                biomout_args  = '--id 0.8'
+                biomout_args2 = "--cluster_fast"
+                biomout_args3 = "--biomout"
+                txtout_args  = '--id 0.8'
+                txtout_args2 = "--cluster_fast"
+                txtout_args3 = "--otutabout"
```
famosab:
```suggestion
                txtout_args  = ""
                txtout_args2 = ""
                txtout_args3 = ""
```

modules/nf-core/biomformat/convert/tests/main.nf.test
```diff
@@ -0,0 +1,217 @@
...
+
+        when {
+            params {
+                module_args = '-table-type="OTU table" --to-hdf5'
+                biomout_args  = '--id 0.8'
+                biomout_args2 = "--cluster_fast"
+                biomout_args3 = "--biomout"
+                txtout_args  = '--id 0.8'
+                txtout_args2 = "--cluster_fast"
+                txtout_args3 = "--otutabout"
```
famosab:
```suggestion
                txtout_args  = ""
                txtout_args2 = ""
                txtout_args3 = ""
```

modules/nf-core/biomformat/convert/tests/main.nf.test
```diff
@@ -0,0 +1,217 @@
...
+            }
+
+            process {
+                """
+                input[0] = GUNZIP_TXTOUT.out.gunzip
+                """
+            }
+        }
```
famosab: Test says to json biom but input is _TXTOUT?

modules/nf-core/biomformat/convert/tests/main.nf.test
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    test("convert - to text") {
+
+        when {
+            params {
+                module_args = '--to-tsv'
+                biomout_args  = '--id 0.8'
+                biomout_args2 = "--cluster_fast"
+                biomout_args3 = "--biomout"
```
famosab:
```suggestion
                biomout_args  = ""
                biomout_args2 = ""
                biomout_args3 = ""
```

modules/nf-core/biomformat/convert/tests/main.nf.test
famosab, line 1: Lets try to be constistent with either "" or '' for params

famosab (commented):
A few comments that might help resolving the tests that do not run.

My guess is because you use the aliases that we have to add the args always even if they are not accessed (weird edge case).

modules/nf-core/biomformat/convert/tests/main.nf.test
jcbioinformatics: Ah, yeah, I see why that's not really intuitive now.

When I made the aliases and the params, I matched it to what the required setup was doing, not what the module being tested outputs.

For the json biom test then, it needs txt tab delimited input to confirm it can appropriately convert that to biom format. I thought it made sense to have the naming reflect the fact that the required VSEARCH setup module produces a txt file.

With that being said, I'm definitely open to changing the naming. Maybe instead of "TXTOUT" and "BIOMOUT", "BIOMTEST" and "TXTTEST" would be easier to follow?

famosab: yes maybe, i would just be consistent - ususally we do the naming based on th inputs not the outputs so maybe that is just my oversight.

modules/nf-core/biomformat/convert/tests/main.nf.test
jcbioinformatics: Thanks!

I basically just copy and pasted the arguments for VSEARCH from https://github.com/nf-core/modules/blob/8b6514c107bac9520a69aac3471f33e2ab07f269/modules/nf-core/vsearch/cluster/tests/nextflow_fast.config#L1-L7

But, now that you mention it, I don't know why they would have needed to switch quote types. I'll fix that in the next commit

jcbioinformatics:
> A few comments that might help resolving the tests that do not run.
> 
> My guess is because you use the aliases that we have to add the args always even if they are not accessed (weird edge case).

Thanks! 

But, it could be me, but it looks like switching to using empty quotes for params that shouldn't be needed locally causes the tests to fail for me now.

I'm not positive, but I think what might be happening is the global setup modules are being retriggered for each test. And, because of that, we need to specify all the params for each test, and we also need to specify all the args so that they run properly.

I feel really weird pushing something that fails locally here, so for now, here's a screenshot of an error for the `convert - to text` test after making the changes to pass empty params when they shouldn't have been needed.

image

This error refers to the setup for `VSEARCH_CLUSTER_TXTOUT`, which is NOT needed for the `convert - to text` test (takes biom file and converts to text).

Let me know if I should still push these changes so you can see exactly what I did

famosab:
This is for sure a very weird behavior... I think we need to try and catch that. Lets bring this up on the slack :) Maybe you can post in #nf-test

jcbioinformatics:
Heh, I'm still getting used to using GitHub for anything besides just pushing code, so I may have did this somewhat weirdly...

Basically, commit 65e62f501e4ac91e42207ffb2962a9cb6fdac7bd should address making quoting consistent, as well as updating the param and setup module names. I'm not sure if the new aliases are any better per se since I basically just dropped the word "out", but I think it's less confusing since the setup is producing one file type but the module itself is producing a different one. 

I realized for the params that I could basically just re-use names and change the value of the args3 for VSEARCH and have it execute alright. That's why I kinda pushed this all together since I'm hoping the tests will all pass, which doesn't seem to have been the case, darn (I swear they all passed locally).

Anyway, I'm planning on asking for help on nf-test tomorrow since both VSEARCH modules (but not the gunzip modules) are being triggered, and I feel like that's definitely not the intended behavior.

image

I'm not sure if the error I'm getting about params with the CI tests is definitely related or not, though I suspect it is.

image

I realize I'm testing locally with Nextflow 25.04.4, but it passed before when I just had module_args in params. 

However, I realize now it did still give a warning about params https://github.com/nf-core/modules/actions/runs/16520459661/job/46721461606

image

Uh, sorry for the wall of text, I got a bit carried away there. I'll try to be more concise when I post to nf-test channel tomorrow :)

modules/nf-core/biomformat/convert/tests/nextflow.config
famosab, line 1: Nice solution!

modules/nf-core/biomformat/convert/tests/main.nf.test
```diff
@@ -0,0 +1,202 @@
...
+                """
+                input[0] = VSEARCH_CLUSTER_BIOM.out.biom
+                """
+            }
+        }
+
+        run ("VSEARCH_CLUSTER", alias: "VSEARCH_CLUSTER_TXT") {
+
+            script "../../../vsearch/cluster/main.nf"
+
```
famosab: Moving the params here did not work right?

modules/nf-core/biomformat/convert/tests/main.nf.test
```diff
@@ -0,0 +1,202 @@
...
+                """
+            }
+        }
+
+        run ("GUNZIP", alias: "GUNZIP_TXT") {
+
+            script "../../../gunzip/main.nf"
+            process {
+                """
+                input[0] = VSEARCH_CLUSTER_TXT.out.otu
```
famosab:
```suggestion
                input[0] = VSEARCH_CLUSTER_TXT.out.otu
```

Is otu the correct output?

modules/nf-core/biomformat/convert/tests/main.nf.test
famosab, line 1: Maybe we do not need the aliases if the inputs are the same and they are just controlled by the handed params. Meaning you only have "VSEARCH_CLUSTER" once in the setup block and then you get only aliases for the gunzip? Or we just stop this madness and move the setup back to the respective test then we do not have to deal with the aliases and call it a day for now.

famosab: Or we move the gunzip part 🤔

modules/nf-core/biomformat/convert/tests/main.nf.test
jcbioinformatics: Yes, that's probably the clearest indication that something's not right with how nf-test is handling params, and my guess is it has to be something with the aliases since [the example you gave](https://github.com/nf-core/modules/blob/6f835e2153334430b0519fd5bb53d146b26cf1a4/modules/nf-core/syri/tests/main.nf.test#L13) works.

I even tested the syri module locally just to be absolutely sure it wasn't something that broke with Nextflow version differences between what I'm using and what the CI's using, and good news for syri, bad news for me, it did work :)

modules/nf-core/biomformat/convert/tests/main.nf.test
jcbioinformatics: Yes :)

Now that you mention it, I realize that does kinda look like a typo. 

Speaking of though, I think nf-test running both VSEARCH setup modules might actually be the intended behavior.

My guess is there must be a check that all expected output channels from the setup modules are available (even if they don't seem to be used), and that's why it was only running both of the VSEARCH ones, whereas only one GUNZIP was being triggered. I could be completely wrong about though, but it'd make sense to me

modules/nf-core/biomformat/convert/tests/main.nf.test
jcbioinformatics: Huh, I tried moving gunzip to local setup in bc60e9440ac63c07a3a01b9991858259d32428e0 to avoid aliasing, and it's still failing. 

I really thought avoiding aliasing entirely would make it work with the CI tests too. It's still strange to me that it works fine, with Singularity at least locally and then fails that test on pushing

jcbioinformatics: Wait, we have hope :)

I just realized I didn't think to doublecheck the [config](https://github.com/nf-core/modules/blob/6f835e2153334430b0519fd5bb53d146b26cf1a4/modules/nf-core/muscle/tests/nextflow.config) for the Muscle example, so I missed the closures for specifying the params for setup. 

Sorry about that, but that should be fixed in 6e408472b6e2ade64eacbcea2d9623117fd26c25

The fact that I couldn't match syri's style of passing the params is weird, but hopefully this will make the muscle-style work with the CI tests too

jcbioinformatics: Okay, so that didn't work, but I realized the Muscle example uses local setup instead, so just to see if we can get it to work with the GitHub CI tests too (I swear these commits have all been passing locally with Singularity), I moved the VSEARCH module to local for all five tests in this commit aceaf6f67dfa8cb9e3124280bb96cfa8046b3671

I definitely don't like how much duplicated code there is this way, but I realized it'd be a lot better to make a simpler reproducible example to submit to the nf-test channel, instead of point them to this PR. 

My hope is to find a way to break the syri test and then telling them exactly how I did that and pointing to this PR as another instance of an issue with params, but that will probably have to wait for next week or the week after at this point

jcbioinformatics: ...I really should wait for the tests to run before commenting.

Anyway, at this point, it should now match Muscle's example, except for having two setup modules instead of one. And, it's failing.

On the bright side, this gives more evidence vmelichar was right in saying having multiple modules in setup is a contributing factor to the params issue https://github.com/nf-core/modules/pull/8812#issuecomment-3126737631 

The kinda cool thing is since it failed even when aliases were not used, we can say it's not aliasing, or at least not just that. 

I think my next attempt is going to be to try to match the varscan/fpfilter module tests, except with aliasing and see how that goes

jcbioinformatics:
Right, just to try to summarize what I think I've learned about the nf-test params issue(s).

I think there's an issue with having more than one module in local setup because my attempt to match [Muscle's test](https://github.com/nf-core/modules/blob/6f835e2153334430b0519fd5bb53d146b26cf1a4/modules/nf-core/muscle/tests/main.nf.test) still failed, and that one has just one module in local 

Commit: 6e408472b6e2ade64eacbcea2d9623117fd26c25

You can have more than one setup module in global though and have it pass as shown by [varscan/fp](https://github.com/nf-core/modules/blob/8476d6a2adf7fb2e2f23229344aeabcd94a14b0b/modules/nf-core/varscan/fpfilter/tests/main.nf.test)

But, I need to have aliases if we do that because of the need for different VSEARCH args for ext.args3.

So, trying to match to that also failed

Commit: 881e2cdcbd4f4fc11747a0859f44c0b12dc20a2d

I think my summary then is you can get around the multiple module issue (by moving to global), and you can get around the aliasing issue (by moving to local), so that's a problem for me.

What is still weird is my commits have consistently passed locally with nf-test 0.9.2, nf-core 3.3.1, and Nextflow 25.04.4. 

I feel like I've hit a road block here, and with there being potentially two different issues, I think it might not be until week after next that I am able to make a simple test example for both to share with nf-test.

I'm about as sure as I can be that these are legitimate issues with nf-test and not just me messing up.

Huh, shoot, sorry for another wall of text.

Anyway, as a temporary workaround, could I switch back to defining the VSEARCH arguments in the nextflow.config file?

jcbioinformatics:
Okay, so that still failed with GitHub CI testing, which is really strange. The silver lining is, this also means the nf-test's params handling may not have been the problem, or at least not the only problem.

Here's my current plan, which will probably only be next week until I get to try it, but I am determined to see this through :)
* Revert to latest commit that passed tests
* Change name of aliases (confirm that still passes on GitHub)
* Change test for json biom output (confirm that passes)
* Change quoting in config (confirm that passes)
  * That should get us to working tests, though with params in the config
*  Switch to using params for setup with global and update param naming to reduce redundancy
    * Fingers-crossed, that'll just workTM
    * If not, then I'll try the local setup method again without aliases, and that will hopefully work 
    * If neither of those work, then we might be stuck with keeping params in the config, and I'll try to come up with a good minimal reproducible example to share with the nf-test channel

I feel like going back thru these changes again like this will let us get to a place where we'll be able to figure things out :)

Sorry this has kinda dragged out a bit. But, thanks a lot for baring with me!

famosab:
@jcbioinformatics I think all tests are passing now. I would suggest to add the stub tests back and then we can get thise merged in :)

jcbioinformatics:
> @jcbioinformatics I think all tests are passing now. I would suggest to add the stub tests back and then we can get thise merged in :)

Thanks!

I think what happened is specifying params within global setup is problematic (or I just did it wrong), but specifying them with the module's `when` block works as expected, as long as all triggered setup modules have access to the required params. 

But, I made the (as it turns out :) very bad assumption that if the stub tests were still failing, the other tests had issues too (without checking the rest of the summaries for the failed tests). Looking back, it turns out the stub tests were failing the entire time https://github.com/nf-core/modules/actions/runs/16535011409 (I was testing with a newer version of Nextflow but didn't put two and two together since I had previously been testing with the same version the CI uses, which also explains why the tests were passing for me but failing on GitHub)

I think the issue with stub tests here is related to https://github.com/nextflow-io/nextflow/issues/5456 

Admittedly, there are plenty of stub tests that are still passing, so I'm not certain that's what's going on. But, I do get that same error with the stub tests locally when I use 24.10 too.

It looks like the CI's still using 24.10.2, so there's a chance the stub tests may just continue to fail for me.

Sorry for yet another wall of text, but I wanted to make sure I noted that a lot of my failures here were probably actually due to the stub tests, not params.

I just re-added the stub tests and pushed (pass locally with nf-test 0.9.2 and Nextflow 25.04.6), so I guess we'll see if I'm right about them continuing to fail and go from there :)

jcbioinformatics:
> > @jcbioinformatics I think all tests are passing now. I would suggest to add the stub tests back and then we can get thise merged in :)
> 
> Thanks!
> 
> I think what happened is specifying params within global setup is problematic (or I just did it wrong), but specifying them with the module's `when` block works as expected, as long as all triggered setup modules have access to the required params.
> 
> But, I made the (as it turns out :) very bad assumption that if the stub tests were still failing, the other tests had issues too (without checking the rest of the summaries for the failed tests). Looking back, it turns out the stub tests were failing the entire time https://github.com/nf-core/modules/actions/runs/16535011409 (I was testing with a newer version of Nextflow but didn't put two and two together since I had previously been testing with the same version the CI uses, which also explains why the tests were passing for me but failing on GitHub)
> 
> I think the issue with stub tests here is related to [nextflow-io/nextflow#5456](https://github.com/nextflow-io/nextflow/issues/5456)
> 
> Admittedly, there are plenty of stub tests that are still passing, so I'm not certain that's what's going on. But, I do get that same error with the stub tests locally when I use 24.10 too.
> 
> It looks like the CI's still using 24.10.2, so there's a chance the stub tests may just continue to fail for me.
> 
> Sorry for yet another wall of text, but I wanted to make sure I noted that a lot of my failures here were probably actually due to the stub tests, not params.
> 
> I just re-added the stub tests and pushed (pass locally with nf-test 0.9.2 and Nextflow 25.04.6), so I guess we'll see if I'm right about them continuing to fail and go from there :)

Okay, next time, I swear, I'll just wait for the tests to finish before submitting the comment :)

The stub tests do still fail https://github.com/nf-core/modules/actions/runs/16803132230

I think we have three options then

1. We could merge in as is, with the stub tests failing, since once the CI is updated to use a newer version of Nextflow, they should be fine

2. We could merge in without the stub tests, and after the CI tests are updated to use Nextflow >25, I could submit a PR to add them back in

3. Or, we could wait until the CI test is updated and then merge in with the stub tests (I'm nearly positive I'm right and they'll pass with an updated Nextflow version since they pass locally)

I don't like the idea of merging in with failing tests, but I'd also feel weird about having something that works just waiting for an update to the CI tests, so I think my preference would be 2, though I'm not opposed to any of those options.

What do you think?

modules/nf-core/biomformat/convert/tests/main.nf.test
```diff
@@ -0,0 +1,187 @@
...
+            params {
+                vsearch_cluster_args   = '--id 0.8'
+                vsearch_cluster_args2  = '--cluster_fast'
+                vsearch_cluster_args3  = '--biomout'
+                module_args            = '--to-tsv'
+            }
+
+            process {
+                """
+                input[0] = GUNZIP_BIOM.out.gunzip
```
famosab, line 145:
```suggestion
                input[0] = [
                    [ id:'test', single_end:false ], // meta map
                    file(params.modules_testdata_base_path + 'genomics/sarscov2/illumina/fastq/test_1.fastq.gz', checkIfExists: true)
                ]
```
Lets be slightly incorrect and not use modules that need setup for the stub test.

modules/nf-core/biomformat/convert/tests/main.nf.test
```diff
@@ -0,0 +1,187 @@
...
+            params {
+                vsearch_cluster_args  = '--id 0.8'
+                vsearch_cluster_args2 = '--cluster_fast'
+                vsearch_cluster_args3 = '--otutabout'
+                module_args           = '--table-type="OTU table" --to-json'
+            }
+
+            process {
+                """
+                input[0] = GUNZIP_TXT.out.gunzip
```
famosab, line 173:
```suggestion
                input[0] = [
                    [ id:'test', single_end:false ], // meta map
                    file(params.modules_testdata_base_path + 'genomics/sarscov2/illumina/fastq/test_1.fastq.gz', checkIfExists: true)
                ]
```
Lets be slightly incorrect and not use modules that need setup for the stub test.

famosab (commented):
Can you check if the stub failure is due to usage of setup? Because I suspect that that is the issue because I never had problems with stub tests. Because the stub sections might not be entirely compatible for all the setup modules :)

famosab:
We can also easily aim for another input file that is more similar to what is actually used. But I think it does not really matter as the stub section does not use the input file at all it just creates different outputs based on the flags and these are the things that we want to test :)

famosab:
The failure occurs because VSEARCH_CLUSTER has no stub section :) I will open an issue for that or link the appropriate one to here. For now I would suggest to do it as I wrote above just use a dummy inout for the stub test and we will be fine.

https://github.com/nf-core/modules/blob/88869fced9dc0c56043eff2dc748a5f47c57495d/modules/nf-core/vsearch/cluster/main.nf

Edit: stubs are tracked here: https://github.com/nf-core/modules/issues/4570

jcbioinformatics:
> The failure occurs because VSEARCH_CLUSTER has no stub section :) I will open an issue for that or link the appropriate one to here. For now I would suggest to do it as I wrote above just use a dummy inout for the stub test and we will be fine.
> 
> https://github.com/nf-core/modules/blob/88869fced9dc0c56043eff2dc748a5f47c57495d/modules/nf-core/vsearch/cluster/main.nf
> 
> Edit: stubs are tracked here: #4570

Thanks for the suggestions! Fingers-crossed, that'll do the trick now :)

Huh, I didn't even think to check if VSEARCH_CLUSTER had a stub section. I guess I just assumed all modules had that, oops

modules/nf-core/biomformat/convert/tests/main.nf.test
```diff
@@ -0,0 +1,193 @@
...
+
+    test("convert - to text - stub") {
+
+        options "-stub"
+
+        when {
+            params {
+                vsearch_cluster_args   = '--id 0.8'
+                vsearch_cluster_args2  = '--cluster_fast'
+                vsearch_cluster_args3  = '--biomout'
```
famosab:
```suggestion
```

modules/nf-core/biomformat/convert/tests/main.nf.test
```diff
@@ -0,0 +1,193 @@
...
+
+    test("convert - to json biom - stub") {
+
+        options "-stub"
+
+        when {
+            params {
+                vsearch_cluster_args  = '--id 0.8'
+                vsearch_cluster_args2 = '--cluster_fast'
+                vsearch_cluster_args3 = '--otutabout'
```
famosab, line 167:
```suggestion
```
I guess we need to remove those otherwise vsearch_cluster is triggered

famosab:
It seems that vsearch_cluster is triggered anyway. Let me add a stub to that so you can finally get this merged in... This is getting annoying. Maybe we have to raise the issue still somewhere though 🤔

jcbioinformatics:
> It seems that vsearch_cluster is triggered anyway. Let me add a stub to that so you can finally get this merged in... This is getting annoying. Maybe we have to raise the issue still somewhere though 🤔

Thanks!

That's weird that it still triggers, but I guess nf-test is just designed to err on the side of running setup modules if it seems like they might be needed in anyway. In hindsight, I think I should've chosen different modules for my first attempted contributions, so thanks again for sticking with me through all of this!

famosab:
If I add this stub we might be able to run it as you set it up once with the params and with vsearch and gunzip :)

modules/nf-core/biomformat/convert/tests/main.nf.test
```diff
@@ -0,0 +1,187 @@
...
+
+    }
+
+    test("convert - to text - stub") {
+
+        options "-stub"
+
+        when {
+            params {
+                module_args            = '--to-tsv'
```
famosab:
```suggestion
                vsearch_cluster_args  = '--id 0.8'
                vsearch_cluster_args2 = '--cluster_fast'
                vsearch_cluster_args3 = '--biomout'
                module_args           = '--to-tsv'
```

modules/nf-core/biomformat/convert/tests/main.nf.test
```diff
@@ -0,0 +1,187 @@
...
+
+    }
+
+    test("convert - to json biom - stub") {
+
+        options "-stub"
+
+        when {
+            params {
+                module_args           = '--table-type="OTU table" --to-json'
```
famosab, line 168:
```suggestion
                vsearch_cluster_args  = '--id 0.8'
                vsearch_cluster_args2 = '--cluster_fast'
                vsearch_cluster_args3 = '--otutabout'
                module_args           = '--table-type="OTU table" --to-json'
```

modules/nf-core/biomformat/convert/tests/main.nf.test
```diff
@@ -0,0 +1,193 @@
...
+                vsearch_cluster_args3 = '--biomout'
+                module_args           = '--to-tsv'
+            }
+
+            process {
+                """
+                input[0] = [
+                    [ id:'test', single_end:false ], // meta map
+                    file(params.modules_testdata_base_path + 'genomics/sarscov2/illumina/fastq/test_1.fastq.gz', checkIfExists: true)
+                ]
```
famosab:
```suggestion
                input[0] = GUNZIP_BIOM.out.gunzip
```
Next we can try to go back to this :)

modules/nf-core/biomformat/convert/tests/main.nf.test
```diff
@@ -0,0 +1,193 @@
...
+                vsearch_cluster_args3 = '--otutabout'
+                module_args           = '--table-type="OTU table" --to-json'
+            }
+
+            process {
+                """
+                input[0] = [
+                    [ id:'test', single_end:false ], // meta map
+                    file(params.modules_testdata_base_path + 'genomics/sarscov2/illumina/fastq/test_1.fastq.gz', checkIfExists: true)
+                ]
```
famosab:
```suggestion
                input[0] = GUNZIP_TXT.out.gunzip
```

modules/nf-core/biomformat/convert/meta.yml
```diff
@@ -0,0 +1,62 @@
+name: "biomformat_convert"
+description: Convert biom table to different format
+keywords:
+  - biom
+  - feature table
+  - formatting
+tools:
+  - "biomformat":
+      description: "Biological Observation Matrix (BIOM) format"
```
famosab: Can we be a bit more descriptive here? Like what does the tool do exept for the format?

modules/nf-core/biomformat/convert/meta.yml
```diff
@@ -0,0 +1,62 @@
+name: "biomformat_convert"
+description: Convert biom table to different format
+keywords:
```
famosab, line 5: You can also add more keywords if you like :)

modules/nf-core/biomformat/convert/main.nf
```diff
@@ -0,0 +1,51 @@
...
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def output = task.ext.args.contains("--to-tsv") ? "${prefix}.txt" : "${prefix}.biom"
+    if( "$output" == "$biom" ) error "ERROR: Input and output names are the same, set prefix in module configuration"
+    """
+    biom convert \\
+        -i $biom \\
+        -o $output \\
+        $args
```
famosab:
```suggestion
        -i ${biom} \\
        -o ${output} \\
        ${args} 
```

modules/nf-core/biomformat/convert/main.nf
```diff
@@ -0,0 +1,51 @@
...
+    END_VERSIONS
+    """
+
+    stub:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def output = task.ext.args.contains("--to-tsv") ? "${prefix}.txt" : "${prefix}.biom"
+    if( "$output" == "$biom" ) error "ERROR: Input and output names are the same, set prefix in module configuration"
+    """
+    touch $output
```
famosab:
```suggestion
    touch ${output}
```

modules/nf-core/biomformat/convert/main.nf
```diff
@@ -0,0 +1,51 @@
...
+    "${task.process}":
+        biom-format: \$(biom --version | cut -f 3 -d ' ')
+    END_VERSIONS
+    """
+
+    stub:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def output = task.ext.args.contains("--to-tsv") ? "${prefix}.txt" : "${prefix}.biom"
+    if( "$output" == "$biom" ) error "ERROR: Input and output names are the same, set prefix in module configuration"
```
famosab:
```suggestion
    if( "${output}" == "${biom}" ) error "ERROR: Input and output names are the same, set prefix in module configuration"
```

modules/nf-core/biomformat/convert/main.nf
```diff
@@ -0,0 +1,51 @@
...
+    path "versions.yml"                            , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def output = task.ext.args.contains("--to-tsv") ? "${prefix}.txt" : "${prefix}.biom"
+    if( "$output" == "$biom" ) error "ERROR: Input and output names are the same, set prefix in module configuration"
```
famosab:
```suggestion
    if( "${output}" == "${biom}" ) error "ERROR: Input and output names are the same, set prefix in module configuration"
```

modules/nf-core/biomformat/convert/tests/main.nf.test
```diff
@@ -0,0 +1,187 @@
...
+                """
+                input[0] = GUNZIP_TXT.out.gunzip
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
```
famosab:
```suggestion
                { assert snapshot(
                    process.out,
                    path(process.out.versions[0]).yaml,
                ).match() }
```
Just to make sure that accessing the version works correctly it makes sense to add this to the stub :)

modules/nf-core/biomformat/convert/tests/main.nf.test
famosab, line 1: One more thing on test names: we have a guideline where we put (organism - fileformat - whatever else you need) as test name :) You can adjust the test names or you leave them as is, I think they are descriptive enough but just for future reference :)

famosab (commented):
Now that we got everything running here are a few more small remarks to give this module the finishing touch ✨

modules/nf-core/biomformat/convert/meta.yml
jcbioinformatics: I think the biom format is mostly used for amplicon analysis, but I know the MetaPhlAn tool (for shotgun data) also supports it, so I added a couple key words in this commit to reflect that 96fe6024bdf9524c98af542349c01f75d7ea54b3

modules/nf-core/biomformat/convert/tests/main.nf.test
jcbioinformatics: Ah, that makes a lot of sense, thanks!

I didn't see that as a formal guideline https://nf-co.re/docs/guidelines/components/modules#test-names

But, I noticed most tools followed that pattern (and two of the examples at that link), so I updated that now in 96fe6024bdf9524c98af542349c01f75d7ea54b3

modules/nf-core/biomformat/convert/meta.yml
jcbioinformatics: I expanded the description in 0927f804682957df5695d3b99e0ccc1ae1de3a43

I guess the tricky thing is, there's both the biom format file format, and the biom format tool :) I think adding the functionality they name for the CLI [here](http://biom-format.org/#the-biological-observation-matrix-biom-format) makes sense though

modules/nf-core/biomformat/convert/tests/main.nf.test
famosab: The section you linked says what I wanted to say: Or at least thats what I understand when I read this :)

famosab (approved):
LGTM 🚀 Thanks for sticking around and solving all these issues 👀

modules/nf-core/biomformat/convert/tests/main.nf.test
jcbioinformatics: Ah, got it.

I guess I was going for a more condensed version of the third given test name there since often times in practice there's data from more than one taxon represented in the input and output tables, but for the purposes of this test, it is just COVID, so oops. I'll make sure to follow that pattern moving forward :)

jcbioinformatics:
> LGTM 🚀 Thanks for sticking around and solving all these issues 👀

Awesome! Thanks for all of your help (and patience :)!

PR #8808, opened 2025-07-24
POC: declare ext variable defaults as directives

nvnieuwk:
This PR is a POC (proof-of-concept) to show how we could reduce redundancy in the modules by declaring the defaults for `ext` variables (like `prefix`) once as a directive instead of once in the script block and once in the stub block


SPPearce:
So, do we want to move to this structure?

nvnieuwk:
I would love to, it would decrease code duplication without changing how the modules work

mahesh-panchal (approved):
I like this idea. I hadn't thought of this. 🚀

modules/nf-core/samtools/sort/main.nf
```diff
@@ -18,12 +19,11 @@ process SAMTOOLS_SORT {
...
 
-    when:
-    task.ext.when == null || task.ext.when
+    when: task.ext.when
 
     script:
-    def args = task.ext.args ?: ''
-    def prefix = task.ext.prefix ?: "${meta.id}"
+    def args = task.ext.args
+    def prefix = task.ext.prefix
```
SPPearce: Do we even need this bit at this point, could we just use task.ext.args directly inside the code?

nvnieuwk: Yeah we could! I've also thought about this and it makes more sense, I'll update this PR after the retreat

mahesh-panchal:
Something I was wondering was whether `ext prefix: "${meta.id}", args: '', when: true` is then needed in every file. It could be set generally for all processes. 

But then I remembered that these modules could be used outside the nf-core template.

nvnieuwk:
Yeah I'd keep them in the module to ensure that it can work on it's own. Especially if we want to use specific defaults per module

JoseEspinosa (approved):
I also like it, specially with @SPPearce suggestion!

mashehu:
is this PoC still up-to-date? 
this doesn't seem like it reduces a lot code...

nvnieuwk:
I'll update it now :)

nvnieuwk:
Done! It should be up to date now

mashehu:
@nf-core-bot fix linting

PR #6383, opened 2024-08-27
add dragen module - v2

marrip:
## PR checklist

Closes #4026 

- [ ] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [ ] Remove all TODO statements.
- [ ] Emit the `versions.yml` file.
- [ ] Follow the naming conventions.
- [ ] Follow the parameters requirements.
- [ ] Follow the input/output options guidelines.
- [x] Add a resource `label`
- [ ] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - For modules:
    - [ ] `nf-core modules test <MODULE> --profile docker`
    - [ ] `nf-core modules test <MODULE> --profile singularity`
    - [ ] `nf-core modules test <MODULE> --profile conda`
  - For subworkflows:
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile docker`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile singularity`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile conda`

old PR: #4111

PR #8823, opened 2025-07-29
Add dragen germline module

marrip:
Hey guys!

After almost one year I finally could fix the issues we had with Dragen and try out the module. I reduced it a lot to only feature inputs and outputs relevant for germline analysis. I haven't added tests yet but wanted to discuss the current state of the module first. Let me know what you think.

Old PR: https://github.com/nf-core/modules/pull/6383

## PR checklist

Closes #XXX 

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/dragen/germline/tests/main.nf.test.snap
+++ b/modules/nf-core/dragen/germline/tests/main.nf.test.snap
@@ -149,11 +149,11 @@
                 "141": [
                     [
                         "20251120_usage.txt:md5,d41d8cd98f00b204e9800998ecf8427e",
-                        "versions.yml:md5,62a0b1f9b24090da6949054fce141347"
+                        "versions.yml:md5,0800f35b86c052ff03f30ce186300e7e"
                     ]
                 ],
                 "142": [
-                    "versions.yml:md5,62a0b1f9b24090da6949054fce141347"
+                    "versions.yml:md5,0800f35b86c052ff03f30ce186300e7e"
                 ],
                 "15": [
                     
@@ -446,7 +446,7 @@
                 "all": [
                     [
                         "20251120_usage.txt:md5,d41d8cd98f00b204e9800998ecf8427e",
-                        "versions.yml:md5,62a0b1f9b24090da6949054fce141347"
+                        "versions.yml:md5,0800f35b86c052ff03f30ce186300e7e"
                     ]
                 ],
                 "baf_bedgraph_gz": [
@@ -849,7 +849,7 @@
                     
                 ],
                 "versions": [
-                    "versions.yml:md5,62a0b1f9b24090da6949054fce141347"
+                    "versions.yml:md5,0800f35b86c052ff03f30ce186300e7e"
                 ],
                 "wgs_contig_mean_cov_csv": [
                     
@@ -870,7 +870,7 @@
             [
                 {
                     "DRAGEN": {
-                        "dragen": "4.2.7"
+                        "dragen": "stub"
                     }
                 }
             ]


famosab:
Thank you for that huge contribution 👀 I was thinking whether there would be a way to simplify this a little so that the module does not have as many input and outputs. But at the same time I think DRAGEN is aimed to be a variant-calling-pipeline itself so that might be hard. 

I thought we could maybe split this into more submodules and then collect those into a DRAGEN subworkflow. But I am unsure whether that is something worth the work. We also have to check if all inputs are required by DRAGEN or if they could also be supplied with ext.args

modules/nf-core/dragen/germline/main.nf
```diff
@@ -0,0 +1,411 @@
+process DRAGEN {
+    tag "$meta.id"
+    label 'process_long'
+
+    // ATTENTION: No conda env or container image as Dragen requires specialized hardware to run
```
famosab, line 5: How would we make this compatible with nf-core pipelines then? 🤔

SPPearce: Something like this: https://github.com/seqeralabs/nf-dragen?tab=readme-ov-file#pipeline-implementation
But not every module in nf-core/modules is necessarily used in nf-core pipelines.

modules/nf-core/dragen/germline/meta.yml
```diff
@@ -0,0 +1,1493 @@
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/meta-schema.json
+name: "dragen"
+description: The Illumina DRAGEN™ Bio-IT Platform is based on the highly reconfigurable DRAGEN Bio-IT Processor, which is integrated on a Field Programmable Gate Array (FPGA) card and is available in a preconfigured server that can be seamlessly integrated into bioinformatics workflows. The platform can be loaded with highly optimized algorithms for many different NGS secondary analysis pipelines.
```
SPPearce: I think this wants to be a description about the germline bit, not just copy/pasted from below.

modules/nf-core/dragen/germline/meta.yml
```diff
@@ -0,0 +1,1493 @@
...
+keywords:
+  - genomics
+  - variant calling
+  - germline
```
SPPearce, line 15: We can have more than 3 keywords, I suspect there could be a lot more here.

modules/nf-core/dragen/germline/meta.yml
```diff
@@ -0,0 +1,1493 @@
...
+      description: Files of adapter sequences to trim from the 3' end of read 1 and 2
+      pattern: "*.{fasta,fa}"
+  - trim_adapter_read_5prime:
+      type: file
+      description: FASTA files that contains adapter sequences to trim from the 5' end of Read 1 and 2; The sequences should be in reverse order (with respect to their appearance in the FASTQ) but not complemented
+      pattern: "*.{fasta,fa}"
+  - variant_annotation_data:
+      type: file
+      description: // keep?
+      pattern: // keep?
```
SPPearce: ?

modules/nf-core/dragen/germline/meta.yml
```diff
@@ -0,0 +1,1493 @@
...
+      description: Specifies the input files for compression or decompression
+      pattern: "*.ora"
+  - bam:
+      type: file
+      description: Input BAM file to send to card
+      pattern: "*.bam"
+  - cram:
+      type: file
+      description: Input CRAM file to send to card
+      pattern: "*.cram"
```
SPPearce: Could we make this one input and figure out based on extension which?

SPPearce:
> Thank you for that huge contribution 👀 I was thinking whether there would be a way to simplify this a little so that the module does not have as many input and outputs. But at the same time I think DRAGEN is aimed to be a variant-calling-pipeline itself so that might be hard.
> 
> I thought we could maybe split this into more submodules and then collect those into a DRAGEN subworkflow. But I am unsure whether that is something worth the work. We also have to check if all inputs are required by DRAGEN or if they could also be supplied with ext.args

I think this should just be one tool, as you say it is a variant calling pipeline in itself.

modules/nf-core/dragen/germline/main.nf
marrip: so we tag it with the label `dragen` instead of `process_long`?

modules/nf-core/dragen/germline/meta.yml
marrip: Any suggestions?

modules/nf-core/dragen/germline/meta.yml
marrip: sorry, I missed this, thanks for pointing it out 😉

modules/nf-core/dragen/germline/meta.yml
marrip: do you want all sequence data containing inputs combined?

marrip:
> > Thank you for that huge contribution 👀 I was thinking whether there would be a way to simplify this a little so that the module does not have as many input and outputs. But at the same time I think DRAGEN is aimed to be a variant-calling-pipeline itself so that might be hard.
> > I thought we could maybe split this into more submodules and then collect those into a DRAGEN subworkflow. But I am unsure whether that is something worth the work. We also have to check if all inputs are required by DRAGEN or if they could also be supplied with ext.args
> 
> I think this should just be one tool, as you say it is a variant calling pipeline in itself.

I agree, we already had a bigger version of this and agreed on reducing it to one use case. ☺️ 

Sorry for the late reply, I was on holiday.

modules/nf-core/dragen/germline/meta.yml
marrip: I added a more specific description. Hope it is not too long?

modules/nf-core/dragen/germline/meta.yml
SPPearce: Well, what does dragen do?
SNVs? indels? SVs? CNV? Other more obscure options?

modules/nf-core/dragen/germline/meta.yml
SPPearce: I don't know, what else comes under that (fastq, bam, cram, anything else?)
Bam and cram are treated pretty much the same in many tools (subject to compliance).
Basically I don't really like modules with lots of input channels as they are quite hard to use as we don't have named inputs.

marrip: I included more keywords. Let me know what you think. ☺️

modules/nf-core/dragen/germline/meta.yml
marrip: I included more keywords. Let me know what you think. ☺️

modules/nf-core/dragen/germline/meta.yml
marrip: I attempted to handle all in one in this commit: https://github.com/nf-core/modules/pull/8823/commits/1a26f3c6e8458ca1d9cd94ded27d028fbe89426a

modules/nf-core/dragen/germline/main.nf
```diff
@@ -0,0 +1,406 @@
...
+    // input data
+    def bam_rex = /\.bam$/
+    def cram_rex = /\.cram$/
+    def fastq_rex = /\.f(ast)?q(\.gz)?$/
+    def ora_rex = /\.ora$/
+    def format_input = ""
+    if (input instanceof List) {
+        if (input.size() != 2) {
+            error "Error: a maximum of 2 input files is supported."
+        }
```
adamrtalbot, line 214: Congrats on a huge job implementing this!

Just fyi, Dragen will accept multiple FASTQ files if you supply the first one in the matching list, e.g. these FASTQs:

```
input_S1_L001_R1_001.fastq.gz
input_S1_L001_R1_002.fastq.gz
input_S1_L001_R1_003.fastq.gz
```

Can be specified with the input arg `-1 input_S1_L001_R1_001.fastq.gz`

You can see it in my implementation here (which only supports FASTQ files): https://github.com/seqeralabs/nf-dragen/blob/master/modules/local/dragen.nf

SPPearce: Where do you get `input_S1_L001_R1_002.fastq.gz` from, I have never seen anything that isn't `001` at the end? Manual renaming?

marrip: thanks, @adamrtalbot 🙏 

About the fastq files. Just to clarify, the fastq files basically belong to the same sample and are from different lanes, for example? I also saw that dragen can take a list of fastq files as input but I am not sure if we want to open up for all these different options. That makes things even more complex. I am not against it, I just think the concatenation can be handled outside of the module as well.

adamrtalbot: > Where do you get `input_S1_L001_R1_002.fastq.gz` from, I have never seen anything that isn't `001` at the end? Manual renaming?

I don't know, it's in the Dragen docs: https://support-docs.illumina.com/SW/DRAGEN_v39/Content/SW/DRAGEN/Inputfiles_fDG.htm

Maybe there's a demultiplex setting for breaking up FASTQ files?

adamrtalbot: Eurgh, link doesn't work. Pasted here for reference:

> If using, bcl2fastq or the DRAGEN BCL command use the following common file naming convention:
> 
> <SampleID>_S<#>_<Lane>_<Read>_<segment#>.fastq.gz
> 
> Older versions of bcl2fastq and DRAGEN could segment FASTQ samples into multiple files to limit file size or to decrease the time to generate them.
> 
> For Example:
> ```
> RDRS182520_S1_L001_R1_001.fastq.gz
> RDRS182520_S1_L001_R1_002.fastq.gz
> ...
> RDRS182520_S1_L001_R1_008.fastq.gz
> ```
> These files do not need to be concatenated to be processed together by DRAGEN. To map/align any sample, provide the first file in the series (-1 <FileName>_001.fastq). DRAGEN reads all segment files in the sample consecutively for both of the FASTQ file sequences specified using the -1 and -2 options for paired-end input and for compressed fastq.gz files. To turn the behavior off, set ‑‑enable-auto-multifile to false on the command line.
> 
> DRAGEN can also optionally read multiple files by the sample name given in the file name, which can be used to combine samples that have been distributed across multiple BCL lanes or flow cells. To enable this feature, set the --combine-samples-by-name option to true.
> 
> If the FASTQ files specified on the command-line use the Casava 1.8 file naming convention shown above and additional files in the same directory share that sample name, those files and all their segments are processed automatically. Note that sample name, read number, and file extension must match. Index barcode and lane number can differ.
> 
> To avoid impacting system performance, input files must be located on a fast file system.

marrip: I can try to implement it if you want just want to be sure we are in agreement that it is something people will use.

adamrtalbot: nah, that would be premature optimization. When someone asks for it make a note of this for future reference!

adamrtalbot: thanks for the hard work!

marrip: thank you 🙂

modules/nf-core/dragen/germline/main.nf
marrip: changed it to `process_dragen` now

marrip:
I am a bit uncertain what the tests should look like, should it be just stubbing tests? Which use-cases should be covered? All sequencing input files? For most other input, there not much changing if it is supplied or not. Let me know what your thoughts are.

mashehu:
yes, just stub tests then.

marrip:
any example for a module that only uses stubbing for nf-test?

mashehu:
https://github.com/nf-core/modules/blob/master/modules/nf-core/cadd/tests/main.nf.test and https://github.com/nf-core/modules/blob/master/modules/nf-core/interproscan/tests/main.nf.test

marrip:
@SPPearce would you like to take another look. I addressed your comments and made some changes. Does it look ok now?

@adamrtalbot does the nf-test with stubbing look ok?

marrip:
> Is it worth having an output channel that just contains everything, for publishing?

You mean like just `*`? Or creating an output directory and targeting that? That makes it more readable but it might be less handy when one wants to use specific output as input for another process. I am open to the idea just want to make sure we consider any use case here.

Did you check the input channel and the process label and is that ok?

modules/nf-core/dragen/germline/main.nf
```diff
@@ -0,0 +1,411 @@
...
+    tuple val(meta), path("${prefix}.wgs_hist.csv")                          , optional: true, emit: wgs_hist_csv
+    tuple val(meta), path("${prefix}.wgs_overall_mean_cov.csv")              , optional: true, emit: wgs_overall_mean_cov_csv
+    tuple val(meta), path("${prefix}_callability.bed")                       , optional: true, emit: callability_bed
+    tuple val(meta), path("${prefix}_chr_start-end.impute.phase.vcf.gz")     , optional: true, emit: chr_start_end_impute_phase_vcf_gz
+    tuple val(meta), path("${prefix}_contig_mean_cov.csv")                   , optional: true, emit: contig_mean_cov_csv
+    tuple val(meta), path("${prefix}_cov_report.bed")                        , optional: true, emit: cov_report_bed
+    tuple val(meta), path("${prefix}_evidence.bam")                          , optional: true, emit: evidence_bam
+    tuple val(meta), path("${prefix}_evidence.cram")                         , optional: true, emit: evidence_cram
+    tuple val(meta), path("${prefix}_evidence.sam")                          , optional: true, emit: evidence_sam
```
SPPearce: Should these just be one channel?

marrip: I have combined them in c26575c6ee8e8255abae4d3dc017b6b9ca167a69

marrip:
> I was thinking emitting an extra channel as "*" which is just everything. Probably doesn't play nicely with `scratch` directive though.

I created the `all` output channel here c26575c6ee8e8255abae4d3dc017b6b9ca167a69

modules/nf-core/dragen/germline/main.nf
```diff
@@ -0,0 +1,410 @@
...
+    path "./sv/workspace/logs/config_log.txt"                                , optional: true, emit: config_log_txt
+    path "./sv/workspace/svLocusGraph.bin"                                   , optional: true, emit: svLocusGraph_bin
+    path "body.txt"                                                          , optional: true, emit: body_txt
+    path "dragen.time_metrics.csv"                                           , optional: true, emit: dragen_time_metrics_csv
+    path "header.txt"                                                        , optional: true, emit: header_txt
+    path "match_log.small_indel_dedup.txt"                                   , optional: true, emit: match_log_small_indel_dedup_txt
+    path "match_log.smn_dedup.txt"                                           , optional: true, emit: match_log_smn_dedup_txt
+    path "streaming_log_*.csv"                                               , optional: true, emit: streaming_log_csv
+    path "*_usage.txt"                                                       , emit: usage_txt
+    path "*"                                                                 , emit: all
```
adamrtalbot: Make it recursive!

```suggestion
    path "**"                                                                , emit: all
```

adamrtalbot: This will omit directories and only list the files:

> Directories are included, unless the ** pattern is used to recurse through directories

So it's up to you.

SPPearce:
> > I was thinking emitting an extra channel as "*" which is just everything. Probably doesn't play nicely with scratch directive though.
> 
> I find this really annoying because I have to filter it afterwards, but I guess it's better to have an `all` channel alongside the specific output channels.

Sure, my thought was I didn't want to have to set up publish directives for all of these output channels ;) 
I'm definitely not saying to have an `all` channel **instead** of the individual ones.

modules/nf-core/dragen/germline/main.nf
marrip: done

marrip:
ready to merge? @SPPearce @adamrtalbot

marrip:
thank you and all other reviewers for your help along the way! 🙏

PR #8838, opened 2025-07-31
Adding mem2 mode to bwameth/index

mbeavitt:
## PR checklist

Added a switch such that when params.use_mem2 = TRUE, the index created will be a mem2 index rather than a bwa mem index. Default behaviour is the same.

RE parameters requirements - is it ok to use params.use_mem2 in a module? Can't think of a better way to do this.

Related:
https://github.com/nf-core/modules/pull/8793
https://github.com/nf-core/methylseq/issues/550


sateeshperi:
are we sure the param is working ? I'm seeing there is no diff in the md5's of mem2 files and the default files

mbeavitt:
Oh, well caught... I thought I'd checked but didn't see that. I think I wrote the test wrong, I'll rewrite tomorrow

mbeavitt:
I'm pretty sure the files are different? Unless I'm doing something wrong.
I used to be able to link sections of source code in a comment on github but I can't seem to do it so, copy pasted below with line numbers...
```
 11                             "genome.fasta.bwameth.c2t:md5,98039984526a41d04d6bd92fcc040c62",
 12                             "genome.fasta.bwameth.c2t.amb:md5,249a4195069071ce47cd0bae68abe376",
 13                             "genome.fasta.bwameth.c2t.ann:md5,46524d4359dcdfb203a235ab3b930dbb",
 14                             "genome.fasta.bwameth.c2t.bwt:md5,84f65df7d42dbe84c9ccfaddfdd5ea6b",
 15                             "genome.fasta.bwameth.c2t.pac:md5,4d8e51cb0bbdeaf24576bdf0264d8653",
 16                             "genome.fasta.bwameth.c2t.sa:md5,d25f6486f5134f57ed5b258f6fbb8673"

/// snip ///

 58                             "genome.fasta.bwameth.c2t:md5,98039984526a41d04d6bd92fcc040c62",
 59                             "genome.fasta.bwameth.c2t.0123:md5,1b92d7081e7b54dafcd97aac45a8076d",
 60                             "genome.fasta.bwameth.c2t.amb:md5,249a4195069071ce47cd0bae68abe376",
 61                             "genome.fasta.bwameth.c2t.ann:md5,46524d4359dcdfb203a235ab3b930dbb",
 62                             "genome.fasta.bwameth.c2t.bwt.2bit.64:md5,52fa820b3a029fe6a9b2e18972f616f2",
 63                             "genome.fasta.bwameth.c2t.pac:md5,4d8e51cb0bbdeaf24576bdf0264d8653"
```

SPPearce:
> RE parameters requirements - is it ok to use params.use_mem2 in a module? Can't think of a better way to do this.

Make this a value channel to the process rather than using `params` please.

sateeshperi:
yes that is the recommended way. @mbeavitt plz pass the param as value channel

mbeavitt:
I've made that change - snapshot is unchanged after running the test, FYI!

mbeavitt:
Looks like I'll have to modify bwameth/align too, since it depends on bwameth/index

modules/nf-core/bwameth/index/tests/main.nf.test
```diff
@@ -31,4 +32,26 @@ nextflow_process {
...
+                ])
+                input[1] = true
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match("mem2") }
```
SPPearce: Don't need the name on the snapshots if there is only one snapshot per test.
```suggestion
                { assert snapshot(process.out).match() }
```

SPPearce:
Yes, you'll need to update the other tests that depend on this one.
Don't forget to update the meta.yml too with the extra channel.

SPPearce:
Oh, and FASTA_INDEX_BISMARK_BWAMETH subworkflow.

mbeavitt:
Need to do the parabricks one as well, just noticed

mbeavitt:
Man, CI is useful. This has also revealed that a couple of subworkflows (fastq_align_dedup_bwameth, fasta_index_bismark_bwameth) use a static index tarball generated from bwa (not mem2 version): https://github.com/nf-core/test-datasets/raw/methylseq/reference/Bwameth_Index.tar.gz

I'll need to add an additional test tarball with the mem2 indexes I guess, but maybe that's more suited to the update I'm planning to the bwameth/align module. Or perhaps a separate update to those SWFs to ensure compatibility with mem2 indexes.

sateeshperi:
a separate PR would be good for reviewers

SPPearce (approved):
Yeah, discrete PRs are generally preferable.

PR #8843, opened 2025-08-02
Improve scrublet batch_col handling

nictru:
Just a small fix, previously empty array inputs (often used as placeholders for missing values) were not handled properly


LeonHafner (approved):
LGTM

PR #8864, opened 2025-08-08
fastplong: use provided reads as input

nschan:
This is a small patch to the fastplong module, which did not use the provided reads as input but assumed a filename based on `meta.id`.

### DIFF
--- a/modules/nf-core/fastplong/main.nf
+++ b/modules/nf-core/fastplong/main.nf
@@ -33,7 +33,7 @@ process FASTPLONG {
     def report_title = task.ext.report_title ?: "${prefix}_fastplong_report"
     """
     fastplong \\
-        --in ${prefix}.fastq.gz \\
+        --in ${reads} \\
         $output_file \\
         --json ${prefix}.fastplong.json \\
         --html ${prefix}.fastplong.html \\


nschan:
Not sure why the test-data cannot be downloaded, I didn't change that, the tests ran fine before, but the url gives 404 right now...

Lupphes:
> This is a small patch to the fastplong module, which did not use the provided reads as input but assumed a filename based on `meta.id`.

I think that the same symlink hack should be used as in `fastp`: `[ ! -f  ${prefix}.fastq.gz ] && ln -sf $reads ${prefix}.fastq.gz`

> Not sure why the test-data cannot be downloaded, I didn't change that, the tests ran fine before, but the URL gives 404 right now...

I have no idea, it should be working. Were there any changes in the main?

nschan:
I don't have strong opinions either way, I don't really see the point in the symlinking. If you prefer it its fine with me, but the current module doesn't work reliably. 
There seems to be some unclear github raw problem, hopefully it works again at some point..

Edit:
Upon rethinking this, I think symlinking is actually a bad idea, `fastplong` can take uncompressed and compressed inputs and switches based on the `.gz` extension. If you pass uncompressed reads they will be linked as a `.gz` file and would then presumably be invalid input?

Lupphes:
> Edit: Upon rethinking this, I think symlinking is actually a bad idea, `fastplong` can take uncompressed and compressed inputs and switches based on the `.gz` extension. If you pass uncompressed reads, they will be linked as a `.gz` file and would then presumably be invalid input?

Ah, no, that's valid, let's use the reads directly then, and if users use different input, it will throw an error.

Lupphes:
Does the tests work locally for you?

nschan:
I have not tried locally, but https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/genomics/homo_sapiens/pacbio/fastq/alz.ccs.fastq.gz is giving 404 so I doubt it would work locally,

Lupphes:
> I have not tried locally, but https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/genomics/homo_sapiens/pacbio/fastq/alz.ccs.fastq.gz is giving 404 so I doubt it would work locally,

Yeah, I tried it as well, and it fails. Is it probably a GitHub problem? Because I can see

https://github.com/nf-core/test-datasets/blob/modules/data/genomics/homo_sapiens/pacbio/fastq/alz.ccs.fastq.gz

but when I click "Raw" -> "Open" I get 404

nschan:
Yup, that seems to affect multiple nf-core pipelines / tests right now. I do not know what is causing it, but I hope that it is just a temporary github problem.

famosab (approved):

Lupphes (approved):
Yeah LGTM

PR #8869, opened 2025-08-08
If `${prefix}.fastq.gz` doesn't exist create a symlink to it

dwells-eit:
Thanks for adding `fastplong` to nf-core!

This PR will:
Create a symlink to `${prefix}.fastq.gz` if it is not already a file.

This is needed because:
If the prefix does not match the file name, currently `fastplong` throws an error because it can't find the reads.

Code taken from the nf-core `fastp` module.

## PR checklist

  - For modules:
    - [x ] `nf-core modules test fastplong --profile docker`

### DIFF
--- a/modules/nf-core/fastplong/main.nf
+++ b/modules/nf-core/fastplong/main.nf
@@ -32,6 +32,8 @@ process FASTPLONG {
     def output_file = discard_trimmed_pass ? '' : "--out ${prefix}.fastplong.fastq.gz"
     def report_title = task.ext.report_title ?: "${prefix}_fastplong_report"
     """
+    [ ! -f  ${prefix}.fastq.gz ] && ln -sf $reads ${prefix}.fastq.gz
+
     fastplong \\
         --in ${prefix}.fastq.gz \\
         $output_file \\


modules/nf-core/fastplong/main.nf
```diff
@@ -32,6 +32,8 @@ process FASTPLONG {
     def output_file = discard_trimmed_pass ? '' : "--out ${prefix}.fastplong.fastq.gz"
     def report_title = task.ext.report_title ?: "${prefix}_fastplong_report"
     """
+    [ ! -f  ${prefix}.fastq.gz ] && ln -sf $reads ${prefix}.fastq.gz
+
     fastplong \\
         --in ${prefix}.fastq.gz \\
```
famosab: I think this should rather take the input reads here. Because the symlinking is already done by nextflow itself!
```suggestion
        --in ${reads} \\
```

modules/nf-core/fastplong/main.nf
```diff
@@ -32,6 +32,8 @@ process FASTPLONG {
     def output_file = discard_trimmed_pass ? '' : "--out ${prefix}.fastplong.fastq.gz"
     def report_title = task.ext.report_title ?: "${prefix}_fastplong_report"
     """
+    [ ! -f  ${prefix}.fastq.gz ] && ln -sf $reads ${prefix}.fastq.gz
+
```
famosab: Then we will not need this!

dwells-eit: Good point, that's much cleaner!
I've made the changes you suggested and rerun the tests. It all looks good!

famosab: I think it was overlooked when the module was approved, because it should be done the way you implemented it now :)

famosab:
Oh I think someone else also realized that: https://github.com/nf-core/modules/pull/8864

famosab (approved):

famosab:
Closing this as the issue was resolved with #8864 :)

PR #8873, opened 2025-08-10
bwameth: allow skip time checks

sateeshperi:


vagkaratzas (approved):
I am not sure if this is pipeline specific or it's fine to be used in the module in general. But since the module is currently only used in methylseq, I'll just approve this :P

PR #8876, opened 2025-08-11
Update duplicate file pattern

LuisHeinzlmeier:
- `SCANPY_HASHSOLO.out.assignment` returned a list of `[assignment, params]` because the pattern wasn't clearly.
- This small PR removes this duplicate file pattern


LuisHeinzlmeier:
Thank you for the review @dialvarezs!

PR #8882, opened 2025-08-11
Fix prettier fail of bff

LuisHeinzlmeier:
Fix prettier fail of bff


LuisHeinzlmeier:
Thank you for the review @dialvarezs!

PR #8893, opened 2025-08-12
Decoupler

antoniasaracco:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


modules/nf-core/decoupler/decoupler/meta.yml
```diff
@@ -50,12 +50,13 @@ input:
...
       ontologies:
         - edam: http://edamontology.org/format_3475 # TSV
-  - gtf:
+  - annot:
       type: file
       description: |
-        Annotation file in GTF format
-      pattern: "*.gtf"
-      ontologies: []
+        Annotation file in tsv format
```
atrigila: Add if this is optional or mandatory

modules/nf-core/decoupler/decoupler/templates/decoupler.py
```diff
@@ -43,50 +43,21 @@ def parse_ext_args(args_string: str):
...
-                attributes[key] = value.replace('"', '').strip()
-            gene_id = attributes.get("gene_id")
-            gene_symbol = attributes.get("gene_name") or attributes.get("gene_symbol") or attributes.get("external_gene_name")
-            if gene_id and gene_symbol:
-                mapping[gene_id] = gene_symbol
-    return mapping
-
 # Parse external arguments
-raw_args = "${task.ext.args}"
+raw_args = """${task.ext.args}"""
```
atrigila: How does the triple """ change the behavior?

modules/nf-core/decoupler/decoupler/meta.yml
atrigila: The tsv file seems to have a specific format (gene_id vs gene_name) so perhaps that should be here

atrigila: Explain the use case that annot is used in mapping ensembl ids only

modules/nf-core/decoupler/decoupler/templates/decoupler.py
```diff
@@ -43,50 +43,21 @@ def parse_ext_args(args_string: str):
...
     try:
-        gene_mapping = parse_gtf("${gtf}")
+        annot_df = pd.read_csv("${annot}", sep="\t")
+        gene_mapping = dict(zip(annot_df["gene_id"], annot_df["gene_name"]))
         new_index = [gene_mapping.get(ens, None) for ens in mat.index]
         mat.index = new_index
         mat = mat[mat.index.notnull()]
         mat = mat[~mat.index.duplicated(keep='first')]
     except Exception as e:
-        print("ERROR: Failed to parse GTF:", e)
```
atrigila, line 89: the error message could  indicate exact required column names to help users debug.

modules/nf-core/decoupler/decoupler/tests/main.nf.test.snap
```diff
@@ -3,7 +3,7 @@
         "content": [
             "[deseq2_treatment_mCherry_hND6_consensus_estimate_decoupler.tsv, deseq2_treatment_mCherry_hND6_mlm_estimate_decoupler.tsv]",
             "[deseq2_treatment_mCherry_hND6_consensus_pvals_decoupler.tsv, deseq2_treatment_mCherry_hND6_mlm_pvals_decoupler.tsv]",
-            "[deseq2_treatment_mCherry_hND6_consensus_estimate_decoupler_plot.png, deseq2_treatment_mCherry_hND6_consensus_pvals_decoupler_plot.png, deseq2_treatment_mCherry_hND6_mlm_estimate_decoupler_plot.png, deseq2_treatment_mCherry_hND6_mlm_pvals_decoupler_plot.png]",
```
atrigila, line 6: Add a note in the PR that the pvals  outputs are no longer generated.

antoniasaracco:
P-value plot files (`*pvals_decoupler_plot.png`) are no longer generated

antoniasaracco:
@pinin4fjords, Hi! I've implemented in this PR the necessary changes to integrate decoupler into the differencial abundance pipeline as discussed. Please let me know if you have any feedback or if any adjustment are needed , thanks!

subworkflows/nf-core/differential_functional_enrichment/main.nf
```diff
@@ -116,6 +136,14 @@ workflow DIFFERENTIAL_FUNCTIONAL_ENRICHMENT {
...
         CUSTOM_TABULARTOGSEACHIP.out.chip.first()
     )
+    // ----------------------------------------------------
+    // Perform enrichment analysis with DECOUPLER
+    // ----------------------------------------------------
+    DECOUPLER(
+        ch_input_for_decoupler,
+        ch_decoupler_network,
+        ch_featuresheet.map{ meta, features, features_id, features_symbol -> features }
+    )
```
pinin4fjords:
```suggestion
    DECOUPLER(
        ch_input_for_other.input.filter{ it[0].functional_method == 'decoupler' },
        ch_input_for_other.genesets.filter{ it[0].functional_method == 'decoupler'},
        ch_featuresheet.map{ meta, features, features_id, features_symbol -> features }
    )
```

If you supply the network in the 'genesets' slot as suggested above, and don't mess with the meta (don't think you need to), then you should be able to get to something like this.

modules/nf-core/decoupler/decoupler/main.nf
```diff
@@ -10,12 +9,12 @@ process DECOUPLER {
     input:
     tuple val(meta), path(mat)
     path(net)
-    path(gtf)
+    path(annot)
```
pinin4fjords: All the inputs should have metas

```suggestion
    tuple val(meta3), path(annot)
```

subworkflows/nf-core/differential_functional_enrichment/main.nf
```diff
@@ -80,6 +81,25 @@ workflow DIFFERENTIAL_FUNCTIONAL_ENRICHMENT {
...
+                id: meta.id,
+                method_differential: meta.params.differential_method
+            ]
+            [new_meta, differential_results]
+        }
+    ch_decoupler_network = ch_input_for_other.input
+        .filter{ it[0].functional_method == 'decoupler' }
+        .map { meta, differential_results ->
+            meta.params.decoupler_network ? file(meta.params.decoupler_network, checkIfExists: true) : []
+        }
```
pinin4fjords:
```suggestion
```

You really shouldn't need to do any of that. This subworkflow is designed to supply files through a consistent interface, so that we can run things with consistent patterns. We rely on the calling workflow to build the channels correctly, so you shouldn't have to be mucking about with the meta, for example.

You can also just pass the network in the `genesets_file slot` of the ch_inputs channel, leaving some elements empty as the comment above illustrates. I realise it's not quite the same thing, but it's analagous in this context (we can look at renaming things later).

subworkflows/nf-core/differential_functional_enrichment/main.nf
pinin4fjords: @suzannejin I don't remember why we don't have the features going through the multiMap, but we probably should (not for this PR).

suzannejin: I think it was simply because the other modules so far (gprofiler2, grea) do not need to receive features.
But I agree, we should add it to ch_input_other too.

antoniasaracco:
@pinin4fjords, Hi! I've implemented in this PR the requested changes.  Please let me know if you have any feedback or if any adjustment are needed , thanks!

modules/nf-core/decoupler/decoupler/templates/decoupler.py
```diff
@@ -43,50 +43,30 @@ def parse_ext_args(args_string: str):
...
 
 if parsed_args.ensembl_ids.upper() == "TRUE":
     try:
-        gene_mapping = parse_gtf("${gtf}")
+        if not os.path.exists("${annot}"):
+            raise FileNotFoundError(f"Annotation file not found: ${annot}")
+
+        annot_df = pd.read_csv("${annot}", sep="\t")
+
+        required_cols = {"gene_id", "gene_name"}
```
suzannejin: I thought that decoupler can be used for any omics data? 
If so, would you still prefer to hard code the feature id and feature name columns?

modules/nf-core/decoupler/decoupler/tests/main.nf.test.snap
```diff
@@ -52,7 +52,7 @@
                         {
                             "id": "test"
                         },
-                        "null_test_decoupler_plot.png:md5,d41d8cd98f00b204e9800998ecf8427e"
+                        "null_test_estimate_decoupler_plot.png:md5,d41d8cd98f00b204e9800998ecf8427e"
```
suzannejin: why the prefix of the output starts with "null_test"?

subworkflows/nf-core/differential_functional_enrichment/main.nf
```diff
@@ -116,6 +118,15 @@ workflow DIFFERENTIAL_FUNCTIONAL_ENRICHMENT {
...
         CUSTOM_TABULARTOGSEACHIP.out.chip.first()
     )
+    // ----------------------------------------------------
+    // Perform enrichment analysis with DECOUPLER
+    // ----------------------------------------------------
+
+    DECOUPLER(
+        ch_input_for_other.input.filter{ it[0].functional_method == 'decoupler' },
+        ch_input_for_other.genesets.filter{ it[0].functional_method == 'decoupler'},
+        ch_featuresheet.map{ meta, features, features_id, features_symbol -> [meta, features] }
```
suzannejin: ch_input_for_other and ch_featuresheet may not have the same number of entries.
I would suggest you to create a ch_input_for_other.features.

subworkflows/nf-core/differential_functional_enrichment/tests/main.nf.test
```diff
@@ -463,6 +465,67 @@ nextflow_workflow {
...
+                    file("https://github.com/nf-core/test-datasets/raw/differentialabundance/modules_testdata/treatment_mCherry_hND6_.deseq2.results.tsv", checkIfExists: true),
+                    file("https://github.com/nf-core/test-datasets/raw/differentialabundance/modules_testdata/progeny/mouse_network.tsv", checkIfExists: true),  // network file as genesets
+                    [],  // background (empty for decoupler)
+                    'decoupler'  // method
+                ])
+
+                ch_contrasts = Channel.of([[], [], [], [], [], []])
+                ch_samplesheet = Channel.of([[], []])
+                ch_featuresheet = Channel.of([
+                    [],
```
suzannejin: to make ch featuresheet properly combine with the rest of part of ch_input_for_other, you would need to parse the meta from ch_input as part of ch_featuresheet

subworkflows/nf-core/differential_functional_enrichment/tests/main.nf.test.snap
```diff
@@ -353,6 +353,45 @@
         },
         "timestamp": "2025-07-21T14:58:54.929796085"
     },
+    "deseq2 + decoupler - mouse": {
+        "content": [
+            [
+                [
+                    "treatment_mCherry_hND6",
+                    [
+                        "deseq2_treatment_mCherry_hND6_consensus_estimate_decoupler.tsv",
```
suzannejin, line 362: don't you want to also include md5 snapshot matching? Maybe at least for the tsv output, cuz the png tends to give issues for md5.

antoniasaracco: Hi @suzannejin, 
I didn’t include it because the TSV hash was stable across local Docker, Singularity and other profiles runs, but changed when the tests ran in GitHub Actions.  To keep the test consistent across environments I switched to value-based checks on the TSV contents, asserting headers and representative rows:
image

Please let me know what you think!

Thanks!

antoniasaracco:
@suzannejin Hi! I've implemented in this PR the requested changes. Please let me know if you have any feedback or if any adjustment are needed , thanks!

modules/nf-core/decoupler/decoupler/main.nf
```diff
@@ -25,9 +25,9 @@ process DECOUPLER {
 
     stub:
     """
-    touch ${task.ext.prefix}_estimate_decoupler.tsv
-    touch ${task.ext.prefix}_pvals_decoupler.tsv
-    touch ${task.ext.prefix}_estimate_decoupler_plot.png
+    touch deseq2_estimate_decoupler.tsv
```
suzannejin, line 28: why 'deseq2'?

modules/nf-core/decoupler/decoupler/meta.yml
```diff
@@ -66,6 +66,18 @@ input:
         required: false
         ontologies:
           - edam: http://edamontology.org/format_3475 # TSV
+    - features_id:
```
suzannejin: As far as I understood, the convention in nf-core is to provide only the compulsory parameters as input for the module. The rest can be still be provided through ext.args.
You can check how this is done in the DESEQ2 module, and how it is called by the differentialabundance pipeline.

modules/nf-core/decoupler/decoupler/tests/main.nf.test.snap
```diff
@@ -36,23 +36,23 @@
                         {
                             "id": "test"
                         },
-                        "null_test_estimate_decoupler.tsv:md5,d41d8cd98f00b204e9800998ecf8427e"
+                        "deseq2_estimate_decoupler.tsv:md5,d41d8cd98f00b204e9800998ecf8427e"
```
suzannejin, line 39: why these files are all empty? Are you stub running?

subworkflows/nf-core/differential_functional_enrichment/main.nf
```diff
@@ -49,8 +49,23 @@ workflow DIFFERENTIAL_FUNCTIONAL_ENRICHMENT {
                 [ meta_with_method, genesets ]
             background:
                 [ meta_with_method, background ]
-        }
 
+        }
+    
+    ch_input_for_decoupler = ch_input
```
suzannejin: maybe i was not clear before, but I think you do not need to specifically create a ch_input_for_decoupler that is essentially doing all the stuff in ch_input_for_other but with featuresheet.
I would suggest you to simply modify ch_input_for_other by adding featuresheet

suzannejin: Basically what you have just did, but for the ch_input_for_other. The reason is that this channel can then be used by other processes, and no code deduplication needed
```suggestion
    ch_input_for_other = ch_input
        .join(ch_featuresheet, by: 0)
        .multiMap {
            meta, file, genesets, background, method, features_sheet, features_id, features_symbol ->
            def meta_with_method = meta + [ 'functional_method': method ]
            input:
                [ meta_with_method, file ]
            genesets:
                [ meta_with_method, genesets ]
            background:
                [ meta_with_method, background ]
            features:
                [ meta_with_method, features_sheet, features_id, features_symbol]
```

subworkflows/nf-core/differential_functional_enrichment/tests/main.nf.test.snap
suzannejin: hi @antoniasaracco . Sorry I have just seen this comment now.
Did you check why this inconsistency happens? Maybe related to long decimals? 
If so, an alternative that other modules (deseq2, etc) do is to provide a rounding flag at testing time

modules/nf-core/decoupler/decoupler/main.nf
antoniasaracco: @suzannejin  In the stub sections I used deseq2 as a placeholder so the filenames and metadata do not contain null. It is one of the valid values that real runs can produce. Outside of stub mode this field is set by the actual method used, so it may be deseq2, limma, etc. The placeholder is only to keep the stub outputs deterministic.

subworkflows/nf-core/differential_functional_enrichment/tests/main.nf.test.snap
antoniasaracco: hi @suzannejin I could not verify the inconsistency. It only shows up in CI and I cannot reproduce it locally, so I cannot confirm whether long decimals are the reason. Locally the TSV content and md5 are stable across repeated runs with Docker and Singularity. To avoid this CI-only mismatch the tests now compare specific table values instead of a snapshot md5.

antoniasaracco:
@suzannejin Hi! I've implemented in this PR the requested changes. Please let me know if you have any feedback or if any adjustment are needed , thanks!

subworkflows/nf-core/differential_functional_enrichment/main.nf
```diff
@@ -40,19 +40,6 @@ workflow DIFFERENTIAL_FUNCTIONAL_ENRICHMENT {
     // Also, reorganize the structure to match them with the modules' input organization
 
     ch_input_for_other = ch_input
```
suzannejin, line 41: thank you! this is great :)

subworkflows/nf-core/differential_functional_enrichment/tests/main.nf.test
```diff
@@ -41,7 +41,12 @@ nextflow_workflow {
                 input[0] = ch_input
                 input[1] = Channel.of([[], [], [], [], [], []])
                 input[2] = Channel.of([[], []])
-                input[3] = Channel.of([[], [], [], []])
+                input[3] = Channel.of([
+                    ['id':'Condition_genotype_WT_KO', 'variable':'Condition genotype', 'reference':'WT', 'target':'KO', 'blocking':'batch'],
+                    [],  //
```
suzannejin, line 46: I don't understand what is this for? Do you necessarily need to provide such a ch_featuresheet with empty stuff?

suzannejin: is it because of join(ch_featureshseet, by:0)?
If so, have you tried to use remainder:true?
I recall I had to do that for the same reasons in the subworkflow "abundance_differential_filter"

antoniasaracco: yes, it's becuase of the join. I'll try that so there is no need to modify the other tests. Thanks for the suggestion!

suzannejin:
> hi @suzannejin I could not verify the inconsistency. It only shows up in CI and I cannot reproduce it locally, so I cannot confirm whether long decimals are the reason. Locally the TSV content and md5 are stable across repeated runs with Docker and Singularity. To avoid this CI-only mismatch the tests now compare specific table values instead of a snapshot md5.


Hi @antoniasaracco, have you tried it with Conda locally? For me, the long decimal issues also happen frequently with Conda.

suzannejin:
Hello @antoniasaracco, sorry for the delay, and thank you for addressing my comments!
I have just left two more related to the questions I asked before.
Once these are solved I think it should be done :)

subworkflows/nf-core/differential_functional_enrichment/tests/main.nf.test
antoniasaracco: Hi @suzannejin,  I added remainder: true to the join:

ch_input.join(ch_featuresheet, by: 0, remainder: true)

And updated the multiMap to handle variable tuple sizes since remainder: true creates tuples with different lengths (6 elements when no featuresheet matches, 8 when it does). 

Please let me know what you think!

antoniasaracco:
Hi @suzannejin , I've implemented the requested changes. Regarding the decoupler test, I also tried with conda locally - tested all profiles (docker, singularity, conda) and couldn't reproduce the inconsistency. The issue seems to be CI-specific.

Please let me know if you have any feedback or if any adjustment are needed , thanks!

subworkflows/nf-core/differential_functional_enrichment/main.nf
```diff
@@ -39,17 +40,23 @@ workflow DIFFERENTIAL_FUNCTIONAL_ENRICHMENT {
...
-        .multiMap {
-            meta, file, genesets, background, method ->
+        .join(ch_featuresheet, by: 0, remainder: true)
+        .multiMap { tuple ->
+            def (meta, file, genesets, background, method) = tuple[0..4]
+            def features_sheet = tuple.size() > 5 ? tuple[5] : null
+            def features_id = tuple.size() > 6 ? tuple[6] : ''
+            def features_symbol = tuple.size() > 7 ? tuple[7] : ''
```
pinin4fjords:
```suggestion
        .join(ch_featuresheet)
        .multiMap { meta, file, genesets, background, method, features_sheet, features_id, features_symbol ->
```

I think this is over-engineered. Let's just assume the workflow is being called as differentialbundance is calling it- with a predictable feature sheet structure.

subworkflows/nf-core/differential_functional_enrichment/main.nf
```diff
@@ -39,17 +40,23 @@ workflow DIFFERENTIAL_FUNCTIONAL_ENRICHMENT {
...
+            
             def meta_with_method = meta + [ 'functional_method': method ]
             input:
                 [ meta_with_method, file ]
             genesets:
                 [ meta_with_method, genesets ]
             background:
                 [ meta_with_method, background ]
+            features:
+                [ meta_with_method, features_sheet ?: [], features_id, features_symbol]
```
pinin4fjords:
```suggestion
                [ meta_with_method, features_sheet, features_id, features_symbol]
```
similarly here

pinin4fjords (changes requested):
Sorry, but let's keep things really simple and assume the feature sheet is supplied correct as diff. ab does. 

I also have other ideas for simplifying this subworkflow, but probably something for a separate PR

pinin4fjords (approved):
Much neater now, thank you

PR #8974, opened 2025-08-30
8705 picardcollecthsmetrics fails for gzipped genomes

### LINKED ISSUE #8705: picard/collecthsmetrics fails for gzipped genomes

### Description of the bug

`picard CollectHsMetrics` requires a `.gzi` file when running on a gzipped fasta. This is not currently an input option for the process. 

You'll get an error in the process that looks like:
```
Exception in thread "main" java.lang.UnsupportedOperationException: Index does not appear to exist for GRCh38_GIABv3_no_alt_analysis_set_maskedGRC_decoys_MAP2K3_KMT2C_KCNJ18.fasta.gz.  samtools faidx can be used to create an index
```

Adding an input of the `.gzi` file (which is also output by samtools faidx when run on a `fasta.gz`) resolves this problem.

```console

```

### Relevant files

_No response_

### System information

_No response_

shivangi-7:
## PR checklist

Closes #8705 

  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/artic/aligntrim/environment.yml
+++ b/modules/nf-core/artic/aligntrim/environment.yml
@@ -4,5 +4,5 @@ channels:
   - conda-forge
   - bioconda
 dependencies:
-  - "bioconda::align_trim=1.0.1"
+  - "bioconda::align_trim=1.0.2"
   - "bioconda::samtools=1.22.1"
--- a/modules/nf-core/artic/aligntrim/main.nf
+++ b/modules/nf-core/artic/aligntrim/main.nf
@@ -4,8 +4,8 @@ process ARTIC_ALIGNTRIM {
 
     conda "${moduleDir}/environment.yml"
     container "${workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container
-        ? 'oras://community.wave.seqera.io/library/align_trim_samtools:3f69e9e238df1a73'
-        : 'community.wave.seqera.io/library/align_trim_samtools:e329909729947df9'}"
+        ? 'oras://community.wave.seqera.io/library/align_trim_samtools:ae17186e78142d18'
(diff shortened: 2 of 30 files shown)


modules/nf-core/picard/collecthsmetrics/main.nf
```diff
@@ -9,9 +9,10 @@ process PICARD_COLLECTHSMETRICS {
 
     input:
     tuple val(meta), path(bam), path(bai), path(bait_intervals, stageAs: "baits/*"), path(target_intervals, stageAs: 'targets/*')
-    tuple val(meta2), path(fasta)
-    tuple val(meta3), path(fai)
-    tuple val(meta4), path(dict)
+    tuple val(meta2), path(ref)
+    tuple val(meta3), path(ref_fai)
+    tuple val(meta4), path(ref_dict)
+    tuple val(meta5), path(ref_gzi) // ref_gzi only required if reference is gzipped
```
akaviaLab, line 15: This might be a silly question, but don't you need to set up some code if ref_gzi is not provided? Right now, in your tests you're providing it as empty, but it is there. If there are pipelines/workflows that use collecthsmetrics and don't provide even an empty ref_gzi, wouldn't they fail after your changes?

Or am I missing something?

khersameesh24: Hi @akaviaLab , for optional inputs, we usually provide empty channels, so it should be fine. The pipelines using collecthsmetrics will fail only if they update the module. Even after updating, the nf-core linter will flag changes to the module and subworkflow, so this should not be an issue.

akaviaLab:
Thank you for clarifying!

shivangi-7:
Thank you for the review, everybody!
Going to merge now :)

PR #8984, opened 2025-09-02
update utils_nfschema_plugin to 2.5.0

nvnieuwk:


SPPearce:
Does this need updating to 2.5.1?

subworkflows/nf-core/utils_nfschema_plugin/tests/nextflow.config
```diff
@@ -1,5 +1,5 @@
 plugins {
-    id "nf-schema@2.4.2"
+    id "nf-schema@2.5.0"
```
nvnieuwk:
```suggestion
    id "nf-schema@2.5.1"
```

PR #8993, opened 2025-09-03
Update BWAMEM2_INDEX

BioWilko:
The previous memory declaration is causing crashes for small references (e.g. viruses) and k8s executor since pods with 5M of RAM do not play nice with the cluster....

This declaration increases on retries and puts a memory floor of 100M in!

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/bwamem2/index/main.nf
+++ b/modules/nf-core/bwamem2/index/main.nf
@@ -2,7 +2,7 @@ process BWAMEM2_INDEX {
     tag "$fasta"
     // NOTE Requires 28N GB memory where N is the size of the reference sequence
     // source: https://github.com/bwa-mem2/bwa-mem2/issues/9
-    memory { 28.B * fasta.size() }
+    memory { ((128.B * fasta.size()) * task.attempt) >= 100.MB ? (128.B * fasta.size()) * task.attempt : 100.MB * task.attempt }
 
     conda "${moduleDir}/environment.yml"
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?


maxulysse:
I managed to circumvent such issues by forcing the memory to a specific value.
But nice to have a better patch, I've been meaning to do so

```
process {
    withName:BWAMEM2_INDEX {
        memory = { 6.GB }
    }
}
```

BioWilko:
Yep that's what I did too to get it to work but ideally it should work out of the box

modules/nf-core/bwamem2/index/main.nf
```diff
@@ -2,7 +2,7 @@ process BWAMEM2_INDEX {
     tag "$fasta"
     // NOTE Requires 28N GB memory where N is the size of the reference sequence
     // source: https://github.com/bwa-mem2/bwa-mem2/issues/9
-    memory { 28.B * fasta.size() }
+    memory { ((128.B * fasta.size()) * task.attempt) >= 100.MB ? (128.B * fasta.size()) * task.attempt : 100.MB * task.attempt }
```
edmundmiller: This looks great. 
I'm 100% for the memory floor.

Only question is why 128 instead of 28. I was just going off this comment: https://github.com/bwa-mem2/bwa-mem2/issues/9#issuecomment-575514285

modules/nf-core/bwamem2/index/main.nf
```diff
@@ -1,8 +1,8 @@
 process BWAMEM2_INDEX {
     tag "$fasta"
-    // NOTE Requires 28N GB memory where N is the size of the reference sequence
+    // NOTE Requires 128N * number of task attempts GB memory where N is the size of the reference sequence with a floor of 100M
     // source: https://github.com/bwa-mem2/bwa-mem2/issues/9
-    memory { 28.B * fasta.size() }
+    memory { ((128.B * fasta.size()) * task.attempt) >= 100.MB ? (128.B * fasta.size()) * task.attempt : 100.MB * task.attempt }
```
muffato: In my pipelines I do something like this
```suggestion
    memory { 280.MB * Math.ceil(fasta.size() / 10000000)}
```
This would preserve the 28N rule, but apply it by steps of 280 MB RAM.
Also `Math.ceil` ensures things are rounded up, and the value before the `*` operator becomes the minimum value sent to the job scheduler, so >100 MB in this case.

You can still throw a `* task.attempt` in if you want extra safety.

edmundmiller (approved):
LGTM!

maxulysse:
I'll investigate why the versions are different

maxulysse:
> I'll investigate why the versions are different

Sorry, started looking at this without coffee, it's ABRA2 tests that are failing due to mismatched versions

modules/nf-core/bwamem2/index/main.nf
BioWilko: I used that due to running into OOM errors suggesting 28N isn't enough in all situations

modules/nf-core/bwamem2/index/main.nf
maxulysse: I love this solution @muffato

BioWilko: Yeah, super elegant

PR #8998, opened 2025-09-04
Dp24 blobtk

DLBPointon:
blobtk is a rust program for the manipulation of blobtooldir data output from the blobtoolkit pipeline.

Tracked here as well where we have done most of the work https://github.com/sanger-tol/nf-core-modules/pull/59

And it has been implemented here: https://github.com/sanger-tol/genomenote/pull/205

Only issue might be the blobtk_plot module which has an enforced errorStrategy explained in the module with:
    // A somewhat nuclear option
    // Linked to issue https://github.com/sanger-tol/genomenote/issues/184
    // Depending on the blob dataset in use, the grid option may not
    // work at all. This is down to the version of blobtoolkit used to
    // generate the blob.
    // Adding a check would overly complicate the module so for now
    // we can ignore errors, with the *assumption* it would only kill
    // runs in which the blobdir doesn't have the right data

### DIFF
--- a/modules/nf-core/blobtk/depth/meta.yml
+++ b/modules/nf-core/blobtk/depth/meta.yml
@@ -31,7 +31,7 @@ input:
         pattern: "*.bam"
         ontologies:
           - edam: http://edamontology.org/format_2572 # BAM
-    - bai:
+    - index:
         type: file
         description: |
           Algined BAM index file.
--- a/modules/nf-core/blobtk/depth/tests/main.nf.test.snap
+++ b/modules/nf-core/blobtk/depth/tests/main.nf.test.snap
@@ -1,9 +1,5 @@
 {
-<<<<<<< HEAD
     "bacteroides fragilis - bam and bai - stub": {
-=======
-    "Cantharis rufa - bam and bai - stub": {
->>>>>>> master
         "content": [
             {
                 "0": [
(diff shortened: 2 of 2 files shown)


DLBPointon:
Hi @charles-plessy thanks for the review!

> * The module also accepts CSI indexes. Maybe rename `bam` to `index`?

That is a fair point, I'll get that changed. Edit: Done now.

> * I do not understand why you need a separate channel for files and for URLs. Nextflow is able to handle both of them transparently.

I tested it as one channel originally, a blobdir url isn't an url to a blobdir but rather a json which describes the blobdir. Where as a local path would be the entirety of blobdir (dozens of files). You quite literally can't use them in the same way. So they are for all intents and purposes two different things. The URL also needs to be the raw string not what Nextflow thinks it should be resolving, for the tool to work.

> * If `blobtk plot --shape grid` may fail depending on the Blob Dir, and if it is an expected behaviour of the tool, then I think that it is indeed a good idea to ignore failures. Mabye you can parse `task.ext.args` to only change the error strategy when `--shape grid` is present?

That is an interesting point, although we know it is `--shape grid` at the moment, this could be more options in the future and I wouldn't want to pin ourselves into a corner with adding a conditional like that. For now, we *know* if something errors or isn't produced then the data simply lacks enough data for that option.

PR #9004, opened 2025-09-05
Collect more genome statistics in the `last` module

charles-plessy:
This PR collects more genome statistics in three `last` modules:

 - Target and query genome size and number of sequences in `last/lastal` and `last/split`, for MultiQC reporting and downstream applications.
 - Target genome sequence name and length in GFF output in `last/mafconvert`, for downstream applications.

In addition I had to add one commit enforcing CRAM format 3.0 to prevent some tests from failing. CRAM format 3.1 became default in samtools 1.22, while the modules are supposed to still use samtools 1.21 but anyway, the commit fixed the failure…

----

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/.github/workflows/nf-test-gpu.yml
+++ b/.github/workflows/nf-test-gpu.yml
@@ -30,7 +30,7 @@ env:
   NXF_SINGULARITY_CACHEDIR: ${{ github.workspace }}/.singularity
   NXF_SINGULARITY_LIBRARYDIR: ${{ github.workspace }}/.singularity
   # renovate: datasource=github-releases depName=nextflow-io/nextflow versioning=semver
-  NXF_VER: "24.10.2"
+  NXF_VER: "25.04.0"
 
 jobs:
   nf-test-changes:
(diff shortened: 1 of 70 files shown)


DLBPointon:
Hi @charles-plessy 

The only real difference between CRAM3.0 and CRAM3.1 is the compression isn't it? https://pubmed.ncbi.nlm.nih.gov/34999766/

So in order to make sure the module stays upto date, would it make sense to add a test for the 3.1 format alongside the current cram test and move the version assignment from `samtools sort -O cram,version=3.0 -o ${prefix}.cram` to an arg?

charles-plessy:
Hi @DLBPointon , the problem is that https://nvnieuwk.github.io/nft-bam/ which is used by `nf-test` does not support CRAM 3.1 as `htsjdk` does not. So I need to either wait for the test toolchain to be updated, or to change the way the output files are tested... Otherwise, I am all fine with defaulting on samtools defaults :)

DLBPointon:
> Hi @DLBPointon , the problem is that https://nvnieuwk.github.io/nft-bam/ which is used by `nf-test` does not support CRAM 3.1 as `htsjdk` does not. So I need to either wait for the test toolchain to be updated, or to change the way the output files are tested... Otherwise, I am all fine with defaulting on samtools defaults :)

Ah ok sorry I missed that bit!

Apologies, in that case i'd be happy to merge.

PR #9024, opened 2025-09-10
Update to correct malformed version

DLBPointon:
Correction to version information in blobtk_plots

A tab got introduced into the version array, which doesn't kill linting but does kill the version collection in a pipeline.


DLBPointon:
Thanks @maxulysse

PR #9043, opened 2025-09-11
Gzi input for mutect2 

### LINKED ISSUE #8789: GATK4_MUTECT2 fails for gzipped references

### Description of the bug

`gatk Mutect2` requires a `.gzi` file when running on a gzipped fasta. This is not currently an input option for the process. 

You'll get an error in the process that looks like:
```
 A USER ERROR has occurred: Fasta bgzip index file file://chr22.fa.gz.gzi for reference file://chr22.fa.gz does not exist. A gzi index can be created using bgzip.
```

Adding an input of the `.gzi` file (which is also output by samtools faidx when run on a `fasta.gz`) resolves this problem.

```console

```

### Relevant files

_No response_

### System information

_No response_

mignardise:
# nf-core/modules pull request

## PR checklist

Closes #8789  

- [✓ ] gzi path input for fasta, added more meta options for fasta, and test using a modified mutect mitochondria mode due to fast testing
- [ ✓ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ✓ ] Remove all TODO statements.
- [✓  ] Emit the `versions.yml` file.
- [✓  ] Follow the naming conventions.
- [ ✓ ] Follow the parameters requirements.
- [ ✓] Follow the input/output options guidelines.
- [ ✓] #Add a resource `bug`
- [ ✓ ] Use BioConda and BioContainers if possible to fulfil software requirements.
  - For modules:
    - [✓ ] `nf-core modules test <MODULE> --profile docker`

### DIFF
--- a/modules/nf-core/slamdunk/map/environment.yml
+++ b/modules/nf-core/slamdunk/map/environment.yml
@@ -1,0 +1,7 @@
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/environment-schema.json
+channels:
+  - conda-forge
+  - bioconda
+dependencies:
+  - "bioconda::slamdunk=0.4.3"
--- a/modules/nf-core/slamdunk/map/main.nf
+++ b/modules/nf-core/slamdunk/map/main.nf
@@ -1,0 +1,60 @@
+process SLAMDUNK_MAP {
+    tag "${meta.id}"
+    label 'process_high'
+
+    conda "${moduleDir}/environment.yml"
+    container "${workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container
+        ? 'https://depot.galaxyproject.org/singularity/slamdunk:0.4.3--py_0'
+        : 'biocontainers/slamdunk:0.4.3--py_0'}"
+
+    input:
+    tuple val(meta), path(input)
+    tuple val(meta2), path(fasta)
+
+    output:
+    tuple val(meta), path("*.bam"), emit: bam
+    path "versions.yml", emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+
+    if ("${input}" == "${prefix}.bam") {
+        error("Input and output names of bam files are the same, set prefix in module configuration to disambiguate!")
+    }
+    """
+    slamdunk \\
+        map \\
+        -r ${fasta} \\
+        -t ${task.cpus} \\
+        -o outputs \\
+        ${args} \\
+        ${input}
+
+    mv outputs/*.bam ${prefix}.bam
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        slamdunk: \$(echo \$(slamdunk --version) | sed 's/^slamdunk //')
+    END_VERSIONS
+    """
+
+    stub:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+
+    if ("${input}" == "${prefix}.bam") {
+        error("Input and output names of bam files are the same, set prefix in module configuration to disambiguate!")
+    }
+    """
+    touch ${prefix}.bam
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        slamdunk: \$(echo \$(slamdunk --version) | sed 's/^slamdunk //')
+    END_VERSIONS
+    """
+}
--- a/modules/nf-core/slamdunk/map/meta.yml
+++ b/modules/nf-core/slamdunk/map/meta.yml
@@ -1,0 +1,66 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/meta-schema.json
+name: "slamdunk_map"
+description: Slamdunk read mapping using NextGenMap’s SLAMSeq alignment settings.
+keywords:
+  - slamseq
+  - rna-seq
+  - mapping
+tools:
+  - "slamdunk":
+      description: "Slamdunk is a software tool for SLAMseq data analysis that performs mapping, filtering, SNP calling, and quantification of metabolic RNA labeling experiments."
+      homepage: "http://t-neumann.github.io/slamdunk/docs.html"
+      documentation: "http://t-neumann.github.io/slamdunk/docs.html"
+      tool_dev_url: "https://github.com/t-neumann/slamdunk"
+      doi: "10.1186/s12859-019-2849-7"
+      licence: ["GNU Affero General Public License v3 (AGPL v3)"]
+      identifier: biotools:slamdunk
+
+input:
+  - - meta:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. `[ id:'sample1' ]`
+    - input:
+        type: file
+        description: Input FASTQ or BAM file(s) for SLAMseq analysis
+        pattern: "*.{bam,fastq,fq,fastq.gz,fq.gz}"
+        ontologies:
+          - edam: "http://edamontology.org/format_1930" # FASTQ
+  - - meta2:
+        type: map
+        description: |
+          Groovy Map containing reference information
+          e.g. `[ id:'reference' ]`
+    - fasta:
+        type: file
+        description: Reference genome FASTA file
+        pattern: "*.{fasta,fa,fas,fna}"
+        ontologies:
+          - edam: "http://edamontology.org/format_1929" # FASTA
+
+output:
+  bam:
+    - - meta:
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. `[ id:'sample1' ]`
+      - "*.bam":
+          type: file
+          description: Mapped BAM file
+          pattern: "*.bam"
+          ontologies:
+            - edam: "http://edamontology.org/format_2572" # BAM
+  versions:
+    - "versions.yml":
+        type: file
+        description: File containing software versions
+        pattern: "versions.yml"
+        ontologies:
+          - edam: "http://edamontology.org/format_3750" # YAML
+
+authors:
+  - "@haidyi"
+maintainers:
+  - "@haidyi"
--- a/modules/nf-core/slamdunk/map/tests/main.nf.test
+++ b/modules/nf-core/slamdunk/map/tests/main.nf.test
@@ -1,0 +1,129 @@
+nextflow_process {
+
+    name "Test Process SLAMDUNK_MAP"
+    script "../main.nf"
+    process "SLAMDUNK_MAP"
+
+    tag "modules"
+    tag "modules_nfcore"
+    tag "slamdunk"
+    tag "slamdunk/map"
+
+    test("slamdunk - map - fastq") {
+
+        when {
+            process {
+                """
+                input[0] = [
+                    [ id:'test', single_end:true ], // meta map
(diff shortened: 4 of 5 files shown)


famosab:
You can join the org by jumping on slack and writing in #github-invitations (https://nf-co.re/join)

.idea/.gitignore
famosab, line 1: please remove all the .idea/... files from the PR :)

modules/nf-core/gatk4/mutect2/main.nf
famosab, line 1: Logic wise I feel like this index (which is an alternative to fai right?) should go below or even in the same tuple as the fai index :)

mignardise:
@famosab gotcha thank you for the feedback will implement those changes soon!

modules/nf-core/gatk4/mutect2/main.nf
```diff
@@ -9,7 +9,7 @@ process GATK4_MUTECT2 {
 
     input:
     tuple val(meta), path(input), path(input_index), path(intervals)
-    tuple val(meta2), path(fasta)
+    tuple val(meta2), path(fasta), path(gzi)
     tuple val(meta3), path(fai)
```
famosab: I know that I suggested to put this here somewhere. But I meant this more as an alternative to the .fai index. Because as I understand it people would either give fasta + fai oder fasta.gz + gzi, right?

modules/nf-core/gatk4/mutect2/meta.yml
famosab, line 1: FRom the changes here I am wondering if we even need a new input channel for .gzi files or whether we should just rename fai to index or fasta_index and then check via the file extension what we are working with.

famosab:
I think it seems like we do not need another input channel but rather we add a test where we supply fasta.gz + .gzi file and use the existing input channels for the respective files. And then we only update the meta :)

modules/nf-core/gatk4/mutect2/main.nf
```diff
@@ -9,7 +9,7 @@ process GATK4_MUTECT2 {
 
     input:
     tuple val(meta), path(input), path(input_index), path(intervals)
-    tuple val(meta2), path(fasta)
+    tuple val(meta2), path(fasta), path(gzi)
```
famosab:
```suggestion
    tuple val(meta2), path(fasta)
```

modules/nf-core/gatk4/mutect2/main.nf
famosab:
```suggestion
    tuple val(meta3), path(fasta_index)
```

modules/nf-core/gatk4/mutect2/meta.yml
```diff
@@ -46,7 +46,12 @@ input:
     - fasta:
         type: file
         description: The reference fasta file
-        pattern: "*.fasta"
+        pattern: "*.{fasta,fasta.gz}"
+        ontologies: []
+    - gzi:
+        type: file
+        description: Index of reference fasta file
+        pattern: "*.{fasta.gz.gzi,fasta.gz.gzi}"
```
famosab:
```suggestion
```

modules/nf-core/gatk4/mutect2/meta.yml
```diff
@@ -46,7 +46,12 @@ input:
...
         ontologies: []
```
famosab:
```suggestion
```

modules/nf-core/gatk4/mutect2/meta.yml
```diff
@@ -56,7 +61,7 @@ input:
     - fai:
```
famosab, line 56:
```suggestion
    - fasta_index:
```

modules/nf-core/gatk4/mutect2/meta.yml
```diff
@@ -56,7 +61,7 @@ input:
...
         type: file
         description: Index of reference fasta file
```
famosab, line 58:
```suggestion
        description: Index of reference fasta file, .fai for .fasta files and .gzi for .fasta.gz files
```

modules/nf-core/gatk4/mutect2/meta.yml
```diff
@@ -56,7 +61,7 @@ input:
...
-        pattern: "*.fasta.fai"
+        pattern: "*.{fasta.fai,fasta.fai.gz}"
```
famosab, line 59:
```suggestion
        pattern: "*.{fasta.fai,fasta.fai.gz,fasta.gz.gzi}"
```

modules/nf-core/gatk4/mutect2/tests/main.nf.test
```diff
@@ -332,6 +338,57 @@ nextflow_process {
...
+                [ file(params.modules_testdata_base_path + 'genomics/homo_sapiens/genome/genome.bed', checkIfExists: true)]
+            ]
+            input[1] = [
+                [ id:'genome' ],
+                [ file(params.modules_testdata_base_path + 'genomics/homo_sapiens/genome/genome.fasta.gz', checkIfExists: true)],
+                [ file(params.modules_testdata_base_path + 'genomics/homo_sapiens/genome/genome.fasta.gz.gzi', checkIfExists: true)]
+            ]
+            input[2] = [
+                [ id:'genome' ],
+                file(params.modules_testdata_base_path + 'genomics/homo_sapiens/genome/genome.fasta.gz.fai', checkIfExists: true)
```
famosab:
```suggestion
                [ file(params.modules_testdata_base_path + 'genomics/homo_sapiens/genome/genome.fasta.gz', checkIfExists: true)]
            ]
            input[2] = [
                [ id:'genome' ],
                 file(params.modules_testdata_base_path + 'genomics/homo_sapiens/genome/genome.fasta.gz.gzi', checkIfExists: true)
```

famosab:
On the second thought I am not completely sure, seems like one still needs both the .fai and the .gzi if we hand bgzip compressed fasta files. But in that case I would say that the .gzi goes alongside the .fai

modules/nf-core/gatk4/mutect2/main.nf
famosab:
```suggestion
    tuple val(meta3), path(fai), path(gzi)
```

modules/nf-core/gatk4/mutect2/meta.yml
```diff
@@ -56,7 +61,7 @@ input:
...
         ontologies: []
```
famosab, line 60:
```suggestion
        ontologies: []
- gzi:
        type: file
        description: Index of bgzipped reference fasta file
        pattern: "*.fasta.gz.gzi"
        ontologies: []
```

modules/nf-core/gatk4/mutect2/meta.yml
famosab:
```suggestion
```

modules/nf-core/gatk4/mutect2/main.nf
apetizerr: would the suggestion still be to put gzi as mutiple inputs into the fai input path

modules/nf-core/gatk4/mutect2/main.nf
mignardise: thank you for the review! i also implemented the input changes on the test file so the inputs are aligned and working

modules/nf-core/gatk4/mutect2/tests/main.nf.test.snap
```diff
@@ -1,7 +1,7 @@
 {
     "human - bam - generate_pon": {
         "content": [
-            "876aa6be01c0c8fc71ad8e99ed842240",
+            "c3b555d00c575478b65626f56e3085c4",
```
famosab: any idea why all of these md5sums change even though we changed nothing about thr module itself? 🤔

famosab:
@apetizerr You will also need to adapt the subworkflows to make the tests pass as we add a new input here they fail. The following subworkflows need to be updated:

- [ ] BAM_TUMOR_NORMAL_SOMATIC_VARIANT_CALLING_GATK
- [ ] BAM_CREATE_SOM_PON_GATK
- [ ] BAM_TUMOR_ONLY_SOMATIC_VARIANT_CALLING_GATK

modules/nf-core/gatk4/mutect2/meta.yml
```diff
@@ -56,7 +56,12 @@ input:
     - fai:
         type: file
         description: Index of reference fasta file
-        pattern: "*.fasta.fai"
+        pattern: "*.{fasta.fai,fasta.fai.gz}"
+        ontologies: []
+- gzi:
```
famosab:
```suggestion
    - gzi:
```

modules/nf-core/gatk4/mutect2/tests/main.nf.test.snap
```diff
@@ -84,7 +84,7 @@
     },
     "human - cram": {
         "content": [
-            "1b65f1a163b517944bf2e4b74230e035",
+            "6a839307b251fbeb1a4b08e3b3692513",
```
famosab: I think these changes should not have happened - I guess that they are due to the platform you ran the tests on. I had these issues sometimes when running directly on my mac. We have multiple possibilities to solve this: either switch to a linux workstation or run it in gitpod :)

modules/nf-core/gatk4/mutect2/tests/main.nf.test.snap
mignardise: i ran the tests for other modules since i was moving the inputs around

famosab: but still the `bam` readsMD5 sum should not change just because you add an empty input

famosab:
It looks good but I think you need to run the tests again on gitpod or a different environment or otherwise the tests will fail because the checksums do not match

famosab:
Yeah the tests are failing because some of the md5sums changed when you ran it locally (this should not happen for tests that you do not touch). I would suggest to try an run it in gitpod :)

mignardise:
tests ran via ona/gitpod

famosab:
Ok lets see if the tests run. In the meantime did you request to be added to the github org in the slack? Maybe you can do that then for the next PR you can directly see the CI tests :)

modules/nf-core/gatk4/mutect2/tests/main.nf.test.snap
```diff
@@ -190,104 +217,48 @@
             "nf-test": "0.9.2",
             "nextflow": "25.04.7"
         },
-        "timestamp": "2025-09-15T16:17:00.767220472"
+        "timestamp": "2025-09-14T23:49:18.290784"
     },
     "human - bam - tumor_normal_pair_f1r2 - stub": {
```
famosab, line 222: Now the results for this test are completely empty....

famosab (commented):
If you give me write access to your repository. I can try and run the tests and push the updated snapshots. I do not understand whats going on with these and could try and see whether I can resolve that.

modules/nf-core/gatk4/mutect2/tests/main.nf.test.snap
mignardise: I requested to join the github org and invited to give access

mignardise: Since there is a limit to the number of times i can use ONA to test, should I set up my environment in windows OS and not mac to avoid the checksum troubles?

famosab: No I think either we use linux or I will try and run it on my linux workstation locally

famosab (approved):
🚀

PR #9064, opened 2025-09-15
fix: Improve samtools flagstat stub to generate realistic output

edmundmiller:
Fixes flagstat stub to generate realistic output instead of empty files.

This resolves parsing failures in stub workflows when processes extract mapped read counts from flagstat output for scale factor calculations.

Related to nf-core/chipseq#470 where this issue was first encountered during bedtools genomecov refactoring.


pinin4fjords (approved):
Thanks! I'll give it the green tick so you can merge once the tests pass

PR #9078, opened 2025-09-16
parabricks: bulk update, language server, meta improvements

famosab:
## PR checklist

Closes #XXX 

- [ ] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [ ] Remove all TODO statements.
- [ ] Emit the `versions.yml` file.
- [ ] Follow the naming conventions.
- [ ] Follow the parameters requirements.
- [ ] Follow the input/output options guidelines.
- [ ] Add a resource `label`
- [ ] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - For modules:
    - [ ] `nf-core modules test <MODULE> --profile docker`
    - [ ] `nf-core modules test <MODULE> --profile singularity`
    - [ ] `nf-core modules test <MODULE> --profile conda`
  - For subworkflows:
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile docker`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile singularity`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile conda`

PR #9082, opened 2025-09-17
Fix parabricks/deepvariant gVCF output detection

HudoGriz:
## Issue
The previous implementation failed when `task.ext.args` was defined as a closure, causing the module to output `.vcf` files even when `--gvcf` was specified, leading to Parabricks errors.

## Description
Fixes gVCF output file detection in the parabricks/deepvariant module.

## Changes
- Replaced regex pattern matching `("--gvcf" =~ args)` with `args.contains("--gvcf")` for more reliable detection


famosab:
I think this is a duplicate of #8726, but good reminder I shall add it to the bigger parabricks PR #9078

HudoGriz:
Yes, looks like the same problem. Thanks, looking forward to it.

famosab:
Closing this as it should be fixed by #9078 - feel free to reopen if something is still not working as expected :)

PR #9085, opened 2025-09-17
Update ultra/pipeline

SPPearce:
Updating ultra/pipeline to add meta maps and capturing only the md5sum of the reads for the sam file to prevent difference with conda


SPPearce:
Something inconsistent in the reads md5sum between docker/conda, but the statistics seem stable.

PR #9125, opened 2025-09-23
add strobealign to fastq_align_dna

matthdsm:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/cooler/balance/environment.yml
+++ b/modules/nf-core/cooler/balance/environment.yml
@@ -4,5 +4,4 @@ channels:
   - conda-forge
   - bioconda
 dependencies:
-  - bioconda::cooler=0.9.2
-  - conda-forge::numpy=1.24.3
+  - bioconda::cooler=0.10.4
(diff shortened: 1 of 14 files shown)


subworkflows/nf-core/fastq_align_dna/tests/nextflow.config
SPPearce, line 1: What is this for?

SPPearce (commented):
I'm a bit surprised that the module is just called strobealign, rather than having a nested structure?
The documentation suggests you can pre-build an index and pass that around, which would suggest a submodule.

subworkflows/nf-core/fastq_align_dna/tests/nextflow.config
matthdsm: Nothing 🤷🏻  was there before, so I just amended it

matthdsm:
> The documentation suggests you can pre-build an index and pass that around, which would suggest a submodule.


There's only has a single executable, so the index generation is handled within the module and some flag magic. The default behaviour for `strobealign` is to generate the index on the fly based of the read lengths and the fasta file,  which should be quite fast.

This also means the index is highly dependent on the input data, and as such is not advised to be generated beforehand.

matthdsm:
For reference, a quick test runs 66 seconds to generate an index for GRCh38

SPPearce (approved):
Ok, fair enough, that is very fast.

PR #9126, opened 2025-09-23
📝 docs: Updated UNTAR meta.yml to include .tar as a target to untar

FloWuenne:
# Description

Added `.tar` to the description of archives that can be processed with UNTAR.

### DIFF
--- a/modules/nf-core/untar/meta.yml
+++ b/modules/nf-core/untar/meta.yml
@@ -1,5 +1,5 @@
 name: untar
-description: Extract files from tar and tar.gz archives
+description: Extract files from tar, tar.gz, tar.bz2, tar.xz archives
 keywords:
   - untar
   - uncompress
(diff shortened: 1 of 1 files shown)


modules/nf-core/untar/meta.yml
```diff
@@ -19,8 +19,8 @@ input:
           e.g. [ id:'test', single_end:false ]
     - archive:
         type: file
-        description: File to be untar
-        pattern: "*.{tar}.{gz}"
+        description: File to be untarred
+        pattern: "*.{tar,tar.gz}"
```
pinin4fjords: Do we need to allow for any other tar.x extensions?

FloWuenne: [suggestion](https://pullpo.io/cc?l=suggestion) 
The two other ones that should work with the code I guess are:
- .tar.bz2 (better compression)
- .tar.xz (best compression, widely supported)

PR #9156, opened 2025-09-30
Update PureCN to support mapping bias and VCF input and update container dependencies

### LINKED ISSUE #9155: [FEATURE] PureCN: support VCF input as part of purecn/run

It's optional, but it's the recommended way to run it. I'm opening this issue to track the change, but I have a PR almost ready already.

Implement `--vcf` in `purecn/run`.

_No response_

### Additional context

_No response_

lbeltrame:
This is actually the recommended way to run PureCN, but the parameter had never been specified.
The stub test (the actual test can't run due to test data being incorrect, see comment) is also updated.

Remove `--stats-file` as this is actually an argument to load MuTect 1 and 2 stats files (and this means it should be in `args`).

In addition, bump PSCBS to 0.68.0 in all modules (including rebuilding containers) to prevent a bug triggered by the `future` package that affected all PSCBS versions < 0.68.0. Testing also showed that MASS was also a required dependency, now added to conda environments and containers.

Also remove Aldo from the maintainers (he's no longer working in our group since more than a year).

## PR checklist

Closes #9155 

### DIFF
--- a/modules/nf-core/purecn/coverage/environment.yml
+++ b/modules/nf-core/purecn/coverage/environment.yml
@@ -1,7 +1,6 @@
 channels:
   - conda-forge
   - bioconda
-
 dependencies:
   - bioconda::bioconductor-dnacopy=1.80.0
   - bioconda::bioconductor-org.hs.eg.db=3.20.0
@@ -11,5 +10,5 @@ dependencies:
   - bioconda::r-pscbs=0.68.0
   - conda-forge::procps-ng=4.0.4
   - conda-forge::r-futile.logger=1.4.3
+  - conda-forge::r-mass=7.3_65
   - conda-forge::r-optparse=1.7.5
-  - conda-forge::r-mass
--- a/modules/nf-core/purecn/intervalfile/environment.yml
+++ b/modules/nf-core/purecn/intervalfile/environment.yml
@@ -1,7 +1,6 @@
 channels:
   - conda-forge
   - bioconda
-
 dependencies:
   - bioconda::bioconductor-dnacopy=1.80.0
   - bioconda::bioconductor-org.hs.eg.db=3.20.0
@@ -11,5 +10,5 @@ dependencies:
   - bioconda::r-pscbs=0.68.0
   - conda-forge::procps-ng=4.0.4
   - conda-forge::r-futile.logger=1.4.3
+  - conda-forge::r-mass=7.3_65
   - conda-forge::r-optparse=1.7.5
-  - conda-forge::r-mass
--- a/modules/nf-core/purecn/normaldb/environment.yml
+++ b/modules/nf-core/purecn/normaldb/environment.yml
@@ -1,7 +1,6 @@
 channels:
   - conda-forge
   - bioconda
-
 dependencies:
   - bioconda::bioconductor-dnacopy=1.80.0
   - bioconda::bioconductor-org.hs.eg.db=3.20.0
@@ -11,5 +10,5 @@ dependencies:
   - bioconda::r-pscbs=0.68.0
   - conda-forge::procps-ng=4.0.4
   - conda-forge::r-futile.logger=1.4.3
+  - conda-forge::r-mass=7.3_65
   - conda-forge::r-optparse=1.7.5
-  - conda-forge::r-mass
--- a/modules/nf-core/purecn/run/environment.yml
+++ b/modules/nf-core/purecn/run/environment.yml
@@ -1,7 +1,6 @@
 channels:
   - conda-forge
   - bioconda
-
 dependencies:
   - bioconda::bioconductor-dnacopy=1.80.0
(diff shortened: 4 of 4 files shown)


modules/nf-core/purecn/run/main.nf
```diff
@@ -46,7 +47,7 @@ process PURECN_RUN {
         --genome ${genome} \\
         --parallel \\
         --cores ${task.cpus} \\
-        --stats-file ${prefix}_stats.txt \\
```
FriederikeHanssen, line 49: are you removing the stats file on purpose?

lbeltrame: Correct, this was an oversight, this doesn't produce a stat file but *reads* one. 

From https://bioconductor.org/packages/devel/bioc/vignettes/PureCN/inst/doc/Quick.html:

> The MuTect 1.1.7 stats file (the main output file besides the VCF) should be provided for better artifact filtering. If the VCF was generated by a pipeline that performs good artifact filtering, this file is not needed. Do NOT provide this file for Mutect 2.

So this option belongs to `args` and not here.

lbeltrame:
Added mapping bias as well, because its use in tandem with the normalDB (it is actually made along with the normalDB if a pool of normal VCFs is provided) is the recommended workflow.

modules/nf-core/purecn/run/main.nf
SPPearce: Well, if the file is an input it'd need to be staged.
But Mutect 1 is really quite outdated now, so I wouldn't recommend supporting that anyway.

lbeltrame: I never ever tested this scenario with a recent PureCN so I don't even know if it would work...

lbeltrame:
Needs an update in the container, testing this found a bug that needs an update in PSCBS to 0.68.0.
~~Only, I need to remember what's in the container...~~ Ok, that is solved.

I can add the bump here, or do you prefer a separate PR with all the modules?

modules/nf-core/purecn/run/main.nf
```diff
@@ -9,8 +9,9 @@ process PURECN_RUN {
         'community.wave.seqera.io/library/bioconductor-dnacopy_bioconductor-org.hs.eg.db_bioconductor-purecn_bioconductor-txdb.hsapiens.ucsc.hg19.knowngene_pruned:cc846801cfba58d6' }"
 
     input:
-    tuple val(meta), path(intervals), path(coverage)
+    tuple val(meta), path(intervals), path(coverage), path(vcf)
     path normal_db
+    path mapping_bias
```
SPPearce, line 14: Do we want meta maps on these file inputs?

SPPearce:
> Needs an update in the container, testing this found a bug that needs an update in PSCBS to 0.68.0. ~Only, I need to remember what's in the container...~ Ok, that is solved.
> 
> I can add the bump here, or do you prefer a separate PR with all the modules?

Don't mind, you can do it either way. Good to keep them in sync between the different submodules though.

lbeltrame:
Ok, I'll get to it next week.

modules/nf-core/purecn/run/main.nf
lbeltrame: It's generally a single-use-for-all-samples input, I'm not sure if a meta map would be useful here.

lbeltrame:
So, if you don't mind I'll add a bump here for all modules (and update PR message and title accordingly).

lbeltrame:
Conda builds are failing, but I can't check the package I used because it doesn't show up in my Seqera Containers history (?) and the search there is not working for conda-forge packages.

EDIT: ok, this is weird... https://wave.seqera.io/view/builds/bd-7fef74d5cbdeecbe_1 reports a successful build with the exact same dependencies...

lbeltrame:
Well, I don't know what's wrong with conda, and not using it at all here makes it difficult to fix it (the containers built just fine, and the error is totally out of whack). 
Help is welcome.

lbeltrame:
Hopefully I was able to handle this, but I don't think it's "right", any version specified for `r-mass` will cause dependency errors on an utterly incompatible version of R. I was able to get it installed (with the same version, nonetheless!) only by specifying the dependency as unversioned.

lbeltrame:
~~And I see lint failing, of course. Unless someone is able to tell me what's wrong, I'll have to keep this open and fork it locally: we work exclusively with containers and I really can't figure out why conda is picking the wrong package.~~

Well, I have no idea why, but now it's actually installing stuff and passing...

lbeltrame:
Thanks! I'll do a follow-up PR (if my tests work) because I noticed there are a few important outputs that aren't caught, namely the sample log and the sample flags ("curation" in PureCN parlance, since after manual curation it can be re-used to rerun estimates).

PR #8978, opened 2025-09-01
First pass at POC for topics update for bwamem2 and dependencies

maxulysse:
## PR checklist

Closes #XXX 

- [ ] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [ ] Remove all TODO statements.
- [ ] Emit the `versions.yml` file.
- [ ] Follow the naming conventions.
- [ ] Follow the parameters requirements.
- [ ] Follow the input/output options guidelines.
- [ ] Add a resource `label`
- [ ] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - For modules:
    - [ ] `nf-core modules test <MODULE> --profile docker`
    - [ ] `nf-core modules test <MODULE> --profile singularity`
    - [ ] `nf-core modules test <MODULE> --profile conda`
  - For subworkflows:
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile docker`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile singularity`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile conda`

PR #9159, opened 2025-09-30
POC: topics

nvnieuwk:
This PR serves as an example of the work being done in https://github.com/nf-core/tools/pull/3779

### DIFF
--- a/modules/nf-core/bwamem2/index/main.nf
+++ b/modules/nf-core/bwamem2/index/main.nf
@@ -14,7 +14,7 @@ process BWAMEM2_INDEX {
 
     output:
     tuple val(meta), path("bwamem2"), emit: index
-    tuple val("${task.process}"), val('bwa-mem2'), eval("bwa-mem2 version 2>&1 | tail -1"),                            topic: versions, emit: versions1
+    tuple val("${task.process}"), val('bwa-mem2'), eval("bwa-mem2 version 2>&1 | tail -1"), topic: versions, emit: versions_bwamem2
 
     when:
     task.ext.when == null || task.ext.when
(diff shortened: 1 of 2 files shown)


maxulysse:
So this one replaces #8978 ?

nvnieuwk:
Not yet, I still need to update the subworkflows here

modules/nf-core/bwamem2/index/main.nf
```diff
@@ -14,7 +14,7 @@ process BWAMEM2_INDEX {
 
     output:
     tuple val(meta), path("bwamem2"), emit: index
-    path "versions.yml"             , emit: versions
+    tuple val("${task.process}"), val('bwa-mem2'), eval("bwa-mem2 version 2>&1 | tail -1"),                            topic: versions, emit: versions1
```
mirpedrol: Just to be picky here, but the harshil alignment is not correct 😂

modules/nf-core/bwamem2/index/meta.yml
```diff
@@ -39,13 +39,27 @@ output:
...
+  - versions:
+      - - process:
+            type: string
+            description: The process the versions were collected from
+        - tool:
+            type: string
+            description: The tool name
+        - version:
+            type: string
+            description: The version of the tool
```
mirpedrol, line 62: this looks good! 
but if we want to automate it somehow, the values of the topic channel should be the same that we add for `versions1`: `${task.process}`, `bwa-mem2` and `bwa-mem2 version 2>&1 | tail -1`.
It doesn't look as nice, but it's the only way we can automate the generation of the meta.yml

nvnieuwk: Thanks the main issue here is that the versions topic can consist of multiple tools here. I'm also not sure this needs automation as this will be in the template anyways. I've tried adding the automation, but it's not that trivial to do :/

modules/nf-core/limma/differential/tests/main.nf.test
eit-maxlcummins, line 1: @wangdepin

SPPearce:
@nvnieuwk , can this PoC now be closed?

nvnieuwk:
Yes, you're right!

PR #9173, opened 2025-10-06
Add diann subworkflow and module

pinin4fjords:
## Overview

This PR ports the DIA-NN workflow from [bigbio/quantms](https://github.com/bigbio/quantms) as an nf-core subworkflow (`dia_proteomics_analysis`), refactored to align with nf-core conventions and best practices.

## Motivation

The original quantms implementation used multiple separate DIA-NN modules with different hardcoded configurations and extensive conditional logic. This PR consolidates and modernizes the workflow architecture to be more maintainable, flexible, and idiomatic to nf-core standards.

### Architecture improvements

- **Single unified DIA-NN module**: Replaced multiple configuration-specific modules with one flexible module. The differences between quantms modules were purely configuration-based, not functional.
  
- **Separation of concerns**: Moved custom pipeline logic from module-level scripts into subworkflow Groovy, keeping modules focused on tool execution.

- **Parameter handling**: The subworkflow no longer directly references `params`. Parameters required for workflow logic are explicitly passed as subworkflow inputs, while tool-specific parameters can be supplied via standard Nextflow config files.

### Data flow improvements

- **Enhanced metadata propagation**: Comprehensive use of meta maps throughout the workflow enables better provenance tracking and allows more complex experimental designs.

- **Channel-driven conditional execution**: Replaced explicit conditionals with Nextflow channel operators (`.ifEmpty(null)`, `.filter()`, `.combine()`), making the data flow more declarative and the workflow logic more transparent.

- **Support for iteration**: The workflow is designed to handle multiple inputs (e.g., multiple FASTA databases, experimental designs) rather than assuming single global inputs, enabling more flexible experimental setups.

### Specific technical improvements

- **Pre-generated inputs support**: Users can optionally supply pre-generated spectral libraries or empirical libraries to skip computationally expensive steps.

- **Mass accuracy extraction**: Automatic extraction of empirical mass accuracy from DIA-NN preliminary logs to inform final quantification parameters.

- **Comprehensive outputs**: The workflow emits multiple output formats (TSV, Parquet, mzTab, MSstats-compatible, Triqler-compatible) to support diverse downstream analysis needs.

## Pipeline stages

The subworkflow implements the complete DIA-NN analysis pipeline:

1. **Configuration generation** (`quantmsutils/dianncfg`): Generate DIA-NN config files based on experimental design
2. **In-silico library generation** (`diann`): Predict spectral library from FASTA (skippable if pre-generated library provided)
3. **Preliminary analysis** (`diann`): Initial pass to generate empirical library (skippable if pre-generated empirical library provided)
4. **Mass accuracy extraction**: Parse DIA-NN logs to extract optimal mass accuracy settings
5. **Individual sample analysis** (`diann`): Per-sample quantification using empirical library
6. **Final quantification** (`diann`): Unified analysis across all samples
7. **Format conversion and statistics** (`quantmsutils/diann2mztab`, `quantmsutils/mzmlstatistics`): Generate standard output formats and QC metrics

## Modules included

- `modules/nf-core/diann/`: Universal DIA-NN wrapper
- `modules/nf-core/quantmsutils/dianncfg/`: DIA-NN configuration file generation
- `modules/nf-core/quantmsutils/diann2mztab/`: Convert DIA-NN output to mzTab format
- `modules/nf-core/quantmsutils/mzmlstatistics/`: Extract mzML file statistics
- `subworkflows/nf-core/dia_proteomics_analysis/`: Main DIA analysis orchestration

## Testing

- All modules and the subworkflow include nf-test suites with appropriate test data
- Tests use snapshot comparisons for reproducible outputs (versions, CSV files) and existence checks for non-reproducible outputs (TSV, mzTab, Parquet files)
- Test data PR: https://github.com/nf-core/test-datasets/pull/1756 (now merged)

## Known limitations

The MSstats LFQ analysis step from [quantms dia.nf](https://github.com/bigbio/quantms/blob/e516ddca04a2ff9eedae1df951995b8793c82c62/workflows/dia.nf#L165) is not yet ported, as it requires custom R scripts that need refactoring into a proper nf-core module. This will be addressed in a follow-up PR.

## Reviewer notes

- The conditional execution pattern using `.ifEmpty(null).filter()` is intentional and documented inline - this allows downstream processes to execute only when pre-generated inputs are not provided
- The large `multiMap` block (lines ~155-161 in `main.nf`) generates all required entity combinations upfront and creates multiple channel views at different metadata granularities - this is documented with detailed comments
- Module meta.yml files follow nf-core schema for both modules and subworkflows, with proper tuple structures and output definitions
- Tests skip Conda for quantmsutils modules (added to `.github/skip_nf_test.json`) as these tools are not available via Conda

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/subworkflows/nf-core/dia_proteomics_analysis/tests/main.nf.test
+++ b/subworkflows/nf-core/dia_proteomics_analysis/tests/main.nf.test
@@ -71,7 +71,6 @@ nextflow_workflow {
             assertAll(
                 { assert workflow.success },
                 { assert path("${workflow.out.diann_report[0][1]}").exists() },
-                { assert path("${workflow.out.diann_report_parquet[0][1]}").exists() },
                 { assert path("${workflow.out.out_triqler[0][1]}").exists() },
                 { assert path("${workflow.out.final_result[0][1]}").exists() },
                 { assert path("${workflow.out.mzml_statistics[0][1]}").exists() },


tests/config/nextflow.config
```diff
@@ -7,7 +7,7 @@ params {
 
 process {
     resourceLimits = [
-        cpus: 4,
+        cpus: 10,
```
maxulysse: [nitpick](https://pullpo.io/cc?l=nitpick) 
Let's not change the default for ALL tests

pinin4fjords: Ooops, sorry, just a debugging change I hadn't reverted

jonasscheid:
FYI I'm going to review. Hope to be finished tomorrow

modules/nf-core/diann/main.nf
```diff
@@ -0,0 +1,116 @@
...
+    // Exit if running this module with -profile conda / -profile mamba
+    if (workflow.profile.tokenize(',').intersect(['conda', 'mamba']).size() >= 1) {
+        error "DIANN module does not support Conda. Please use Docker / Singularity / Podman instead."
+    }
+    def prefix = task.ext.prefix ?: "${meta.id}"
+
+    """
+    # Library outputs
+    touch ${prefix}.predicted.speclib
+    touch empirical_library.speclib
```
jonasscheid: Why no prefix here?

modules/nf-core/diann/main.nf
```diff
@@ -0,0 +1,116 @@
...
+    def ms_input = ''
+    if (ms_files && ms_files != []) {
+        // ms_files provided (actual file paths for preliminary/assembly/individual analysis)
+        ms_input = ms_files instanceof List ? ms_files.collect{ "--f ${it}" }.join(' ') : "--f ${ms_files}"
+    } else if (ms_file_names && ms_file_names != []) {
+        // ms_file_names provided (just basenames for final quantification with --use-quant)
+        ms_input = ms_file_names instanceof List ? ms_file_names.collect{ "--f ${it}" }.join(' ') : "--f ${ms_file_names}"
+    }
+    
+    def fasta_input = fasta && fasta.name != 'NO_FASTA_FILE' ? "--fasta ${fasta}" : ''
```
jonasscheid: Would check for [] instead of 'NO_FASTA_FILE' make more sense here 🤔?

modules/nf-core/diann/meta.yml
```diff
@@ -0,0 +1,218 @@
...
+          pattern: "*.predicted.speclib"
+  empirical_library:
+    - - meta:
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. `[ id:'sample1' ]`
+      - "empirical_library.*":
+          type: file
+          description: Empirical spectral library in .speclib format (from assembly)
```
jonasscheid: What do you mean with assembly here

modules/nf-core/diann/meta.yml
```diff
@@ -0,0 +1,218 @@
...
+          pattern: "empirical_library.*"
+  final_speclib:
+    - - meta:
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. `[ id:'sample1' ]`
+      - "empirical_library.tsv":
+          type: file
+          description: Final spectral library in TSV format
```
jonasscheid: I don't understand the diff between final_speclib and emperical_speclib. Could you clarify for me (and maybe in here as well)?

modules/nf-core/diann/tests/main.nf.test
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    test("proteomics - in-silico library generation") {
+        config "./nextflow.config"
+        
+        when {
+            process {
+                """
+                input[0] = Channel.of([
+                    [id:'ecoli_test', config: '--cut K*,R*,!*P --fixed-mod Carbamidomethyl,57.021464,C --var-mod Oxidation,15.994915,M'],
```
jonasscheid, line 33: Could these additional params also go into the nextflow.config?

modules/nf-core/quantmsutils/mzmlstatistics/main.nf
```diff
@@ -0,0 +1,59 @@
+process QUANTMSUTILS_MZMLSTATISTICS {
+    tag "$meta.id"
+    label 'process_very_low'
```
jonasscheid: duplicated ressource labels

modules/nf-core/quantmsutils/mzmlstatistics/main.nf
```diff
@@ -0,0 +1,59 @@
...
+    tuple val(meta), path(ms_file)
+
+    output:
+    tuple val(meta), path("*_ms_info.parquet"), emit: ms_statistics
+    tuple val(meta), path("*_ms2_info.parquet"), emit: ms2_statistics, optional: true
+    tuple val(meta), path("*_feature_info.parquet"), emit: feature_statistics, optional: true
+    path "versions.yml", emit: versions
+    tuple val(meta), path("*.log"), emit: log
+
+    when:
```
jonasscheid, line 20: I thought this one is not needed anymore 🤔

subworkflows/nf-core/dia_proteomics_analysis/main.nf
```diff
@@ -0,0 +1,426 @@
...
+    // MODULE: Generate DIA-NN configuration. Needs to run once each unique enzyme, fixed modifications, and variable modifications combination.
+    //
+    // Only run config generation and speclib generation if no speclib is provided
+    // Conditional execution pattern:
+    //   - combine with ifEmpty(null) pairs each input with either existing speclib or null
+    //   - filter keeps only tuples containing null (i.e., no pre-generated speclib exists)
+    //   - downstream processes only execute when needed
+
+    ch_config_input = input.enzyme_mods.unique() // [meta_enzyme_mods, enzyme, fixed_mods, variable_mods]
+        .combine(ch_speclib.ifEmpty(null))
```
jonasscheid, line 232: ifEmpty(null) needed?

subworkflows/nf-core/dia_proteomics_analysis/main.nf
```diff
@@ -0,0 +1,426 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    IMPORT NF-CORE MODULES
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+*/
+
+include { QUANTMSUTILS_DIANNCFG          } from '../../../modules/nf-core/quantmsutils/dianncfg/main'
+include { QUANTMSUTILS_MZMLSTATISTICS    } from '../../../modules/nf-core/quantmsutils/mzmlstatistics/main'
+include { QUANTMSUTILS_DIANN2MZTAB       } from '../../../modules/nf-core/quantmsutils/diann2mztab/main'
+
+include { DIANN as DIANN_INSILICOLIBRARYGENERATION } from '../../../modules/nf-core/diann/main'
```
jonasscheid, line 11: Style option: A bit more compressed would be to:` { DIANN as X; Y; Z; ..}`

subworkflows/nf-core/dia_proteomics_analysis/main.nf
```diff
@@ -0,0 +1,426 @@
...
+    ch_final_quantification_input = ch_individual_analysis_input // [meta_with_settings, ms_file, fasta, empirical_library]
+        .join(DIANN_INDIVIDUALANALYSIS.out.diann_quant)          // [meta_with_settings, ms_file, fasta, empirical_library, diann_quant]
+        .map { [it[0].experiment] + it.drop(1)}                  // [meta_exp_searchdb, ms_file, fasta, empirical_library, diann_quant]
+        .groupTuple()                                            // [meta_exp_searchdb, [ms_file], [fasta], [empirical_library], [diann_quant]]
+        .map{ sortListsByPathName(it, 1) }                       // [meta_exp_searchdb, [sorted_ms_file], [sorted_fasta], [sorted_empirical_library], [sorted_diann_quant]]
+        .map { meta, ms_files, fasta, empirical_library, diann_quant ->
+            [meta, [], ms_files.collect{ it.name}, fasta, empirical_library, diann_quant] // Use ms_file_names instead of ms_files for final quant
+        }
+
+    DIANN_FINALQUANTIFICATION(ch_final_quantification_input)
```
jonasscheid, line 367: Does this process quantify across runs?

jonasscheid:
@pinin4fjords are you aware that DIA-NN moved from open-source to license-restricted access beyond version 1.8.2? It's still free for academic use, but container distribution is not allowed in newer versions (latest version is 2.2.0 with quite some improvements..). So my general question would be what the version update strategy of this nf-core subworkflow is? Don't get me wrong, it's nice that this subworkflow is part of nf-core, but we need to think about updating to future versions eventually.

pinin4fjords:
> @pinin4fjords are you aware that DIA-NN moved from open-source to license-restricted access beyond version 1.8.2? It's still free for academic use, but container distribution is not allowed in newer versions (latest version is 2.2.0 with quite some improvements..). So my general question would be what the version update strategy of this nf-core subworkflow is? Don't get me wrong, it's nice that this subworkflow is part of nf-core, but we need to think about updating to future versions eventually.

Hi @jonasscheid , thanks for the comments, will address them today. For complete clarity this in contract work done at the request of @FelixKrueger .

In terms of updates, yes, we're aware of the licensing issues. It will be relatively trivial for folks to use configuration to override `container` directives and `ext.args` to suit newer versions, which covers some of the answer. But where the newer version would require larger-scale changes to the module/ subworkflow structure I don't have a good answer.

subworkflows/nf-core/dia_proteomics_analysis/main.nf
pinin4fjords: That would be  a very long line and I would feel less able to be descriptive in the aliases. I'm going to respectfully disagree on this one.

subworkflows/nf-core/dia_proteomics_analysis/main.nf
pinin4fjords: In this case, yes (appreciate it's been widely removed for other use cases).

I'm using it as a way of avoiding conditionals. By doing it this way, I can use a filter immediately after to tell me if there was any pre-made speclib input. Only if that is not the case do we populate the channel at the end of this chain and proceed with the downstream steps.

subworkflows/nf-core/dia_proteomics_analysis/main.nf
pinin4fjords: Yes, the same as the quantms workflow. The groupTuple in the ops above is how we merge all the inputs.

subworkflows/nf-core/dia_proteomics_analysis/main.nf
jonasscheid: Makes sense!

modules/nf-core/diann/meta.yml
DongzeHE: I think they are the same. These two are the same file published from two different modules, [assemble_empirical_library](https://github.com/bigbio/quantms/blob/e516ddca04a2ff9eedae1df951995b8793c82c62/modules/local/diann/assemble_empirical_library/main.nf#L18) and  [final_quantification](https://github.com/bigbio/quantms/blob/e516ddca04a2ff9eedae1df951995b8793c82c62/modules/local/diann/final_quantification/main.nf#L35). I checked a run I had. The latter is a symbolic link pointing to the former.

modules/nf-core/diann/main.nf
```diff
@@ -0,0 +1,116 @@
...
+
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://containers.biocontainers.pro/s3/SingImgsRepo/diann/v1.8.1_cv1/diann_v1.8.1_cv1.img' :
+        'docker.io/biocontainers/diann:v1.8.1_cv1' }"
+
+    input:
+    tuple val(meta), path(ms_files), val(ms_file_names), path(fasta), path(library), path(quant, stageAs: 'quant/*')
+
+    output:
+    // Library outputs
```
DongzeHE, line 13: I think we can probably optimize this part, because
1. we need to define `--out-lib empirical_library` in order to ensure the files are properly named (see [here](https://github.com/vdemichev/DiaNN#:~:text=from%20this%20one-,%2D%2Dout%2Dlib,-%5Bfile%20name%5D))
2. skyline speclib was introduced in DIANN 1.9, and will not work until the future version of skyline is released (see [here](https://github.com/vdemichev/DiaNN#:~:text=two%20visualisation%20options.-,Skyline,-.%20To%20visualise%20chromatograms))
3. the actual empirical library that will be passed to `-lib` in the second round of searching is called `empirical_library.speclib`, which is not being captured explicitly.
4. A library, either a predicted or empirical one, will always be the input of DIANN, which means there will alwyas be a subset of these files being published in all DIANN calls, which is suboptimal. 
5. the `predict_speclib` is always called `lib.predicted.speclib` if it is generated by DIANN

modules/nf-core/diann/main.nf
```diff
@@ -0,0 +1,116 @@
...
+    tuple val(meta), path("diann_report.manifest.txt"), emit: report_manifest, optional: true
+    tuple val(meta), path("diann_report.protein_description.tsv"), emit: protein_description, optional: true
+    tuple val(meta), path("diann_report.stats.tsv"), emit: report_stats, optional: true
+    tuple val(meta), path("diann_report.pr_matrix.tsv"), emit: pr_matrix, optional: true
+    tuple val(meta), path("diann_report.pg_matrix.tsv"), emit: pg_matrix, optional: true
+    tuple val(meta), path("diann_report.gg_matrix.tsv"), emit: gg_matrix, optional: true
+    tuple val(meta), path("diann_report.unique_genes_matrix.tsv"), emit: unique_gene_matrix, optional: true
+    
+    // Common outputs
+    tuple val(meta), path("*.log.txt"), emit: log
```
DongzeHE, line 33: This file is `lib.log.txt` for insilico library generation or `report.log.txt` for all other modes

modules/nf-core/diann/main.nf
```diff
@@ -0,0 +1,116 @@
...
+    } else if (ms_file_names && ms_file_names != []) {
+        // ms_file_names provided (just basenames for final quantification with --use-quant)
+        ms_input = ms_file_names instanceof List ? ms_file_names.collect{ "--f ${it}" }.join(' ') : "--f ${ms_file_names}"
+    }
+    
+    def fasta_input = fasta && fasta.name != 'NO_FASTA_FILE' ? "--fasta ${fasta}" : ''
+    def lib_input = library && library.name != 'NO_LIB_FILE' ? "--lib ${library}" : ''
+    
+    // Determine temp directory based on whether quant files are provided
+    def temp_dir = quant && quant.name != 'NO_QUANT_DIR' ? "--temp ./quant/" : "--temp ./"
```
DongzeHE: if we have temp_dir, `--use-quant` should be specified together with `--temp`

modules/nf-core/diann/main.nf
```diff
@@ -0,0 +1,116 @@
...
+    script:
+    // Exit if running this module with -profile conda / -profile mamba
+    if (workflow.profile.tokenize(',').intersect(['conda', 'mamba']).size() >= 1) {
+        error "DIANN module does not support Conda. Please use Docker / Singularity / Podman instead."
+    }
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    
+    // Handle MS files input: use ms_files if provided, otherwise fall back to ms_file_names. 
+    //Allows us to provide file names without staging the files, useful in certain scenarios.
```
DongzeHE: "in certain scenarios" means when preprocessed files are provided in the `quant/` folder, and we don't want nextflow to create links for these files again. 

To be specific, that is saying, `--use-quant --temp quant/` + `--f ms_file_names` will avoid the raw files being staged in the working directory. Not sure if users will be confused by these two options

modules/nf-core/diann/main.nf
```diff
@@ -0,0 +1,116 @@
...
+    // Library outputs
+    tuple val(meta), path("*.predicted.speclib"), emit: predict_speclib, optional: true
+    tuple val(meta), path("empirical_library.*"), emit: empirical_library, optional: true
+    tuple val(meta), path("empirical_library.tsv"), emit: final_speclib, optional: true
+    tuple val(meta), path("empirical_library.tsv.skyline.speclib"), emit: skyline_speclib, optional: true
+    
+    // Quantification outputs
+    tuple val(meta), path("*.quant"), emit: diann_quant, optional: true
+    
+    // Report outputs (from final quantification)
```
DongzeHE, line 21: The file names are decided by `--out` ([ref](https://github.com/vdemichev/DiaNN#:~:text=to%20DIA%2DNN-,%2D%2Dout%20%5Bfile%20name%5D,-specifies%20the%20name)). We might want to hardcode the parameter somewhere if these file names are hardcoded in the `output` section

modules/nf-core/diann/main.nf
DongzeHE: Maybe pass it to `--out` and/or `out-lib`?

modules/nf-core/diann/main.nf
```diff
@@ -0,0 +1,116 @@
...
+    touch diann_report.pg_matrix.tsv
+    touch diann_report.gg_matrix.tsv
+    touch diann_report.unique_genes_matrix.tsv
+    
+    # Common outputs
+    touch ${prefix}.log
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        DIA-NN: 1.8.1
```
DongzeHE, line 115: maybe just get the version directly from DIANN as did in script?

modules/nf-core/diann/meta.yml
```diff
@@ -0,0 +1,218 @@
...
+input:
+  - - meta:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. `[ id:'sample1' ]`
+    - ms_files:
+        type: file
+        description: |
+          MS data file(s) in mzML, Thermo RAW, or Bruker .d format (can be single file or list).
```
DongzeHE: Only Linux DIANN 2.0 and later can take `.raw` files directly. In older versions, they have to be converted to `.d` or `.mzml` as [here](https://github.com/bigbio/quantms/blob/e516ddca04a2ff9eedae1df951995b8793c82c62/subworkflows/local/file_preparation/main.nf#L64)

modules/nf-core/diann/meta.yml
```diff
@@ -0,0 +1,218 @@
...
+          For preliminary/assembly/individual analysis, these are actual file paths.
+          For final quantification with --use-quant, this should be an empty list.
+        pattern: "*.{mzML,raw,d}"
+    - ms_file_names:
+        type: string
+        description: |
+          MS file basenames (not paths) as strings (can be single name or list).
+          Used for final quantification step with --use-quant where only filenames are needed.
```
DongzeHE, line 37: This can be applied whenever the `--use-quant` is specified. So, [assemble_empirical_library](https://github.com/bigbio/quantms/tree/master/modules/local/diann/assemble_empirical_library) is also applicable.

DongzeHE: Maybe we should couple `--use-quant` `--temp`, and `ms_file_names` together.

modules/nf-core/diann/meta.yml
```diff
@@ -0,0 +1,218 @@
...
+        type: file
+        description: |
+          Spectral library file in .speclib or .tsv format.
+          Use a placeholder file (e.g., 'NO_LIB_FILE') if not needed for the specific analysis step.
+        pattern: "*.{speclib,tsv}"
+    - quant:
+        type: directory
+        description: |
+          Directory containing .quant files from previous DIA-NN analysis.
+          Required for empirical library assembly and final quantification.
```
DongzeHE: I don't think it is "required", we can still pass files via `--f`. It is just a good-to-have.

modules/nf-core/quantmsutils/dianncfg/main.nf
```diff
@@ -0,0 +1,56 @@
...
+    path "*.log", emit: log
+    path "versions.yml", emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    // Exit if running this module with -profile conda / -profile mamba
+    if (workflow.profile.tokenize(',').intersect(['conda', 'mamba']).size() >= 1) {
+        error "QUANTMSUTILS_DIANNCFG module does not support Conda. Please use Docker / Singularity / Podman instead."
```
DongzeHE: I think it does support conda. It is a standalone python package [here](https://github.com/bigbio/quantms-utils)

modules/nf-core/quantmsutils/diann2mztab/main.nf
```diff
@@ -0,0 +1,61 @@
...
+    tuple val(meta), path("*.mzTab"), optional: true, emit: out_mztab
+    tuple val(meta), path("*.log"), emit: log
+    path "versions.yml", emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    // Exit if running this module with -profile conda / -profile mamba
+    if (workflow.profile.tokenize(',').intersect(['conda', 'mamba']).size() >= 1) {
```
DongzeHE: it is on pip (see its [github repo](https://github.com/bigbio/quantms-utils))

modules/nf-core/quantmsutils/dianncfg/meta.yml
```diff
@@ -0,0 +1,69 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/meta-schema.json
+name: "quantmsutils_dianncfg"
+description: |
+  Generate DIA-NN configuration arguments from enzyme and modification parameters.
+  This module uses quantms-utils to convert enzyme and modification specifications into DIA-NN-compatible command-line arguments.
+  It validates modifications against the Unimod database and handles special cases for labeling reagents.
```
DongzeHE: As far as I can remember this can only take unimod modifications. For custom modifications they need to go to DIANN directly.

modules/nf-core/quantmsutils/mzmlstatistics/main.nf
```diff
@@ -0,0 +1,59 @@
+process QUANTMSUTILS_MZMLSTATISTICS {
```
DongzeHE, line 1: As the name suggests, this module only works for mzML files, no `.d` or `.raw` files. We might want to address this in the subworkflow

modules/nf-core/quantmsutils/mzmlstatistics/main.nf
```diff
@@ -0,0 +1,59 @@
...
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        quantms-utils: \$(pip show quantms-utils | grep "Version" | awk -F ': ' '{print \$2}')
+    END_VERSIONS
+    """
+
+    stub:
+    // Exit if running this module with -profile conda / -profile mamba
+    if (workflow.profile.tokenize(',').intersect(['conda', 'mamba']).size() >= 1) {
```
DongzeHE: Same, quantms-utils is on pip

subworkflows/nf-core/dia_proteomics_analysis/main.nf
```diff
@@ -0,0 +1,426 @@
...
+    //   - downstream processes only execute when needed
+
+    ch_config_input = input.enzyme_mods.unique() // [meta_enzyme_mods, enzyme, fixed_mods, variable_mods]
+        .combine(ch_speclib.ifEmpty(null))
+        .filter{tuple -> tuple.any { it == null }} 
+        .map{ meta, enzyme, fixed_mods, variable_mods, _ ->
+            [meta, enzyme, fixed_mods, variable_mods]
+        }
+    
+    QUANTMSUTILS_DIANNCFG(ch_config_input)
```
DongzeHE, line 238: As this only works for unimod molecules, not sure if we want to expose a "costom_mods" parameter to take more mods

subworkflows/nf-core/dia_proteomics_analysis/main.nf
```diff
@@ -0,0 +1,426 @@
...
+
+    //
+    // MODULE: In-silico library generation. Needs to run once for every unique config/ FASTA dataset combination.
+    //
+
+    ch_insilico_library_input = input.search_db_by_enzyme    // [meta_enzyme_mods, meta_fasta_enzyme, meta_fasta, fasta]
+        .combine(QUANTMSUTILS_DIANNCFG.out.diann_cfg, by: 0) // [meta_enzyme_mods, meta_fasta_enzyme, meta_fasta, fasta, cfg_file]
+        .unique() // Multiple inputs might have the same config
+        .map { meta_enzyme_mods, meta_fasta_enzyme, meta_fasta, fasta, cfg_file -> 
+            [meta_fasta_enzyme + [config: cfg_file.text], [], [], fasta, [], []] // Added ms_file_names as []
```
DongzeHE, line 249: As we have one lib for each fasta, [enzyme], [fix_mods], [var_mods] combination, should we create a `meta_fasta_enzyme_mods` and use here?

subworkflows/nf-core/dia_proteomics_analysis/nextflow.config
```diff
@@ -0,0 +1,139 @@
...
+
+    // DIANN_INSILICOLIBRARYGENERATION - In-silico library generation
+    // What makes this "In-Silico Library Generation":
+    //   - Input: FASTA file + enzyme/modification config (from meta.config)
+    //   - Key flags: --fasta-search --predictor --gen-spec-lib --met-excision
+    //   - Output: *.predicted.speclib (predicted spectral library from sequences)
+    //   - No MS data input required at this stage
+    withName: 'DIANN_INSILICOLIBRARYGENERATION' {
+        ext.args = {[
+            meta.config ?: '',
```
DongzeHE, line 28: I am a little bit skeptical about it: meta is passed in the `input` section, but the config is added to `ext.args` here. Maybe put it in the module?

subworkflows/nf-core/dia_proteomics_analysis/nextflow.config
```diff
@@ -0,0 +1,139 @@
...
+            params.max_peptide_length ? "--max-pep-len ${params.max_peptide_length}" : '',
+            params.min_precursor_charge ? "--min-pr-charge ${params.min_precursor_charge}" : '',
+            params.max_precursor_charge ? "--max-pr-charge ${params.max_precursor_charge}" : '',
+            params.max_mods != null ? "--var-mods ${params.max_mods}" : '',
+            params.diann_debug ? "--verbose ${params.diann_debug}" : '',
+            "--fasta-search",
+            "--predictor",
+            "--gen-spec-lib",
+            "--met-excision"
+        ].findAll { it != '' }.join(' ').trim()}
```
DongzeHE, line 45: we probably want to add (i) `--out-lib` (here or somewhere in the module) to define the output file name and (ii) `--rt-profile` because of [this](https://github.com/bigbio/quantms/issues/481#issuecomment-2674005402)

subworkflows/nf-core/dia_proteomics_analysis/nextflow.config
```diff
@@ -0,0 +1,139 @@
...
+    //   - Uses --individual-mass-acc and --individual-windows (if automatic mode enabled)
+    //   - Output: empirical_library.* (refined library based on actual data)
+    //   - Purpose: Build experiment-specific library using preliminary results to improve identifications
+    withName: 'DIANN_ASSEMBLEEMPIRICALLIBRARY' {
+        ext.args = [
+            params.diann_debug ? "--verbose ${params.diann_debug}" : '',
+            "--rt-profiling",
+            "--use-quant",
+            "--gen-spec-lib",
+            "--out-lib empirical_library",
```
DongzeHE: I am not sure if having hardcoded file names in the `output` section is good as we specify file names here.

subworkflows/nf-core/dia_proteomics_analysis/nextflow.config
```diff
@@ -0,0 +1,139 @@
...
+    //   - Input: All MS files + predicted library + preliminary *.quant files
+    //   - Key flags: --rt-profiling --use-quant --gen-spec-lib --out-lib empirical_library
+    //   - Uses --individual-mass-acc and --individual-windows (if automatic mode enabled)
+    //   - Output: empirical_library.* (refined library based on actual data)
+    //   - Purpose: Build experiment-specific library using preliminary results to improve identifications
+    withName: 'DIANN_ASSEMBLEEMPIRICALLIBRARY' {
+        ext.args = [
+            params.diann_debug ? "--verbose ${params.diann_debug}" : '',
+            "--rt-profiling",
+            "--use-quant",
```
DongzeHE: This probably need to pair with --temp and `ms_file_names`

subworkflows/nf-core/dia_proteomics_analysis/nextflow.config
```diff
@@ -0,0 +1,139 @@
...
+    //   - Purpose: Re-analyze each file individually with optimal settings before final aggregation
+    withName: 'DIANN_INDIVIDUALANALYSIS' {
+        ext.args = {
+            [
+                params.diann_debug ? "--verbose ${params.diann_debug}" : '',
+                "--no-ifs-removal",
+                "--no-main-report",
+                "--relaxed-prot-inf",
+                meta.pg_level ? "--pg-level ${meta.pg_level}" : '',
+                meta.mass_acc_ms2 ? "--mass-acc ${meta.mass_acc_ms2}" : '',
```
DongzeHE, line 98: I don't want to be nitty-picky but parsing meta entries here looks strange. Please educate me if it is the best practice!

subworkflows/nf-core/dia_proteomics_analysis/main.nf
```diff
@@ -0,0 +1,426 @@
...
+    )
+    ch_versions = ch_versions.mix(DIANN_PRELIMINARYANALYSIS.out.versions)
+
+    //
+    // MODULE: Assemble empirical library with all inputs from the same experiment + search DB
+    //
+    ch_empirical_input = ch_preliminary_analysis_input    // [meta_input_fasta, ms_file, speclib]
+        .join(DIANN_PRELIMINARYANALYSIS.out.diann_quant)  // [meta_input_fasta, ms_file, speclib, diann_quant]
+        .map { [it[0].experiment] + it.drop(1)}           // [meta_exp_searchdb, ms_file, speclib, diann_quant]
+        .groupTuple()                                     // [meta_exp_searchdb, [ms_file], [speclib], [diann_quant]]
```
DongzeHE, line 309: Not sure if it is a concern but each experiment should only have one speclib, so there should only be a single entry in`[speclib]` .

subworkflows/nf-core/dia_proteomics_analysis/main.nf
```diff
@@ -0,0 +1,426 @@
...
+        }
+
+    DIANN_ASSEMBLEEMPIRICALLIBRARY(ch_empirical_input)
+    ch_versions = ch_versions.mix(DIANN_ASSEMBLEEMPIRICALLIBRARY.out.versions)
+
+    //
+    // Derive suggested settings from DIA-NN log
+    //
+
+    ch_empirical_output = DIANN_ASSEMBLEEMPIRICALLIBRARY.out.empirical_library // [meta_exp_searchdb, empirical_library]
```
DongzeHE: `empirical_library.*` will match multiple files. So this will be a list

subworkflows/nf-core/dia_proteomics_analysis/main.nf
```diff
@@ -0,0 +1,426 @@
...
+    //
+    // MODULE: Final quantification
+    //
+
+    ch_final_quantification_input = ch_individual_analysis_input // [meta_with_settings, ms_file, fasta, empirical_library]
+        .join(DIANN_INDIVIDUALANALYSIS.out.diann_quant)          // [meta_with_settings, ms_file, fasta, empirical_library, diann_quant]
+        .map { [it[0].experiment] + it.drop(1)}                  // [meta_exp_searchdb, ms_file, fasta, empirical_library, diann_quant]
+        .groupTuple()                                            // [meta_exp_searchdb, [ms_file], [fasta], [empirical_library], [diann_quant]]
+        .map{ sortListsByPathName(it, 1) }                       // [meta_exp_searchdb, [sorted_ms_file], [sorted_fasta], [sorted_empirical_library], [sorted_diann_quant]]
+        .map { meta, ms_files, fasta, empirical_library, diann_quant ->
```
DongzeHE, line 363: This actually can be applied to assemble_empirical_library as well. Only file names are needed

subworkflows/nf-core/dia_proteomics_analysis/main.nf
```diff
@@ -0,0 +1,426 @@
...
+        .combine(ch_speclib_in.ifEmpty(null))
+        .combine(ch_empirical_library_in.ifEmpty(null))
+        .combine(ch_empirical_log_in.ifEmpty(null))
+        .multiMap{ 
+            meta_exp, 
+            expdesign,
+            meta_fasta, 
+            fasta, 
+            meta_input, 
+            ms_file,
```
DongzeHE, line 174: Because DIANN 1.8.1 cannot take `.raw` files. I suggested to add the thermorawfileparser module here to parse `[ms_file]` if they are `.raw` files

DongzeHE (commented):
Thank you very much @pinin4fjords . This is a lottt of work!

DongzeHE:
I also want to comment on the discussion of DIANN versions. I think many lines of code are related to features only in DIANN 1.9 and later, for example the `skyline.speclib` and the different types of report (`params.diann_export_xic` etc). The module and the subworkflow should work with DIANN 1.9 and later!

modules/nf-core/diann/tests/main.nf.test
pinin4fjords: For the purpose of testing alone, yes they could.

But I'm trying to use this as a form of documentation, since this is how we actually do it in the subworkflow, because we have a process that produces configuration, which we then stuff into the meta for use in this way.

modules/nf-core/diann/main.nf
pinin4fjords: yes, you're right, I'll correct that (along with the other options)

modules/nf-core/diann/main.nf
pinin4fjords: This comment made me realise I'd not been deploying the prefixing as well as I could- thanks! Fix incoming.

modules/nf-core/diann/meta.yml
pinin4fjords: Yep, thanks guys. My merge of the formerly separate modules was a little incomplete.

modules/nf-core/diann/meta.yml
pinin4fjords: clarified

modules/nf-core/diann/main.nf
pinin4fjords: I've cleaned up now, and made proper use of the nf-core `prefix` standard for both `--out` and `--out-lib`.

I had an incomplete merge previously.

pinin4fjords: We could tweak the defaults if you want though

modules/nf-core/diann/main.nf
pinin4fjords: Simpler with the prefix usage (it will always be `${prefix}.log.txt`).

modules/nf-core/diann/main.nf
pinin4fjords: Thanks! Done.

modules/nf-core/diann/main.nf
pinin4fjords: I've used the nf-core prefix standard for this, hopefully simpler now.

modules/nf-core/diann/main.nf
pinin4fjords: This is just the stub, doesn't really matter

modules/nf-core/diann/meta.yml
pinin4fjords: I don't _think_ we should, that would differ from the quantms implementation. Here's an AI-assisted explanation, let me know if you think it's wrong:

In quantms, both empirical library assembly and final quantification use --use-quant + --temp ./quant/ when quant files are present:

  - Empirical assembly: Uses path(ms_files) (stages actual files) + quant + --use-quant
  - Final quantification: Uses val(ms_files) (basenames only) + quant + --use-quant

The difference is that empirical assembly needs to access the raw MS data (so files are staged), while final quantification only needs the .quant files (so we just pass basenames to avoid staging large files). The choice between ms_files vs ms_file_names is about staging optimization, not about whether to use quant files. The current logic correctly triggers --use-quant/--temp based on whether quant files are provided.

modules/nf-core/quantmsutils/dianncfg/main.nf
pinin4fjords: Ahh yes- fixed

subworkflows/nf-core/dia_proteomics_analysis/main.nf
pinin4fjords: I feel like that might be better handled by the calling workflow via `ext.args`

subworkflows/nf-core/dia_proteomics_analysis/nextflow.config
pinin4fjords: This is about nf-core standards. Optional non-file inputs are supposed to go via ext.args.

subworkflows/nf-core/dia_proteomics_analysis/nextflow.config
pinin4fjords: Done!

subworkflows/nf-core/dia_proteomics_analysis/nextflow.config
pinin4fjords: temp yes, not sure about the names (see other comment)

subworkflows/nf-core/dia_proteomics_analysis/nextflow.config
pinin4fjords: Yep, again, this is nf-core standards. We keep modules simple and portable, push complexity like this out to config. Modules shouldn't make assumptions on `meta` keys, but calling workflows can.

subworkflows/nf-core/dia_proteomics_analysis/main.nf
pinin4fjords: Fixed now!

modules/nf-core/diann/main.nf
pinin4fjords: Addressed!

modules/nf-core/diann/meta.yml
pinin4fjords: addressed

subworkflows/nf-core/dia_proteomics_analysis/main.nf
pinin4fjords: That's basically what this is doing with the `+`

subworkflows/nf-core/dia_proteomics_analysis/nextflow.config
pinin4fjords: Agreed, that was my bad, fixed.

subworkflows/nf-core/dia_proteomics_analysis/main.nf
pinin4fjords: Don't think this structure should cause a problem for that

subworkflows/nf-core/dia_proteomics_analysis/main.nf
pinin4fjords: My understanding is as follows, please correct me if I'm wrong:

While empirical assembly does use --use-quant --temp ./quant/, it also uses --rt-profiling and --gen-spec-lib, which require DIA-NN to access the raw MS data to build the empirical spectral library from actual measurements. Only final quantification can use just file names because it's purely aggregating pre-computed .quant results without needing to access raw data.

subworkflows/nf-core/dia_proteomics_analysis/main.nf
pinin4fjords: I added a commit for this: https://github.com/nf-core/modules/pull/9173/commits/949280172582a023aa495929c441cc1fc39386df

... but I'm a little bit skeptical. It seems neater to do the conversions externally and keep the subworkflow simple. What do you think?

modules/nf-core/diann/main.nf
DongzeHE: This looks good! Thanks!

modules/nf-core/diann/meta.yml
DongzeHE: You are absolutely correct! Thanks for the explanation!

subworkflows/nf-core/dia_proteomics_analysis/main.nf
DongzeHE: I think you are absolutely correct. Thanks for pointing this out!

subworkflows/nf-core/dia_proteomics_analysis/main.nf
DongzeHE: Ahhh I think both ways work. I raised this up because I had hard a time finding the correct way to convert .raw to .mzml when building my first linux DIANN module.

I am Ok with separating the file conversion step with the DIANN subworkflow, but we should definitely write a note about this in the documentation so that ppl are aware of the converter module we provide.

subworkflows/nf-core/dia_proteomics_analysis/main.nf
```diff
@@ -0,0 +1,454 @@
...
+    ch_empirical_input = ch_preliminary_analysis_input    // [meta_input_fasta, ms_file, speclib]
+        .join(DIANN_PRELIMINARYANALYSIS.out.diann_quant)  // [meta_input_fasta, ms_file, speclib, diann_quant]
+        .map { [it[0].experiment] + it.drop(1)}           // [meta_exp_searchdb, ms_file, speclib, diann_quant]
+        .groupTuple()                                     // [meta_exp_searchdb, [ms_file], [speclib], [diann_quant]]
+        .map{ sortListsByPathName(it, 1) }                // [meta_exp_searchdb, [sorted_ms_file], [sorted_speclib], [sorted_diann_quant]]
+        .map { meta, ms_files, speclib, diann_quant ->
+            [meta, ms_files, [], [], speclib, diann_quant]    // DIANN module input: add ms_file_names and fasta placeholders
+        }
+
+    DIANN_ASSEMBLEEMPIRICALLIBRARY(ch_empirical_input)
```
DongzeHE, line 315: I have one more thought here: In all-in-one DIANN, the quantification quant tables generated from the first pass search results, including gg_matrix.tsv, pg_matrix.tsv and unique_genes_matrix.tsv, are reported explicitly. 

Sometimes, we found these tables very important in quality assessment and downstream data analysis, because, by compared with MBR results, we know how much we gain/lose from the second pass search, and even sometimes use the first-pass results in downstream analysis and discard the second-pass results. 

So, do you think we should generate these first-pass quant tables in this subworkflow? If the answer is yes, I think we can either add these quantification-related parameters into DIANN_ASSEMBLEEMPIRICALLIBRARY, or call DIANN_FINALQUANTIFICATION twiec using the `.quant`  files from the first pass search and second pass search, respectively?

subworkflows/nf-core/dia_proteomics_analysis/main.nf
pinin4fjords: OK, I've reverted the conversion, and added that note to the meta.yml

subworkflows/nf-core/dia_proteomics_analysis/main.nf
pinin4fjords: I feel like this is extension work that deviates from how the quantms workflow operates?

I don't 100% follow your suggestions in any case, I would need your help for implementation.

Maybe we could merge this as a foundation of behaviour roughtly equivalent to how quantms does it, and then build from there via subsequent PRs?

DongzeHE: Sounds good!

jonasscheid (approved):

jonasscheid:
LGTM

I hope DIA-NN implements some sort of an authentication mechanism (tokens) in the future that could be passend by env variable. Then we could use the most recent versions 🙏🏼

@vdemichev are there any plans towards that?

vdemichev:
Hi All,

A general comment: since 2.0 we keep all inputs/outputs stable & backward compatible, i.e. even .quant files are backward compatible. So in the future it's very likely that newer versions will work without changing any code in the pipeline. That said, if one wishes to implement InfinDIA, this will probably result in very minor changes (it creates '.pre.quant' files instead of '.quant' files; also does not require spectral library as input), but this is like sth separate that should exist in addition to the main workflow - just for super large search spaces.

@jonasscheid, you mean like AI models? :) 

Best,
Vadim

jonasscheid:
Thanks for the quick response @vdemichev and good to know that input output structures will remain quite stable.

>@jonasscheid, you mean like AI models? :)

No by token I meant some sort of a license key that can be purchased for DIA-NN (or free license key for academy members). In the grand scheme of things that would enable the distribution of DIA-NN containers (and evenutally conda) that could be used in workflows like these here.

vdemichev:
Well, this is exactly how the Enterprise version works :) But redistribtion is an issue regardless, so we don't enable it. I would think it's OK to just always have the pipeline download DIA-NN & create container on the fly? This apparently is not possible in Galaxy, but here should work just fine?

pinin4fjords:
> Well, this is exactly how the Enterprise version works :) But redistribtion is an issue regardless, so we don't enable it. I would think it's OK to just always have the pipeline download DIA-NN & create container on the fly? This apparently is not possible in Galaxy, but here should work just fine?

Not how things work right now, unfortunately. Upgrades right now will have to work via folks building their containers externally, and then they'd override the `container` directives.

We have other commercial software here, which work via license keys. The software won't work without them, but you can still download the containers etc. Sentieon is an example: https://github.com/nf-core/modules/blob/master/modules/nf-core/sentieon/bwamem/main.nf. Maybe a model worth considering. I think Sentieon provide a license key to allow the CI testing.

vdemichev:
> Not how things work right now, unfortunately.

But that's how it works for quantms, seems all is OK with it?

> Maybe a model worth considering

Will not work in our case unfortunately.

pinin4fjords:
> > Not how things work right now, unfortunately.
> 
> But that's how it works for quantms, seems all is OK with it?

This nf-core repo has more structured requirements, we have a standardised approach to container provisioning etc and I don't _think_ this will fit, though I'll ask around.

PR #9185, opened 2025-10-10
Add fastq seqkit sana pair

vagkaratzas:
New `fastq` file sanitization/pre-processing subworkflow, to be used instead of just `seqkit/sana`, that automatically breaks paired-end reads (`seqkit/sana` can only process one file at a time), mixes with single-end reads in one channel, runs `seqkit/sana` and finally merges back original pairs with `seqkit/pair`.

## PR checklist

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/subworkflows/nf-core/fastq_sanitise_seqkit/main.nf
+++ b/subworkflows/nf-core/fastq_sanitise_seqkit/main.nf
@@ -14,8 +14,14 @@ workflow FASTQ_SANITISE_SEQKIT {
         // seqkit/sana can only receive one file at a time
         .flatMap  { meta, reads ->
             if (meta.single_end) {
+                if (reads instanceof List && reads.size() != 1) {
+                    error("Error: Check your meta.single_end value. Single-end reads should contain one file only.")
+                }
                 return [[ meta + [strandness: 'single'], reads ]]
             } else {
+                if (!(reads instanceof List) || reads.size() != 2) {
+                    error("Error: Check your meta.single_end value. Paired-end reads should contain two files; a forward and a reverse.")
+                }
                 return [
                     [ meta + [strandness: 'R1'], reads[0] ],
                     [ meta + [strandness: 'R2'], reads[1] ]


subworkflows/nf-core/fastq_seqkit_sana_pair/main.nf
```diff
@@ -0,0 +1,57 @@
...
+    ch_reads = ch_reads
+        .branch {
+            meta, fastq ->
+                single_end: meta.single_end
+                    return [ meta, fastq ]
+                paired_end: !meta.single_end
+                    return [ meta, fastq ]
+        }
+
+    ch_reads = ch_reads.single_end.mix(ch_reads.paired_end.transpose()) // seqkit/sana can only receives one file at a time
```
prototaxites: I think you can just transpose the input reads channel for the same result?

subworkflows/nf-core/fastq_seqkit_sana_pair/main.nf
```diff
@@ -0,0 +1,57 @@
...
+workflow FASTQ_SEQKIT_SANA_PAIR {
+
+    take:
+    ch_reads // channel: [ val(meta), [ fastq ] ]
+
+    main:
+
+    ch_versions = Channel.empty()
+
+    ch_reads = ch_reads
```
prototaxites: Overwriting subworkflow inputs here

subworkflows/nf-core/fastq_seqkit_sana_pair/main.nf
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+    ch_sanitized_paired_reads = ch_sanitized_reads.paired_end
+        .map { meta, reads ->
+            def sorted_reads = reads.sort { it.size() } // sorting by file size, to keep snapshot order same
+            def renamed = sorted_reads.indexed().collect { i, file_path ->
+                def base = file(file_path).getSimpleName()
+                def ext  = file(file_path).getName() - file(file_path).getSimpleName()
+                def new_name = "${base}_${i + 1}${ext}"
+                file(file_path).copyTo(new_name)
```
prototaxites: Is there a reason to rename files here, rather than specifying the output prefix in a supplied with the subworkflow nextflow.config file?

subworkflows/nf-core/fastq_seqkit_sana_pair/main.nf
vagkaratzas: Oh my..! Indeed, you are a genius! xD

subworkflows/nf-core/fastq_seqkit_sana_pair/main.nf
vagkaratzas: Yee, won't need the original input later on

subworkflows/nf-core/fastq_seqkit_sana_pair/main.nf
vagkaratzas: For paired-end reads, because they share the same meta id, `seqkit/sana` will always produce an output file with the same name for both. This block of code ensures they will be copied and renamed *_1 and *_2 no matter what. Couldn't think of a better way to implement this. We could change the meta id after we transpose and before `sana`, but then it will require even more complicated computations to figure out how to merge back, I think..

subworkflows/nf-core/fastq_sanitise_seqkit/main.nf
jfy133, line 1: Based on: https://nf-co.re/docs/guidelines/components/subworkflows#name-format-of-subworkflow-files

I think ti should be called `fastq_sanitise_pair_seqkit` or something

subworkflows/nf-core/fastq_seqkit_sana_pair/main.nf
```diff
@@ -0,0 +1,46 @@
+include { SEQKIT_SANA } from '../../../modules/nf-core/seqkit/sana/main'
+include { SEQKIT_PAIR } from '../../../modules/nf-core/seqkit/pair/main'
+
+workflow FASTQ_SEQKIT_SANA_PAIR {
+
+    take:
+    ch_reads // channel: [ val(meta), [ fastq ] ]
+
+    main:
+
```
jfy133:
```suggestion
```

subworkflows/nf-core/fastq_seqkit_sana_pair/meta.yml
```diff
@@ -0,0 +1,42 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/subworkflows/yaml-schema.json
+name: "fastq_seqkit_sana_pair"
+description: Sanitize and pair FASTQ files using seqkit sana and seqkit pair
+keywords:
+  - fastq
+  - quality control
+  - sanitization
```
jfy133: What does sanitised mean here?

Maybe other stuff like `repair` or `clean` would be helpful for non-native English speakers

subworkflows/nf-core/fastq_seqkit_sana_pair/meta.yml
```diff
@@ -0,0 +1,42 @@
...
+        Structure: [ val(meta), [ fastq ] ]
+        Where meta is a map containing at least:
+        - id: sample identifier
+        - single_end: boolean indicating if data is single-end (true) or paired-end (false)
+      pattern: "*.{fastq,fastq.gz,fq,fq.gz}"
+output:
+  - reads:
+      type: channel
+      description: |
+        Channel containing sanitized and paired FASTQ files.
```
jfy133: What does sanitised mean here?

subworkflows/nf-core/fastq_sanitise_seqkit/tests/main.nf.test
jfy133, line 1: If you are testing single-end, is the `pair` correct in teh name?

subworkflows/nf-core/fastq_seqkit_sana_pair/main.nf
jfy133: I am with Jim here, copyTo will result a huge amount of IO on the headnode which is not what we want.

I would rather update the subworkflow to update the meta during transpose to add a meta field to tag that it's either `R1` or `R2`, use that in the `prefix` when executing teh module, then drop the meta when grouping back together again

prototaxites: Agreed, I was going to suggest the same solution! Maybe you can `flatMap` the input reads channel to do this effectively in a single operation.

subworkflows/nf-core/fastq_seqkit_sana_pair/main.nf
prototaxites: It's probably paranoia on my end and Nextflow handles this fine, but I'm always very wary of overwriting channels 😅 

Personally would just give it a new name but fine as-is if you prefer it!

vagkaratzas: I am my own garbage collector.

subworkflows/nf-core/fastq_seqkit_sana_pair/main.nf
vagkaratzas: And I agree with both of you. Will change

subworkflows/nf-core/fastq_sanitise_seqkit/main.nf
vagkaratzas: Maybe go with `fastq_sanitise_seqkit` since pair is optional (only for paired-end)?

subworkflows/nf-core/fastq_seqkit_sana_pair/meta.yml
vagkaratzas: Updated

subworkflows/nf-core/fastq_seqkit_sana_pair/meta.yml
vagkaratzas: Updated

subworkflows/nf-core/fastq_sanitise_seqkit/tests/main.nf.test
vagkaratzas: pair is used only for paired-end reads, so will probably rename to `fastq_sanitise_seqkit`

subworkflows/nf-core/fastq_seqkit_sana_pair/main.nf
vagkaratzas: Done

subworkflows/nf-core/fastq_seqkit_sana_pair/main.nf
vagkaratzas: but changed

subworkflows/nf-core/fastq_sanitise_seqkit/main.nf
vagkaratzas: Done

subworkflows/nf-core/fastq_sanitise_seqkit/main.nf
```diff
@@ -0,0 +1,52 @@
...
+        // seqkit/sana can only receive one file at a time
+        .flatMap  { meta, reads ->
+            if (meta.single_end) {
+                return [[ meta + [strandness: 'single'], reads ]]
+            } else {
+                return [
+                    [ meta + [strandness: 'R1'], reads[0] ],
+                    [ meta + [strandness: 'R2'], reads[1] ]
+                ]
+            }
```
prototaxites, line 29: So I think this section is fine as-is - but I just want to check if we need to catch other input problems. It's possible that a channel entry with `meta.single_end` set might have multiple read fastqs, for example. Do we want to make some guarantees about the structures - like checking that for single end, reads is not a list, and for paired end, that reads is a list of size 2?

subworkflows/nf-core/fastq_sanitise_seqkit/tests/nextflow.config
```diff
@@ -0,0 +1,11 @@
+process {
+
+    withName: SEQKIT_SANA {
+        ext.prefix = { "${meta.id}_${meta.strandness}" }
+    }
+
+    withName: SEQKIT_PAIR {
+        ext.args   = '-u'
+    }
```
prototaxites, line 9: Can you include this also in a `nextflow.config` file next to `main.nf`? I would imagine these args are important to set for the subworkflow to run, so we should increase their prominence with an includable config file.

prototaxites (approved):
LGTM! 

Two minor comments - I don't think they're blocking, but definitely the config file one would IMO be nice-to-have.

subworkflows/nf-core/fastq_sanitise_seqkit/tests/nextflow.config
vagkaratzas: Good one, will do, but only for the `SEQKIT_SANA`, the `SEQKIT_PAIR` one shouldn't be mandatory

subworkflows/nf-core/fastq_sanitise_seqkit/main.nf
vagkaratzas: We could, and exit with an error if not, I guess?

prototaxites: Sounds sensible!

subworkflows/nf-core/fastq_sanitise_seqkit/main.nf
```diff
@@ -0,0 +1,52 @@
...
+
+    // Add strandness information to meta
+    ch_reads_with_strandness = ch_reads
+        // seqkit/sana can only receive one file at a time
+        .flatMap  { meta, reads ->
+            if (meta.single_end) {
+                return [[ meta + [strandness: 'single'], reads ]]
+            } else {
+                return [
+                    [ meta + [strandness: 'R1'], reads[0] ],
```
jfy133, line 26:
```suggestion
                    [ meta + [strandness: 'R1'], reads.sort()[0] ],
```

Maybe? For safety? IdK, optional

subworkflows/nf-core/fastq_sanitise_seqkit/main.nf
```diff
@@ -0,0 +1,52 @@
...
+            def clean_meta = meta.findAll { key, value -> key != 'strandness' }
+            return [ clean_meta, fastq ]
+        }
+        .groupTuple(by: 0)
+        .branch {
+            meta, fastq ->
+                single_end: meta.single_end
+                    return [ meta, fastq ]
+                paired_end: !meta.single_end
+                    return [ meta, fastq ]
```
jfy133, line 47: Maybe 
```suggestion
                    return [ meta, fastq.sort() ]
```

Again to make sure R1 is coming first?

subworkflows/nf-core/fastq_sanitise_seqkit/main.nf
prototaxites: IMO correct sorting of the reads should be the responsibility of the calling pipeline. There’s a hypothetical user out there whose forward reads are called foo.fq and their reverse reads are called baz.fq…

vagkaratzas: Good point Jim..I'll leave the sorting out

subworkflows/nf-core/fastq_sanitise_seqkit/main.nf
vagkaratzas: Done and tested!

vagkaratzas:
Thank you both!

PR #9220, opened 2025-10-21
update module dysgu to dysgu/run

nvnieuwk:
Renames the `DYSGU` module to `DYSGU_RUN` to enable creation of other subtools in the dysgu toolchain


nvnieuwk:
Pinging @poddarharsh15 and @famosab here to let you guys know that the name of this module will be changed

PR #497, opened 2021-05-12
Update comments with new style

JoseEspinosa:
As explained on this issue #493 comment style has been updated.

## PR checklist

Closes #493 

- [x] This comment contains a description of changes (with reason).

PR #9304, opened 2025-10-29
fix: gprofiler2 output files missing gene names in intersection columns #497

mohe1linux:
The gprofiler2 module outputs were missing the actual gene names/IDs in the expected columns, making it impossible to identify which specific genes contribute to pathway enrichment.

Expected behavior:

*.gprofiler2.all_enriched_pathways.tsv should contain an intersection column with gene names/IDs
*.gprofiler2.[source].sub_enriched_pathways.tsv should contain actual gene names in the DE_genes_names column

Actual behavior:

all_enriched_pathways.tsv file lacks the intersection column entirely
sub_enriched_pathways.tsv files have DE_genes_names column containing numeric values (same as DE_genes) instead of gene names

Now with the fix 

Enable g:Profiler evidence codes so the `intersection` column is emitted.
Populate sub-tables with *both* Ensembl IDs and symbols:
DE_genes_ids`  = original `intersection` IDs
DE_genes_names` = gene symbols (from DE table where available, else gprofiler2::gconvert), fallback to IDs if unmapped

nextflow run . -profile test,docker
--gprofiler2_run true
--gprofiler2_organism mmusculus
--gprofiler2_evcodes true
--outdir test_gprofile_symbols

- `*all_enriched_pathways.tsv` now contains `intersection`.
- `*sub_enriched_pathways.tsv` now has `DE_genes_ids` and `DE_genes_names` (symbols present; IDs used as fallback).

**Notes**
- No changes to output file names besides adding `DE_genes_ids` in sub tables.

### DIFF
--- a/modules/nf-core/gprofiler2/gost/templates/gprofiler2_gost.R
+++ b/modules/nf-core/gprofiler2/gost/templates/gprofiler2_gost.R
@@ -39,14 +39,14 @@
 #' @return named list of options and values similar to optparse
 
 parse_args <- function(x) {
-    args_list <- unlist(strsplit(x, ' ?--')[[1]])[-1]
-    args_vals <- lapply(args_list, function(x) scan(text=x, what='character', quiet = TRUE))
-
-    # Ensure the option vectors are length 2 (key/ value) to catch empty ones
-    args_vals <- lapply(args_vals, function(z) { length(z) <- 2; z})
-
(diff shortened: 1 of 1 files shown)


famosab:
Can you have a look at the tests for this module? Because I am confused that this did not lead to a change in md5sum because you literally changed what goes in the file :D

PR #9346, opened 2025-11-05
Swap tabix to only one output channel

SPPearce:
The tabix modules produce either a tbi or csi index.
Very large genomes (mostly plant genomes) can't be indexed with a .tbi index, they have to have a .csi. 
This means subworkflows and pipelines (such as sarek for instance) that currently only use the .tbi index don't actually work with e.g. wheat, as they haven't joined the .tbi and .csi channels, even if you give them additional arguments to make the csi via a config.
This PR changes the output to whichever of them is generated into a single channel rather than two separate ones. 
[This is consistent with the guidelines](https://nf-co.re/docs/guidelines/components/modules#one-output-channel-per-output-file-type), which have stated (since last summer): 

> In some cases the file format can be different between files of the same type or for the same function (e.g. indices: .bai and .crai). These different file formats SHOULD be part of the same output channel since they are they serve the same purpose and are mutually exclusive.


SPPearce:
Hmm, seems to be failing on singularity for one of the subworkflows (VCF_ANNOTATE_ENSEMBLVEP_SNPEFF)

PR #9350, opened 2025-11-06
change in staging - to avoid summary.txt editing (ref)

EladH1:
## PR checklist

Closes #XXX 

[+] This comment contains a description of changes (with reason).
Description: Fixed BBSplit staging issues by adding stageInMode 'copy' and removing complex timestamp fixing logic that was causing problems with staged index files.
Reason: The previous script had staging issues that prevented proper index file handling.
The code of bbsplit modified in place the .summary.txt --> sometimes even deleting the last modified line.
This change also improves the resume of bbsplit as the .summary.txt file stay the same.

[+ ] If you've fixed a bug or added code that should be tested, add tests!
I fixed a bug in the BBSplit module. This is now run better
- [- ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [x] Add a resource `label`
  - For modules:
 INFO     All tests passed!                                 

    Test Process BBMAP_BBSPLIT                                                                                                                                  │
│                                                                                                                                                             │
│   Test [b500f41c] 'sarscov2_se_fastq_fasta_chr22_fasta - index' PASSED (17.659s)                                                                            │
│   Test [684e797b] 'sarscov2_se_fastq_fasta_chr22_fasta - index - stub' PASSED (12.41s)                                                                      │
│   Test [2b2e4760] 'sarscov2_se_fastq_fasta_chr22_fasta' PASSED (18.886s)                                                                                    │
│   Test [5fe89f57] 'sarscov2_se_fastq_fasta_chr22_fasta - stub' PASSED (15.803s)             
    INFO     All tests passed!

### DIFF
--- a/modules/nf-core/strobealign/meta.yml
+++ b/modules/nf-core/strobealign/meta.yml
@@ -10,14 +10,14 @@ keywords:
   - bam
   - sam
 tools:
-- "strobealign":
-    description: "Align short reads using dynamic seed size with strobemers"
-    homepage: "https://github.com/ksahlin/strobealign"
-    documentation: "https://github.com/ksahlin/strobealign?tab=readme-ov-file#usage"
-    tool_dev_url: "https://github.com/ksahlin/strobealign"
-    doi: "10.1186/s13059-022-02831-7"
-    licence: ['MIT']
-    identifier: biotools:strobealign
+  - "strobealign":
+      description: "Align short reads using dynamic seed size with strobemers"
+      homepage: "https://github.com/ksahlin/strobealign"
+      documentation: "https://github.com/ksahlin/strobealign?tab=readme-ov-file#usage"
+      tool_dev_url: "https://github.com/ksahlin/strobealign"
+      doi: "10.1186/s13059-022-02831-7"
+      licence: ["MIT"]
+      identifier: biotools:strobealign
 
 input:
   - - meta:
@@ -136,7 +136,7 @@ output:
         ontologies:
           - edam: http://edamontology.org/format_3750 # YAML
 authors:
-- "@matthdsm"
+  - "@matthdsm"
 maintainers:
-- "@matthdsm"
-- "@nvnieuwk"
+  - "@matthdsm"
+  - "@nvnieuwk"


modules/nf-core/bbmap/bbsplit/tests/main.nf.test
```diff
@@ -132,7 +132,7 @@ nextflow_process {
 
             assertAll(
                 { assert process.success },
-                { assert path(process.out.log[0][1]).text.contains("If you wish to regenerate the index") },
+                { assert path(process.out.log[0][1]).text.contains("already exists; skipping merge") },
```
pinin4fjords, line 135: This was specifically testing for index re-use, why are you changing it? If the index is not longer being re-used, then this has broken the fix in https://github.com/nf-core/modules/pull/5005

pinin4fjords (changes requested):
I totally agree that this module should be modified to prevent it from making changes to input files, happy to help with that (and my bad for making that error).

BUT I believe this has completely stripped my fix in https://github.com/nf-core/modules/pull/5005 that allowed index re-use. Yes, it's very, very (VERY) ugly but it took me forever to craft and get it working, and I don't want to lose it. You've even updated the test that showed you you'd broken index re-use! 

I'll suggest an alternate fix.

pinin4fjords:
I've suggested an alternate fix in https://github.com/nf-core/modules/pull/9356. If those tests pass, I think that would be a better way to go.

**Edit:** confirmed that #9356 fixes the issue of editing inputs, and preserves index re-use at least. @EladH1 could you verify that fix in your use case please?

modules/nf-core/bbmap/bbsplit/tests/main.nf.test
EladH1: Hi,
Im not trying to diminish your work - just want it to use bbsplit with existing index. 
The current fix [https://github.com/nf-core/modules/pull/5005] - doesn't work in current version of the pipeline.
when you feed it with index. (the index saved contains summary.txt)

also the test not working when you **include** "summary.txt" in the outputs.

we can work together to improve 
Best,
Elad

pinin4fjords: Sorry, the point I was trying to get across is that the change in the logging output, which caused you to change that message, indicates that the module is NOT using the existing index any more (which is what the former message was diagnostic for). This indicates that your change has in fact broken the usage of input indices and bbsplit is re-indexing.

I'm reasonably confident this other PR will do the trick now: https://github.com/nf-core/modules/pull/9356

It excludes the index from outputs when an input index is provided, and prevents the module from editing the input index in-place. I believe that should resolve your issues without breaking the index re-use.

SPPearce:
Resolved in #9356

PR #9356, opened 2025-11-07
Fix BBSplit to prevent input file modification

pinin4fjords:
## Problem

The current BBSplit module modifies input index files in-place, which causes issues when those files are symlinked (Nextflow's default staging behavior). The timestamp-fixing logic directly modifies `summary.txt` files in the staged index directory, potentially affecting the original files in the work directory.

PR #9350 attempted to fix this with `stageInMode 'copy'` globally, but that approach removed the essential timestamp/path fixing logic that prevents BBSplit from re-indexing on every run (the original issue fixed in PR #5005).

## Solution

This PR takes a surgical approach that preserves the fix from PR #5005 while addressing the input modification concern:

### Key Changes

1. **Explicit index naming** (main.nf:13): 
   ```groovy
   path index, name: 'input_index'
   ```
   Stages the index with an explicit name `input_index` to prevent naming conflicts with the output index directory. Previously, the input index could be staged as `bbsplit`, causing the modified input to be incorrectly captured as output.

2. **Copy before modify** (main.nf:76):
   ```bash
   rm -rf index_writable && cp -rL input_index index_writable
   ```
   Creates a writable copy of the input index before any modifications, protecting the original input files. The `-L` flag ensures symlinks are followed and actual content is copied.

3. **Fix paths and timestamps** (main.nf:78-83):
   Updates both paths and timestamps in summary.txt so BBSplit can find the reference files in the copied location and recognizes the index as valid

4. **Use the copy** (main.nf:59):
   ```groovy
   index_files = "path=index_writable"
   ```
   BBSplit uses the copied `index_writable` directory instead of the input

5. **Atomic index creation** (main.nf:53, 99-106):
   When building a new index:
   - BBSplit builds into temporary `bbsplit_build` directory
   - Fix paths in summary.txt to reference final `bbsplit_index` name
   - Atomically rename `bbsplit_build` to `bbsplit_index`
   - Ensures Nextflow only sees fully-populated index, preventing race conditions

6. **Test improvements**:
   - Renamed tests to clearly indicate when prebuilt index is used ("build index" vs "split with prebuilt index")
   - Fixed stub logic to only create bbsplit_index directory when actually building an index
   - Corrected test inputs: When using a prebuilt index for splitting, FASTA files are no longer passed (they were redundant)
   - **Significantly reduced snapshot size**: The "split with prebuilt index" test snapshot now correctly contains only splitting outputs (fastq files, stats, versions) instead of incorrectly capturing index files. This is a direct result of fixing the input/output naming conflict - the prebuilt index is no longer incorrectly captured as an index output.
   - Removed snapshot assertions from stub tests (stubs now only verify process success)

## Benefits

✅ **Preserves PR #5005's fix**: BBSplit reuses indices without re-indexing  
✅ **Addresses PR #9350's concern**: Input files are never modified in-place  
✅ **Surgical approach**: Only copies what needs to be modified (the index), not all inputs  
✅ **Clear data flow**: `input_index` → `index_writable` (with fixes) → BBSplit uses `index_writable`  
✅ **Better test clarity**: Test names and structure clearly show different usage scenarios  
✅ **Cleaner test logic**: Removed complex file filtering code; snapshots now correctly distinguish between index inputs and outputs  
✅ **Atomic operations**: Index building uses atomic rename to prevent partial captures

## Relationship to Other PRs

- **PR #5005** (merged): Added timestamp/path fixing logic to prevent re-indexing. **This PR preserves and extends that fix**.
- **PR #9350** (open): Identified the input modification issue but removed the timestamp fixing logic. **This PR provides the correct solution**.

## Testing

The existing tests verify:
- ✅ Pre-built indices are reused (check log for "If you wish to regenerate the index")
- ✅ No input file modification occurs (symlinked inputs remain unmodified)
- ✅ Resume functionality works correctly
- ✅ Index building scenarios work as before

## PR checklist

- [x] Add a resource `label`

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/modules/nf-core/bbmap/bbsplit/main.nf
+++ b/modules/nf-core/bbmap/bbsplit/main.nf
@@ -73,7 +73,7 @@ process BBMAP_BBSPLIT {
     # disturbed, which bbsplit doesn't like. Fix the time stamps in its summaries.
     # This needs to be done via Java to match what bbmap does.
     if [ "$use_index" == "true" ]; then
-        rm -rf index_writable &&  cp -rL input_index index_writable
+        cp -rL input_index index_writable
 
         for summary_file in \$(find index_writable/ref/genome -name summary.txt); do
             # Extract the path from summary.txt and update it to point to index_writable


pinin4fjords:
Something's up, working on it...

SPPearce:
Oh, why we have to do such things to get tools to work in sensible ways...

pinin4fjords:
> Oh, why we have to do such things to get tools to work in sensible ways...

Yep, I'm not proud....

EladH1:
So why not work with https://github.com/nf-core/modules/pull/9350 ??

less fix on fix. 
leave the input as input and not modify in place
there is clear understanding when things go wrong.
the tests shows the summary changes (https://github.com/nf-core/modules/pull/9356) makes bugs.

pinin4fjords:
Sorry @EladH1 , this is the right way to do it, working around the limitations of BBSplit (which is is what creates the need).

The modification is non-negotiable, it's what allows the index to be re-used. We just need to make sure the modifications aren't disruptive (which is what I'm doing here)

modules/nf-core/bbmap/bbsplit/main.nf
```diff
@@ -67,15 +68,18 @@ process BBMAP_BBSPLIT {
...
 
-    # When we stage in the index files the time stamps get disturbed, which
-    # bbsplit doesn't like. Fix the time stamps in its summaries. This needs to
-    # be done via Java to match what bbmap does
+    # If using a pre-built index, copy it to avoid modifying input files in place,
+    # then fix timestamps. When we stage in the index files the time stamps get
+    # disturbed, which bbsplit doesn't like. Fix the time stamps in its summaries.
+    # This needs to be done via Java to match what bbmap does.
+    if [ "$use_index" == "true" ]; then
+        rm -rf index_writable &&  cp -rL input_index index_writable
```
SPPearce: Why are we `rm -rf` this folder?

pinin4fjords: Yeah, this can go, was being paranoid

pinin4fjords:
```suggestion
        cp -rL input_index index_writable
```

PR #9360, opened 2025-11-07
Add module: `TRIDENT_FETCH`

TCLamnidis:
This PR adds modules for `trident`:
 - [x] trident fetch
 - [ ] trident validate
 - [ ] trident forge

## PR checklist

Closes #XXX 

- [x] ~Emit the `versions.yml` file.~ Emit versions topic.
- [x] Add a resource `label`
  - For modules:
  - For subworkflows:


modules/nf-core/trident/fetch/tests/main.nf.test
```diff
@@ -0,0 +1,104 @@
...
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
+            )
+        }
+
+    }
+
+    test("2012_MeyerScience-2.1.1 - fetchString - stub") {
```
itrujnara, line 151: A test for `fetchFile` is missing

modules/nf-core/trident/fetch/main.nf
```diff
@@ -0,0 +1,70 @@
+process TRIDENT_FETCH {
+    tag ""
```
itrujnara: Can we somehow infer the tag from the input values?

modules/nf-core/trident/fetch/main.nf
```diff
@@ -0,0 +1,70 @@
...
+    label 'process_single'
+
+    conda "${moduleDir}/environment.yml"
+    container "${workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container
+        ? 'https://depot.galaxyproject.org/singularity/poseidon-trident:1.6.7.1--hebebf5b_1'
+        : 'biocontainers/poseidon-trident:1.6.7.1--hebebf5b_1'}"
+
+    input:
+    tuple path(archive_dir), val(fetch_s), path(fetch_fn)
```
itrujnara: Are you sure that a meta map will never be useful here?

modules/nf-core/trident/fetch/meta.yml
```diff
@@ -0,0 +1,101 @@
...
+        type: file
+        description: The POSEIDON.yml file of any newly downloaded package(s).
+        pattern: "output_archive/*/POSEIDON.yml"
+        ontologies: []
+  geno:
+    - "output_archive/*/*.{geno,bed,vcf,geno.gz,bed.gz,vcf.gz}":
+        type: file
+        description: The genotype files of any newly downloaded package(s).
+        pattern: "output_archive/*/*.{geno,bed,vcf,geno.gz,bed.gz,vcf.gz}"
+        ontologies: []
```
itrujnara: Some outputs (at least this one) should have some suitable ontology. Can you please check?

itrujnara (commented):
Hi @TCLamnidis, nice submission. I have left some minor comments.

SPPearce:
@TCLamnidis , are you planning on finishing this off?

TCLamnidis:
yes! I've just been tied up on eager atm 🙈

modules/nf-core/trident/fetch/main.nf
TCLamnidis: good idea. adding that now

modules/nf-core/trident/fetch/meta.yml
TCLamnidis: added!

modules/nf-core/trident/fetch/main.nf
TCLamnidis: The way the tool runs, there is never a reason to split it up into subsets of runs. 
You would always want to merge all your requests before running the module, and run the tool once to get any datasets you're missing.
So I doubt there's any use to a meta for the input.

modules/nf-core/trident/fetch/tests/main.nf.test
TCLamnidis: Added!

TCLamnidis:
TODO: To follow nf-core guidelines I will need to split the `geno`, `snp`, and `ind` output channels by the output file type. This will make all trident modules more complicated to implement into pipelines overall, but it would be the only way to fit within nf-core.

TCLamnidis:
Thank you for the review @itrujnara, and sorry for taking a while to get back to it.
I updated the module according to the review, and the nf-core guidelines. 
I believe the failed linting is an issue with nf-core/tools (see https://github.com/nf-core/tools/issues/3901), but happy to update the meta.yml if not. :)

modules/nf-core/trident/fetch/main.nf
```diff
@@ -0,0 +1,64 @@
+process TRIDENT_FETCH {
+    tag "${fetch_fn ?: ''} ${fetch_s ?: ''}"
+    label 'process_single'
+
+    conda "${moduleDir}/environment.yml"
+    container "${workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container
+        ? 'https://depot.galaxyproject.org/singularity/poseidon-trident:1.6.7.1--hebebf5b_1'
+        : 'biocontainers/poseidon-trident:1.6.7.1--hebebf5b_1'}"
```
SPPearce:
```suggestion
        : 'quay.io/biocontainers/poseidon-trident:1.6.7.1--hebebf5b_1'}"
```

modules/nf-core/trident/fetch/tests/main.nf.test
```diff
@@ -0,0 +1,182 @@
...
+                    [], // fetchFile
+                ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
```
SPPearce, line 31:
```suggestion
                { assert snapshot(sanitizeOutput(process.out)).match() }
```

modules/nf-core/trident/fetch/tests/main.nf.test
```diff
@@ -0,0 +1,182 @@
...
+                    )
+                }
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
```
SPPearce, line 64:
```suggestion
                { assert snapshot(sanitizeOutput(process.out)).match() }
```

modules/nf-core/trident/fetch/tests/main.nf.test
```diff
@@ -0,0 +1,182 @@
...
+                        tuple(archive_dir, [], fetch_fn)
+                    }
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
```
SPPearce, line 109:
```suggestion
                { assert snapshot(sanitizeOutput(process.out)).match() }
```

modules/nf-core/trident/fetch/tests/main.nf.test
```diff
@@ -0,0 +1,182 @@
...
+                tuple(archive_dir, fetchString, [])
+                }
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
```
SPPearce, line 151:
```suggestion
                { assert snapshot(sanitizeOutput(process.out)).match() }
```

modules/nf-core/trident/fetch/tests/main.nf.test
```diff
@@ -0,0 +1,182 @@
...
+                    [], // fetchFile
+                ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
```
SPPearce, line 176:
```suggestion
                { assert snapshot(sanitizeOutput(process.out)).match() }
```

TCLamnidis:
The linting issue I mentioned (described in [this issue](https://github.com/nf-core/tools/issues/3901)) is still happening with nf-core tools 2.1.0.

Not sure what I can do here...
Maybe if you are happy with the manual meta.yml I have (only diverging form the automated one in the `{geno,snp,ind}_*` attributes of the output), we can ignore linting for the module's meta somehow?

itrujnara (approved):
It's been a marathon, but nothing more from me, good to go once CI passes.

SPPearce:
Can you try `path("output_archive/*/*{bed,bed.gz}")` instead of `path "output_archive/*/*{bed,bed.gz}" `, I think that might help.
Also, can you reorder the meta.yml and main.nf to be in the same order for the output channels

TCLamnidis:
`path("output_archive/*/*{bed,bed.gz}")` has the same automatic linting output, and fails in the same way. 
Reordering output channels now

TCLamnidis:
@SPPearce @itrujnara 
After asking around in slack, I am considering restructuring the output in a way that is less standard, which makes sense for the module. 
In particular, instead of each output file being in its own channel, a single output channel is emitted, that has the entire package contents as one element (multiple elements for multiple packages).

I originally avoided it based on this directive from the [tutorial on writing nf-core modules](https://nf-co.re/docs/developing/tutorials/writing-nf-core-modules/5-writing-modules#the-output-block):
> Give each file type its own emit entry, usually one per format.

But that approach would definitely simplify downstream processing with other trident modules, since they all ultimately work with Poseidon packages, which are entire directories, not single files.

Before I go off working on that, would you object to such a solution?

SPPearce:
No, not at all, I think it would be sensible to just emit the whole lot ;)

TCLamnidis:
I updated the module outputs. Now the process emits 2 channels, one containing the input archive (if any, otherwise the `output_archive` directory), and one listing the downloaded packages.

In irl operations, the intended publishDir where trident should fetch packages would be provided as the archive_dir in the input as well, ensuring that packages only ever get downloaded once and only if missing. The pickup of `output_archive` is a workaround, to allow testing that no download happens with existing packages, without needing to create a dummy archive directory within nf-test. 

Conceptually, this should allow users to set the archive, run through fetch, and use the `local_archive` emission downstream, ensuring that downstream processes only ran after the needed packages have been downloaded by fetch.

PR #9363, opened 2025-11-07
Updates for souporcell

LuisHeinzlmeier:
- add `clusters` to the first input tuple as it belongs to the same `meta` as `bam` and `barcode`
- change output files
- add testing for `ext.args`
- unfortunately, `souporcell` does not seem to have a flag for its version ([see this issue](https://github.com/wheaton5/souporcell/issues/262))
- add the version of `souporcell` as discussed [here](https://nf-co.re/docs/guidelines/components/modules#emission-of-versions) in the nf-core documentation

### DIFF
--- a/modules/nf-core/diamond/blastp/environment.yml
+++ b/modules/nf-core/diamond/blastp/environment.yml
@@ -4,4 +4,4 @@ channels:
   - conda-forge
   - bioconda
 dependencies:
-  - bioconda::diamond=2.1.12
+  - bioconda::diamond=2.1.16
--- a/modules/nf-core/diamond/blastp/main.nf
+++ b/modules/nf-core/diamond/blastp/main.nf
@@ -4,8 +4,8 @@ process DIAMOND_BLASTP {
 
     conda "${moduleDir}/environment.yml"
     container "${workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container
-        ? 'https://depot.galaxyproject.org/singularity/diamond:2.1.12--hdb4b4cc_1'
-        : 'biocontainers/diamond:2.1.12--hdb4b4cc_1'}"
+        ? 'https://depot.galaxyproject.org/singularity/diamond:2.1.16--h13889ed_0'
+        : 'biocontainers/diamond:2.1.16--h13889ed_0'}"
 
     input:
     tuple val(meta), path(fasta)
--- a/modules/nf-core/diamond/blastp/tests/main.nf.test.snap
+++ b/modules/nf-core/diamond/blastp/tests/main.nf.test.snap
@@ -29,7 +29,7 @@
                     
                 ],
                 "7": [
-                    "versions.yml:md5,75db7b2f0c2a5129e5a67e014f19a597"
+                    "versions.yml:md5,b48dcfbb4862d283c42c972c5001db4c"
                 ],
                 "blast": [
                     
@@ -55,7 +55,7 @@
                     ]
                 ],
                 "versions": [
-                    "versions.yml:md5,75db7b2f0c2a5129e5a67e014f19a597"
+                    "versions.yml:md5,b48dcfbb4862d283c42c972c5001db4c"
                 ],
                 "xml": [
                     
@@ -64,9 +64,9 @@
         ],
         "meta": {
             "nf-test": "0.9.2",
-            "nextflow": "25.04.2"
+            "nextflow": "25.10.0"
         },
-        "timestamp": "2025-06-05T10:51:07.898268369"
+        "timestamp": "2025-11-11T20:02:54.776422616"
     },
     "txt_gz": {
         "content": [
@@ -98,7 +98,7 @@
                     
                 ],
                 "7": [
-                    "versions.yml:md5,75db7b2f0c2a5129e5a67e014f19a597"
+                    "versions.yml:md5,b48dcfbb4862d283c42c972c5001db4c"
                 ],
                 "blast": [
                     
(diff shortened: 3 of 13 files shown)


modules/nf-core/souporcell/main.nf
```diff
@@ -37,21 +38,25 @@ process SOUPORCELL {
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
-        souporcell: \$(souporcell_pipeline.py --version 2>&1 || echo "N/A")
+        souporcell: $VERSION
+        gxx: \$( g++ --version | head -n1 | awk '{print \$NF}' )
```
SPPearce: Hmm, is this just the version number? On mac I get `(clang-1500.3.9.4)`.

nictru: Yes on linux and in the container I get clean version strings (e.g. `9.5.0`, no parenthesis).

But I think the version capture for `gxx` is not very necessary here - the reason why it is in the `environment.yml` is because it is required at runtime in certain situations, and the conda recipe is not configured properly. Also I think the version of `gxx` does not affect reproducibility.

LuisHeinzlmeier: Thanks for reviewing!

If I just run `g++ --version` in the pipeline, I get (Mac):
```
g++ (conda-forge gcc 9.5.0-19) 9.5.0
Copyright (C) 2019 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```
So I would say that it should be fine. Should we keep the version from gxx or remove it?

nictru: Okay so given your terminal output we can see that `g++ --version | head -n1 | awk '{print $NF}'` will evaluate to `9.5.0` which is good so we know that it is consistent across OS. I think the way `g++` is installed affects this, but as in the context of nf-core `g++` will always come from installation via conda, it will be consistent.

So in this case I think you can keep it, what do you think @SPPearce?

modules/nf-core/souporcell/tests/main.nf.test
```diff
@@ -17,14 +28,66 @@ nextflow_process {
...
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(
+                    process.out.tsv,
+                    process.out.versions,
```
SPPearce, line 47: This will capture the versions output as a string in plain-text, so it is easy to verify.
```suggestion
                    path(process.out.versions[0]).yaml,
```

modules/nf-core/souporcell/main.nf
SPPearce: I would agree that we don't need to capture the version here, and the pinning is only required in the conda recipe to fix broken dependencies.
However, if you do want to keep it, that is fine, but can you explicitly capture the version in the nf-test snapshot at that point.
There are two ways to do that.
The first is to use `path(process.out.versions[0]).yaml` (see the suggestion).
The second is to move to using a `topic` channel, which we have just started the implementation of. See [this PR](https://github.com/nf-core/website/pull/3613) for the website - [the preview is here](https://deploy-preview-3613--nf-core-main-site.netlify.app/blog/2025/version_topics).

LuisHeinzlmeier: Perhaps I expressed myself somewhat ambiguously. `g++ --version` inside the container looks like this,
```
g++ (conda-forge gcc 9.5.0-19) 9.5.0
Copyright (C) 2019 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```
 while the `g++ --version` on Mac looks like this:
```
 Apple clang version 17.0.0 (clang-1700.0.13.5)
Target: arm64-apple-darwin24.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```
 Thank you for your feedback. Which option do you prefer, @nictru?

nictru: the only way this will be executed outside of a container is if the pipeline is executed on conda. So what you should test is the same command in the conda env of the process. What you sent in your last message was the log message of your global `g++` installed at `/Library/Developer/CommandLineTools/usr/bin` which is not the same as would be used in an actual nextflow execution.

However I think you should just remove it, since even if we find out that it would work, nobody would benefit from having this

modules/nf-core/souporcell/main.nf
```diff
@@ -37,21 +38,24 @@ process SOUPORCELL {
...
-    touch ${prefix}/cluster_genotypes.vcf
     touch ${prefix}/clusters.tsv
+    touch ${prefix}/cluster_genotypes.vcf
+    touch ${prefix}/ambient_rna.txt
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
-        souporcell: \$(souporcell --version)
+        souporcell: $VERSION
+        gxx: \$( g++ --version | head -n1 | awk '{print \$NF}' )
```
nictru: Also remove it from the `stub`

nictru:
@LuisHeinzlmeier if you still could [do this](https://github.com/nf-core/modules/pull/9363#discussion_r2509669886) then I think we're good to go. It's nice to have because it allows to see easily which versions changed

modules/nf-core/souporcell/tests/main.nf.test
LuisHeinzlmeier: Included in `test("homo sapiens - bam and barcodes")` and `test("homo sapiens - bam and barcodes - ext.args")`!

nictru:
I think you need to have both `process.out.versions` and the `path(process.out.versions[0]).yaml` since apparently the linting is failing currently - sorry for the inconvenience

LuisHeinzlmeier:
Ah, I see. No problem, thanks for your help!

LuisHeinzlmeier:
LGTM now! Ready to merge?

PR #9398, opened 2025-11-13
Syntax updates to the picard modules

nvnieuwk:
This PR updates the syntax of all Picard modules to fix warnings and errors of the language server

### DIFF
--- a/modules/nf-core/picard/crosscheckfingerprints/main.nf
+++ b/modules/nf-core/picard/crosscheckfingerprints/main.nf
@@ -22,8 +22,8 @@ process PICARD_CROSSCHECKFINGERPRINTS {
     def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
 
-    def input1_cmd = input1.collect{"--INPUT $it"}.join(' ')
-    def input2_cmd = input2.collect{"--SECOND_INPUT $it"}.join(' ')
+    def input1_cmd = input1.collect{ f1 -> "--INPUT $f1"}.join(' ')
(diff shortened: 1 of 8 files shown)


modules/nf-core/picard/crosscheckfingerprints/main.nf
```diff
@@ -22,8 +22,8 @@ process PICARD_CROSSCHECKFINGERPRINTS {
     def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
 
-    def input1_cmd = input1.collect{"--INPUT $it"}.join(' ')
-    def input2_cmd = input2.collect{"--SECOND_INPUT $it"}.join(' ')
+    def input1_cmd = input1.collect{ f1 -> "--INPUT $f1"}.join(' ')
+    def input2_cmd = input2.collect{ f2 -> "--SECOND_INPUT $f2"}.join(' ')
```
maxulysse, line 26: why `f1` and `f2` and not `file` or `input` or whatever?

nvnieuwk: doesn't really matter much here :) it's short for `file_input1` and `file_input2`

maxulysse: I see, but I'd rather be explicit to be honest, or we should have a proper guidelines on what to do.
That's the kind of thing that will tend to diverge from one person to the other

nvnieuwk: Agh sorry I didn't see your response here :/ Guidelines on this would be good!

PR #9405, opened 2025-11-14
add eval as channel type

mashehu:
useful for topic channels

### DIFF
--- a/modules/meta-schema.json
+++ b/modules/meta-schema.json
@@ -42,7 +42,7 @@
             "type": "array",
             "description": "Extra arguments for the module",
             "items": {
-                "type" : "object",
+                "type": "object",
                 "properties": {
                     "description": {
                         "type": "string",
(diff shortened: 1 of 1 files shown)


modules/meta-schema.json
```diff
@@ -55,24 +55,26 @@
...
+                            "type": "object",
+                            "patternProperties": {
+                                ".*": {
+                                    "$ref": "#/definitions/elementProperties"
+                                }
+                            }
+                        }
+                    },
+                    {
                         "type": "object",
```
mirpedrol, line 71: is keeping this object outside of an array ok?

mashehu: git just shows it in a weird way, what actually changed is that the double curly braces on line 75 were resovled by prettier

mirpedrol: ok! I was confused by the github diff

PR #9412, opened 2025-11-17
update topics structure in meta.yml

mirpedrol:


modules/nf-core/bamstats/generalstats/meta.yml
```diff
@@ -49,15 +49,15 @@ output:
...
+  versions:
+    - - process:
+          type: string
+          description: The process the versions were collected from
+      - tool:
+          type: string
+          description: The tool name
+      - version:
+          type: string
+          description: The version of the tool
```
mashehu:
```suggestion
    - - ${task.process}:
          type: string
          description: The name of the process
      - bamstats:
          type: string
          description: The name of the tool
      - 'bamstats --version | grep "version: " | sed -e s"/version: //':
          type: eval
          description: The expression to obtain the version of the tool
```
that's what we have in the meta.yml template

PR #9416, opened 2025-11-17
run pre-commit rules on all files

mashehu:


maxulysse:
Did a quick review, all seems to be just lint as expected.
I'd say let's merge as is

PR #9420, opened 2025-11-18
lint: fix linting warnings in GAWK module

prototaxites:
Fix linting warnings identified by `nextflow lint` in the GAWK module.

## PR checklist

Closes #XXX 

- [ ] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [ ] Remove all TODO statements.
- [ ] Emit the `versions.yml` file.
- [ ] Follow the naming conventions.
- [ ] Follow the parameters requirements.
- [ ] Follow the input/output options guidelines.
- [ ] Add a resource `label`
- [ ] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - For modules:
    - [ ] `nf-core modules test <MODULE> --profile docker`
    - [ ] `nf-core modules test <MODULE> --profile singularity`
    - [ ] `nf-core modules test <MODULE> --profile conda`
  - For subworkflows:
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile docker`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile singularity`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile conda`

PR #9435, opened 2025-11-21
Refactor GAWK implicit closures

DLBPointon:
Using `nextflow lint` I get `Implicit closure parameter is deprecated, declare an explicit parameter instead`

Granted i'm using nf 25.10 and testing for 26 so this isn't _needed_ now but it doesn't hurt.


prototaxites:
I think your branch is out of date, I fixed this already: https://github.com/nf-core/modules/pull/9420

DLBPointon:
Yeah, I was planning on doing more modules. But ran out of time. Thanks for sorting it!

PR #9453, opened 2025-11-25
fix prek in fix-linting action

mashehu:
switch also fix linting action to check/fix all files

### DIFF
--- a/nf-test.config
+++ b/nf-test.config
@@ -13,13 +13,12 @@ config {
 
     // load the necessary plugins
     plugins {
-        load "nft-bam@0.5.0"
-        load "nft-compress@0.1.0"
-        load "nft-vcf@1.0.7"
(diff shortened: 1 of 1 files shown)


nf-test.config
```diff
@@ -13,13 +13,12 @@ config {
...
     // load the necessary plugins
     plugins {
-        load "nft-bam@0.5.0"
-        load "nft-compress@0.1.0"
-        load "nft-vcf@1.0.7"
+        load "nft-anndata@0.1.0"
+        load "nft-bam@0.6.0"
         load "nft-csv@0.1.0"
-        load "nft-utils@0.0.7"
+        load "nft-compress@0.1.0"
```
maxulysse, line 19: can we get sorting?

mashehu: did that in a parallel commit 😁

PR #9455, opened 2025-11-25
fix issue 9424: MultiQC report handling for cloud based workDir

### LINKED ISSUE #9424: Sendmail issue when running on AWS batch with S3 as workDir

### Description of the bug

In the `utils_nfcore_pipeline` workflow, the `completionEmail` function is defined to send email to users when the pipeline run completes. However, it will not work when the pipeline runs on AWS batch using S3 as workDir.

Specifically, in line https://github.com/nf-core/modules/blob/master/subworkflows/nf-core/utils_nfcore_pipeline/main.nf#L277 , `mqc_report` is the a path.

When running on local or HPC, the path is the local path which is accessible by the head job.

When running on AWSbatch using S3 as the workDir, suppose the workDir is "s3://BUCKET/scratch", the path will be something like: "/BUCKET/scratch/4dmcVVKo90LpUC/12/3d1d669c37458ddf6cd8c53ebf0520/multiqc_report.html" Note that the "S3:/" prefix is gone now as that is the virtual local path for worker EC2. They are mounted as a local path, and it is accessible to worker EC2. But the head job cannot access that file with a local path.

I tried some debug script:
```
    def mqc_report = getSingleReport(multiqc_report)

    log.info("MultiQC report: ${mqc_report}")
    log.info("Type of MultiQC report: ${mqc_report.getClass()}")
    log.info("MultiQC report absolute path: ${mqc_report.toAbsolutePath()}")
    log.info("MultiQC report URI: ${mqc_report.toUriString()}")
```

They gave me:

> MultiQC report: /BUCKET/scratch/4dmcVVKo90LpUC/12/3d1d669c37458ddf6cd8c53ebf0520/multiqc_report.html
Type of MultiQC report: class nextflow.cloud.aws.nio.S3Path
MultiQC report absolute path: /BUCKET/scratch/4dmcVVKo90LpUC/12/3d1d669c37458ddf6cd8c53ebf0520/multiqc_report.html
MultiQC report URI: s3://BUCKET/scratch/4dmcVVKo90LpUC/12/3d1d669c37458ddf6cd8c53ebf0520/multiqc_report.html

I also checked "s3://BUCKET/scratch/4dmcVVKo90LpUC/12/3d1d669c37458ddf6cd8c53ebf0520/multiqc_report.html" is the acutal file which exists.

Then I did:
```
            mqc_report_file = new File(mqc_report.toUriString())
            log.info("MultiQC report file path: ${mqc_report_file.getAbsolutePath()}")
            log.info("MultiQC report file size: ${mqc_report_file.length()} bytes")
```

or 

```
            mqc_report_file = new File(mqc_report)
            log.info("MultiQC report file path: ${mqc_report_file.getAbsolutePath()}")
            log.info("MultiQC report file size: ${mqc_report_file.length()} bytes")
```
 It gave me the same:

> MultiQC report file path: /s3:/BUCKET/scratch/4dmcVVKo90LpUC/12/3d1d669c37458ddf6cd8c53ebf0520/multiqc_report.html
MultiQC report file size: 0 bytes

This file path does not look right. It is "/s3:/". I did not change it.

The reason that I call `new File` first is because `def mqcFileObj = new File("$mqcFile")` does not work in the "sendmail_template.txt" as in the wrapper the File class is the native Java file class, which only works for local path not S3.

Anyway, the send mail function will fail silentily as in "sendmail_template.txt", `mqcFile` is not empty but `mqcFileObj` is empty in `def mqcFileObj = new File("$mqcFile")`

Would like to hear others' thoughts about this.

The workaround I made it work on AWS batch with S3 as workDir is to do the following.
I had to create a new temp file to make it work.

```
    if (mqc_report != null) {
        try {
            // Check if we're using S3 workDir and need to download file
            if (workflow.workDir.scheme == 's3' && mqc_report.toUriString().startsWith('s3://')) {
                def tempFile = new File("${workflow.workDir}/multiqc_report.html")
                tempFile.text = mqc_report.text
                mqc_report_file = tempFile.absolutePath
            } else {
                // Local path - just pass the path string (template will create File object)
                mqc_report_file = mqc_report.toString()
            }
        } catch (Exception e) {
            log.warn("Error accessing MultiQC report: ${e.message}")
        }
    }
```

```console

```

### Relevant files

_No response_

### System information

Nextflow version 25.04

SZhengP:
MultiQC report handling for AWS batch with S3 workDir

Fix https://github.com/nf-core/modules/issues/9424

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/subworkflows/nf-core/utils_nfcore_pipeline/main.nf
+++ b/subworkflows/nf-core/utils_nfcore_pipeline/main.nf
@@ -276,6 +276,26 @@ def completionEmail(summary_params, email, email_on_fail, plaintext_email, outdi
     // On success try attach the multiqc report
     def mqc_report = getSingleReport(multiqc_report)
 
+    // Fix issue https://github.com/nf-core/modules/issues/9424 when running on AWS batch with S3 workDir. mqc_report is not downloaded locally.
+    def mqc_report_file = null
+
+    if (mqc_report != null) {
+        try {
+            // Check if we're using S3 for the report file and need to download it so the head node can access it
+            if (mqc_report.toUriString().startsWith('s3://')) {
+                def tempFile = file("${workflow.launchDir}/multiqc_report_${workflow.sessionId}.html")
+                mqc_report.copyTo(tempFile)
+                mqc_report_file = tempFile.toString()
+            } else {
+                // Local path - just pass the path string (template will create File object)
+                mqc_report_file = mqc_report.toString()
+            }
+        } catch (Exception e) {
+            log.warn("Error accessing MultiQC report: ${e.message}")
+        }
+    }
+
+
     // Check if we are only sending emails on failure
     def email_address = email
     if (!email && email_on_fail && !workflow.success) {
@@ -295,7 +315,7 @@ def completionEmail(summary_params, email, email_on_fail, plaintext_email, outdi
 
     // Render the sendmail template
     def max_multiqc_email_size = (params.containsKey('max_multiqc_email_size') ? params.max_multiqc_email_size : 0) as MemoryUnit
-    def smail_fields           = [email: email_address, subject: subject, email_txt: email_txt, email_html: email_html, projectDir: "${workflow.projectDir}", mqcFile: mqc_report, mqcMaxSize: max_multiqc_email_size.toBytes()]
+    def smail_fields           = [email: email_address, subject: subject, email_txt: email_txt, email_html: email_html, projectDir: "${workflow.projectDir}", mqcFile: mqc_report_file, mqcMaxSize: max_multiqc_email_size.toBytes()]
     def sf                     = new File("${workflow.projectDir}/assets/sendmail_template.txt")
     def sendmail_template      = engine.createTemplate(sf).make(smail_fields)
     def sendmail_html          = sendmail_template.toString()
--- a/subworkflows/nf-core/utils_nfcore_pipeline/main.nf
+++ b/subworkflows/nf-core/utils_nfcore_pipeline/main.nf
@@ -276,13 +276,14 @@ def completionEmail(summary_params, email, email_on_fail, plaintext_email, outdi
     // On success try attach the multiqc report
     def mqc_report = getSingleReport(multiqc_report)
 
-    // Fix issue https://github.com/nf-core/modules/issues/9424 when running on AWS batch with S3 workDir. mqc_report is not downloaded locally.
+    // Fix issue https://github.com/nf-core/modules/issues/9424 when running on cloud based executor such as AWS batch with S3 workDir.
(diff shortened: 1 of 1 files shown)


SPPearce (commented):
@SZhengP , is this still needed?

SZhengP:
Hi @SPPearce ,

Yes. This is still needed.
But I am hesitating which way to go to fix the nf-core lint issue.

It looks like that for some subworks (like [utils_nfcore_pipeline](https://github.com/nf-core/modules/tree/master/subworkflows/nf-core/utils_nfcore_pipeline/tests) and [utils_nextflow_pipeline](https://github.com/nf-core/modules/tree/master/subworkflows/nf-core/utils_nextflow_pipeline/tests)) the `tests` folder for nf-test always has:

- `main.function.nf.test`
- `main.workflow.nf.test`

For modules, pipelines and other subworkslow, the `tests` folder always has `main.nf.test`.

We can either **add `main.nf.test` for the subworkflows** or **modify nf-core tools lint logic**.

It is a design question.
Would like to hear some suggestions here @mashehu @mirpedrol.

SPPearce:
I think @maxulysse is keen to remove these subworkflows entirely in favour of a plugin, does the plugin work correctly?

maxulysse:
I've been using the plugin in question in rnavar and no issues encountered so far

SZhengP:
The plug-in solution looks like a cleaner solution.
I took a look of the plug-in codes, which have the same issue. 
I could make PR for the plug-in too.

But I wonder is there an annoucement or roadmap to switch from subworkflows to one plug-in?
As for most of nf-core pipelines, they are still using the subworkflows.
We had to made the changes internally to multiple pipelines to make sending email work with AWSbatch & S3 workDir.

I feel that before the switch is made in nf-core tool pipeline template, we do need the changes in the subworkflows too.

SPPearce:
I suspect this is no longer required and can be closed?

SZhengP:
> I suspect this is no longer required and can be closed?

The email sending is still broken.
Still a lot of pipelines use this subworkflow. 
I still think this should be merged.

SZhengP:
Also, the plug-in is more broken than this subworkflow when it comes to the `mqc_report` handling when using cloud executor. I made [another issue](https://github.com/nf-core/nf-core-utils/issues/65) in the plug-in repo

maxulysse:
> Also, the plug-in is more broken than this subworkflow when it comes to the `mqc_report` handling when using cloud executor. I made [another issue](https://github.com/nf-core/nf-core-utils/issues/65) in the plug-in repo

Thanks for the update, I'll investigate and report ASAP

Issue #9456, opened 2025-11-25
new module: PBMARKDUP


sainsachiko:
### Is there an existing module for this?

- [x] I have searched for the existing module

### Is there an open PR for this?

- [x] I have searched for existing PRs

### Is there an open issue for this?

- [x] I have searched for existing issues

### Further Information

**Description:** pbmarkdup identifies and marks duplicate reads in PacBio HiFi (CCS) data. It clusters highly similar CCS reads to detect PCR duplicates and flags them in the BAM output (duplicate bit 0x400), optionally removing duplicates. 
**Homepage:** https://github.com/PacificBiosciences/pbmarkdup
**Documentation:** https://github.com/PacificBiosciences/pbmarkdup
**Available via Bioconda:** https://bioconda.github.io/

### Are you going to work on this?

- [x] If I'm planning to work on this module, I added myself to the `Assignees` to facilitate tracking who is working on the module

PR #9457, opened 2025-11-25
Add module PBMARKDUP

sainsachiko:
Add [pbmarkdup](https://github.com/PacificBiosciences/pbmarkdup) to identifies and marks duplicate reads in PacBio HiFi (CCS) data. 
https://github.com/nf-core/modules/issues/9456
Thank you for reviewing!

## PR checklist

Closes #XXX 

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/pbmarkdup/meta.yml
+++ b/modules/nf-core/pbmarkdup/meta.yml
@@ -12,7 +12,7 @@ tools:
   - pbmarkdup:
       description: |
         pbmarkdup identifies and marks duplicate reads in PacBio HiFi (CCS) data. It clusters
-        highly similar CCS reads to detect PCR duplicates and flags them in the output files 
+        highly similar CCS reads to detect PCR duplicates and flags them in the output files
         (BAM,FASTQ,FASTA) (duplicate bit 0x400), optionally removing duplicates.
         (duplicate bit 0x400), optionally removing duplicates.
       homepage: https://github.com/PacificBiosciences/pbmarkdup


modules/nf-core/pbmarkdup/meta.yml
```diff
@@ -0,0 +1,62 @@
...
+keywords:
+  - markdup
+  - bam
+  - fastq
+  - fasta
+tools:
+  - pbmarkdup:
+      description: |
+        pbmarkdup identifies and marks duplicate reads in PacBio HiFi (CCS) data. It clusters
+        highly similar CCS reads to detect PCR duplicates and flags them in the BAM output
```
DLBPointon: Change the bit about it being BAM output, as there are 3 formats it can output.

modules/nf-core/pbmarkdup/main.nf
```diff
@@ -0,0 +1,50 @@
...
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/pbmarkdup:1.2.0--h9ee0642_0' :
+        'biocontainers/pbmarkdup:1.2.0--h9ee0642_0' }"
+
+    input:
+    tuple val(meta), path(input)
+
+    output:
+    tuple val(meta), path("${prefix}.${suffix}"), emit: markduped
+    path "versions.yml"                         , emit: versions
```
DLBPointon, line 17: I feel like the `--dup-file dups.fasta` flag needs better support, otherwise it won't get caught and output.

Perhaps an output channel, and something to capture the name the user will use in the config?

```
output:
tuple val(meta), path("${dup_file}"), optional: true, emit: duplicates

script:

// This little chunk would be soley to have a string to give to the output tuple
// not needed in the script as it exists in the args.
def matcher = (task.ext.args =~ /--dup-file\s+(\S+)/)
dup_file = matcher.find() ? matcher[0][1] : ""

"""
    pbmarkdup \\
        -j ${task.cpus} \\
        $input \\
        ${prefix}.${suffix} \\
        $args
"""
```
Feel free to get another reviewer here, but otherwise you can't capture the dupes file unless you simplify your existing output by removing the suffix, but then you have to deal with a tuple[meta, file, file] in the workflow.

modules/nf-core/pbmarkdup/main.nf
```diff
@@ -0,0 +1,50 @@
...
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args  = task.ext.args  ?: ''
+
```
DLBPointon, line 29: pbmarkdup can take a list of files as input, so i think you may just need to do some double checks to make sure names don't conflict.

Make sure input and output file names don't conflict, which seems like it could happen.

Adapted from CAT_CAT
```
    if(file_list.contains("${prefix}.{suffix}") {
        error "PBMARKDUP: The name of the input file can't be the same as the output: " +
        "Change the prefix to avoid conflict."
    }
```

As it can take a list of input files, perhaps a check to make sure they are in fact unique files too? However i think this can be optional as it should be checked in the input_check or workflow of the pipeline.

```
    def input_files = [input].collect { file.baseName }
    if (input_files.size() != input_files.unique().size()) {
        error "PBMARKDUP: Input files must have unique names. Found duplicates: ${input_files}" +
        "Check your input reads"
    }
```

DLBPointon (commented):
làm tốt lắm!

I just have a few comments to flesh it out more.

It'll mean adding another test, sorry. Feel free to get another opinion though.

modules/nf-core/pbmarkdup/main.nf
sainsachiko: Thanks @DLBPointon for kindly reviewing this. I have updated the code regarding your reviews, please take a look, thank you!

modules/nf-core/pbmarkdup/meta.yml
```diff
@@ -0,0 +1,89 @@
...
```
sainsachiko:
```suggestion
        highly similar CCS reads to detect PCR duplicates and flags them in the output files 
        (BAM,FASTQ,FASTA) (duplicate bit 0x400), optionally removing duplicates.
```

PR #9472, opened 2025-11-27
qsv/cat: add option to skip input format checking

dialvarezs:
By default, qsv infers input formats from file extensions. If the extension isn’t supported, it errors out, even with a custom `--delimiter`. Currently, the only way to bypass this is via an environment variable.

I think this behavior shouldn’t always be enabled as it can prevent errors, and don't think it's common enough for a dedicated input, so I’ve exposed it through `ext` instead.

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/qsv/cat/tests/main.nf.test
+++ b/modules/nf-core/qsv/cat/tests/main.nf.test
@@ -23,6 +23,7 @@ nextflow_process {
                 ]
                 input[1] = "rows"
                 input[2] = "csv"
+                input[3] = false
                 """
             }
         }
@@ -50,6 +51,7 @@ nextflow_process {
                 ]
                 input[1] = "rows"
                 input[2] = "tsv"
+                input[3] = false
                 """
             }
         }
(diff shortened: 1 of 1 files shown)


modules/nf-core/qsv/cat/main.nf
```diff
@@ -22,7 +22,12 @@ process QSV_CAT {
     script:
     prefix = task.ext.prefix ?: "${meta.id}"
     def args = task.ext.args ?: ''
+    // QSV by default checks input format based on the file extension,
+    // and the only way to skip this check is to set an env variable
+    def skip_format_check_cmd = task.ext.skip_format_check ? 'export QSV_SKIP_FORMAT_CHECK=1' : ''
```
jfy133: We don't want to support custom `exts` as they are not supported in documentation etc, if it is really necessary I would do this via another boolean input channel.

dialvarezs: I was not 100% sure either, but I have seen this pattern in other modules, so that's why I went with this.
Will do by boolean input instead.

dialvarezs: It should be ready now.

PR #9474, opened 2025-11-28
feat(fastq_qc_trim_filter_setstrandedness): Add Bowtie2 as alternative rRNA removal tool

pinin4fjords:
## Summary

Add bowtie2 as a third option for rRNA removal alongside sortmerna and ribodetector, providing users with more flexibility in choosing their preferred alignment-based rRNA filtering approach.

### Paired-end Read Handling
For paired-end reads, proper rRNA removal requires filtering pairs where **either** mate aligned to rRNA. The implementation:
1. Aligns reads with BOWTIE2_ALIGN_PE
2. Filters BAM with `samtools view -f 12` (flag 12 = both mates unmapped)
3. Converts filtered BAM back to FASTQ with `samtools fastq`

### Single-end Read Handling
For single-end reads, bowtie2's `--un-gz` output is used directly via `save_unaligned=true`.

### U→T Conversion
rRNA reference FASTAs may contain uracil (U) since they're RNA sequences, but sequencing reads contain thymine (T). The implementation converts U→T when building the bowtie2 index.

## Changes

- Add `BOWTIE2_ALIGN`, `BOWTIE2_ALIGN_PE`, and `BOWTIE2_BUILD` module imports
- Add `SAMTOOLS_VIEW` and `SAMTOOLS_FASTQ` for paired-end filtering
- Add `ch_bowtie2_index` input channel for pre-built indexes
- Add `make_bowtie2_index` parameter for on-the-fly index building
- Implement single-end/paired-end branching with appropriate filtering strategy
- Update meta.yml with bowtie2 in `ribo_removal_tool` enum and new components
- Add test cases for both paired-end and single-end bowtie2 rRNA removal
- Update nextflow.config with bowtie2 and samtools process configurations

## Test Results

| Test | Input | After Trim | rRNA Detected | Output |
|------|-------|------------|---------------|--------|
| PE fastp bowtie2 | 4169 pairs | 1137 pairs | 17 mates (10 pairs) | 1127 pairs |
| SE fastp bowtie2 | 4169 reads | 4162 reads | 10 reads | 4152 reads |

The paired-end bowtie2 test achieves the same output (1127 pairs) as sortmerna, correctly removing all 10 synthetic rRNA pairs.

## Test plan

- [x] Run existing sortmerna and ribodetector tests to ensure no regression
- [x] Run paired-end bowtie2 test
- [x] Run single-end bowtie2 test
- [x] Verify bowtie2 log output is captured for MultiQC

## Proposed refactor before merge

> **Note:** This subworkflow has grown substantially with three rRNA removal tools (sortmerna, ribodetector, bowtie2), each with their own index building and filtering logic. I propose we factor out the rRNA removal components into a dedicated `fastq_remove_rrna` subworkflow before merging this PR, once we confirm bowtie2 should be retained as a method.

### DIFF
--- a/modules/nf-core/ribodetector/tests/main.nf.test
+++ b/modules/nf-core/ribodetector/tests/main.nf.test
@@ -29,8 +29,8 @@ nextflow_process {
                 { assert process.success },
                 { assert process.out.fastq },
                 { assert process.out.log },
-                { assert path(process.out.log[0][1]).getText().contains("Writing output non-rRNA sequences") },
-                { assert snapshot(process.out.versions).match() }
+                { assert path(process.out.log[0][1]).getText().contains("Writing output non-rRNA sequences") }
+                // Note: versions collected via topic, not snapshotted
             )
         }
 
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,3 @@
 {
-    "nextflow.formatting.harshilAlignment": true,
-    "nextflow.telemetry.enabled": false
+    "nextflow.formatting.harshilAlignment": true
 }
--- a/modules/nf-core/ribodetector/main.nf
+++ b/modules/nf-core/ribodetector/main.nf
@@ -4,8 +4,8 @@ process RIBODETECTOR {
 
 	conda "${moduleDir}/environment.yml"
 	container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
-		'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/4d/4de8fe74d21198e6fc8218cb3209d929b3d7dab750678501b096b0ccc324307b/data':
-		'community.wave.seqera.io/library/ribodetector:0.3.2--cbe1c77fa14eeb53' }"
+        'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/4d/4de8fe74d21198e6fc8218cb3209d929b3d7dab750678501b096b0ccc324307b/data' :
+        'community.wave.seqera.io/library/ribodetector:0.3.2--cbe1c77fa14eeb53' }"
 
 	input:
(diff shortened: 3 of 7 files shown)


suhrig (approved):
LGTM! Thanks for this great addition (at amazing speed)!

I agree that moving the rRNA removal into a separate space would be sensible, but if bandwidth is limited, then it should be okay the way it is.

pinin4fjords:
@suhrig Factoring out complete!

The failure in CI is because the factoring-out triggered the issue described in https://github.com/hzi-bifo/RiboDetector/pull/59 for ribodetector (I think). I think we'll need that fix to propagate to Conda channels before we can set a seed for testing purposes and this can be completed. 

Edit: I also hadn't sorted the reads

modules/nf-core/ribodetector/tests/main.nf.test
```diff
@@ -29,8 +29,8 @@ nextflow_process {
                 { assert process.success },
                 { assert process.out.fastq },
                 { assert process.out.log },
-                { assert path(process.out.log[0][1]).getText().contains("Writing output non-rRNA sequences") },
-                { assert snapshot(process.out.versions).match() }
+                { assert path(process.out.log[0][1]).getText().contains("Writing output non-rRNA sequences") }
+                // Note: versions collected via topic, not snapshotted
```
maxulysse, line 33: We should be snapshoting the versions. There's a snippet for that

maxulysse (approved):

modules/nf-core/ribodetector/tests/main.nf.test
pinin4fjords: sorry, it was set to auto-merge! Will pr a fix

pinin4fjords: https://github.com/nf-core/modules/pull/9478

PR #9482, opened 2025-12-01
plastid metagene_generate, make_wiggle, psite

suhrig:
This PR adds plastid with three submodules:
- metagene_generate
- psite
- make_wiggle

These functions are required to compute p-sites from Ribo-seq data.

I have also added test data here: https://github.com/nf-core/test-datasets/pull/1809

## PR checklist

- [X] Add a resource `label`
  - For modules:

### DIFF
--- a/modules/nf-core/strdrop/call/environment.yml
+++ b/modules/nf-core/strdrop/call/environment.yml
@@ -4,4 +4,7 @@ channels:
   - conda-forge
   - bioconda
 dependencies:
-  - bioconda::trgt=5.0.0
+  - conda-forge::pip=25.3
+  - conda-forge::python=3.14.1
+  - pip:
+      - strdrop==0.3
--- a/modules/nf-core/strdrop/call/main.nf
+++ b/modules/nf-core/strdrop/call/main.nf
@@ -1,0 +1,51 @@
+process STRDROP_CALL {
+    tag "$meta.id"
+    label 'process_single'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/4f/4fc3114f6d67fadb826981d29ae67f8564ee81283184893c2db677d919b5b32d/data':
+        'community.wave.seqera.io/library/pip_strdrop:df8d5dc993ea6848' }"
+
+    input:
+    tuple val(meta), path(vcf)
+    tuple val(meta2), path(training_set_json)
+    tuple val(meta3), path(training_set_vcfs, stageAs: 'input/*')
+
+    output:
+    tuple val(meta), path("*.vcf.gz"), emit: vcf
+    tuple val("${task.process}"), val('strdrop'), eval("strdrop --version | sed 's/.* //g'"), topic: versions, emit: versions_strdrop
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args         = task.ext.args ?: ''
+    def prefix       = task.ext.prefix ?: "${meta.id}"
+    def training_set = training_set_json ? "--training-set ${training_set_json}" : '--training-set ./input'
+
+    if (training_set_json && training_set_vcfs) {
+        error("Please provide only one of 'training_set_json' or 'training_set_vcfs' as training set input.")
+    }
+    """
+    strdrop \\
+        call \\
+        $args \\
+        $training_set \\
+        $vcf \\
+        ${prefix}.vcf.gz
+    """
+
+    stub:
+    def args   = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+
+    if (training_set_json && training_set_vcfs) {
+        error("Please provide only one of 'training_set_json' or 'training_set_vcfs' as training set input.")
+    }
+    """
+    echo $args
+
+    echo | gzip > ${prefix}.vcf.gz
+    """
+}
--- a/modules/nf-core/strdrop/call/meta.yml
+++ b/modules/nf-core/strdrop/call/meta.yml
@@ -1,0 +1,90 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/meta-schema.json
+name: "strdrop_call"
+description: Detect drops in STR coverage
+keywords:
+  - str
+  - vcf
+  - lrs
+tools:
+  - "strdrop":
+      description: "Flag STR coverage drops in LRS data"
+      homepage: "https://github.com/dnil/strdrop"
+      documentation: "https://github.com/dnil/strdrop/blob/main/docs/README.md"
+      tool_dev_url: "https://github.com/dnil/strdrop"
+      licence: ["MIT"]
+
+input:
+  - - meta:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. `[ id:'sample1' ]`
+    - vcf:
+        type: file
+        description: Input STR call VCF file
+        pattern: "*.{vcf,vcf.gz,bcf,bcf.gz}"
+        ontologies:
+          - edam: "http://edamontology.org/format_3016" # VCF
+          - edam: "http://edamontology.org/format_3020" # BCF
+  - - meta2:
+        type: map
+        description: |
+          Groovy Map containing training set information
+          e.g. `[ id:'sample1' ]`
+    - training_set_json:
+        type: file
+        description: Input training set as json
+        pattern: "*.json"
+        ontologies:
+          - edam: "http://edamontology.org/format_3464" # JSON
+  - - meta3:
+        type: map
+        description: |
+          Groovy Map containing training set information
+          e.g. `[ id:'sample1' ]`
+    - training_set_vcfs:
+        type: file
+        description: Input training set as VCF files
+        pattern: "*.{vcf,vcf.gz,bcf,bcf.gz}"
+        ontologies:
+          - edam: "http://edamontology.org/format_3016" # VCF
+          - edam: "http://edamontology.org/format_3020" # BCF
+output:
+  vcf:
+    - - meta:
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. `[ id:'sample1' ]`
+      - "*.vcf.gz":
+          type: file
+          description: Output annotated VCF file
+          pattern: "*.{vcf.gz}"
+          ontologies:
+            - edam: "http://edamontology.org/format_3016" # VCF
+  versions_strdrop:
+    - - "${task.process}":
+          type: string
+          description: The name of the process
+      - "strdrop":
+          type: string
+          description: The name of the tool
+      - "strdrop --version | sed 's/.* //g'":
+          type: eval
+          description: The expression to obtain the version of the tool
+
+topics:
+  versions:
+    - - ${task.process}:
+          type: string
+          description: The name of the process
+      - strdrop:
+          type: string
+          description: The name of the tool
+      - strdrop --version | sed 's/.* //g':
+          type: eval
+          description: The expression to obtain the version of the tool
+authors:
+  - "@fellen31"
+maintainers:
+  - "@fellen31"
--- a/modules/nf-core/strdrop/call/tests/main.nf.test
+++ b/modules/nf-core/strdrop/call/tests/main.nf.test
@@ -1,0 +1,209 @@
+nextflow_process {
+
+    name "Test Process STRDROP_CALL"
+    script "../main.nf"
+    process "STRDROP_CALL"
+
+    tag "modules"
+    tag "modules_nfcore"
+    tag "strdrop"
+    tag "strdrop/call"
+    tag "samtools/faidx"
+    tag "gunzip"
+    tag "trgt/genotype"
+    tag "strdrop/build"
+
+    setup {
+        run("GUNZIP"){
+            script "../../../gunzip/main.nf"
+            process {
+                """
+                input[0] = [
+                    [ id : 'chr22' ],
+                    file(params.modules_testdata_base_path + 'genomics/homo_sapiens/genome/chr22_chr22_KI270734v1_random/sequence/genome.fa.gz', checkIfExists: true)
+                ]
+            """
+            }
+        }
+        run("SAMTOOLS_FAIDX"){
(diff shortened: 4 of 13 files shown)


modules/nf-core/plastid/psite/main.nf
```diff
@@ -0,0 +1,60 @@
...
+    val min_length
+    val max_length
+    val require_upstream
+    val default_offset
+
+    output:
+    tuple val(meta), path("*_metagene_profiles.txt"), emit: metagene_profiles
+    tuple val(meta), path("*_p_offsets.png")        , emit: p_offsets_png
+    tuple val(meta), path("*_p_offsets.txt")        , emit: p_offsets
+    path "versions.yml"                             , emit: versions
```
pinin4fjords, line 19: Might be sensible to use the new topics syntax instead

pinin4fjords: https://github.com/nf-core/modules/blob/c93d7d10f9e67e868d27073e600114ae0ee1adef/modules/nf-core/ribodetector/main.nf#L17

modules/nf-core/plastid/psite/main.nf
```diff
@@ -0,0 +1,60 @@
...
+    label "process_single"
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/plastid:0.6.1--py39had3e4b6_2':
+        'biocontainers/plastid:0.6.1--py39had3e4b6_2' }"
+
+    input:
+    tuple val(meta), path(bam), path(bam_index)
+    path rois_txt
```
pinin4fjords: Pretty much everything should have metas these days

modules/nf-core/plastid/psite/main.nf
```diff
@@ -0,0 +1,60 @@
...
+
+    script:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def args = task.ext.args ?: ""
+    if (require_upstream) args += "--require_upstream"
+    """
+    psite \
+        "$rois_txt" \\
+        "$prefix" \\
+        --min_length "$min_length" --max_length "$max_length" \\
```
pinin4fjords: The are optional, no? those should normally come via ext.args.

modules/nf-core/plastid/psite/main.nf
```diff
@@ -0,0 +1,60 @@
...
+    tuple val(meta), path("*_p_offsets.txt")        , emit: p_offsets
+    path "versions.yml"                             , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def args = task.ext.args ?: ""
+    if (require_upstream) args += "--require_upstream"
```
pinin4fjords: Seems like another optional thing that should come via ext.args

modules/nf-core/plastid/psite/main.nf
```diff
@@ -0,0 +1,60 @@
...
+    stub:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    touch ${prefix}_metagene_profiles.txt
+    touch ${prefix}_p_offsets.png
+    touch ${prefix}_p_offsets.txt
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        plastid: 0.6.1
```
pinin4fjords: Does the tool not have a version command? We don't like to hard code these

modules/nf-core/plastid/psite/main.nf
```diff
@@ -0,0 +1,60 @@
...
+    sed -i '/^##/d' *.txt # remove variable comment header
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        plastid: 0.6.1
+    END_VERSIONS
+    """
+
+    stub:
+    def prefix = task.ext.prefix ?: "${meta.id}"
```
pinin4fjords, line 42: The stub should reflect all of the handling logic from the main script

modules/nf-core/plastid/metagene_generate/main.nf
```diff
@@ -0,0 +1,44 @@
...
+    tag "$gtf"
+    label "process_low"
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/plastid:0.6.1--py39had3e4b6_2':
+        'biocontainers/plastid:0.6.1--py39had3e4b6_2' }"
+
+    input:
+    path gtf
```
pinin4fjords: You definitely need a meta, and you probably want to bundle these two inputs together (unless the val is at all optional, in which it should come via ext.args)

modules/nf-core/plastid/metagene_generate/main.nf
```diff
@@ -0,0 +1,44 @@
...
+    val landmark
+
+    output:
+    path "*_rois.txt"  , emit: rois_txt
```
pinin4fjords: Outputs will need the metas too

modules/nf-core/plastid/metagene_generate/main.nf
```diff
@@ -0,0 +1,44 @@
...
+    path "*_rois.bed"  , emit: rois_bed
+    path "versions.yml", emit: versions
```
pinin4fjords, line 17: Again, can probably do with topics

modules/nf-core/plastid/metagene_generate/main.nf
```diff
@@ -0,0 +1,44 @@
...
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    """
+    metagene generate "${gtf.baseName}" --landmark "$landmark" --annotation_files "$gtf" $args
```
pinin4fjords: Are any of these optional? Would `metagene generate` run without them? e.g. 'landmark'? If so, that should go via ext.args

modules/nf-core/plastid/metagene_generate/main.nf
```diff
@@ -0,0 +1,44 @@
...
+    sed -i '/^##/d' *_rois.* # remove variable comment header
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        plastid: 0.6.1
```
pinin4fjords: again with the hard code

pinin4fjords: if it needs to be a static version it should go like: https://nf-co.re/docs/guidelines/components/modules#emission-of-versions

modules/nf-core/plastid/make_wiggle/main.nf
```diff
@@ -0,0 +1,64 @@
...
+    label "process_single"
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/plastid:0.6.1--py39had3e4b6_2':
+        'biocontainers/plastid:0.6.1--py39had3e4b6_2' }"
+
+    input:
+    tuple val(meta), path(bam), path(bam_index), path(p_offsets)
+    val min_length
```
pinin4fjords: again, just checking, but if a `val` argument isn't absolutely required for the command operation it should come via ext.args.

modules/nf-core/plastid/make_wiggle/main.nf
```diff
@@ -0,0 +1,64 @@
...
+    path "versions.yml", emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def extension = output_format == "bedgraph" ? "bedgraph" : "wig"
+    def args = task.ext.args ?: ""
+    if (fiveprime_variable) args += "--fiveprime_variable"
```
pinin4fjords: I suspect we won't need this with good ext.args handling

modules/nf-core/plastid/make_wiggle/main.nf
```diff
@@ -0,0 +1,64 @@
...
+    make_wiggle \\
+        --count_files "$bam" \\
+        --offset "$p_offsets" \\
+        --min_length "$min_length" \\
+        --output_format "$output_format" \\
+        --verbose \\
+        -o "$prefix" \\
+        $args
+
+    sed -i '/^##/d' *.wig # remove variable comment header
```
pinin4fjords: Let's not edit command output. If you're doing this for the snapshot purposes there are other workarounds (like checking for content rather than a full snapshot)

modules/nf-core/plastid/make_wiggle/main.nf
```diff
@@ -0,0 +1,64 @@
...
+
+    if [ "$output_format" = "bedgraph" ]; then
+        for FILE in *.wig; do
+            mv "\$FILE" "\${FILE%.wig}.bedgraph"
```
pinin4fjords, line 41: Why the rename?

modules/nf-core/plastid/make_wiggle/main.nf
```diff
@@ -0,0 +1,64 @@
...
+    stub:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def extension = output_format == "bedgraph" ? "bedgraph" : "wig"
+    """
+    touch ${prefix}_fw.$extension
+    touch ${prefix}_rc.$extension
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        plastid: 0.6.1
```
pinin4fjords: ditton on the version comment for other modules

modules/nf-core/plastid/make_wiggle/main.nf
suhrig: The tool names the files .wig, even when bedgraph is chosen as output format. I don't mind leaving it like this, but it feels wrong and I thought it could lead to confusion with downstream tools that determine the file type from the extension. What do you recommend?

suhrig:
Thanks for the comprehensive feedback! I will address the points and change the code accordingly.

modules/nf-core/plastid/psite/main.nf
suhrig: I searched. It doesn't. I will use the `VERSION` variable workaround with the warning comment.

modules/nf-core/plastid/make_wiggle/main.nf
pinin4fjords: I see, OK, guess that's fair.

modules/nf-core/plastid/psite/main.nf
suhrig: I have added the meta to all input/output files.

modules/nf-core/plastid/psite/main.nf
suhrig: All optional parameters have been removed.

modules/nf-core/plastid/make_wiggle/main.nf
suhrig: This is the only parameter that is mandatory. `make_wiggle` accepts p-site offsets to be supplied in two ways: as a static integer value or as a file listing a different value for each read length. Unfortunately, in both cases the argument is supplied via the parameter `--offset`. Nextflow cannot deal with the different data types. It has to be either a val() or a path(). I decided to go for path(), because I believe this is more commonly used and also more flexible. Users who wish to use a static value for all read lengths can still supply a file with just one entry `default <some integer>`. When a path() is used, then the parameter `--fiveprime_variable` is mandatory.

modules/nf-core/plastid/make_wiggle/main.nf
suhrig: I'm using content-based checks now.

modules/nf-core/plastid/psite/main.nf
suhrig: As discussed on Slack I will stick with the traditional method of collecting versions. I hope that's okay.

suhrig:
Thanks again for your feedback. All issues should be resolved now. Let me know if you would like any further changes.

modules/nf-core/plastid/make_wiggle/main.nf
```diff
@@ -0,0 +1,61 @@
...
+    script:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def extension = output_format == "bedgraph" ? "bedgraph" : "wig"
+    def args = task.ext.args ?: ""
+    def VERSION = "0.6.1" // WARN: Version information not provided by tool on CLI. Please update this string when bumping container versions.
+    """
+    make_wiggle \\
+        --count_files "$bam" \\
+        --offset "$p_offsets" \\
+        --fiveprime_variable \\
```
pinin4fjords: You're still hard-coding things that look optional (correct me if I'm wrong). These should go via ext.args, and you can do that via nextflow.config files in the test directory and the appropriate `withName` selectors. There's lots of examples of this in the modules repo if you look about.

suhrig: `make_wiggle` doesn't run when `--fiveprime_variable` is absent and the argument to `--offset` is  a file. I can move `--fiveprime_variable` to `ext.args` if you like, but this will mean that anyone using the module will have to do this.

pinin4fjords: I've opened a PR to your branch with some nf-core compliance tweaks by way of illustration: https://github.com/suhrig/modules/pull/1

Since 'a' mapping rule is mandatory (one of `--fiveprime`, `--threeprime`, `--center`, `--fiveprime_variable`), but `--fiveprime_variable` is not, I've added it as a mapping_rule val input with an enum. 

The p_offsets file is then only passed when mapping_rule == 'fiveprime_variable'. Moved output_format to ext.args since it's optional, but we can still respond to it for the extension.

pinin4fjords:
Hi @suhrig, I've opened another PR to your branch with suggestions to fix a CI reproducibility issue: https://github.com/suhrig/modules/pull/2

The psite PNG has non-reproducible md5sums due to matplotlib rendering differences across conda builds, so I've excluded it from the snapshot (just checking existence instead).

Also tidied up the test assertions while I was at it - consolidated snapshots and removed snapshots of empty stub files.

pinin4fjords:
@suhrig sorry, maybe I added some things to the snapshotting that shouldn't be there, let me fix it up

PR #9490, opened 2025-12-02
5787 new module whatshap/stats

### LINKED ISSUE #5787: new module: WHATSHAP/STATS

eliottBo:
Add module whatshap/stats.

## PR checklist

Closes #5787 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/subworkflows/nf-core/bam_impute_stitch/main.nf
+++ b/subworkflows/nf-core/bam_impute_stitch/main.nf
@@ -32,6 +32,9 @@ workflow BAM_IMPUTE_STITCH {
         .combine(ch_parameters)
         .map{
             metaI, bam, bai, bampath, bamname, metaPC, posfile, gmap, chr, start, end ->
+            if (!chr) {
+                error "ERROR: chromosome is not provided in ch_chunks."
+            }
             def regionout = "${chr}"
             if (start != [] && end != []) {
                 regionout = "${chr}:${start}-${end}"
--- a/subworkflows/nf-core/bam_impute_stitch/tests/main.nf.test
+++ b/subworkflows/nf-core/bam_impute_stitch/tests/main.nf.test
@@ -16,7 +16,7 @@ nextflow_workflow {
     tag "glimpse2"
     tag "glimpse2/ligate"
 
-    test("Impute with stitch one individual, no renaming, posfile, no chunks, no map, no fasta") {
+    test("Impute with stitch one individual, no renaming, posfile, no chunks whole chromosome, no map, no fasta") {
         when {
             workflow {
                 """
@@ -183,8 +183,8 @@ nextflow_workflow {
                 bamname = Channel.of("MySample1").collectFile(name: 'bamname.txt', newLine: true)
                 ch_samples = channel.of([
                     [id: "allid"],
-                    file(params.pipelines_testdata_base_path + "hum_data/individuals/NA12878/NA12878.s.bam", checkIfExist:true),
-                    file(params.pipelines_testdata_base_path + "hum_data/individuals/NA12878/NA12878.s.bam.bai", checkIfExist:true)
+                    file(params.modules_testdata_base_path + "genomics/homo_sapiens/illumina/bam/NA12878.chr21_22.1X.bam", checkIfExist:true),
+                    file(params.modules_testdata_base_path + "genomics/homo_sapiens/illumina/bam/NA12878.chr21_22.1X.bam.bai", checkIfExist:true)
                 ])
                 input[0] = ch_samples.combine(bampath).combine(bamname) // input
                 input[1] = channel.of(
@@ -212,6 +212,37 @@ nextflow_workflow {
             assertAll(
                 { assert workflow.failed },
                 { assert workflow.errorMessage.contains("ERROR: join operation resulted in an empty channel. Please provide a valid ch_chunks and ch_map channel as input.") }
+            )
+        }
+    }
+
+    test("homo_sapiens - error no chromosome provided - stub") {
+        options "-stub"
+        when {
+            workflow {
+                """
+                bampath = Channel.of("NA12878.chr21_22.1X.bam").collectFile(name: 'bampath.txt', newLine: true)
+                bamname = Channel.of("MySample1").collectFile(name: 'bamname.txt', newLine: true)
+                ch_samples = channel.of([
+                    [id: "allid"],
+                    file(params.modules_testdata_base_path + "genomics/homo_sapiens/illumina/bam/NA12878.chr21_22.1X.bam", checkIfExist:true),
+                    file(params.modules_testdata_base_path + "genomics/homo_sapiens/illumina/bam/NA12878.chr21_22.1X.bam.bai", checkIfExist:true)
+                ])
+                input[0] = ch_samples.combine(bampath).combine(bamname) // input
+                input[1] = channel.of([[id: "1000GP", chr: "chr22"], []]) // posfile
+                input[2] = channel.of([[chr: "chr22", id: "1000GP"], [], [], []]) // Empty chunks
+                input[3] = Channel.of([[id: "1000GP", chr: "chr22"], []],) // map
+                input[4] = channel.of([ [id: "GRCh38"], [], [] ]).collect()
+                input[5] = 2
+                input[6] = 100
+                input[7] = 1
+                """
+            }
+        }
+        then {
+            assertAll(
+                { assert workflow.failed },
+                { assert workflow.errorMessage.contains("ERROR: chromosome is not provided in ch_chunks.") }
             )
         }
     }
--- a/subworkflows/nf-core/bam_impute_stitch/tests/main.nf.test.snap
+++ b/subworkflows/nf-core/bam_impute_stitch/tests/main.nf.test.snap
@@ -1,5 +1,5 @@
 {
-    "Impute with stitch one individual, no renaming, posfile, no chunks, no map, no fasta": {
+    "Impute with stitch one individual, no renaming, posfile, no chunks whole chromosome, no map, no fasta": {
         "content": [
             [
                 [


modules/nf-core/whatshap/stats/main.nf
```diff
@@ -0,0 +1,42 @@
...
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/whatshap:2.8--py39h2de1943_0':
+        'biocontainers/whatshap:2.8--py39h2de1943_0' }"
+
+    input:
+    tuple val(meta), path(vcf)
+
+    output:
+    tuple val(meta), path("*.tsv"),                                            emit: tsv
+    tuple val(meta), path("*.txt"),                                            emit: txt
```
nschan: I would recommend in both cases:
`path("${prefix}.tsv")` and `path("${prefix}.txt")`

nschan:
This looks overall fine. My concern is that whatshap stats has a bunch of arguments that are not clearly documented on readthedocs (https://github.com/whatshap/whatshap/blob/main/whatshap/cli/stats.py):

```
    add("--gtf", metavar="FILE", help="Write phased blocks as GTF with each block represented as a "
        "'gene'. If blocks are interleaved or nested, they are split into multiple 'exons'.")
    add("--block-list", metavar="FILE", help="Write list of all blocks to FILE (one block per "
        "line). Nested/interleaved blocks are not split.")
    add("--sample", metavar="SAMPLE", help="Name of the sample "
        "to process. If not given, use first sample found in VCF.")
    add("--chr-lengths", metavar="FILE",
        help="Override chromosome lengths in VCF with those from FILE (one line per chromosome, "
        "tab separated '<chr> <length>'). Lengths are used to compute NG50 values.")
    add("--tsv", metavar="FILE", help="Write statistics in tab-separated value format to FILE")
    add("--only-snvs", default=False, action="store_true", help="Only process SNVs "
        "and ignore all other variants.")
    add("--chromosome", dest="chromosomes", metavar="CHROMOSOME", default=[], action="append",
        help="Name of chromosome(s) to process. If not given, all chromosomes in the "
        "input VCF are considered. Can be used multiple times and accepts a comma-separated list. ")
    add("vcf", metavar="VCF", help="Phased VCF file")
```

And it looks like the current implementation uses `--tsv`, but I think it would be nice to support maybe an input switch for format, to toggle between `gtf`, `block-list` and `tsv`? I think the other arguments can come via `$args` but additional output formats would not be captured as it is now.

eliottBo:
> This looks overall fine. My concern is that whatshap stats has a bunch of arguments that are not clearly documented on readthedocs (https://github.com/whatshap/whatshap/blob/main/whatshap/cli/stats.py):
> 
> ```
>     add("--gtf", metavar="FILE", help="Write phased blocks as GTF with each block represented as a "
>         "'gene'. If blocks are interleaved or nested, they are split into multiple 'exons'.")
>     add("--block-list", metavar="FILE", help="Write list of all blocks to FILE (one block per "
>         "line). Nested/interleaved blocks are not split.")
>     add("--sample", metavar="SAMPLE", help="Name of the sample "
>         "to process. If not given, use first sample found in VCF.")
>     add("--chr-lengths", metavar="FILE",
>         help="Override chromosome lengths in VCF with those from FILE (one line per chromosome, "
>         "tab separated '<chr> <length>'). Lengths are used to compute NG50 values.")
>     add("--tsv", metavar="FILE", help="Write statistics in tab-separated value format to FILE")
>     add("--only-snvs", default=False, action="store_true", help="Only process SNVs "
>         "and ignore all other variants.")
>     add("--chromosome", dest="chromosomes", metavar="CHROMOSOME", default=[], action="append",
>         help="Name of chromosome(s) to process. If not given, all chromosomes in the "
>         "input VCF are considered. Can be used multiple times and accepts a comma-separated list. ")
>     add("vcf", metavar="VCF", help="Phased VCF file")
> ```
> 
> And it looks like the current implementation uses `--tsv`, but I think it would be nice to support maybe an input switch for format, to toggle between `gtf`, `block-list` and `tsv`? I think the other arguments can come via `$args` but additional output formats would not be captured as it is now.

Thanks for your review @nschan , I now updates my PR for the user to be able to choose which output they want to make it more flexible.

nschan:
Thanks, that looks very good to me. I have one additional question, probably because I have no experience with it: independent of any `--gtf` or `--tsv` setting, will it always print to stdout, i.e. can you capture the stats with `> ${prefix}.txt` even if a different output format is specified?

eliottBo:
> Thanks, that looks very good to me. I have one additional question, probably because I have no experience with it: independent of any `--gtf` or `--tsv` setting, will it always print to stdout, i.e. can you capture the stats with `> ${prefix}.txt` even if a different output format is specified?

I thought about always capturing the output in the `.txt` file but from a user point of view I thought to get only the output you want would be better. As it is now the user can definately get the stdout which will end up in the `.txt` file. The combination of all output is possible. If, for example, the user is just interested in the `.tsv `or .`gtf` it is maybe not necessary to **always** create this `.txt` file with the stdout. But if you think it is better to always have it, I can add it by default.

nschan:
No, that is not what I meant, I only wanted to clarify if whatshap stats will always output the same format to stdout independent of the format specified.
However, this made me remember something: nf-core guidelines recommend using `tee` when capturing stdout (https://nf-co.re/docs/guidelines/components/modules#capturing-stdout-and-stderr). I think this is more of a "nice to do right" thing, but maybe it would be useful?

modules/nf-core/whatshap/stats/main.nf
```diff
@@ -0,0 +1,59 @@
...
+    val(include_tsv_output)    // value:   [ true | false ]
+    val(include_gtf_output)    // value:   [ true | false ]
+    val(inlude_block_output)   // value:   [ true | false ]
+    val(include_txt_output)    // value:   [ true | false ]
+
+    output:
+    tuple val(meta), path("${prefix}_whap_stats.tsv"),                         emit: tsv,   optional: true
+    tuple val(meta), path("${prefix}_whap_stats.gtf"),                         emit: gtf,   optional: true
+    tuple val(meta), path("${prefix}_whap_stats_block.txt"),                   emit: block, optional: true
+    tuple val(meta), path("${prefix}_whap_stats.txt"),                         emit: txt,   optional: true
```
fellen31: I suggest giving the user of the module the opportunity to choose to include e.g. `_whap_stats` via the prefix, if they want to, rather than hardcoding it. Although we would need some distinction between the block (which is also tab-separated) and txt output (calling the txt log for example, which I'd argue that it is, especially when running with the `--debug` flag), if it's kept. 

```suggestion
    tuple val(meta), path("${prefix}.tsv"),                         emit: tsv,   optional: true
    tuple val(meta), path("${prefix}.gtf"),                         emit: gtf,   optional: true
    tuple val(meta), path("${prefix}.txt"),                   emit: block, optional: true
    tuple val(meta), path("${prefix}.log"),                         emit: log,   optional: true
```

modules/nf-core/whatshap/stats/main.nf
```diff
@@ -0,0 +1,59 @@
...
+    def output_block = inlude_block_output ? "--block-list ${prefix}_whap_stats_block.txt" : ''
+    def output_txt   = include_txt_output  ? "> ${prefix}_whap_stats.txt"                  : ''
+    """
+    whatshap stats \\
+        $args \\
+        $output_tsv \\
+        $output_gtf \\
+        $output_block \\
+        $vcf \\
+        $output_txt \\
```
fellen31:
```suggestion
        $output_txt
```

modules/nf-core/whatshap/stats/main.nf
```diff
@@ -0,0 +1,59 @@
...
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args   ?: ''
+    prefix   = task.ext.prefix ?: "${meta.id}"
+
+    def output_tsv   = include_tsv_output  ? "--tsv ${prefix}_whap_stats.tsv"              : ''
+    def output_gtf   = include_gtf_output  ? "--gtf ${prefix}_whap_stats.gtf"              : ''
+    def output_block = inlude_block_output ? "--block-list ${prefix}_whap_stats_block.txt" : ''
+    def output_txt   = include_txt_output  ? "> ${prefix}_whap_stats.txt"                  : ''
```
fellen31: I think having it like this is OK, but I would also be down to scrap the stdout output altogether. I think it's more of a log rather than an output really. No one in their right mind would choose to process that over the TSV 😅 (used by e.g. MultiQC). The printed output would still available in the stdout files.

If kept, since the tool will aways write this to stdout, regardless of whether any of the other output formats are selected, I think it would be okay to always output this log as default as well. The user doesn't have to use it if they don't want to.

modules/nf-core/whatshap/stats/main.nf
```diff
@@ -0,0 +1,59 @@
...
+    prefix           = task.ext.prefix     ?: "${meta.id}"
+    def output_tsv   = include_tsv_output  ? "--tsv ${prefix}_whap_stats.tsv" : ''
+    def output_gtf   = include_gtf_output  ? "--gtf ${prefix}_whap_stats.gtf" : ''
+    def output_block = inlude_block_output ? "--block-list ${prefix}_whap_stats_block.txt" : ''
+    def output_txt   = include_txt_output  ? "> ${prefix}_whap_stats.txt" : ''
+    """
+    touch ${prefix}_whap_stats.tsv
+    touch ${prefix}_whap_stats.gtf
+    touch ${prefix}_whap_stats_block.txt
+    touch ${prefix}_whap_stats.txt
```
fellen31: Something like this would make the stub section mimic the script section correctly.

```suggestion
    def tsv_touch_cmd   = include_tsv_output  ? "--tsv ${prefix}_whap_stats.tsv" : ''
    def gtf_touch_cmd   = include_gtf_output  ? "touch ${prefix}.tsv" : ''
    def block_touch_cmd = inlude_block_output ? "touch ${prefix}.txt" : ''
    def log_touch_cmd   = include_txt_output  ? "touch ${prefix}.log" : ''
    """
    echo $args
    
    $tsv_touch_cmd
    $gtf_touch_cmd
    $block_touch_cmd
    $txt_touch_cmd
```

modules/nf-core/whatshap/stats/main.nf
eliottBo: Good suggestion, more flexible

modules/nf-core/whatshap/stats/main.nf
eliottBo: I decided to go for the second option and always have the .log file in order to have at least one output by default. I used the tee command to have it in the file and in the stdout too.

modules/nf-core/whatshap/stats/main.nf
```diff
@@ -18,6 +18,7 @@ process WHATSHAP_STATS {
     tuple val(meta), path("${prefix}.tsv"),                                    emit: tsv,   optional: true
     tuple val(meta), path("${prefix}.gtf"),                                    emit: gtf,   optional: true
     tuple val(meta), path("${prefix}.txt"),                                    emit: block, optional: true
+    tuple val(meta), path("${prefix}.log"),                                    emit: log,   optional: true
```
fellen31: If always output, it should not be optional :)

```suggestion
    tuple val(meta), path("${prefix}.log"),                                    emit: log,   optional: true
```

fellen31 (approved):
LGTM 🚀

PR #9503, opened 2025-12-04
meta.yml schema: add `containers` section, fix order and simplify

mashehu:
probably easier to review each commit, because git is bad with order changes


modules/meta-schema.json
```diff
@@ -225,9 +99,133 @@
...
+            "description": "Container images for the module",
+            "properties": {
+                "docker": {
+                    "type": "object",
+                    "description": "Docker containers for different architectures",
+                    "additionalProperties": {
+                        "$ref": "#/definitions/container"
+                    }
+                },
+                "singularity": {
```
ewels: Should we make `https` required?

modules/meta-schema.json
```diff
@@ -194,17 +76,9 @@
...
-                                    {
-                                        "type": "string",
-                                        "pattern": "^biotools:.*$"
-                                    },
-                                    {
-                                        "type": "string",
-                                        "maxLength": 0
-                                    }
-                                ]
+                                "pattern": "^(biotools:.*)?$"
```
mirpedrol, line 81: Will this allow for empty strings? It would be good to doublecheck this

modules/meta-schema.json
```diff
@@ -261,13 +259,49 @@
...
+        },
+        "container": {
+            "type": "object",
+            "description": "Container information for Docker or Singularity",
+            "properties": {
+                "name": {
+                    "type": "string",
+                    "description": "Container name/URI"
+                },
+                "https": {
```
mirpedrol, line 279: https is required for singularity containers, would it be worth it to have two separate objects for docker and singularity to be able to make this required?

modules/meta-schema.json
```diff
@@ -225,9 +99,133 @@
...
+            }
+        },
+        "containers": {
+            "type": "object",
+            "description": "Container images for the module",
+            "properties": {
+                "docker": {
+                    "type": "object",
+                    "description": "Docker containers for different architectures",
+                    "additionalProperties": {
```
mirpedrol, line 172: Aren't we missing one level? for `linux_amd64` and `linux_arm64`

modules/meta-schema.json
mashehu: yes, because of the `?`

modules/meta-schema.json
mashehu: will add it to the `allOf` above

modules/meta-schema.json
mashehu: additionalProperties handles `linux_*` I got also confused for a bit

PR #9520, opened 2025-12-08
fix(tximport): handle tx2gene files with extra columns

pinin4fjords:
## Summary

Fixes nf-core/rnaseq#1652

- Fix data corruption when using multiple `--gtf_extra_attributes` (e.g., `gene_name,gene_biotype`)
- Read tx2gene file with proper header handling instead of hardcoded column names
- Select columns by name with positional fallback for compatibility with non-standard GTFs
- Add configurable column names via `ext.args` (`--tx_col`, `--gene_id_col`, `--gene_name_col`)

## Problem

When users specify multiple comma-separated values for `--gtf_extra_attributes`, the `custom/tx2gene` module produces a file with more than 3 columns. The previous `read.csv()` call used hardcoded `col.names = c("tx", "gene_id", "gene_name")`, which caused R to drop the first column and shift data left, corrupting the transcript-to-gene mapping and producing incorrect gene-level counts.

## Solution

- Read the tx2gene file with `header = TRUE` without hardcoded column names
- Select columns by name (`transcript_id`, `gene_id`, `gene_name`) with positional fallback (columns 1, 2, 3) if names don't match
- This handles both standard and non-standard GTF attribute names

## Test plan

- [x] Added test case `saccharomyces_cerevisiae - kallisto - gtf - extra_attributes` with `gene_name,gene_biotype` to verify 4-column tx2gene handling
- [x] Verified output md5sums match the standard 3-column test (confirming correct gene aggregation)
- [x] Existing tests pass

### DIFF
--- a/modules/nf-core/bbmap/bbsplit/main.nf
+++ b/modules/nf-core/bbmap/bbsplit/main.nf
@@ -79,7 +79,7 @@ process BBMAP_BBSPLIT {
         done
         find index_writable/ref/genome -name summary.txt | while read -r summary_file; do
             src=\$(grep '^source' "\$summary_file" | cut -f2- -d\$'\\t' | sed 's|.*/ref/|index_writable/ref/|')
-            mod=\$(echo "System.out.println(java.nio.file.Files.getLastModifiedTime(java.nio.file.Paths.get(\\"\$src\\")).toMillis());" | jshell -J-Djdk.lang.Process.launchMechanism=vfork -)
+            mod=\$(echo "System.out.println(java.nio.file.Files.getLastModifiedTime(java.nio.file.Paths.get(\\"\$src\\")).toMillis());" | jshell -J-Djdk.lang.Process.launchMechanism=vfork - 2>/dev/null | grep -oE '^[0-9]{12,14}\$')
             sed -e 's|bbsplit_index/ref|index_writable/ref|' -e "s|^last modified.*|last modified\\t\$mod|" "\$summary_file" > \${summary_file}.tmp && mv \${summary_file}.tmp \${summary_file}
         done
     fi
--- a/modules/nf-core/bedgovcf/environment.yml
+++ b/modules/nf-core/bedgovcf/environment.yml
@@ -5,3 +5,4 @@ channels:
   - bioconda
 dependencies:
   - bioconda::bedgovcf=0.1.1
+  - bioconda::htslib=1.20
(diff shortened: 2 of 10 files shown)


modules/nf-core/tximeta/tximport/tests/main.nf.test.snap
```diff
@@ -5,63 +5,63 @@
                 "0": [
                     [
                         [
-
+
```
LouisLeNezet, line 8: This is strange that empty spaces are added here.

modules/nf-core/tximeta/tximport/tests/main.nf.test
```diff
@@ -71,6 +71,65 @@ nextflow_process {
...
+                    process.out.tpm_transcript,
+                    process.out.versions).match()
+                }
+            )
+        }
+    }
+
     test("saccharomyces_cerevisiae - kallisto - gtf - stub") {
 
         options "-stub"
```
LouisLeNezet, line 135: Could you remove the UNTAR and CUSTOM TXGENE in the stub test as they aren't used ?
Also adding a id to the first meta map channel would avoid the strange "[].gene*" names in the snapshot.

modules/nf-core/tximeta/tximport/templates/tximport.r
```diff
@@ -117,6 +145,21 @@ create_summarized_experiment <- function(counts, abundance, length, col_data, ro
...
 
+# Set defaults and classes
+opt <- list(
+    tx_col = "transcript_id",
+    gene_id_col = "gene_id",
+    gene_name_col = "gene_name"
+)
+
+# Apply parameter overrides from ext.args
+args_opt <- parse_args('$task.ext.args')
```
LouisLeNezet, line 156: Could you add a test where you use these optional arguments with other values, to ensure that they works ?

modules/nf-core/tximeta/tximport/tests/main.nf.test
pinin4fjords: Oh yeah, fair

modules/nf-core/tximeta/tximport/tests/main.nf.test.snap
pinin4fjords: I think nf-test must have added them...

LouisLeNezet (approved):
Fine for me !

PR #9333, opened 2025-11-03
Merge modules with new topics syntax

mashehu:
See https://deploy-preview-3613--nf-core-main-site.netlify.app/blog/2025/version_topics for details about this effort

PR #8895, opened 2025-08-13
add writing index option to samtools sort

ljwharbers:
This adds the ability to include an index of choice using `samtools sort`. Currently it was only possible to use the `ext.args = "--write-index"` which will only output a `.csi` index. This adds a new `val` channel input that will output the selected bam/sam/cram index. If the channel is left empty `''` no index will be written and the old behaviour is maintained.

This is the same as the use case in the latest version of `samtools/view` and `minimap2/align` and will reduce a lot of unecessary `SAMTOOLS_SORT` into `SAMTOOLS_INDEX` processes.

Due to many modules/subworkflows use `samtools sort` (for instance in `BAM_SORT_STATS_SAMTOOLS`), other tests are also updated to include an empty third input channel to maintain the old behaviour. 

Optionally `BAM_SORT_STATS_SAMTOOLS` can also be updated to now entirely skip `SAMTOOLS_INDEX` and simply directly write the `bai` index within `SAMTOOLS_SORT`

## PR checklist

Closes #XXX 

- [x] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [ ] Remove all TODO statements.
- [ ] Emit the `versions.yml` file.
- [ ] Follow the naming conventions.
- [ ] Follow the parameters requirements.
- [ ] Follow the input/output options guidelines.
- [ ] Add a resource `label`
- [ ] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - For modules:
    - [ ] `nf-core modules test <MODULE> --profile docker`
    - [x] `nf-core modules test <MODULE> --profile singularity`
    - [ ] `nf-core modules test <MODULE> --profile conda`
  - For subworkflows:
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile docker`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile singularity`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile conda`

PR #9521, opened 2025-12-08
Add draft of FASTQ_REMOVE_ADAPTERS_AND_MERGE subworkflow with tests

kornkv:
draft of FASTQ_REMOVE_ADAPTERS_AND_MERGE subworkflow with tests

### DIFF
--- a/subworkflows/nf-core/fastq_removeadapters_merge/tests/main.nf.test.snap
+++ b/subworkflows/nf-core/fastq_removeadapters_merge/tests/main.nf.test.snap
@@ -216,7 +216,7 @@
     },
     "sarscov2 - fastq - adapterremoval - paired-end - merge": {
         "content": [
-            464,
+            336,
             [
                 "test.discarded.fastq.gz"
             ],
@@ -235,7 +235,7 @@
             "nf-test": "0.9.3",
             "nextflow": "25.10.2"
         },
-        "timestamp": "2026-01-12T14:40:54.554084134"
+        "timestamp": "2026-01-14T14:17:56.133855749"
     },
     "sarscov2 - fastq - fastp - single-end": {
         "content": [


subworkflows/nf-core/fastq_removeadapters_merge/meta.yml
```diff
@@ -0,0 +1,144 @@
...
+      description: |
+        The trimmed/modified fastq reads merged into a single file.
+        Only meaningful for paired-end reads.
+      pattern: "*.fastq.gz"
+  - discarded_reads:
+      type: file
+      description: |
+        All files containing discarded reads throughout the steps.
+      pattern: "*.fastq.gz"
+  - trimmomatic_trim_log:
```
miraep8: For some of these outputs where the underlying tool is optional, should these outputs be marked as optional?

subworkflows/nf-core/fastq_removeadapters_merge/tests/main.nf.test
```diff
@@ -0,0 +1,205 @@
+nextflow_workflow {
+
```
miraep8, line 2: I think it might be nice to add a test skipping bbduk, trimgalore or trimmomatic (or all of them).  Right now, I believe the only step that is skipped/tested is `cutadapt` (in `sarscov2 - fastq - single-end - stub`) which isn't connected to any unique outputs that may disappear when it is skipped.

vagkaratzas: Thanks! I added another nf-test with-skips!

subworkflows/nf-core/fastq_removeadapters_merge/meta.yml
vagkaratzas: More or less everything is kinda optional here since we have the skip flags anyway, so I mostly copy-pasted from each tool's module yaml file :D

pinin4fjords:
## Note for reviewers: `setup-nextflow` deletion is intentional cleanup

The diff shows deletion of a `setup-nextflow` submodule entry - this is **correct and beneficial**, not an error.

### Background

- `setup-nextflow` is the [nf-core/setup-nextflow](https://github.com/nf-core/setup-nextflow) GitHub Action for installing Nextflow in CI
- It was **accidentally committed** as an orphaned git submodule in PR #8895 (2025-08-18) - likely from a contributor's local dev environment
- There's no corresponding `.gitmodules` file, making it a broken/orphaned gitlink
- Previous removal attempts in PRs #8977 and #9095 were undone when PR #9333 merge reintroduced it

The removal commit by @vagkaratzas (`9239334ea - "remove unused folder"`) correctly cleans this up.

subworkflows/nf-core/fastq_removeadapters_merge/meta.yml
```diff
@@ -0,0 +1,144 @@
...
+      type: file
+      description: trimmomatic log file, from the trim_log parameter
+      pattern: "*.log"
+  - trimmomatic_summary:
```
pinin4fjords: At least some of these outputs could have structure annotations:

```
  - trim_json:
      type: file
      description: |
        Structure: [ val(meta), path(trim_json) ]
        Results in JSON format
```

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
```diff
@@ -0,0 +1,148 @@
...
+// both SE and PE, plus merging
+include { FASTP                               } from '../../../modules/nf-core/fastp/main'
+include { ADAPTERREMOVAL as ADAPTERREMOVAL_SE } from '../../../modules/nf-core/adapterremoval/main'
+include { ADAPTERREMOVAL as ADAPTERREMOVAL_PE } from '../../../modules/nf-core/adapterremoval/main'
+
+workflow FASTQ_REMOVEADAPTERS_MERGE {
+
+    take:
+    reads                       // channel: [ val(meta), [ reads ] ]
+    skip_trimmomatic            // boolean
```
pinin4fjords: You didn't leave any notes in the PR description, so I'm not sure. Is it your intention that folks would want to run multiple trimming methods, which is why you've avoided a simpler `trimmer` param or similar, as `fastq_qc_trim_filter_setstrandedness` does (for example)?

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
```diff
@@ -0,0 +1,148 @@
...
+    skip_cutadapt               // boolean
+    skip_trimgalore             // boolean
+    skip_bbduk                  // boolean
+    contaminants                // channel: [ reads ] // fasta, adapters to remove
```
pinin4fjords: Would this benefit from metas? I think we normally require those these days (not all of the modules/ subworkflows have caught up)

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
```diff
@@ -0,0 +1,148 @@
...
+    skip_leehom                 // boolean
+    skip_fastp                  // boolean
+    fastp_discard_trimmed_pass  // boolean
+    fastp_save_trimmed_fail     // boolean
+    save_merged                 // boolean
+    skip_adapterremoval         // boolean
+    text_adapters               // channel: [ txt ] // adapters to remove, in adapterremoval text format
```
pinin4fjords: Same, think this could do with a meta

subworkflows/nf-core/fastq_removeadapters_merge/meta.yml
vagkaratzas: Done

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
vagkaratzas: Exactly, the idea is to select 1 or more adapter removal (and/or merge) tools. This subworkflow is going to be part of the mega fastq QC/preprocessing subworkflow described [here](https://docs.google.com/document/d/1Jx_L_-evFLfGn6HkouUSmCrWjZxs8SIhJFNCKqTwY1E/edit?tab=t.0#heading=h.tsfqmacx3ysf).
Potentially, with pipeline chaining, this could one day become its own standalone pipeline that would be executed before all fastq short read pipelines.

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
vagkaratzas: At this point in time, I just sourced the inputs in the format that they are being used by the included modules. Will update accordingly when and if the module inputs are updated at module side ;)

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
vagkaratzas: Same as above

pinin4fjords (approved):
I wonder a bit how it fits into the landscape of other trimming/ preprocessing workflows, but nothing particularly objectionable

vagkaratzas:
> I wonder a bit how it fits into the landscape of other trimming/ preprocessing workflows, but nothing particularly objectionable

Thanks! I'll wait for @jfy133 for a final review pass and will then merge.
Stay tuned for the following updates :D

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
```diff
@@ -0,0 +1,148 @@
...
+        ch_trimmomatic_trim_log       = TRIMMOMATIC.out.trim_log
+        ch_trimmomatic_summary        = TRIMMOMATIC.out.summary
+        ch_versions                   = ch_versions.mix(TRIMMOMATIC.out.versions.first())
+        ch_multiqc_files              = ch_multiqc_files.mix(TRIMMOMATIC.out.out_log)
+    }
+
+    if (!skip_cutadapt) {
+        CUTADAPT( ch_reads )
+        ch_reads         = CUTADAPT.out.reads
+        ch_multiqc_files = ch_multiqc_files.mix(CUTADAPT.out.log)
```
jfy133: missing cut adapt versions mixing

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    if (!skip_trimgalore) {
+        TRIMGALORE( ch_reads )
+        ch_reads               = TRIMGALORE.out.reads
+        ch_discarded_reads     = ch_discarded_reads.mix(TRIMGALORE.out.unpaired)
+        ch_trimgalore_log      = TRIMGALORE.out.log
+        ch_trimgalore_html     = TRIMGALORE.out.html
```
jfy133: missing timmomatic versions mixing

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
```diff
@@ -0,0 +1,148 @@
...
+        ch_reads        = ADAPTERREMOVAL_SE.out.singles_truncated.mix(ADAPTERREMOVAL_PE.out.paired_truncated)
+        ch_merged_reads = ADAPTERREMOVAL_SE.out.collapsed
+            .mix(
+                ADAPTERREMOVAL_PE.out.collapsed,
+                ADAPTERREMOVAL_SE.out.collapsed_truncated,
+                ADAPTERREMOVAL_PE.out.collapsed_truncated
+            )
+        ch_discarded_reads                    = ch_discarded_reads.mix(ADAPTERREMOVAL_SE.out.discarded, ADAPTERREMOVAL_PE.out.discarded)
+        ch_adapterremoval_paired_interleaved  = ADAPTERREMOVAL_SE.out.paired_interleaved.mix(ADAPTERREMOVAL_PE.out.paired_interleaved)
+        ch_versions                           = ch_versions.mix(ADAPTERREMOVAL_SE.out.versions.first(), ADAPTERREMOVAL_PE.out.versions.first())
```
jfy133: Does this work if only one of the two modules are run? I would rather put a dedicated `mix` after each module invocation independently

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
```diff
@@ -0,0 +1,148 @@
...
+include { LEEHOM                              } from '../../../modules/nf-core/leehom/main'
+// both SE and PE, plus merging
+include { FASTP                               } from '../../../modules/nf-core/fastp/main'
+include { ADAPTERREMOVAL as ADAPTERREMOVAL_SE } from '../../../modules/nf-core/adapterremoval/main'
+include { ADAPTERREMOVAL as ADAPTERREMOVAL_PE } from '../../../modules/nf-core/adapterremoval/main'
+
+workflow FASTQ_REMOVEADAPTERS_MERGE {
+
+    take:
+    reads                       // channel: [ val(meta), [ reads ] ]
```
jfy133:
```suggestion
    input_reads                       // channel: [ val(meta), [ reads ] ]
```

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
```diff
@@ -0,0 +1,148 @@
...
+                ADAPTERREMOVAL_PE.out.collapsed_truncated
+            )
+        ch_discarded_reads                    = ch_discarded_reads.mix(ADAPTERREMOVAL_SE.out.discarded, ADAPTERREMOVAL_PE.out.discarded)
+        ch_adapterremoval_paired_interleaved  = ADAPTERREMOVAL_SE.out.paired_interleaved.mix(ADAPTERREMOVAL_PE.out.paired_interleaved)
+        ch_versions                           = ch_versions.mix(ADAPTERREMOVAL_SE.out.versions.first(), ADAPTERREMOVAL_PE.out.versions.first())
+        ch_multiqc_files                      = ch_multiqc_files.mix(ADAPTERREMOVAL_PE.out.settings, ADAPTERREMOVAL_SE.out.settings)
+    }
+
+    emit:
+    reads                              = ch_reads                               // channel: [ val(meta), [ fastq.gz ] ]
```
jfy133:
```suggestion
    processed_reads                              = ch_reads                               // channel: [ val(meta), [ fastq.gz ] ]
```

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
jfy133: Technically all these inputs should be prefixed with `ch_` and `val_`, [^1] but you do that after so I will let it slide

[^1]: https://nf-co.re/docs/guidelines/components/subworkflows#input-channel-name-structure

jfy133: For all of these please indicate if they are mandatory or optional [^1]

[^1]: https://nf-co.re/docs/guidelines/components/subworkflows#code-comment-of-channel-structure

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
jfy133: My instinct was with Jon, but fair enough. We could try and push pipeline develoeprs to update teh original module by requiring meta here (even if we drop it in the subworkflow)

subworkflows/nf-core/fastq_removeadapters_merge/tests/main.nf.test
```diff
@@ -0,0 +1,246 @@
...
+    tag "trimmomatic"
+    tag "cutadapt"
+    tag "trimgalore"
+    tag "bbmap"
+    tag "bbmap/bbduk"
+    tag "leehom"
+    tag "fastp"
+    tag "adapterremoval"
+
+    test("sarscov2 - fastq - single-end") {
```
jfy133: If you switch to the thinbg we talked on slack (i.e., you can only run one clipping tool a ta time), we should have at least one test for each module + one test

subworkflows/nf-core/fastq_removeadapters_merge/tests/main.nf.test
```diff
@@ -0,0 +1,246 @@
...
+                input[10] = params.save_merged
+                input[11] = false               // skip_adapterremoval
+                input[12] = []                  // text adapters
+                """
+            }
+        }
+
+        then {
+            assert workflow.success
+            assert snapshot(
```
jfy133: Missing `assertAll()`? [^1]

[^1]: https://nf-co.re/docs/guidelines/components/subworkflows#assertall but I don't know why this is in the guidelines exactl;y 😅

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
```diff
@@ -0,0 +1,148 @@
...
+    ch_fastp_log                         = channel.empty()
+    ch_adapterremoval_paired_interleaved = channel.empty()
+    ch_versions                          = channel.empty()
+    ch_multiqc_files                     = channel.empty()
+
+    if (!skip_trimmomatic) {
+        TRIMMOMATIC( ch_reads )
+        ch_reads                      = TRIMMOMATIC.out.trimmed_reads
+        ch_discarded_reads            = ch_discarded_reads.mix(TRIMMOMATIC.out.unpaired_reads.transpose()) // .transpose() because paired reads have 2 unpaired files in an array
+        ch_trimmomatic_trim_log       = TRIMMOMATIC.out.trim_log
```
jfy133: If you go with single tool in single subworkflow invocation, I would suggest having a single 'log' output channel, rather than one per tool

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
```diff
@@ -0,0 +1,148 @@
...
+        ch_trimmomatic_summary        = TRIMMOMATIC.out.summary
```
jfy133: This could just be called 'report' (e.g. FASTP's htm could go here too)

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
```diff
@@ -0,0 +1,148 @@
...
+
+    if (!skip_fastp) {
+        FASTP(
+            ch_reads.map { meta, files ->  [ meta, files, contaminants ] },
+            fastp_discard_trimmed_pass,
+            fastp_save_trimmed_fail,
+            save_merged
+        )
+        ch_reads              = FASTP.out.reads
+        ch_merged_reads       = FASTP.out.reads_merged
```
jfy133: These are sort of mututally exclusive, I _think_ it makes more sense that you have an if/elkse around it - if `save_merged` (I presume is what controls this) then you mix in `if true ch_reads = FASTP.out.reads_merged else ch_reads= FASTP.out.reads`

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
```diff
@@ -0,0 +1,148 @@
...
+        ch_discarded_reads    = ch_discarded_reads.mix(FASTP.out.reads_fail.transpose()) // .transpose() because paired reads have 3 fail files in an array
+        ch_fastp_html         = FASTP.out.html
+        ch_fastp_log          = FASTP.out.log
+        ch_versions           = ch_versions.mix(FASTP.out.versions.first())
+        ch_multiqc_files      = ch_multiqc_files.mix(FASTP.out.json)
+    }
+
+    if (!skip_adapterremoval) {
+        ch_adapterremoval_in = ch_reads
```
jfy133: Another Question here just to double check, do all the other tools not care how they receive their data (i.e., can handle easily both)?

The reason why we have it separate for ADAPTEREMOVAL is more about the mkixing the output channels more than any thing else

jfy133 (commented):
Some initial comments - @vagkaratzas and I spoke on Slack in more detail about what I had in my mind when I suggested this subworkflow vs what is implemented here.

 I can see why Jon was a bit confused about the purpose of the subworkflow, hopefully it'll be clearer based on the refactor we discussed on Slack

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
vagkaratzas: We don't do that for modules that have been updated to the new version output syntax ;)
As in, it's kinda impossible without the pieces of code at the end of the new pipeline's template.
@maxulysse for back-up!

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
vagkaratzas: Same as above

subworkflows/nf-core/fastq_removeadapters_merge/tests/main.nf.test
vagkaratzas: Dog ate it

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
vagkaratzas: Yep! that's how it seems on the tests where nothing is skipped for both single and paired ends

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
vagkaratzas: After discussing with @jfy133, it seems this subworkflow will act as a plug-in interface, that will hide the complexity of the underlying tools from the user, while letting them pick their favourite adapter removal tool of choice.
i.e., one selected tool, and not chaining tools.
Updating accordingly..

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
vagkaratzas: Done

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
vagkaratzas: Done

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
vagkaratzas: Yep, seemed to be working fine during tests

subworkflows/nf-core/fastq_removeadapters_merge/tests/main.nf.test
vagkaratzas: Done

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
vagkaratzas: Done

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
```diff
@@ -0,0 +1,128 @@
...
+
+    main:
+
+    ch_discarded_reads    = channel.empty() // from trimmomatic, trimgalore, leehom, fastp, adapterremoval
+    ch_paired_interleaved = channel.empty() // from adapterremoval
+    ch_log                = channel.empty() // from trimmomatic, trimgalore, fastp
+    ch_report             = channel.empty() // from trimmomatic, trimgalore, fastp
+    ch_versions           = channel.empty()
+    ch_multiqc_files      = channel.empty() // from trimmomatic, cutadapt, bbduk, leehom, fastp, adapterremoval
+
```
jfy133, line 31: I would add a validation check here to make sure whatever is given to `val_adapter_tool` is recognised by the subworkflow.

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
```diff
@@ -0,0 +1,128 @@
...
+include { FASTP                               } from '../../../modules/nf-core/fastp/main'
+include { ADAPTERREMOVAL as ADAPTERREMOVAL_SE } from '../../../modules/nf-core/adapterremoval/main'
+include { ADAPTERREMOVAL as ADAPTERREMOVAL_PE } from '../../../modules/nf-core/adapterremoval/main'
+
+workflow FASTQ_REMOVEADAPTERS_MERGE {
+
+    take:
+    ch_input_reads                 // channel: [mandatory] meta, reads
+    val_adapter_tool               // string:  [mandatory] tool_name // choose from: ["trimmomatic", "cutadapt", "trimgalore", "bbduk", "leehom", "fastp", "adapterremoval"]
+    ch_adapters                    // channel: [optional]  {fasta,txt} // fasta, for bbduk or fastp, or txt, for adapterremoval
```
jfy133: For clarity of a pipeline dev maybe `ch_adapters_file` or a `ch_custom_adapters_file`?

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
```diff
@@ -0,0 +1,128 @@
...
+    ch_log                = channel.empty() // from trimmomatic, trimgalore, fastp
+    ch_report             = channel.empty() // from trimmomatic, trimgalore, fastp
+    ch_versions           = channel.empty()
+    ch_multiqc_files      = channel.empty() // from trimmomatic, cutadapt, bbduk, leehom, fastp, adapterremoval
+
+    if (val_adapter_tool == "trimmomatic") {
+        TRIMMOMATIC( ch_input_reads )
+
+        ch_processed_reads = TRIMMOMATIC.out.trimmed_reads
+        ch_discarded_reads = ch_discarded_reads.mix(TRIMMOMATIC.out.unpaired_reads.transpose()) // .transpose() because paired reads have 2 unpaired files in an array
```
jfy133: What do you mean by 2 unpaired files in an array? Do they represent different 'discarded' files? Or the discarded reads from each of the R1/R2? In the latter case does ti makes sense to separated them?

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
```diff
@@ -0,0 +1,128 @@
...
+        if (val_save_merged) {
+            ch_processed_reads = ADAPTERREMOVAL_SE.out.collapsed
+                .mix(
+                    ADAPTERREMOVAL_PE.out.collapsed,
+                    ADAPTERREMOVAL_SE.out.collapsed_truncated,
+                    ADAPTERREMOVAL_PE.out.collapsed_truncated
+                )
+        } else {
+            ch_processed_reads = ADAPTERREMOVAL_SE.out.singles_truncated.mix(ADAPTERREMOVAL_PE.out.paired_truncated)
+        }
```
jfy133: Please check against

https://github.com/nf-core/eager/blob/52da7b420a47b3703cc1bf12941409ad9ab61455/subworkflows/local/preprocessing_adapterremoval.nf#L38-L92

Preparing the AdapterRemoval reads is a little more complicated than just mixing, and you appear to have a mistake (there is  no `ADAPTERREMOVAL_SE.out.collapsed_truncated` -> collapsed here means merged, but you don't get merged SE data).

What I would recommend is to offer as 'single' output:

- SE (singles_truncated)
- Unmerged Pairs (paird_truncated)
- Merged pairs (collapsed + collapsed_truncated)

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
```diff
@@ -0,0 +1,128 @@
...
+        ch_discarded_reads    = ch_discarded_reads.mix(ADAPTERREMOVAL_SE.out.discarded, ADAPTERREMOVAL_PE.out.discarded)
+        ch_paired_interleaved = ADAPTERREMOVAL_SE.out.paired_interleaved.mix(ADAPTERREMOVAL_PE.out.paired_interleaved)
```
jfy133: To be hones,t I think it's very rare anyone uses these... IU would maybe drop support for them to simplify

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
vagkaratzas: There is this `else` condition at the end:
```
else {
        error('Please choose one of the available adapter removal and merging tools: ["trimmomatic", "cutadapt", "trimgalore", "bbduk", "leehom", "fastp", "adapterremoval"]')
    }
```
Is this not enough?

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
vagkaratzas: `ch_custom_adapters_file` seems like the best option, since each tool already uses its defaults ones. Updating

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
vagkaratzas: It's the latter (rephrased the comment a bit). `leehom` does something similar I think a bit further down the code. I would keep as it is and let the user handle it differently in the pipeline level if they wish.

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
vagkaratzas: Done

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
vagkaratzas: I copied over from eager; is this correct/needed though:
`ADAPTERREMOVAL_PAIRED.out.singles_truncated,`
?

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
jfy133: Ah yeah, missed that 😅 

I _personally_ would put it separately right at the beginning as I find it easier to 'read' and know the rest of subwf won't start until it passes. But that's just personal prefereence, this still works.

subworkflows/nf-core/fastq_removeadapters_merge/main.nf
```diff
@@ -0,0 +1,141 @@
...
+        ADAPTERREMOVAL_SE( ch_adapterremoval_in.single, ch_custom_adapters_file )
+        ADAPTERREMOVAL_PE( ch_adapterremoval_in.paired, ch_custom_adapters_file )
+
+        if (val_save_merged) { // merge
+            ch_concat_fastq = channel.empty()
+                .mix(
+                    ADAPTERREMOVAL_PE.out.collapsed,
+                    ADAPTERREMOVAL_PE.out.collapsed_truncated,
+                    ADAPTERREMOVAL_PE.out.singles_truncated,
+                    ADAPTERREMOVAL_PE.out.paired_truncated
```
jfy133: Remaining pairs (i.e., `non-collapsed`) should not be included here.
```suggestion
```

Here we want: reads that merged (collapsed) that were not trimmed, reads that were merged and adapter trimmed, and singletons (i.e., reads where one of the two in the pair got discarded for whatever reason, but the remaining read still being valid).

jfy133: Eager does something special there so it's slightly different but not really ideal.

PR #9525, opened 2025-12-08
Update xenium ranger modules and subworkflows

an-altosian:
This PR updates the xeniumranger version in its modules, and restructured the code to follow nf-core standard. Because xeniumranger is not on conda, I tested it using docker.

## PR checklist

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/bcftools/concat/tests/main.nf.test
+++ b/modules/nf-core/bcftools/concat/tests/main.nf.test
@@ -19,12 +19,12 @@ nextflow_process {
                 input[0] = [
                     [ id:'test3' ], // meta map
                     [
-                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878.chr22.1X.vcf.gz', checkIfExists: true),
-                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878_GIAB.chr22.vcf.gz', checkIfExists: true)
+                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878.chr21_22.1X.glimpse2.vcf.gz', checkIfExists: true),
+                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878_GIAB.chr21_22.vcf.gz', checkIfExists: true)
                     ],
                     [
-                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878_GIAB.chr22.vcf.gz.csi', checkIfExists: true),
-                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878.chr22.1X.vcf.gz.csi', checkIfExists: true)
+                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878_GIAB.chr21_22.vcf.gz.csi', checkIfExists: true),
+                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878.chr21_22.1X.glimpse2.vcf.gz.csi', checkIfExists: true)
                     ]
                 ]
                 """
@@ -50,12 +50,12 @@ nextflow_process {
                 input[0] = [
                     [ id:'test3' ], // meta map
                     [
-                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878.chr22.1X.vcf.gz', checkIfExists: true),
-                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878_GIAB.chr22.vcf.gz', checkIfExists: true)
+                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878.chr21_22.1X.glimpse2.vcf.gz', checkIfExists: true),
+                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878_GIAB.chr21_22.vcf.gz', checkIfExists: true)
                     ],
                     [
-                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878_GIAB.chr22.vcf.gz.csi', checkIfExists: true),
-                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878.chr22.1X.vcf.gz.csi', checkIfExists: true)
+                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878_GIAB.chr21_22.vcf.gz.csi', checkIfExists: true),
+                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878.chr21_22.1X.glimpse2.vcf.gz.csi', checkIfExists: true)
                     ]
                 ]
                 """
@@ -87,12 +87,12 @@ nextflow_process {
                 input[0] = [
                     [ id:'test3' ], // meta map
                     [
-                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878.chr22.1X.vcf.gz', checkIfExists: true),
-                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878_GIAB.chr22.vcf.gz', checkIfExists: true)
+                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878.chr21_22.1X.glimpse2.vcf.gz', checkIfExists: true),
+                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878_GIAB.chr21_22.vcf.gz', checkIfExists: true)
                     ],
                     [
-                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878_GIAB.chr22.vcf.gz.csi', checkIfExists: true),
-                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878.chr22.1X.vcf.gz.csi', checkIfExists: true)
+                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878_GIAB.chr21_22.vcf.gz.csi', checkIfExists: true),
+                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878.chr21_22.1X.glimpse2.vcf.gz.csi', checkIfExists: true)
                     ]
                 ]
                 """
@@ -124,12 +124,12 @@ nextflow_process {
                 input[0] = [
                     [ id:'test3' ], // meta map
                     [
-                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878.chr22.1X.vcf.gz', checkIfExists: true),
-                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878_GIAB.chr22.vcf.gz', checkIfExists: true)
+                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878.chr21_22.1X.glimpse2.vcf.gz', checkIfExists: true),
+                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878_GIAB.chr21_22.vcf.gz', checkIfExists: true)
                     ],
                     [
-                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878_GIAB.chr22.vcf.gz.csi', checkIfExists: true),
-                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878.chr22.1X.vcf.gz.csi', checkIfExists: true)
+                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878_GIAB.chr21_22.vcf.gz.csi', checkIfExists: true),
+                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878.chr21_22.1X.glimpse2.vcf.gz.csi', checkIfExists: true)
                     ]
                 ]
                 """
@@ -162,8 +162,8 @@ nextflow_process {
                 input[0] = [
                     [ id:'test3' ], // meta map
                     [
-                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878.chr22.1X.vcf.gz', checkIfExists: true),
-                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878_GIAB.chr22.vcf.gz', checkIfExists: true)
+                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878.chr21_22.1X.glimpse2.vcf.gz', checkIfExists: true),
+                        file(params.modules_testdata_base_path + 'genomics/homo_sapiens/illumina/vcf/NA12878_GIAB.chr21_22.vcf.gz', checkIfExists: true)
                     ],
                     []
                 ]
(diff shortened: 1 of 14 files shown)


modules/nf-core/xeniumranger/import-segmentation/tests/nextflow.config
```diff
@@ -0,0 +1,13 @@
+process {
+    withName: 'XENIUMRANGER_IMPORT_SEGMENTATION' {
+        ext.args = [
+            '--units=pixels'
```
khersameesh24: This might not always be true, running import-segmentation with baysor/proseg would need the `micron` coordinate space. Better to handle it like if `--coordinate-transform` then `units=microns` else `units=pixels`

modules/nf-core/xeniumranger/import-segmentation/main.nf
```diff
@@ -23,38 +23,28 @@ process XENIUMRANGER_IMPORT_SEGMENTATION {
...
+    prefix = task.ext.prefix ?: "${meta.id}"
+    def assembled_args = []
+    if (task.ext.args) { assembled_args << task.ext.args.trim() }
+    if (nuclei) { assembled_args << "--nuclei=\"${nuclei}\"" }
+    if (cells) { assembled_args << "--cells=\"${cells}\"" }
+    if (transcript_assignment) { assembled_args << "--transcript-assignment=\"${transcript_assignment}\"" }
+    if (viz_polygons) { assembled_args << "--viz-polygons=\"${viz_polygons}\"" }
+    if (coordinate_transform) {
+        assembled_args << "--coordinate-transform=\"${coordinate_transform}\""
+        assembled_args = assembled_args.replaceAll("--units=pixels", "--units=microns")
```
khersameesh24: ok, this works too. Ignore the earlier comment

modules/nf-core/xeniumranger/import-segmentation/meta.yml
```diff
@@ -28,58 +28,39 @@ input:
...
+        description: Optional transcript assignment CSV passed to `--transcript-assignment`
+    - viz_polygons:
+        type: file
+        optional: true
+        description: Optional visualization polygons GeoJSON passed to `--viz-polygons`
+    - coordinate_transform:
+        type: file
+        optional: true
+        description: Optional image alignment transform passed to `--coordinate-transform`
 output:
```
khersameesh24, line 105: It would be good to update the `meta.yml` file with `topics` for xeniumranger now

khersameesh24: https://nf-co.re/blog/2025/version_topics

modules/nf-core/xeniumranger/relabel/tests/main.nf.test
```diff
@@ -9,14 +9,14 @@ nextflow_process {
...
     tag "xeniumranger"
     tag "xeniumranger/relabel"
-    tag "unzip"
+    tag "untar"
 
     setup {
-            run("UNZIP") {
-                script "modules/nf-core/unzip/main.nf"
+            run("UNTAR") {
+                script "../../../untar/main.nf"
```
khersameesh24: Are absolute paths not working anymore? It would be better to have `modules/nf-core/untar/main.nf` here as the nf-tests are run from the root directory

modules/nf-core/xeniumranger/import-segmentation/meta.yml
an-altosian: Thanks! I will update it accordingly!

modules/nf-core/xeniumranger/resegment/main.nf
```diff
@@ -23,28 +19,25 @@ process XENIUMRANGER_RESEGMENT {
...
+    def args = (task.ext.args ?: "").trim()
+    def args_block = args ? "\\\n        ${args}" : ""
+    prefix = task.ext.prefix ?: "${meta.id}"
 
     """
+    rm -rf "${prefix}"
+
     xeniumranger resegment \\
-        --id="${prefix}" \\
+        --id="XENIUMRANGER_RESEGMENT" \\
```
khersameesh24, line 29: this should be reverted back to `$prefix`

subworkflows/nf-core/xeniumranger/import_segmentation_transcript/main.nf
```diff
@@ -0,0 +1,41 @@
+include { XENIUMRANGER_IMPORT_SEGMENTATION } from '../../../../modules/nf-core/xeniumranger/import-segmentation/main'
```
khersameesh24: subworkflows usually have 2 or more modules being run, I don't think there is a need for a xeniumranger subworkflows. I do not see any use case for these in any pipeline as a single module subworkflow, instead the module can be used itself in that case

modules/nf-core/xeniumranger/import-segmentation/tests/nextflow.config
an-altosian: Yaa, I did not put units in the input channel cuz it is not mandatory. Do you think we should add it in?

modules/nf-core/xeniumranger/relabel/tests/main.nf.test
an-altosian: Absolute paths work. I will update accordingly.

modules/nf-core/xeniumranger/resegment/main.nf
an-altosian: My thought is, as we only need the `${prefix}/outs/`, we can name the entire output dir as a fixed name, like `XENIUMRANGER_RESEGMENT`, move the `outs/` from it, and mark that as the outdir. By doing so, the `outdir` is basically the updated bundle dir, so the input bundle dir and output bundle dir have the same structure. What do you think?

subworkflows/nf-core/xeniumranger/import_segmentation_transcript/main.nf
an-altosian: Good point. I imagined these two sub-workflows as the abstractions of the underlying module, because the module itself has two sets of mutually exclusive parameters (which is pretty annoying I would say). If we don't want this, could I ask your thoughts on a better way to making sure folks know how to use the module? Or ... am I just overthinking?

khersameesh24: You can always include comments in the `main.nf` file. Also a README can be included I guess. But I think the way modules and submodules are built in nf-core we should ideally have just one import-segmentation module.

modules/nf-core/xeniumranger/import-segmentation/tests/nextflow.config
khersameesh24: Not as a input channel but a condition to check if the segmentation polygons are passed with the
 `--coordinate-transform` option, if so, then `--units=microns` else pixels

modules/nf-core/xeniumranger/import-segmentation/main.nf
khersameesh24: This is correct, it can be removed from `ext.args`

modules/nf-core/xeniumranger/resegment/main.nf
khersameesh24: this would work, but then, there would be some changes to the way it is used in `spatialxe`, that is my concern. These modules are only part of spatialxe, no other pipeline/subworkflow use these in nf-core, they might be used elsewhere ofcourse, but lets keep it catered to `spatialxe` . The only step we need to do there is then to update it with `nf-core modules update`

an-altosian: This is a valuable concern! I think both of use are trying to make the nf-core version of the modules to be more flexible and follow the nf-core standards. Because the changes to the `versions` channel, we need to modify spatialxe anyhow. So, I am slightly prefer to output `outs/` here to make the module output cleaner. If you agree, I can help make a PR to spatialxe accordingly to adopt these changes.

subworkflows/nf-core/xeniumranger/import_segmentation_transcript/main.nf
DongzeHE: Cool. I removed the subworkflows.

modules/nf-core/xeniumranger/import-segmentation/main.nf
DongzeHE: I still used your old spec and made the units as an argument. It is neater this way.

DongzeHE:
Hi @khersameesh24 , I think the PR is ready. I tried my best to address all your comments. Please let me know if you have further thoughts. Some notes: 
1. I removed the subworkflows for import-segmentataion and added a check in the module to ensure the image mode args and transcript mode args are not specified simultaneously.
2. I still think exporting just the `outs/` dir, which corresponds to the updated bundle, is neater. So I left the change there.
3. I added `--units`  back to the arg list of import-segmentation, because the replaceAll part looked ugly.

modules/nf-core/xeniumranger/import-segmentation/meta.yml
```diff
@@ -28,66 +33,95 @@ input:
...
+        description: Path to the Xenium output bundle generated by the Xenium
+          Onboard Analysis pipeline
+    - transcript_assignment:
+        type: file
+        optional: true
+        description: |
+          Transcript assignment CSV with cell assignment, such as from Baysor v0.6, (transcript-based mode).
+          Mutually exclusive with image-based inputs (`nuclei`, `cells`). Required when using
+          transcript-based mode. Passed to `--transcript-assignment`.
+        ontologies: []
```
khersameesh24: I think if you add `pattern:` , then the linter would be able to update the ontologies with the 
`nf-core modules lint xeniumranger --fix` command

modules/nf-core/xeniumranger/rename/tests/main.nf.test.snap
```diff
@@ -7,59 +7,116 @@
...
+                            ],
+                            "nucleus_boundaries.csv.gz:md5,e417b6e293298870956d42c7106cbd0c",
+                            "nucleus_boundaries.parquet:md5,bacbfc3c2e956d899e1d8ccba5dd7c5e",
+                            "transcripts.parquet:md5,203cb05ee7689805cc505ebda9557551",
+                            "transcripts.zarr.zip:md5,807e63a2ef8340e085cd899507f45395"
+                        ]
+                    ]
+                ],
+                "versions": [
+                    "versions.yml:md5,0c04440360057514cedbad984e64ea93"
```
khersameesh24: `versions.yml` does not exist anymore. Were the snapshots updated? Maybe try 
`nf-core modules test xeniumranger --update`

modules/nf-core/xeniumranger/nextflow.config
khersameesh24, line 1: this one can be removed I guess

modules/nf-core/xeniumranger/README.md
khersameesh24, line 1: Please keep the readme file here

modules/nf-core/xeniumranger/import-segmentation/main.nf
```diff
@@ -2,80 +2,67 @@ process XENIUMRANGER_IMPORT_SEGMENTATION {
...
-    path(cells)
-    path(transcript_assignment)
-    path(viz_polygons)
+    tuple val(meta), path(xenium_bundle, stageAs: "bundle/"), path(transcript_assignment), path(viz_polygons), path(nuclei), path(cells), path(coordinate_transform), val(units)
 
     output:
-    tuple val(meta), path("**/outs/**"), emit: outs
-    path "versions.yml", emit: versions
+    tuple val(meta), path("${prefix}"), emit: outs
+    tuple val("${task.process}"), val("xeniumranger"), eval("xeniumranger -V | sed -e 's/xeniumranger-/- /g'"), emit: versions_xeniumranger, topic: versions
```
khersameesh24:
```suggestion
    tuple val("${task.process}"), val("xeniumranger"), eval("xeniumranger -V | sed -e 's/.*xenium-//'"), emit: versions_xeniumranger, topic: versions
```

We just need the version number now in the snap file

modules/nf-core/xeniumranger/relabel/main.nf
```diff
@@ -2,55 +2,46 @@ process XENIUMRANGER_RELABEL {
...
     input:
-    tuple val(meta), path(xenium_bundle)
-    path(gene_panel)
+    tuple val(meta), path(xenium_bundle, stageAs: "bundle/"), path(panel)
 
     output:
-    tuple val(meta), path("**/outs/**"), emit: outs
-    path "versions.yml", emit: versions
+    tuple val(meta), path("${prefix}"), emit: outs
+    tuple val("${task.process}"), val("xeniumranger"), eval("xeniumranger -V | sed -e 's/xeniumranger-/- /g'"), emit: versions_xeniumranger, topic: versions
```
khersameesh24:
```suggestion
    tuple val("${task.process}"), val("xeniumranger"), eval("xeniumranger -V | sed -e 's/.*xenium-//'"), emit: versions_xeniumranger, topic: versions
```

modules/nf-core/xeniumranger/rename/main.nf
```diff
@@ -2,57 +2,49 @@ process XENIUMRANGER_RENAME {
...
-    tuple val(meta), path(xenium_bundle)
-    val(region_name)
-    val(cassette_name)
+    tuple val(meta), path(xenium_bundle, stageAs: "bundle/"), val(region_name), val(cassette_name)
 
     output:
-    tuple val(meta), path("**/outs/**"), emit: outs
-    path "versions.yml", emit: versions
+    tuple val(meta), path("${prefix}"), emit: outs
+    tuple val("${task.process}"), val("xeniumranger"), eval("xeniumranger -V | sed -e 's/xeniumranger-/- /g'"), emit: versions_xeniumranger, topic: versions
```
khersameesh24:
```suggestion
    tuple val("${task.process}"), val("xeniumranger"), eval("xeniumranger -V | sed -e 's/.*xenium-//'"), emit: versions_xeniumranger, topic: versions
```

modules/nf-core/xeniumranger/resegment/main.nf
```diff
@@ -2,69 +2,44 @@ process XENIUMRANGER_RESEGMENT {
...
-    val(dapi_filter)
-    val(boundary_stain)
-    val(interior_stain)
+    tuple val(meta), path(xenium_bundle, stageAs: "bundle/")
 
     output:
-    tuple val(meta), path("**/outs/**"), emit: outs
-    path "versions.yml", emit: versions
+    tuple val(meta), path("${prefix}"), emit: outs
+    tuple val("${task.process}"), val("xeniumranger"), eval("xeniumranger -V | sed -e 's/xeniumranger-/- /g'"), emit: versions_xeniumranger, topic: versions
```
khersameesh24:
```suggestion
    tuple val("${task.process}"), val("xeniumranger"), eval("xeniumranger -V | sed -e 's/.*xenium-//'"), emit: versions_xeniumranger, topic: versions
```

modules/nf-core/xeniumranger/relabel/tests/main.nf.test.snap
```diff
@@ -38,29 +38,41 @@
...
+                            "experiment.xenium:md5,d41d8cd98f00b204e9800998ecf8427e"
+                        ]
                     ]
                 ],
                 "1": [
-                    "versions.yml:md5,ab2584177544560d5a9e9c36f7d24354"
+                    [
+                        "XENIUMRANGER_RELABEL",
+                        "xeniumranger",
+                        "xeniumranger xenium-4.0.1.1"
```
khersameesh24: this section gets updated if the suggestions above are implemented for the version string

modules/nf-core/xeniumranger/rename/meta.yml
```diff
@@ -24,35 +24,45 @@ input:
...
+            Ranger
+          pattern: "${prefix}"
+  versions_xeniumranger:
+    - - "${task.process}":
+          type: string
+          description: The process the versions were collected from
+      - xeniumranger:
+          type: string
+          description: The tool name
+      - "xeniumranger -V | sed -e 's/xeniumranger-/- /g'":
```
khersameesh24: this would also need an update as the version string is updated from the suggestion above

khersameesh24 (commented):
Hi @an-altosian , left some comments. Would just need a small update, everything else looks good. Thanks for working on this again.

modules/nf-core/xeniumranger/rename/tests/main.nf.test.snap
an-altosian: Good point. The github workflows passed but the outdated version files were still there. I think the reason is, because ther is not version files anymore, some tests now do not involve checksum verification at all, so the sections for these tests in the snapshots were ignored. Tag @jfy133 here, because I am not sure if this behavior is the optimal.

I deleted the snapshots and remade them. Now the version files are gone.

There is another interesting behavior. Say I have two tests, the first test is for `scripts` and second is for `stub`. Only the `stub` test has entries in the snapshot and the real test doesn't. In this case, the stub test's entries will be written on the top of the snapshot. Later, if I add some checksum checks (for the version topic) to the `scripts` test, the entries of the `scripts` test will be appended to the snapshot. That is saying, the two tests are recorded in the opposite order in `main.nf.test` and `main.nf.test.snap`. In this case, `nf-core modules lint`, will report an error saying `test_snap_versions: versions not found in snapshot`, even though this is recorded at the bottom of the file. I think it is probably a bug.

modules/nf-core/xeniumranger/nextflow.config
an-altosian: Thanks. I will remove it when we are ready to merge because the tests runs forever :(

DongzeHE:
Hi @khersameesh24 , all addressed. Thank you for the comments and suggestions!

PR #9530, opened 2025-12-09
remove topics from multiqc

nvnieuwk:
https://nfcore.slack.com/archives/C09LJTQQ3EY/p1764864053598659

### DIFF
--- a/modules/nf-core/multiqc/meta.yml
+++ b/modules/nf-core/multiqc/meta.yml
@@ -73,18 +73,6 @@ output:
         description: Plots created by MultiQC
         pattern: "*_data"
         ontologies: []
-  versions_multiqc:
-    - - ${task.process}:
-          type: string
-          description: The process the versions were collected from
-      - multiqc:
-          type: string
-          description: The tool name
-      - multiqc --version | sed "s/.* //g":
-          type: eval
-          description: The expression to obtain the version of the tool
-
-topics:
   versions:
     - - ${task.process}:
           type: string


nvnieuwk:
Linting will fail because there's no topic channel for the versions output here, but that's exactly what we want here

modules/nf-core/multiqc/main.nf
```diff
@@ -18,7 +18,7 @@ process MULTIQC {
     path "*.html"      , emit: report
     path "*_data"      , emit: data
     path "*_plots"     , optional:true, emit: plots
-    tuple val("${task.process}"), val('multiqc'), eval('multiqc --version | sed "s/.* //g"'), topic: versions, emit: versions_multiqc
+    tuple val("${task.process}"), val('multiqc'), eval('multiqc --version | sed "s/.* //g"'), emit: versions_multiqc
```
mashehu: how if we do `emit: version`, then linting might not compalin

nvnieuwk: Sounds like a good workaround :)

modules/nf-core/multiqc/main.nf
```diff
@@ -18,7 +18,7 @@ process MULTIQC {
     path "*.html"      , emit: report
     path "*_data"      , emit: data
     path "*_plots"     , optional:true, emit: plots
-    tuple val("${task.process}"), val('multiqc'), eval('multiqc --version | sed "s/.* //g"'), topic: versions, emit: versions_multiqc
+    tuple val("${task.process}"), val('multiqc'), eval('multiqc --version | sed "s/.* //g"'), emit: versions
```
nvnieuwk, line 21:
```suggestion
    tuple val("${task.process}"), val('multiqc'), eval('multiqc --version | sed "s/.* //g"'), emit: versions
    // MultiQC should not push its versions to the `versions` topic. Its input depends on the versions topic to be resolved thus outputting to the topic will let the pipeline hang forever
```

nvnieuwk:
Good idea! I added a comment

PR #9543, opened 2025-12-10
Migrate cat/fastq to topic channel

Aratz:
Migrate the cat/fastq module to topic channel

#### Note
* `nf-core modules lint` fails with 3.5.1 but this is fixed in 3.6.0dev


mashehu (approved):

mashehu:
shouldn't the subworkflow snapshot also change 🤔

Aratz:
I thought so to, but it seems like it the version channel is not included in the snapshot. Should I edit the test and fix it?

mashehu:
that would be great, but understand if you don't want to go to deep down this rabbit hole

PR #9562, opened 2025-12-11
Allow `dream` to process non-formula contrasts

delfiterradas:
## PR checklist

Closes https://github.com/nf-core/differentialabundance/issues/578 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/variancepartition/dream/tests/main.nf.test
+++ b/modules/nf-core/variancepartition/dream/tests/main.nf.test
@@ -20,7 +20,7 @@ nextflow_process {
                 input[0] = Channel.of([
                     'id': 'genotype_WT_KO_treatment_Control_Treated',
                     'formula': '~ genotype * treatment',
-                    'comparison': 'genotypeKO - treatmentTreated' // should be a 'make.names() string'
+                    'comparison': 'genotypeWT - treatmentTreated' // should be a 'make.names() string'
                 ])
                 .map {
                     tuple(it, it.variable, it.reference, it.target, it.formula, it.comparison)
@@ -39,8 +39,8 @@ nextflow_process {
             assertAll(
                 { assert process.success },
                 { assert path(process.out.results[0][1]).getText().contains("gene_id\tlogFC\tAveExpr\tt\tP.Value\tadj.P.Val\tB") },
-                { assert path(process.out.results[0][1]).getText().contains("0.83") },
-                { assert path(process.out.results[0][1]).getText().contains("13.5\t21.12\t4.41") },
+                { assert path(process.out.results[0][1]).getText().contains("1309.83") },
+                { assert path(process.out.results[0][1]).getText().contains("557\t279.88\t18.81") },
                 { assert snapshot(process.out.model, process.out.versions).match() }
             )
         }
@@ -54,7 +54,7 @@ nextflow_process {
             }
             process {
                 """
-                input[0] = Channel.of(['id': 'treatment_mCherry_hND6', 'formula':'~ treatment', 'comparison':'treatmenthND6'])
+                input[0] = Channel.of(['id': 'treatment_mCherry_hND6', 'formula':'~ treatment', 'comparison':'treatmentmCherry'])
                     .map{
                         tuple(it, it.variable, it.reference, it.target, it.formula, it.comparison)
                     }
@@ -73,8 +73,8 @@ nextflow_process {
             assertAll(
                 { assert process.success },
                 { assert path(process.out.results[0][1]).getText().contains("gene_id\tlogFC\tAveExpr\tt\tP.Value\tadj.P.Val\tB") },
-                { assert path(process.out.results[0][1]).getText().contains("849.67") },
-                { assert path(process.out.results[0][1]).getText().contains("1050\t549\t3.78") },
+                { assert path(process.out.results[0][1]).getText().contains("-849.67") },
+                { assert path(process.out.results[0][1]).getText().contains("-1050\t549\t-3.78") },
                 { assert snapshot(process.out.model, process.out.versions).match() }
             )
         }
--- a/modules/nf-core/variancepartition/dream/tests/main.nf.test.snap
+++ b/modules/nf-core/variancepartition/dream/tests/main.nf.test.snap
@@ -24,17 +24,24 @@
     "RNAseq - Feature Counts - formula + comparison contrast string - interaction": {
         "content": [
             [
-                
+                [
+                    {
+                        "id": "genotype_WT_KO_treatment_Control_Treated",
+                        "formula": "~ genotype * treatment",
+                        "comparison": "genotypeWT - treatmentTreated"
+                    },
+                    "genotype_WT_KO_treatment_Control_Treated.dream.model.txt:md5,e67d5a9c42dd9374b370b6426832d08f"
+                ]
             ],
             [
-                
+                "versions.yml:md5,fc1f26eb2194018e99fc2916332676b7"
             ]
         ],
         "meta": {
             "nf-test": "0.9.3",
             "nextflow": "25.04.2"
         },
-        "timestamp": "2025-12-23T19:04:58.864804168"
+        "timestamp": "2025-12-26T15:10:51.980662299"
     },
     "Mus musculus - contrasts - matrix - no formula": {
         "content": [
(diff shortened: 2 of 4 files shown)


grst:
I think you still need to take the baseline level into account when building contrasts, as done here for DESeq2: 

https://github.com/nf-core/modules/blob/8846461fd688f29921c28dd4182913969a4f36c1/modules/nf-core/deseq2/differential/templates/deseq2_differential.R#L450-L457

DESeq2 allows to test contrasts with a `c(variable, target, reference)` tuple. I don't know of DREAM offers the same. If it doesn't, you'll need to manually build the contrast or choose appropriate baseline levels for the factor with the contrast variable.

modules/nf-core/variancepartition/dream/templates/dream.R
```diff
@@ -128,15 +175,27 @@ if (!is.null(opt\$exclude_samples_col) && !is.null(opt\$exclude_samples_values))
 }
 
 # Load count data
-countMatrix <- read_delim_flexible(opt\$count_file, header = TRUE, stringsAsFactors = FALSE)
+countMatrix <- read_delim_flexible(opt\$count_file, header = TRUE, stringsAsFactors = FALSE, check.names = FALSE)
 rownames(countMatrix) <- countMatrix\$gene_id # count_file/matrix must have a gene_id column.
 countMatrix <- countMatrix[, rownames(metadata), drop = FALSE]
 
-# Construct model formula using user-provided formula if available; if not, default to contrast variable only
+# Construct model formula using user-provided formula if available; if not, generate formula from variables
```
pinin4fjords, line 219: Not part of your changes, but maybe the check below should be `if (is_valid_string(opt$formula))` as well?

modules/nf-core/variancepartition/dream/templates/dream.R
```diff
@@ -95,7 +99,7 @@ for (ao in names(args_opt)) {
 }
 
 # If there is no option supplied, convert string "null" to NULL
-keys <- c("formula", "contrast_string", "contrast_variable", "contrast_reference")
+keys <- c("formula", "contrast_string", "contrast_variable", "blocking_variable", "contrast_reference")
```
pinin4fjords:
```suggestion
keys <- c("formula", "contrast_string", "contrast_variable", "blocking_variables", "contrast_reference")
```

modules/nf-core/variancepartition/dream/templates/dream.R
```diff
@@ -116,6 +120,49 @@ if (!is.null(opt\$round_digits)){
...
+        )
+      )
+    }
+  }
+
+  # Optionally, subset to only the samples involved in the contrast
+  if (opt\$subset_to_contrast_samples){
+    sample_selector <- metadata[[contrast_variable]] %in% c(opt\$contrast_target, opt\$contrast_reference)
+    selected_samples <- metadata[sample_selector, opt\$sample_id_col]
+    count.table <- count.table[, selected_samples]
```
pinin4fjords: Looking at the code flow - the subset happens before countMatrix is defined. In DESeq2/limma this works because the count table is loaded before the validation block. Here, the order is:

  1. Load metadata
  2. Validation + subset ← tries to subset count.table which doesn't exist
  3. Load countMatrix

modules/nf-core/variancepartition/dream/templates/dream.R
```diff
@@ -7,6 +7,9 @@ library("BiocParallel")
 library("limma")
 
 # Load auxiliary helping functions
+is_valid_string <- function(input) {
```
pinin4fjords, line 22: Just for tidiness, could you give this the roxygen2-style docs please, like the other functions?

modules/nf-core/variancepartition/dream/templates/dream.R
```diff
@@ -172,28 +244,41 @@ fitmm <- eBayes(fitmm, proportion = opt\$proportion,
...
+    contrast_string <- opt\$contrast_string
+} else if (has_intercept) {
+  contrast_string <- ""
+    # Intercept model → use coef (reference level is implicit)
+    coef_name <- paste0(opt\$contrast_variable, opt\$contrast_target)
+    cat("Using default contrast matrix:", coef_name, "\n")
+
+    results <- topTable(fitmm, coef = coef_name, number = Inf,
+                        adjust.method = opt\$adjust.method, p.value = opt\$p.value,
+                        lfc = opt\$lfc, confint = opt\$confint)
```
grst: Do we need this case? 
I thought we only accept a formula in combination with a contrast_string. 

If we built the formula within this script, we already know that it is a zero-intercept model. 

I think combining custom formula with (baseline, variable, target) tuples will open up too many edge cases. So rather simplify here and add a check for that.

delfiterradas: I have adapted the logic from `limma` for creating the contrasts taking this into account. Do you think it makes sense now?

modules/nf-core/variancepartition/dream/templates/dream.R
```diff
@@ -174,26 +246,49 @@ print(colnames(fitmm\$design))
...
+        # Only the target level is present in the design matrix
+        # The reference level may have been omitted due to collinearity or being set as the baseline
+        # We compare the target level to zero (implicit reference)
+        contrast_string <- ""
+        coef_name <- paste0(opt\$contrast_variable, opt\$contrast_target)
+        cat("Using default contrast matrix:", coef_name, "\n")
+
+        results <- topTable(fitmm, coef = coef_name, number = Inf,
+                            adjust.method = opt\$adjust.method, p.value = opt\$p.value,
+                            lfc = opt\$lfc, confint = opt\$confint)
```
grst: I think we can error in this case?

Perfect colinearity of the factors would show that it is an incorrect model. 
Reference level as baseline can't happen if we always create a zero-intercept model.

grst:
So regarding the eBayes, it is necessary to call it after contrasts.fit, according to this example from the limma user guide: 
image
https://bioconductor.org/packages//2.11/bioc/vignettes/limma/inst/doc/usersguide.pdf

It wouldn't be necessary to call it *before* contrasts.fit, but it doesn't harm. (we still need it if we don't call contrasts.fit at all, but results() directly on the first fit).

modules/nf-core/variancepartition/dream/templates/dream.R
grst: @pinin4fjords, I'm inclined to keep things simple and avoid possible error sources by making the following options mutually exclusive: 

 * specify a formula and a constrast_string
 * specify a contrast tuple `(variable, baseline, target)` and `blocking_factors`. 

I.e. I would raise an error if someone specifies a formula together with a contrast tuple. Does that sound alright to you?

pinin4fjords:
I'll defer to @grst on approval for this

grst:
ok, then here are my last two requests: 

- [ ] Can you please add a check that (variable, reference, target) contrasts are not combined with a custom formula? 
- [ ] remove the special case when factors are missing from the design matrix, just fail instead. 

Like that we focus on the usecases we currently need for differentialabundance and keep it simple.

delfiterradas:
@grst 
> ok, then here are my last two requests:
> 
> * [ ]  Can you please add a check that (variable, reference, target) contrasts are not combined with a custom formula?
> * [x]  remove the special case when factors are missing from the design matrix, just fail instead.
> 
> Like that we focus on the usecases we currently need for differentialabundance and keep it simple.

I will remove the special case now, but I am not sure how to approach the check for the contrast tuple since the script currently uses the `contrast_variable` and `contrast_reference` to relevel the metadata:
```
if (!is.null(opt\$contrast_reference) && opt\$contrast_reference != "") {
    metadata[[opt\$contrast_variable]] <- factor(metadata[[opt\$contrast_variable]])
    metadata[[opt\$contrast_variable]] <- relevel(metadata[[opt\$contrast_variable]], ref = opt\$contrast_reference)
}
```
I could make the check only for `blocking_variables` and `contrast_reference` but then that wouldn't allow subsetting to contrast samples.

Would the check be necessary if we already have a validation where `contrast_string` must be provided if using `formula`?

grst:
I don't think the relevel or subset to contrast samples should work together with a custom contrast/formula. At the pipeline level, when specifying a formula and a contrast string, contrast variable, target and reference won't be set. 

You can write the check such that this possibility is excluded.

delfiterradas:
Okay, I have added the validation so that the module fails when a formula is used together with the contrast tuple. 

Some of the tests are failing because no explicit reference level is set and therefore no releveling is performed, so sometimes the "target" level is absorbed into the intercept by the default:
```
    >   ~genotype * treatment
    >   Warning message:
    >   Zero sample variances detected, have been offset away from zero 
    >           (Intercept) genotypeWT treatmentTreated genotypeWT:treatmentTreated
    >   Sample1           1          1                0                           0
    >   Sample2           1          1                0                           0
    >   Sample3           1          1                0                           0
    >   [1] "(Intercept)"                 "genotypeWT"                 
    >   [3] "treatmentTreated"            "genotypeWT:treatmentTreated"
    >   Using contrast string: genotypeKO - treatmentTreated 
    >   Error in eval(ej, envir = levelsenv) : object 'genotypeKO' not found
```
In this case, genotype (KO) is encoded in the intercept by default and therefore does not exist as a standalone coefficient in the design matrix and cannot be directly referenced in a contrast expression.

@grst Do you think this is behavior that users are generally expected to understand and handle themselves?

grst:
yes, users are expected to understand how this works if they use formulas and contrast strings. 

One thing we should consider is how to explicitly set the baseline level of the factors (I believe currently it's the default, i.e. alphabetically). But that's a different topic that we should discuss at the pipeline level first.

PR #9579, opened 2025-12-17
Add vcfpgloader/load module

### LINKED ISSUE #9578: new module: `vcf-pg-loader`
## Description

Add a new module for **vcf-pg-loader**, a high-performance VCF to PostgreSQL loader using asyncpg for bulk variant ingestion.

## Tool Information

- **Name**: vcf-pg-loader
- **Homepage**: https://github.com/Zacharyr41/vcf-pg-loader
- **License**: MIT
- **Description**: Async PostgreSQL VCF loader using cyvcf2 and asyncpg for loading VCF files into postgres databases with high throughput. 
## Use Case

This module enables loading annotated VCF files directly into PostgreSQL databases, which is useful for:
- Clinical variant analysis pipelines
- Building searchable variant databases
- Integration with downstream analysis tools

## Notes

- Tool is containerized on GHCR (`ghcr.io/zacharyr41/vcf-pg-loader`)
- Not yet on Bioconda, but PR is submitted [there too](https://github.com/bioconda/bioconda-recipes/pull/61423) (module includes conda profile error check)
- Uses Nextflow secrets for database password handling

Zacharyr41:
Closes #9578

## Summary

High-throughput VCF to PostgreSQL loader using asyncpg for bulk variant ingestion.

- Adds `vcfpgloader/load` module for loading VCF variants into PostgreSQL databases
- Uses cyvcf2 for VCF parsing and asyncpg for high-performance database operations
- Supports batch loading with configurable batch size and parallel workers
- Outputs JSON report with loading statistics and detailed log

## Test plan

- [x] Stub tests pass with sarscov2 and homo_sapiens test data
- [x] `nf-core modules lint vcfpgloader/load` passes (43/43 tests)
- [x] Pre-commit hooks pass

### DIFF
--- a/modules/nf-core/vcfpgloader/load/tests/main.nf.test.snap
+++ b/modules/nf-core/vcfpgloader/load/tests/main.nf.test.snap
@@ -9,6 +9,22 @@
                     "test_human.load_report.json:md5,0c600325e85cdf50ee557916606fe2b1"
                 ]
             ],
+            [
+                [
+                    {
+                        "id": "test_human"
+                    },
+                    "test_human.load.log:md5,d41d8cd98f00b204e9800998ecf8427e"
+                ]
+            ],
+            [
+                [
+                    {
+                        "id": "test_human"
+                    },
+                    "0"
+                ]
+            ],
             {
                 "versions_vcfpgloader": [
                     [
@@ -43,6 +59,15 @@
                         "family": "TEST_FAM"
                     },
                     "test_sample.load.log:md5,d41d8cd98f00b204e9800998ecf8427e"
+                ]
+            ],
+            [
+                [
+                    {
+                        "id": "test_sample",
+                        "family": "TEST_FAM"
+                    },
+                    "0"
                 ]
             ],
             {


Zacharyr41:
@nf-core/modules-team tagging per [this instruction](https://nf-co.re/docs/contributing/modules#uploading-to-nf-coremodules:~:text=config%20%2Dstub%2Drun-,Uploading%20to%20nf%2Dcore/modules,-When%20you%20are)

This is my first PR here so I apologize in advance if I missed anything!

Zacharyr41:
This is waiting on a PR from bioconda to bump to version 0.5.4 (since there was some weird version drift going on), at which point I will request review from nf-core. Just commenting here to document progress.

modules/nf-core/vcfpgloader/load/main.nf
famosab, line 1: Can you run
```
```
on this file? :)

modules/nf-core/vcfpgloader/load/main.nf
```diff
@@ -0,0 +1,67 @@
...
+    conda "${moduleDir}/environment.yml"
+    container "ghcr.io/zacharyr41/vcf-pg-loader:0.5.4"
+
+    input:
+    tuple val(meta), path(vcf), path(tbi)
+    val(db_host)
+    val(db_port)
+    val(db_name)
+    val(db_user)
+    val(db_schema)
```
famosab: I would suggest to have all these val inputs in the first input channel. That way we ensure that they are fixed to the respective input.
```
tuple val(meta), path(vcf), path(tbi), val(db_host), val(dp_port) ...
```

modules/nf-core/vcfpgloader/load/main.nf
```diff
@@ -0,0 +1,67 @@
...
+        $args \\
+        $vcf
+
+    ROWS_LOADED=\$(jq -r '.variants_loaded' ${prefix}.load_report.json)
+    """
+
+    stub:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    echo '{"status": "stub", "variants_loaded": 0, "elapsed_seconds": 0}' > ${prefix}.load_report.json
```
famosab: We do not really need to put anything in these stub files but you can leave it as is if you want.

modules/nf-core/vcfpgloader/load/tests/tags.yml
famosab, line 1: This file is not needed anymore :) You can delete it.

modules/nf-core/vcfpgloader/load/tests/nextflow.config
```diff
@@ -0,0 +1,14 @@
+params {
+    modules_testdata_base_path = 'https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/'
+}
+
```
famosab: We should not need this here.
```suggestion
```

modules/nf-core/vcfpgloader/load/tests/main.nf.test
```diff
@@ -0,0 +1,82 @@
...
+
+    tag "modules"
+    tag "modules_nfcore"
+    tag "vcfpgloader"
+    tag "vcfpgloader/load"
+    tag "database"
+
+    test("sarscov2 - vcf.gz - stub") {
+        tag "stub"
+        options "-stub"
```
famosab, line 16: Is there no way to actually test the tool without stub?

The stub tests look fine to me.

modules/nf-core/vcfpgloader/load/tests/main.nf.test
```diff
@@ -0,0 +1,82 @@
...
+        then {
+            assertAll(
+                { assert process.success },
+                { assert path(process.out.report.get(0).get(1)).exists() },
+                { assert path(process.out.log.get(0).get(1)).exists() },
+                { assert snapshot(
+                    process.out.report,
+                    process.out.log,
+                    process.out.findAll { key, val -> key.startsWith("versions")}
+                ).match() }
```
famosab: We do not need to assert if the files exist if we also add them to the snapshot. And we want to add the row_count here I think.

```suggestion
                { assert snapshot(
                    process.out.report,
                    process.out.log,
                    process.out.row_count,
                    process.out.findAll { key, val -> key.startsWith("versions")}
                ).match() }
```

modules/nf-core/vcfpgloader/load/tests/main.nf.test
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert path(process.out.report.get(0).get(1)).exists() },
+                { assert snapshot(
+                    process.out.report,
+                    process.out.findAll { key, val -> key.startsWith("versions")}
+                ).match() }
```
famosab, line 77: Why are we not evaluating log and row_count?
```suggestion
                { assert process.success },
                { assert snapshot(
                    process.out.report,
                    process.out.log,
                    process.out.row_count,
                    process.out.findAll { key, val -> key.startsWith("versions")}
                ).match() }
```

modules/nf-core/vcfpgloader/load/main.nf
```diff
@@ -0,0 +1,67 @@
...
+//   env {
+//       PGPASSWORD = secrets.PGPASSWORD
+//   }
+
+process VCFPGLOADER_LOAD {
+    tag "$meta.id"
+    label 'process_medium'
+
+    conda "${moduleDir}/environment.yml"
+    container "ghcr.io/zacharyr41/vcf-pg-loader:0.5.4"
```
famosab: This needs to either be pushed to the nf-core repo or you could use seqera containers to create docker and singularity containers from your bioconda package. See https://seqera.io/containers/. Also for bioconda packages there should be a biocontainer available. But if you need both packages (bioconda::vcf-pg-loader=0.5.4 and conda-forge::jq=1.7) for this to work I would suggest using seqera containers. And then you can add both a docker and a singularity container here (see other modules for how they did it).

modules/nf-core/vcfpgloader/load/main.nf
Zacharyr41: Resolved almost all the issues. There was one regarding the `ROWS_LOADED` env var that seems to be a linter bug, although I could be completely wrong? I added a commit message explaining this, but let me know if this is still an issue.

modules/nf-core/vcfpgloader/load/main.nf
Zacharyr41: 🙏

modules/nf-core/vcfpgloader/load/tests/tags.yml
Zacharyr41: Done :)

modules/nf-core/vcfpgloader/load/main.nf
Zacharyr41: Done :)

modules/nf-core/vcfpgloader/load/tests/main.nf.test
Zacharyr41: Yeah, the tool needs a running PostgreSQL database with the vcf-pg-loader schema set up, so real integration testing in CI gets a bit complicated - we'd need to spin up a Postgres container, init the schema, run tests, then clean up.

Since stub tests cover the module interface (inputs, outputs, file patterns) and the tool itself is tested in its own repo, I figured stub-only was the pragmatic choice here. But happy to look into testcontainers if you'd prefer real integration tests! (Really happy to do whatever the standard practice is here, just didn't know since this is my first PR :) )

modules/nf-core/vcfpgloader/load/main.nf
Zacharyr41: Removed the jq dependency and now it should depend on only bioconda and so I switched to just biocontainers.

Zacharyr41:
@famosab thank you so much for the thoughtful and comprehensive review,  I really appreciate it 🙏 . Learned a lot by implementing your feedback. Let me know if there is anything that I misinterpreted/did not implement properly. Really thank you so much!

modules/nf-core/vcfpgloader/load/tests/main.nf.test
famosab: Ah yes that makes a lot of sense to me. Honestly I do not know too much about running PostgreSQL with the CI. There might be other modules that use some kind of SQL as well, so it might be worth a quick search through this repository to see if there are other cases.

I do agree that in theory the tool does not need testing. But we usually aim to test the tool within the module as in "Does this tool work as expected in the Nextflow framework". Creating snapshots also enables us to see whether the outputs are stable across different containers etc. As this is "only" a load module its probably enough to leave it at a stub test though.

Zacharyr41: Cool! I'd be happy to look into this a bit and search for prior art here on integrating postgres into testing and then I'll get back to you! 🙏

Zacharyr41: Thanks for the suggestion! I searched through the repo and couldn't find any prior art for testing against 
PostgreSQL or other database servers in CI - vcf-pg-loader appears to be the first module that connects to an actual database server. Previous iterations testing against databases like sqlite test against the `.db` files but not an active DB connection. Given that and your point about stub tests being sufficient for a load module, I'm happy to leave it as-is with stub tests.

Once I get a bit more familiar with the nf-core ecosystem, however, I'd be happy to submit a PR allowing for testing against containerized dependencies such as a postgres database, if that would be helpful. 

Wondering if you could lend your LGTM now :)

famosab (approved):
Lets go 🚀

Issue #9574, opened 2025-12-16
new module: blast/rpsblast -> local-cd-search


Ales-ibt:
### Is there an existing module for this?

- [x] I have searched for the existing module

### Is there an open PR for this?

- [x] I have searched for existing PRs

### Is there an open issue for this?

- [x] I have searched for existing issues

### Further Information

The plan is to add `blast/rpsblast`, `blast/makeprofiledb`, and `rpsbproc` modules to complete the suit of tools needed to run [RPS-BLAST](https://www.ncbi.nlm.nih.gov/books/NBK279684/table/appendices.T.rpsblast_application_option/). I will make a PR per module using the `feature/rpsblast` branch.

### Are you going to work on this?

- [x] If I'm planning to work on this module, I added myself to the `Assignees` to facilitate tracking who is working on the module

Ales-ibt:
I had closed the PRs and removed the branches for rpsblast and cdddownloader as we decided to change strategy after discussion with the nf-core community. The new approach is creating an nf-core module for [local-cd-search](https://github.com/apcamargo/local-cd-search).

PR #9582, opened 2025-12-17
Module to download CDD databases

Ales-ibt:
This module was developed to download and process the CDD databases necessary to run rpsblast from the FTP site of the NCBI.

## PR checklist
- [X] Add a resource `label`
  - For modules:
    - [X] `nf-core modules test blast/cdddownloader --profile docker`


modules/nf-core/blast/cdddownloader/meta.yml
```diff
@@ -0,0 +1,77 @@
...
+      homepage: "https://www.gnu.org/software/wget/"
+      documentation: "https://www.gnu.org/software/wget/manual/wget.html"
+      licence: ["GPL"]
+      identifier: ""
+  - untar:
+      description: |
+        Extract tar, tar.gz, tar.bz2, tar.xz files.
+      documentation: https://www.gnu.org/software/tar/manual/
+      licence: ["GPL-3.0-or-later"]
+      identifier: ""
```
vagkaratzas: As I see it, here you describe the tool under which the modules falls, in this case that should be `blast`, and you are making a misc downloader for it; not the wget and untar ones.
See the `arriba/download` and `checkm2/databasedownload` modules for reference.

modules/nf-core/blast/cdddownloader/main.nf
```diff
@@ -0,0 +1,70 @@
...
+    rm -f ${db_name}.tar.gz
+
+    echo "Downloading metadata files"
+
+    wget https://ftp.ncbi.nih.gov/pub/mmdb/cdd/cddid.tbl.gz -O ./data/cddid.tbl.gz && gzip -d ./data/cddid.tbl.gz
+    wget https://ftp.ncbi.nih.gov/pub/mmdb/cdd/cdtrack.txt -O ./data/cdtrack.txt
+    wget https://ftp.ncbi.nih.gov/pub/mmdb/cdd/family_superfamily_links -O ./data/family_superfamily_links
+    wget https://ftp.ncbi.nih.gov/pub/mmdb/cdd/cddannot.dat.gz -O ./data/cddannot.dat.gz && gzip -d ./data/cddannot.dat.gz
+    wget https://ftp.ncbi.nih.gov/pub/mmdb/cdd/cddannot_generic.dat.gz -O ./data/cddannot_generic.dat.gz && gzip -d ./data/cddannot_generic.dat.gz
+    wget https://ftp.ncbi.nih.gov/pub/mmdb/cdd/bitscore_specific.txt -O ./data/bitscore_specific.txt
```
vagkaratzas: Are these necessary to be downloaded? Next to every db every time?

Ales-ibt: One set of them is enough for any database. These are "universal"

vagkaratzas: I am wondering how this module will handle multiple dbs. You'll have to call it as many times as dbs you want, and then all these files will be downloaded again, and then you'll have multiple copies of the `cdd_databases` folder.
I am not familiar with the tree structure, but what's the plan for moving everything in a single cdd_databases folder to be used by the pipeline for multiple dbs? Or is the idea to only use one db at a time?

Ales-ibt: The rpsblast will run with one databse at the time. The Cdd database contains all the rest of the DBs. In case someone want to run the analysis versus all the databases, should choose Cdd or a the corresponding subset. The idea is to do the check if exists in the calling workflow and run the downloader only when the database is not present. I am not sure how to tell the pipeline to fix the location of the downloaded DB. We probably should do something like this:
https://nf-co.re/funcscan/3.0.0/parameters/#database-downloading-options

Ales-ibt: I will add an input for the downloading dir and checks for existing dbs

modules/nf-core/blast/cdddownloader/meta.yml
Ales-ibt: Fixed

modules/nf-core/blast/cdddownloader/main.nf
Ales-ibt: I have added handlers to deal with existing db folders

Ales-ibt:
I am closing this PR as we decided to change strategy after discussion with the nf-core community. The new approach is creating an nf-core module for [local-cd-search](https://github.com/apcamargo/local-cd-search) instead of the rpsblast and cdddownloader. The corresponding branches will be removed and [issue 9574](https://github.com/nf-core/modules/issues/9574) will be updated accordingly.

PR #9595, opened 2025-12-21
Swap mpileup to topics

SPPearce:
Swap samtools/mpileup to topics

### DIFF
--- a/modules/nf-core/samtools/mpileup/tests/main.nf.test
+++ b/modules/nf-core/samtools/mpileup/tests/main.nf.test
@@ -27,10 +27,7 @@ nextflow_process {
         then {
             assert process.success
             assertAll(
-                { assert snapshot(
-                    process.out.mpileup,
-                    process.out.findAll { key, val -> key.startsWith("versions") }
-                    ).match() }
+                { assert snapshot(process.out).match()}
             )
         }
     }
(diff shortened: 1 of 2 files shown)


modules/nf-core/poolsnp/main.nf
```diff
@@ -1,6 +1,6 @@
 process POOLSNP {
     tag "$meta.id"
-    label 'process_medium'
+    label 'process_single'
```
SPPearce:
```suggestion
    label 'process_medium'
```

modules/nf-core/samtools/mpileup/tests/main.nf.test
```diff
@@ -28,8 +28,8 @@ nextflow_process {
             assert process.success
             assertAll(
                 { assert snapshot(
-                    process.out,
-                    path(process.out.versions[0]).yaml
+                    process.out.mpileup,
+                    process.out.findAll { key, val -> key.startsWith("versions") }
```
mashehu: why not just `snapshot(process.out)` these are already all outputs?

modules/nf-core/samtools/mpileup/meta.yml
```diff
@@ -54,16 +54,30 @@ output:
...
+      - samtools:
+          type: string
+          description: Name of the tool
+      - samtools --version | head -1 | sed -e "s/samtools //":
+          type: eval
+          description: The expression to obtain the version of the tool
+
 authors:
-  - "@drpatelh"
   - "@joseespinosa"
```
mashehu, line 82: not sure about removing people from `authors`. `maintainers` yes, but `authors` feels more like a record

SPPearce: Fair, although I don't really see the point of the authors field.

PR #9598, opened 2025-12-21
Migrate samtools dict/faidx/merge to topics

SPPearce:
Migrate samtools dict/faidx/merge to topics

### DIFF
--- a/modules/nf-core/samtools/faidx/tests/main.nf.test
+++ b/modules/nf-core/samtools/faidx/tests/main.nf.test
@@ -9,7 +9,7 @@ nextflow_process {
     tag "samtools"
     tag "samtools/faidx"
     config "./nextflow.config"
-    
+
     test("test_samtools_faidx") {
 
         when {


modules/nf-core/samtools/faidx/tests/main.nf.test
```diff
@@ -12,6 +12,9 @@ nextflow_process {
     test("test_samtools_faidx") {
```
LouisLeNezet, line 13: You defined module_args but you did not add config `./nextflow.config` in the header

subworkflows/nf-core/fasta_index_methylseq/tests/main.nf.test.snap
```diff
@@ -141,16 +140,15 @@
                 
             ],
             [
-                "versions.yml:md5,1a7e59286a0e918ba2400750ddefd444",
```
LouisLeNezet, line 144: The fact that the topic channel is not present in the snapshot worries me a bit.
We lose the versions md5sum of the tool that did the migration.

modules/nf-core/samtools/faidx/tests/main.nf.test
SPPearce: Good spot. Added it back in.

modules/nf-core/samtools/faidx/tests/main.nf.test
```diff
@@ -8,10 +8,14 @@ nextflow_process {
     tag "modules_nfcore"
     tag "samtools"
     tag "samtools/faidx"
-
+    config "./nextflow.config"
```
LouisLeNezet, line 11: Could you modify the nextflow.config to use params.module_args instead of "ext.args = 'MT192765.1 -o extract.fa'"

LouisLeNezet (changes requested):
Just a small change needed for the config

PR #9599, opened 2025-12-22
Update vireo

LuisHeinzlmeier:
Add two optional output files for the hadge pipeline.

### DIFF
--- a/modules/nf-core/vireo/main.nf
+++ b/modules/nf-core/vireo/main.nf
@@ -23,6 +23,8 @@ process VIREO {
 
     script:
     def args   = task.ext.args   ?: ''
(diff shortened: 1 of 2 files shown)


LuisHeinzlmeier:
Thank you @SPPearce for the review! I added the optional files to `stub`. There were only consistent results for `filtered_variants.tsv` if I set `--randSeed` in `nextflow.config`. I hope this is the correct way to handle randomness in modules!?

SPPearce:
Personally I like to hardcode a default random seed like:
https://github.com/nf-core/modules/blob/61cd0d3823f5596e86f09d55d21f96c871845d19/modules/nf-core/umitools/group/main.nf#L33

PR #9612, opened 2025-12-31
Swap mosdepth to topic

SPPearce:


modules/nf-core/mosdepth/main.nf
```diff
@@ -49,15 +49,17 @@ process MOSDEPTH {
...
-    cat <<-END_VERSIONS > versions.yml
-    "${task.process}":
-        mosdepth: \$(mosdepth --version 2>&1 | sed 's/^.*mosdepth //; s/ .*\$//')
-    END_VERSIONS
     """
 
     stub:
+    def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
+    if (bed && args.contains("--by")) {
```
fellen31: Should also check the short option `-b`.

modules/nf-core/mosdepth/main.nf
```diff
@@ -49,15 +49,17 @@ process MOSDEPTH {
...
+        error "'--by' can only be specified once when running mosdepth! Either remove input BED file definition or remove '--by' from 'ext.args' definition"
+    }
+    if (!bed && args.contains("--thresholds")) {
```
fellen31: I don't understand why this check needed? 

`--thresholds` should be allowed even without bed, as long as args contains `--by`.  But, should't the software be able to give that error itself?

SPPearce: I don't know, it was just in the module for the `script` block so I just copied to the `stub` block too. It is more useful in the `stub` block actually, as that doesn't actually use the tool so wouldn't fail.

fellen31: Since this was just a copy we can fix it later if you want. But `--thresholds` should be allowed even without bed, since the tool allows it.

fellen31 (approved):
LGTM, but there are some if statements in this module that should be updated at some point.

PR #9636, opened 2026-01-12
Fix nf-schema version for language server

nvnieuwk:
The language server complains about the subworkflow if the test version is lower than 2.6.1, this fixes that issue

### DIFF
--- a/.github/actions/nf-test-action/action.yml
+++ b/.github/actions/nf-test-action/action.yml
@@ -25,7 +25,7 @@ runs:
     - name: Set up Nextflow
       uses: nf-core/setup-nextflow@v2
       with:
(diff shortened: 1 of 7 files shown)


nvnieuwk:
No this only bumps that version for the subworkflow test. The pipeline itself still uses the configured version (no functional changes here)

mashehu:
ah, my bad

nvnieuwk:
No worries

mashehu:
but i guess we need to bump the nextflow version of the test then

JoseEspinosa:
What happened with the tests... 😢

JoseEspinosa:
Ah the snapshots need to be updated probably

nvnieuwk:
Yes on it!

PR #9666, opened 2026-01-15
fix(fastk/merge): use topic versions and update snapshots

prototaxites:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


modules/nf-core/fastk/merge/meta.yml
```diff
@@ -48,7 +48,7 @@ output:
           description: |
             Groovy Map containing sample information
             e.g. [ id:'test', single_end:false ]
-      - "*.ktab*":
+      - '"*.ktab*", hidden: true':
```
charles-plessy: I tried `"*.ktab*"` on my computer and the linting went without error. Is it really needed to hide something here?

modules/nf-core/fastk/merge/meta.yml
```diff
@@ -58,17 +58,31 @@ output:
           description: |
             Groovy Map containing sample information
             e.g. [ id:'test', single_end:false ]
-      - "*.{prof,pidx}*":
+      - '"*.{prof,pidx}*", hidden: true':
```
charles-plessy: Same here, is `hidden: true` really needed?

modules/nf-core/fastk/merge/tests/main.nf.test
```diff
@@ -49,7 +49,13 @@ nextflow_process {
         then {
             assertAll(
                 { assert process.success },
-                { assert snapshot(process.out).match() }
```
charles-plessy, line 52: I tried in my computer with `{ assert snapshot(process.out).match() }` and it works fine. Is it really needed to check each output channel one by one?

modules/nf-core/fastk/merge/meta.yml
prototaxites: It seems to be a regression in the dev version of tools: https://nfcore.slack.com/archives/CE5LG7WMB/p1768480946931369

I agree we should just use `*.ktab*` here ideally! But we do need to capture the hidden files in the module as the kmer databases consist of a normal index file and N hidden files with a `.` prefix.

modules/nf-core/fastk/merge/tests/main.nf.test
prototaxites: This one was also passing fine locally on my computer, but the snapshot was not stable when run on CI. I think the merging of two tables is non-deterministic in some way that might depend on e.g. the locale...

prototaxites:
@mashehu This last one that requires a force-merge, if you have a moment? 🙏

PR #9672, opened 2026-01-15
feat(merquryfk/merquryfk): use topic versions and emit all images in single channel

prototaxites:
A few changes here:
- merquryfk/merquryfk and merquryfk/hapmaker both use topic versions
- merquryfk/merquryfk emits the image files in a single output channel
- I fixed the tests for merquryfk/merquryfk, as it requires the outputs from merqury/hapmaker, and the channel joins in the old tests were also resulting in empty inputs. In doing this I discovered that in trio mode, there is some library loading issue that stops the test from completing. I've commented out the trio test in the mean time, and we can investigate if it is a data or a container issue at a later date.

edit: this appears to be a bug in merquryfk, it refers to a wrong file name internally. Not sure if it's been fixed as there hasn't been a new version tagged in a while...

```
    rm: error while loading shared libraries: libc.so.6: cannot open shared object file: Error 24
    Logex: Cannot open table test_mat
    MerquryFK: Command 'Logex -T2 -H1000 '._CN_PLT.1ZLS2M.0=B-A' '._CN_PLT.1ZLS2M.1=B&.A[1]' '._CN_PLT.1ZLS2M.2=B&.A[2]' '._CN_PLT.1ZLS2M.3=B&.A[3]' '._CN_PLT.1ZLS2M.4=B&.A[4]' '._CN_PLT.1ZLS2M.5=B&.A[5-]' '._CN_PLT.1ZLS2M.6=A-B' ._MQY_A1.xou5Lf test_mat' failed

```

edit2: Yes, this is a bug in the tagged version of merquryFK - I can run the command in `.command.sh` fine with the latest commit. I'll endeavour to get Gene to tag a release, but this might take some time or not be possible.

edit3: MerquryFK bumped to 1.2 and added a trio test dataset. Now the tests pass and the tool works correctly :) 

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/merquryfk/merquryfk/tests/main.nf.test
+++ b/modules/nf-core/merquryfk/merquryfk/tests/main.nf.test
@@ -82,11 +82,8 @@ nextflow_process {
                 { assert process.success },
                 { assert snapshot(
                     process.out.stats,
-                    process.out.bed,
                     process.out.qv,
                     process.out.assembly_qv,
-                    process.out.phased_block_bed,
-                    process.out.phased_block_stats,
                     process.out.images.collect { meta, images -> [ meta, images.collect { img -> file(img).name } ] },
                     process.out.findAll { key, val -> key.startsWith("versions")}
                 ).match() }
@@ -122,11 +119,8 @@ nextflow_process {
                 { assert process.success },
                 { assert snapshot(
                     process.out.stats,
-                    process.out.bed,
                     process.out.qv,
                     process.out.assembly_qv,
-                    process.out.phased_block_bed,
-                    process.out.phased_block_stats,
                     process.out.images.collect { meta, images -> [ meta, images.collect { img -> file(img).name } ] },
                     process.out.findAll { key, val -> key.startsWith("versions")}
                 ).match() }
@@ -163,7 +157,6 @@ nextflow_process {
                 { assert process.success },
                 { assert snapshot(
                     process.out.stats,
-                    process.out.bed,
                     process.out.qv,
                     process.out.assembly_qv,
                     process.out.phased_block_bed,
--- a/modules/nf-core/merquryfk/merquryfk/tests/main.nf.test.snap
+++ b/modules/nf-core/merquryfk/merquryfk/tests/main.nf.test.snap
@@ -16,15 +16,6 @@
                         "id": "test",
                         "single_end": false
                     },
-                    "test.Hap1_only.bed:md5,d41d8cd98f00b204e9800998ecf8427e"
-                ]
-            ],
-            [
-                [
-                    {
-                        "id": "test",
-                        "single_end": false
-                    },
                     "test.qv:md5,781d72926d1c9a9be5761ce1601df4f8"
                 ]
             ],
@@ -36,12 +27,6 @@
                     },
                     "test.Hap1.qv:md5,bf3fb8514963e67ce7663f642b15be8e"
                 ]
-            ],
-            [
-                
-            ],
-            [
-                
             ],
             [
                 [
@@ -87,7 +72,7 @@
             "nf-test": "0.9.3",
             "nextflow": "25.10.0"
         },
-        "timestamp": "2026-01-21T10:37:29.544498"
+        "timestamp": "2026-01-21T10:52:00.275823"
     },
     "MerquryFK - Synthetic trio data - both haps - no trio": {
         "content": [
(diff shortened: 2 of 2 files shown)


modules/nf-core/merquryfk/hapmaker/main.nf
```diff
@@ -17,7 +17,10 @@ process MERQURYFK_HAPMAKER {
     output:
     tuple val(meta) , path("*${input_mat}.hap.ktab*", hidden: true), emit: mat_hap_ktab
     tuple val(meta2), path("*${input_pat}.hap.ktab*", hidden: true), emit: pat_hap_ktab
-    path "versions.yml"                                            , emit: versions
+    // WARN: Version information not provided by tool on CLI. Please update this string when bumping container versions.
+    tuple val("${task.process}"), val('merquryfk'), eval('echo 1.1.1'), emit: versions_merquryfk, topic: versions
+    // WARN: Version information not provided by tool on CLI. Please update this string when bumping container versions.
+    tuple val("${task.process}"), val('fastk'), eval('echo 1.1'), emit: versions_fastk, topic: versions
```
charles-plessy: I wondered why `1.1.0` was changed to `1.1` and I checked on GitHub. Release says `1.1` and the git tag says `1.1.0`. The sources do not mention any version number at all. Is there a compelling reason to change the version number reported by the module?

modules/nf-core/merquryfk/hapmaker/tests/main.nf.test
```diff
@@ -65,7 +65,11 @@ nextflow_process {
         then {
             assertAll(
                 { assert process.success },
-                { assert snapshot(process.out).match() }
```
charles-plessy, line 68: I am curious why it was necessary to stop using `{ assert snapshot(process.out).match() }`?

modules/nf-core/merquryfk/hapmaker/main.nf
prototaxites: I changed it to be in line with what the other FastK modules report, which is 1.1: https://github.com/nf-core/modules/blob/master/modules/nf-core/fastk/fastk/main.nf

modules/nf-core/merquryfk/hapmaker/tests/main.nf.test
prototaxites: Yeah, I'm not sure what my logic was here either... I'll revert this one.

prototaxites:
I tried bumping to the latest 1.1.3 release to see if it fixes the error with trio inputs, but no dice. So I'm happy to leave that disabled for now and to see if we can get the issue fixed later.

edit: I have found the bug and created a PR to fix it upstream, also: https://github.com/thegenemyers/MERQURY.FK/pull/20

prototaxites:
OK, I've spoken with the author and he is has written and will tag a fix for this bug. So I'll hold off on merging this PR until that's sorted and the new version is live, then incorporate that so we have a working module.

modules/nf-core/merquryfk/hapmaker/main.nf
```diff
@@ -17,7 +17,10 @@ process MERQURYFK_HAPMAKER {
     output:
     tuple val(meta) , path("*${input_mat}.hap.ktab*", hidden: true), emit: mat_hap_ktab
     tuple val(meta2), path("*${input_pat}.hap.ktab*", hidden: true), emit: pat_hap_ktab
-    path "versions.yml"                                            , emit: versions
+    // WARN: Version information not provided by tool on CLI. Please update this string when bumping container versions.
+    tuple val("${task.process}"), val('merquryfk'), eval('echo 1.2'), emit: versions_merquryfk, topic: versions
+    // WARN: Version information not provided by tool on CLI. Please update this string when bumping container versions.
+    tuple val("${task.process}"), val('fastk'), eval('echo 1.2'), emit: versions_fastk, topic: versions
```
mashehu, line 23: why no R version like in the other module?

prototaxites: This one doesn't actually call R, it's actually a wrapper around some of the FastK tools. I just kept the environment and container the same to save some space on disk, as you're unlikely to ever use this tool without running MerquryFK proper

prototaxites:
OK, almost there - this module needs some real trio test data as the test fails due to a floating point error. I'll add this to test-datasets, then we should be good to go.

prototaxites:
https://github.com/nf-core/test-datasets/pull/1841

prototaxites:
OK, I think this is good to go! @mashehu, might you be able to force-merge again? 🙏 This one also suffers from the `hidden: true` linting issue.

mashehu:
👍🏻 will wait for the tests to pass though

prototaxites:
@mashehu And tests passing here 🙏

mashehu:
a-woman-in-a-helmet-with-the-words-did-we-do-it-below-her

PR #9674, opened 2026-01-15
Update amulety modules

ggabernet:
Update amulety modules: in amulety>=2.0 the command syntax and subcommands have changed and I updated the nf-core modules are accordingly.


Joon-Klaps (commented):
There are some rules on deprecating modules. https://nf-co.re/docs/guidelines/components/module_deprecation

Think it might be better to make a `amulety2`?  What do you think?

ggabernet:
Ah thanks, I wasn't aware of this. I don't think it's needed to use AMULETY2, since the deprecated commands are no longer available. Instead of deleting the old modules I will deprecate them.

ggabernet:
Hi @Joon-Klaps I've deprecated the old modules now instead of just removing them. Let me know if the PR now looks good to you when you have a minute

modules/nf-core/amulety/embed/meta.yml
```diff
@@ -0,0 +1,82 @@
...
+        type: file
+        description: TSV with BCR/TCR nucleotide sequences in AIRR rearrangement format
+        pattern: "*.{tsv}"
+        ontologies:
+          - edam: http://edamontology.org/format_3475 # TSV
+  - chain:
+      type: string
+      description: |
+        The chain to use for the embeddings. Can be either 'H' (high diversity chain: BCR heavy or TCR beta or delta),
+        'L' (low diversity chain: BCR light or TCR alpha or gamma) or 'HL' (BCR heavy + light, and TCR alpha + beta or gamma + delta).
```
Joon-Klaps:
```suggestion
  - chain:
      type: string
      description: |
        The chain to use for the embeddings. Can be either 'H' (high diversity chain: BCR heavy or TCR beta or delta),
        'L' (low diversity chain: BCR light or TCR alpha or gamma) or 'HL'/'LH' (BCR heavy + light, and TCR alpha + beta or gamma + delta).
       enum: ['L', 'H', 'HL', 'LH'] 
```

modules/nf-core/amulety/embed/meta.yml
```diff
@@ -0,0 +1,82 @@
...
+    - - meta:
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. `[ id:'sample1', single_end:false ]`
+      - "*.tsv":
+          type: file
+          description: TSV with embeddings of BCR or TCR amino acid sequences.
+          pattern: "*.tsv"
+
```
Joon-Klaps:
```suggestion
```

modules/nf-core/amulety/embed/main.nf
```diff
@@ -0,0 +1,41 @@
...
+    output:
+    tuple val(meta), path("*.tsv"), emit: embedding
+    tuple val("${task.process}"), val('amulety'), eval("amulety --help 2>&1 | grep -o 'version [0-9\\.]\\+' | grep -o '[0-9\\.]\\+'"), emit: versions_amulety, topic: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args   = task.ext.args   ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
```
Joon-Klaps, line 25: You don't suspect a filename clash, as both input and output files are `tsv` files? If so, here is the suggestion
```suggestion
    def prefix = task.ext.prefix ?: "${meta.id}"
    if ("${tsv}" == "${prefix}.tsv") error "Input and output names are the same, use \"task.ext.prefix\" to disambiguate!"

```

modules/nf-core/amulety/embed/main.nf
```diff
@@ -0,0 +1,41 @@
...
+        embed \\
+        $args \\
+        --input-airr $tsv \\
+        --chain $chain \\
+        --model $model \\
+        --output-file-path ${prefix}.tsv
+    """
+
+    stub:
+    def prefix = task.ext.prefix ?: "${meta.id}"
```
Joon-Klaps, line 39:
```suggestion
    def prefix = task.ext.prefix ?: "${meta.id}"
    if ("${input}" == "${prefix}") error "Input and output names are the same, use \"task.ext.prefix\" to disambiguate!"

```

modules/nf-core/amulety/embed/main.nf
ggabernet: Yes, good point!

ggabernet:
Thanks for the review @Joon-Klaps , I've now addressed the points and the tests are passing again

PR #9676, opened 2026-01-16
fix(picard/markduplicates): add missing bai index file to stub output

adamrtalbot:
## Why
The stub output for `picard/markduplicates` was missing the `.bai` index file, causing downstream processes that expect the index file to fail during stub runs/testing.

## What
- Added missing `touch ${prefix}.${suffix}.bai` to stub block to generate the expected BAI index file alongside the BAM output


mashehu:
just update to the current version of `master` and it should be a lot less 😉

PR #9679, opened 2026-01-16
version topics, strict syntax, stub tests

vagkaratzas:
version topics, strict syntax, stub tests

## PR checklist

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/subworkflows/nf-core/mafft_align/main.nf
+++ b/subworkflows/nf-core/mafft_align/main.nf
@@ -1,0 +1,14 @@
+include { MAFFT_ALIGN as MAFFT_ALIGN_MODULE } from '../../../modules/nf-core/mafft/align/main'
+
+workflow MAFFT_ALIGN {
+
+    take:
+    ch_fasta // channel: [ val(meta), fasta ]
+
+    main:
+
+    MAFFT_ALIGN_MODULE ( ch_fasta, [[], []], [[], []], [[], []], [[], []], [[], []], [] )
+
+    emit:
+    alignment = MAFFT_ALIGN_MODULE.out.fas      // channel: [ val(meta), *.fas ]
+}
--- a/subworkflows/nf-core/mafft_align/meta.yml
+++ b/subworkflows/nf-core/mafft_align/meta.yml
@@ -1,0 +1,50 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/subworkflows/yaml-schema.json
+name: "mafft_align"
+description: Prepare channels for running MAFFT/align
+keywords:
+  - msa
+  - mafft
+  - align
+components:
+  - mafft/align
+input:
+  - ch_fasta:
+      type: file
+      description: |
+        The input channel containing the FASTA files
+      pattern: "*.{bam/cram/sam}"
+      structure:
+        - - meta:
+              type: map
+              description: |
+                Groovy Map containing sample information
+                e.g. `[ id:'sample1', single_end:false ]`
+          - fasta:
+              type: file
+              description: Input sequences in FASTA format
+              pattern: "*.{fa,fasta}"
+              ontologies:
+                - edam: http://edamontology.org/format_1929
+output:
+  - alignment:
+      type: file
+      description: Channel containing the alignment file in fasta format.
+      structure:
+        - - meta:
+              type: map
+              description: |
+                Groovy Map containing sample information
+                e.g. `[ id:'test']`
+          - "*.aln.gz":
+              type: file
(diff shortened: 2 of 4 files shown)


modules/nf-core/mafft/align/tests/main.nf.test
```diff
@@ -29,7 +29,10 @@ nextflow_process {
         then {
             assertAll(
                 { assert process.success },
-                { assert snapshot(process.out).match("SARS-CoV-2 scaffolds fasta - uncompressed")}
+                { assert snapshot(
+                    process.out.fas,
+                    process.out.findAll { key, val -> key.startsWith("versions")}
```
mashehu:
```suggestion
                    process.out
```

subworkflows/nf-core/mafft_align/main.nf
mashehu, line 1: needs a deprecation, or what happened here?

vagkaratzas: This was a subworkflow with 1 module (that shouldn't exist based on the nf-core subworkflow rules), that wasn't used in any nf-core pipeline. It was a test for mafft_align (which is a module) essentially as far as I understand, that has its own nf-tests.

mashehu: @mirpedrol because you added this subworkflow, is it okay to be removed?

mirpedrol: don't remove it please, I have work in progress to add this to the multiplesequencealign pipeline. 
The reason of this subworkflow is that we are working on a way to benchmark modules that are interchangeable, this means that they must perform the same operation and have the same input and output. For this module mafft/align, we need some manual tuning on how the module is called, so I wrapped in into a subworkflow. The subworkflow basically allows all the automation to work.
If this is super against the guidelines I am open to discuss, but I think it doesn't harm to leave it here :)

vagkaratzas: Then maybe something like [this](https://github.com/nf-core/modules/blob/master/subworkflows/nf-core/fastq_removeadapters_merge/main.nf)?
But again, something like this would need at least two different tools, whose inputs / outputs would be normalized to work the same. 
Anyway, I'll put it back for now, but I'll have to remove the versions since mafft/align does not output them anymore, with the new topics syntax.

mirpedrol: What do you mean by the example you posted?
This is an edge case. I don't think it can follow the standard rules; it doesn't make sense to add an additional tool that is not used just to make it fit in the rules.
The problem is not that the channels have to be modified, but the way the automation to include this module works in our POC, basically this module has several channels that have exactly the same structure, but to call it in a standard way only one has to be provided, the rest are optional, and this has to be selected manually.
I can point to the open PRs if you are interested :) 

I can also work on updating this subworkflo once the module is updated, but I don't have time to do it now. If it is better for you, I would suggest to ignore the subworkflow for now and open an issue about it. I will tackle this as soon as I find a moment.

Thanks!

vagkaratzas: I see, no problem in ignoring it at all.
The subworkflow I posted provides 7 different tools to choose 1 from, but expects the same inputs and outputs (more or less) no matter which one is used. I guessed that you wanted to do something similar.
Just curious, is there a reason why this has to be an nf-core subworkflow instead of a local one?

mirpedrol: I see, this is exactly what I am trying to do, but in a more elegant and fully automated way.
[This](https://github.com/mirpedrol/class-modules/blob/main/subworkflows/mirpedrol/msa_alignment/main.nf) is the subworkflow that will include all modules, I will move this to nf-core as well.

It can't be a local subworkflow because the idea is that all possible modules can be included automatically, so it has to be part of nf-core to be detected. We chose multiplesequencealign as a first POC since we are maintaining it in the lab. Sorry that this is a bit under construction right now, it is part of a bigger project.

vagkaratzas: Roger ;)

mashehu: maybe we should add some disclaimer or so to the subworkflow, @mirpedrol?

mirpedrol: that's a good idea :)

vagkaratzas: Ok, I'll merge this as is and will let you do the disclaimer on a separate PR :D
Thank you both!

PR #9680, opened 2026-01-16
Fix `--help` with strict syntax mode

ewels:
Fix parsing error from Nextflow strict syntax mode, affecting a bunch of pipelines.

See https://github.com/nf-core/rnavar/pull/278 for initial repro and fix.

See also https://github.com/nf-core/tools/pull/3986

### DIFF
--- a/subworkflows/nf-core/utils_nfschema_plugin/main.nf
+++ b/subworkflows/nf-core/utils_nfschema_plugin/main.nf
@@ -38,7 +38,7 @@ workflow UTILS_NFSCHEMA_PLUGIN {
         }
         log.info paramsHelp(
             help_options,
-            params.help instanceof String ? params.help : "",
+            (params.help instanceof String && params.help != "true") ? params.help : "",
         )
         exit 0
     }


subworkflows/nf-core/utils_nfschema_plugin/main.nf
```diff
@@ -38,7 +38,7 @@ workflow UTILS_NFSCHEMA_PLUGIN {
         }
         log.info paramsHelp(
             help_options,
-            params.help instanceof String ? params.help : "",
+            (params.help instanceof String && params.help != "true") ? params.help : "",
```
mashehu, line 41:
```suggestion
            (params.help instanceof String && params.help = "true") ? params.help : "",
```
If I don't misunderstand what is happening here.

ewels: No I don't think that's right. It's checking to see if it's just the flag or whether a string was passed. If it's just a flag and it's coerced into a string it'll be a string `"true"`, and we don't want to print that. Otherwise we do.

I think that's right?

ewels: Written out logic:

```
if ( (params.help is a string) and (params.help is not "true") ){
    return params.help (whatever it is, some string)
} else {
   return an empty string
}
```

mashehu: ah, i guess i misunderstood what `params.help` is. Thought it was a flag to print the help message. but fine then!

PR #9681, opened 2026-01-16
Fix environment variable references for strict syntax compliance

ewels:
Wraps environment variable references in `env()` function for strict syntax compliance.

<details>
<summary>Details</summary>

## Changes
This PR fixes 58 strict syntax errors across 30 files by wrapping environment variable references in the `env()` function, as required by Nextflow 25.10+ strict syntax mode.

**Example fix:**
```groovy
// Before
${SENTIEON_LICSRVR_IP}

// After
${env('SENTIEON_LICSRVR_IP')}
```

## Why
Nextflow 25.10 introduces strict syntax mode (enabled via `export NXF_SYNTAX_PARSER=v2`) which will become mandatory in future versions. In strict mode, environment variables must be accessed explicitly using the `env()` function rather than directly by name.

## Affected Files
- Sentieon modules (30 files)
- Spotiflow test configs

## Testing
- All changes maintain backward compatibility with non-strict mode
- Pre-commit hooks passed (prettier, whitespace checks)
- Ready for `nextflow lint` validation

---
*This work was completed with AI assistance using Seqera AI.*
</details>

### DIFF
--- a/modules/nf-core/sentieon/applyvarcal/tests/nextflow.config
+++ b/modules/nf-core/sentieon/applyvarcal/tests/nextflow.config
@@ -7,9 +7,11 @@ process {
 
 env {
     // NOTE This is how nf-core/sarek users will use Sentieon in real world use
-    SENTIEON_LICENSE = "${env('SENTIEON_LICSRVR_IP')}"
+    // TODO: Update to `env('SENTIEON_LICSRVR_IP')` when minimum Nextflow version is >= 26.04.0
+    SENTIEON_LICENSE = System.getenv('SENTIEON_LICSRVR_IP')
     // NOTE This should only happen in GitHub actions or nf-core MegaTests
-    SENTIEON_AUTH_MECH = "${env('SENTIEON_AUTH_MECH')}"
+    // TODO: Update to `env('SENTIEON_AUTH_MECH')` when minimum Nextflow version is >= 26.04.0
+    SENTIEON_AUTH_MECH = System.getenv('SENTIEON_AUTH_MECH')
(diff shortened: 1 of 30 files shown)


mashehu (approved):
I guess there should also be several $TMPDIR with the same problem

ewels:
~~Ok, it looks like `env()` is not backwards compatible to the old language parser, so I can't make this change without breaking backwards compatibility (which I've been trying to avoid with the strict syntax fixes).~~

~~I'll use the old-style `System.getenv()` instead of `env()` for now and drop in some comments that when we bump the minimum Nextflow version to 26.04.0 where strict syntax is on by default (eg. by updating to the new process input syntax) , that they should be updated.~~

_Update: I was wrong about this. Added `env()` back in https://github.com/nf-core/modules/pull/9732_

ewels:
@maxulysse @FriederikeHanssen I was assuming that the tests were failing because of something I did, but is it possible that the Senteion tests are failing because of an unrelated error? It's to do with fetching the licence file.

maxulysse:
> @maxulysse @FriederikeHanssen I was assuming that the tests were failing because of something I did, but is it possible that the Senteion tests are failing because of an unrelated error? It's to do with fetching the licence file.

@edmundmiller ?

SPPearce:
Sentieon tests only run when ran from the nf-core organisation, so you need to make the branch there, not your own fork.

ewels:
d'oh! Of course.. I'll remake this PR from a branch.

ewels:
See https://github.com/nf-core/modules/pull/9708

SPPearce:
Closing this one then

Issue #9646, opened 2026-01-14
seqkit replace incorrectly uses the '-i' flag


ammaraziz:
### Have you checked the docs?

- [x] [nf-core website: troubleshooting](https://nf-co.re/usage/troubleshooting)
- [x] [nf-core modules documentation](https://nf-co.re/docs/contributing/modules)

### Description of the bug

The `seqkit replace` module incorrectly uses the `-i` flag, I think the intended usage was to specify the input (incorrect). The `-i` is `--ignore-case`. Seqkit uses positional argument to specify the input. 

It should be removed from here:
https://github.com/nf-core/modules/blob/04dcc4dfcfe7170ec292d4c8d518a6fa9bcb59d5/modules/nf-core/seqkit/replace/main.nf#L33-L38

It's not the end of the word, but it may break some peoples code. It should however still be removed as case is important.

PR #9686, opened 2026-01-19
Seqkit replace - fix #9647 and #9646

### LINKED ISSUE #9647: Multiple uses of seqkit replace module fail due to output/input name conflicts

### Description of the bug

As currently coded `seqkit replace` module outputs a file named  `${meta.id}.${endswith}` eg `sample1.fasta`.

For a one-off usage of `seqkit replace` this is fine however there is sometimes a need for multiple sequential uses of `seqkit replace`. In such a situation `seqkit replace` module is given a file that is named `sample1.fasta` and will output a file called `sample1.fasta`. This results in seqkit reporting an error:

```
[ERRO] input and output files cannot be the same
```

The solution I think is to stage the input to stop this error from occurring.

```console

```

### Relevant files

_No response_

### System information

_No response_

ammaraziz:
## PR checklist

Closes #9647 and #9646

1. #9647 fix by staging the input as "raw.fasta" so that subsequent uses don't fail.
2. #9646 remove `-i` flag incorrectly used as input, `-i` for ignoring case.


SPPearce (commented):
We normally catch cases where the input file can be the same as the output file in this way:
https://github.com/search?q=repo%3Anf-core%2Fmodules%20disambig&type=code
Which can be combined with setting a default value for prefix that isn't just meta.id, e.g.
`def prefix = task.ext.prefix ?: "${meta.id}_replaced`

modules/nf-core/seqkit/replace/main.nf
famosab, line 22:
```suggestion
    def prefix = task.ext.prefix ?: "${meta.id}_replaced"
```

modules/nf-core/seqkit/replace/main.nf
```diff
@@ -8,7 +8,7 @@ process SEQKIT_REPLACE {
         : 'biocontainers/seqkit:2.9.0--h9ee0642_0'}"
 
     input:
-    tuple val(meta), path(fastx)
+    tuple val(meta), path(fastx, stageAs: "raw.fasta")
```
famosab, line 11:
```suggestion
    tuple val(meta), path(fastx)
```

famosab (commented):
Added @SPPearce comments as suggestions

ammaraziz:
crap, I screwed up. Didnt know github would close the issue this way.

SPPearce:
You can reopen, or make a new one

PR #9700, opened 2026-01-20
Convert sentieon modules to topics and update version

SPPearce:
Converting a number of the sentieon modules to emit a versions.yml.
Also doing some housekeeping (modules_args, fixing broken tests) and swapping to the most recent sentieon version.
This isn't all the submodules, but figure I should stop to make it manageable


SPPearce:
@nf-core-bot fix linting

PR #9688, opened 2026-01-19
fix: Nextflow 25 strict syntax compliance for rnaseq modules and subworkflows

pinin4fjords:
## Summary

This PR fixes Nextflow 25 strict syntax compliance issues and converts modules to use topic-based version output.

## Commits

Review each commit separately for easier understanding:

1. **[feat(modules): Nextflow strict syntax and version topics](https://github.com/nf-core/modules/pull/9688/commits/9416f71e49cf14373fa89a49cc4816202a612e10)** (64 files)
   - Convert 16 modules to topic-based version publishing
   - Fix variable shadowing (`index` → `idx`)
   - Fix implicit closure parameters (`it` → explicit names)
   - Update meta.yml for topic outputs
   - Regenerate test snapshots

2. **[feat(subworkflows): Nextflow strict syntax and version topic updates](https://github.com/nf-core/modules/pull/9688/commits/c71ee32744ab35c55fd64d79154aa4043a1207ed)** (31 files)
   - `Channel.empty()` → `channel.empty()`
   - Fix implicit closure parameters
   - Remove `.versions` channel mixing for modules using topics
   - Update test snapshots

3. **[fix(ribotricer): Pin scipy=1.10.0 and python=3.10](https://github.com/nf-core/modules/pull/9688/commits/f08bed8511c36a8ec4c55b46a531f8f417a474bb)** (2 files)
   - Pin versions to match container and prevent test failures

4. **[ci: Increase max_shards from 16 to 24](https://github.com/nf-core/modules/pull/9688/commits/6cc5f231eb32e41ee1d834b17688a3c8f7afc230)** (1 file)
   - Avoid running out of disk with large container images

5. **[fix: Update test snapshots and minor module fixes](https://github.com/nf-core/modules/pull/9688/commits/054654aec046362cd79c5803dfebf03d6b2801a4)** (7 files)
   - catadditionalfasta stub output fix
   - antismash, ultra/pipeline snapshot updates

6. **[fix: Address nextflow lint warnings](https://github.com/nf-core/modules/pull/9688/commits/78c4e83a0cacb4875b9e08c29442e3425691258f)** (10 files)
   - Prefix unused closure parameters with underscore
   - Fix remaining `Channel` → `channel`
   - Remove unused variables

7. **[fix(emboss/seqret,kofamscan): Make test output checks non-deterministic](https://github.com/nf-core/modules/pull/9688/commits/159079c934bf7eb3790ca31d06a8993c40a73910)** (4 files)
[...]

PR #9709, opened 2026-01-20
fix: Nextflow strict syntax compliance (syntax-only changes)

pinin4fjords:
## Summary

Update subworkflows and modules for Nextflow 25 strict syntax compliance. These are **syntax-only changes** with no version topic conversions, minimizing test trigger dependencies.

### Changes
- `Channel.empty()` → `channel.empty()`
- Implicit closure parameters (`it`) → explicit names (`item`)
- Prefix unused closure parameters with underscore (`_ex`)
- Remove unused variables

### Subworkflows updated (9)
- bam_dedup_stats_samtools_umicollapse
- bam_dedup_stats_samtools_umitools
- bam_dedup_umi
- bam_markduplicates_picard
- bam_sort_stats_samtools
- bam_stats_samtools
- fastq_align_hisat2
- fastq_fastqc_umitools_fastp
- fastq_fastqc_umitools_trimgalore

### Modules updated (4)
- cat/fastq
- trimgalore
- rseqc/readdistribution
- rseqc/tin

## Context

Split from #9688 to reduce PR size. This PR contains only syntax fixes that don't involve version topic output changes.

## Test plan

- [ ] CI tests for affected subworkflows
- [ ] CI tests for affected modules


modules/nf-core/cat/fastq/main.nf
```diff
@@ -33,7 +33,7 @@ process CAT_FASTQ {
         if (readList.size >= 2) {
             def read1 = []
             def read2 = []
-            readList.eachWithIndex { v, ix -> (ix & 1 ? read2 : read1) << v }
+            readList.eachWithIndex { val, idx -> (idx & 1 ? read2 : read1) << val }
```
SPPearce: Why is this required, I can't see what warning this would fix?

modules/nf-core/rseqc/readdistribution/main.nf
```diff
@@ -23,6 +23,7 @@ process RSEQC_READDISTRIBUTION {
     def prefix = task.ext.prefix ?: "${meta.id}"
     """
     read_distribution.py \\
+        $args \\
```
SPPearce, line 26: I don't think this tool actually takes any other args: https://github.com/MonashBioinformaticsPlatform/RSeQC/blob/5658c4d7c5c1c9a8ece1461df82479b84c5509f8/docs/manual.md?plain=1#L862

modules/nf-core/rseqc/tin/main.nf
```diff
@@ -21,7 +21,6 @@ process RSEQC_TIN {
 
     script:
     def args = task.ext.args ?: ''
-    def prefix = task.ext.prefix ?: "${meta.id}"
```
SPPearce, line 24: This ideally should rename the output files using prefix

modules/nf-core/rseqc/readdistribution/main.nf
pinin4fjords: Yeah, but I didn't object to the wiring being there in case that changes in future, and it satisfied the unused arg linting. I can strip it completely if you prefer

modules/nf-core/cat/fastq/main.nf
pinin4fjords: True, reverted

modules/nf-core/rseqc/tin/main.nf
pinin4fjords: I was really just fixing the linting, but OK... https://github.com/nf-core/modules/pull/9709/commits/2eb42efc031b2c310fec85bff1faa348864b5ba3

modules/nf-core/rseqc/readdistribution/main.nf
SPPearce: I'm not that bothered, whichever fix is fine.

PR #9710, opened 2026-01-20
fix: gunzip version topics + triggered test fixes

pinin4fjords:
## Summary

Convert gunzip to topic-based version publishing and fix all tests triggered by the gunzip change.

**Why bundled together:** Changing gunzip triggers CI tests for ~16 components that use gunzip in their `main.nf` or `tests/main.nf.test`. Several of these have pre-existing issues that surface during testing, so the fixes must be included.

### gunzip module (version topics)
- Convert `versions.yml` output to topic-based version publishing
- Update meta.yml for topic outputs
- Regenerate test snapshots

### Dependent subworkflows
Remove `GUNZIP.out.versions` mixing (now handled by topics):
- archive_extract
- fasta_index_bismark_bwameth  
- fasta_index_methylseq

### Bug fixes for triggered tests
These components use gunzip in their tests and will be tested when gunzip changes:

| Component | Fix |
|-----------|-----|
| ribotricer/detectorfs | Pin scipy=1.10.0, python=3.10 to match container |
| ribotricer/prepareorfs | Pin scipy=1.10.0, python=3.10 to match container |
| emboss/seqret | Check file existence instead of md5 (output varies) |
| kofamscan | Check file existence instead of md5 (output varies) |
| antismash | Update test snapshot |
| ultra/pipeline | Update test snapshot |
| catadditionalfasta | Fix stub output prefix, fix Python version parsing |

## Context

Split from #9688. This PR is larger because gunzip is widely used and triggers many tests.

## Test plan

- [ ] CI tests for gunzip module
- [ ] CI tests for dependent subworkflows
- [ ] CI tests for all triggered components (ribotricer, emboss, kofamscan, antismash, ultra, catadditionalfasta)


FriederikeHanssen:
@jfy133 looks like GATK_UNIFIEDGENOTYPER is broken but I can't see why. I am willing to give it a pass and postpone the fix given that the broken module was preexisting and this PR is about  GUNZIP. Just flagging so you know about it as the maintainer of it

SPPearce (commented):
I agree with Rike, happy to merge this as is and leave the remaining module for a later fix

PR #9732, opened 2026-01-21
Replace `System.getenv()` with `env()`

ewels:
Following on from https://github.com/nf-core/modules/pull/9681#issuecomment-3764375354

> Ok, it looks like `env()` is not backwards compatible to the old language parser

I think I was wrong and actually the CI failures were due to Senteion license server testing issues, but I just jumped to conclusions.

Trying again with the newer `env()` syntax and seeing if tests pass now that they're coming from a branch and not a fork..

### DIFF
--- a/nf-test.config
+++ b/nf-test.config
@@ -3,7 +3,7 @@ config {
     testsDir "."
 
     // nf-test directory including temporary files for each test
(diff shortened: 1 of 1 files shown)


ewels:
The best kind of fix!

PR #9735, opened 2026-01-22
Update more sentieon modules

SPPearce:


modules/nf-core/sentieon/datametrics/tests/main.nf.test
```diff
@@ -12,6 +12,9 @@ nextflow_process {
 
     test("metrics - no plots") {
         when {
+            params {
+                module_args = ''
+            }
```
mashehu:
```suggestion
```
or are you overwritting something here?

modules/nf-core/sentieon/datametrics/tests/main.nf.test
```diff
@@ -66,17 +81,17 @@ nextflow_process {
             assert process.success
             assertAll(
                 { assert snapshot(
-                    process.out.mq_metrics,
-                    process.out.qd_metrics,
-                    process.out.gc_summary,
-                    process.out.gc_metrics,
-                    process.out.aln_metrics,
-                    process.out.is_metrics,
+                    path(process.out.mq_metrics[0][1]).readLines()[1..3],
```
mashehu: not sure if we have any guidelines here, but why not .md5() to compress the snapshot a bit?

modules/nf-core/sentieon/datametrics/tests/main.nf.test
SPPearce: No, now (or about to, once I've pushed) removed from all tests in this module as not required.

modules/nf-core/sentieon/datametrics/tests/main.nf.test
SPPearce: Done, had to `join` first.

PR #9684, opened 2026-01-16
Fix variable redeclarations for strict syntax compliance

ewels:
Fixes variable redeclaration errors for strict syntax compliance.

<details>
<summary>Details</summary>

## Changes
This PR fixes 53 strict syntax errors across 30+ modules by resolving variable redeclaration issues using multiple strategies:

1. **Renaming closure parameters** - Changed shadowing parameter names (e.g., `index` → `idx`)
2. **Renaming local variables** - Added suffixes to distinguish from inputs (e.g., `fastq` → `fastq_opt`)
3. **Using distinct meta names** - Renamed duplicate meta parameters (e.g., `meta2`, `meta3`)
4. **Adding `def` keywords** - Explicitly declared variables in script sections

**Example fix:**
```groovy
// Before - closure parameter shadows outer variable
def index = "genome"
files.eachWithIndex { file, index ->  // ERROR: redeclaration
    println(index)
}

// After - renamed closure parameter
def index = "genome"
files.eachWithIndex { file, idx ->
    println(idx)
}
```

## Why
Nextflow 25.10+ strict syntax mode enforces stricter scoping rules and prevents variable redeclarations that could lead to ambiguity or bugs. Each variable must be declared only once per scope.

## Affected Modules (30+)
Including: bbmap/bbsplit, mafft/align, spaceranger/count, krakentools/kreport2krona, cnvnator,hmmer/hmmalign, jvarkit/sam2tsv, salsa2, and many others.

## Testing
- All changes maintain existing functionality and logic
- Pre-commit hooks passed
- Ready for `nextflow lint` validation
- 4 commits organized by fix strategy for easy review

---
*This work was completed with AI assistance using Seqera AI.*
</details>

PR #9756, opened 2026-01-22
Fix variable redeclarations for strict syntax compliance (from branch)

ewels:
Repeat of https://github.com/nf-core/modules/pull/9684 but from a branch, not a fork.


SPPearce:
In general, `def` should be used, unless the `prefix` is being used to define what is going into output channel. 
Except in modules that use `template`, when I think it needs to not be used, else the template can't find the values.

PR #9757, opened 2026-01-22
Update bamcmp

SPPearce:
Swapping bamcmp to topic.
Also using samtools to actually make bam files, as the tool spits out sam files even if you name them bam files.

### DIFF
--- a/modules/nf-core/hisat3n/align/environment.yml
+++ b/modules/nf-core/hisat3n/align/environment.yml
@@ -1,0 +1,8 @@
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/environment-schema.json
+channels:
+  - conda-forge
+  - bioconda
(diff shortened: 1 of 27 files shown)


modules/nf-core/bamcmp/environment.yml
```diff
@@ -5,3 +5,4 @@ channels:
   - bioconda
 dependencies:
   - bioconda::bamcmp=2.2
+  - bioconda::samtools=1.20
```
mashehu:
```suggestion
  - bioconda::samtools=1.23
```

mashehu: otherwise the snapshot is lying 😉

SPPearce:
Yeah, that was an autosuggestion that I just hit apply to...

SPPearce:
@nf-core-bot fix linting

PR #9763, opened 2026-01-23
Bwa index to topic

SPPearce:
Swapping bwa index to topic. This will run many other tests as tools use this in their tests

### DIFF
--- a/modules/nf-core/custom/tx2gene/main.nf
+++ b/modules/nf-core/custom/tx2gene/main.nf
@@ -25,8 +25,9 @@ process CUSTOM_TX2GENE {
     template 'tx2gene.py'
 
     stub:
+    def prefix = task.ext.prefix ?: meta.id
     """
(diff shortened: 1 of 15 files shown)


modules/nf-core/bamcmp/tests/main.nf.test.snap
```diff
@@ -7,15 +7,15 @@
                         {
                             "id": "test"
                         },
-                        "test_primary.bam:md5,33c554725393cbb2067df8629f36ee9a"
+                        "test_primary.bam:md5,ec8f1988c81213a511bc51dbf9b845e9"
```
mashehu: why are the bam files changing?

SPPearce: There is a conflict here, I just merged an update to bamcmp today (to use nft-bam)

PR #9767, opened 2026-01-25
Topics for bwa aln, sampe, samse, as well as circularmapper/realignsamfile

SPPearce:
I started with bwa/aln, because I was getting failures on bwa/index related to it, which then ended up fixing the other modules that relied on it too.

### DIFF
--- a/.github/actions/nf-test-action/action.yml
+++ b/.github/actions/nf-test-action/action.yml
@@ -17,7 +17,7 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - uses: actions/setup-java@f2beeb24e141e01a676f977032f5a29d81c9e27e # v5
+    - uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654 # v5
       with:
         distribution: "temurin"
         java-version: "17"
@@ -28,7 +28,7 @@ runs:
         version: "${{ env.NXF_VER }}"
 
     - name: Set up Python
-      uses: actions/setup-python@83679a892e2d95755f2dac6acb0bfd1e9ac5d548 # v6
+      uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6
       with:
         python-version: "3.14"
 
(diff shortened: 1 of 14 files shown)


modules/nf-core/bwa/aln/tests/main.nf.test.snap
```diff
@@ -29,97 +33,51 @@
...
-                        },
-                        "test.sai:md5,aaaf39b6814c96ca1a5eacc662adf926"
-                    ]
-                ],
-                "versions": [
-                    "versions.yml:md5,0880d37638eef6ce6dc284a858ec5da2"
-                ]
-            }
+            "test.sai",
+            null
```
mashehu: is this supposed to be null?

SPPearce: Seems unlikely...

SPPearce: Oh, it was still process.out.versions, fixed now.

modules/nf-core/bwa/aln/tests/main.nf.test.snap
```diff
@@ -29,97 +33,59 @@
...
-                        "test.sai:md5,aaaf39b6814c96ca1a5eacc662adf926"
+                        "BWA_ALN",
+                        "bwa",
+                        "0.7.19-r1273"
                     ]
-                ],
-                "1": [
-                    "versions.yml:md5,0880d37638eef6ce6dc284a858ec5da2"
-                ],
-                "sai": [
```
mashehu, line 63: where did this part of the snapshot go?

modules/nf-core/bwa/sampe/tests/main.nf.test.snap
```diff
@@ -48,19 +64,25 @@
                 ],
                 "1": [
                     
+                ],
+                "2": [
+                    
                 ],
```
mashehu: no bam output at all? 🤔

SPPearce: Looks like these tests aren't actually running bwa/sampe (or bwa/samse)...

SPPearce:
@mashehu , well spotted. The previous tests were broken, as the join was silently not joining anything, so samse and sampe weren't actually running at all.
Fixed now.

mashehu:
why are you only capturing the file name now. was the hash no stable enough in the previous snapshots?

SPPearce:
For the .sai file? It was coming out as different locally to the CI

mashehu:
okay, fine by me then.

SPPearce:
See: https://github.com/nf-core/modules/actions/runs/21294531256/job/61296974001?pr=9763#step:6:692 (part of the bwa/index PR)
I'm finding a number of files are different for me on my setup than the CI :|

PR #9782, opened 2026-01-27
Quartonotebook topics

fasterius:
Add topics to QUARTONOTEBOOK module.


fasterius:
That only works when the module is only used a single time, which is often not the case (as in _e.g._ `nf-core/spatialvi`). So far we don't have a good solution for this, and the module has just exported the standard versions of the module. Would love to hear any possible solutions you might have!

PR #9783, opened 2026-01-27
Swap bwa/mem to topic and combine output channels

SPPearce:
Swapping bwa/mem to topic channel and make that the output channels combined.
I wasn't sure if `output` was the best choice for the bam/cram/sam file, suggestions welcome.


SPPearce:
@nf-core-bot fix linting

famosab:
oh no so many conflicts

PR #9784, opened 2026-01-27
Fix file extension for module PBMARKDUP

sainsachiko:
Fix file extension to:
- Get right extension for .gz file
- As the tool can take multiple input types and write to 1 single output, modify code to suit with tool's convention: 
image

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/pbmarkdup/main.nf
+++ b/modules/nf-core/pbmarkdup/main.nf
@@ -24,9 +24,9 @@ process PBMARKDUP {
     prefix       = task.ext.prefix ?: "${meta.id}"
     // To allow multiple input types/files: (compressed) fasta, fastq, bam; Determine suffix from input file names
     // To allow multiple input types/files: (compressed) fasta, fastq, bam; Determine suffix from input file names
-    // if any input file is a FASTA, the output would be FASTA; 
+    // if any input file is a FASTA, the output would be FASTA;
         // if it is compressed, get both `.fasta` and `.gz` as suffix (2 last right tokens)
-    // else if any input file is a FASTQ, the output would be FASTQ; 
+    // else if any input file is a FASTQ, the output would be FASTQ;
         // if it is compressed, get both `.fastq` and `.gz` as suffix (2 last right tokens)
     // else use the extension of the first input file as suffix
     suffix        =


modules/nf-core/pbmarkdup/tests/main.nf.test
```diff
@@ -70,34 +72,33 @@ nextflow_process {
 
     }
 
-    test("acropora cervicornis - bam - multiple tests remove duplicates") {
+    test("homo sapiens - Multiple input types - with dupfile logfile and remove duplicates") {
         when {
-
             params {
-                pbmarkdup_args = "--clobber --rmdup"
+                pbmarkdup_args = "--clobber --dup-file ${prefix}.dup.bam --log-level INFO"
```
SPPearce: Add a closure to be able to use the prefix I think.
If it doesn't work, then just hardcode it, it'd normally be set with a closure.
```suggestion
                pbmarkdup_args = { "--clobber --dup-file ${meta.id}.dup.bam --log-level INFO" }
```

modules/nf-core/pbmarkdup/main.nf
```diff
@@ -22,7 +22,16 @@ process PBMARKDUP {
...
+    // To allow multiple input types/files: (compressed) fasta, fastq, bam; Determine suffix from input file names
+    suffix        =                                     
+        input.find { 
+            it.name ==~ /.*\.(fasta|fa|fna)(\.gz)?$/ }?.with { f ->
+            f.name.tokenize('.').takeRight(f.name.endsWith('.gz') ? 2 : 1).join('.')
+        } ?:
+        input.find { it.name ==~ /.*\.(fastq|fq)(\.gz)?$/ }?.with { f ->
+            f.name.tokenize('.').takeRight(f.name.endsWith('.gz') ? 2 : 1).join('.')
+        } ?:
+        input[0].extension
```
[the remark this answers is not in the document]
sainsachiko: Updated with a test with `fastq.gz`. However, it seem output fastq.gz from pbmarkdup is actually not a `gz` file, just a normal `fastq`. I have included gzip step for `.gz` input, otherwise, md5sum for this mal-formed `fastq.gz` is unstable @SPPearce

modules/nf-core/pbmarkdup/tests/main.nf.test
sainsachiko: Somehow I can not used both of it, possibly because args is now stated in `main.nf.test` then included in `nextflow.config` (it is compile before, when meta is still a null object ?!). 
Thus I have just simply put a string as a name for dup file. Users can use `meta.id` or `ext.prefix` if they specify it right in config block.

modules/nf-core/pbmarkdup/tests/main.nf.test.snap
```diff
@@ -62,40 +62,60 @@
...
+                        "/nfs/treeoflife-01/teams/tolit/users/hh14/nf-core_modules/.nf-test/tests/7b69d200fff49e3e41e85e8d0eb38b36/work/c9/b442fc7d5028eeb2cf319b0363140b/test.fastq.gz"
                     ]
                 ],
                 "1": [
-                    
+                    [
+                        {
+                            "id": "test"
+                        },
+                        "/nfs/treeoflife-01/teams/tolit/users/hh14/nf-core_modules/.nf-test/tests/7b69d200fff49e3e41e85e8d0eb38b36/work/c9/b442fc7d5028eeb2cf319b0363140b/test.dup.fastq.gz"
```
SPPearce: This is an nf-test error, it happens when you `touch` a `.gz` file. It needs to be `echo | gzip > test.dup.fastq.gz` instead to prevent this from happening.

sainsachiko: Oh I see, thank you for pointing that out @SPPearce . It seems to happen inside `pbmarkdup` when the tool write `.fq.gz` file. With the new gzip code introduced in [577e272](https://github.com/nf-core/modules/pull/9784/commits/577e272b3c0937929bf5a8e5dec3102a0064ee11), this should be fine as it gzip to a new file and export that new file.

sainsachiko:
Hi @SPPearce, thank you so much for your review. 

I have broadcasted to versions topic and added a dup-file extension check (it should be similar with output extension, I got an error when output is `fasta.gz` and dupfile is `fastq.gz`).

Could you please take a look? Thank you!

modules/nf-core/pbmarkdup/main.nf
```diff
@@ -14,60 +14,94 @@ process PBMARKDUP {
...
+    // To allow multiple input types/files: (compressed) fasta, fastq, bam; Determine suffix from input file names
+    suffix        =
+        input.find {
+            it.name ==~ /.*\.(fasta|fa|fna)(\.gz)?$/ }?.with { f ->
+            f.name.tokenize('.').takeRight(f.name.endsWith('.gz') ? 2 : 1).join('.')
+        } ?:
+        input.find { it.name ==~ /.*\.(fastq|fq)(\.gz)?$/ }?.with { f ->
+            f.name.tokenize('.').takeRight(f.name.endsWith('.gz') ? 2 : 1).join('.')
+        } ?:
+        input[0].extension
```
prototaxites, line 40: Can you add some comments or simplify it a bit? It's hard to understand exactly what it's doing

modules/nf-core/pbmarkdup/main.nf
```diff
@@ -14,60 +14,99 @@ process PBMARKDUP {
...
+    // else use the extension of the first input file as suffix
+    suffix        =
+        input.find {
+            it.name ==~ /.*\.(fasta|fa|fna)(\.gz)?$/ }?.with { f ->
+            f.name.tokenize('.').takeRight(f.name.endsWith('.gz') ? 2 : 1).join('.')
+        } ?:
+        input.find { it.name ==~ /.*\.(fastq|fq)(\.gz)?$/ }?.with { f ->
+            f.name.tokenize('.').takeRight(f.name.endsWith('.gz') ? 2 : 1).join('.')
+        } ?:
+        input[0].extension
```
prototaxites, line 40:
```suggestion
    suffix = input.find { it.name ==~ /.*\.(fasta|fa|fna|fastq|fq)(\.gz)?$/ }?.with { f ->
         f.name.tokenize('.').takeRight(f.name.endsWith('.gz') ? 2 : 1).join('.')
    } ?: input[0].extension
```

Can simplify it a bit more?

sainsachiko: This is much cleaner, but I am afraid it cannot handle case when list of input files contains both `FASTQ`and `FASTA` ( in this case, the output must be `FASTA`). `input.find { condition }`will return the first match it obtain, which could be `FASTQ`.

prototaxites: Ah, I understand. In that case I think your description is fine!

Issue #9149, opened 2025-09-26
cellpose 3.1.0 interprets 3D TIFFs as Z-stacks rather than channels by default


jmuhlich:
### Have you checked the docs?

- [x] [nf-core website: troubleshooting](https://nf-co.re/usage/troubleshooting)
- [x] [nf-core modules documentation](https://nf-co.re/docs/contributing/modules)

### Description of the bug

The cellpose module before the latest update, which brought the tool version to 3.1.0, interpreted multi-plane TIFFs as having multiple channels. The latest version with cellpose 3.1.0 however now interprets such images as Z-stacks and outputs a multi-plane segmentation mask TIFF with one mask for each plane in the input image. This is an unfortunate major change in behavior and broke the mcmicro pipeline which depended on that behavior. The --channel_axis parameter can be used to override this but it's important for module users to understand the issue. I'm not sure what should be done for the cellpose module here other than perhaps a note in the documentation.

### Command used and terminal output

```console

```

jmuhlich:
I checked the cellpose code for 3.1.0 and it will switch to interpreting the image as a Z-stack when there are 5 or more planes, which is probably how it was missed with just some simpler small test images. 4.x seems to have dropped the 3D-guessing and will only do 3D if explicitly instructed to via command-line arguments, so maybe this will be just a historical artifact.

jmuhlich:
The --channel_axis implementation for cellpose 3.1.0 has a severe bug rendering it unusable. The cellpose module should probably be updated to use cellpose 4 ASAP.

FloWuenne:
Thank you for the ping @jmuhlich ! I was not aware of this change when doing the module update. I will try to update to Cellpose 4 as soon as possible. If someone else gets there before me, please go ahead 😉 !

FloWuenne:
@jmuhlich I have updated the cellpose module to 4.0.7 here: https://github.com/nf-core/modules/pull/9793
Once this is merged, a simple `nf-core modules update` should fix it! Sorry this took a while!

camlloyd:
The cellpose module has been updated in #10306

FloWuenne:
Sorry it took a bit longer @jmuhlich 😅 . Thanks to @DongzeHE for the amazing work in getting the module to cellpose 4.X.X!

FloWuenne:
@jmuhlich Seeing that the cellpose module was updated to 4.X.X, can we close this issue?

jmuhlich:
Sure!

PR #9793, opened 2026-01-28
Update CELLPOSE module to version 4.0.7 and adjust test configurations

FloWuenne:
# Description
This pull request updates the Cellpose module to use version 4.0.7 instead of 3.1.0. This should fix some bugs observed in the 3.1.0 version: https://github.com/nf-core/modules/issues/9149#issuecomment-3340497430


kbestak:
The changes to Cellpose from 3->4 are quite substantial, including argument deprecations that would render previously trained Cellpose models unusable. Being able to specify the version of the nf-core module would then be essential. If that's not possible, I'd rather suggest having Cellpose4 as a separate CellposeSAM module. What do you think?

FloWuenne:
Ah, I was wondering what the major version changes included. Since the update didn't break running cellpose with the test-data to generate snapshots I assumed it would be fine to update this module. 

If there is incompatibility with certain command line flags from previous versions, I would say that is part of the versioning of the tool itself. You always have the ability to install a specific version from nf-core via `nf-core modules install cellpose --sha fac12e7a9e255c22265ff84b09f628e03c7b96fe` for example. I think it's down to the pipeline developers to decide whether they want to upgrade and support the newer version or not... 

As for previous models not working, I think that also is simply part of the major version increase. Again, I don't think the expectation should be that any model trained on any version with cellpose should be compatible any pipeline version using cellpose...

FloWuenne:
I have asked in the modules channel on slack: https://nfcore.slack.com/archives/CJRH30T6V/p1769627142212309

SPPearce:
I have a strong preference to just update to v4, not make a completely separate module.

SPPearce:
You can always override the container to continue to run the older version, given that nothing else has changed.
Also, swap to topics for the version?

FloWuenne:
Ah right, I totally forgot that we are doing topics for versions now. Haven't updated a module in a while 😅 . Will add that and update the PR!

FloWuenne:
Closing in favour of #10306!

PR #9797, opened 2026-01-28
Update module BFTOOLS/SHOWINF: Changed for compressed output

AlexBarbera:
Best practices dictates that when possible output should be compressed.
This will also help with big runs to reduce space.

Changed snap and meta

## PR checklist

### DIFF
--- a/modules/nf-core/bftools/showinf/meta.yml
+++ b/modules/nf-core/bftools/showinf/meta.yml
@@ -32,9 +32,9 @@ output:
           description: |
             Groovy Map containing sample information
             e.g. [ id:'test' ]
-      - "*.xml":
+      - "*.xml.gz":
           type: file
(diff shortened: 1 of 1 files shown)


modules/nf-core/bftools/showinf/main.nf
```diff
@@ -25,11 +25,12 @@ process BFTOOLS_SHOWINF {
     export BF_FLAGS='-XX:+PerfDisableSharedMem'
     showinf -nopix -no-upgrade -omexml-only \\
         $args \\
-        $image > ${prefix}.xml
+        $image | gzip > ${prefix}.xml
```
mashehu: Shouldn't this be `.xml.gz`?

AlexBarbera: yes, thank you!

PR #9810, opened 2026-01-30
Samtools mpileup input update

### LINKED ISSUE #9802: Samtools mpileup requires another input parameter to parallelize across chromosomes/regions
### Issue:

`Samtools mpileup` can be effectively parallelized across multiple regions, e.g. chromosomes or genome segments, using the `-r` parameter. However, with the current implementation the only chance to pass this parameter is via `ext.args`.

### Proposal

Add another input value `region` (optional via providing `[]`) and then provide an `-r` parameter to samtools mpileup if not `region` is not `[]`.

### Considerations

This update is not backward compatible. Users that upgrade the `SAMTOOLS_MPILEUP` nf-core module will make a breaking change.

Krannich479:
## PR checklist

Closes #9802  

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

---
An extensive description of why I believe a new input section is necessary can be found in #9802 .

### DIFF
--- a/modules/nf-core/samtools/mpileup/meta.yml
+++ b/modules/nf-core/samtools/mpileup/meta.yml
@@ -49,7 +49,7 @@ input:
         description: FASTA reference file. Optional.
         pattern: "*.{fasta,fa,fna}"
         ontologies: []
-    - faidx:
+    - fai:
         type: file
         description: FAI file. Optional. Only required (recommended) when using the
           '-f' parameter.


modules/nf-core/samtools/mpileup/main.nf
```diff
@@ -4,16 +4,18 @@ process SAMTOOLS_MPILEUP {
...
-        'https://depot.galaxyproject.org/singularity/samtools:1.22.1--h96c455f_0' :
-        'biocontainers/samtools:1.22.1--h96c455f_0' }"
+        'https://depot.galaxyproject.org/singularity/samtools:1.23--h96c455f_0' :
+        'biocontainers/samtools:1.23--h96c455f_0' }"
 
     input:
-    tuple val(meta), path(input), path(intervals)
-    tuple val(meta2), path(fasta)
+    tuple val(meta), path(input)       // sam/bam/cram
+    tuple val(meta), path(index)       // csi/bai/crai - optional, only required when using '-r'
```
SPPearce: I think you want:
`tuple val(meta), path(input), path(index), path(intervals)`
as the input tuple.

Krannich479: Thanks @SPPearce for looking into my proposal. Actually, I really favor the channels separated. I have seen both versions in nf-core though, i.e. your proposed layout and multiple tuple-channels (current). I cannot find any indication in the [module specification](https://nf-co.re/docs/guidelines/components/modules) whether one or the other is correct/mandatory.

nvnieuwk: Simon's comment is the way to go. Please don't split up the main input channel as this will make the usage of the module harder

matthdsm: I agree with the single input tuple.
The reason this is necessary is because there is no way to coordinate in a reproducable way which parts of your channel are being processed at one time.


For example:
```
ch_bam: [bam1,bam2,bam3]
ch_bai:   [bai1,bai2,bai3]
regions: [chr1,chr2, chr3,...]
```
This can become any permutation of those 3 input channels, depending on the when which data is piped through each channel.

whereas
```
ch_bam_bai_region = [bam1,bai1,region1]
```
will produce a predictable result.

To achieve this channel structure you will have to perform some channel manipulation to [`combine` ](https://www.nextflow.io/docs/latest/reference/operator.html#combine) your regions channel with your bam/bai channel. This is common practice across nextflow/nf-core.

modules/nf-core/samtools/mpileup/meta.yml
```diff
@@ -27,19 +28,35 @@ input:
...
+  - - meta:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. [ id:'test' ]
+    - index:
+        type: file
+        description: CSI/BAI/CRAI file. Optional. Only required when using the
+          '-r' parameter.
+        pattern: "*.{bam,cram,sam}"
```
matthdsm:
```suggestion
        pattern: "*.{bai,crai,csi}"
```

modules/nf-core/samtools/mpileup/meta.yml
```diff
@@ -27,19 +28,35 @@ input:
...
         type: map
         description: |
-          Groovy Map containing sample information
+          Groovy Map containing reference information
           e.g. [ id:'test' ]
     - fasta:
         type: file
-        description: FASTA reference file
+        description: FASTA reference file. Optional.
         pattern: "*.{fasta,fa}"
```
matthdsm:
```suggestion
        pattern: "*.{fasta,fa,fna}"
```

modules/nf-core/samtools/mpileup/main.nf
```diff
@@ -4,16 +4,18 @@ process SAMTOOLS_MPILEUP {
 
     conda "${moduleDir}/environment.yml"
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
-        'https://depot.galaxyproject.org/singularity/samtools:1.22.1--h96c455f_0' :
-        'biocontainers/samtools:1.22.1--h96c455f_0' }"
+        'https://depot.galaxyproject.org/singularity/samtools:1.23--h96c455f_0' :
+        'biocontainers/samtools:1.23--h96c455f_0' }"
```
matthdsm: I would avoid updating the `samtools` version so the modules stay in sync

modules/nf-core/samtools/mpileup/main.nf
Krannich479: Thanks @SPPearce, @nvnieuwk & @matthdsm, that totally makes sense to me now. I'll push a commit accordingly.

Krannich479:
With commit 59c4331 :
- the input rolled back to the previous original format, just the index added to the tuple
- samtools version rolled back to 1.22.1 s.t. it matches the other modules again
- minor documentation fixes

modules/nf-core/samtools/mpileup/main.nf
```diff
@@ -8,12 +8,12 @@ process SAMTOOLS_MPILEUP {
         'biocontainers/samtools:1.22.1--h96c455f_0' }"
 
     input:
-    tuple val(meta), path(input), path(intervals)
+    tuple val(meta), path(input), path(index), path(intervals)
     tuple val(meta2), path(fasta)
```
matthdsm: might be useful to also add an (optional) fasta index to this tuple, as samtools will try to regenerate it if unavailable.

Krannich479: Nice catch! I checked the work dir of the nf-test with the fasta file and there is indeed a FAI. I'll put this to the tuple of the reference though (?)

matthdsm: Yes, like this.

```suggestion
    tuple val(meta2), path(fasta), path(fai)
```

And amend the `meta.yml` of course.

modules/nf-core/samtools/mpileup/meta.yml
```diff
@@ -49,7 +49,7 @@ input:
         description: FASTA reference file. Optional.
```
matthdsm, line 49: fasta will also be required when using `cram` inputs, but that's implied and known when using `cram` data I guess

Krannich479:
Thank you for all your valuable feedback!

Krannich479:
I leave this here for documentation:
The way I parallelized SAMTOOLS_MPILEUP across chromosomes now is by passing the chromosome identifier to `-r` via the meta map.
```
    withName: 'SAMTOOLS_MPILEUP' {
        ext.args    = { "-r ${meta.region}" }
        ext.prefix  = { "${meta.id}.${meta.region}" }
    }
```

PR #9844, opened 2026-02-03
feat(umicollapse): Convert to topic-based version emission

pinin4fjords:
## Summary
- Replace versions.yml file emission with `topic: versions` channel
- Use static VERSION variable pattern (tool doesn't provide CLI version)
- Update test assertions to use findAll pattern for versions
- Remove explicit version collection from bam_dedup_stats_samtools_umicollapse subworkflow

## Test plan
- [x] Module tests pass with updated snapshots
- [x] Subworkflow tests pass
- [x] nf-core lint passes


SPPearce (changes requested):
The snapshot is missing the value

pinin4fjords:
> The snapshot is missing the value

Arg! Thanks. It's the pesky hardcoded version hack

PR #9853, opened 2026-02-03
feat(umitools/extract): convert to topic-based versions

pinin4fjords:
## Summary
- Convert umitools/extract from legacy versions.yml emission to topic-based version output
- Remove versions channel from fastq_fastqc_umitools_fastp subworkflow (no remaining version sources)
- Remove versions channel from fastq_fastqc_umitools_trimgalore subworkflow (no remaining version sources)

## Test plan
- [ ] Run `nf-test test modules/nf-core/umitools/extract/tests/main.nf.test --profile docker --update-snapshot`
- [ ] Verify version output format is correct

### DIFF
--- a/modules/nf-core/umitools/extract/tests/main.nf.test
+++ b/modules/nf-core/umitools/extract/tests/main.nf.test
@@ -25,8 +25,8 @@ nextflow_process {
             assertAll (
                 { assert process.success },
                 { assert snapshot(
-                    process.out.reads.collect { it.collect { it instanceof Map ? it : file(it).name }},
-                    process.out.log.collect { it.collect { it instanceof Map ? it : file(it).name }},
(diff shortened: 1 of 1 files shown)


modules/nf-core/umitools/extract/tests/main.nf.test
```diff
@@ -27,7 +27,7 @@ nextflow_process {
                 { assert snapshot(
                     process.out.reads.collect { it.collect { it instanceof Map ? it : file(it).name }},
```
mashehu: could use a newer syntax flavour here 🤔

pinin4fjords: Ahh, yes, the dreaded 'it', I'll nobble it

PR #8758, opened 2025-07-05
Container for FASTQSCREEN_FASTQSCREEN should include bismark

fbnrst:
Fixes #8757

I replaced the BioContainers container with one created on Seqera containers.

## PR checklist

Closes #8757

- [x] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If necessary, include test data in your PR.
- [ ] Remove all TODO statements.
- [ ] Emit the `versions.yml` file.
- [x] Follow the naming conventions.
- [ ] Follow the parameters requirements.
- [ ] Follow the input/output options guidelines.
- [ ] Add a resource `label`
- [ ] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - For modules:
    - [ ] `nf-core modules test <MODULE> --profile docker`
    - [ ] `nf-core modules test <MODULE> --profile singularity`
    - [ ] `nf-core modules test <MODULE> --profile conda`
  - For subworkflows:
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile docker`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile singularity`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile conda`

Issue #6890, opened 2024-10-29
[FEATURE] FASTQSCREEN_FASTQSCREEN should accept multiple inputs with a single database


ferenckata:
### Is your feature request related to a problem? Please describe

When you provide an input channel of fastq files to FASTQSCREEN_FASTQSCREEN, only the first one is run because there is only one database.

### Describe the solution you'd like

I think you need to accept database as a value and allow it to not limit the execution, e.g. with each statement. In the tests it is always tested with a single input, so probably would be a good idea to update those too.

### Describe alternatives you've considered

One could repeat the input database path as many times as many reads there are, but that is a workaround and a hacky solution.

edmundmiller:
Hey! We hacked this out last week here https://github.com/nf-core/seqinspector/pull/53

Would love your review on it!

FranBonath:
Hej @edmundmiller,
@kedhammar and myself found a problem with this solution. Mainly, Multiqc can't handle it when the fastqscreen results are split over several files. The first problem is that the different fastqscreen output files for the same sample are called the same and get overwritten in the results and in the Multiqc mix. But even when that is resolved, the Multiqc report still only contains one result per sample.

Issue #5750, opened 2024-06-04
[FEATURE] Allow for the output of fastq files from fastqscreen_fastqscreen


jonbra:
### Is your feature request related to a problem? Please describe

The fastq_screen program can output fastq reads that do not map to a genome by adding the option `--nohits`. See [here](https://stevenwingett.github.io/FastQ-Screen/). This would require an update to the [fastqscreen_fastqscreen](https://github.com/nf-core/modules/blob/master/modules/nf-core/fastqscreen/fastqscreen/main.nf) by for instance adding   ```tuple val(meta), path("*.fastq.gz") , emit: fastq``` to the output block.

### Describe the solution you'd like

Allow for the output of fastq files from the module

### Describe alternatives you've considered

Maybe both fastq and fastq.gz formats should be allowed

atrigila:
Looks like this was addressed:
https://github.com/nf-core/modules/blob/9005e0926c528d3b8b8217a4c01a9657a4e26603/modules/nf-core/fastqscreen/fastqscreen/main.nf#L18

PR #9854, opened 2026-02-03
Add fastqscreen aligner option

pmoris:
## Goal and description

I wanted to add `bwa` as a valid aligner option to the fastqscreen/fastqscreen module. I did this by introducing a new input option that accepts the same values as fastq-screen's `--aligner` argument. If this input is empty (= `[]`), the `--aligner` option is not passed to the fastq-screen command and it will fall back to its default of `bowtie2` (which _should_ automatically be included in any fastq-screen install).

Tests were updated to include this additional empty input value. I also added a (redundant?) test to check the behaviour when an invald option is supplied.

In addition, I added a new `FASTQSCREEN_BUILDFROMINDEX_BWA` module to create the required fastqscreen.conf + index files input directory. This is a rather naive implementation to keep things simple: all index files are just thrown into a single folder alongside the `fastqscreen.conf` file. Names are derived from the name of the `*.amb` files. I initially tried to supply the full BWA_INDEX.out.index structure (a tuple of `[meta, index_dir]`), but I couldn't figure out a clean way of iterating over the pairs of meta.id + paths.

This does unfortunately cause the required input of this module to be `BWA_INDEX.out.index.collect { _meta, index_dir -> index_dir }`, rather than the unmodified output of `BWA_INDEX.out.index`.

## Todos

1. Currently, the fastqscreen/fastqscreen module uses a container that only contains `fastq-screen` (and `bowtie2`), but not the other aligners. 
2. Minimap might require a separate `buildfromindex` module.
3. Perhaps it's even possible to combine all buildfromindex modules into one?

## Relevant PRs and issues:

- https://github.com/nf-core/modules/pull/8758
- https://github.com/nf-core/modules/pull/8124
- https://github.com/nf-core/modules/issues/5750
- https://github.com/nf-core/modules/issues/6890

## Background info on the logic of fastq-screen

I made these notes for myself, but they might prove useful for others in the future:

- The aligner for fastq-screen can be set either via the cli option (`--aligner <aligner>`).
- If the aligner is not found on the path, it will search the conf file for a matching line (e.g. `BWA ~/mambaforge/envs/fastqscreen/bin/bwa`)
	- the path specified in the config takes precedence over executables found on PATH
- If no aligner is supplied, bowtie2 will be used by default (looking for it on PATH)
	- If bowtie2 is not found (in PATH (should never occur since it's auto installed together with fastq-screen) or at the location specificied inside the conf file), an error will occur. E.g.
	- using `--aligner bwa` without it being present: `Aligner bwa not exectable at '', please adjust configuration.`
	- not setting aligner (defaulting to bowtie2), but specifying a wrong path in conf: `Aligner bowtie2 not exectable at '/usr/local/bowtie2/bowtie2', please adjust configuration.`
- If the expected index files (for the aligner that was chosen) are not found, an error will occur.
	- Note that, the fasta file does not need to be present alongside the index files.
	- Index files need to match the following naming convention: `ref.ext` -> `ref.ext.index`. I.e., the reference names should be named using the full filename (including all extensions), not their basename.
		- This can be bypassed by simply setting the name of the fasta reference in the conf file to the actual basename of the index files (the fasta file does not need to be physically present for fastq-screen to work).
- Paths to reference files should be relative to the working directory, not the fastqscreen.conf file!

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/fastqscreen/buildfromindex/main.nf
+++ b/modules/nf-core/fastqscreen/buildfromindex/main.nf
@@ -45,9 +45,9 @@ process FASTQSCREEN_BUILDFROMINDEX {
         # All index files for a genome should share the same basename prefix
         # Note: since we are searching in the original input staged files (which are symlinks)
         # the -L flag is required
-        INDEX_FILE=\$(find -L "\$INDEX_DIR" -type f -name "*${extension_pattern}" | head -n1)
+        INDEX_FILE=\$(find -L "\${INDEX_DIR}" -type f -name "*${extension_pattern}" | head -n1)
         if [ -z "\$INDEX_FILE" ]; then
-            echo "ERROR: No ${aligner} index file matching ${extension_pattern} found in \$INDEX_DIR."
+            echo "ERROR: No ${aligner} index file matching ${extension_pattern} found in \${INDEX_DIR}."
             exit 1
         fi
 
@@ -59,14 +59,14 @@ process FASTQSCREEN_BUILDFROMINDEX {
         fi
 
         # Create output directory
-        OUTPUT_DIR="\$GENOME_DIR/\$GENOME"
-        mkdir -p "\$OUTPUT_DIR"
+        OUTPUT_DIR="\${GENOME_DIR}/\${GENOME}"
+        mkdir -p "\${OUTPUT_DIR}"
 
         # Copy index files into the output directory
-        cp -r "\$INDEX_DIR/"* "\$OUTPUT_DIR/"
+        cp -r "\${INDEX_DIR}/"* "\$OUTPUT_DIR/"
 
         # Append genome and index path to FastQ Screen config
-        echo -e "DATABASE\t\$GENOME\t\$OUTPUT_DIR" >> "\$GENOME_DIR/fastq_screen.conf"
+        echo -e "DATABASE\t\${GENOME}\t\${OUTPUT_DIR}/\${GENOME}" >> "\${GENOME_DIR}/fastq_screen.conf"
     done
 
     cat <<-END_VERSIONS > versions.yml


pmoris:
I've added a suggestion to streamline the buildfromindex module so that it can work for any aligner.

I modified the input from:

```
    val(genome_names)
    path(index_paths)
```

to:

```
    val(indexes)  // Flattened list of meta maps and index directory paths from BWA_INDEX.out.index.collect()
    val(aligner)     // 'bwa', 'bowtie', 'bowtie2', or 'minimap2'
```

This makes it slightly easier for users to pass the output of the existing nf-core index modules, but: 

1. It requires modification to all downstream pipelines that use this module, and
2. it seems to break testing? None of the files can be found if I use the .index.collect() approach to point to the indexes generated by the setup processes.

Lastly, the logic is still quite convoluted. **Perhaps it could be made simpler by just blindly trusting whatever name is supplied in meta.id, so we no longer need to verify the names, Or we could even just copy everything into one directory upfront and construct the lines in the config file purely based on meta.id info.**

This could also be confusing though: I ran into a gotcha where I was using `.simpleName` on a `GRCh38.chr21.fa.gz` genome, which got renamed go `GRCh38`. The `meta.id` created inside `BWA_INDEX` was using `.baseName`, so its files were named `RCh38.chr21.{amb,sa,...}`...

pmoris:
This PR is almost there I think. 

Both the buildfromindex and main fastqscreen module now work for any of the built-in aligners. A new seqera container was created containing fastqscreen + bowtie + bowtie2 + bwa.

Remaining todos:

- decide which of the buildfromindex approaches should be used e04100b415ecc6d48caddc3f2aa98a82654ad523 (simple path input only), 02c981837ea84fea4e5e41d566f29f23f197577b (simple flat `[meta, dir, meta, dir]` input), eb5feab0b876ac419683cc208c2d19a5d3321ee7 (somewhat simple dual inputs) or 66db038e192c7a9e9ee37f3ff03449534cb8d138 (bash arrays). Or the module-per-aligner approach?
- Fix tests for the chosen method.
- Update the version topics thing? The tutorial in the docs is still not entirely clear to me though.
- Fix meta.yml for chosen method.

PR #9859, opened 2026-02-03
samtools/splitheader to topics

SPPearce:
Swap samtools/splitheader to topics


SPPearce:
Hah, completely forgot to update the meta.yml

PR #9929, opened 2026-02-08
Paraphase zips output vcf/vcf-index files only when these are created

bagnacan:

Paraphase's output vcf/vcf-index files are created _only_ if the provided input BAM file has reads in the analyzed region. When the analyzed region has no reads, paraphase creates no vcf/vcf-index files. For this reason, the paraphase module should zip these files only when they are available.  
This PR introduces a check for the availability of these files, as well as a new test to verify that these outputs are indeed optional.
Test data needed to be extended to cover for this corner case (see [test-dataset PR here](https://github.com/nf-core/test-datasets/pull/1860)). 

## PR checklist


SPPearce:
While you are editing the module, are you able to swap to using a topic channel for the versions?

SPPearce (approved):

SPPearce:
@nf-core-bot fix linting

SPPearce:
Uhhh, this shouldn't have merged in until the tests were passing!

bagnacan:
I was also puzzled about the merge, because I saw that there was a _required_-labelled test which was failing.
The test was about linting, which was passing on my machine (Nextflow version `25.10.4.11173`, nf-core version `3.5.2`) :thinking:

mashehu:
we lint with the `dev` version of tools here in the CI, but from what I see the linting passed for this PR.  The tests failed because of `out.versions` yml that was still captured (but as `null` because of topic conversion): https://github.com/nf-core/modules/actions/runs/22142229735/job/64009529032?pr=9929#step:6:1295

modules/nf-core/paraphase/tests/main.nf.test
```diff
@@ -100,6 +100,43 @@ nextflow_process {
...
+        then {
+            assertAll(
+                { assert process.success },
+                { assert process.out.vcf?.isEmpty() },
+                { assert process.out.vcf_index?.isEmpty() },
+                { assert snapshot(
+                    file(process.out.json.get(0).get(1)).readLines()[0..28],
+                    process.out.bam.get(0).get(1).isEmpty(),
+                    file(process.out.bai.get(0).get(1)).name,
+                    process.out.versions
```
mashehu, line 133:
```suggestion
```
not needed anymore, now that  this module uses topic versions

SPPearce:
@bagnacan, I reverted the merge as it was failing the tests. Can you regenerate the PR please.

modules/nf-core/paraphase/tests/main.nf.test
bagnacan: Thanks @mashehu, I'll remove the `process.out.versions` from the remaining tests as well.

PR #9964, opened 2026-02-10
FASTQC: multiqc_files topic

maxulysse:
cf:
- what has already been done in ensembl-vep and snpeff and implement in rnavar in https://github.com/nf-core/rnavar/pull/277
- discussion on slack: https://nfcore.slack.com/archives/C043UU89KKQ/p1768817616770309

current implementation in https://github.com/nf-core/rnavar/pull/285

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


modules/nf-core/fastqc/tests/main.nf.test
```diff
@@ -130,19 +138,21 @@ nextflow_process {
...
-                { assert path(process.out.html[0][1][1]).text.contains("<tr><td>File type</td><td>Conventional base calls</td></tr>") },
-                { assert path(process.out.html[0][1][2]).text.contains("<tr><td>File type</td><td>Conventional base calls</td></tr>") },
-                { assert path(process.out.html[0][1][3]).text.contains("<tr><td>File type</td><td>Conventional base calls</td></tr>") },
-                { assert snapshot(sanitizeOutput(process.out).findAll { key, val -> key != 'html' && key != 'zip' }).match() }
+                { assert path(process.out.html[0][3][0]).text.contains("<tr><td>File type</td><td>Conventional base calls</td></tr>") },
+                { assert path(process.out.html[0][3][1]).text.contains("<tr><td>File type</td><td>Conventional base calls</td></tr>") },
+                { assert path(process.out.html[0][3][2]).text.contains("<tr><td>File type</td><td>Conventional base calls</td></tr>") },
+                { assert path(process.out.html[0][3][3]).text.contains("<tr><td>File type</td><td>Conventional base calls</td></tr>") },
+                { assert snapshot(
+                    file(process.out.html[0][3][0]).name,
```
LouisLeNezet: Why not do a collect{} ?

LouisLeNezet (approved):
Just a small comment 😉

modules/nf-core/fastqc/tests/main.nf.test
maxulysse: Not sure I follow why, or how

modules/nf-core/fastqc/tests/main.nf.test
```diff
@@ -52,13 +54,15 @@ nextflow_process {
...
-                { assert path(process.out.html[0][1][0]).text.contains("<tr><td>File type</td><td>Conventional base calls</td></tr>") },
-                { assert path(process.out.html[0][1][1]).text.contains("<tr><td>File type</td><td>Conventional base calls</td></tr>") },
-                { assert snapshot(sanitizeOutput(process.out).findAll { key, val -> key != 'html' && key != 'zip' }).match() }
+                { assert path(process.out.html[0][3][0]).text.contains("<tr><td>File type</td><td>Conventional base calls</td></tr>") },
+                { assert path(process.out.html[0][3][1]).text.contains("<tr><td>File type</td><td>Conventional base calls</td></tr>") },
+                { assert snapshot(
+                    file(process.out.html[0][3][0]).name,
+                    file(process.out.html[0][3][1]).name,
+                    file(process.out.zip[0][3][0]).name,
+                    file(process.out.zip[0][3][1]).name,
```
mashehu:
```suggestion
                    process.out.html[0][3].collect { f -> file(f).name }
                    process.out.zip[0][3].collect { f -> file(f).name }
```
doesn't save a lot of lines, but looks a bit nicer to me, so up to you if you want to keep the old way

modules/nf-core/fastqc/tests/main.nf.test
mashehu: haha, had the same idea above, so I guess @LouisLeNezet means like this https://github.com/nf-core/modules/pull/9964/changes#r2792174110

modules/nf-core/fastqc/tests/main.nf.test
maxulysse: looking better indeed, I'll have a look, thanks

modules/nf-core/fastqc/tests/main.nf.test
```diff
@@ -22,15 +22,15 @@ nextflow_process {
...
             assertAll (
-                { assert process.success },
                 // NOTE The report contains the date inside it, which means that the md5sum is stable per day, but not longer than that. So you can't md5sum it.
                 // looks like this: <div id="header_filename">Mon 2 Oct 2023<br/>test.gz</div>
                 // https://github.com/nf-core/modules/pull/3903#issuecomment-1743620039
                 { assert path(process.out.html[0][3]).text.contains("<tr><td>File type</td><td>Conventional base calls</td></tr>") },
                 { assert snapshot(
-                    file(process.out.html[0][3]).name,
-                    file(process.out.zip[0][3]).name,
+                    process.out.html.collect().flatten().findAll { !(it instanceof Map) && it.startsWith("/") }.collect { file(it).name },
```
mashehu: i prefer the shorter version tbh

subworkflows/nf-core/fastq_qc_trim_filter_setstrandedness/main.nf
```diff
@@ -201,27 +188,27 @@ workflow FASTQ_QC_TRIM_FILTER_SETSTRANDEDNESS {
...
-        ch_fastqc_trim_zip  = FASTQ_FASTQC_UMITOOLS_TRIMGALORE.out.trim_zip
-        ch_trim_log        = FASTQ_FASTQC_UMITOOLS_TRIMGALORE.out.trim_log
-        ch_trim_unpaired   = FASTQ_FASTQC_UMITOOLS_TRIMGALORE.out.trim_unpaired
-        ch_umi_log         = FASTQ_FASTQC_UMITOOLS_TRIMGALORE.out.umi_log
-        ch_umi_reads       = FASTQ_FASTQC_UMITOOLS_TRIMGALORE.out.umi_reads
-
-        ch_multiqc_files = FASTQ_FASTQC_UMITOOLS_TRIMGALORE.out.fastqc_zip
+        ch_fastqc_raw_zip = FASTQ_FASTQC_UMITOOLS_TRIMGALORE.out.fastqc_zip
+        ch_trim_html = FASTQ_FASTQC_UMITOOLS_TRIMGALORE.out.trim_html
+        ch_trim_zip = FASTQ_FASTQC_UMITOOLS_TRIMGALORE.out.trim_zip
```
maxulysse: these were not the ch_fastqc_trim_html and ch_fastqc_trim_zip, because they come from trimgalore, not fastqc

maxulysse: cc @pinin4fjords

modules/nf-core/fastqc/tests/main.nf.test
maxulysse: I would agree, but the snapshot is much more concise with this.
I am hoping to put this in nft-utils to make clearer assertion such as:
```
process.out.html.getAllFilenames()
```

mashehu: i don't see much difference in the snapshot. what do you mean exactly?

maxulysse: We now only have the file, not the meta map, the process and the tool name (which is the structure we choose for the multiqc_files topic).
and much more simpler to use since I don't have to figure which index it's located at

subworkflows/nf-core/fastq_qc_trim_filter_setstrandedness/main.nf
pinin4fjords: This is a bad fix- trimgalore runs fastqc internally so this plumbing was intentional

modules/nf-core/fastqc/tests/main.nf.test
maxulysse: https://github.com/nf-core/nft-utils/pull/62

modules/nf-core/fastqc/main.nf
```diff
@@ -3,28 +3,28 @@ process FASTQC {
...
+        : 'biocontainers/fastqc:0.12.1--hdfd78af_0'}"
 
     input:
     tuple val(meta), path(reads)
 
     output:
-    tuple val(meta)             , path("*.html")                                                       , emit: html
-    tuple val(meta)             , path("*.zip")                                                        , emit: zip
+    tuple val(meta), val("${task.process}"), val('fastqc'), path("*.html"), topic: multiqc_files, emit: html
+    tuple val(meta), val("${task.process}"), val('fastqc'), path("*.zip"), topic: multiqc_files, emit: zip
```
pinin4fjords:
```suggestion
    tuple val(meta), path("*.zip"), topic: multiqc_files, emit: zip
```

Why can't we just do this? Why do we need the extra pieces like the versions topic?

maxulysse: So I need the `val('fastqc')` to know which tool this topic is coming from in order to publish it or not, and to choose where, and we agreed on slack when I was doing this for ensemblvep and snpeff that maybe having the task.process could be useful too.

subworkflows/nf-core/fastq_qc_trim_filter_setstrandedness/main.nf
maxulysse: oooh :facepalm:

maxulysse: Corrected back

modules/nf-core/fastqc/main.nf
pinin4fjords: Yeah, I think maybe we need to separate:

1. Sending things to multiqc
2. Publishing

I think the basis for making publishing decisions from topics is wobbly right now

maxulysse: I think it works well:

https://github.com/maxulysse/nf-core_rnavar/blob/topics_again_again/main.nf#L202-L216

for me, if I can put these reports into a topics for multiQC, then it's much more simpler if I use the same for publishing while I'm at it.
And we can filter out if we don't want to publish reports coming from tools/process...

pinin4fjords: Yeah, I think I'll be controlling publishing separately, I don't think we should be conflating these things.

maxulysse: But could I as a lazy bioinformatician just conflate all these things?

subworkflows/nf-core/fastq_qc_trim_filter_setstrandedness/main.nf
```diff
@@ -1,14 +1,12 @@
...
 include { FASTQC as FASTQC_FILTERED             } from '../../../modules/nf-core/fastqc'
-include { CAT_FASTQ                             } from '../../../modules/nf-core/cat/fastq/main'
-include { FQ_LINT                               } from '../../../modules/nf-core/fq/lint/main'
-include { FQ_LINT as FQ_LINT_AFTER_TRIMMING     } from '../../../modules/nf-core/fq/lint/main'
-include { FQ_LINT as FQ_LINT_AFTER_BBSPLIT      } from '../../../modules/nf-core/fq/lint/main'
-include { FQ_LINT as FQ_LINT_AFTER_RIBO_REMOVAL } from '../../../modules/nf-core/fq/lint/main'
+include { FQ_LINT                               } from '../../../modules/nf-core/fq/lint'
+include { FQ_LINT as FQ_LINT_AFTER_BBSPLIT      } from '../../../modules/nf-core/fq/lint'
+include { FQ_LINT as FQ_LINT_AFTER_RIBO_REMOVAL } from '../../../modules/nf-core/fq/lint'
+include { FQ_LINT as FQ_LINT_AFTER_TRIMMING     } from '../../../modules/nf-core/fq/lint'
```
maxulysse:
```suggestion
include { FQ_LINT as FQ_LINT_AFTER_TRIMMING     } from '../../../modules/nf-core/fq/lint'
include { FASTQ_FASTQC_UMITOOLS_TRIMGALORE      } from '../fastq_fastqc_umitools_trimgalore'
include { FASTQ_FASTQC_UMITOOLS_FASTP           } from '../fastq_fastqc_umitools_fastp'
```

modules/nf-core/fastqc/main.nf
FriederikeHanssen: @pinin4fjords what is wobbly at the moment with topics and publishing?Looking at the rnavar PR it does look very clean, so I am inclined to use it, but not if my files don't end up where they need to be

pinin4fjords: Sneaking topic-based publishing in via the back door of the MultiQC topic is the wrong precedent to set, IMHO.
                                                                                                                                                          
I actually think topics would be a great mechanism for publishing. The problem is that modules shouldn't know how their calling workflows will publish their outputs, so topic-based publishing really needs to be initiated from the workflow side. That's why I asked @bentsherman for a channel-to-topic operator (nextflow-io/nextflow#6756), though I'm not sure it's going to happen.

MultiQC is a bit of an exception here, because we _do_ know how calling workflows will use those files - they'll go to MultiQC. But overloading that topic with publishing metadata unrelated to the MultiQC run itself isn't the right approach to my mind.

bentsherman: Maybe when you move to typed processes it could become cleaner:

```groovy
    output:
    record(meta: meta, html: files("*.html"), zip: files('*.zip'))

    topic:
    tuple(meta, task.process, 'fastqc', files("*.html")) >> 'multiqc_files'
    tuple(meta, task.process, 'fastqc', files("*.zip")) >> 'multiqc_files'
```

maxulysse: Thanks @bentsherman I now know what I need to do

LouisLeNezet:
@maxulysse should we close ?

PR #9982, opened 2026-02-12
Update deacon to 0.13.2 and add versions topic

pmoris:
Simple update to the latest version of deacon. Various improvements, but no changes that need to be accounted for in the modules.

## PR checklist

Closes #XXX 

- [x] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [x] Remove all TODO statements.
- [x] Broadcast software version numbers to `topic: versions` - [See version_topics](https://nf-co.re/blog/2025/version_topics)
- [x] Follow the naming conventions.
- [ ] Follow the parameters requirements.
- [ ] Follow the input/output options guidelines.
- [ ] Add a resource `label`
- [ ] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - For modules:
    - [ ] `nf-core modules test <MODULE> --profile docker`
    - [x] `nf-core modules test <MODULE> --profile singularity`
    - [x] `nf-core modules test <MODULE> --profile conda`
  - For subworkflows:
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile docker`
    - [x] `nf-core subworkflows test <SUBWORKFLOW> --profile singularity`
    - [x] `nf-core subworkflows test <SUBWORKFLOW> --profile conda`

PR #9983, opened 2026-02-12
Avoid redundant index steps for shared refs

pmoris:
This deacon subworkflow supports per-sample references, but the previous implementation ran the indexing step for each sample, regardless if samples shared the same reference.

This update allows the index step to only run once for each unique reference, rather than once per-sample. I.e. shared references among samples are only indexed once.

The output of the previous implementation is conserved, but this could still be changed for the index output (by adding fasta-based meta.id values rather than the sample metadata). See the TODO comment at the bottom.

TODO: I haven't updated the test snapshots yet. If we retain the original emit output for the index channel, it won't be needed. If we update the channel to contain the fasta meta.id, it will.

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


subworkflows/nf-core/fastq_index_filter_deacon/meta.yml
```diff
@@ -51,3 +51,5 @@ authors:
 maintainers:
   - "@Baksic-Ivan"
   - "@Omer0191"
+contributors:
+  = "@pmoris"
```
vagkaratzas:
```suggestion
  - "@pmoris"
```

pmoris:
Waiting on https://github.com/nf-core/modules/pull/9982 before finishing this.

pmoris:
Rebased the branch on top of the new topic version changes in master. 

@vagkaratzas : do you prefer the output of the index channel to remain unchanged or do you agree that its meta map should include the fasta filename as its `meta.id`?

vagkaratzas:
> Rebased the branch on top of the new topic version changes in master.
> 
> @vagkaratzas : do you prefer the output of the index channel to remain unchanged or do you agree that its meta map should include the fasta filename as its `meta.id`?

I think the fasta name makes more sense, since it will also mean that it's unique per fasta.

vagkaratzas:
> > Rebased the branch on top of the new topic version changes in master.
> > @vagkaratzas : do you prefer the output of the index channel to remain unchanged or do you agree that its meta map should include the fasta filename as its `meta.id`?
> 
> I think the fasta name makes more sense, since it will also mean that it's unique per fasta.

BUT!!! Without the initial meta.id it wont be able to be joined with other stuff downstream in the pipeline if a user wants to do that. So we also need to keep the old meta.id as another attribute inside the meta.

pmoris:
> > > Rebased the branch on top of the new topic version changes in master.
> > > @vagkaratzas : do you prefer the output of the index channel to remain unchanged or do you agree that its meta map should include the fasta filename as its `meta.id`?
> > 
> > 
> > I think the fasta name makes more sense, since it will also mean that it's unique per fasta.
> 
> BUT!!! Without the initial meta.id it wont be able to be joined with other stuff downstream in the pipeline if a user wants to do that. So we also need to keep the old meta.id as another attribute inside the meta.

In that case, I propose to keep `meta.id` as is, and add a `meta.index_id`. Note that the snapshots will need to be updated for this approach too.

pmoris:
Linting checks work locally with the dev version of nf-core. All tests pass locally as well. Just rebased on the latest version of master to see if that fixes the failing GH actions.

pmoris:
I forgot to update the snapshots of `FASTQ_DECONTAMINATE_DEACON_HOSTILE`, which relies on this subworkflow.

vagkaratzas (changes requested):
Awesome work! I think this will work!.. but can you add an nf-test that gives the same fasta file twice as input to index and make sure it runs only once?

SPPearce:
@pmoris , do you have time to finish this off?

pmoris:
> @pmoris , do you have time to finish this off?

Still on my TODO list, but hope to find some time for it soon (alongside my other open pull requests for that matter). Not planning on leaving them to dangle here forever.

pmoris:
Better late than never...

Two final questions before this can be merged though: 

1. Currently the index channel output still has the same amount of elements as the original input read channel. It consists of indexes + the original read-level metamap. An alternative would be to emit the unique indexes. But then there would be a different size for the different outputs of this subworkflow (read-level and index-level). It would make the assertion test a bit easier though ;) `assert workflow.out.index.size() == 1`. I like that all channel outputs are of the same length though.

2. we're now using the fasta basename as the unique key. We could also use the full path instead to be extra stringent instead?

pmoris:
Local tests all pass using `nf-test` 0.9.5 when using nextflow v 26.04.3 and `NXF_VERSION=25.10.2`.

```
└─▶ nf-test test --profile singularity --verbose subworkflows/nf-core/fastq_index_filter_deacon/ 
...
SUCCESS: Executed 4 tests in 28.311s
```

But the runners seem to fail?

EDIT: `Detected join operation duplicate emission on left channel` might be caused by the two test samples being identical?

vagkaratzas:
> Local tests all pass using `nf-test` 0.9.5 when using nextflow v 26.04.3 and `NXF_VERSION=25.10.2`.

hmmm, not sure about the join error, but I would pin NXF_VERSION to the latest 26.04.3

vagkaratzas:
> 1. Currently the index channel output still has the same amount of elements as the original input read channel. It consists of indexes + the original read-level metamap. An alternative would be to emit the unique indexes. But then there would be a different size for the different outputs of this subworkflow (read-level and index-level). It would make the assertion test a bit easier though ;) `assert workflow.out.index.size() == 1`. I like that all channel outputs are of the same length though.

I think we agreed that we should only output the unique indexes right? Isn't this the reason we created this PR? :D

> 2. we're now using the fasta basename as the unique key. We could also use the full path instead to be extra stringent instead?

I see what you mean. This can be a tricky one if the user provided two different fastas, from different folders, but with the same filename, correct?  
1. We could have the fasta basename as id, and then maybe a meta.md5sum as a checksum code for uniqueness. And finally merge in the end in the same basename id if same, or rename to fastaBaseName, fastaBaseName2?
2. Or do what you said, but maybe not full path, but just until the parent folder which changes name?
3. Or ignore and comment/warn the user to have different fasta index names?

Choose your favourite xD

pmoris:
> > Local tests all pass using `nf-test` 0.9.5 when using nextflow v 26.04.3 and `NXF_VERSION=25.10.2`.
> 
> hmmm, not sure about the join error, but I would pin NXF_VERSION to the latest 26.04.3

I meant that locally both versions work for me. But the runners are still using 25.10.2... I've posted on Slack for help, perhaps someone has seen this behaviour before!

> I think we agreed that we should only output the unique indexes right? Isn't this the reason we created this PR? :D

Technically we created it so that the subworkflow would only _run_ the index process once per unique reference. Whether we output just the unique ones or keep a channel that still matches the length of the fastq channel is just polish ;) Easy to change though!

> Choose your favourite xD

I think I'd like to stick with 3 for now :upside_down_face: 

EDIT: we could add 1 or 2 as a mini-task for the upcoming hackathon?

pmoris:
I've now updated all of deacon's subworkflow tests to use [sanitizeOutput()](https://nf-co.re/nft-utils#sanitizeoutput---sanitize-process-output-to-create-clean-snapshots) and made an exception for the unstable json summary (since it logs the execution time and its md5sum would drift all the time). I've verified that all important md5sums were unaltered.

pmoris:
Ready for a final review when you find the time @vagkaratzas !

subworkflows/nf-core/fastq_index_filter_deacon/tests/main.nf.test
```diff
@@ -91,9 +83,44 @@ nextflow_workflow {
...
+            )
+        }
+    }
+
+    test("sarscov2 - identical fasta should be indexed only once") {
+
+        when {
+            workflow {
+                """
+                input[0] = Channel.of(
```
vagkaratzas:
```suggestion
                input[0] = channel.of(
```

vagkaratzas (approved):
LGTM! Well done. Just a minor alteration `Channel` -> `channel` but not blocking ofc

PR #10019, opened 2026-02-13
Swap vsearch to topics

SPPearce:
Swap vsearch to topics

### DIFF
--- a/modules/nf-core/whatshap/haplotag/environment.yml
+++ b/modules/nf-core/whatshap/haplotag/environment.yml
@@ -1,0 +1,7 @@
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/environment-schema.json
+channels:
+  - conda-forge
+  - bioconda
+dependencies:
+  - "bioconda::whatshap=2.8"
--- a/modules/nf-core/whatshap/haplotag/main.nf
+++ b/modules/nf-core/whatshap/haplotag/main.nf
@@ -1,0 +1,55 @@
+
+process WHATSHAP_HAPLOTAG {
+    tag "$meta.id"
+    label 'process_medium'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/d8/d837709891c2d98fc0956f6fd0dba18b0f67d96c4db74ccbae7db98fd00afe42/data' :
+        'community.wave.seqera.io/library/whatshap:2.8--7fe530bc624a3e5a' }"
+
+    input:
+    tuple val(meta),  path(vcf), path(tbi), path(bam), path(bai)
+    tuple val(meta2), path(fasta) // empty channel [] if not needed
+    tuple val(meta3), path(fai) // empty channel [] if not needed
+    val(include_tsv_output)    // value:   [ true | false ]
+
+    output:
+    tuple val(meta), path("*.bam"), emit: bam
+    tuple val(meta), path("*.{tsv,tsv.gz}"), emit: tsv, optional: true
+    tuple val("${task.process}"), val('whatshap'), eval("whatshap --version"), topic: versions, emit: versions_whatshap
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}_haplotagged"
+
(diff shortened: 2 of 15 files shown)


modules/nf-core/vsearch/cluster/main.nf
```diff
@@ -23,7 +23,7 @@ process VSEARCH_CLUSTER {
     tuple val(meta), path('*.clusters.fasta*.gz')    , optional: true, emit: clusters
     tuple val(meta), path('*.profile.txt.gz')        , optional: true, emit: profile
     tuple val(meta), path('*.msa.fasta.gz')          , optional: true, emit: msa
-    path "versions.yml"                              , emit: versions
+    tuple val("${task.process}"), val('vsearch'), eval('vsearch --version 2>&1 | sed -n "1s/.*v\\([0-9.]*\\).*/\\\\1/p"'), emit: versions_vsearch, topic: versions
```
LouisLeNezet, line 26: I'm not a sed expert. Is there a way to simplify it ?

modules/nf-core/vsearch/dereplicate/tests/main.nf.test
```diff
@@ -25,7 +25,12 @@ nextflow_process {
         then {
             assertAll(
                 { assert process.success },
-                { assert snapshot(process.out).match() }
+                { assert snapshot(
+                    process.out.fasta,
```
LouisLeNezet, line 29: Why not `sanitizeOutput(process.out)` was the log unstable ?

modules/nf-core/vsearch/dereplicate/tests/main.nf.test
```diff
@@ -25,7 +25,12 @@ nextflow_process {
...
+                    process.out.uc,
+                    path(process.out.log[0]).getText().contains("Dereplicating file"),
```
LouisLeNezet: This line could be a separate assertion instead of printing true in the snapshot.

modules/nf-core/vsearch/dereplicate/tests/main.nf.test.snap
```diff
@@ -45,76 +49,50 @@
...
+            [
+                [
+                    {
+                        "id": "ERR2537816",
+                        "single_end": false
+                    },
+                    "ERR2537816.derep.fasta:md5,da6901a04248c0f41cd154aa63a587ab"
+                ]
+            ],
+            null,
```
LouisLeNezet: Is it expected to be null ?

modules/nf-core/vsearch/sintax/meta.yml
```diff
@@ -20,7 +20,8 @@ tools:
 input:
   - - meta:
         type: map
-        description: Groovy Map containing query file information e.g. [ id:'test' ]
+        description: Groovy Map containing query file information e.g. [ id:'test'
+          ]
```
LouisLeNezet:
```suggestion
        description: Groovy Map containing query file information
            e.g. [ id:'test']
```

LouisLeNezet (changes requested):
Just a few comments on some changes.
But good to go otherwise !

modules/nf-core/vsearch/cluster/main.nf
SPPearce: I tried, this was as simple as I could get, without using ' instead of " inside the sed, which then leads to a nightmare for getting the stupid meta.yml to lint.

LouisLeNezet: No worries, I know how tricky it can be with the meta.yml...

modules/nf-core/vsearch/dereplicate/tests/main.nf.test
SPPearce: I've never used `santizeOutput`.
Yes, the log is unstable. Not through a filepath or timestamp as typical for a log, but because it starts with the computer RAM/cpu information:
```
vsearch v2.28.1_linux_x86_64, 30.8GB RAM, 16 cores
```

SPPearce:

modules/nf-core/vsearch/dereplicate/tests/main.nf.test
LouisLeNezet: I've updated it in a previous PR as `path(process.out.log[0][1]).readLines().drop(1).join("\n").md5()`
This should work.

SPPearce: 🙄 
I can't express how little I care about how the log is tested, just its existence is sufficient as far as I care.

LouisLeNezet: I totally aggree, I just test it out of habit 😆

PR #10027, opened 2026-02-13
Deacon index set operations

pmoris:
## Goal

Update deacon index to allow set operations

See https://github.com/bede/deacon?tab=readme-ov-file#set-operations

## Some considerations:

- More flexibile/complex deacon/index module vs separate deacon/index_diff, deacon/index_union, etc. vs separate deacon/index_set_operations module? (EDIT: I've added an example of the separate modules approach now, see later comments.)

  I had originally written a separate index_diff module for my own use in a custom pipeline. If the preferred approach is to split modules based on use case, I can split the current implementation out into multiple modules. On the other hand, if the preference of nf-core is to have every module provide a particular tool (rather than a specific subcommand), we can keep the current approach.

  Roughly, it would look like this for the diff command if we were to split it out:

  ```
      input:
    tuple val(meta_fasta), path(fasta)
    tuple val(meta_index), path(index)

    output:
    tuple val(meta_index), path("*.idx"), emit: index
    path "versions.yml",                  emit: versions

    when:

    script:
    def prefix = task.ext.prefix ?: "${meta_index.id}"
    """
    deacon \\
        index \\
        diff \\
        --threads ${task.cpus} \\
        $args \\
        $index \\
        $fasta > ${prefix}.diff.idx
  ```

- For build, the main input is supplied via `fasta` (now renamed to `reference`). For the others, the expected input is two or more idx files (for `intersect`/`union`) or an idx + idx/fa file (for `diff`). Either we re-use the main fasta input of the build command for supplying the first index to operate on, and supply the rest via the `set_genomes`, or we can leave `reference` empty for the set operations and provide all inputs for the set operation via `set_genomes`. For the latter approach, they must be supplied in exact order (e.g., for `diff` I'd argue it makes sense to provide them separately so that it is clear what is subtracted from what), and it makes it more difficult to provide an idx + fa file. Ergo I opted for option 1.

- Construction of output command:
  Currently I use a ternary operator that constructs a different set of cli arguments depending on the selected subcommand (`${subcommand == 'build' ? "${reference}" : "${reference} ${set_genomes}"} > ${prefix}.idx`), but it could be simplified into

  ```
    deacon \\
        index \\
        build \\
        --threads ${task.cpus} \\
        $args \\
        $fasta 
        $set_genomes > ${prefix}.idx
  ```

  because set_genomes would be empty for the build subcommand anyway.

- There are a few extra index subcommands, like dump (creates .fa based on idx) and info, but since these result in different output types than the others, I've omitted them for now.

- I still need to add new tests for the set operation commands. For this we need idx files. Are there any in nf-test's dataset yet or can we create them on the fly in chained tests somehow?

## PR checklist

Closes #XXX 

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:


pmoris:
I've added the alternative option of splitting the various `deacon index` subcommands into separate modules. No tests yet until we decide which option to use.

If we go with this option, we'd still need to decide how to construct the inputs.  

 In the end it comes down to how the upstream genomes are most likely to be constructed. If these are individual `[meta, index]` tuples in a channel, you can:
 
 1. naively combine them via `index_ch.collect()`, and then the input to these modules would have to be a `val` (list of tuples).
 2. map them into a single element with `[meta, path(index_files)`: then the module only has a single meta.id to deal with.
 3. map them into a tuple consisting of two lists [ meta_list, path(index_files)]`, which retains the individual meta info for each genome requires the input channel to go through a mapping operation.

I'm not sure which assumptions these modules should make in this regard.

---

Additionally, for the union and intersect modules, you could argue that it makes more sense to split their inputs into a main index + a set of indexes, just like diff, but here order of operations does not matter so I combined everything into a single input.

pmoris:
See relevant discussion on Slack: https://nfcore.slack.com/archives/CE6SDBX2A/p1771355056637259

Conclusion: based on the nf-core guidelines (https://nf-co.re/docs/guidelines/components/modules#module-granularity), we'll use separate modules for the different set operations. Names should be changed to `deacon/indexbuildunion`.

For the existing index build module, we can either keep it as is (backwards compatibility + sane default) or rename it to `indexbuild` and deprecate the old one (https://nf-co.re/docs/guidelines/components/module_deprecation).

For now, I'll proceed with adding tests to the set modules and leave the existing index module as is.

famosab:
@pmoris Can this PR be closed then?

pmoris:
> @pmoris Can this PR be closed then?

No, this PR contains a draft for what was decided in the Slack discussion; it just needs to be finalized (clean up + tests).

SoonTM....

pmoris:
This one is also nearly there. The sub-sub-commands have been split over different modules. The main index build one keeps its original name for now, since it is arguably the main deacon index command and this way it won't break existing functionality.

What remains to be done is add some idx test data to the test-datasets repository so that I can finish the unit tests for intersect/union.

pmoris:
I think this is nearly done, @vagkaratzas!

I had to update deacon and fix the container registry URLs to satisfy the linter, and now all that's left is the failure on the `meta_*` naming I used. Is this a hard-rule for nf-core though? I find meta_index and meta_genome much more readable than meta and meta2.

EDIT: Slack says please use the convention.

vagkaratzas:
> I think this is nearly done, @vagkaratzas!
> 
> I had to update deacon and fix the container registry URLs to satisfy the linter, and now all that's left is the failure on the `meta_*` naming I used. Is this a hard-rule for nf-core though? I find meta_index and meta_genome much more readable than meta and meta2.
> 
> EDIT: Slack says please use the convention.

Yep, meta, meta2, sorry :P

modules/nf-core/deacon/indexdiff/tests/main.nf.test
```diff
@@ -0,0 +1,172 @@
...
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(
+                        process.out.index,
+                        process.out.findAll { key, val -> key.startsWith("versions") }
+                    ).match()
```
vagkaratzas:
```suggestion
               { assert snapshot(sanitizeOutput(process.out)).match() }
```
This is the preferred way nowadays, when everything is md5sum matched as expected. Can you update throughout please?

modules/nf-core/deacon/indexdiff/main.nf
```diff
@@ -0,0 +1,41 @@
...
+    label 'process_low'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine in ['singularity', 'apptainer'] && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/deacon:0.15.0--hdd79491_0':
+        'quay.io/biocontainers/deacon:0.15.0--hdd79491_0' }"
+
+    input:
+    tuple val(meta), path(index)      // main deacon .idx index file
+    tuple val(meta2), path(genome)    // a single fasta or .idx file to subtract from the main index
```
vagkaratzas, line 12: not sure if `genome` is the best name for this param, since it can also be an index? I'll leave that to you to decide

modules/nf-core/deacon/indexintersect/main.nf
```diff
@@ -0,0 +1,39 @@
+process DEACON_INDEXINTERSECT {
+    tag "fasta"
+
```
vagkaratzas:
```suggestion
   tag "${meta.id}"
```
oopsy

modules/nf-core/deacon/indexunion/main.nf
```diff
@@ -0,0 +1,39 @@
+process DEACON_INDEXUNION {
+    tag "fasta"
+
```
vagkaratzas:
```suggestion
    tag "${meta.id}"
```

vagkaratzas (commented):
tags should be overriden through the pipeline module.config files where necessary

modules/nf-core/deacon/indexintersect/main.nf
pmoris: Good eye...

modules/nf-core/deacon/indexdiff/tests/main.nf.test
pmoris: Sure! How do I deal with cases like https://github.com/nf-core/modules/blob/3ae65622c8c8e6ab8390a73dd75a698308a4971b/modules/nf-core/deacon/filter/tests/main.nf.test#L94 and https://github.com/nf-core/modules/blob/3ae65622c8c8e6ab8390a73dd75a698308a4971b/modules/nf-core/deacon/filter/tests/main.nf.test#L48 though?

vagkaratzas: leave those as you have already ;)

modules/nf-core/deacon/indexdiff/main.nf
pmoris: It's not ideal, but for now it'll do I hope. It's a quick change if someone comes up with something better at some point.

vagkaratzas (approved):
Good work @pmoris  :D

PR #10048, opened 2026-02-16
Update Windowmasker Modules

DLBPointon:
Updates for:
- Windowmasker/convert
- Windowmasker/mkcounts
- Windowmasker/ustat

Updates include
- Remove unnecessary .config for mkcounts
- Update to use version topics
- Update to BLAST:2.17.0 from BLAST:2.15.0 (This is unnecessary as windowmasker hasn't been updated but helps with caching if also using blast modules etc).
- Update to formatting
- Update to meta.yml

### DIFF
--- a/modules/nf-core/windowmasker/convert/environment.yml
+++ b/modules/nf-core/windowmasker/convert/environment.yml
@@ -1,7 +1,7 @@
 ---
 # yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/environment-schema.json
 channels:
-    - conda-forge
-    - bioconda
+  - conda-forge
+  - bioconda
 dependencies:
-    - bioconda::blast=2.17.0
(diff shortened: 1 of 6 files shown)


modules/nf-core/windowmasker/convert/environment.yml
```diff
@@ -4,4 +4,4 @@ channels:
   - conda-forge
   - bioconda
 dependencies:
-  - bioconda::blast=2.15.0
+  - bioconda::blast=2.17.0
```
prototaxites, line 7: Is the package not available on conda? Where does the windowmasker binary come from?

edit: is it part of BLAST?

DLBPointon: Along with a handful of others are shipped only with BLAST ¯\_(ツ)_/¯

- https://github.com/ncbi/docker/blob/master/blast/2.11.0/Dockerfile

PR #10061, opened 2026-02-16
merqury versions topic

nschan:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/alignoth/environment.yml
+++ b/modules/nf-core/alignoth/environment.yml
@@ -4,4 +4,4 @@ channels:
   - conda-forge
   - bioconda
 dependencies:
-  - bioconda::bedops=2.4.42
+  - bioconda::alignoth=1.4.6
--- a/modules/nf-core/alignoth/main.nf
+++ b/modules/nf-core/alignoth/main.nf
@@ -1,0 +1,59 @@
+process ALIGNOTH {
+    tag "${meta.id}"
+    label 'process_single'
+
+    conda "${moduleDir}/environment.yml"
+    container "${workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container
+        ? 'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/2f/2f879512a9cc25a7baa078e8a87e17b585464cf222f71be231e8f8a921354493/data'
+        : 'community.wave.seqera.io/library/alignoth:1.4.6--049c1999033885bf'}"
+
+    input:
+    tuple val(meta), path(bam), path(bai), path(fasta), path(fai), path(bed), path(vcf), path(tbi)
+
+    output:
+    tuple val(meta), path("${prefix}.html"),    emit: html,    optional: true
+    tuple val(meta), path("${prefix}"),         emit: output,  optional: true
+    tuple val(meta), path("${prefix}.vl.json"), emit: vl_json, optional: true
+    tuple val("${task.process}"), val('alignoth'), eval("alignoth --version | sed 's/alignoth //g'"), topic: versions, emit: versions_alignoth
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "${meta.id}"
+    def output_cmd = args.contains('--html')
+        ? "--html > ${prefix}.html"
+        : args.contains('--json') ? "> ${prefix}.vl.json" : "--output ${prefix}"
+    def dir_cmd = output_cmd.contains('--output') ? "mkdir ${prefix}" : ""
+    def vcf_cmd = vcf ? "--vcf ${vcf}" : ""
+    def bed_cmd = bed ? "--bed ${bed}" : ""
+    // --html is added with output_cmd and --json is not a flag used by the tool and is just here for easier module usage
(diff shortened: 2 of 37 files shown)


modules/nf-core/merqury/merqury/main.nf
```diff
@@ -28,15 +28,15 @@ process MERQURY_MERQURY {
     tuple val(meta), path("${prefix}.*.qv")      , emit: scaffold_qv
     tuple val(meta), path("*.hist.ploidy")       , emit: read_ploidy
     tuple val(meta), path("*.hapmers.blob.png")  , emit: hapmers_blob_png           , optional: true
-    path "versions.yml"                          , emit: versions
+    // WARN: Version information not provided by tool on CLI. Please update this string when bumping container versions.
+    tuple val("${task.process}"), val('merqury'), eval('echo 1.3'), emit: versions_merqury, topic: versions
```
mashehu:
```suggestion
    tuple val("${task.process}"), val('merqury'), val('1.3'), emit: versions_merqury, topic: versions
```
easier that way

modules/nf-core/merqury/merqury/meta.yml
```diff
@@ -204,13 +205,27 @@ output:
...
-        ontologies:
-          - edam: http://edamontology.org/format_3750 # YAML
+    - - ${task.process}:
+          type: string
+          description: The name of the process
+      - merqury:
+          type: string
+          description: The name of the tool
+      - echo 1.3:
+          type: eval
```
mashehu:
```suggestion
          type: string
```

modules/nf-core/merqury/hapmers/main.nf
```diff
@@ -18,7 +18,8 @@ process MERQURY_HAPMERS {
     tuple val(meta), path('*_inherited_hapmers.fl.png') , emit: inherited_hapmers_fl_png
     tuple val(meta), path('*_inherited_hapmers.ln.png') , emit: inherited_hapmers_ln_png
     tuple val(meta), path('*_inherited_hapmers.st.png') , emit: inherited_hapmers_st_png
-    path "versions.yml"                                 , emit: versions
+    // WARN: Version information not provided by tool on CLI. Please update this string when bumping container versions.
+    tuple val("${task.process}"), val('merqury'), eval('echo 1.3'), emit: versions_merqury, topic: versions
```
mashehu:
```suggestion
    tuple val("${task.process}"), val('merqury'), val('1.3'), emit: versions_merqury, topic: versions
```

modules/nf-core/merqury/hapmers/meta.yml
```diff
@@ -85,17 +84,31 @@ output:
...
-        ontologies:
-          - edam: http://edamontology.org/format_3750 # YAML
+    - - ${task.process}:
+          type: string
+          description: The name of the process
+      - merqury:
+          type: string
+          description: The name of the tool
+      - echo 1.3:
+          type: eval
```
mashehu:
```suggestion
          type: string
```

LouisLeNezet:
The issue with conda is quite tricky.
It only happens on the runner.
And it disappear when you unfold the whole: `mat_hapmer_meryl`, `pat_hapmer_meryl`.
But if you do so then nf-core linting complain due to empty file...

nschan:
There is another issue here, I think the module linter does not like '1.3' as a string and keeps complaining?

nschan:
I have added the meryl module update into this PR, since meryl is tested by merqury tests.

LouisLeNezet:
For meryl you might need to update the current version to use the:
`https://depot.galaxyproject.org/singularity/meryl:1.4.1--h4ac6f70_0`
And also pin the exact commit version in the environment.yml as it will create issues
` - bioconda::meryl=1.4.1=h4ac6f70_0` doing so will produce the correct `1.4.1`.

nschan:
I have found a way to get the linter to stop complaining about `"1.3"` in `meta.yml` but it is kinda strange and I don't think it is very good. What do you think @mashehu ?

mashehu:
no, we have other modules that work with a fixed version that work, see wittyer for example. there is something else wrong in the meta.yml, looking into it.

nschan:
@LouisLeNezet thanks for pointing out the version issue, it took me a bit to figure out what exactly you meant since the container was inconsistent within the different meryl modules and the one I looked at first happened to have the correct version.. 
The sed expression used was pretty complex, I made no effort to figure out what it does and replaced it with a simpler one.

nschan:
> no, we have other modules that work with a fixed version that work, see wittyer for example. there is something else wrong in the meta.yml, looking into it.

@mashehu Did you have a chance to look into this? I checked `wittyer` but that also seems to use an `eval`? We can also go back to `eval('echo 1.3')` if the string causes problems, at least this would be somewhat consistent in the sense that it is usually an `eval`?

Edit:
I noticed that this seems to be okay now, not sure why, thanks  if you fixed it!

nschan:
> The issue with conda is quite tricky. It only happens on the runner. And it disappear when you unfold the whole: `mat_hapmer_meryl`, `pat_hapmer_meryl`. But if you do so then nf-core linting complain due to empty file...

I have switched to checking names only, since this seems like this is some strange combination of `nf-core modules lint` and `nf-test` causing a problem on an empty file in specific contexts.

PR #10099, opened 2026-02-18
Paraphase update

bagnacan:

Paraphase's output vcf/vcf-index files are created only if the provided input BAM file has reads in the analyzed region. When the analyzed region has no reads, paraphase creates no vcf/vcf-index files. For this reason, the paraphase module should zip these files only when they are available.

This PR introduces a check for the availability of these files, as well as a new test to verify that these outputs are indeed optional. Test data needed to be extended to cover for this corner case (see [test-dataset PR here](https://github.com/nf-core/test-datasets/pull/1860)).  
Finally, this PR bumps the versions of _paraphase_ and _samtools_, and modifies the module to use topic channels.

## PR checklist

Relates to #9929.


modules/nf-core/paraphase/tests/main.nf.test
```diff
@@ -92,8 +91,43 @@ nextflow_process {
...
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert process.out.vcf?.isEmpty() },
+                { assert process.out.vcf_index?.isEmpty() },
+                { assert snapshot(
+                    file(process.out.json.get(0).get(1)).readLines()[0..28],
```
mashehu: does this need to be clear text or would a sha sum be enough or checking that it contains a specific string?

modules/nf-core/paraphase/main.nf
```diff
@@ -19,7 +19,9 @@ process PARAPHASE {
     tuple val(meta), path("*.paraphase.bam.bai")                        , emit: bai
     tuple val(meta), path("${prefix}_paraphase_vcfs/*.vcf.gz")          , emit: vcf      , optional: true
     tuple val(meta), path("${prefix}_paraphase_vcfs/*.vcf.gz.{csi,tbi}"), emit: vcf_index, optional: true
-    path "versions.yml"                                                 , emit: versions
+    tuple val("${task.process}"), val('minimap2'),  eval('minimap2 --version')      , emit: versions_minimap2   , topic: versions
+    tuple val("${task.process}"), val('paraphase'), eval('paraphase --version')     , emit: versions_paraphase  , topic: versions
+    tuple val("${task.process}"), val('samtools'),  eval('samtools --version')      , emit: versions_samtools   , topic: versions
```
mashehu:
```suggestion
    tuple val("${task.process}"), val('samtools'),  eval('samtools --version |  sed '1!d;s/.* //'')      , emit: versions_samtools   , topic: versions
```
to capture only the version and not the rest

bagnacan: Ah, good point, thank you!

modules/nf-core/paraphase/tests/main.nf.test
bagnacan: A sha sum would be enough, thanks for the tip!

modules/nf-core/paraphase/main.nf
```diff
@@ -5,8 +5,8 @@ process PARAPHASE {
     conda "${moduleDir}/environment.yml"
 
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
-        'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/65/65e79cfe3a98637330bad85befbbb0baf72366040ce1911f60be8485eae28d55/data':
-        'community.wave.seqera.io/library/minimap2_paraphase_samtools:ab39f9ad1f898e08' }"
+        'oras://community.wave.seqera.io/library/minimap2_paraphase_samtools:0a1655b13210ceb6':
```
mashehu: Please keep the https url

bagnacan: With `https` as protocol, the singularity image won't be found.
I substituted the former `https` url with the `oras` url because the former url was pointing to a docker image where the version of _paraphase_ and _samtools_ were older than the ones I'm trying to update to. I therefore created a singularity image on community.wave.seqera, and used it to run the local tests with the singularity profile.  
If `https` is preferred, then I need some help, because using it won't find the image:
```bash
  Caused by:
    Failed to pull singularity image
      command: singularity pull  --name community.wave.seqera.io-library-minimap2_paraphase_samtools-0a1655b13210ceb6.img.pulling.1772521472854 https://community.wave.seqera.io/library/minimap2_paraphase_samtools:0a1655b13210ceb6 > /dev/null
      status : 255
      hint   : Try and increase singularity.pullTimeout in the config (current is "20m")
      message:
        INFO:    Downloading network image
        FATAL:   the requested image was not found
```

I also tried by pulling the image locally before running the test. Pulling will only work with the `oras` protocol.

Do you have any recommendation?

mashehu: Please follow these steps to get the correct URL (we will soon have an CLI way for this) https://nf-co.re/docs/tutorials/nf-core_components/using_seqera_containers

bagnacan: Thanks @mashehu. I was confused because I thought I already went through all the steps in the procedure that you linked, but just I realized that I missed the last part: select `https`.

modules/nf-core/paraphase/tests/main.nf.test
```diff
@@ -53,8 +53,7 @@ nextflow_process {
                     path(process.out.vcf.get(0).get(1)).vcf.summary,
                     bam(process.out.bam.get(0).get(1)).getReadsMD5(),
                     file(process.out.bai.get(0).get(1)).name,
-                    file(process.out.vcf_index.get(0).get(1)).name,
-                    process.out.versions
```
mashehu: please catpure the version with `process.out.findAll { key, val -> key.startsWith('versions') }` instead of just dropping `process.out.versions`

bagnacan:
Closing this PR because #10325 updates paraphase an even newer release, as well as switching over to versions topic channel

PR #10118, opened 2026-02-20
Decouple Parabricks GPU count from accelerator directive

pinin4fjords:
(not sure if this is right yet, for discussion only until we have agreement in https://nfcore.slack.com/archives/CQY2U5QU9/p1771579583230519)

## Summary

- Replace `task.accelerator.request` with `task.ext.num_gpus` for the `--num-gpus` flag across all 11 Parabricks modules

All Parabricks modules currently read `task.accelerator.request` to set `--num-gpus` for pbrun. This couples two separate concerns:

1. **Executor GPU scheduling** (`accelerator`) - tells cloud executors (AWS Batch, Google Batch, K8s) to provision GPU instances
2. **Tool GPU count** (`--num-gpus`) - tells pbrun how many GPUs to use

This coupling prevents pipelines from making `accelerator` executor-conditional (as some nf-core pipelines do, e.g. Sarek), because omitting `accelerator` on local/SLURM also drops `--num-gpus` from the tool command.

Using `task.ext.num_gpus` instead lets pipeline configs set them independently:

```groovy
withLabel: process_gpu {
    ext.num_gpus = 1
    accelerator  = { task.executor in ['awsbatch','google-batch','k8s'] ? 1 : null }
}
```

Context: nf-core/rnaseq#1710, nf-core/rnaseq#1711

## Affected modules

- `parabricks/applybqsr`
- `parabricks/dbsnp`
- `parabricks/deepvariant`
- `parabricks/fq2bam`
- `parabricks/fq2bammeth`
- `parabricks/haplotypecaller`
- `parabricks/indexgvcf`
- `parabricks/minimap2`
- `parabricks/mutectcaller`
- `parabricks/rnafq2bam`
- `parabricks/starfusion`

### DIFF
--- a/modules/nf-core/parabricks/applybqsr/main.nf
+++ b/modules/nf-core/parabricks/applybqsr/main.nf
@@ -30,7 +30,7 @@ process PARABRICKS_APPLYBQSR {
     def args             = task.ext.args    ?: ''
     def prefix           = task.ext.prefix  ?: "${meta.id}"
     def interval_command = intervals        ? intervals.collect { interval -> "--interval-file ${interval}" }.join(' ') : ""
-    def num_gpus         = task.accelerator ? "--num-gpus ${task.accelerator.request}" : ''
+    def num_gpus         = task.ext.num_gpus ? "--num-gpus ${task.ext.num_gpus}" : ''
     """
     pbrun \\
         applybqsr \\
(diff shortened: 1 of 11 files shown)


pinin4fjords:
Closing this - after discussion on nf-core/rnaseq#1711, we're keeping the upstream `task.accelerator.request` pattern. `accelerator` is silently ignored on unsupported executors (including SLURM), so the value is still available via `task.accelerator.request` for `--num-gpus` even when it doesn't drive scheduling. Users who want multiple GPUs override `accelerator` in their config, which feeds both executor scheduling and the tool parameter. No need for `ext.num_gpus`.

PR #10125, opened 2026-02-21
qcatch: Add scikit-image dependency and enable doublet detection tests

an-altosian:
## Summary
- Add `scikit-image==0.26.0` as a pip dependency in `environment.yml` — required by `scanpy.pp.scrublet()` for doublet detection when no explicit threshold is provided
- Update container images to include scikit-image (`pip_qcatch_scikit-image` variants)
- Enable `--remove_doublets --visualize_doublets` flags in test config to exercise the Scrublet doublet detection code path

## Test plan
- [x] nf-test passes locally with Docker (both real test and stub)
- [x] Ran twice to confirm deterministic h5ad output
- [x] CI passes all shards (docker, conda, singularity)

### DIFF
--- a/modules/nf-core/qcatch/tests/main.nf.test.snap
+++ b/modules/nf-core/qcatch/tests/main.nf.test.snap
@@ -1,13 +1,13 @@
 {
     "test_qcatch - 1k_pbmc_v3": {
         "content": [
-            5127487,
+            4049945,
             [
                 [
                     {
                         "id": "1k_pbmc_v3"
                     },
-                    "1k_pbmc_v3_filtered_quants.h5ad:md5,2bf3a686b65c54d9e8f97c3f6728f354"
+                    "1k_pbmc_v3_filtered_quants.h5ad:md5,0642f9ce60e77af039b00a99447c1aca"
                 ]
             ],
             [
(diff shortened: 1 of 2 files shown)


modules/nf-core/qcatch/tests/main.nf.test
```diff
@@ -50,6 +50,44 @@ nextflow_process {
...
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert process.out.report },
+                { assert process.out.filtered_h5ad },
+                { assert process.out.metrics_summary },
+                { assert process.out.versions }
```
mashehu:
```suggestion
                { assert snapshot(process.out)}
```

an-altosian: Thanks for the comment! As I could not make the HTML report file deterministic (probably because of time stamp), I have to split this into multiple lines.

an-altosian:
@mashehu Good point, thanks for the link! Added the doublet detection test back (`test_qcatch - remove_doublets`) using the nf-core assertion guidelines for non-deterministic outputs:

- `{ assert process.success }` 
- File size > 0 checks for report, h5ad, and CSV (scrublet's numerical results vary across CPU architectures due to BLAS/LAPACK differences)
- `snapshot(process.out.versions).match()` for the deterministic versions.yml

The base test without doublet args keeps full md5 snapshots for h5ad and CSV. 

PR #10126

PR #10135, opened 2026-02-23
Filter to primary alignments before UMI-tools dedup

pinin4fjords:
## Summary

Add an optional `val_primary_only` parameter to the `bam_dedup_stats_samtools_umitools` subworkflow (and thread it through `bam_dedup_umi`) that filters BAM files to primary alignments only (`samtools view -F 0x900`) before running UMI-tools dedup.

### Problem

UMI-tools dedup processes all alignments in the input BAM, including secondary and supplementary alignments from multi-mapping reads. This inflates the dedup statistics, making them disconnected from the rest of the QC flow.

For example, from a real nf-core/riboseq run: a sample with ~26M primary alignments (from ~73M total alignment records including multi-mappers) reported 73M input reads and 12M output in the UMI-tools log. The actual primary-only dedup numbers were 26M in / 7.9M out, giving a true duplication rate of ~70% rather than the misleading ~84% implied by the raw UMI-tools stats. UMI-tools has no built-in option to restrict to primary alignments ([CGATOxford/UMI-tools#454](https://github.com/CGATOxford/UMI-tools/issues/454)).

### Changes

- **`bam_dedup_stats_samtools_umitools`**: New `val_primary_only` boolean input. When true, runs `SAMTOOLS_VIEW_PRIMARY` (aliased `samtools/view` with `-F 0x900 -b`) and `SAMTOOLS_INDEX_PRIMARY` before `UMITOOLS_DEDUP`. When false, passes input directly through (backwards compatible).
- **`bam_dedup_umi`**: New `umitools_dedup_primary_only` input, threaded to both genome and transcriptome `BAM_DEDUP_STATS_SAMTOOLS_UMITOOLS` calls.
- **meta.yml**: Updated for both subworkflows with new input documentation.
- **Tests**: Updated to pass the new parameter (set to `false` to preserve existing test behaviour and snapshots).

### Breaking change

Both subworkflows gain a new required input parameter. Pipelines calling these subworkflows will need to pass the additional boolean. Setting it to `false` preserves existing behaviour.

## Test plan

- [ ] Existing tests pass with `val_primary_only = false` (no snapshot changes expected)
- [ ] Add test case with `val_primary_only = true` (if suitable test data with secondary alignments is available)


subworkflows/nf-core/bam_dedup_stats_samtools_umitools/tests/main.nf.test
```diff
@@ -30,6 +32,41 @@ nextflow_workflow {
...
+        }
+
+        then {
+            assertAll(
+                { assert workflow.success},
+                { assert workflow.out.deduplog.get(0).get(1) ==~ ".*.log"},
+                { assert workflow.out.bam.get(0).get(1) ==~ ".*.bam"},
+                { assert workflow.out.bai.get(0).get(1) ==~ ".*.bai"},
+                { assert snapshot(
+                    bam(workflow.out.bam[0][1]).getSamLinesMD5(),
```
LouisLeNezet, line 44: Wouldn't it be better to collect the filename in the snapshot ?

subworkflows/nf-core/bam_dedup_stats_samtools_umitools/main.nf
```diff
@@ -2,20 +2,52 @@
...
+            .join(SAMTOOLS_INDEX_PRIMARY.out.bai, by: [0], remainder: true)
+            .join(SAMTOOLS_INDEX_PRIMARY.out.csi, by: [0], remainder: true)
+            .map {
+                meta, bam, bai, csi ->
+                    if (bai) {
+                        [ meta, bam, bai ]
+                    } else {
+                        [ meta, bam, csi ]
+                    }
+            }
```
LouisLeNezet: The following should be enough, no ?
```suggestion
        ch_dedup_input = SAMTOOLS_VIEW_PRIMARY.out.bam
            .join(SAMTOOLS_INDEX_PRIMARY.out.bai
                .mix(SAMTOOLS_INDEX_PRIMARY.out.csi),
            by: [0], remainder: true)
```

subworkflows/nf-core/bam_dedup_stats_samtools_umitools/tests/main.nf.test
pinin4fjords: Why not the content assertion?

subworkflows/nf-core/bam_dedup_stats_samtools_umitools/main.nf
pinin4fjords: good point!

subworkflows/nf-core/bam_dedup_stats_samtools_umitools/tests/main.nf.test
LouisLeNezet: It would be even better if it's stable.

pinin4fjords: The content MD5 via `getSamLinesMD5()` is a stronger assertion than a filename check - it verifies the actual BAM content is correct and deterministic, not just that a file was produced. This is the same pattern used by the existing `sarscov2_bam_bai` test (and many other subworkflow tests in this repo). `getSamLinesMD5()` strips headers (which can contain timestamps), so the hash is stable across runs.

LouisLeNezet: I was thinking about the log and the bai.

pinin4fjords: Well, that's not the line you highlighted! But OK ;-)

LouisLeNezet (approved):
Seems fine for me !

pinin4fjords:
Thanks @LouisLeNezet !

PR #9726, opened 2026-01-21
Add nextflow lint

maxulysse:
## PR checklist

Closes #XXX 

- [ ] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [ ] Remove all TODO statements.
- [ ] Broadcast software version numbers to `topic: versions` - [See version_topics](https://nf-co.re/blog/2025/version_topics)
- [ ] Follow the naming conventions.
- [ ] Follow the parameters requirements.
- [ ] Follow the input/output options guidelines.
- [ ] Add a resource `label`
- [ ] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - For modules:
    - [ ] `nf-core modules test <MODULE> --profile docker`
    - [ ] `nf-core modules test <MODULE> --profile singularity`
    - [ ] `nf-core modules test <MODULE> --profile conda`
  - For subworkflows:
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile docker`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile singularity`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile conda`

PR #10241, opened 2026-02-24
Add nextflow-lint pre-commit hook

mashehu:
Supersedes #8391 (too many files changed in that PR) and #9726 (simpler setup)

### DIFF
--- a/modules/nf-core/coverm/genome/tests/nextflow.config
+++ b/modules/nf-core/coverm/genome/tests/nextflow.config
@@ -1,8 +1,6 @@
 process {
     withName: COVERM_GENOME {
-    ext {
-        reference_mode = 'dir'   // force directory branch
-        args = '--genome-fasta-extension fasta'
-        }
+        ext.reference_mode = 'dir' // force directory branch
+        ext.args = '--genome-fasta-extension fasta'
     }
 }


.pre-commit-config.yaml
LouisLeNezet, line 1: Do we expect the nextflow version to be higher than 26 ?
Or should we add: `NXF_SYNTAX_PARSER=v2`

mashehu: good idea

.pre-commit-config.yaml
```diff
@@ -60,3 +60,11 @@ repos:
     hooks:
       - id: hadolint-docker
         args: ["--failure-threshold", "error"]
+
+  - repo: https://github.com/seqeralabs/nf-lint-pre-commit
+    rev: v0.2.0
+    hooks:
+      - id: nextflow-lint
+        files: '\.nf$|nextflow\.config$'
```
adamrtalbot, line 68: This is already on the pre-commit configuration, do you want to include `*.nextflow` files there?

https://github.com/seqeralabs/nf-lint-pre-commit/blob/d187e4c7d6fa592d1bd8cf00fa260b3ca3c10898/.pre-commit-hooks.yaml#L6

```suggestion
```

mashehu: good point

.pre-commit-config.yaml
mashehu: @adamrtalbot should we add this variable as an input-option to the hook (on by default)?

.pre-commit-config.yaml
mashehu: okay, that breaks the CI now...

.pre-commit-config.yaml
```diff
@@ -60,3 +60,10 @@ repos:
     hooks:
       - id: hadolint-docker
         args: ["--failure-threshold", "error"]
+
+  - repo: https://github.com/seqeralabs/nf-lint-pre-commit
+    rev: v0.2.0
```
mashehu, line 65:
```suggestion
    rev: 65c80e14d50363fd621e72e7bb0327fd6878f99a
```
testing the batched version

.pre-commit-config.yaml
```diff
@@ -60,3 +60,10 @@ repos:
     hooks:
       - id: hadolint-docker
         args: ["--failure-threshold", "error"]
+
+  - repo: https://github.com/seqeralabs/nf-lint-pre-commit
+    rev: 65c80e14d50363fd621e72e7bb0327fd6878f99a
```
mashehu:
```suggestion
    rev: 65b2539f4d124ba9f4f36b7562f1b9d226e15523
```

LouisLeNezet:
Shouldn't the pre-commit be ran only on the modified files ?
Currently it seems to run on all existing files.

The other issue, is that using the latest version of nextflow, generate new errors that I did not yet fix.
I think it would be best to use the stable release for this GA.

Similarly to what maxime did in his PR ?

```
      - name: Nextflow lint module ${{ matrix.module }}
        run: NXF_SYNTAX_PARSER=v2 nextflow lint modules/nf-core/${{ matrix.module }}
```

mashehu:
> Shouldn't the pre-commit be ran only on the modified files ?
> Currently it seems to run on all existing files.

We had things slip through otherwise in the past. It is fast enough to run on all files, so this will keep the repo in better shape.

> The other issue, is that using the latest version of nextflow, generate new errors that I did not yet fix.
> I think it would be best to use the stable release for this GA.

I just wanted to test something, now back at latest stable (with the same error 🙁)

LouisLeNezet:
The error is indeed strange as it does not appear locally.
However the incriminated modules does have an issue.
It should be
```
process {
    withName: COVERM_GENOME {
        ext.reference_mode = 'dir'   // force directory branch
        ext.args = '--genome-fasta-extension fasta'
    }
}
```

LouisLeNezet:
Should we merge @mashehu ?

PR #10246, opened 2026-02-24
refactor: move optional RG tags to ext.args for STAR, Sentieon STAR, and HISAT2

pinin4fjords:
## Summary
- Remove `seq_platform` and `seq_center` val inputs from **star/align** and **sentieon/staralign**
- Remove `params.seq_center` and `params.save_unaligned` references from **hisat2/align**, adding `save_unaligned` as an explicit val input
- All three aligners now build only mandatory RG tags (ID, SM) by default, with an `args.contains()` guard so callers can fully override via `ext.args`
- Optional RG tags (PL, CN) are injected through `ext.args` config closures by the calling pipeline
- Updates all subworkflows and downstream test files that call these modules

## Design principle

- File inputs and mandatory non-file inputs should be explicit module inputs
- Optional non-file inputs (like seq_platform, seq_center) should go via `ext.args`
- This lets pipelines use config closures with per-sample values from `meta` (e.g. `--outSAMattrRGline 'ID:${meta.id}' 'SM:${meta.id}' 'PL:${meta.seq_platform}'`) without forcing a single value through subworkflow plumbing
- Test configs should only be as complex as they need to be - module tests exercise the default RG behavior (ID + SM) rather than explicitly overriding it

### star/align
- Removed `val seq_platform` and `val seq_center` inputs
- Simplified `attrRG` to `--outSAMattrRGline 'ID:$prefix' 'SM:$prefix'` (guarded by `args.contains("--outSAMattrRGline")`)
- Test configs unchanged - module defaults provide correct RG tags

### sentieon/staralign
- Same as star/align

### hisat2/align
- Removed `params.seq_center` and `params.save_unaligned` references
- Added `val save_unaligned` as explicit input (controls output filenames, can't be done via ext.args)
- Simplified RG to `--rg-id $prefix --rg SM:$prefix` with `args.contains("--rg-id")` guard
- Test configs unchanged - module defaults provide correct RG tags

### fastq_align_star subworkflow
- Removed `val_seq_platform` and `val_seq_center` from take block and STAR_ALIGN call
- Test inputs renumbered; test configs unchanged (module defaults handle RG)

### fastq_align_hisat2 subworkflow
- Added `save_unaligned` to take block and HISAT2_ALIGN call

### Downstream test fixes
- **arriba/arriba**: removed old STAR_ALIGN setup inputs (seq_platform, seq_center); config unchanged
- **arriba/visualisation**: same
- **ctatsplicing/startocancerintrons**: removed old STAR_ALIGN setup inputs; config unchanged

## Breaking change

Callers that previously passed `seq_platform` or `seq_center` as val inputs to STAR/Sentieon will need to move those to ext.args closures instead. HISAT2 callers that relied on `params.seq_center` or `params.save_unaligned` will need to pass `save_unaligned` explicitly and use ext.args for RG tags.

## Context

Related to [nf-core/rnaseq#1701](https://github.com/nf-core/rnaseq/pull/1701) which adds `seq_platform` support. This PR establishes a consistent pattern across all three aligners for handling optional read group metadata.

### DIFF
--- a/modules/nf-core/sentieon/staralign/tests/nextflow.arriba.config
+++ b/modules/nf-core/sentieon/staralign/tests/nextflow.arriba.config
@@ -5,7 +5,7 @@ process {
     }
 
     withName: SENTIEON_STARALIGN {
-        ext.args = { "--readFilesCommand zcat --outSAMtype BAM Unsorted --outSAMunmapped Within --outBAMcompression 0 --outFilterMultimapNmax 50 --peOverlapNbasesMin 10 --alignSplicedMateMapLminOverLmate 0.5 --alignSJstitchMismatchNmax 5 -1 5 5 --chimSegmentMin 10 --chimOutType WithinBAM HardClip --chimJunctionOverhangMin 10 --chimScoreDropMax 30 --chimScoreJunctionNonGTAG 0 --chimScoreSeparation 1 --chimSegmentReadGapMax 3 --chimMultimapNmax 50 --outSAMattrRGline 'ID:${meta.id}' 'SM:${meta.id}' 'PL:illumina'" }
+        ext.args = '--readFilesCommand zcat --outSAMtype BAM Unsorted --outSAMunmapped Within --outBAMcompression 0 --outFilterMultimapNmax 50 --peOverlapNbasesMin 10 --alignSplicedMateMapLminOverLmate 0.5 --alignSJstitchMismatchNmax 5 -1 5 5 --chimSegmentMin 10 --chimOutType WithinBAM HardClip --chimJunctionOverhangMin 10 --chimScoreDropMax 30 --chimScoreJunctionNonGTAG 0 --chimScoreSeparation 1 --chimSegmentReadGapMax 3 --chimMultimapNmax 50'
     }
 
 }
--- a/modules/nf-core/sentieon/staralign/tests/nextflow.config
+++ b/modules/nf-core/sentieon/staralign/tests/nextflow.config
@@ -5,7 +5,7 @@ process {
     }
 
     withName: SENTIEON_STARALIGN {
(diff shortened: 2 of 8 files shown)


modules/nf-core/hisat2/align/main.nf
```diff
@@ -11,6 +11,9 @@ process HISAT2_ALIGN {
     tuple val(meta), path(reads)
     tuple val(meta2), path(index)
     tuple val(meta3), path(splicesites)
+    val seq_platform
+    val seq_center
```
maxulysse: I'd rather we use the meta map for this

modules/nf-core/hisat2/align/main.nf
```diff
@@ -33,9 +36,11 @@ process HISAT2_ALIGN {
         strandedness = meta.single_end ? '--rna-strandness R' : '--rna-strandness RF'
     }
     ss = "$splicesites" ? "--known-splicesite-infile $splicesites" : ''
-    def seq_center = params.seq_center ? "--rg-id ${prefix} --rg SM:$prefix --rg CN:${params.seq_center.replaceAll('\\s','_')}" : "--rg-id ${prefix} --rg SM:$prefix"
+    def seq_platform_arg = seq_platform ? "--rg PL:${seq_platform.replaceAll('\\s','_')}" : ''
+    def seq_center_arg   = seq_center   ? "--rg CN:${seq_center.replaceAll('\\s','_')}" : ''
+    def rg = "--rg-id ${prefix} --rg SM:${prefix} ${seq_center_arg} ${seq_platform_arg}".trim()
```
maxulysse: Can't we just use a single meta map entry for the read group here?

https://github.com/nf-core/sarek/blob/dev/workflows/sarek/main.nf#L649-L668 for reference

maxulysse: Used in
https://github.com/nf-core/sarek/blob/dev/conf/modules/aligner.config#L69

which still allow for more complex metamap usage depending on the mapper:
https://github.com/nf-core/sarek/blob/dev/conf/modules/aligner_parabricks.config#L20-L23

PR #10262, opened 2026-02-25
Fix silent read loss when Salmon strandedness inference fails

pinin4fjords:
## Summary

- Add `failOnMismatch: true` to the `.join()` between `lib_format_counts` and `auto_strand` channels in `fastq_qc_trim_filter_setstrandedness`, so the pipeline fails hard instead of silently dropping samples
- Add a diagnostic check that compares sample IDs entering auto-strandedness detection against those present in `lib_format_counts`, logging a clear error naming the affected samples and suggesting the user check their Salmon index or set strandedness explicitly
- Add a test that verifies the error fires when Salmon produces no `lib_format_counts`

Previously, if Salmon quant failed to produce `lib_format_counts` for any sample (e.g. due to a mismatched index), the inner join silently dropped those samples. Downstream processes saw an empty channel, did nothing, and the pipeline exited 0 with no warning.

Ref: nf-core/rnaseq#1693

## Test plan

- [ ] Verify the new failure test passes (`auto strandedness - should fail when salmon produces no lib_format_counts`)
- [ ] Verify existing passing tests are unaffected (fastp/ribodetector, fastp/bowtie2, single-end)
- [ ] Manually test with a mismatched Salmon index to confirm the diagnostic error message appears


pinin4fjords:
Thanks @maxulysse !

PR #10263, opened 2026-02-25
fix outdated eval for version of fastk

mashehu:


mashehu:
@nf-core-bot fix linting

PR #10297, opened 2026-03-02
fix: a few purge_dups tweaks

prototaxites:
Noticed a couple of little points for purge_dups that could be fixed:

- purgedups/purgedups outputs non-compressed BED file but purgedups/getseqs is able to read a compressed BED file, so this is compressed now.
- purgedups/getseqs hard-codes the `-e` option, which is not necessary for operation, but is recommended. Sets this as the default arg in case `ext.args` is not set, but explicitly captures the case that `ext.args` is an empty string and will pass this through in this case, allowing the `-e` option to be unset. 
- Wrap all variables in `${}` cause it's prettier :) 

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/.github/actions/nf-test-action/action.yml
+++ b/.github/actions/nf-test-action/action.yml
@@ -13,7 +13,7 @@ inputs:
   paths:
     description: "Test paths"
     required: true
-  tag:
+  tags:
     description: "Tag to pass as argument for nf-test --tag parameter"
     required: false
     default: ""
--- a/.github/workflows/nf-test-gpu.yml
+++ b/.github/workflows/nf-test-gpu.yml
@@ -128,7 +128,7 @@ jobs:
           shard: ${{ matrix.shard }}
           total_shards: ${{ env.TOTAL_SHARDS }}
           paths: "${{ join(fromJson(needs.nf-test-changes.outputs.paths), ' ') }}"
-          tag: ${{ matrix.tag }}
+          tags: ${{ matrix.tag }}
 
   confirm-pass-gpu:
     runs-on:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -67,5 +67,5 @@ repos:
       - id: nextflow-lint
         files: '\.nf$|nextflow\.config$'
         args: ["-output", "json"]
-        # TODO add formating at a later stage
+        # TODO add formatting at a later stage
         # args: ["-format", "-sort-declarations", "-spaces", "4", "-harshil-alignment", "-output", "concise"]
--- a/modules/nf-core/gawk/environment.yml
+++ b/modules/nf-core/gawk/environment.yml
@@ -4,4 +4,4 @@ channels:
   - conda-forge
   - bioconda
 dependencies:
-  - conda-forge::gawk=5.3.0
+  - conda-forge::gawk=5.3.1
--- a/modules/nf-core/gawk/main.nf
+++ b/modules/nf-core/gawk/main.nf
@@ -4,8 +4,8 @@ process GAWK {
 
     conda "${moduleDir}/environment.yml"
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
-        'https://depot.galaxyproject.org/singularity/gawk:5.3.0' :
-        'biocontainers/gawk:5.3.0' }"
+        'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/a1/a125c778baf3865331101a104b60d249ee15fe1dca13bdafd888926cc5490a34/data' :
+        'community.wave.seqera.io/library/gawk:5.3.1--e09efb5dfc4b8156' }"
 
     input:
     tuple val(meta), path(input, arity: '0..*')
@@ -51,7 +51,7 @@ process GAWK {
 
     stub:
     prefix = task.ext.prefix ?: "${meta.id}"
-    suffix = task.ext.suffix ?: "${input.getExtension()}"
+    suffix = task.ext.suffix ?: "${input.collect{ file -> file.getExtension()}.get(0)}"
     def create_cmd = suffix.endsWith("gz") ? "echo '' | gzip >" : "touch"
 
     """
--- a/modules/nf-core/gawk/tests/main.nf.test
+++ b/modules/nf-core/gawk/tests/main.nf.test
@@ -34,6 +34,33 @@ nextflow_process {
         }
     }
 
+    test("Convert fasta to bed - stub") {
+
+        options "-stub"
+
+        when {
+            params {
+                gawk_suffix = "bed"
+                gawk_args2  = '\'BEGIN { FS = OFS = "\t"}; { print \$1, "0", \$2 }\''
+            }
+            process {
+                """
+                input[0] = [
+                    [ id:'test' ],
+                    file(params.modules_testdata_base_path + 'genomics/homo_sapiens/genome/genome.fasta.fai', checkIfExists: true)
+                ]
+                input[1] = []
+                input[2] = false
+                """
+            }
+        }
+
+        then {
+            assert process.success
+            assert snapshot(process.out).match()
+        }
+    }
+
     test("Convert fasta to bed with program file") {
         when {
             params {
--- a/modules/nf-core/gawk/tests/main.nf.test.snap
+++ b/modules/nf-core/gawk/tests/main.nf.test.snap
@@ -14,16 +14,16 @@
                     [
                         "GAWK",
                         "gawk",
-                        "5.3.0"
+                        "5.3.1"
                     ]
                 ]
             }
         ],
+        "timestamp": "2026-03-04T11:31:50.761549948",
         "meta": {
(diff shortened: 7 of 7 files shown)


modules/nf-core/purgedups/getseqs/tests/main.nf.test.snap
```diff
@@ -7,7 +7,7 @@
                     {
                         "id": "test"
                     },
-                    "test.purged.fa:md5,70d0b87ecc04353767b1a713992f00fe"
+                    "test.purged.fa:md5,5593b561f5dac649dd02adbc1ff440e9"
```
maxulysse: why is the fa md5 changing?

prototaxites: Removing the `-e` option changes the operation of the tool

modules/nf-core/purgedups/getseqs/main.nf
```diff
@@ -24,10 +24,10 @@ process PURGEDUPS_GETSEQS {
     def prefix = task.ext.prefix ?: "${meta.id}"
     """
     get_seqs \\
-        $args \\
-        -e $bed \\
```
PeterKneale, line 28: given the below: do you think it would make sense to have the `-e` parameter set as a default if no other configuration is supplied? would that be a safer transition for people already consuming this module?

Image

prototaxites: Yes, that's probably a good idea

prototaxites: OK, fixed. I explicitly check if task.ext.args is a string, and if so we use it even if it's empty. Otherwise we fall back to `-e`, catching the case where it is unset and thus null.

prototaxites:
@nf-core-bot fix linting

modules/nf-core/purgedups/getseqs/main.nf
```diff
@@ -20,14 +20,16 @@ process PURGEDUPS_GETSEQS {
     task.ext.when == null || task.ext.when
 
     script:
-    def args = task.ext.args ?: ''
+    // By default, we include the '-e' option in task.ext.args as this is considered the 'safe default'
```
prototaxites, line 23:
```suggestion
    // By default, we include the '-e' option in task.ext.args as this is considered the 'safe default'
    // Check if we have a CharSequence (both String and GString inherit) and thus we can apply an empty string, which is otherwise 'falsy'
```

prototaxites:
Up to you, but FYI any nf-core member is welcome to review and approve code :) 

See https://nf-co.re/docs/contributing/how_to_contribute_to_nf-core

modules/nf-core/purgedups/getseqs/tests/main.nf.test
```diff
@@ -22,6 +23,36 @@ nextflow_process {
...
+    test("sarcov2 - fasta - bed - no e option") {
+
+        // load a config file where we load ext.args as an empty string to unset -e option
+        config './nextflow.config'
+
+        when {
+            process {
+                """
+                ch_bed = Channel.of('MT192765.1\\n0\\t29828\\tJUNK\\n').collectFile(name: 'test.bed' )
+                input[0] = Channel.of([
```
vagkaratzas:
```suggestion
                ch_bed = channel.of('MT192765.1\\n0\\t29828\\tJUNK\\n').collectFile(name: 'test.bed' )
                input[0] = channel.of([
```
Let's try to keep them like that throughout already

modules/nf-core/purgedups/purgedups/main.nf
```diff
@@ -24,16 +24,18 @@ process PURGEDUPS_PURGEDUPS {
...
+        -c ${basecov} \\
+        ${paf} | gzip \\
+        > ${prefix}.dups.bed.gz \\
+        2>| >(tee ${prefix}.purge_dups.log >&2)
     """
 
     stub:
     def prefix = task.ext.prefix ?: "${meta.id}"
     """
-    touch ${prefix}.dups.bed
```
vagkaratzas, line 36: echo the args as well

modules/nf-core/purgedups/purgedups/main.nf
```diff
@@ -24,16 +24,18 @@ process PURGEDUPS_PURGEDUPS {
     def prefix = task.ext.prefix ?: "${meta.id}"
     """
     purge_dups \\
-        $args \\
-        -T $cutoff \\
-        -c $basecov \\
-        $paf > ${prefix}.dups.bed 2>| >(tee ${prefix}.purge_dups.log >&2)
+        ${args} \\
+        -T ${cutoff} \\
```
vagkaratzas, line 28: is this a mandatory input? else remove and let it be provided by ext.args

modules/nf-core/purgedups/purgedups/main.nf
```diff
@@ -24,16 +24,18 @@ process PURGEDUPS_PURGEDUPS {
...
+        -c ${basecov} \\
```
vagkaratzas, line 29: same for this (cant see inputs here to see if this is a file)

modules/nf-core/purgedups/histplot/main.nf
```diff
@@ -27,9 +27,9 @@ process PURGEDUPS_HISTPLOT {
     def prefix = task.ext.prefix ?: "${meta.id}"
     """
     hist_plot.py \\
-        -c $cutoff \\
-        $args \\
-        $statfile \\
+        -c ${cutoff} \\
```
vagkaratzas: If this is not mandatory, provide through ext.args instead

modules/nf-core/purgedups/getseqs/tests/main.nf.test
```diff
@@ -22,6 +23,36 @@ nextflow_process {
...
+                    [ id:'test'],
+                    file(params.modules_testdata_base_path + 'genomics/sarscov2/genome/genome.fasta', checkIfExists: true)
+                ]).combine(ch_bed)
+                """
+            }
+        }
+
```
vagkaratzas, line 88: Maybe also add a test with a config that passes somethign else rather than empty or -e? Else it is not obvious how this can get overriden

modules/nf-core/purgedups/purgedups/main.nf
prototaxites: I believe these are mandatory for this one

modules/nf-core/purgedups/histplot/main.nf
prototaxites: done

prototaxites:
test with args unset (`null`):
```
#!/bin/bash -ue
get_seqs \
    -e \
    test.bed \
    -p test \
    genome.fasta
```

test with args set to empty string:
```
#!/bin/bash -ue
get_seqs \
     \
    test.bed \
    -p test \
    genome.fasta
```

test with args set explicitly to another string, `-es`:
```
#!/bin/bash -ue
get_seqs \
    -es \
    test.bed \
    -p test \
    genome.fasta
```

modules/nf-core/purgedups/getseqs/tests/main.nf.test
prototaxites: I have added an extra test that adds some explicit args, to demonstrate that setting args always overrides the default `-e`.

modules/nf-core/purgedups/getseqs/tests/main.nf.test
```diff
@@ -10,18 +10,90 @@ nextflow_process {
...
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(
+                    file(process.out.haplotigs[0][1]).name, // Empty file
```
vagkaratzas: hmm when is this non-empty then? does it need another test maybe?

modules/nf-core/purgedups/purgedups/meta.yml
```diff
@@ -44,10 +44,10 @@ output:
           description: |
             Groovy Map containing sample information
             e.g. [ id:'test', single_end:false ]
-      - "*.dups.bed":
+      - "*.dups.bed.gz":
           type: file
           description: A bed file of sequences purged of false duplications
-          pattern: "*.bed"
+          pattern: "*.bed.gz"
           ontologies: []
```
vagkaratzas: this can probably be fille dbefore merging

modules/nf-core/purgedups/getseqs/tests/main.nf.test
prototaxites: It's non-empty in real cases where there are haplotigs actually being purged from a genome assembly. I'm not sure how to make this work in this test case...

prototaxites: The test BED file was broken... fixed and the file is not empty!

modules/nf-core/purgedups/purgedups/meta.yml
prototaxites: Done!

PR #10299, opened 2026-03-02
fix nextflow lint for samtools/addreplacerg

mashehu:


mashehu:
> Shouldn't it be both with def ?

I thought it should only be done once per script block, no? I am no expert there

PR #10303, opened 2026-03-02
fix long nextflow linting message in CI

mashehu:
not the prettiest but better then thousands of lines (and some more empty ones)


LouisLeNezet (approved):
This will indeed be at least a bit better

PR #10306, opened 2026-03-02
feat(cellpose): update to v4.0.8 with GPU support

an-altosian:
## Summary

- Upgrade cellpose from 3.1.0 to 4.0.8 (includes Cellpose-SAM integration)
- Replace `biocontainers/cellpose:3.1.0_cv1` with GPU-enabled `dongzehe/cellpose-gpu:4.0.8`
- Add GPU support via `process_gpu` label and `task.ext.use_gpu` flag
- Change `model` input from `path` to `val` (pretrained model name string)
- Add `cells` output (`*seg.npy`) for downstream analysis
- Outputs moved to `${prefix}/` directory for consistency
- Add `Dockerfile` for reproducible container builds

## Changes

| File | Change |
|------|--------|
| `main.nf` | Upgrade to v4.0.8, GPU container, `val model` input, `task.ext.use_gpu`, `cells` output |
| `meta.yml` | Match new interface: `model` as `val` string, add `cells` output |
| `Dockerfile` | New: GPU container based on `pytorch/pytorch:2.5.1-cuda12.4-cudnn9-runtime` |
| `tests/main.nf.test` | Update inputs for new interface |
| `tests/main.nf.test.snap` | Updated snapshots for v4.0.8 outputs |

## Container

`dongzehe/cellpose-gpu:4.0.8` (hosted on quay.io, public)
- Base: `pytorch/pytorch:2.5.1-cuda12.4-cudnn9-runtime`
- Packages: cellpose==4.0.8, numpy, scipy, matplotlib, scikit-image, opencv-python-headless
- Falls back to CPU automatically when no GPU is available

## GPU control

```groovy
withName: 'CELLPOSE' {
    ext.use_gpu = true  // passes --use_gpu to cellpose CLI
}
```

When `task.ext.use_gpu` is not set or false, cellpose runs on CPU only.

## Usage

```groovy
CELLPOSE(
    ch_images,      // tuple val(meta), path(image)
    "cyto3"         // model name: 'cyto3', 'nuclei', 'CPSAM', or ''
)
```

## Licensing

- Cellpose: BSD-3-Clause
- SAM (`segment_anything`): Apache 2.0
- Model weights NOT bundled (downloaded at runtime)

### DIFF
--- a/modules/nf-core/cellpose/tests/main.nf.test.snap
+++ b/modules/nf-core/cellpose/tests/main.nf.test.snap
@@ -34,7 +34,7 @@
                     [
                         "CELLPOSE",
                         "torch",
-                        "2.10.0"
+                        "2.10.0+cu128"
                     ]
                 ]
             }
@@ -43,7 +43,7 @@
             "nf-test": "0.9.3",
             "nextflow": "25.10.2"
         },
-        "timestamp": "2026-03-06T15:04:19.24771681"
+        "timestamp": "2026-03-09T17:04:23.355608629"
     },
     "cellpose - stub": {
         "content": [
@@ -66,21 +66,21 @@
                     [
                         "CELLPOSE",
                         "cellpose",
-                        ""
+                        "4.0.9"
                     ]
                 ],
                 "4": [
                     [
                         "CELLPOSE",
                         "python",
-                        ""
+                        "3.12.12"
                     ]
                 ],
                 "5": [
                     [
                         "CELLPOSE",
                         "torch",
-                        ""
+                        "2.10.0+cu128"
                     ]
                 ],
                 "cells": [
@@ -101,21 +101,21 @@
                     [
                         "CELLPOSE",
                         "cellpose",
-                        ""
+                        "4.0.9"
                     ]
                 ],
                 "versions_python": [
                     [
                         "CELLPOSE",
                         "python",
-                        ""
+                        "3.12.12"
                     ]
                 ],
                 "versions_torch": [
                     [
                         "CELLPOSE",
                         "torch",
-                        ""
+                        "2.10.0+cu128"
                     ]
                 ]
             }
@@ -124,6 +124,6 @@
             "nf-test": "0.9.3",
             "nextflow": "25.10.2"
         },
-        "timestamp": "2026-03-09T16:54:30.847780039"
+        "timestamp": "2026-03-09T17:04:40.207673084"
     }
 }


an-altosian:
Hi there, I saw that the cellpose module was updated three weeks ago but is still on cellpose 3. Here is a draft of updating it to cellpose v4 with GPU support, which has its cellpose-SAM based model compatibility. It works great on my side.

modules/nf-core/cellpose/Dockerfile
mashehu, line 1: is this dockerfile really needed? can't we just build them via seqera containers https://nf-co.re/docs/tutorials/nf-core_components/using_seqera_containers

an-altosian: Thanks for the comment! The reason we have to have this Dockerfile is that I want to add GPU support for this module, but I could not get cuda runtime installed via seqera containers. So I started with `pytorch/pytorch:2.5.1-cuda12.4-cudnn9-runtime`. If there is a way to do specify base image in seqera containers or other ways to add cuda supportm, for example specify env var `CONDA_OVERRIDE_CUDA`, please do let me know, so that I can remove the dockerfile!

modules/nf-core/cellpose/meta.yml
```diff
@@ -22,36 +24,55 @@ input:
...
         ontologies: []
-  - model:
-      type: file
-      description: Optional input file. Cellpose 2 model trained by user using human-in-the-loop
-        approach.
-      ontologies: []
+  - - model_name:
+        type: string
+        description: |
+          Cellpose pretrained model name (e.g. 'cyto3', 'nuclei', 'cyto2', 'CPSAM').
```
camlloyd: Are you certain these are valid model arguments for cellpose4?

modules/nf-core/cellpose/meta.yml
```diff
@@ -60,6 +81,24 @@ output:
...
+notes: |
+  This module requires a Docker/Singularity/Podman container (conda is not supported
+  due to complex CUDA/PyTorch dependencies).
+
+  GPU support: Set `task.ext.use_gpu = true` to pass `--use_gpu` to cellpose.
+  The container includes CUDA 12.4 and falls back to CPU automatically when no GPU
+  is available. Use the `process_gpu` label to request GPU resources from your executor.
+
+  Model selection via the model input channel (tuple of name and file):
+    - Pretrained model:  ["cyto3", []]
```
camlloyd: Is this also valid?

modules/nf-core/cellpose/meta.yml
an-altosian: Good catch! Those were leftovers from cellpose3. updated!

modules/nf-core/cellpose/Dockerfile
an-altosian: Could you let me know what is the best practice: Should I provide a GPU ready container, or specifying CPU only containers inthe module and provide a Dockerfile and note in the meta.yml saying for GPU runs, please build the GPU ready container by themself?

modules/nf-core/cellpose/meta.yml
an-altosian: Addressed!

modules/nf-core/cellpose/tests/main.nf.test
```diff
@@ -28,7 +28,7 @@ nextflow_process {
                 { assert process.success },
                 { assert snapshot(process.out.mask).match("mask") },
                 { assert snapshot(process.out.flows).match("flows") },
-                { assert snapshot(process.out.versions).match("versions") }
+                { assert snapshot(process.out.versions_cellpose).match("versions") }
```
camlloyd: See https://nf-co.re/docs/tutorials/migrate_to_topics/update_modules#updating-standard-modules

modules/nf-core/cellpose/tests/main.nf.test.snap
```diff
@@ -14,14 +14,18 @@
     "versions": {
         "content": [
             [
-                "versions.yml:md5,398393e73a80fc622873765256d5ec79"
+                [
+                    "CELLPOSE",
+                    "cellpose",
```
camlloyd: This version doesn't look quite right. Should be `4.0.8`?

camlloyd: Are you also able to capture:

- python version
- torch version

an-altosian:
Great. I figured out how to build GPU ready containes using seqera containers. Thanks everybody!

modules/nf-core/cellpose/tests/main.nf.test.snap
an-altosian: addressed!

camlloyd:
I want to highlight there appears to be some overlap with https://github.com/nf-core/modules/pull/9793. Worth coordinating to avoid duplicating effort.

an-altosian:
Thank you @camlloyd. I have merged the changes in that PR here. 

Hi @FloWuenne, Sorry I did not notice your PR when I was doing this. I just wanted to update cellpose so that I can use it in spatialxe. 🙏

modules/nf-core/cellpose/tests/main.nf.test.snap
camlloyd: What if you capture these from the other lines output by `cellpose --version`?

modules/nf-core/cellpose/tests/nextflow.config
```diff
@@ -0,0 +1,9 @@
+docker {
+    fixOwnership = true
+}
+
+process {
+    withName: "CELLPOSE" {
+        ext.args = '--diameter 9 --channel_axis 0 --no_npy --save_flows'
```
camlloyd: Is `--diameter 9` necessary here?

DongzeHE: You know cellpose toooo well! 🤣

DongzeHE: all addressed!

modules/nf-core/cellpose/main.nf
```diff
@@ -1,56 +1,63 @@
 process CELLPOSE {
-    tag "$meta.id"
-    label 'process_medium'
+    tag "${meta.id}"
+    label 'process_high'
```
FloWuenne: Are we sure we need to change the label to `process_high`? This is highly dependent on the input images. 42GB (the process_medium default) feels like a pretty high base amount already. I think keeping process_medium would be fine.

modules/nf-core/cellpose/main.nf
```diff
@@ -1,56 +1,63 @@
...
 
     script:
-    // Exit if running this module with -profile conda / -profile mamba
-    if (workflow.profile.tokenize(',').intersect(['conda', 'mamba']).size() >= 1) {
-        error "I did not manage to create a cellpose module in Conda that works in all OSes. Please use Docker / Singularity / Podman instead."
-    }
     def args = task.ext.args ?: ''
-    def model_command = model ? "--pretrained_model $model" : ""
+    def model_command = model ? "--pretrained_model ${model}" : ""
+    def gpu_flag = task.ext.use_gpu ? "--use_gpu" : ""
```
FloWuenne:
```suggestion
    def gpu_flag = task.accelerator ? "--use_gpu" : ""
```

Instead of setting an additional ext flag here, we should be able to use accelerator so that when a GPU was actually requested for the task, it will use it. Accelerator should become true based on the gpu label for the process!

modules/nf-core/cellpose/meta.yml
```diff
@@ -37,32 +42,115 @@ output:
...
+          description: The name of the tool
+      - "pip show torch | sed -n 's/^Version: //p'":
+          type: eval
+          description: The expression to obtain the version of the tool
+notes: |
+  GPU support: Set `task.ext.use_gpu = true` to pass `--use_gpu` to cellpose.
+  The container (built via Seqera Containers) includes PyTorch 2.10.0 with CUDA 12.8
+  and falls back to CPU automatically when no GPU is available. Use the `process_gpu`
+  label to request GPU resources from your executor. When running with conda/mamba,
+  GPU support depends on having a CUDA-enabled PyTorch installation in your environment.
```
FloWuenne:
```suggestion
When `accelerator` is set (e.g. `accelerator = 1`), the module
  automatically passes `--use_gpu` to cellpose. The container (built via Seqera
  Containers) includes PyTorch 2.10.0 with CUDA 12.8 and falls back to CPU
  automatically when no GPU is available. Use the `process_gpu` label to request
  GPU resources from your executor. When running with conda/mamba, GPU support
  depends on having a CUDA-enabled PyTorch installation in your environment.
```

FloWuenne:
@an-altosian The last thing potentially missing is an extra test profile for the gpu functionality with a config that sets accelerator to 1 for Cellpose. We could basically just replicate the existing test profile and point to the gpu config. This is the pattern that would be needed in the config:

`process {
    withName: "CELLPOSE" {
        ext.args    = '--no_npy --save_flows'
        accelerator = 1
    }
}
`
@mashehu What's the best approach for this so we can utilize gpu runners to test that specific nf-test?

FloWuenne (changes requested):

modules/nf-core/cellpose/main.nf
DongzeHE: changed back to medium! I am updating with Xenium data in my mind, which is always huge. But I agree, usually, we only need proces_medium.

modules/nf-core/cellpose/main.nf
DongzeHE: Done!

modules/nf-core/cellpose/meta.yml
DongzeHE: accepted!

modules/nf-core/cellpose/tests/main.nf.test
mashehu, line 9:
```suggestion
    tag "cellpose"
    tag "gpu"
```
to test them with gpu runners

FloWuenne: @mashehu Since this can run on cpu or gpu, use different tags throughout the same main.nf.test or create a second nf.test file specifically for GPU runners so it's tested on both cpu and gpu?

mashehu: i would go for a seperate test file to get both options tested

FloWuenne: @DongzeHE Can you add an additional nf-test file with the gpu tag please? It can be exactly the same tests, just need that gpu tag so it triggers the gpu runners. That would round up the module update to also have both cpu and gpu tests included and I think we can merge after!

an-altosian: Done!

modules/nf-core/cellpose/tests/main.nf.test.snap
DongzeHE: updated!

camlloyd: Please check your stub test snapshots

DongzeHE: oooops, forgot to specify docker profile while running it. Updated!

FloWuenne:
I think this all looks great now @DongzeHE! GPU nf-test seems to have been run successfully as well! 🚀

an-altosian:
Thank you very much. Could you merge the PR?

PR #10328, opened 2026-03-04
Fix paraphase

bagnacan:

Paraphase's output vcf/vcf-index files are created only if the provided input BAM file has reads in the analyzed region. When the analyzed region has no reads, paraphase creates no vcf/vcf-index files. For this reason, the paraphase module should zip these files only when they are available.

This PR introduces a check for the availability of these files, as well as a new test to verify that these outputs are indeed optional. Test data needed to be extended to cover for this corner case (see [test-dataset PR here](https://github.com/nf-core/test-datasets/pull/1860)).  
Finally, this PR bumps the versions of _paraphase_ and _samtools_, and modifies the module to use topic channels.

## PR checklist

Relates to #9929 and #10099.

### DIFF
--- a/modules/nf-core/paraphase/tests/main.nf.test
+++ b/modules/nf-core/paraphase/tests/main.nf.test
@@ -3,6 +3,7 @@ nextflow_process {
     name "Test Process PARAPHASE"
     script "../main.nf"
     process "PARAPHASE"
+    config "./nextflow.config"
 
     tag "modules"
     tag "modules_nfcore"
@@ -24,8 +25,6 @@ nextflow_process {
     }
 
     test("homo_sapiens - [ bam, bai ], []") {
-
-        config "./nextflow.config"
 
         when {
             params {
@@ -67,8 +66,6 @@ nextflow_process {
 
     test("homo_sapiens - [ bam, bai ], config") {
 
-        config "./nextflow.config"
-
         when {
             params {
                 module_args = "--gene PRODH"
@@ -107,8 +104,6 @@ nextflow_process {
     }
 
     test("homo_sapiens - [ bam, bai ], no_vcf") {
-
-        config "./nextflow.config"
 
         when {
             params {
@@ -151,6 +146,9 @@ nextflow_process {
         config "./nextflow.csi_index.config"
 
         when {
+            params {
+                module_args = "--gene PRODH"
+            }
             process {
                 """
                 input[0] = [
@@ -187,7 +185,6 @@ nextflow_process {
     test("homo_sapiens - [ bam, bai ], [] - stub") {
 
         options "-stub"
-        config "./nextflow.config"
 
         when {
             params {
@@ -220,7 +217,6 @@ nextflow_process {
     test("homo_sapiens - [ bam, bai ], config - stub") {
 
         options "-stub"
-        config "./nextflow.config"
 
         when {
             params {
@@ -256,6 +252,9 @@ nextflow_process {
         config "./nextflow.csi_index.config"
 
         when {
+            params {
+                module_args = "--gene PRODH"
+            }
             process {
                 """
                 input[0] = [
--- a/modules/nf-core/paraphase/tests/nextflow.csi_index.config
+++ b/modules/nf-core/paraphase/tests/nextflow.csi_index.config
@@ -1,6 +1,5 @@
 process {
     withName: 'PARAPHASE' {
-        ext.args = '--gene PRODH'
         ext.args3 = '--csi'
     }
 }


modules/nf-core/paraphase/tests/main.nf.test
```diff
@@ -93,7 +93,44 @@ nextflow_process {
...
+                    process.out.findAll { key, val -> key.startsWith("versions") }
+                    ).match()
+                }
+            )
+        }
+    }
+
+    test("homo_sapiens - [ bam, bai ], no_vcf") {
+
+        config "./nextflow.no_vcf.config"
```
fellen31: Could you change to use just one nextflow.config and https://nf-co.re/docs/guidelines/components/modules#configuration-of-extargs-in-tests?

modules/nf-core/paraphase/tests/main.nf.test
```diff
@@ -93,7 +93,44 @@ nextflow_process {
...
+                ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert process.out.vcf?.isEmpty() },
+                { assert process.out.vcf_index?.isEmpty() },
```
fellen31: Can `process.out.vcf` ever be `null`? 

```suggestion
                { assert process.out.vcf.isEmpty() },
                { assert process.out.vcf_index.isEmpty() },
```

modules/nf-core/paraphase/tests/main.nf.test
```diff
@@ -93,7 +93,44 @@ nextflow_process {
...
+                { assert snapshot(
+                    file(process.out.json.get(0).get(1)).readLines()[0..28],
+                    process.out.bam.get(0).get(1).isEmpty(),
```
fellen31: Shouldn't there still be a bam output? Looks like it in the in the snapshot since `process.out.bam.get(0).get(1).isEmpty()` is `false`. Could we at least snapshot the name if `bam(process.out.bam.get(0).get(1)).getReadsMD5()` doesn't work?

modules/nf-core/paraphase/tests/main.nf.test
bagnacan: Can be removed here. I think I left it there when I was checking paraphase's behavior before updating the module's script.

modules/nf-core/paraphase/tests/main.nf.test
bagnacan: Good catch, I'll change that line.

modules/nf-core/paraphase/tests/main.nf.test
bagnacan: Thanks for the pointer. I didn't know what one _should_ use a single nextflow.config when testing.

However, when I add a `when.params` block:
```
    test("homo_sapiens - [ bam, bai ], no_vcf") {

        //config "./nextflow.no_vcf.config"

        when {
            params {
                args = "--gene RIMBP3B"
            }
            process {
                ...
```
the `gene` parameter does not overwrite the one provided by `./nextflow.config`. In fact, a `PRODH` VCF file is always created.  
Since the point of this test is to show that the process _optionally_ creates VCFs, then the presence of the `PRODH` VCF file invalidates the purpose of this test.  
Am I defining the params block incorrectly? Otherwise, I propose to keep the 2nd `nextflow.no_vcf.config`.  
What do you think?

fellen31: I think you need to update both the no_vcf test, and the previous tests with `params { ... }` blocks.

bagnacan: It still does not work if I remove the `config` directive from any/all of the tests (and replace it with a reative `when.params` block).  
Just to be sure, I even commented out every other test, and run the `no_vcf` test alone. It still produces a PRODH VCF file. So I think that the `./nextflow_config` is always loaded unless ovewritten.  
Do you have any recommendation?

fellen31: Could you push your changes so I can take a look?

bagnacan: Thanks, @fellen31.
I pushed 3 commits, the last of which tries to address your 1st comment.

modules/nf-core/paraphase/tests/nextflow.no_vcf.config
fellen31, line 1: Remove this.

modules/nf-core/paraphase/tests/main.nf.test
fellen31: Cool, so what's missing is that you need to update `.nextflow.config` from 

```
process {
  withName: 'PARAPHASE' {
    ext.args = '--gene PRODH'
  }
}
```

to

```
process {
  withName: 'PARAPHASE' {
    ext.args = params.module_args
  }
}
```

so that `--gene PRODH` is not hardcoded, and `params.module_args` can be picked up instead and sent to the module (via ext.args). 

Then use `module_args`, not `ext.args` in main.nf.test‎, e.g.: 

```
params {
    module_args = '--gene RIMBP3B'
  }
```

for your no_vcf tests, and replace it with 

```
params {
    module_args = '--gene PRODH'
  }
```

in the other tests.

bagnacan: Ah, very nice, I didn't know I could do:

```
process {
  withName: 'PARAPHASE' {
    ext.args = params.module_args
  }
}
```

Thanks for the tip!

modules/nf-core/paraphase/tests/main.nf.test
fellen31, line 7:
```suggestion
process "PARAPHASE"
config "./nextflow.config"

```

modules/nf-core/paraphase/tests/main.nf.test
```diff
@@ -28,6 +28,9 @@ nextflow_process {
         config "./nextflow.config"
```
fellen31:
```suggestion
```

modules/nf-core/paraphase/tests/main.nf.test
```diff
@@ -67,6 +70,9 @@ nextflow_process {
...
```
fellen31:
```suggestion
```

modules/nf-core/paraphase/tests/main.nf.test
```diff
@@ -93,7 +99,47 @@ nextflow_process {
...
+                    ).match()
+                }
+            )
+        }
+    }
+
+    test("homo_sapiens - [ bam, bai ], no_vcf") {
+
+        config "./nextflow.config"
+
```
fellen31:
```suggestion
```

modules/nf-core/paraphase/tests/main.nf.test
```diff
@@ -174,6 +223,9 @@ nextflow_process {
         config "./nextflow.config"
```
fellen31:
```suggestion
```

modules/nf-core/paraphase/tests/main.nf.test
```diff
@@ -144,6 +190,9 @@ nextflow_process {
...
```
fellen31:
```suggestion
```

PR #10332, opened 2026-03-04
feat(stardist): update to Seqera Containers GPU image with model input

an-altosian:
## Summary
- Switch stardist container to Seqera Containers with CUDA support (stardist 0.9.2, TensorFlow 2.20.0, Python 3.12) with singularity/docker conditional
- Add `process_gpu` label for GPU resource requests; TensorFlow auto-detects GPU at runtime
- Add `[model_name, model_path]` tuple input for model selection (`-m` is required by stardist-predict2d):
  - Pretrained model: `[ '2D_versatile_fluo', [] ]`
  - Custom model directory: `[ '', file("/path/to/model") ]`
  - Via ext.args only: `[ '', [] ]`
- Replace `versions.yml` with `eval()` topic versions via `pip show` (stardist, python, tensorflow, tifffile)
- Update `environment.yml` with pip dependencies including NVIDIA CUDA 12 packages and imagecodecs
- Add `docker { fixOwnership = true }` to test config for self-hosted runners

## Context
Follows the same pattern as the cellpose update in #10306.

## Test plan
- [x] `nf-test test modules/nf-core/stardist/tests/main.nf.test --profile docker_self_hosted` — both full and stub tests pass
- [x] Verified version strings via `pip show` inside the container match snapshot values
- [ ] CI nf-test run on GitHub Actions

### DIFF
--- a/modules/nf-core/stardist/tests/gpu.nf.test
+++ b/modules/nf-core/stardist/tests/gpu.nf.test
@@ -1,0 +1,61 @@
+nextflow_process {
+
+    name "Test Process STARDIST on GPU"
+    script "../main.nf"
+    process "STARDIST"
+
+    tag "modules"
+    tag "modules_nfcore"
+    tag "stardist"
+    tag "gpu"
+
+    test("stardist2d - gpu - tif") {
+
+        when {
+            config "./nextflow.config"
+            process {
+                """
+                input[0] = [
+                    [ id:'test' ], // meta map
+                    file(params.modules_testdata_base_path + 'imaging/segmentation/nuclear_image.tif', checkIfExists: true)
+                ]
+                input[1] = [ '2D_versatile_fluo', [] ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
+            )
+        }
+
+    }
+
+    test("stardist2d - gpu - tif - stub") {
+
+        options "-stub"
+
+        when {
+            process {
+                """
+                input[0] = [
+                    [ id:'test' ], // meta map
+                    file(params.modules_testdata_base_path + 'imaging/segmentation/nuclear_image.tif', checkIfExists: true)
+                ]
+                input[1] = [ '2D_versatile_fluo', [] ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
+            )
+        }
+
+    }
+
+}


an-altosian:
Hi @mashehu @camlloyd ,

Sorry for bothering you again. This time is for stardist. I basically did the same thing as in #10306 and applied what I learned from there. Please take a look!

Best,
Dongze

modules/nf-core/stardist/tests/main.nf.test
```diff
@@ -17,14 +17,16 @@ nextflow_process {
...
                 """
             }
         }
 
         then {
             assertAll(
                 { assert process.success },
-                { assert snapshot(process.out).match() }
+                { assert snapshot(process.out.mask).match("mask") },
+                { assert snapshot(process.out.findAll { key, val -> key.startsWith('versions') }).match("versions") }
```
SPPearce: If you put all the snapshots into one, then you don't have to specify the name of the snapshot.
```suggestion
                { assert snapshot(
                    process.out.mask,
                    process.out.findAll { key, val -> key.startsWith('versions') }
                    ).match() }
```

modules/nf-core/stardist/main.nf
```diff
@@ -1,48 +1,41 @@
...
     tuple val(meta), path(image)
+    tuple val(model_name), path(model_path)
 
     output:
     tuple val(meta), path("*.stardist.tif"), emit: mask
-    path "versions.yml"                    , emit: versions
+    tuple val("${task.process}"), val('stardist'), eval("pip show stardist | grep '^Version:' | sed 's/Version: //'"), topic: versions, emit: versions_stardist
+    tuple val("${task.process}"), val('python'), eval("python --version | sed 's/Python //'"), topic: versions, emit: versions_python
+    tuple val("${task.process}"), val('tensorflow'), eval("pip show tensorflow | grep '^Version:' | sed 's/Version: //'"), topic: versions, emit: versions_tensorflow
+    tuple val("${task.process}"), val('tifffile'), eval("pip show tifffile | grep '^Version:' | sed 's/Version: //'"), topic: versions, emit: versions_tifffile
```
SPPearce: You could shorten these:
`grep '^Version:' | sed 's/Version: //'` can become `sed -n 's/^Version: //p'`
(`sed -n` doesn't automatically print any line, then the `p` at the end of the replacement says to print after it finds a match).

modules/nf-core/stardist/tests/main.nf.test
DongzeHE: thank you! addressed. Iet's just do `{ assert snapshot(process.out).match() }`

modules/nf-core/stardist/main.nf
DongzeHE: addressed!

an-altosian:
Any other comments?

an-altosian:
Hi @kbestak do you have any comments on this PR before I merge it?

kbestak:
Hi @an-altosian, looks great to me, looking forward to trying it out. Amazing push you've been doing!

PR #10413, opened 2026-03-11
Migrate atlas/pmd to topic channels

garciafranciscomartn:
Closes nf-core/modules#10208

### DIFF
--- a/modules/nf-core/atlas/pmd/main.nf
+++ b/modules/nf-core/atlas/pmd/main.nf
@@ -17,7 +17,7 @@ process ATLAS_PMD {
     tuple val(meta), path("*_PMD_input_Exponential.txt"), emit: exponential
     tuple val(meta), path("*_PMD_Table_counts.txt")     , emit: counts
     tuple val(meta), path("*_PMD_Table.txt")            , emit: table
-    tuple val("${task.process}"), val('atlas'), eval('atlas | sed -e "2!d;s/^[ \\t]*Atlas //"'), emit: versions_atlas, topic: versions
+    tuple val("${task.process}"), val('atlas'), eval('atlas | sed -e "2!d;s/.*Atlas //"'), emit: versions_atlas, topic: versions
 
     when:
     task.ext.when == null || task.ext.when
--- a/modules/nf-core/atlas/pmd/meta.yml
+++ b/modules/nf-core/atlas/pmd/meta.yml
@@ -107,7 +107,7 @@ output:
       - "atlas":
           type: string
           description: The name of the tool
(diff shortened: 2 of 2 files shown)


modules/nf-core/atlas/pmd/main.nf
```diff
@@ -17,7 +17,7 @@ process ATLAS_PMD {
     tuple val(meta), path("*_PMD_input_Exponential.txt"), emit: exponential
     tuple val(meta), path("*_PMD_Table_counts.txt")     , emit: counts
     tuple val(meta), path("*_PMD_Table.txt")            , emit: table
-    path "versions.yml"                                 , emit: versions
+    tuple val("${task.process}"), val('atlas'), eval('(atlas 2>&1) | grep Atlas | head -n 1 | sed -e "s/^[ \\t]*Atlas //"'), emit: versions_atlas, topic: versions
```
LouisLeNezet: Could you please simplify the evaluated expression.
You can use `-n` or `'2!d;s...` for the sed command to remove the grep and head call

Not sure if the following would work, just check in singularity what's the output of `atlas` 😉 
```suggestion
    tuple val("${task.process}"), val('atlas'), eval('atlas | sed -e "2!d;s/^[ \\t]*Atlas //"'), emit: versions_atlas, topic: versions
```

garciafranciscomartn: Great suggestion

modules/nf-core/atlas/pmd/main.nf
```diff
@@ -17,7 +17,7 @@ process ATLAS_PMD {
     tuple val(meta), path("*_PMD_input_Exponential.txt"), emit: exponential
     tuple val(meta), path("*_PMD_Table_counts.txt")     , emit: counts
     tuple val(meta), path("*_PMD_Table.txt")            , emit: table
-    path "versions.yml"                                 , emit: versions
+    tuple val("${task.process}"), val('atlas'), eval('atlas | sed -e "2!d;s/^[ \\t]*Atlas //"'), emit: versions_atlas, topic: versions
```
LouisLeNezet: You can simplify even more with
```suggestion
    tuple val("${task.process}"), val('atlas'), eval('atlas | sed -e "2!d;s/.*Atlas //"'), emit: versions_atlas, topic: versions
```

LouisLeNezet: The backslash are to be avoided.
Is what I proposed working ?

garciafranciscomartn: No, linting verification fails when doing it, I needed to revert changes

PR #10481, opened 2026-03-11
Migrate argnorm to topic channels

garciafranciscomartn:
Closes nf-core/modules#10192


fmalmeida:
Hi @garciafranciscomartn ,
Normally we keep one module change per PR so we can clearly separate and debug the automated tests of each

fmalmeida:
The issue being closed mentions argnorm, but we can see more files changed

garciafranciscomartn (commented):
remove files from other modules

akaviaLab:
Hi
You shouldn't delete, the files from other modules.
Try reverting the commits that touched other modules, so the changelog won't include that. We should only see the files under angrom in the Changes tab.
Failing that, you could make a new branch and copy the angrom changes over to it (like cherry picking)

modules/nf-core/argnorm/meta.yml
```diff
@@ -52,14 +52,28 @@ output:
           description: Normalized argNorm output
           pattern: "*.tsv"
           ontologies:
-            - edam: "http://edamontology.org/format_3475" # TSV
+            - edam: "http://edamontology.org/format_3475"
```
fmalmeida: Can you re-add the comments of the ontologies?

akaviaLab:
I agree with the previous comment that you should add the commented tags to the onthologies. It should also still require updating to the latest branch before you try merging.
If you do all of these, it looks good to me, so provisionally approved.

fmalmeida (approved):
LGTM

PR #10549, opened 2026-03-11
nf-core/mag local catpack subworkflow extraction

mberacochea:
We needed this nf-core/mag local subworkflow (https://github.com/nf-core/mag/blob/main/subworkflows/local/catpack/main.nf) in nf-core/seqsubmit so I've tried to generalize it a bit (not too much) so we can re-use it.

I'll probably add 1 more unit test, the download catpack step is not being executed in any of the current ones.

I'm not too sure about the structure as in nf-core/mag has a few params which I tried to accomodate here.

## PR checklist

This subworkflow will be used here - https://github.com/nf-core/seqsubmit/issues/11

- [ ] Add a resource `label`
  - For subworkflows:

### DIFF
--- a/modules/nf-core/bismark/coverage2cytosine/main.nf
+++ b/modules/nf-core/bismark/coverage2cytosine/main.nf
@@ -16,7 +16,7 @@ process BISMARK_COVERAGE2CYTOSINE {
     tuple val(meta), path("*.cov.gz")                      , emit: coverage,  optional: true
     tuple val(meta), path("*report.txt.gz")                , emit: report
     tuple val(meta), path("*cytosine_context_summary.txt") , emit: summary
-    path  "versions.yml"                                   , emit: versions
+    tuple val("${task.process}"), val('bismark'), eval("bismark --version 2>&1 | sed -n 's/^.*Bismark Version: v//p'"), emit: versions_bismark, topic: versions
 
     when:
     task.ext.when == null || task.ext.when
@@ -31,11 +31,6 @@ process BISMARK_COVERAGE2CYTOSINE {
         --output ${prefix} \\
         --gzip \\
         ${args}
-
-    cat <<-END_VERSIONS > versions.yml
-    "${task.process}":
-        bismark: \$(echo \$(bismark -v 2>&1) | sed 's/^.*Bismark Version: v//; s/Copyright.*\$//')
-    END_VERSIONS
     """
 
     stub:
@@ -44,10 +39,5 @@ process BISMARK_COVERAGE2CYTOSINE {
     echo | gzip > ${prefix}.cov.gz
     echo | gzip > ${prefix}.report.txt.gz
     touch ${prefix}.cytosine_context_summary.txt
-
-    cat <<-END_VERSIONS > versions.yml
-    "${task.process}":
-        bismark: \$(echo \$(bismark -v 2>&1) | sed 's/^.*Bismark Version: v//; s/Copyright.*\$//')
-    END_VERSIONS
     """
 }
--- a/modules/nf-core/bismark/coverage2cytosine/meta.yml
+++ b/modules/nf-core/bismark/coverage2cytosine/meta.yml
@@ -19,7 +19,8 @@ tools:
       homepage: https://github.com/FelixKrueger/Bismark
       documentation: https://github.com/FelixKrueger/Bismark/tree/master/Docs
       doi: 10.1093/bioinformatics/btr167
-      licence: ["GPL-3.0-or-later"]
+      licence:
+        - "GPL-3.0-or-later"
       identifier: biotools:bismark
 input:
   - - meta:
@@ -87,13 +88,27 @@ output:
           description: Cyotosine context summary report.
           pattern: "*cytosine_context_summary.txt"
           ontologies: []
+  versions_bismark:
+    - - ${task.process}:
+          type: string
+          description: The name of the process
+      - bismark:
+          type: string
+          description: The name of the tool
+      - "bismark --version 2>&1 | sed -n 's/^.*Bismark Version: v//p'":
+          type: eval
+          description: The expression to obtain the version of the tool
+topics:
   versions:
-    - versions.yml:
-        type: file
-        description: File containing software versions
-        pattern: "versions.yml"
-        ontologies:
-          - edam: http://edamontology.org/format_3750 # YAML
+    - - ${task.process}:
+          type: string
+          description: The name of the process
+      - bismark:
+          type: string
+          description: The name of the tool
+      - "bismark --version 2>&1 | sed -n 's/^.*Bismark Version: v//p'":
+          type: eval
+          description: The expression to obtain the version of the tool
 authors:
   - "@ewels"
 maintainers:
--- a/modules/nf-core/bismark/coverage2cytosine/tests/main.nf.test
+++ b/modules/nf-core/bismark/coverage2cytosine/tests/main.nf.test
@@ -16,7 +16,7 @@ nextflow_process {
                 script "../../genomepreparation/main.nf"
                 process {
                     """
-                    input[0] = Channel.of([
+                    input[0] = channel.of([
                         [ id:'test' ], // meta map
                         file(params.modules_testdata_base_path + 'genomics/sarscov2/genome/genome.fasta', checkIfExists: true)
                     ])
@@ -28,9 +28,9 @@ nextflow_process {
                 script "../../methylationextractor/main.nf"
                 process {
                     """
-                    input[0] = Channel.of([
-                                [ id:'test', single_end:false ],
-                                file(params.modules_testdata_base_path + 'genomics/sarscov2/illumina/bam/test.paired_end.methylated.bam', checkIfExists: true)
+                    input[0] = channel.of([
+                        [ id:'test', single_end:false ],
+                        file(params.modules_testdata_base_path + 'genomics/sarscov2/illumina/bam/test.paired_end.methylated.bam', checkIfExists: true)
                     ])
                     input[1] = BISMARK_GENOMEPREPARATION.out.index
                     """
@@ -44,7 +44,10 @@ nextflow_process {
             process {
                 """
                 input[0] = BISMARK_METHYLATIONEXTRACTOR.out.coverage
-                input[1] = input[1] = Channel.of([ [ id:'sarscov2'], file(params.modules_testdata_base_path + 'genomics/sarscov2/genome/genome.fasta', checkIfExists: true) ])
+                input[1] = channel.of([
+                    [ id:'sarscov2'],
+                    file(params.modules_testdata_base_path + 'genomics/sarscov2/genome/genome.fasta', checkIfExists: true)
+                ])
                 input[2] = BISMARK_GENOMEPREPARATION.out.index
                 """
             }
@@ -53,10 +56,7 @@ nextflow_process {
         then {
             assert process.success
             assertAll(
-                { assert snapshot(
-                    process.out,
-                    path(process.out.versions[0]).yaml
-                ).match() }
+                { assert snapshot(sanitizeOutput(process.out)).match() }
             )
         }
     }
@@ -68,7 +68,10 @@ nextflow_process {
             process {
                 """
                 input[0] = BISMARK_METHYLATIONEXTRACTOR.out.coverage
-                input[1] = input[1] = Channel.of([ [ id:'sarscov2'], file(params.modules_testdata_base_path + 'genomics/sarscov2/genome/genome.fasta', checkIfExists: true) ])
+                input[1] = channel.of([
+                    [ id:'sarscov2'],
+                    file(params.modules_testdata_base_path + 'genomics/sarscov2/genome/genome.fasta', checkIfExists: true)
+                ])
                 input[2] = BISMARK_GENOMEPREPARATION.out.index
                 """
             }
@@ -77,10 +80,7 @@ nextflow_process {
         then {
             assert process.success
             assertAll(
-                { assert snapshot(
-                    process.out,
-                    path(process.out.versions[0]).yaml
-                ).match() }
+                { assert snapshot(sanitizeOutput(process.out)).match() }
             )
         }
     }
--- a/modules/nf-core/bismark/coverage2cytosine/tests/main.nf.test.snap
+++ b/modules/nf-core/bismark/coverage2cytosine/tests/main.nf.test.snap
@@ -2,30 +2,6 @@
     "sarscov2 genome [fasta] | sarscov2 paired-end methylated [bam]": {
         "content": [
             {
-                "0": [
-                    
-                ],
-                "1": [
-                    [
-                        {
-                            "id": "test",
-                            "single_end": false
-                        },
-                        "test.CpG_report.txt.gz:md5,0b83c760c7b3df21a320f054bbb865fb"
-                    ]
-                ],
-                "2": [
-                    [
-                        {
-                            "id": "test",
-                            "single_end": false
-                        },
-                        "test.cytosine_context_summary.txt:md5,5ed486bfa844a5e476a7ab054a7a8d6b"
-                    ]
-                ],
-                "3": [
-                    "versions.yml:md5,98c6acbd0cc567eb739d17ecbbdd6711"
-                ],
                 "coverage": [
                     
                 ],
@@ -47,49 +23,24 @@
                         "test.cytosine_context_summary.txt:md5,5ed486bfa844a5e476a7ab054a7a8d6b"
                     ]
                 ],
-                "versions": [
-                    "versions.yml:md5,98c6acbd0cc567eb739d17ecbbdd6711"
+                "versions_bismark": [
+                    [
+                        "BISMARK_COVERAGE2CYTOSINE",
+                        "bismark",
+                        "0.25.1"
+                    ]
                 ]
-            },
-            {
-                "BISMARK_COVERAGE2CYTOSINE": {
-                    "bismark": "0.25.1"
-                }
             }
         ],
+        "timestamp": "2026-03-12T15:46:17.269623714",
         "meta": {
-            "nf-test": "0.9.2",
-            "nextflow": "25.04.6"
-        },
-        "timestamp": "2025-09-25T20:48:11.823558427"
+            "nf-test": "0.9.4",
+            "nextflow": "25.10.4"
+        }
     },
     "sarscov2 genome [fasta] | sarscov2 paired-end methylated [bam] | stub": {
         "content": [
             {
-                "0": [
-                    
-                ],
-                "1": [
-                    [
-                        {
-                            "id": "test",
-                            "single_end": false
-                        },
-                        "test.report.txt.gz:md5,68b329da9893e34099c7d8ad5cb9c940"
-                    ]
-                ],
-                "2": [
-                    [
-                        {
-                            "id": "test",
-                            "single_end": false
-                        },
-                        "test.cytosine_context_summary.txt:md5,d41d8cd98f00b204e9800998ecf8427e"
-                    ]
-                ],
-                "3": [
-                    "versions.yml:md5,98c6acbd0cc567eb739d17ecbbdd6711"
-                ],
                 "coverage": [
                     
                 ],
@@ -111,20 +62,19 @@
                         "test.cytosine_context_summary.txt:md5,d41d8cd98f00b204e9800998ecf8427e"
                     ]
                 ],
-                "versions": [
-                    "versions.yml:md5,98c6acbd0cc567eb739d17ecbbdd6711"
+                "versions_bismark": [
+                    [
+                        "BISMARK_COVERAGE2CYTOSINE",
+                        "bismark",
+                        "0.25.1"
+                    ]
                 ]
-            },
-            {
-                "BISMARK_COVERAGE2CYTOSINE": {
-                    "bismark": "0.25.1"
-                }
             }
         ],
+        "timestamp": "2026-03-12T15:46:31.405076755",
         "meta": {
-            "nf-test": "0.9.2",
-            "nextflow": "25.04.6"
-        },
-        "timestamp": "2025-09-25T20:48:23.237287933"
+            "nf-test": "0.9.4",
+            "nextflow": "25.10.4"
+        }
     }
 }
--- a/modules/nf-core/blobtk/plot/main.nf
+++ b/modules/nf-core/blobtk/plot/main.nf
@@ -21,9 +21,11 @@ process BLOBTK_PLOT {
     path(local_path)                // Genuine path location must be a path.
     val(online_path)                // HTTPS location needs to remain a value
     val extra_args                  // In format [name: "", args: ""]
+    val format                      // Output format, e.g. png or svg
 
     output:
-    tuple val(meta), path("*.png"), emit: png
+    tuple val(meta), path("*.png"), optional: true, emit: png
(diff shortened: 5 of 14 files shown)


subworkflows/nf-core/fasta_classify_catpack/main.nf
```diff
@@ -0,0 +1,134 @@
...
+    //
+
+    // Handle pre-built db: untar if compressed, or use directory directly
+    ch_cat_db
+        .branch { _meta, db ->
+            tar: db.name.endsWith('.tar.gz')
+            dir: true
+        }
+        .set { ch_cat_db_input }
+
```
prototaxites:
```suggestion
    ch_cat_db
        .branch { _meta, db ->
            tar: db.name.endsWith('.tar.gz')
            dir: db.isDirectory()
            other: true
         }
        .set { ch_cat_db_input }

ch_cat_db.other.subscribe { db -> 
    error("Error: A DB was provided to FASTA_CLASSIFY_CATPACK that is not a `.tar.gz` or a directory!")
}
```

Maybe we can explicitly capture the case we get unexpected inputs and exit?

subworkflows/nf-core/fasta_classify_catpack/main.nf
```diff
@@ -0,0 +1,134 @@
...
+    CAT_DB_UNTAR(ch_cat_db_input.tar)
+    ch_versions = ch_versions.mix(CAT_DB_UNTAR.out.versions_untar.first())
```
prototaxites: How easy would it be to future-proof this subWF and port the modules to use topics?

subworkflows/nf-core/fasta_classify_catpack/main.nf
```diff
@@ -0,0 +1,134 @@
...
+    CATPACK_PREPARE(
+        CATPACK_DOWNLOAD.out.fasta,
+        CATPACK_DOWNLOAD.out.names.map   { _meta, names -> names },
+        CATPACK_DOWNLOAD.out.nodes.map   { _meta, nodes -> nodes },
+        CATPACK_DOWNLOAD.out.acc2tax.map { _meta, acc2tax -> acc2tax },
+    )
+    ch_versions = ch_versions.mix(CATPACK_PREPARE.out.versions_catpack.first())
+
+    // Combine db sources - one of these channels will be empty depending on inputs
+    ch_db       = ch_prepared_from_dir.db.mix(CATPACK_PREPARE.out.db).first()
```
prototaxites, line 71:
```suggestion
    ch_db       = ch_prepared_from_dir.db.mix(CATPACK_PREPARE.out.db).first()
    
    ch_db.collect().subscribe { db_list -> 
        if(db_list.size() > 1) {
            error("Error: More than 1 DB has been provided to FASTA_CLASSIFY_CATPACK! Probably you supplied a DB and downloaded it too.")
        }
    }
```

Error handling for >1 db?

subworkflows/nf-core/fasta_classify_catpack/main.nf
```diff
@@ -0,0 +1,134 @@
...
+    ch_taxonomy = ch_prepared_from_dir.taxonomy.mix(CATPACK_PREPARE.out.taxonomy).first()
```
prototaxites, line 72: And similar error handling here as above?

subworkflows/nf-core/fasta_classify_catpack/main.nf
```diff
@@ -0,0 +1,134 @@
...
+    //
+    // Bin taxonomic classification
+    //
+
+    CATPACK_BINS(
+        ch_bins,
+        ch_db,
+        ch_taxonomy,
+        [[:], []],
+        [[:], []],
```
prototaxites, line 83: Would it ever be useful to be able to specify these as inputs?

subworkflows/nf-core/fasta_classify_catpack/main.nf
```diff
@@ -0,0 +1,134 @@
...
+        bin_suffix,
+    )
+    ch_versions = ch_versions.mix(CATPACK_BINS.out.versions_catpack.first())
+
+    CATPACK_ADDNAMES_BINS(CATPACK_BINS.out.bin2classification, ch_taxonomy)
+    ch_versions = ch_versions.mix(CATPACK_ADDNAMES_BINS.out.versions_catpack.first())
+
+    ch_summarise_bins = channel.empty()
+    if (!allow_unofficial_lineages) {
```
prototaxites: How does this interact with the unofficial lineage argument? A user would have to specify both the `--only_official` arg in a config and ensure that this flag is set too.

Maybe this param could just be re-named `run_catpack_summarise` with a note in the meta that this is only functional if `--only-official` is set? 

Or alternatively make `--only-official` a mandatory input to CATPACK_ADDNAMES and then we use this boolean here and as an input?

subworkflows/nf-core/fasta_classify_catpack/main.nf
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    //
+    // Unbinned contigs taxonomic classification (optional)
+    //
+
+    ch_unbins_addnames  = channel.empty()
+    ch_summarise_unbins = channel.empty()
+
+    if (classify_unbinned) {
```
prototaxites: if ch_unbins was `channel.empty()` the following would all be skipped, so you don't need this flag?

subworkflows/nf-core/fasta_classify_catpack/meta.yml
```diff
@@ -0,0 +1,118 @@
...
+        trigger automatic database download using ch_cat_db_download_id.
+        Structure: [ val(meta), path(db) ]
+      ontologies:
+        - edam: "http://edamontology.org/data_1049"
+  - ch_cat_db_download_id:
+      type: string
+      description: |
+        Channel containing the database identifier to download via CATPACK_DOWNLOAD (e.g. 'nr').
+        Only used when ch_cat_db is channel.empty(). Provide channel.empty() when supplying a
+        pre-built database via ch_cat_db.
```
prototaxites: > Only used when ch_cat_db is channel.empty(). Provide channel.empty() when supplying a pre-built database via ch_cat_db.

There's no actual code check that you can't supply both, and if you do, you will end up with a channel of 2 dbs.

prototaxites (commented):
Hi @mberacochea, looking good - some thoughts, mostly about error handling and process inputs.

subworkflows/nf-core/fasta_classify_catpack/meta.yml
```diff
@@ -0,0 +1,118 @@
...
+        - edam: "http://edamontology.org/format_3475"
+  - unbinned_summary:
+      type: file
+      description: |
+        Summary of unbinned contig classifications. Empty channel when classify_unbinned is
+        false or allow_unofficial_lineages is true.
+        Structure: [ val(meta), path(txt) ]
+      pattern: "*.txt"
+      ontologies:
+        - edam: "http://edamontology.org/format_3475"
```
vagkaratzas: Can you also add comments at the end of the edam lines?
```suggestion
        - edam: "http://edamontology.org/format_3475" # Textual format
```

subworkflows/nf-core/fasta_classify_catpack/main.nf
```diff
@@ -0,0 +1,134 @@
...
+    take:
+    ch_bins                   // channel: [ val(meta), path(fasta) ] - binned MAGs/contigs
+    ch_unbins                 // channel: [ val(meta), path(fasta) ] - unbinned contigs
+    ch_cat_db                 // channel: [ val(meta), path(db) ] - pre-built db as directory (with db/ and tax/ subdirs) or .tar.gz
+                              //          provide channel.empty() to trigger automatic download via ch_cat_db_download_id
+    ch_cat_db_download_id     // channel: [ val(meta), val(db_id) ] - db ID for CATPACK_DOWNLOAD (e.g. 'nr')
+                              //          provide channel.empty() if supplying a pre-built db via ch_cat_db
+    classify_unbinned         // val: boolean - whether to classify unbinned contigs
+    allow_unofficial_lineages // val: boolean - whether to allow unofficial lineages in the summarise step
+    bin_suffix                // val: string - file extension of bin FASTA files (e.g. '.fa')
```
vagkaratzas: Can we change this to not include the dot? e.g. 'fa'

subworkflows/nf-core/fasta_classify_catpack/main.nf
mberacochea: I like this, better to fail explicitly

subworkflows/nf-core/fasta_classify_catpack/tests/main.nf.test
```diff
@@ -0,0 +1,119 @@
...
+            }
+        }
+    }
+
+    test("sarscov2 - bins only - summarise") {
+
+        when {
+            workflow {
+                """
+                input[0] = Channel.of([
```
vagkaratzas:
```suggestion
                input[0] = channel.of([
```
strict syntax throughout please

subworkflows/nf-core/fasta_classify_catpack/main.nf
dialvarezs: All the modules used in this subwf already use topics, so I guess it's already done.
Maybe @mberacochea you can just drop the "manual" version collection.

mberacochea: Done :)

subworkflows/nf-core/fasta_classify_catpack/main.nf
mberacochea: Maybe, I was thinking of leaving this out for this iteration.

subworkflows/nf-core/fasta_classify_catpack/main.nf
mberacochea: Great point, I've changed this a bit following your first suggestion.

subworkflows/nf-core/fasta_classify_catpack/main.nf
mberacochea: Indeed, removed.

subworkflows/nf-core/fasta_classify_catpack/meta.yml
mberacochea: Done

subworkflows/nf-core/fasta_classify_catpack/main.nf
mberacochea: Not really, "CATPACK_BINS" uses that as an argument

subworkflows/nf-core/fasta_classify_catpack/tests/main.nf.test
mberacochea: Done

subworkflows/nf-core/fasta_classify_catpack/main.nf
prototaxites: Fair!

subworkflows/nf-core/fasta_classify_catpack/main.nf
```diff
@@ -0,0 +1,128 @@
...
+        bin_suffix,
+    )
+
+    CATPACK_ADDNAMES_BINS(CATPACK_BINS.out.bin2classification, ch_taxonomy)
+
+    ch_summarise_bins = channel.empty()
+    if (run_summarise) {
+        CATPACK_SUMMARISE_BINS(CATPACK_ADDNAMES_BINS.out.txt, [[:], []])
+        ch_summarise_bins = CATPACK_SUMMARISE_BINS.out.txt
+    }
```
prototaxites: Alternative, for your delectation - avoid having to pre-declare empty channels:


```suggestion
        CATPACK_SUMMARISE_BINS(CATPACK_ADDNAMES_BINS.out.txt.filter { it -> run_summarise }, [[:], []])
        ch_summarise_bins = CATPACK_SUMMARISE_BINS.out.txt
```

subworkflows/nf-core/fasta_classify_catpack/tests/main.nf.test.snap
```diff
@@ -0,0 +1,227 @@
...
+                ],
+                "bat_summary": [
+                    
+                ],
+                "unbinned_classification": [
+                    
+                ],
+                "unbinned_summary": [
+                    
+                ],
```
prototaxites, line 88: Should these be empty?

subworkflows/nf-core/fasta_classify_catpack/main.nf
```diff
@@ -0,0 +1,128 @@
...
+        CATPACK_SUMMARISE_UNBINS(ch_unbin_input.names, ch_unbin_input.contigs)
+        ch_summarise_unbins = CATPACK_SUMMARISE_UNBINS.out.txt
+    }
+
+    emit:
+    bat_classification      = CATPACK_ADDNAMES_BINS.out.txt   // channel: [ val(meta), path(txt) ]
+    bat_summary             = ch_summarise_bins               // channel: [ val(meta), path(txt) ]
+    unbinned_classification = CATPACK_ADDNAMES_UNBINS.out.txt // channel: [ val(meta), path(txt) ]
+    unbinned_summary        = ch_summarise_unbins             // channel: [ val(meta), path(txt) ]
+    versions                = channel.topic('versions')       // channel: versions
```
vagkaratzas: I dont think we use this

vagkaratzas: 1. if there is at least one of the old versions output, we do it the old way
2. else, completely remove it from here (it is handled only at the pipeline level)
```suggestion
```
which I guess is the case here

mberacochea: ahh, live and learn. We can remove it yup.

subworkflows/nf-core/fasta_classify_catpack/tests/main.nf.test.snap
mberacochea: image

let me check, I did several iterations of the tests so this could be a mistake

subworkflows/nf-core/fasta_classify_catpack/main.nf
```diff
@@ -0,0 +1,124 @@
+/*
+ * CAT/BAT/RAT: tools for taxonomic classification of contigs and metagenome-assembled genomes (MAGs)
+ */
+include { CATPACK_ADDNAMES as CATPACK_ADDNAMES_BINS     } from '../../../modules/nf-core/catpack/addnames/main'
+include { CATPACK_ADDNAMES as CATPACK_ADDNAMES_UNBINS   } from '../../../modules/nf-core/catpack/addnames/main'
+include { CATPACK_BINS                                  } from '../../../modules/nf-core/catpack/bins/main'
+include { CATPACK_CONTIGS as CATPACK_UNBINS             } from '../../../modules/nf-core/catpack/contigs/main'
```
ochkalova: I vote for renaming those "unbins" to contigs in this subworkflow. Less confusing, and name inherited from catpack_contigs module

When I see unbins, I think "waaat?" 🫤

subworkflows/nf-core/fasta_classify_catpack/main.nf
```diff
@@ -0,0 +1,124 @@
...
+workflow FASTA_CLASSIFY_CATPACK {
+
+    take:
+    ch_bins               // channel: [ val(meta), path(fasta) ] - binned MAGs/contigs
+    ch_unbins             // channel: [ val(meta), path(fasta) ] - unbinned contigs; provide channel.empty() to skip unbinned classification
+    ch_cat_db             // channel: [ val(meta), path(db) ] - pre-built db as directory (with db/ and tax/ subdirs) or .tar.gz
+                          //          provide channel.empty() to trigger automatic download via ch_cat_db_download_id
+    ch_cat_db_download_id // channel: [ val(meta), val(db_id) ] - db ID for CATPACK_DOWNLOAD (e.g. 'nr')
+                          //          provide channel.empty() if supplying a pre-built db via ch_cat_db
+                          //          supplying both ch_cat_db and ch_cat_db_download_id will cause a runtime error
```
ochkalova, line 23: Is it possible to print error and exit in this case instead of runtime error?

subworkflows/nf-core/fasta_classify_catpack/main.nf
prototaxites: 🙏🙏🙏🙏🙏

Completely agree, I've always found the name "unbins" very strange

subworkflows/nf-core/fasta_classify_catpack/main.nf
```diff
@@ -0,0 +1,124 @@
...
+        ch_taxonomy,
+        [[:], []],
+        [[:], []],
+        bin_suffix,
+    )
+
+    CATPACK_ADDNAMES_BINS(CATPACK_BINS.out.bin2classification, ch_taxonomy)
+
+        CATPACK_SUMMARISE_BINS(CATPACK_ADDNAMES_BINS.out.txt.filter { it -> run_summarise }, [[:], []])
+        ch_summarise_bins = CATPACK_SUMMARISE_BINS.out.txt
```
ochkalova: Looks complicated, why not use `if` here? Also looks like there is an extra tab before this lines

subworkflows/nf-core/fasta_classify_catpack/main.nf
```diff
@@ -0,0 +1,124 @@
...
+    ch_prepared_from_dir.db.combine(CATPACK_PREPARE.out.db).subscribe {
+        error("Error: Both a pre-built DB and a download ID were provided to FASTA_CLASSIFY_CATPACK! Provide only one via ch_cat_db or ch_cat_db_download_id.")
+    }
+
+    ch_db       = ch_prepared_from_dir.db.mix(CATPACK_PREPARE.out.db).first()
+    ch_taxonomy = ch_prepared_from_dir.taxonomy.mix(CATPACK_PREPARE.out.taxonomy).first()
+
+    //
+    // Bin taxonomic classification
+    //
```
ochkalova, line 76: I guess this will be skipped as well if channel is empty
Maybe using the same comment as before CATPACK_UNBINS?

subworkflows/nf-core/fasta_classify_catpack/main.nf
```diff
@@ -0,0 +1,124 @@
...
+        ch_unbin_input = CATPACK_ADDNAMES_UNBINS.out.txt
+            .join(ch_unbins)
+            .multiMap { meta, names, contigs ->
+                names:   [meta, names]
+                contigs: [meta, contigs]
+            }
+
+        CATPACK_SUMMARISE_UNBINS(ch_unbin_input.names, ch_unbin_input.contigs)
+        ch_summarise_unbins = CATPACK_SUMMARISE_UNBINS.out.txt
+    }
```
ochkalova, line 120: And here you use `if`... Why? 👀

subworkflows/nf-core/fasta_classify_catpack/main.nf
```diff
@@ -0,0 +1,124 @@
...
+
+    emit:
+    bat_classification      = CATPACK_ADDNAMES_BINS.out.txt   // channel: [ val(meta), path(txt) ]
+    bat_summary             = ch_summarise_bins               // channel: [ val(meta), path(txt) ]
+    unbinned_classification = CATPACK_ADDNAMES_UNBINS.out.txt // channel: [ val(meta), path(txt) ]
+    unbinned_summary        = ch_summarise_unbins             // channel: [ val(meta), path(txt) ]
```
ochkalova: Add CAT_pack output before ADDNAMES please

subworkflows/nf-core/fasta_classify_catpack/main.nf
ochkalova: Could you please explain this choice? I don't like this syntax because it's less readable. This `.filter()` is easier to loose in opposite to `if (run_summarise)`

subworkflows/nf-core/fasta_classify_catpack/main.nf
ochkalova: I saw it was a suggestion from @prototaxites, so @mberacochea ignore this for now

subworkflows/nf-core/fasta_classify_catpack/tests/main.nf.test.snap
mberacochea: Yes, for this one it is fine that these are empty (no unbins and run_summarise is false)

subworkflows/nf-core/fasta_classify_catpack/main.nf
mberacochea: yes, there is a bit of code that does that:
```nextflow
    // Combine db sources - one of these channels will be empty depending on inputs
    // Guard: fail if both ch_cat_db and ch_cat_db_download_id are provided simultaneously.
    // .combine() only emits when both channels have at least one element.
    ch_prepared_from_dir.db.combine(CATPACK_PREPARE.out.db).subscribe {
        error("Error: Both a pre-built DB and a download ID were provided to FASTA_CLASSIFY_CATPACK! Provide only one via ch_cat_db or ch_cat_db_download_id.")
    }
```

subworkflows/nf-core/fasta_classify_catpack/main.nf
prototaxites: I often use this approach as it means you never need to catch non-existent channels that get declared in an `if` block, as you just empty the channel with a boolean. IMO it's cleaner, though in this case I appreciate it doesn't really achieve anything as there's only one variable to catch.

subworkflows/nf-core/fasta_classify_catpack/main.nf
mberacochea: I've removed the tab, from @prototaxites comment I got that this is the preferred way (instead of nesting code).

prototaxites: It's not necessarily "the" preferred way as much as /my/ preferred way 😇 You are very welcome to ignore the suggestion if you dislike it

mberacochea: Well, I'll revert this.. I think the code is easier to read with the ifs otherwise the input/flag is kind of hidden.

subworkflows/nf-core/fasta_classify_catpack/main.nf
mberacochea: Reverted the previuos one to a good ol' `if`

subworkflows/nf-core/fasta_classify_catpack/main.nf
mberacochea: I think it makes reading this bit easier, unbin across the whole subworkflow

dialvarezs: > I've always found the name "unbins" very strange

Yeah, it's a bit strange, but it's a shortened form of “unbinned contig” in the context of mag. I assume that was the reasoning behind it, right @jfy133?

However, since this is meant to be a more general subworkflow, I agree that we should use “contigs” instead.

subworkflows/nf-core/fasta_classify_catpack/main.nf
mberacochea: done

subworkflows/nf-core/fasta_classify_catpack/main.nf
mberacochea: fair enough, I've renamed them to contings 👍🏽 [#c925c9471](https://github.com/nf-core/modules/pull/10549/changes/c925c9471a0b0e506ed4f7b5e11d42af0f1018ae)

PR #10574, opened 2026-03-11
Migration to topic channels: anndata/getsize

### LINKED ISSUE #10181: Migration to topic channels: anndata/getsize
This issue tracks migration of `anndata/getsize` to topic channels.

Parent meta issue: #9978
Migration guide: https://nf-co.re/docs/tutorials/migrate_to_topics/update_modules

Checklist:
- [ ] Module migrated to topic channels
- [ ] Tests passing
- [ ] meta.yml updated
- [ ] nf-core lint passing

juliaapolonio:
Closes #10181 

Things modified and why:
meta.yml, main.nf and main.nf.test - default changes to migrate to topic channels
getsize.py - shebang removed to fix nf-test error

### DIFF
--- a/modules/nf-core/amplify/predict/main.nf
+++ b/modules/nf-core/amplify/predict/main.nf
@@ -2,7 +2,6 @@ process AMPLIFY_PREDICT {
     tag "$meta.id"
     label 'process_single'
 
-    // WARN: Version information not provided by tool on CLI. Please update version string below when bumping container versions.
     conda "${moduleDir}/environment.yml"
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
         'https://depot.galaxyproject.org/singularity/amplify:2.0.0--py36hdfd78af_1':
@@ -14,7 +13,7 @@ process AMPLIFY_PREDICT {
 
     output:
     tuple val(meta), path('*.tsv'), emit: tsv
-    path "versions.yml"           , emit: versions
+    tuple val("${task.process}"), val('AMPlify'), eval( "AMPlify --help 2>&1 | sed -n 's/AMPlify v//p'" ), emit: versions_amplify, topic: versions
 
     when:
     task.ext.when == null || task.ext.when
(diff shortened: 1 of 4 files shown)


SPPearce:
@nf-core-bot fix linting

juliaapolonio:
> Seems a bit odd to have to remove the shebang?

Yes it is... nf-test was not happy and the only solution I found was to remove the shebang. Now I think it is failing prek because of that. 
Before removing the shebang I was having an error like "this is not a bash script". Have you had a similar issue with a module from an R/python tool so I can have a look at it?

juliaapolonio:
Hi, Simon! The shebang is back, I didn't realize that migrating with a template was different. Can you have a look?

modules/nf-core/anndata/getsize/tests/main.nf.test
```diff
@@ -52,10 +49,7 @@ nextflow_process {
         then {
             assertAll(
                 { assert process.success },
-                { assert snapshot(
-                    process.out,
-                    path(process.out.versions[0]).yaml
```
famosab, line 57: I guess keeping this line once would be nice to really spot if the version number is still accessed correctly :)

juliaapolonio:
Hi @famosab, thanks for the suggestion. I implemented it but Docker and  Singularity checks are failing because of lack of disk space. Is there something I can do about that?

SPPearce:
> Hi @famosab, thanks for the suggestion. I implemented it but Docker and Singularity checks are failing because of lack of disk space. Is there something I can do about that?

Updating the branch to the most recent changes (as I've just done) should hopefully fix that.

modules/nf-core/anndata/getsize/tests/main.nf.test
```diff
@@ -84,6 +82,7 @@ nextflow_process {
                     process.out,
                     path(process.out.versions[0]).yaml
                 ).match() }
+                { assert snapshot(sanitizeOutput(process.out)).match() },
```
famosab:
```suggestion
{ assert snapshot(sanitizeOutput(
process.out,
path(process.out.versions[0]).yaml)
).match() }
```
We do not need both snapshots though :)

LouisLeNezet: Done

modules/nf-core/anndata/getsize/tests/main.nf.test
LouisLeNezet: Done

LouisLeNezet:
Thanks for your contribution !
This PR is now ready and will be merged to clean up the repository.

PR #10587, opened 2026-03-12
Remove SAMTOOLS_INDEX from BAM_SORT_STATS_SAMTOOLS workflow

### LINKED ISSUE #10390: Skip indexing for bam_sort_stats_samtools
Samtools sort has been updated a while ago so that it can automatically add the index (crai/bai) of a bam file. I feel like the index step in the `BAM_SORT_STATS_SAMTOOLS` subworkflow is unneeded and can be replaced by taking an extra argument of the type of index you want and simply use the samtools sort to directly sort and index sam/bam files.

yinshiyi:
Removes `SAMTOOLS_INDEX` from `BAM_SORT_STATS_SAMTOOLS` and lets `SAMTOOLS_SORT` write the index in the same pass (`--write-index`), saving one process per sample.
Ideation from https://github.com/nf-core/modules/issues/10390 and https://github.com/nf-core/modules/pull/8895.

### Changes

- `bam_sort_stats_samtools`
  - `SAMTOOLS_INDEX` dropped; index comes from `SAMTOOLS_SORT.out.index`
  - **Breaking:** new mandatory take `index_format` (`'bai'`, `'csi'` or `'crai'`), passed straight to `SAMTOOLS_SORT`. Nothing is hardcoded, so large genomes can request `csi` (review feedback from @SPPearce).
  - `meta.yml` documents the new input; tests pass `'bai'`, a `csi` test case was added, `samtools/index` tag removed.
- All nine callers updated to pass `'bai'` so their behaviour is unchanged: `fastq_align_bwa`, `fastq_align_bowtie2`, `fastq_align_hisat2`, `fastq_align_chromap`, `fastq_align_mapad`, `fastq_align_dedup_bwamem`, `fastq_align_star`, `fastq_align_bamcmp_bwa`, `bam_dedup_umi`. Stale transitive `samtools/index` component/tag entries in those subworkflows were removed.
- Caller snapshots change because the sorted BAM header (`@PG` line) and hence the BAM and `.bai` md5s differ when the index is written by `samtools sort`. A few stale `samtools stats` md5s in untouched snapshots were refreshed along the way (CI reproduces them).
- `fastq_align_star` test configs used `params.outdir` in `publishDir` without defining it, which fails on CI's Nextflow; the line was removed.

## PR checklist

Closes #10390

- [x] Add a resource `label`
  - For subworkflows:
    - [x] `nf-core subworkflows test <SUBWORKFLOW> --profile docker` (via CI)
    - [x] `nf-core subworkflows test <SUBWORKFLOW> --profile conda` (via CI)


yinshiyi:
New shave off 2s
image

Old
image

SPPearce (changes requested):
Removing the module makes sense, but you shouldn't hardcode bai here. 
Large genomes can't use bai, they need to use csi files.
Now personally I think that samtools index should emit the bai, csi and crai indexes in one channel, not be split out into three.

yinshiyi:
> Removing the module makes sense, but you shouldn't hardcode bai here. Large genomes can't use bai, they need to use csi files. Now personally I think that samtools index should emit the bai, csi and crai indexes in one channel, not be split out into three.

I agree. 
if one needs to emit bai, csi and crai, then without using `samtools index`, the `samtool_sort` need to be run 3 times. 
I think the new samtool_sort, takes one val of `index_choice` and run it.
Please let me know if my understanding is correct.
Thank you.

ljwharbers:
I agree with not having bai as default. I would vote for to go with no default and people have to specify which they want.

yinshiyi:
> I agree with not having bai as default. I would vote for to go with no default and people have to specify which they want.

I am thinking about, but
```
workflow BAM_SORT_STATS_SAMTOOLS {
    take:
    ch_bam   // channel: [ val(meta), [ bam ] ]
    ch_fasta // channel: [ val(meta), path(fasta) ]
    index_choice // index choice for samtools sort, either 'bai' or 'csi'
```
This adding val will likely break many older calls

just leaving  `SAMTOOLS_SORT ( ch_bam, ch_fasta, '')` does not seem to generate index.

Any suggestions? thanks

ljwharbers:
Either of three possibilities in my eyes.
1. Old things will break and will need to be adjusted. (This also happened with samtools sort update)
2. Have a default (like you have now)
3. Have an if/else to check for the index value and run samtools index if its not present.

The third option makes little sense to me since then you might as well have a default value in the subworkflow. I dont hold a strong opinion for option 1 or 2 but in leaning towards 1.

Any suggestions @SPPearce ?

SPPearce:
If setting a default, csi is better I think, because all genomes support csi, but bai doesn't support really large ones.
But don't worry about adding a new parameter, people can update their pipelines when they incorporate changes.

yinshiyi:
> If setting a default, csi is better I think, because all genomes support csi, but bai doesn't support really large ones. But don't worry about adding a new parameter, people can update their pipelines when they incorporate changes.

Thank you!
I give it a shot with option 1 then. by adding a val input specific the type of index user require.
I likely need to update nf-test for this subworkflow also.

SPPearce:
FYI, I started on a plan to change SAMTOOLS_INDEX and SAMTOOLS_SORT to emit all the indexes into one channel, rather than splitting bai and csi seperately. 
Samtools index has the PR here: #10614.

ljwharbers:
Ah great, I see that it was merged. Maybe change this accordingly and then it looks good to me! @yinshiyi (would also need to fix the nf-tests I assume)

SPPearce:
Well, the plan here is to remove samtools index anyway ;) 
I haven't started on samtools sort yet.
There is also a PR to mass update all the samtools versions too.

ljwharbers:
Ahh yes of course, oops :')

SPPearce:
Anyone plan on finishing this off? Otherwise we can close it.

ljwharbers:
On it.

ljwharbers:
I've let claude do the updates on this so below the message. I did check the changes myself of course.

@SPPearce I picked this up and finished it off on this branch (description updated):

- merged master; `SAMTOOLS_INDEX` dropped, index comes from `SAMTOOLS_SORT.out.index`
- new mandatory take `index_format` (`'bai'`/`'csi'`/`'crai'`) instead of a hardcoded `bai`, as discussed above
- all nine callers (`fastq_align_bwa/bowtie2/hisat2/chromap/mapad/dedup_bwamem/star/bamcmp_bwa`, `bam_dedup_umi`) pass `'bai'` so their behaviour is unchanged; stale transitive `samtools/index` component/tag entries removed
- tests: added a `csi` case, snapshots regenerated (BAM/`.bai` md5s change because the sorted BAM `@PG` header now records `--write-index`)
- `fastq_align_star` test configs referenced an undefined `params.outdir` in `publishDir`, which broke on CI; removed

CI is green. Since I ended up writing most of it, a review from someone else would be good.

ljwharbers (approved):

PR #10614, opened 2026-03-12
Swap samtools index to emit just one channel

SPPearce:
Swapping samtools index to just emit one channel, rather than splitting bai and csi especially into two.
This stops the need to join the bai and csi channels back together afterwards.
bai indexes don't work for large genomes, they have to be csi indexed. Which means that many of the current subworkflows that are currently hardcoded to only use the bai will not work for these genomes.

Also I noticed that fastq_align_dedup_bwameth had a channel error, where the fastq files were being passed as separate parts of a tuple.


matthdsm:
Can you also update samtools to v1.23 so I don't have to re-test everything when merging with my bulk update PR?

maxulysse (approved):

PR #10683, opened 2026-03-12
Update pretext snapshot to 0.0.7

DLBPointon:
Update pretextsnapshot to 0.0.7
Added new input for the module, a custom order file
Update tests to include test the new custom order flag
Update test names now there's more than 2

### DIFF
--- a/modules/nf-core/pretextsnapshot/tests/main.nf.test.snap
+++ b/modules/nf-core/pretextsnapshot/tests/main.nf.test.snap
@@ -178,30 +178,15 @@
     },
     "jaGalFasc40_2 - 'full' snapshot pretextmap": {
         "content": [
+            [
+                [
+                    {
+                        "id": "test"
+                    },
+                    "test_FullMap.png"
+                ]
+            ],
             {
-                "0": [
-                    [
-                        {
-                            "id": "test"
-                        },
-                        "test_FullMap.png:md5,40ac67b97fe859769dd5af99c9899585"
-                    ]
-                ],
-                "1": [
-                    [
-                        "PRETEXTSNAPSHOT",
-                        "PretextSnapshot",
-                        "0.0.7"
-                    ]
-                ],
-                "image": [
-                    [
-                        {
-                            "id": "test"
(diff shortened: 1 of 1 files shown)


modules/nf-core/pretextsnapshot/tests/main.nf.test.snap
```diff
@@ -73,7 +73,7 @@
                     [
                         "PRETEXTSNAPSHOT",
                         "PretextSnapshot",
-                        "0.0.5"
+                        "0.0.6"
```
prototaxites: Version is wrong - either need to fix the tool release or temporarily fix the version like in purge_dups: https://github.com/nf-core/modules/blob/master/modules/nf-core/purgedups/purgedups/main.nf

modules/nf-core/pretextsnapshot/tests/main.nf.test.snap
```diff
@@ -84,6 +84,98 @@
...
+                            "test_scaffold_3.png:md5,d41d8cd98f00b204e9800998ecf8427e",
+                            "test_scaffold_4.png:md5,d41d8cd98f00b204e9800998ecf8427e"
+                        ]
+                    ]
+                ],
+                "1": [
+                    [
+                        "PRETEXTSNAPSHOT",
+                        "PretextSnapshot",
+                        "0.0.7"
```
prototaxites, line 22: But here the tool version is 0.0.7... 🤔

modules/nf-core/pretextsnapshot/tests/main.nf.test.snap
DLBPointon: All good, I don't think nf-test cleaned the snapshot correctly after I renamed the tests.

DLBPointon: Green across the board now

DLBPointon:
Damned png format

there will be metadata changes between runs due to png metadata timestamps. I'll change the snapshot to only check name.

PR #10728, opened 2026-03-12
Add kmercounter to nf-core

DLBPointon:
Add the cobiontid/kmercounter module to nf-core

Module creates a npy file containing kmer count data

### DIFF
--- a/modules/nf-core/bwa/index/main.nf
+++ b/modules/nf-core/bwa/index/main.nf
@@ -2,7 +2,7 @@ process BWA_INDEX {
     tag "$fasta"
     // NOTE requires 5.37N memory where N is the size of the database
     // source: https://bio-bwa.sourceforge.net/bwa.shtml#8
-    memory { 6.B * fasta.size() }
+    memory { 7.B * fasta.size() }
 
     conda "${moduleDir}/environment.yml"
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
--- a/modules/nf-core/mobster/templates/main_script.R
+++ b/modules/nf-core/mobster/templates/main_script.R
@@ -91,7 +91,7 @@ lapply(samples, function(sample_name) {
     fit = run_mobster_fit(joint_table=input_table %>% dplyr::filter(sample_id == !!sample_name),
                           descr=paste(description, sample_name, sep=":"))
 
-    if (any(fit[["fits.table"]][["tail"]])) {
+    if (fit[["best"]][["fit.tail"]]) {
         evoparams = evolutionary_parameters(fit)
         fit[["evolutionary_parameters"]] = evoparams
         fit[["best"]][["evolutionary_parameters"]] = evoparams
--- a/modules/nf-core/multiqc/tests/main.nf.test
+++ b/modules/nf-core/multiqc/tests/main.nf.test
@@ -29,13 +29,26 @@ nextflow_process {
 
         then {
             assert process.success
-            assertAll(
-                { assert snapshot(
-                    file(process.out.report[0][1]).name,
-                    file(process.out.data[0][1]).name,
-                    process.out.findAll { key, val -> key.startsWith("versions")
-                }).match() }
-            )
+            assert snapshot(
+                sanitizeOutput(process.out).collectEntries { key, val ->
+                        if (key == "data") {
+                            return [key, val.collect { [path(it[1]).list().collect { file(it.toString()).name }] }]
+                        }
+                        else if (key == "plots") {
+                            return [key, val.collect { [
+                                "pdf",
+                                path("${it[1]}/pdf").list().collect { file(it.toString()).name },
+                                "png",
+                                path("${it[1]}/png").list().collect { file(it.toString()).name },
+                                "svg",
+                                path("${it[1]}/svg").list().collect { file(it.toString()).name }] }]
+                        }
+                        else if (key == "report") {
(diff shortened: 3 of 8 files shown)


modules/nf-core/cobiontid/kmercounter/meta.yml
```diff
@@ -0,0 +1,80 @@
...
+  - - meta:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. `[ id:'sample1' ]`
+    - fasta:
+        type: file
+        description: |
+          Input fasta assembly file
+        ontologies: []
```
prototaxites: Can you add ontologies?

modules/nf-core/cobiontid/kmercounter/main.nf
```diff
@@ -0,0 +1,37 @@
...
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/kmer-counter:0.1.2--h4349ce8_0' :
+        'biocontainers/kmer-counter:0.1.2--h4349ce8_0' }"
+
+    input:
+    tuple val(meta), path(fasta)
+    val kmer_size
+
+    output:
+    tuple val(meta), path(fasta), path( "*.npy" ) , emit: npy
```
prototaxites: Is there a reason to pass the fasta back out?

modules/nf-core/cobiontid/kmercounter/tests/main.nf.test.snap
```diff
@@ -0,0 +1,88 @@
...
+                        },
+                        "genome.fasta:md5,d41d8cd98f00b204e9800998ecf8427e",
+                        "Homo_sapiens_kmer_counts.npy:md5,d41d8cd98f00b204e9800998ecf8427e"
+                    ]
+                ],
+                "versions_kmercounter": [
+                    [
+                        "KMERCOUNTER",
+                        "kmer-counter",
+                        "K-mer counter 0.1.2"
```
prototaxites: Version string doesn't remove tool name

modules/nf-core/cobiontid/kmercounter/main.nf
DLBPointon: No there isn't, removed.

modules/nf-core/cobiontid/kmercounter/meta.yml
DLBPointon: Done

modules/nf-core/cobiontid/kmercounter/tests/main.nf.test.snap
DLBPointon: One 'V' to many in the sed. Done.

modules/nf-core/cobiontid/kmercounter/main.nf
```diff
@@ -0,0 +1,37 @@
+process KMERCOUNTER {
```
prototaxites: I think this module should just be called kmercounter and you should remove cobiontid from all name elements, as cobiontid is not actually a tool?

DLBPointon: I guess not, CobiontID is the collection of tools. I was debating on adding the python script I wrote to generate the kmer sequences that goes with this tool, i though keeping them grouped would make sense?

prototaxites: I guess if you were to do that, you need to make sure this module is called COBIONTID_KMERCOUNTER?

DLBPointon: That's fair. I'll sort that out.

DLBPointon: Done, updated tests too

DLBPointon:
> Can you just make the module support gzipped fata files via gunzip inside the module?

I could but it doesn't need it, i'll change the test to use the chr21 fasta. Unless you think it would benefit from having gunzip included anyway?

prototaxites:
It would be nice to be able to use the module with gzipped FASTA files - would make the module more versatile?

DLBPointon:
> It would be nice to be able to use the module with gzipped FASTA files - would make the module more versatile?


> Can you just make the module support gzipped fata files via gunzip inside the module?

Updated the module to contain the same logic as blastn as that seemed the simplest approach.

modules/nf-core/cobiontid/kmercounter/main.nf
```diff
@@ -0,0 +1,44 @@
...
+    script:
+    def prefix          = task.ext.prefix   ?: "${meta.id}"
+    def args            = task.ext.args     ?: ""
+    def is_compressed   = fasta.getExtension() == "gz" ? true : false
+    def fasta_name      = is_compressed ? fasta.getBaseName() : fasta
+
+    """
+    if [ "${is_compressed}" == "true" ]; then
+        gzip -c -d ${fasta} > ${fasta_name}
+    fi
```
prototaxites, line 30: Can you clean up the decompressed FASTA at the end of the module?

DLBPointon: Can do this at the end or anything simpler?

```
        rm ${fasta_name}
    fi
```

DLBPointon: Done

prototaxites (approved):

PR #10750, opened 2026-03-12
Crabs modules topics migration

### LINKED ISSUE #10711: Migration to topics channels: crabs/import

### LINKED ISSUE #10712: Migration to topics channels: crabs/insilicopcr

mantisps:
## PR checklist

Closes #10711
Closes #10712

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/bff/main.nf
+++ b/modules/nf-core/bff/main.nf
@@ -4,7 +4,7 @@ process BFF {
 
     conda "${moduleDir}/environment.yml"
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
-        'oras://community.wave.seqera.io/library/bioconductor-cellhashr_r-seurat:1c94360d8ed188c4':
+        'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/21/21d3acad5fd1818f00b1c267fcc8c8be88c930cf31e74fb6b5ce444f96827604/data':
         'community.wave.seqera.io/library/bioconductor-cellhashr_r-seurat:25c4bc76749af5ac' }"
 
     input:
@@ -14,7 +14,7 @@ process BFF {
     tuple val(meta), path("*_assignment_bff.csv"), emit: assignment
     tuple val(meta), path("*_metrics_bff.csv")   , emit: metrics
     tuple val(meta), path("*_params_bff.csv")    , emit: params
-    path "versions.yml"                          , emit: versions
+    path "versions.yml"                          , emit: versions_bff, topic: versions
 
     when:
     task.ext.when == null || task.ext.when
--- a/modules/nf-core/bff/meta.yml
+++ b/modules/nf-core/bff/meta.yml
@@ -1,5 +1,3 @@
----
-# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/meta-schema.json
 name: "bff"
 description: Generating cell hashing calls from a matrix of count data.
 keywords:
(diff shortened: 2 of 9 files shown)


modules/nf-core/crabs/import/main.nf
```diff
@@ -27,10 +27,7 @@ process CRABS_IMPORT {
     def is_compressed  = fasta.name.endsWith(".gz")
     def fasta_name     = fasta.name.replace(".gz", "")
     def import_fmt_cmd = "--import-format ${import_format}"
-    def version_cmd    = "\$(crabs --help 2>/dev/null | grep 'CRABS |' | sed 's/.*CRABS | v\\([0-9.]*\\).*/\\1/')"
-    if (workflow.profile.tokenize(',').intersect(['conda', 'mamba']).size() >= 1) {
-        version_cmd    = '1.0.7'
-    }
```
atrigila, line 33: This is interesting, did it have a different version when using conda vs. docker? Does that not happen anymore?

modules/nf-core/crabs/import/main.nf
```diff
@@ -16,7 +16,7 @@ process CRABS_IMPORT {
 
     output:
     tuple val(meta), path("*.txt"), emit: crabsdb
-    path "versions.yml"           , emit: versions
+    tuple val("${task.process}"), val('crabs'), eval("crabs --help 2>/dev/null | grep -oE 'v[0-9.]+' | cut -c2-"), emit: versions_crabs, topic: versions
```
LouisLeNezet, line 19: Is it possible to simplify the expression ?
The problem with using grep is that it isn't always available when using conda on different computer.
Something like:
```suggestion
    tuple val("${task.process}"), val('crabs'), eval("crabs --help 2&>1 | sed -n "s/.*CRABS.*v//p'"), emit: versions_crabs, topic: versions
```

modules/nf-core/crabs/import/main.nf
mantisps: Not sure, but it was overly complicated. I made it more simple and checked Docker, Singularity and Conda tests > all passed!

modules/nf-core/crabs/insilicopcr/main.nf
```diff
@@ -12,43 +12,25 @@ process CRABS_INSILICOPCR {
 
     output:
     tuple val(meta), path("${prefix}.txt"), emit: txt
-    path "versions.yml"                   , emit: versions
+    tuple val("${task.process}"), val('crabs'), eval("crabs --help 2>/dev/null | grep -oE 'v[0-9.]+' | cut -c2-"), emit: versions_crabs, topic: versions
```
LouisLeNezet, line 15: Same here

modules/nf-core/crabs/import/main.nf
mantisps: That one doesn't work. But I just tested this one `crabs --help 2>&1 | sed -nE 's/.*CRABS.*v([0-9.]+).*/\1/p'` and it works fine. I'll update the code

modules/nf-core/crabs/import/main.nf
atrigila: awesome thank you

modules/nf-core/crabs/import/main.nf
mantisps: I updated the command to only use grep and for some reason it doesn't pass the conda test here. It passed on my side. Also it was passing the test before when using the grep command.

image

LouisLeNezet:
The conda environment failing are a bit strange.
It could be nice to update the tool to a newer version:
https://bioconda.github.io/recipes/crabs/README.html

mantisps:
> The conda environment failing are a bit strange. It could be nice to update the tool to a newer version: https://bioconda.github.io/recipes/crabs/README.html

I checked that, but the newest version also doesn't have a way to output the tool version. I'll try another approach

mantisps:
None of the commits I pushed using the `sed` command for the version extraction passed the conda test here (but they all passed it locally..which is weird.. maybe worth investigating). Probably this was the reason why the module used a different approach for the conda environment before.

I'm going back to the previous version that worked (without the use of `sed`). I checked and there are many other modules using grep in the version command, so I believe it shouldn't be an issue.

PR #10812, opened 2026-03-13
new module: stitchr

Clara0611:
New module: stitchr 

Stitchr is a tool that generates full nucleotide and amino acid TCR sequences from V-, J- and CDR3 information. 

Testdata is still missing, waiting for PR

## PR checklist

- [x] If necessary, include test data in your PR. -> Waiting for testdata PR
- [x] Add a resource `label`
- [ ] Use BioConda and BioContainers if possible to fulfil software requirements. -> Not possible, added Dockerfile
  - For modules:


Clara0611:
Draft PR - still waiting for testdata merge ([here](https://github.com/nf-core/test-datasets/pull/1923)

modules/nf-core/stitchr/Dockerfile
mashehu, line 1: looks like these are just python libraries, no need for a dockerfile then. Easiest to use seqera containers instead https://nf-co.re/docs/tutorials/nf-core_components/using_seqera_containers (also please add an `environment.yml`file to also support conda)

Clara0611: Hi, thank you for taking the time to look at my module! 
Unfortunately, Stitchr is not available via conda (bioconda, conda-forge), only via pip. I left the environment.yml file out as well, because the instructions for adding a module (https://nf-co.re/docs/guidelines/components/modules, section 7.2) specify that this file is only required if the tool is available via conda. Of course I would be happy to add the (empty) file back in, if that is the preferred style. 
The reason for including our own container is that stitchr requires a data download (and further manipulation of this data), handled by an internal wrapper script. This script requires root permission, which is not available in the container (with the given global configs). I've tried to circumvent this issue in several ways, e.g. with a local config, but couldn't find a way to make it work. The external container already has the necessary data baked in. However, if you have an immediate idea on something I might try, I'd be very open to that.

mashehu: for the first part: you can install pip-only dependencies via the `environment.yml` see for example https://github.com/nf-core/modules/blob/f899f6adbdb9fc68b4098119e25340488746e776/modules/nf-core/pygenprop/info/environment.yml#L7

mashehu: for the second part: why not make the download command a separate module then?

Clara0611: Thanks for the example! I've added conda just now, but without the container, the species data is missing, so the module cannot run with conda in its current state. Sorry, I forgot that this was an additional concern with conda when I first replied.

About the separate module: We discussed this locally, and decided against it because the download wrapper script (stitchrdl) is so closely intertwined with the rest of the tool, and not really something that would be run on its own. Also, we fear that this might simply move the problem: With two modules, we would first need to run stitchrdl inside a download module (which doesn't work due to permission issues - there is no option to determine where data is saved, stitchrdl deposits the data were stitchr expects to find it), and then move the data to the correct place again, this time manually, in the second module (here, moving over root could perhaps be avoided). Maybe there's something I'm missing though, as this is my first contribution.

jfy133: Jumping in at the request of @mashehu !

In metagenomics modules, we've had a lot of cases of tools coming with databases etc.. In the vast majority of cases (actually, all, from the the top of my head), we've ended up splitting data from the execution. 

This is partly because the databases are very large - and you don't want to repeatedly pull large containers on each execution node (as this can be very slow), and also embedding data inside the container means you then cannot update the data.

Can you maybe go into more detail what the technical issue is (for someone who is not familiar with the tool/the conversation :sweat_smile: )?

I don't understand what you mean by 'due to permission issues', for example. Could you just mount to the container where `strichr` wants thte data to go the work directory for downloading, for example?

Clara0611: Hi, thanks for taking the time :) 

That makes sense, I'll split the download off into another submodule. 

On the topic of permission errors: The issue is that stitchr provides a wrapper script for data downloading, which is not flexible in terms of the destination folder. This is not a problem when running the script with conda, outside of a container. However, when running the script inside of a container, it produces a permission error when it tries to move the downloaded data to the appropriate folder. From what I understand, this is because it tries to move the data into a folder that is immutable in a container setting (/opt/...)? I'm fairly new to containers, and couldn't find a way to circumvent this - if you know of one, I'm very happy to try!

I just reproduced the error by running stitchrdl with a seqera container and profile docker:
"
    OSError: [Errno 18] Invalid cross-device link: 'HUMAN' -> '/opt/conda/lib/python3.14/site-packages/Data/HUMAN'
    
    During handling of the above exception, another exception occurred:
    
    PermissionError: [Errno 13] Permission denied: '/opt/conda/lib/python3.14/site-packages/Data/HUMAN'
    
"   
    It would probably not be allowed to create modules that only support conda, and don't provide a container at all?

akaviaLab: Why is the download script not flexible? Could you post the full log from the Seqera build attempt, not just the final error?
I'm not familiar with the tool, but I can try and debug the docker container and what's going on.

mapo9: Picked this up and played around with the split you asked for, @jfy133 / @mashehu. Sorry for the long silence.

**What changed**

The module is now two: `stitchr/stitchrdl` downloads the IMGT reference and emits it as a `Data/` directory, and `stitchr/thimble` takes that directory as an input. The custom Dockerfile is gone and both modules use stock Seqera containers.

**How the permission problem got solved**

@akaviaLab, this was the answer to your question. `stitchrdl` writes into the installed package's own `Data/` folder, which isn't writable in a container running as a non-root user. But `Data` turns out to be an ordinary importable Python package (`stitchrfunctions.py` resolves it with `importlib.resources.files("Data")`), so `PYTHONPATH` redirects it. Both modules copy or stage a writable `Data` into the work directory and point `PYTHONPATH` at it. No custom image needed.

`thimble` stages its input with `stageAs: 'Data'`, which matters: Python imports the package by that literal name, and if it's called anything else it silently falls back to the empty copy in the container.

**Worth knowing**

The output is byte-identical to what the old baked container produced (`test.tsv:md5,1dc2864a6189891ae9e6bc4d4b377897`), so nothing was lost in the split and the reference data can now be updated.

akaviaLab: It looks like this change solved my question. Generally, the PR looks good to me.

modules/nf-core/stitchr/stitchrdl/tests/main.nf.test
```diff
@@ -0,0 +1,74 @@
...
+    tag "modules"
+    tag "modules_nfcore"
+    tag "stitchr"
+    tag "stitchr/stitchrdl"
+
+    test("human") {
+
+        when {
+            params {
+                module_args = ''
```
jonasscheid: This can go, right? Injecting empty string in module args not needed (same in stub)

modules/nf-core/stitchr/stitchrdl/main.nf
```diff
@@ -0,0 +1,56 @@
...
+    # site-packages is not writable when the container runs as a non-root user, so seed a writable copy of the packaged Data package in the work directory
+    packaged_data=\$(python -c "import importlib.resources as r; print(r.files('Data'))")
+    cp -rL "\$packaged_data" Data
+    chmod -R u+w Data
+
+    # resolve the "Data" package import to the writable copy
+    export PYTHONPATH="\$PWD\${PYTHONPATH:+:\$PYTHONPATH}"
+
+    stitchrdl \\
+        ${args} \\
```
jonasscheid:
```suggestion
        $args \\
```

modules/nf-core/stitchr/stitchrdl/main.nf
```diff
@@ -0,0 +1,56 @@
...
+        -s ${species}
```
jonasscheid:
```suggestion
        -s $species
```

modules/nf-core/stitchr/thimble/tests/main.nf.test
```diff
@@ -0,0 +1,87 @@
...
+                """
+            }
+        }
+    }
+
+    test("human - tsv") {
+
+        when {
+            params {
+                module_args = ''
```
jonasscheid: Double check if this is needed here as well

modules/nf-core/stitchr/thimble/tests/nextflow.config
jonasscheid, line 1: if `module_args` not needed, this file can go as well

modules/nf-core/stitchr/thimble/main.nf
```diff
@@ -0,0 +1,47 @@
...
+    prefix = task.ext.prefix ?: "${meta.id}"
+    // optional files:
+    def codon_usage = alt_codon_usage ? "-cu ${alt_codon_usage}" : ""
+    def allele_preference = allele_preferences ? "-p ${allele_preferences}" : ""
+    """
+    # resolve the "Data" package import to the IMGT reference staged by stitchrdl, otherwise not readable (root issues)
+    export PYTHONPATH="\$PWD\${PYTHONPATH:+:\$PYTHONPATH}"
+
+    thimble \\
+        ${args} \\
```
jonasscheid: nitpick: rm {} here, not needed.

jonasscheid (changes requested):

jonasscheid (approved):

PR #10855, opened 2026-03-13
Cnvpytor/partition topics migration

### LINKED ISSUE #10659: Migration to topics channels: cnvpytor/partition

gonzagrau:
## PR checklist

Closes #10659


modules/nf-core/cnvpytor/partition/tests/main.nf.test.snap
```diff
@@ -21,53 +25,42 @@
...
     "test_cnvpytor_partition": {
         "content": [
             {
                 "0": [
-                    [
-                        {
-                            "id": "test2"
-                        },
-                        "test2.pytor:md5,aa03a8fa15b39f77816705a48e10312a"
-                    ]
```
camlloyd, line 49: Are the test snapshots up-to-date?

famosab:
@gonzagrau I think you need to rerun the tests and update the snapshots once more then we can get this merged :)

famosab:
Also the bug seems to be fixed so we can export the versions properly

modules/nf-core/cnvpytor/partition/tests/main.nf.test.snap
LouisLeNezet: Updated

modules/nf-core/cnvpytor/partition/environment.yml
```diff
@@ -6,3 +6,5 @@ channels:
 dependencies:
   - bioconda::cnvpytor=1.3.1
   - conda-forge::numpy=1.24.3
+  - conda-forge::python=3.10.8
+  - conda-forge::setuptools=67.7.2
```
famosab, line 10: If they are added maybe we should export all versions via topics? https://nf-co.re/docs/specifications/components/modules/general#each-tool-in-a-module-must-report-a-version

LouisLeNezet: Good question, `setuptools` is only to mitigate a bug with `pkg_resources` for all the python modules.
The python version could definitely be added.
I think I will standardize all the CNVpytor modules whenever the hackathon contribution will be merged.

famosab: ok then lets merge this as is, maybe we can have someone do a regex and check whether we always collect all versions (I am a bit torn but I think if you give the tool explicetly in the environment file then it should be added to the versions).

famosab (approved):
🚀

modules/nf-core/cnvpytor/partition/environment.yml
LouisLeNezet: I agree, there is many subtle thing that aren't yet in the nf-core modules lint that I would like to standardize at some point :smile:

LouisLeNezet:
Hi,

Thanks for your contribution.
Similarly to importreadepth the version has been fixed to 1.3.1 while waiting for the new version.
This PR is now ready and will be merged to cleanup the repository.

PR #9652, opened 2026-01-14
Fix cellranger multi's singplex flex failure and add tests

tgelafr-pfzr:
Closes #9639 (https://github.com/nf-core/scrnaseq/issues/515), and adds a new test that would have captured the bug. Corresponding test dataset here: https://github.com/nf-core/test-datasets/pull/1832

In draft until the test-datasets PR is merged so the CI can pass.

- [x] This comment contains a description of changes (with reason).
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [x] If necessary, include test data in your PR.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - For modules:
    - [ ] `nf-core modules test <MODULE> --profile docker`
    - [ ] `nf-core modules test <MODULE> --profile singularity`
    - [ ] `nf-core modules test <MODULE> --profile conda`
  - For subworkflows:
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile docker`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile singularity`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile conda`

PR #10922, opened 2026-03-16
Fix logic to accept singplex flex libraries in cellranger_multi

delfiterradas:
## PR checklist

Closes https://github.com/nf-core/scrnaseq/issues/529 
In reference to https://github.com/nf-core/modules/pull/9652

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/cellranger/multi/tests/singleplex_flex.config
+++ b/modules/nf-core/cellranger/multi/tests/singleplex_flex.config
@@ -2,7 +2,7 @@ process {
 
     withName: CELLRANGER_MULTI {
         stageInMode = 'copy'
-        ext.singleplex_flex = true
+        ext.force_include_frna = true
     }
 
 }
--- a/modules/nf-core/cellranger/multi/main.nf
+++ b/modules/nf-core/cellranger/multi/main.nf
@@ -44,9 +44,9 @@ process CELLRANGER_MULTI {
     if ([ocm_barcodes, cmo_barcodes, frna_sampleinfo].count { input -> input } >= 2) {
         error "The ocm barcodes; cmo barcodes and frna probes are mutually exclusive features. Please use only one per sample, or reach out in slack in case it is really intended."
     }
-    args            = task.ext.args            ?: ''
-    prefix          = task.ext.prefix          ?: "${meta.id}"
-    singleplex_flex = task.ext.singleplex_flex ?: ''
+    args               = task.ext.args               ?: ''
+    prefix             = task.ext.prefix             ?: "${meta.id}"
+    force_include_frna = task.ext.force_include_frna ?: ''
 
     // if references + FASTQ are empty, then don't run corresponding analyses
     // get names of references, if they exist
@@ -77,7 +77,7 @@ process CELLRANGER_MULTI {
     target_panel = gex_targetpanel_name != '' ? "target-panel,./$gex_targetpanel_name" : ''
 
     // fixed RNA reference (not sample info!) also goes under GEX section
-    frna_probeset = ( include_frna || singleplex_flex ) && gex_frna_probeset_name != '' ? "probe-set,./$gex_frna_probeset_name" : ''
+    frna_probeset = ( include_frna || force_include_frna ) && gex_frna_probeset_name != '' ? "probe-set,./$gex_frna_probeset_name" : ''
 
     // VDJ inner primer set
     primer_index = vdj_primer_index ? "inner-enrichment-primers,./references/primers/${vdj_primer_index.getName()}" : ''


modules/nf-core/cellranger/multi/tests/main.nf.test
```diff
@@ -879,7 +879,14 @@ nextflow_process {
     }
     test("cellranger - multi - 10k - kidney - flex - singleplex") {
 
+        config "./singleplex_flex.config"
+
         when {
+
+            params {
+                singleplex_flex = true
```
fmalmeida: Since it is not a parameter, and actually you already have a "singleplex.config" I would say is better to just place it there directly that the task.ext variable is true.

fmalmeida:
Hi @delfiterradas , many thanks for opening this PR.

@grst , I discussed yesterday with Delfina regarding this change, and I personally think it is the best. 

As we keep the normal behaviour but, still give the ability to force the singleplex.

When the first edition was made regarding singleplex, I had commented something in these lines.

What do you think of it?

modules/nf-core/cellranger/multi/tests/singleplex_flex.config
```diff
@@ -0,0 +1,8 @@
+process {
+
+    withName: CELLRANGER_MULTI {
+        stageInMode = 'copy'
+        ext.singleplex_flex = true
```
fmalmeida: Hi @delfiterradas ,
I just have one last comment. Do you oversee any other type of library or run that we could make use of it?

I am just thinking if this is the best variable name.

Or if, for example, instead of saying singleplex_fex we could go with something more explicit and generic like:

`task.ext.{force_include_frna,include_frna,include_frna_probeset}`

Etc. just to make is generic enough and clear that this will be used to force the inclusion of those specific config lines.

delfiterradas: I am not sure about the other libraries, but I agree that it makes sense to modify the name

fmalmeida (approved):
Thanks!

mahesh-panchal:
Hi. My apologies but new `ext` keys are not permitted unless approved by proposal.

Can you explain the purpose of the `ext.force_include_frna` and why it's needed before I make a PR reverting this specific addition.

delfiterradas:
@mahesh-panchal as I mentioned in https://github.com/nf-core/scrnaseq/issues/529, I found a bug in the `test_cellrangermulti` profile, were I would get the following error:

> These library types are not compatible with Flex: Multiplexing Capture


This appears to be related to how the FRNA probeset is added in main.nf. The error can be avoided by modifying the logic introduced in [this commit](https://github.com/nf-core/modules/commit/2501026a8888cac0d2ddac55207a7fdae0ef1f0f):


Specifically, changing:

`frna_probeset = gex_frna_probeset_name != '' ? "probe-set,./$gex_frna_probeset_name" : ''`

to

`frna_probeset = include_frna && gex_frna_probeset_name != '' ? "probe-set,./$gex_frna_probeset_name" : ''`

However, this would then cause the initial error (for which these changes were originally made) to reappear (https://github.com/nf-core/scrnaseq/issues/515). 

The solution proposed with the addition of the `ext.force_include_frna` would satisfy both conditions: It would avoid the library incompatibilities while also having the option to force the inclusion of the probeset when needed to allow running the module with Singleplex Flex.

Please let me know what you think :)

mahesh-panchal:
This is not my field of expertise so I need more information about how and what the tool is doing. The custom ext field has to go. They're often used as ducttape rather than implementing a robust solution. 

I don't see it's use in nf-core/scrnaseq, but given this is a very new addition I guess a change might have been incoming. Or does this issue only appear on the module test?

Can you help fill in what I don't understand?
What I think I understand. The module needs to be smart enough to recognize a Singleplex Flex run.

```
// 1. Determine if this is a Fixed RNA Profiling (Flex) run at all
is_flex = gex_frna_probeset_name != '' && include_gex != ''

// 2. The probeset is required if it's Flex, regardless of multiplexing status
frna_probeset = is_flex ? "probe-set,./$gex_frna_probeset_name" : ''

// 3. The [samples] section header is only needed if multiplexing info is actually provided
// This prevents the "Multiplexing Capture" conflict for singleplex runs
include_frna = (gex_frna_probeset_name && frna_sampleinfo) ? '[samples]' : ''
```
Is this a solution? 3 is already there. This changes from using `include_frna` to using `is_flex` instead. If `sample_info` is not present then this is a singleplex run.

mahesh-panchal:
@fmalmeida I could really use your feedback here on whether changing `frna_probeset` to check `include_gex` rather than `frna_sampleinfo` is a viable solution. 

The change to remove `ext.force_include_frna` is necessary to bring this module in line with nf-core guidelines.

fmalmeida:
Hi @mahesh-panchal ,

I was reviewing the conditionals and trying to think on the two use cases.

What you suggested makes sense, I have to test them a bit today.

@delfiterradas can you try it?

On other way of thinking,

@mahesh-panchal would have something in the meta or real parameter rather than task.ext suffice?

mahesh-panchal:
`meta` maps can have any keys they want, but only permitted meta keys are allowed to be hard-coded into the nf-core modules, in order to preserve interoperability. Values passed through `input:` are an option if they are vital to the correct functioning of the module. 

If you let me know which tests to run locally I can potentially assist.

fmalmeida:
Hi @mahesh-panchal ,

On that note, I believe having the parameter would make sense ... I tried a bit but could not think in an "easy" logic to implement, since both use cases would have gex ... the only difference would be the presence of frna_samplesheet.

You can test like:

**Using nf-core/scrnaseq dev branch**:
```console
nextflow run main.nf -profile test_cellrangermulti,human,podman -c custom.config --outdir results
```

where the custom config is:
```groovy
process {
            withName: 'NFCORE_SCRNASEQ:SCRNASEQ:CELLRANGER_MULTI_ALIGN:CELLRANGER_MKGTF' {
                ext.args = "--attribute=gene_biotype:protein_coding --attribute=gene_biotype:IG_V_gene --attribute=gene_biotype:IG_J_gene --attribute=gene_biotype:IG_D_gene --attribute=gene_biotype:IG_C_gene --attribute=gene_biotype:TR_V_gene --attribute=gene_biotype:TR_J_gene --attribute=gene_biotype:TR_D_gene --attribute=gene_biotype:TR_C_gene"
            }
        }
```

And, you can ensure the other use cases still run by doing (also in scrnaseq dev branch):
```console
nf-test test tests/main_pipeline_cellrangermulti.nf.test
```

delfiterradas:
Hi @fmalmeida and @mahesh-panchal, thank you for your input! Throughout the day I will test the module switching from the ext argument to a parameter via `input:` and create a new PR

mahesh-panchal:
My issue here is that I still don't quite understand the process inputs which are necessary. I also don't know if there are any meta keys that could be used in a `.filter` in the pipeline call to not supply certain inputs. 

Is the problem the presence of the gene expression reference so probeset would always be there or that it's missing but the probeset still needs to be provided? A table of the possible combinations the module needs to handle for this subset would be helpful. 

I'll try my best though to investigate in the mean time.

fmalmeida:
Hi @mahesh-panchal ,

I believe for that case regarding the dataset itself, @grst and @nictru would have more insights than me.

I believe to make it easier to track, we can revert the PR as you mentioned and continue exploring the possibilities and the real fix in the issue that we can re-open.

As you said, it may indeed be that the problem is the other dataset that may be solved by removing the probeset in case it is not used.

@grst and @nictru ,
Would you be able to join the discussions in the ticket to track down a deconvolution of what data combinations make sense is the first step.

mahesh-panchal:
Is there documentation anywhere what values meta.chemistry can take and if it can be used to distinguish between the different analysis types?

mahesh-panchal:
The scrnaseq test seems to have both standard assays and Flex. Is this intended? Should these be split into two separate tests or should the pipeline be able to handle standard and flex at the same time?

fmalmeida:
Hi @mahesh-panchal ,

In the pipeline the two tests are added as sub-profiles so when executed they are done separately (rat or human).

If that is what you meant.

From the discussion, I understand we have a clear homework to do that can be addressed first in the scrnaseq repo ... so I am re-opening the issue there for us to further work on this and have a better answer.

That is:

* Do the full mapping of possible combination of data and files
* Ensure that the testings a correct and we are not for example adding a file that should not be in the combination 
* Based on the above, then think on the solution for the code logic.

---

So let's revert to remove the prohibited task.ext variable

---

@delfiterradas -- the bullets above of reference of what we can do to ensure we know the fix is being done in the correct context.

mahesh-panchal:
I meant within the human samplesheet. I also just noticed something is emptying the assets/fb_reference.csv. When I run the test above the content of `fb_reference.csv` is deleted.

PR #10929, opened 2026-03-17
fix(subworkflows): preserve meta in channel emits

pinin4fjords:
## Summary

Two subworkflows were stripping the `meta` map from emitted channels, which breaks downstream consumers that need sample metadata for per-sample processing.

### Changes

- **`bam_dedup_umi`**: The `ch_multiqc_files` emit collected stats/flagstat/idxstats, transposed them, then discarded the meta with `.map { item -> item[1] }`. This PR removes that `.map` so the emitted channel retains the `[ val(meta), path(file) ]` tuple structure.

- **`fastq_qc_trim_filter_setstrandedness`**: The `multiqc_files` emit similarly discarded meta via `.map { _meta, file -> file }` after `.transpose()`. This PR removes that `.map` to preserve the tuple structure.

### Motivation

Keeping meta in emitted channels is consistent with nf-core conventions and enables downstream workflows to route files per-sample (e.g. for per-sample MultiQC reports). Without meta, consuming workflows cannot determine which sample a file belongs to.

This is needed by nf-core/rnaseq's `--skip_quantification_merge` feature (nf-core/rnaseq#1744), but is a general improvement for any consumer of these subworkflows.

**Note**: This is a breaking change for any pipeline that consumes these emits as bare paths rather than tuples. Consumers will need to add `.map { _meta, file -> file }` on their side if they don't need the meta, or update to handle tuples.

## Test plan
- [ ] Verify existing subworkflow tests pass
- [ ] Check that consuming pipelines (nf-core/rnaseq) handle the new tuple structure

### DIFF
--- a/subworkflows/nf-core/fastq_qc_trim_filter_setstrandedness/tests/main.nf.test
+++ b/subworkflows/nf-core/fastq_qc_trim_filter_setstrandedness/tests/main.nf.test
@@ -108,7 +108,7 @@ nextflow_workflow {
                 .join('\n') // Join the processed lines back into a single text block
 
             // Parse sortmerna log for rRNA detection stats
-            def sortmernaLog = path(workflow.out.multiqc_files.find { it[1].toString().endsWith('.sortmerna.log') }[1])
+            def sortmernaLog = path(workflow.out.multiqc_files.find { _meta, file -> file.toString().endsWith('.sortmerna.log') }[1])
                 .getText()
             def sortmernaRrnaMatch = (sortmernaLog =~ /Total reads passing E-value threshold = (\d+)/)
             def sortmernaRrnaCount = sortmernaRrnaMatch ? sortmernaRrnaMatch[0][1].toInteger() : -1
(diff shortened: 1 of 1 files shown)


subworkflows/nf-core/fastq_qc_trim_filter_setstrandedness/tests/main.nf.test
```diff
@@ -108,7 +108,7 @@ nextflow_workflow {
                 .join('\n') // Join the processed lines back into a single text block
 
             // Parse sortmerna log for rRNA detection stats
-            def sortmernaLog = path(workflow.out.multiqc_files.find { it.toString().endsWith('.sortmerna.log') })
+            def sortmernaLog = path(workflow.out.multiqc_files.find { it[1].toString().endsWith('.sortmerna.log') }[1])
```
mashehu: I guess this needs to be updated to strict syntax as well

pinin4fjords: Thanks, replaced all the implicit 'its'

pinin4fjords:
Thanks @maxulysse @mashehu !

PR #10873, opened 2026-03-13
topic migration for deseq2

lourdeslabrador:
## PR checklist

Closes #10765 

- [ ] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [ ] Remove all TODO statements.
- [ ] Broadcast software version numbers to `topic: versions` - [See version_topics](https://nf-co.re/blog/2025/version_topics)
- [ ] Follow the naming conventions.
- [ ] Follow the parameters requirements.
- [ ] Follow the input/output options guidelines.
- [ ] Add a resource `label`
- [ ] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - For modules:
    - [ ] `nf-core modules test <MODULE> --profile docker`
    - [ ] `nf-core modules test <MODULE> --profile singularity`
    - [ ] `nf-core modules test <MODULE> --profile conda`
  - For subworkflows:
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile docker`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile singularity`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile conda`

PR #10931, opened 2026-03-17
feat: add topic version emission to quant-related modules

pinin4fjords:
## Summary

- Convert `custom/tx2gene`, `tximeta/tximport`, and `summarizedexperiment/summarizedexperiment` to emit versions via `topic: versions` channels
- Update meta.yml files with `topics` section and updated version output descriptions

This enables downstream subworkflows (e.g. `quant_tximport_summarizedexperiment`) to collect versions automatically via topic channels instead of manual `ch_versions.mix()`, which is needed for conditional execution paths like #10928.

Follows the pattern I observed in #10873 for templated modules.

## Test plan
- [ ] Existing module tests pass (snapshot update expected for versions format)
- [ ] `process.out.versions` still works in tests (emit is preserved alongside topic)

### DIFF
--- a/modules/nf-core/custom/tx2gene/main.nf
+++ b/modules/nf-core/custom/tx2gene/main.nf
@@ -16,7 +16,7 @@ process CUSTOM_TX2GENE {
 
     output:
     tuple val(meta), path("*tx2gene.tsv"), emit: tx2gene
-    path "versions.yml"                  , emit: versions
+    path "versions.yml"                  , emit: versions, topic: versions
 
     when:
     task.ext.when == null || task.ext.when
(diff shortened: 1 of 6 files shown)


modules/nf-core/custom/tx2gene/main.nf
```diff
@@ -16,7 +16,7 @@ process CUSTOM_TX2GENE {
 
     output:
     tuple val(meta), path("*tx2gene.tsv"), emit: tx2gene
-    path "versions.yml"                  , emit: versions
+    path "versions.yml"                  , emit: versions, topic: versions
```
maxulysse, line 19:
```suggestion
    path "versions.yml"                  , emit: versions_tx2gene, topic: versions
```

maxulysse: that's the pattern

pinin4fjords:
Thanks @maxulysse !

PR #10951, opened 2026-03-18
Add `.gzi` to `BCFTOOLS_MPILEUP`

LouisLeNezet:
## PR checklist

When using compressed fasta file, bcftools require a `.gzi` file.
If not present it will be create on the fly, however in some rare (random?) case this file will not be available during bcftools calls when piping multiple command.
This will induce an error message in the log similar to the following even if all the files are properly staged and contains all the sequence necessary.

```log
Note: none of --samples-file, --ploidy or --ploidy-file given, assuming all sites are diploid
[mpileup] 1 samples in 1 input files
[mpileup] maximum number of reads per input file set to -d 250
[E::faidx_adjust_position] The sequence "chr22" was not found
[E::faidx_adjust_position] The sequence "chr22" was not found
```

More problematic,, the process will still properly finish, eventhough the output is empty (only header in the vcf present).

See the [conversation on slack](https://nfcore.slack.com/archives/CE6SDBX2A/p1772652501457779) for more details.

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


modules/nf-core/bcftools/mpileup/main.nf
```diff
@@ -9,7 +9,7 @@ process BCFTOOLS_MPILEUP {
 
     input:
     tuple val(meta), path(bam), path(intervals_mpileup, stageAs: 'mpileup_intervals/*'), path(intervals_call, stageAs: 'call_intervals/*')
-    tuple val(meta2), path(fasta), path(fai)
+    tuple val(meta2), path(fasta), path(fai), path(gzi)
```
maxulysse, line 12: How about this instead?
```suggestion
    tuple val(meta2), path(fasta), path(index)
```

maxulysse: I assume we don't use fai and gzi at the same time

modules/nf-core/bcftools/mpileup/meta.yml
maxulysse, line 45: need to add the .gz here

LouisLeNezet:
After discussion `.gzi` should simply replace `.fai` when working with compressed fasta.
We just need to add `.gz` in the meta.yml.
I will start a new pr for that.

PR #10959, opened 2026-03-18
mcstaging/imc2mc topic update

kbestak:
* mcstaging/imc2mc - topic version outputs
* removing UUID in output file, as done in the ASHLAR module to ensure output stability
* added ontologies, corrected meta

## PR checklist

Closes https://github.com/nf-core/modules/issues/10956

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/mcstaging/imc2mc/main.nf
+++ b/modules/nf-core/mcstaging/imc2mc/main.nf
@@ -8,7 +8,7 @@ process MCSTAGING_IMC2MC {
 
     output:
     tuple val(meta), path("*.tif"), emit: tif
-    path "versions.yml"           , emit: versions
+    tuple val("${task.process}"), val('imc2mc'), eval("python /imc2mc/scripts/imc2mc.py --version | sed 's/v//g'"), emit: versions_imc2mc, topic: versions
 
     when:
     task.ext.when == null || task.ext.when
@@ -20,17 +20,13 @@ process MCSTAGING_IMC2MC {
     }
     def args   = task.ext.args   ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
-
     """
     python /imc2mc/scripts/imc2mc.py \
         -i ${txtfile} \
         -o "${prefix}.tif" \
         $args
 
-    cat <<-END_VERSIONS > versions.yml
-    "${task.process}":
-        imc2mc: \$(python /imc2mc/scripts/imc2mc.py --version | sed 's/v//g')
-    END_VERSIONS
+    sed -i -E 's/UUID="urn:uuid:[[:xdigit:]]{8}-[[:xdigit:]]{4}-[[:xdigit:]]{4}-[[:xdigit:]]{4}-[[:xdigit:]]{12}"/                                                    /g' ${prefix}.tif
     """
 
     stub:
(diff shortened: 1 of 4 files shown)


modules/nf-core/mcstaging/imc2mc/main.nf
```diff
@@ -8,7 +8,7 @@ process MCSTAGING_IMC2MC {
 
     output:
     tuple val(meta), path("*.tif"), emit: tif
-    path "versions.yml"           , emit: versions
+    tuple val("${task.process}"), val('imc2mc'), eval("python /imc2mc/scripts/imc2mc.py --version | sed 's/v//g'"), emit: versions_imc2mc, topic: versions
```
maxulysse, line 11: If the tool is imc2mc, why is this in mcstaging?

kbestak: These are all staging modules adapting data into mcmicro-friendly inputs - imc2mc, phenoimager2mc, and macsima2mc.
macsima2mc recently was approved so I took the version outputs as they were written there: https://github.com/nf-core/modules/blob/master/modules/nf-core/mcstaging/macsima2mc/main.nf

famosab: the problem is there is no description in the docs (or at least I cannot find it) that tells you to use the module and not the submodule for the value

famosab, line 11:
```suggestion
    tuple val("${task.process}"), val('mcstaging'), eval("python /imc2mc/scripts/imc2mc.py --version | sed 's/v//g'"), emit: versions_mcstaging, topic: versions
```

kbestak: The catch here is that all of the aforementioned tools are individual tools that we grouped here into mcstaging given their shared purpose to keep things more organised. Was this a mistake? Given that both macsima2mc and phenoimager2mc have been merged with a versions_platform2mc wording, should all be corrected or is this acceptable?

famosab: Coming from the docs that I have in the back of my mind I would argue that this has been a mistake in the first place. We do not group the modules by purpose. Usually we have tool/command as the convention. For example the tool `bcftools` has many command. Each of the command is a submodule: `bcftools/view` runs the command `bcftools view` etc. Maybe you can link the original PR / issue where this was decided and discussed but judging from what you told us here I would argue that this needs to be changed. 

Maybe other @nf-core/maintainers have opinions on this :)

matthdsm: Modules is "general purpose", so alle modules MUST be TOOL/SUBTOOL.
As these are "modified" modules for a specific pipeline, these should be local to the pipeline IMO.

If the changes to the original tool are minimal AND can not be done through configuration files, then you can work with module patches in the pipeline.

mahesh-panchal: My opinion is that modules should be tool/command and not grouped for a purpose/pipeline. If a module needs to be adapted to a pipeline, it should be included in the pipeline, and `nf-core modules patch` should be used to make it pipeline specific. 

The path to the python script is also hardcoded to the root and this cannot be portable. Should it really be this way?

It's absolutely fine to have local modules in pipelines and I feel like this should remain there. I would only bring it into nf-core modules if the module needs to be used by other pipelines and in which case this python script should then be properly packaged.

maxulysse: Agreeing with @matthdsm and @famosab here, modules are only organized by names, but you could use any info in a tag, and that would be shown on the website

kbestak:
I'm closing this pull request as it was easier to create a new branch, and leave the MCSTAGING/IMC2MC mostly unchanged (except the deprecation message), then add IMC2MC as its individual module.

Thank you all for highlighting the issue --> requested changes applied here: https://github.com/nf-core/modules/pull/11109

PR #10972, opened 2026-03-19
[WIP] Add cpu and allow bigger resource label

### LINKED ISSUE #10968: AGAT_SPKEEPLONGESTISOFORM  fails on small ish genomes due to memory

### Description of the bug

I suggest we may add extra memory to the module. 
I am running on an SGE HPC, 

On Bombus 
[d5/084177] NOTE: Process `AGAT_SPKEEPLONGESTISOFORM (Bombus_terrestris)` terminated with an error exit status (140) -- Execution is retried (1)

We don't need more cpu, as it only runs on 1. Just more memory, but `process_high_memory` will be way too much

```console

```

### Relevant files

_No response_

### System information

_No response_

chriswyatt1:
## PR checklist

Closes #10968 

I added the cpu flag and gave more resources to this module, as I think this will speed it up

But before we merge this, I want to test it, hence WIP. Maybe splitting up with cpu could make it slower, if IPC::Shareable in this perl script is causing the stalling, when /dev/shm gets full

  - For modules:

### DIFF
--- a/modules/nf-core/agat/spkeeplongestisoform/main.nf
+++ b/modules/nf-core/agat/spkeeplongestisoform/main.nf
@@ -1,6 +1,6 @@
 process AGAT_SPKEEPLONGESTISOFORM {
     tag "$meta.id"
-    label 'process_single'
+    label 'process_medium'
 
     conda "${moduleDir}/environment.yml"
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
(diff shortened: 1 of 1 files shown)


dialvarezs:
Ups, I already gave you a ✅ after I saw the changes. But let's wait until your testing is done.

chriswyatt1:
Thanks!. 
Actually, I already ran again on HPC and local, and it does seem a lot faster, so I don't see why we can't add these changes. Not sure if we always need process_medium,, but now it runs in 2 minutes instead of 2 hours+, so I think its worth the change :)

PR #10976, opened 2026-03-20
refactor(dia_proteomics): centralize mass accuracy control via defineMassAccuracySettings

an-altosian:
## Summary

Centralizes mass accuracy parameter control across all DIA-NN pipeline steps by calling `defineMassAccuracySettings` twice — once before preliminary analysis (from user params or null if automatic) and once before individual analysis (updating from DIA-NN log only when automatic mode is enabled).

**Problem**: Mass accuracy, MS1 accuracy, and scan window were controlled differently across steps — some via `params.*` in config, some via `meta.*`. This was fragile and caused a bug where explicit mass accuracy params weren't passed to `DIANN_ASSEMBLEEMPIRICALLIBRARY` when `mass_acc_automatic=false`.

**Solution**: All process configs now use the same `meta.*`-based pattern:
- `meta.mass_acc_ms2 ? "--mass-acc ${meta.mass_acc_ms2}" : ''`
- `meta.mass_acc_ms1 ? "--mass-acc-ms1 ${meta.mass_acc_ms1}" : ''`
- `meta.scan_window ? "--window ${meta.scan_window}" : ''`

### Changes

**`main.nf`**:
- Refactored `defineMassAccuracySettings` to separate mass accuracy and scan window logic with null-safe navigation
- Added `ch_initial_mass_settings` / `ch_initial_mass_settings_by_exp` channels computed once upfront
- Joined mass settings into preliminary analysis and empirical library inputs at process call time
- Stripped mass settings keys from meta after empirical library output to prevent key pollution in downstream `combine`

**`nextflow.config`**:
- `DIANN_PRELIMINARYANALYSIS`: Conditional `meta.*` flags (null in automatic mode → DIA-NN decides)
- `DIANN_ASSEMBLEEMPIRICALLIBRARY`: Falls back to `--individual-mass-acc` / `--individual-windows` when meta values are null
- `DIANN_INDIVIDUALANALYSIS`: Unconditional `--mass-acc`, `--mass-acc-ms1`, `--window` (always populated at this stage)

### Behavior

| Step | automatic=true | automatic=false (ppm) |
|------|----------------|----------------------|
| Preliminary analysis | No mass-acc flags (DIA-NN decides) | Explicit user values |
| Empirical library | `--individual-mass-acc --individual-windows` | Explicit user values |
| Individual analysis | Values from DIA-NN log | User values (unchanged) |

## Test plan
- [x] nf-test passes with `--profile docker` (manual mode with ppm tolerances of 20)
- [x] Snapshot updated — md5 change expected due to explicit mass accuracy flags now being passed
- [x] Verified `DIANN_ASSEMBLEEMPIRICALLIBRARY` receives explicit `--mass-acc`/`--mass-acc-ms1` when automatic=false
- [x] Verified `--individual-mass-acc`/`--individual-windows` used when automatic=true (meta values null)
- [x] nf-core lint passes for `dia_proteomics_analysis` subworkflow


an-altosian:
Hi @pinin4fjords , I noticed a discrepancy with quantms. Here, if the ms-acc-automatic is not set, no ms-acc and ms1-acc parameters will be set. In quantms, they will fall back to the parameters in meta, which come from sdrf or params. As this is not using sdrf, I added the fall back to params. This applies to three stages, first pass search, empirical lib generation, and second pass search. One example is the empirical lib step in quantms listed below.

https://github.com/bigbio/quantms/blob/e719f43a9fc3ef6530330a659818a8456114fd34/modules/local/diann/assemble_empirical_library/main.nf#L31

subworkflows/nf-core/dia_proteomics_analysis/nextflow.config
```diff
@@ -95,9 +97,8 @@ process {
                 "--no-main-report",
                 "--relaxed-prot-inf",
                 meta.pg_level ? "--pg-level ${meta.pg_level}" : '',
-                meta.mass_acc_ms2 ? "--mass-acc ${meta.mass_acc_ms2}" : '',
-                meta.mass_acc_ms1 ? "--mass-acc-ms1 ${meta.mass_acc_ms1}" : '',
-                meta.scan_window ? "--window ${meta.scan_window}" : ''
+                meta.mass_acc_ms2 && meta.mass_acc_ms1 && params.mass_acc_automatic ? "--mass-acc ${meta.mass_acc_ms2} --mass-acc-ms1 ${meta.mass_acc_ms1}" : (params.library_mass_acc && params.library_ms1_acc ? "--mass-acc ${params.library_mass_acc} --mass-acc-ms1 ${params.library_ms1_acc}" : ""),
```
pinin4fjords: This one I'm less sure about. The `defineMassAccuracySettings()` function in `main.nf` already handles the automatic vs manual decision and writes the resolved values into `meta.mass_acc_ms2`, `meta.mass_acc_ms1`, and `meta.scan_window`. The current config just reads from meta, which already has the correct values for both the automatic and non-automatic cases.

Adding a `params.mass_acc_automatic` guard here creates a second decision path for the same logic, split across two files. When `mass_acc_automatic` is false, this would use `params.library_mass_acc` instead of `meta.mass_acc_ms2` (which was set from per-sample `precursor_tolerance` / `fragment_tolerance`). That means per-sample tolerances get silently ignored in favor of the global library params.

If the intent is to prefer `library_mass_acc` over per-sample tolerances, I think the cleaner approach would be to update `defineMassAccuracySettings()` in `main.nf` to incorporate that priority, so the config stays a simple consumer of meta.

an-altosian: Thanks for pointing me to this function. I tried another solution. Basically, we call defineMassAccuracySettings twice. 
- Once before preliminary analysis without diann log, this is to decide if we want to use `automatic` for the three parameters or the provided input parameter values. 
- The second call is the same as what you wrote, but this time we call it with the DIANN log, the purpose is to update the three parameters if `automatic`. By doing so, we don't need to have complex logics in the config file. We will always specify the three parameters based on the values in `meta`.

pinin4fjords (approved):

PR #10997, opened 2026-03-21
Add gcta/bivariatereml module

lyh970817:
This PR adds the `gcta/bivariatereml` module, which runs bivariate REML analysis with a single dense GRM.

This PR should merge after the `gcta/makegrm` module in https://github.com/nf-core/modules/pull/11650.

## PR checklist

- [x] Add a resource `label`
- [x] `nf-core modules test gcta/bivariatereml --profile docker`
- [x] `nf-core modules test gcta/bivariatereml --profile singularity`
- [x] `nf-core modules test gcta/bivariatereml --profile conda`


modules/nf-core/gcta/bivariatereml/main.nf
```diff
@@ -0,0 +1,48 @@
...
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def pair_id = prefix
+    def qcovar_param = quant_covariates_file ? "--qcovar ${quant_covariates_file}" : ''
+    def covar_param = cat_covariates_file ? "--covar ${cat_covariates_file}" : ''
+    """
+    gcta \\
+        --reml-bivar 1 2 \\
```
lyh970817: `1 2` is to select columns from `phenotype_file`. Is it okay for them to hard-coded? Or should they be part of the `phenotype_file` tuple with defaults? @SPPearce

SPPearce: I don't understand the options, presumably there are times when this would be different?
Does it have a default? Presumably not, else you wouldn't be specifying it here.
It could be an optional value channel, after the other 4 current inputs (and take "1 2" as the detail)

lyh970817: But the "1 2" values are closely related to `phenotype_file` which, meaning to select a column pair 1 and 2 in `phenotype_file` for the computation. They can indeed be different such as "1 4" or "2 3" to select a different pair of columns. 

As this is close related to `phenotype_file` I wonder if it'd make more sense to have it in the same tuple, i.e.  `tuple val(meta2), path(phenotype_file) val(col1) val(col2)`.

SPPearce: Fair enough, you can do that if it is more appropriate.

lyh970817: Addressed in d9ce888c4. The phenotype tuple now includes two values, `phenotype_col1` and `phenotype_col2`, which are passed to `--reml-bivar` when non-empty. Passing both values empty leaves `--reml-bivar` without explicit columns so GCTA uses its default bivariate phenotype columns. `nf-core modules test gcta/bivariatereml --profile docker` passes with both explicit `1, 2` and empty/default test cases.

modules/nf-core/gcta/bivariatereml/main.nf
```diff
@@ -0,0 +1,48 @@
...
+    tuple val(meta), path("*.log"), emit: log_file
+    tuple val("${task.process}"), val("gcta"), eval("gcta --version | sed -En 's/^[*] version v([0-9.]*).*/\\1/p'"), emit: versions_gcta, topic: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def phenotype_columns = [phenotype_col1, phenotype_col2].findAll { column -> column }
```
SPPearce: What is this findAll doing?

lyh970817: Removed the `findAll` helper in df522722a. The command now directly uses `--reml-bivar ${phenotype_col1} ${phenotype_col2}` when both tuple values are present, otherwise it emits bare `--reml-bivar` for the default case.

SPPearce (approved):

PR #10999, opened 2026-03-21
Add gcta/calculateldscores module

lyh970817:
This PR adds the `gcta/calculateldscores` module, which calculates LD scores with GCTA and derives LD-based SNP-group files.

## PR checklist

- [x] Add a resource `label`
- [x] `nf-core modules test gcta/calculateldscores --profile docker`
- [x] `nf-core modules test gcta/calculateldscores --profile singularity`
- [x] `nf-core modules test gcta/calculateldscores --profile conda`


modules/nf-core/gcta/calculateldscores/main.nf
```diff
@@ -0,0 +1,73 @@
...
+    def bfile_prefix = bed.baseName
+    """
+    gcta \\
+        --bfile ${bfile_prefix} \\
+        --ld-score-region ${ld_score_region} \\
+        --out ${prefix}_gcta_ld \\
+        --thread-num ${task.cpus} \\
+        ${args}
+
+    Rscript - ${prefix}_gcta_ld.score.ld ${prefix} <<'EOF'
```
SPPearce: If you are going to do this, you might as well just have the `${prefix}_gcta_ld.score.ld` defined directly in the read.table

modules/nf-core/gcta/calculateldscores/main.nf
```diff
@@ -0,0 +1,73 @@
...
+      header = TRUE,
+      colClasses = c("character", rep("numeric", 8))
+    )
+
+    quartiles <- summary(lds_seg\$ldscore_SNP)
+
+    lb1 <- which(lds_seg\$ldscore_SNP <= quartiles[2])
+    lb2 <- which(lds_seg\$ldscore_SNP > quartiles[2] & lds_seg\$ldscore_SNP <= quartiles[3])
+    lb3 <- which(lds_seg\$ldscore_SNP > quartiles[3] & lds_seg\$ldscore_SNP <= quartiles[5])
+    lb4 <- which(lds_seg\$ldscore_SNP > quartiles[5])
```
SPPearce: What is going on here? Where did quartile 4 go? What is lb1-4 mean?

SPPearce (commented):
Not so sure on this one, why is it doing some post-processing in R?

lyh970817:
The R post-processing is intentional here. It follows the official GCTA GREML-LDMS tutorial, where `--ld-score-region` first produces the `*.score.ld` file and a following R step stratifies SNPs into LD-score groups before GRM construction with `--extract`: https://yanglab.westlake.edu.cn/software/gcta/#gremlinwgsorimputeddata

This output is also needed for a planned GREML-LDMS subworkflow. The future subworkflow will use this module to produce the LD-score SNP groups, then pass those group files into downstream GCTA GRM-building steps and finally into multi-GRM REML. So the R step is part of the workflow contract, not incidental post-processing.

I agree the current implementation could make that intent clearer. I’ll add clearer naming around the LD-score grouping and make the relationship to the GCTA GREML-LDMS tutorial explicit in the module description/test context.

modules/nf-core/gcta/calculateldscores/main.nf
lyh970817: This mirrors the official GCTA GREML-LDMS R example for LD-score SNP grouping: https://yanglab.westlake.edu.cn/software/gcta/#step2option1stratifythesnpsbyldscoresofindividualsnpsinr

The naming is definitely unclear though. `summary()` returns min, Q1, median, mean, Q3, and max, so the code is using Q1, median, and Q3 as thresholds:
- group 1: `ldscore_SNP <= Q1`
- group 2: `Q1 < ldscore_SNP <= median`
- group 3: `median < ldscore_SNP <= Q3`
- group 4: `ldscore_SNP > Q3`

I’ll rename the variables so they describe the LD-score groups rather than using `lb1`-`lb4`, and I’ll make the threshold definitions explicit so the grouping is easier to review.

SPPearce: You can use `quantile` rather than `summary` then.

modules/nf-core/gcta/calculateldscores/main.nf
```diff
@@ -0,0 +1,69 @@
...
+
+    input:
+    tuple val(meta), path(bed), path(bim), path(fam)
+    val ld_score_region
+
+    output:
+    tuple val(meta), path("*_gcta_ld.score.ld"), emit: ld_scores
+    tuple val(meta), path("*_snp_group*.txt"), emit: snp_group_files
+    tuple val("${task.process}"), val("gcta"), eval("gcta --version | sed -En 's/^[*] version v([0-9.]*).*/\\1/p'"), emit: versions_gcta, topic: versions
+
```
SPPearce, line 18: Add the r-base version.

modules/nf-core/gcta/calculateldscores/main.nf
```diff
@@ -0,0 +1,69 @@
...
+    lb1 <- which(lds_seg\$ldscore_SNP <= quartiles[2])
+    lb2 <- which(lds_seg\$ldscore_SNP > quartiles[2] & lds_seg\$ldscore_SNP <= quartiles[3])
+    lb3 <- which(lds_seg\$ldscore_SNP > quartiles[3] & lds_seg\$ldscore_SNP <= quartiles[5])
+    lb4 <- which(lds_seg\$ldscore_SNP > quartiles[5])
+
+    write.table(lds_seg\$SNP[lb1], "${prefix}_snp_group1.txt", row.names = FALSE, quote = FALSE, col.names = FALSE, append = TRUE)
+    write.table(lds_seg\$SNP[lb2], "${prefix}_snp_group2.txt", row.names = FALSE, quote = FALSE, col.names = FALSE, append = TRUE)
+    write.table(lds_seg\$SNP[lb3], "${prefix}_snp_group3.txt", row.names = FALSE, quote = FALSE, col.names = FALSE, append = TRUE)
+    write.table(lds_seg\$SNP[lb4], "${prefix}_snp_group4.txt", row.names = FALSE, quote = FALSE, col.names = FALSE, append = TRUE)
```
SPPearce: A lot of redundant code here.
Could just add `readr` to the container and use `write_tsv`

Why append = TRUE, unless the file is generated by the gcta tool?

modules/nf-core/gcta/calculateldscores/main.nf
lyh970817: Updated.

The grouping now uses `quantile(lds_seg$ldscore_SNP, probs = c(0.25, 0.50, 0.75), names = FALSE)` directly instead of indexing into `summary()`.

I also replaced the `lb1`-`lb4` variables with named `snp_group_filters`. The fourth group is now explicit as SNPs with `ldscore_SNP > Q3`, written to the fourth SNP-group file.

modules/nf-core/gcta/calculateldscores/main.nf
lyh970817: Added.

The module now emits a separate `versions_rbase` topic entry using `Rscript -e "cat(as.character(getRversion()))"`.

I also documented it in `meta.yml` and updated the nf-test snapshots so both `gcta` and `r-base` versions are checked.

modules/nf-core/gcta/calculateldscores/main.nf
lyh970817: Updated.

The repeated `write.table(..., append = TRUE)` calls have been replaced with a loop over the four SNP-group filters using `writeLines(...)`, so the files are written directly rather than appended.

I kept this in base R rather than adding `readr`, because the current stable community Wave image already contains `gcta` and `r-base` but not `r-readr`. Adding `readr` would require changing the frozen module container, and base R is sufficient now that the repeated writes and unnecessary append behavior are gone.

modules/nf-core/gcta/calculateldscores/tests/main.nf.test
```diff
@@ -132,7 +130,8 @@ nextflow_process {
                     assert groupedSnps.toSet() == ldScoreSnpsSet
                 },
                 {
-                    assert snapshot(sanitizeOutput(process.out)).match()
+                    assert snapshot(process.out.ld_scores, process.out.snp_group_files).match()
+                    assert snapshot(process.out.findAll { key, val -> key.startsWith("versions") }).match("versions_stub")
```
SPPearce: Not sure where this has come from, we don't do multiple snapshots (any more), and the `snapshot(sanitizeOutput(process.out)).match()` should be fine.

SPPearce: If there is something you need to add to `unstableKeys`, do that.

modules/nf-core/gcta/calculateldscores/tests/main.nf.test
```diff
@@ -66,16 +66,14 @@ nextflow_process {
                     assert groupedSnps.toSet() == ldScoreSnpsSet
                 },
                 {
-                    assert snapshot(
-                        process.out.ld_scores,
-                        process.out.versions_gcta
-                    ).match()
+                    assert snapshot(process.out.ld_scores).match()
+                    assert snapshot(process.out.findAll { key, val -> key.startsWith("versions") }).match("versions")
```
SPPearce: Same here.

modules/nf-core/gcta/calculateldscores/tests/main.nf.test
lyh970817: Updated to use a full sanitized output snapshot.

The real execution now uses `snapshot(sanitizeOutput(process.out, unstableKeys: ["snp_group_files"])).match()`. The `unstableKeys` exception is only for `snp_group_files` because this fixture validly produces two empty SNP-group files, and snapshotting those by md5 triggers nf-core lint's empty-file md5 check. The surrounding assertions still check the SNP-group files exist, have the expected names, and partition the LD-score SNP set.

The obsolete standalone `versions` snapshots were removed, and the combined snapshot now includes `ld_scores`, `snp_group_files`, `versions_gcta`, and `versions_rbase`.

modules/nf-core/gcta/calculateldscores/tests/main.nf.test
lyh970817: Updated this one as well.

The stub test now uses `snapshot(sanitizeOutput(process.out)).match()`, so outputs and versions are captured together in a single snapshot. The old separate `versions_stub` snapshot entry was removed.

SPPearce (approved):

PR #11003, opened 2026-03-21
Add gcta/makegrmpart module

lyh970817:
This PR adds the `gcta/makegrmpart` module, which computes one partition of a GCTA genetic relationship matrix from PLINK inputs.

Merge dependencies: none. This PR is self-contained and does not require any other split GCTA PR to merge first.

## PR checklist

- [x] Add a resource `label`
- [x] `nf-core modules test gcta/makegrmpart --profile docker`
- [x] `nf-core modules test gcta/makegrmpart --profile singularity`
- [x] `nf-core modules test gcta/makegrmpart --profile conda`

### DIFF
--- a/modules/nf-core/gcta/makegrmpart/main.nf
+++ b/modules/nf-core/gcta/makegrmpart/main.nf
@@ -1,0 +1,49 @@
+process GCTA_MAKEGRMPART {
+    tag "part ${meta.part_gcta_job} of ${meta.nparts_gcta} (${meta.id})"
+    label 'process_medium'
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'docker://community.wave.seqera.io/library/gcta:1.94.1--9bc35dc424fcf6e9' :
+        'community.wave.seqera.io/library/gcta:1.94.1--9bc35dc424fcf6e9' }"
+
+    input:
+    tuple val(meta), path(mfile), path(bed_pgen), path(bim_pvar), path(fam_psam)
+    tuple val(meta2), path(snp_group_file)
+
+    output:
+    tuple val(meta), path("*.part_${meta.nparts_gcta}_${meta.part_gcta_job}.grm.id"), path("*.part_${meta.nparts_gcta}_${meta.part_gcta_job}.grm.bin"), path("*.part_${meta.nparts_gcta}_${meta.part_gcta_job}.grm.N.bin"), emit: grm_files
+    tuple val("${task.process}"), val("gcta"), eval("gcta --version 2>&1 | grep 'version v' | tr -s ' ' | cut -d' ' -f3 | sed 's/^v//'"), emit: versions_gcta, topic: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def part_gcta_job = meta.part_gcta_job
+    def nparts_gcta = meta.nparts_gcta
+    def extract_cmd = snp_group_file ? "--extract ${snp_group_file}" : ''
+    def extra_args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def genotype_files = bed_pgen instanceof List ? bed_pgen : [bed_pgen]
+    def genotype_extension = genotype_files[0].name.tokenize('.').last()
+    def multi_file_flag = genotype_extension == 'pgen' ? '--mpfile' : '--mbfile'
+
+    """
+    set -euo pipefail
+
+    gcta \\
+        ${multi_file_flag} ${mfile} \\
+        --make-grm-part ${nparts_gcta} ${part_gcta_job} \\
+        ${extract_cmd} \\
+        --maf 0.01 \\
+        --thread-num ${task.cpus} \\
+        --out ${prefix} ${extra_args}
+    """
+
+    stub:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    touch ${prefix}.part_${meta.nparts_gcta}_${meta.part_gcta_job}.grm.id
+    touch ${prefix}.part_${meta.nparts_gcta}_${meta.part_gcta_job}.grm.bin
+    touch ${prefix}.part_${meta.nparts_gcta}_${meta.part_gcta_job}.grm.N.bin
+    """
+}
--- a/modules/nf-core/gcta/makegrmpart/main.nf
+++ b/modules/nf-core/gcta/makegrmpart/main.nf
@@ -1,25 +1,25 @@
 process GCTA_MAKEGRMPART {
-    tag "part ${meta.part_gcta_job} of ${meta.nparts_gcta} (${meta.id})"
+    tag "part ${part_gcta_job ?: 1} of ${nparts_gcta ?: 1} (${meta.id})"
     label 'process_medium'
     conda "${moduleDir}/environment.yml"
-    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
-        'docker://community.wave.seqera.io/library/gcta:1.94.1--9bc35dc424fcf6e9' :
-        'community.wave.seqera.io/library/gcta:1.94.1--9bc35dc424fcf6e9' }"
+    container "${workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container
+        ? 'docker://community.wave.seqera.io/library/gcta:1.94.1--9bc35dc424fcf6e9'
+        : 'community.wave.seqera.io/library/gcta:1.94.1--9bc35dc424fcf6e9'}"
 
     input:
-    tuple val(meta), path(mfile), path(bed_pgen), path(bim_pvar), path(fam_psam)
+    tuple val(meta), val(nparts_gcta), val(part_gcta_job), path(mfile), path(bed_pgen), path(bim_pvar), path(fam_psam)
     tuple val(meta2), path(snp_group_file)
 
     output:
-    tuple val(meta), path("*.part_${meta.nparts_gcta}_${meta.part_gcta_job}.grm.id"), path("*.part_${meta.nparts_gcta}_${meta.part_gcta_job}.grm.bin"), path("*.part_${meta.nparts_gcta}_${meta.part_gcta_job}.grm.N.bin"), emit: grm_files
-    tuple val("${task.process}"), val("gcta"), eval("gcta --version 2>&1 | grep 'version v' | tr -s ' ' | cut -d' ' -f3 | sed 's/^v//'"), emit: versions_gcta, topic: versions
+    tuple val(meta), path("*.part_${nparts_gcta ?: 1}_${part_gcta_job ?: 1}.grm.id"), path("*.part_${nparts_gcta ?: 1}_${part_gcta_job ?: 1}.grm.bin"), path("*.part_${nparts_gcta ?: 1}_${part_gcta_job ?: 1}.grm.N.bin"), val(nparts_gcta), val(part_gcta_job), emit: grm_files
+    tuple val("${task.process}"), val("gcta"), eval("gcta --version | sed -En 's/^[*] version v([0-9.]*).*/\\1/p'"), emit: versions_gcta, topic: versions
 
     when:
     task.ext.when == null || task.ext.when
 
     script:
-    def part_gcta_job = meta.part_gcta_job
-    def nparts_gcta = meta.nparts_gcta
+    def nparts = nparts_gcta ?: 1
+    def part = part_gcta_job ?: 1
     def extract_cmd = snp_group_file ? "--extract ${snp_group_file}" : ''
     def extra_args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
@@ -32,7 +32,7 @@ process GCTA_MAKEGRMPART {
 
     gcta \\
         ${multi_file_flag} ${mfile} \\
-        --make-grm-part ${nparts_gcta} ${part_gcta_job} \\
+        --make-grm-part ${nparts} ${part} \\
         ${extract_cmd} \\
         --maf 0.01 \\
         --thread-num ${task.cpus} \\
@@ -40,10 +40,12 @@ process GCTA_MAKEGRMPART {
     """
 
     stub:
+    def nparts = nparts_gcta ?: 1
+    def part = part_gcta_job ?: 1
     def prefix = task.ext.prefix ?: "${meta.id}"
     """
-    touch ${prefix}.part_${meta.nparts_gcta}_${meta.part_gcta_job}.grm.id
-    touch ${prefix}.part_${meta.nparts_gcta}_${meta.part_gcta_job}.grm.bin
-    touch ${prefix}.part_${meta.nparts_gcta}_${meta.part_gcta_job}.grm.N.bin
+    touch ${prefix}.part_${nparts}_${part}.grm.id
+    touch ${prefix}.part_${nparts}_${part}.grm.bin
+    touch ${prefix}.part_${nparts}_${part}.grm.N.bin
     """
 }
--- a/modules/nf-core/gcta/makegrmpart/main.nf
+++ b/modules/nf-core/gcta/makegrmpart/main.nf
@@ -11,7 +11,7 @@ process GCTA_MAKEGRMPART {
     tuple val(meta2), path(snp_group_file)
 
     output:
-    tuple val(meta), path("*.part_${nparts}_${part}.grm.id"), path("*.part_${nparts}_${part}.grm.bin"), path("*.part_${nparts}_${part}.grm.N.bin"), val(nparts_gcta), val(part_gcta_job), emit: grm_files
+    tuple val(meta), path("*.part_${nparts}_${part}.grm.*"), val(nparts_gcta), val(part_gcta_job), emit: grm_files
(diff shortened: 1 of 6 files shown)


modules/nf-core/gcta/makegrmpart/main.nf
```diff
@@ -0,0 +1,49 @@
...
+    output:
+    tuple val(meta), path("*.part_${meta.nparts_gcta}_${meta.part_gcta_job}.grm.id"), path("*.part_${meta.nparts_gcta}_${meta.part_gcta_job}.grm.bin"), path("*.part_${meta.nparts_gcta}_${meta.part_gcta_job}.grm.N.bin"), emit: grm_files
+    tuple val("${task.process}"), val("gcta"), eval("gcta --version 2>&1 | grep 'version v' | tr -s ' ' | cut -d' ' -f3 | sed 's/^v//'"), emit: versions_gcta, topic: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def part_gcta_job = meta.part_gcta_job
+    def nparts_gcta = meta.nparts_gcta
```
SPPearce: We don't like adding extra required meta maps, what are these doing here?

modules/nf-core/gcta/makegrmpart/main.nf
```diff
@@ -0,0 +1,49 @@
...
+        'docker://community.wave.seqera.io/library/gcta:1.94.1--9bc35dc424fcf6e9' :
+        'community.wave.seqera.io/library/gcta:1.94.1--9bc35dc424fcf6e9' }"
+
+    input:
+    tuple val(meta), path(mfile), path(bed_pgen), path(bim_pvar), path(fam_psam)
+    tuple val(meta2), path(snp_group_file)
+
+    output:
+    tuple val(meta), path("*.part_${meta.nparts_gcta}_${meta.part_gcta_job}.grm.id"), path("*.part_${meta.nparts_gcta}_${meta.part_gcta_job}.grm.bin"), path("*.part_${meta.nparts_gcta}_${meta.part_gcta_job}.grm.N.bin"), emit: grm_files
+    tuple val("${task.process}"), val("gcta"), eval("gcta --version 2>&1 | grep 'version v' | tr -s ' ' | cut -d' ' -f3 | sed 's/^v//'"), emit: versions_gcta, topic: versions
```
SPPearce: Can you simplify this version extraction, you should be able to do it in one command rather than grep, tr, cut, sed.
LLMs are very good at simplifying these.
Also do you need the 2>&1 redirect (depends on whether the tool prints to stderr or stdout)

modules/nf-core/gcta/makegrmpart/main.nf
lyh970817: These are for parallelisation - splitting the computation into multiple parts. `nparts_gcta` is the total number of parts (splits) and `part_gcta_job` is the part currently being computed.

I can move them as values to the input section and group them in a tuple?

famosab: You can also take a look at other modules usually we do the parallelization based on task.ext.cpus

modules/nf-core/gcta/makegrmpart/main.nf
```diff
@@ -0,0 +1,49 @@
+process GCTA_MAKEGRMPART {
+    tag "part ${meta.part_gcta_job} of ${meta.nparts_gcta} (${meta.id})"
```
famosab:
```suggestion
    tag "${meta.id}"
```

modules/nf-core/gcta/makegrmpart/main.nf
famosab: also I am a little confused, where does the splitting happen. This sort of functionality seems to be something that needs to be handled in the pipeline or subworkflow level

famosab: This is probably related to here:
https://yanglab.westlake.edu.cn/software/gcta/#Tutorial

I would say this is something that should be handled in an extra subworkflow. And the module should be agnostic. It could be that someone has a very small input file which then means that they do not need parts (hypothetically)

lyh970817: `nparts_gcta` is intended to be a user-supplied parameter for the number of parts to split the computation into, and `part_gcta_job` is the current split being computed.

I do have a subworkflow for another PR that creates the input channel based on `nparts_gcta` for the module. Should I submit a PR for that now as well for review?

I can have a routing logic within the module to allow for running the module without splitting into parts. There are different ways of parallelisation - the by part method is to split by groups of samples (individuals) and the link points to the method of splitting by chromosome. There is also configuring the number of cpus  - I'll add this.

modules/nf-core/gcta/makegrmpart/main.nf
lyh970817: The process is intended for a channel with multiple elements (with different `part_gcta_job`). Is it okay to use a shared `meta.id` for them?

modules/nf-core/gcta/makegrmpart/main.nf
jfy133: > We don't like adding extra required meta maps, what are these doing here?

Yes, custom meta entries are explicitly not allowed as per the module specs : https://nf-co.re/docs/specifications/components/modules/general#types-of-meta-map-keys

modules/nf-core/gcta/makegrmpart/main.nf
SPPearce: I think it is reasonable to set the tag using the part_gct_job and nparts_gcta, but not from the meta map (as discussed)

SPPearce:
I think the correct thing here is to have the total number of parts and the current part to run as additional inputs inside the tuple (with the meta).
They can be extracted from the meta just before the module is ran. If they aren't supplied, they could both default to 1, so that the case of the small testdataset is covered.

lyh970817:
> I think the correct thing here is to have the total number of parts and the current part to run as additional inputs inside the tuple (with the meta). They can be extracted from the meta just before the module is ran. If they aren't supplied, they could both default to 1, so that the case of the small testdataset is covered.

I see. Thanks for the suggestion. Do you mean something like `tuple val(meta), val(nparts_gcta), val(part_gcta_job), path(mfile), path(bed_pgen), path(bim_pvar), path(fam_psam)`?

I thought the only `val()` allowed in the tuple is `val(meta)` and the other elements should be file paths. Is this an exception made for this module or can I use it for my other modules?

SPPearce:
> > I think the correct thing here is to have the total number of parts and the current part to run as additional inputs inside the tuple (with the meta). They can be extracted from the meta just before the module is ran. If they aren't supplied, they could both default to 1, so that the case of the small testdataset is covered.
> 
> I see. Thanks for the suggestion. Do you mean something like `tuple val(meta), val(nparts_gcta), val(part_gcta_job), path(mfile), path(bed_pgen), path(bim_pvar), path(fam_psam)`?
> 
> I thought the only `val()` allowed in the tuple is `val(meta)` and the other elements should be file paths. Is this an exception made for this module or can I use it for my other modules?

You can have multiple `val` input in the tuple if appropriate, other modules do that.
I might be tempted to put it at the end of the tuple, but that it is often difficult to figure out how to order things (until we get named inputs).

lyh970817:
> > > I think the correct thing here is to have the total number of parts and the current part to run as additional inputs inside the tuple (with the meta). They can be extracted from the meta just before the module is ran. If they aren't supplied, they could both default to 1, so that the case of the small testdataset is covered.
> > 
> > 
> > I see. Thanks for the suggestion. Do you mean something like `tuple val(meta), val(nparts_gcta), val(part_gcta_job), path(mfile), path(bed_pgen), path(bim_pvar), path(fam_psam)`?
> > I thought the only `val()` allowed in the tuple is `val(meta)` and the other elements should be file paths. Is this an exception made for this module or can I use it for my other modules?
> 
> You can have multiple `val` input in the tuple if appropriate, other modules do that. I might be tempted to put it at the end of the tuple, but that it is often difficult to figure out how to order things (until we get named inputs).

Thanks! Can I similarly have `val` in the output tuples along with the file path?

SPPearce:
> Thanks! Can I similarly have val in the output tuples along with the file path?
I think that is fine.

SPPearce:
@lyh970817 , do you need anything more before you modify this one?

lyh970817:
No, thanks. I’ve updated this now based on the discussion: partition values are explicit `val(...)` inputs and output context, custom `meta` keys are removed, and the small-input case defaults internally to `1/1`.

modules/nf-core/gcta/makegrmpart/main.nf
lyh970817: Updated. `nparts_gcta` and `part_gcta_job` are now explicit `val(...)` tuple inputs instead of custom `meta` keys, and `meta` is back to standard keys only.

The caller handles the split/fan-out; this module runs one partition per tuple. For the non-split case, both values now default to `1` internally when `null` is passed.

I also carry the partition context at the end of the `grm_files` output tuple with `val(...)`, and added a `1/1` test.

modules/nf-core/gcta/makegrmpart/main.nf
lyh970817: Updated. I simplified this to a single `gcta --version | sed -En ...` command and removed the `2>&1 | grep | tr | cut` chain.

modules/nf-core/gcta/makegrmpart/main.nf
lyh970817: Updated. The tag still includes the part context, but it now uses the explicit partition values rather than custom `meta` entries.

modules/nf-core/gcta/makegrmpart/main.nf
```diff
@@ -0,0 +1,51 @@
...
+    container "${workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container
+        ? 'docker://community.wave.seqera.io/library/gcta:1.94.1--9bc35dc424fcf6e9'
+        : 'community.wave.seqera.io/library/gcta:1.94.1--9bc35dc424fcf6e9'}"
+
+    input:
+    tuple val(meta), val(nparts_gcta), val(part_gcta_job), path(mfile), path(bed_pgen), path(bim_pvar), path(fam_psam)
+    tuple val(meta2), path(snp_group_file)
+
+    output:
+    tuple val(meta), path("*.part_${nparts_gcta ?: 1}_${part_gcta_job ?: 1}.grm.id"), path("*.part_${nparts_gcta ?: 1}_${part_gcta_job ?: 1}.grm.bin"), path("*.part_${nparts_gcta ?: 1}_${part_gcta_job ?: 1}.grm.N.bin"), val(nparts_gcta), val(part_gcta_job), emit: grm_files
```
SPPearce: You should be able to do this if you remove the `def` from below.
```suggestion
    tuple val(meta), path("*.part_${nparts}_${part}.grm.id"), path("*.part_${nparts}_${part}.grm.bin"), path("*.part_${nparts}_${part}.grm.N.bin"), val(nparts_gcta), val(part_gcta_job), emit: grm_files
```

modules/nf-core/gcta/makegrmpart/main.nf
```diff
@@ -0,0 +1,51 @@
...
+    tuple val("${task.process}"), val("gcta"), eval("gcta --version | sed -En 's/^[*] version v([0-9.]*).*/\\1/p'"), emit: versions_gcta, topic: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def nparts = nparts_gcta ?: 1
+    def part = part_gcta_job ?: 1
```
SPPearce:
```suggestion
    nparts = nparts_gcta ?: 1
    part = part_gcta_job ?: 1
```

modules/nf-core/gcta/makegrmpart/meta.yml
```diff
@@ -0,0 +1,125 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/meta-schema.json
+name: "gcta_makegrmpart"
+description: Compute one partition of a GCTA genetic relationship matrix
+keywords:
+  - gcta
+  - grm
+  - genetics
```
SPPearce, line 9: Can you add keywords where you spell out the acronyms, for those who don't know what gcta/grm are.
Anything else that would be potential search targets?

modules/nf-core/gcta/makegrmpart/main.nf
lyh970817: Updated. The output path patterns now use the defaulted `nparts`/`part` values, while `nparts_gcta` and `part_gcta_job` remain appended at the end as `val(...)` output context.

modules/nf-core/gcta/makegrmpart/main.nf
lyh970817: Updated. Removed `def` in both `script:` and `stub:` so `nparts` and `part` are visible to the output declarations.

modules/nf-core/gcta/makegrmpart/tests/nextflow.config
SPPearce, line 1: You shouldn't need this.

modules/nf-core/gcta/makegrmpart/main.nf
```diff
@@ -0,0 +1,51 @@
...
+    def extract_cmd = snp_group_file ? "--extract ${snp_group_file}" : ''
+    def extra_args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def genotype_files = bed_pgen instanceof List ? bed_pgen : [bed_pgen]
+    def genotype_extension = genotype_files[0].name.tokenize('.').last()
+    def multi_file_flag = genotype_extension == 'pgen' ? '--mpfile' : '--mbfile'
+
+    """
+    set -euo pipefail
+
```
SPPearce: This should be set automatically by the nextflow.config
```suggestion
```

modules/nf-core/gcta/makegrmpart/main.nf
```diff
@@ -0,0 +1,51 @@
...
+    gcta \\
+        ${multi_file_flag} ${mfile} \\
+        --make-grm-part ${nparts} ${part} \\
+        ${extract_cmd} \\
+        --maf 0.01 \\
```
SPPearce: Why are we hardcoding this?
Is a value required, does it have a default value?

SPPearce:
I think this is basically ready, just a couple of minor things

modules/nf-core/gcta/makegrmpart/main.nf
lyh970817: Removed in 74b18d92e. `--maf` is not required for `--make-grm-part`, so the module no longer hardcodes an analysis filter. Callers can still pass something like `--maf 0.01` through `task.ext.args` when they want that filter.

modules/nf-core/gcta/makegrmpart/main.nf
lyh970817: Removed in 74b18d92e. I dropped the redundant shell safety line because Nextflow task wrappers already handle task shell failure behavior.

modules/nf-core/gcta/makegrmpart/main.nf
```diff
@@ -0,0 +1,48 @@
+process GCTA_MAKEGRMPART {
+    tag "${meta.id}: part ${part_gcta_job ?: 1} of ${nparts_gcta ?: 1}"
+    label 'process_medium'
+    conda "${moduleDir}/environment.yml"
+    container "${workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container
+        ? 'docker://community.wave.seqera.io/library/gcta:1.94.1--9bc35dc424fcf6e9'
```
SPPearce: This should be a singularity image, preferably with the http link rather than the oras link. There is a toggle on Seqera Containers page. 
```suggestion
        ? 'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/46/46b0d05f0daa47561d87d2a9cac5e51edc2c78e26f1bbab439c688386241a274/data'
```

modules/nf-core/gcta/makegrmpart/main.nf
```diff
@@ -0,0 +1,48 @@
...
+        : 'community.wave.seqera.io/library/gcta:1.94.1--9bc35dc424fcf6e9'}"
+
+    input:
+    tuple val(meta), val(nparts_gcta), val(part_gcta_job), path(mfile), path(bed_pgen), path(bim_pvar), path(fam_psam)
+    tuple val(meta2), path(snp_group_file)
+
+    output:
+    tuple val(meta), path("*.part_${nparts}_${part}.grm.id"), path("*.part_${nparts}_${part}.grm.bin"), path("*.part_${nparts}_${part}.grm.N.bin"), val(nparts_gcta), val(part_gcta_job), emit: grm_files
```
SPPearce: Up to you, but I suspect that you don't ever need to seperate out these files. In which case, you could emit them all as one part of the tuple rather than splitting them out:

```suggestion
    tuple val(meta), path("*.part_${nparts}_${part}.*"), val(nparts_gcta), val(part_gcta_job), emit: grm_files
```

modules/nf-core/gcta/makegrmpart/main.nf
```diff
@@ -0,0 +1,48 @@
+process GCTA_MAKEGRMPART {
+    tag "${meta.id}: part ${part_gcta_job ?: 1} of ${nparts_gcta ?: 1}"
```
SPPearce:
```suggestion
    tag "${meta.id}: part ${part_gcta_job} of ${nparts_gcta}"
```

SPPearce (approved):
Two minor changes and a question, but otherwise I think this is fine now, thank you for your patience.

PR #11008, opened 2026-03-21
Add regenie/step1 module

lyh970817:
This PR adds the `regenie/step1` module, which runs REGENIE Step 1 whole-genome regression to fit LOCO ridge models for downstream association testing workflows.

Merge dependencies: none. This PR is self-contained and does not require any other split REGENIE PR to merge first.

## PR checklist

- [x] Add a resource `label`
- [x] `nf-core modules test regenie/step1 --profile docker`
- [x] `nf-core modules test regenie/step1 --profile singularity`
- [x] `nf-core modules test regenie/step1 --profile conda`


modules/nf-core/regenie/step1/tests/nextflow.config
famosab, line 1: This file should not be needed

modules/nf-core/regenie/step1/main.nf
```diff
@@ -0,0 +1,55 @@
+process REGENIE_STEP1 {
+    tag "${meta.id}:${meta2.pheno_col}"
```
famosab:
```suggestion
    tag "${meta.id}"
```

modules/nf-core/regenie/step1/main.nf
```diff
@@ -0,0 +1,55 @@
...
+        : 'community.wave.seqera.io/library/regenie:4.1.2--5d361f9fcb2f85cf'}"
+
+    input:
+    tuple val(meta), path(plink_genotype_file), path(plink_variant_file), path(plink_sample_file)
+    tuple val(meta2), path(pheno)
+    tuple val(meta3), path(covar)
+    val(bsize)
+
+    output:
+    tuple val(meta2), path("*_pred.list"), path("*.loco.gz"), emit: predictions
```
famosab: one output per file

modules/nf-core/regenie/step1/main.nf
```diff
@@ -0,0 +1,55 @@
...
+    tuple val(meta2), path("*.log"), emit: log
```
famosab:
```suggestion
    tuple val(meta), path("*.log"), emit: log
```

modules/nf-core/regenie/step1/main.nf
```diff
@@ -0,0 +1,55 @@
...
+
+    script:
+    def args = task.ext.args ?: ''
+    def binary_arg = meta2.is_binary ? '--bt' : ''
+    def covar_arg = covar ? "--covarFile ${covar}" : ''
+    def pheno_col = meta2.pheno_col
+    def genotype_flag = plink_genotype_file.name.endsWith('.pgen') ? '--pgen' : '--bed'
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def bsize_arg = bsize ?: 1000
+
```
famosab:
```suggestion
```

modules/nf-core/regenie/step1/main.nf
```diff
@@ -0,0 +1,55 @@
...
+        --threads ${task.cpus} \\
+        ${args} \\
+        --out ${prefix}
+    """
+
+    stub:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    touch ${prefix}_pred.list
+    printf '' | gzip > ${prefix}_1.loco.gz
```
famosab:
```suggestion
    echo | gzip > ${prefix}_1.loco.gz
```

modules/nf-core/regenie/step1/main.nf
```diff
@@ -0,0 +1,55 @@
...
+    input:
+    tuple val(meta), path(plink_genotype_file), path(plink_variant_file), path(plink_sample_file)
+    tuple val(meta2), path(pheno)
+    tuple val(meta3), path(covar)
+    val(bsize)
+
+    output:
+    tuple val(meta2), path("*_pred.list"), path("*.loco.gz"), emit: predictions
+    tuple val(meta2), path("*.log"), emit: log
+    tuple val("${task.process}"), val('regenie'), eval("regenie --version 2>&1 | head -n 1"), topic: versions, emit: versions_regenie
```
famosab:
```suggestion
    tuple val("${task.process}"), val('regenie'), eval("regenie --version | head -n 1"), topic: versions, emit: versions_regenie
```

famosab: and please update the statement so only the bare version ends up in the output
in this case 4.1.2

famosab (commented):
Looks good a few comments

modules/nf-core/regenie/step1/main.nf
lyh970817: Here `meta.id` is only the identifier for the genetic file, but a single genetic file can be paired with multiple phenotypes. Is it okay to only include the genetic file identifier in the tag?

modules/nf-core/regenie/step1/tests/main.nf.test.snap
```diff
@@ -0,0 +1,153 @@
...
+                    },
+                    "plink_simulated_1.loco.gz:md5,93aa1d97d8f164e57cd8fb2551a482f3"
+                ]
+            ],
+            {
+                "versions_regenie": [
+                    [
+                        "REGENIE_STEP1",
+                        "regenie",
+                        "v4.1.2.gz"
```
mashehu:
```suggestion
                        "4.1.2"
```
you need to strip the `v` and `.gz` in the version extraction command

modules/nf-core/regenie/step1/main.nf
famosab: I think we can have both but then I would also use the id of the phenotype 🤔

lyh970817: I set the id of the phenotype (`meta2.id`) to be the genotype id (so the same as `meta.id`), as I conceptualise it as representing sample id - different phenotypes can be used but they must come from the same individuals with the same genotypes. So later if I have multiple genotypes and multiple phenotypes I know which corresponds to which.

I'm quite uncertain of the approach to take. What would you suggest the design to be?

famosab: I see your issue, however we do not allow custom meta map keys, so we need to find another solution here. If you have two inputs you could map the meta2.id. The id is recording unique file identifiers associated with a file. So the meta.id should be unique to the file its matched to.

modules/nf-core/regenie/step1/main.nf
lyh970817: Updated: the reported REGENIE version is now normalized to `4.1.2`.

modules/nf-core/regenie/step1/main.nf
lyh970817: Applied: the stub now creates the LOCO output with `echo | gzip`.

modules/nf-core/regenie/step1/tests/main.nf.test.snap
lyh970817: Updated: the snapshot now matches the normalized bare version output.

modules/nf-core/regenie/step1/main.nf
lyh970817: Updated: the process tag now uses `meta.id`. `meta2.id` and `meta3.id` also represent genotype id (so the same as `meta.id`).

modules/nf-core/regenie/step1/main.nf
lyh970817: Updated: the `log` output now uses the genotype `meta` tuple.

modules/nf-core/regenie/step1/main.nf
lyh970817: Applied.

modules/nf-core/regenie/step1/tests/nextflow.config
lyh970817: Updated: I removed `modules/nf-core/regenie/step1/tests/nextflow.config` from the PR diff and kept the local helper copy out of the PR.

modules/nf-core/regenie/step1/main.nf
lyh970817: Updated: split the combined `predictions` output so `*_pred.list` stays on `predictions` and `*.loco.gz` is emitted separately on `loco`.

SPPearce:
@lyh970817 , are you waiting on a review here?

lyh970817:
> @lyh970817 , are you waiting on a review here?

Yes. It's good for a review. Thanks!

modules/nf-core/regenie/step1/main.nf
```diff
@@ -0,0 +1,54 @@
...
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def binary_arg = is_binary ? '--bt' : ''
+    def covar_arg = covar ? "--covarFile ${covar}" : ''
+    def genotype_flag = plink_genotype_file.name.endsWith('.pgen') ? '--pgen' : '--bed'
+    def prefix = plink_genotype_file.baseName
```
SPPearce: We usually allow this to be overridden via `ext.prefix`

modules/nf-core/regenie/step1/main.nf
```diff
@@ -0,0 +1,54 @@
...
+    tuple val(meta), path("*.loco.gz"), val(pheno_col), val(is_binary), emit: loco
+    tuple val(meta), path("*.log"), val(pheno_col), val(is_binary), emit: log
+    tuple val("${task.process}"), val('regenie'), eval('regenie --version 2>&1 | sed -n "1{s/^v//;s/\\.gz$//;p}"'), topic: versions, emit: versions_regenie
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def binary_arg = is_binary ? '--bt' : ''
```
SPPearce: If this is not essential input for the tool, it would normally be passed via ext.args

lyh970817: I anticipate use cases where the user can supply multiple phenotype files in a channel,  and each phenotype file could be continuous or binary. If `--bt` were supplied via `ext.args`, all the phenotype files would have to be analysed all as continuous or binary.

modules/nf-core/regenie/step1/main.nf
lyh970817: Thanks. I've addressed this.

modules/nf-core/regenie/step1/main.nf
SPPearce: No, they can still be set via a closure:
`ext.args = { meta.bt ? "--bt" : "" }`

lyh970817: Thanks. It's now fixed. There was another val `pheno_col` which is also optional and I removed that too.

SPPearce (approved):

PR #9765, opened 2026-01-23
feat: Add emit statements to support Nextflow workflow outputs

pinin4fjords:
## Summary

Add individual output channel emits to subworkflows to support the new Nextflow workflow outputs syntax (25.10+). These changes allow pipelines to publish outputs centrally rather than via `publishDir` directives.

This PR ports the subworkflow changes from nf-core/rnaseq#1679 to the modules repo.

## Changes

**Updated subworkflows:**

| Subworkflow | New Emits |
|-------------|-----------|
| bam_dedup_stats_samtools_umitools | `tsv_edit_distance`, `tsv_per_umi`, `tsv_umi_per_position` |
| bam_dedup_umi | `tsv_edit_distance`, `tsv_per_umi`, `tsv_umi_per_position` (when using umitools) |
| fastq_fastqc_umitools_fastp | `umi_reads` |
| fastq_fastqc_umitools_trimgalore | `umi_reads` |
| fastq_qc_trim_filter_setstrandedness | 20+ individual output channels for QC/trim/filter outputs |
| fastq_remove_rrna | `sortmerna_log`, `ribodetector_log`, `seqkit_stats`, `bowtie2_log`, `bowtie2_index`, `seqkit_prefixed`, `seqkit_converted` |
| quantify_pseudo_alignment | `tx2gene` |

## Test plan

- [ ] Snapshot updates needed for tests using `snapshot(workflow.out)`:
  - bam_dedup_stats_samtools_umitools (1 test)
  - fastq_fastqc_umitools_fastp (8 tests)
  - fastq_fastqc_umitools_trimgalore (3 tests)
  - quantify_pseudo_alignment (2 tests)

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #11024, opened 2026-03-23
update subworkflows for differential and functional enrichment analysis

suzannejin:
Update subworkflows for differential and functional enrichment analysis.
This update is needed for the workflow output migration [here](https://github.com/nf-core/differentialabundance/pull/662)

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


subworkflows/nf-core/differential_functional_enrichment/main.nf
```diff
@@ -82,63 +84,141 @@ workflow DIFFERENTIAL_FUNCTIONAL_ENRICHMENT {
...
+                .join(GSEA_GSEA.out.report_tsvs_target)
+        )
+
+    // other results
+    ch_artifacts = ch_artifacts
+        .mix(
+            GSEA_GSEA.out.rpt
+                .mix(GSEA_GSEA.out.index_html)
+                .mix(GSEA_GSEA.out.heat_map_corr_plot)
+                .mix(GSEA_GSEA.out.report_tsvs_ref)
```
pinin4fjords: Did you intend to mix these here as well as above?

subworkflows/nf-core/differential_functional_enrichment/main.nf
```diff
@@ -152,23 +232,12 @@ workflow DIFFERENTIAL_FUNCTIONAL_ENRICHMENT {
...
-    // gsea-specific outputs
-    gsea_report           = GSEA_GSEA.out.report_tsvs_ref.join(GSEA_GSEA.out.report_tsvs_target)
-
-    // decoupler-specific outputs
-    decoupler_dc_estimate = DECOUPLER_DECOUPLER.out.dc_estimate
-    decoupler_dc_pvals = DECOUPLER_DECOUPLER.out.dc_pvals
-    decoupler_png = DECOUPLER_DECOUPLER.out.png
-
+    // main results
+    results = ch_results
```
pinin4fjords: I'm unconvinced by outputting everything in a single channel this way. You're mixing channel 'arity' and forcing anyone who wants to access specific channels to sort though a heterogenous set. Can we leave these separate please, and defer any combinations to the workflow?

pinin4fjords:
e.g. here's how I prepared subworkflows for workflow output syntax: https://github.com/nf-core/modules/pull/9765

Essentially just making sure that everything was available via channels. No need to combine everything artificially into a small number of outputs.

subworkflows/nf-core/differential_functional_enrichment/main.nf
suzannejin: yes.. i was trying to see if i can simplify, but didnt like it either. reverting back to the previous commit now.

suzannejin:
hi @pinin4fjords , sorry for the confusion, i was trying to see if extra simplification works, but it did not. 
Just reverted back to the previous commit, this should work for the workflow output migration purpose.

subworkflows/nf-core/abundance_differential_filter/main.nf
```diff
@@ -195,6 +195,15 @@ workflow ABUNDANCE_DIFFERENTIAL_FILTER {
         .mix(DESEQ2_NORM.out.vst_counts)
         .groupTuple()
 
+    ch_plots = DESEQ2_DIFFERENTIAL.out.dispersion_plot_png
+        .mix(LIMMA_DIFFERENTIAL.out.md_plot)
+
+    ch_other = DESEQ2_DIFFERENTIAL.out.rdata
```
pinin4fjords: Sorry to be a pain, thinking about this we should probably emit disparate things separately, the 'other' doesn't really work. If I want the session info in a consuming workflow I don't want to have to filter it out.

I'm inclined to say the same about the plots unless they're equivalent plots (which I don't think is the case here).

suzannejin: I splitted the output for the ABUNDANCE_DIFFERENTIAL_FILTER subworkflow.
Now, for the GSEA artifacts, you want to emit a different output for each of the 21 artifacts?
That's a lot of duplicated verbosity to deal with at all subworkflow, workflow levels.

pinin4fjords: It just doesn't make sense to combine disparate things into channels. I agree the current implementation makes things awkward- it's why I put this PR on hold for rnaseq: https://github.com/nf-core/rnaseq/pull/1679

Nextflow will have some improvements, notably record types, that may help this soon. It's fine to put this on pause until that happens if you feel the same here as I did for rnaseq.

suzannejin: The migration here is mainly needed to solve this [issue](https://github.com/nf-core/differentialabundance/issues/471) affecting pipeline cache.
That's why I wouldn't pause it for this PR. I'd vote to have a less perfect subworkflow output structure for the moment (it can always be adapted later on).

pinin4fjords: Could you explain why separating the outputs would block that work?

suzannejin: What meant is that pausing the migration will block that work, not splitting the output itself.

However, the entire idea about pipeline hub is that alternative methods (in this case functional enrichment) can be exchanged easily. This means adding a new method will ideally only require changes in the subworkflow, not the pipeline itself. By splitting outputs, instead of classifying outputs into the same category, will make this even more difficult. 

But yes, i agree that having them in artifacts/other is also not ideal. I will split the outputs as temporal solution. With syntax improvements in the future this may be better addressed.

pinin4fjords: When different methods have clear analogs, putting them in the same channel makes sense (perhaps after some normalisation of the structure- for future). But combining e.g. size factors and session info will never make sense at the component level.

subworkflows/nf-core/differential_functional_enrichment/main.nf
```diff
@@ -153,19 +157,44 @@ workflow DIFFERENTIAL_FUNCTIONAL_ENRICHMENT {
     // nf-core/differentialabundance pipeline
 
     // gprofiler2-specific outputs
+    gprofiler2_plot_html  = GPROFILER2_GOST.out.plot_html
     gprofiler2_all_enrich = GPROFILER2_GOST.out.all_enrich
     gprofiler2_sub_enrich = GPROFILER2_GOST.out.sub_enrich
-    gprofiler2_plot_html  = GPROFILER2_GOST.out.plot_html
+    gprofiler2_artifacts = GPROFILER2_GOST.out.plot_png
```
pinin4fjords: Same deal with these things. Arguably makes sense to group e.g. plots and reports, but we shouldn't be combining all the disparate things

pinin4fjords: To be fair, I know why you want to do this- I wanted something similar when I opened https://github.com/nextflow-io/nextflow/issues/6756, i.e. to have channels mapping directly to publish locations. 

But we can't make assumptions in components about how pipelines will want to publish.

suzannejin: ok nice it looks like with records, things will become easier.

PR #11033, opened 2026-03-24
Add new module: telogator2

pinin4fjords:
## Summary

- Add nf-core module for [telogator2](https://github.com/zstephens/telogator2), a tool for allele-specific telomere length (TL) estimation and telomere variant repeat (TVR) characterization from long-read sequencing data (ONT/PacBio)
- Uses bioconda package `telogator2=2.2.3` with Wave-built Docker and Singularity containers
- Accepts BAM/CRAM input with optional reference genome, uses topic-based version reporting

## Inputs

| Input | Type | Description |
|-------|------|-------------|
| reads | BAM/CRAM | Long-read alignments (PacBio HiFi or ONT) |
| reads_index | BAI/CRAI | Index for reads |
| fasta + fai | FASTA + FAI | Optional reference genome (single channel) |

## Outputs

| Output | Type | Description |
|--------|------|-------------|
| tlens | TSV | Telomere lengths per allele (`tlens_by_allele.tsv`) |
| plots | PNG | Allele and violin plots (optional) |
| qc | Directory | QC stats, read length distributions, run metadata |

## Notes

- The tool exits with code 1 when no telomere reads are found in the input, with a clear error message. This is not caught by the module - pipelines should handle this if mixed samples are expected.
- Test data uses real ONT telomeric reads from GIAB HG002 (nf-core/test-datasets#1947), plus a PacBio BAM without telomere reads to test the failure path.
- TL values are non-deterministic across runs (minimap2 alignment), so tests assert the TSV header structure and output existence rather than exact md5 checksums. A fixed random seed (`--rng 42`) is set in the test config.

## Test plan

- [x] `nf-test` passes for all 3 tests (telomere reads, no telomere reads, stub) with `--profile docker`
- [x] Checked stderr for silent errors - none found
- [ ] CI nf-test validation

### DIFF
--- a/modules/nf-core/telogator2/main.nf
+++ b/modules/nf-core/telogator2/main.nf
@@ -12,9 +12,13 @@ process TELOGATOR2 {
     tuple val(meta2), path(fasta), path(fai)
 
     output:
-    tuple val(meta), path("${prefix}/tlens_by_allele.tsv"), emit: tlens
-    tuple val(meta), path("${prefix}/*.png")              , emit: plots
-    tuple val(meta), path("${prefix}/qc")                 , emit: qc
+    tuple val(meta), path("${prefix}/tlens_by_allele.tsv")  , emit: tlens
+    tuple val(meta), path("${prefix}/*.png")                , emit: plots
+    tuple val(meta), path("${prefix}/qc/cmd.txt")           , emit: cmd
+    tuple val(meta), path("${prefix}/qc/stats.txt")         , emit: stats
+    tuple val(meta), path("${prefix}/qc/qc_readlens.png")   , emit: qc_readlens
+    tuple val(meta), path("${prefix}/qc/readlens.npz")      , emit: readlens
+    tuple val(meta), path("${prefix}/qc/rng.txt")           , emit: rng
     tuple val("${task.process}"), val('telogator2'), eval("telogator2 --version | sed 's/telogator2 //'"), emit: versions_telogator2, topic: versions
 
     when:
@@ -40,6 +44,10 @@ process TELOGATOR2 {
     touch ${prefix}/tlens_by_allele.tsv
     touch ${prefix}/all_final_alleles.png
     touch ${prefix}/violin_atl.png
+    touch ${prefix}/qc/cmd.txt
+    touch ${prefix}/qc/qc_readlens.png
+    touch ${prefix}/qc/readlens.npz
+    touch ${prefix}/qc/rng.txt
     touch ${prefix}/qc/stats.txt
     """
 }
--- a/modules/nf-core/telogator2/tests/main.nf.test
+++ b/modules/nf-core/telogator2/tests/main.nf.test
@@ -32,11 +32,14 @@ nextflow_process {
                 { assert path(process.out.tlens.get(0).get(1)).readLines().first() ==
                     '#chr\tposition\tref_samp\tallele_id\tTL_p75\tread_TLs\tread_lengths\tread_mapq\ttvr_len\ttvr_consensus\tsupporting_reads' },
                 { assert process.out.plots },
-                { assert process.out.qc },
                 { assert snapshot(
                     file(process.out.tlens.get(0).get(1)).name,
                     process.out.plots.collect { meta, pngs -> pngs.collect { png -> file(png).name } },
-                    process.out.qc.collect { meta, dir -> file(dir).list().sort() },
+                    process.out.cmd,
+                    process.out.stats,
+                    process.out.qc_readlens,
+                    process.out.readlens,
+                    process.out.rng,
                     process.out.findAll { key, val -> key.startsWith('versions') }
                 ).match() }
             )
--- a/modules/nf-core/telogator2/meta.yml
+++ b/modules/nf-core/telogator2/meta.yml
@@ -77,15 +77,61 @@ output:
           pattern: "*/*.png"
           ontologies:
             - edam: http://edamontology.org/format_3603
-  qc:
+  cmd:
     - - meta:
           type: map
           description: |
             Groovy Map containing sample information
             e.g. [ id:'test' ]
-      - ${prefix}/qc:
-          type: directory
-          description: Directory containing QC statistics, read length distributions, and run metadata
+      - ${prefix}/qc/cmd.txt:
+          type: file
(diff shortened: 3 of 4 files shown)


pinin4fjords:
~~TODO: Generate telomere-containing long-read test data so the tests exercise telogator2's actual analysis path, not just the "no telomere reads found" fallback.~~

modules/nf-core/telogator2/tests/nextflow.config
```diff
@@ -0,0 +1,5 @@
+process {
+    withName: 'TELOGATOR2' {
+        ext.args = '--rng 42'
```
fellen31: Could use module args: https://nf-co.re/docs/guidelines/components/modules#configuration-of-extargs-in-tests

modules/nf-core/telogator2/tests/main.nf.test
```diff
@@ -0,0 +1,86 @@
...
+        then {
+            assertAll(
+                { assert process.success },
+                { assert path(process.out.tlens.get(0).get(1)).readLines().first() ==
+                    '#chr\tposition\tref_samp\tallele_id\tTL_p75\tread_TLs\tread_lengths\tread_mapq\ttvr_len\ttvr_consensus\tsupporting_reads' },
+                { assert process.out.plots },
+                { assert process.out.qc },
+                { assert snapshot(
+                    process.out.findAll { key, val -> key.startsWith('versions') }
+                ).match() }
```
fellen31, line 44: Perhaps you could snapshot at least the names of the output files? So it's clear it matches the stub for example.

modules/nf-core/telogator2/main.nf
```diff
@@ -0,0 +1,45 @@
...
+        ${args}
+    """
+
+    stub:
+    prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    mkdir -p ${prefix}/qc
+    touch ${prefix}/tlens_by_allele.tsv
+    touch ${prefix}/all_final_alleles.png
+    touch ${prefix}/violin_atl.png
```
fellen31, line 46: Looks like the png files are optional, could you touch these in the stub only when they should be output?

modules/nf-core/telogator2/tests/nextflow.config
pinin4fjords: Done - switched to the `params.module_args` pattern. The config now just has `ext.args = params.module_args`, and each test sets `params { module_args = '--rng 42' }` in the `when` block.

modules/nf-core/telogator2/tests/main.nf.test
pinin4fjords: Good idea. The telomere test now snapshots output file names (`tlens_by_allele.tsv`, `all_final_alleles.png`, `violin_atl.png`, qc directory listing) plus versions. We can't snapshot md5s for the main outputs because telogator2 results are non-deterministic across runs (minimap2 alignment), but we also assert the TSV header structure to verify the output format is correct.

modules/nf-core/telogator2/main.nf
pinin4fjords: Checked the telogator2 CLI - the two main plots (`all_final_alleles.png`, `violin_atl.png`) are always produced on a successful run and there's no flag to disable them. So they're not truly optional - removed `optional: true` and added them back to the stub.

modules/nf-core/telogator2/main.nf
```diff
@@ -0,0 +1,45 @@
...
+        : 'community.wave.seqera.io/library/telogator2:2.2.3--01b2748e09721f3b' }"
+
+    input:
+    tuple val(meta), path(reads), path(reads_index)
+    tuple val(meta2), path(fasta), path(fai)
+
+    output:
+    tuple val(meta), path("${prefix}/tlens_by_allele.tsv"), emit: tlens
+    tuple val(meta), path("${prefix}/*.png")              , emit: plots
+    tuple val(meta), path("${prefix}/qc")                 , emit: qc
```
fellen31: Since there are only 5 QC-files perhaps they could be output into separate channels. But, I'm not sure you would want to do anything with them downstream, so I'm fine with this.

modules/nf-core/telogator2/main.nf
```diff
@@ -0,0 +1,45 @@
...
+    """
+
+    stub:
+    prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    mkdir -p ${prefix}/qc
+    touch ${prefix}/tlens_by_allele.tsv
+    touch ${prefix}/all_final_alleles.png
+    touch ${prefix}/violin_atl.png
+    touch ${prefix}/qc/stats.txt
```
fellen31, line 51: If these are always output, I would always touch them in the stub.                
```
                    "cmd.txt",
                    "qc_readlens.png",
                    "readlens.npz",
                    "rng.txt",
                    "stats.txt"
```

modules/nf-core/telogator2/tests/nextflow.config
fellen31: 👍

modules/nf-core/telogator2/main.nf
fellen31: Alright!

modules/nf-core/telogator2/tests/main.nf.test
fellen31: Nice!

modules/nf-core/telogator2/main.nf
pinin4fjords: I did consider it and picked this way, but don't have strong feelings, so can do that.

pinin4fjords: Done - split the QC directory into 5 individual output channels: `cmd`, `stats`, `qc_readlens`, `readlens`, `rng`. All are now touched in the stub too.

modules/nf-core/telogator2/main.nf
pinin4fjords: Done - all 5 QC files are now individual output channels and touched in the stub.

fellen31: Awesome!

modules/nf-core/telogator2/main.nf
fellen31: Cool!

PR #11034, opened 2026-03-24
Big slice imp modules

SkyLexS:
## PR checklist

Closes #XXX 

- [x] This update includes new arguments for the bigslice tool.
- [x] Add a resource `label`
- [ x Use BioConda and BioContainers if possible to fulfil software requirements.
  - For modules:
  - For subworkflows:


modules/nf-core/bigslice/main.nf
```diff
@@ -9,21 +9,28 @@ process BIGSLICE {
...
     tuple val(meta), path(bgc, stageAs: 'bgc_files/*')
-    path hmmdb
+    path(hmmdb)
 
     output:
-    tuple val(meta), path("${prefix}/result/data.db"), emit: db
+    tuple val(meta), path("${prefix}/result/data.db")   , emit: db
     tuple val(meta), path("${prefix}/result/tmp/**/*.fa"), emit: fa
-    // WARN: Version information not provided by tool on CLI. Please update this string when bumping container versions.
-    tuple val("${task.process}"), val('bigslice'), val("2.0.2"), topic: versions, emit: versions_bigslice
```
camlloyd, line 18: Please keep this

SkyLexS: working on it rn sorry

modules/nf-core/bigslice/main.nf
```diff
@@ -9,21 +9,28 @@ process BIGSLICE {
...
+    def args   = task.ext.args ?: ''
+    def args2  = task.ext.args2 ?: ''
+    prefix     = task.ext.prefix ?: "${meta.id}"
     def sample = meta.id
+    def export_tsv_cmd = args2 ? """
+    bigslice \\
+        --export-tsv ${prefix}/result/tsv_export \\
+        --program_db_folder ${hmmdb} \\
+        ${prefix}
+    """ : ''
```
jfy133: I do not understand this implementation, what is `args2`  trying to do here to activate the execution?

Shouldn't you just have a boolean input value channel that, if `true`, just injects `--export-tsv` in to the main command? Why the whole extra command?

vagkaratzas: Agreed, ext.args can have more than one param - value pair inside. You don't need a new args* value for each param possible. `--export-tsv` can be passed and checked through ext.args

SkyLexS: --export-tsv is a separate BiG-SLiCE subcommand that must be called as a second bigslice invocation after clustering completes, so it cannot simply be appended to the main command. Two approaches exist: using ext.args2 to explicitly separate the two commands (clear intent, but adds a non-standard extra args variable), or detecting --export-tsv inside ext.args, stripping it from the main command and running it as a post-step (current approach), which keeps a single ext.args entry point. If you have any other suggestions please tell :')

jfy133: OK then yeah I would go for a boolean input channel, and inject the second command (with $args2) if requested. 
For variable-based command injection, keep the command on one line rather than across multiple though

vagkaratzas: Definitely add a new nf-test case testing this new funcitonality.

I am confused as to which method to choose, because I can see the `--export-tsv` being used twice; both here `--export-tsv ${prefix}/result/tsv_export \\` and in the `${export_tsv_cmd}` command afterwards.

I'd definitely skip args2, and parse ext.args for finding it though.

jfy133: Given it fundamentally changes the command, I feel input channel is clearer/more explicit

SkyLexS:
no more results folder it will be sample/(bigslice output) fully working parameters with catch for errors if you have any suggestions please tell

modules/nf-core/bigslice/tests/main.nf.test
```diff
@@ -67,6 +67,38 @@ nextflow_process {
...
+
+    test("streptomyces_coelicolor - bigslice - gbk - export_tsv") {
+
+        when {
+            process {
+                """
+                // Flatten the GBK directory into a list of individual GBK files with meta
+                input[0] = UNTAR_GBK.out.untar.map { meta, dir ->
+                    def gbk_files = []
+                    dir.eachFileRecurse { if (it.name.endsWith('.gbk')) gbk_files << it }
```
jfy133, line 110: This might be suffiicent: https://nf-co.re/docs/contributing/nf-test/assertions#snapshotting-variable-files-in-a-channel-emitting-a-directory

modules/nf-core/bigslice/main.nf
```diff
@@ -47,23 +44,25 @@ process BIGSLICE {
         --program_db_folder ${hmmdb} \\
         ${prefix}
 
-    mkdir -p ${prefix}/result/tsv_export
     ${export_tsv_cmd}
+
+    mv ${prefix}/result/data.db ${prefix}/data.db
+    mv ${prefix}/result/tmp    ${prefix}/tmp
+    rm -rf ${prefix}/result
```
jfy133: I don't think it's necessary to do these move operations (and actually it's not recommened apparently), youc an just emit `prefix` as is

modules/nf-core/bigslice/tests/main.nf.test
```diff
@@ -67,16 +67,51 @@ nextflow_process {
...
         then {
             assert process.success
+            def resultDir = file(process.out.output[0][1])
             assertAll(
+                { assert resultDir.isDirectory() },
+                { assert file("${resultDir}/data.db").exists() },
+                { assert resultDir.list().any { it.endsWith('.fa') || new File("${resultDir}/tmp").exists() } },
+                { assert snapshot(
+                    process.out.findAll { key, val -> key.startsWith("versions")}
+                ).match() }
```
famosab, line 96: Can you add other files to this snapshot as well? Ideally we want all outputs to be at least present by name in the snapshot.

modules/nf-core/bigslice/main.nf
```diff
@@ -8,25 +8,27 @@ process BIGSLICE {
         : 'biocontainers/bigslice:2.0.2--pyh8ed023e_0'}"
 
     input:
-    tuple val(meta), path(bgc, stageAs: 'bgc_files/*')
-    path hmmdb
+    tuple val(meta), path(bgc, stageAs: 'bgc_files/s*/*')
+    path(hmmdb)
+    val(export_tsv)
```
famosab, line 13: Can we put all these inputs into one tuple? That will make sure you're sure that EVERY time everything comes together in the right combination.

modules/nf-core/bigslice/main.nf
```diff
@@ -39,16 +41,22 @@ process BIGSLICE {
         -i input \\
         --program_db_folder ${hmmdb} \\
         ${prefix}
+
+    ${export_tsv_cmd}
```
famosab, line 46: Should we add this tsv then as optional output and have it be accessible for downstream analyses?

modules/nf-core/bigslice/main.nf
SkyLexS: hmmdb is a shared reference database (not sample-specific) and export_tsv is a boolean flag (not a file), so neither belongs in the sample tuple. In nf-core, tuples group a meta map with the data files of that specific sample mixing in shared resources or behaviour flags would break this convention.

SkyLexS:
@famosab 
The *.fa files in result/tmp/ are non-deterministic, bigslice generates them based on HMM scoring, and the exact set of hits varies between runs/environments due to floating-point differences. Snapshotting their names will always break on CI. So instead we just assert that at least one .fa exists there, which is enough to confirm the tool ran correctly.

modules/nf-core/bigslice/main.nf
famosab: I see where you are coming from but I would argue that the meta map is assigned to the processed samples and everything that is needed for processing can be adhered to that. But even if you want to leave out the hmmdb (in the end it does not make a difference whether we do a mapping step in the pipeline or whether we have it separate here) the boolean flag should be in the input tuple because only then I can decide on a per sample basis whether I want the tsv export or not which makes the module more flexible :)

jfy133: > I see where you are coming from but I would argue that the meta map is assigned to the processed samples and everything that is needed for processing can be adhered to that. But even if you want to leave out the hmmdb (in the end it does not make a difference whether we do a mapping step in the pipeline or whether we have it separate here) the boolean flag should be in the input tuple because only then I can decide on a per sample basis whether I want the tsv export or not which makes the module more flexible :)

I don't think it makes sense here - like @SkyLexS the first input channel are essentially samples, the second is a database, but the final flag that controls the optional output you would very VERY unlikely to do on per-sample basis (and if you really want to do it. You can use a multi map).

I've not actually seen anywhere the pattern of including module control options in the input tuples, is this new guidance form somewhere I have missed?

jfy133: > > I see where you are coming from but I would argue that the meta map is assigned to the processed samples and everything that is needed for processing can be adhered to that. But even if you want to leave out the hmmdb (in the end it does not make a difference whether we do a mapping step in the pipeline or whether we have it separate here) the boolean flag should be in the input tuple because only then I can decide on a per sample basis whether I want the tsv export or not which makes the module more flexible :)
> 
> I don't think it makes sense here - like @SkyLexS the first input channel are essentially samples, the second is a database, but the final flag that controls the optional output you would very VERY unlikely to do on per-sample basis (and if you really want to do it. You can use a multi map).
> 
> I've not actually seen anywhere the pattern of including module control options in the input tuples, is this new guidance form somewhere I have missed?

In fact, are you arguing that all parameters for a sample should be added into every single sample channel now ?!

famosab: I think we can leave it as is - I am convinced I went down the wrong rabbit hole somewhere :D

SkyLexS:
Is there something else I can implement or do to this module ? I am planing to do something like antismash to automatically download the hmmdb without the user to manually download it.

modules/nf-core/bigslice/main.nf
famosab: I had issues before getting to the files afterwards when I did not do the mv (where are the recommendations on this?)

famosab:
From my side we are good but I guess this should be @vagkaratzas deciding since he wrote the module at some point :)

PR #11036, opened 2026-03-24
Update to topics `telseq`

pinin4fjords:
## Summary
- Migrate telseq module from legacy `versions.yml` file emission to topic-based version channels
- Both `telseq` and `samtools` versions now emitted via `topic: versions`
- Pin `bamtools=2.5.2` in environment.yml to fix conda runtime failure (`libbamtools.so.2.5.2` not found when solver picks bamtools 2.5.3)
- Rebuild containers via Wave from the updated environment
- Hardcode telseq version as `0.0.2` (the binary `--help` misreports `0.0.1`)

## Test plan
- [x] All 4 existing nf-test cases pass with `--update-snapshot` (docker)
- [x] Version commands verified in container: `telseq 0.0.2`, `samtools 1.20`
- [x] Hardcoded version pattern matches established modules (bamcmp, cafe, adapterremovalfixprefix, etc.)

### DIFF
--- a/modules/nf-core/telseq/meta.yml
+++ b/modules/nf-core/telseq/meta.yml
@@ -92,7 +92,7 @@ output:
       - telseq:
           type: string
           description: The name of the tool
-      - 'echo 0.0.2':
+      - "echo 0.0.2":
           type: eval
           description: The expression to obtain the version of telseq
   versions_samtools:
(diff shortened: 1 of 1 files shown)


pinin4fjords:
Conda failures were caused by bamtools 2.5.3 being resolved instead of 2.5.2, breaking the telseq shared library link. Fixed by pinning `bamtools=2.5.2` in environment.yml and rebuilding containers via Wave.

PR #11039, opened 2026-03-25
Update pairtools to 1.1.3 and migrating to topic version

birkiy:
- Updated the pairtools version to 1.1.3. 
- Migrated to topic versions.


birkiy:
@SPPearce Seems that `pairtools 1.1.3` works with `numpy 1.26.4`

Tested as below: 
`nf-core modules test pairtools --profile conda`

Below are the test results. 

Screenshot 2026-03-26 at 6 05 43 PM
Screenshot 2026-03-26 at 6 05 50 PM

Is this okay?

SPPearce:
The recipe: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/pairtools/meta.yaml doesn't pin a specific version of numpy, so conda will likely use the most recent whenever it is run.
Whereas the docker/singularity containers will remain pinned to whatever version that they have inside them currently.
Over time, this can lead to a drift in the numerical output, which is likely why they were pinned previously

birkiy:
@SPPearce The new container has `pairtools 1.1.3` and `numpy 1.26.4`. 

Screenshot 2026-03-27 at 11 17 04 AM

The bioconda recipe pairtools 1.1.3, [link](https://bioconda.github.io/recipes/pairtools/README.html), pinned numpy lower than v2, and the last version before v2 is 1.26.4. 

The mentioned issue has been closed; https://github.com/open2c/pairtools/issues/170

Would you like me to pin numpy 1.26.4?

SPPearce:
No, don't pin the older version then, as the issue has now been resolved.

SPPearce:
You can merge now if you are happy

birkiy:
Thank you, merging in a while.

PR #11045, opened 2026-03-25
Update bioawk

DLBPointon:
Bioawk update
- Remove need for GZ output
  - Added a bool if block for gzipping an output
- Allow for both a GZ and unzipped optional output
- Added suffix and zip_bool input vals
- Added test which produces tsv file
  - [X] NEEDS ADDING TO TEST_DATA REPO
- Updated meta.yml
- Updated tests + snapshot
- Add stub

### DIFF
--- a/modules/nf-core/bioawk/main.nf
+++ b/modules/nf-core/bioawk/main.nf
@@ -34,7 +34,8 @@ process BIOAWK {
 
     stub:
     def prefix      = task.ext.prefix ?: "${meta.id}"
+    def compress_output = output_file_extension.endsWith(".gz") ? " | gzip " : ""
     """
-    echo "" > "${prefix}.${output_file_extension}"
+    echo "" ${compress_output} > "${prefix}.${output_file_extension}"
     """
 }


modules/nf-core/bioawk/main.nf
```diff
@@ -9,26 +9,39 @@ process BIOAWK {
 
     input:
     tuple val(meta), path(input)
+    val suffix
+    val zip_bool
 
     output:
-    tuple val(meta), path("*.gz"), emit: output
+    tuple val(meta), path("${file_output}"),    optional: true, emit: output
+    tuple val(meta), path("*.gz"),              optional: true, emit: gz_output
```
prototaxites: Would it not be better to write all output to a single channel?

modules/nf-core/bioawk/main.nf
```diff
@@ -9,26 +9,39 @@ process BIOAWK {
...
     bioawk \
         $args \
         $input \
-        > ${prefix}
+        > ${file_output}
 
-    gzip ${prefix}
+    if [ "${zip_bool}" = "true" ]; then
+        gzip "${file_output}"
+    fi
```
prototaxites: How about something like:

```
def compress_output = suffix.endsWith(".gz") ? " | gzip " : ""
"""
bioawk \
        $args \
        $input \
        ${compress_output} > ${prefix}.${suffix}
```

That way compression is handled automatically based on the input suffix and you don't need a third input.

modules/nf-core/bioawk/main.nf
```diff
@@ -9,26 +9,39 @@ process BIOAWK {
 
     input:
     tuple val(meta), path(input)
+    val suffix
```
prototaxites:
```suggestion
    val output_file_extension
```

IMO this is more descriptive

modules/nf-core/bioawk/meta.yml
```diff
@@ -31,9 +29,28 @@ input:
         description: |
           Input biological sequence file (optionally gzipped) to be manipulated via the program specified in `$args`.
         pattern: "*.{bed,gff,sam,vcf,fastq,fasta,tab,bed.gz,gff.gz,sam.gz,vcf.gz,fastq.gz,fasta.gz,tab.gz}"
-
+        ontologies:
+          - edam: http://edamontology.org/format_1930
+          - edam: http://edamontology.org/format_3475
```
prototaxites: Could you add the formats in comments?

modules/nf-core/bioawk/tests/nextflow_telomere_check.config
```diff
@@ -0,0 +1,6 @@
+process {
+    withName: BIOAWK {
+        ext.args = "-c fastx \'{s = toupper(\$seq); copy_s = s; g = gsub(/G/, \"\", s); pct = 100*g/length(copy_s); rev = (pct < 30); out = rev ? revcomp(\$seq) : \$seq; printf \"%s\t%d\t%.2f\t%s\t%s\\n\", out, g, pct, (rev ? \"true\" : \"false\"), copy_s}\'"
```
prototaxites: This is quite complicated for a test? 

And can you set this as a param and then define param value inside the test (and for the other test). Then we only need 1 config file.

modules/nf-core/bioawk/main.nf
DLBPointon: agreed, much cleaner. Updated now.

modules/nf-core/bioawk/main.nf
DLBPointon: done

modules/nf-core/bioawk/main.nf
DLBPointon: that's fair, done.

modules/nf-core/bioawk/tests/nextflow_telomere_check.config
DLBPointon: it is, it's just because it was an example I had on hand to be honest. I'll make a simpler one and sort out params.

modules/nf-core/bioawk/meta.yml
DLBPointon: if the ontology website was working

Edit: i just searched for them all in the other meta's

modules/nf-core/bioawk/tests/nextflow_telomere_check.config
DLBPointon: done

modules/nf-core/bioawk/tests/main.nf.test
```diff
@@ -10,16 +10,49 @@ nextflow_process {
     tag "modules_nfcore"
     tag "bioawk"
 
-    test("test-bioawk") {
+    test("fasta bioawk fasta.gz") {
+        config "./nextflow.config"
```
prototaxites:
```suggestion
```

Not needed as it's declared at top of the script too

modules/nf-core/bioawk/meta.yml
```diff
@@ -31,44 +29,51 @@ input:
...
         pattern: "*.{bed,gff,sam,vcf,fastq,fasta,tab,bed.gz,gff.gz,sam.gz,vcf.gz,fastq.gz,fasta.gz,tab.gz}"
-
+        ontologies:
+          - edam: "http://edamontology.org/format_3003"
+          - edam: "http://edamontology.org/format_3016"
+          - edam: "http://edamontology.org/format_3989"
+          - edam: "http://edamontology.org/format_1975"
+          - edam: "http://edamontology.org/format_1929"
+          - edam: "http://edamontology.org/format_1930"
+          - edam: http://edamontology.org/format_3475
```
prototaxites:
```suggestion
          - edam: http://edamontology.org/format_3003 # BED
          - edam: http://edamontology.org/format_3016 # VCF
          - edam: http://edamontology.org/format_3989 # GZIP
          - edam: http://edamontology.org/format_1975 # GFF3
          - edam: http://edamontology.org/format_1929 # FASTA
          - edam: http://edamontology.org/format_1930 # FASTQ
          - edam: http://edamontology.org/format_3475 # TSV
```

modules/nf-core/bioawk/main.nf
```diff
@@ -9,26 +9,32 @@ process BIOAWK {
 
     input:
     tuple val(meta), path(input)
+    val output_file_extension
 
     output:
-    tuple val(meta), path("*.gz"), emit: output
+    tuple val(meta), path("*.${output_file_extension}"), emit: output
     tuple val("${task.process}"), val('bioawk'), val("1.0"), emit: versions_bioawk, topic: versions
     // WARN: Version information not provided by tool on CLI. Please update version string above when bumping container versions.
```
prototaxites:
```suggestion
// WARN: Version information not provided by tool on CLI. Please update version string above when bumping container versions.
tuple val("${task.process}"), val('bioawk'), val("1.0"), emit: versions_bioawk, topic: versions
```

modules/nf-core/bioawk/tests/nextflow.config
```diff
@@ -1,6 +1,6 @@
 process {
     withName: BIOAWK {
-        ext.args = "-c fastx \'{print \">\" \$name ORS length(\$seq)}\'"
-        ext.prefix = "sample_1.fa"
+        ext.args = "${params.bioawk_command}"
+        ext.prefix = "${params.bioawk_prefix}"
```
prototaxites:
```suggestion
        ext.args = { "${params.bioawk_command}" }
        ext.prefix = { "${params.bioawk_prefix}" }
```

Not sure if this is necessary but doesn't hurt

modules/nf-core/bioawk/meta.yml
DLBPointon: AAAaaaah I added those in and then linted... thank you for adding back in.

modules/nf-core/bioawk/main.nf
```diff
@@ -9,26 +9,32 @@ process BIOAWK {
...
-    def args  = task.ext.args ?: ''
-    def prefix = task.ext.prefix ?: "${meta.id}"
-    if ("${input}" == "${prefix}") error "Input and output names are the same, use \"task.ext.prefix\" to disambiguate."
+    def args        = task.ext.args ?: ''
+    def prefix      = task.ext.prefix ?: "${meta.id}"
+    if ("${input}" == "${prefix}.${output_file_extension}") error "Input and output names are the same, use \"task.ext.prefix\" to disambiguate."
 
+    def compress_output = output_file_extension.endsWith(".gz") ? " | gzip " : ""
     """
     bioawk \
```
prototaxites:
```suggestion
    bioawk \\
```

modules/nf-core/bioawk/main.nf
```diff
@@ -9,26 +9,32 @@ process BIOAWK {
...
-        $args \
-        $input \
-        > ${prefix}
+            $args \
+            $input \
+            ${compress_output} > ${prefix}.${output_file_extension}
```
prototaxites, line 32:
```suggestion
            $args \\
            $input \\
```

modules/nf-core/bioawk/tests/main.nf.test
```diff
@@ -10,16 +10,48 @@ nextflow_process {
...
+                { assert snapshot(process.out).match() }
+            )
+        }
+    }
+
+    test("fasta bioawk tsv") {
+
+        when {
+            params {
+                bioawk_command = "-c gfastx \'{s = toupper(\$seq); g = gsub(/G/, \"\", s); printf \"%s\\t%d\\n\", s, g}\'"
```
prototaxites:
```suggestion
                bioawk_command = '''-c gfastx '{s = toupper(\$seq); g = gsub(/G/, "", s); printf "%s\t%d\n", s, g}''''
```

I wonder if the problem is an escaping issue? Claude gives me this to try instead

DLBPointon: I've got it, it was a spelling mistake it's `fastx` not `gfastx`. I've been looking at `gfastats` too much.

modules/nf-core/bioawk/main.nf
```diff
@@ -9,26 +9,32 @@ process BIOAWK {
...
-    bioawk \
-        $args \
-        $input \
-        > ${prefix}
 
-    gzip ${prefix}
+    stub:
+    def prefix      = task.ext.prefix ?: "${meta.id}"
+    """
+    echo "" > "${prefix}.${output_file_extension}"
```
prototaxites:
```suggestion
    def prefix      = task.ext.prefix ?: "${meta.id}"
    def compress_output = output_file_extension.endsWith(".gz") ? " | gzip " : ""
    """
    echo "" ${compress_output} > "${prefix}.${output_file_extension}"
```

nf-test will complain if a file with a .gz extension isn't actually gzipped

Issue #8716, opened 2025-06-25
[FEATURE] Update modules to remove undocumented task.ext fields


mahesh-panchal:
### Is your feature request related to a problem? Please describe

task.ext fields were introduced allow users to override developer settings in pipelines. However some modules are simply using them as an alternative `input:` method for parameters that don't always need to be supplied.

```
% find modules -name "*.nf" -exec grep -H "task.ext." {} \; | grep -vE "task\.ext\.(singularity|args|prefix|when|suffix)"
modules/nf-core/picard/fixmateinformation/main.nf:    def STRINGENCY = task.ext.stringency ?: "STRICT"
modules/nf-core/artic/minion/main.nf:    def hd5_plugin_path = task.ext.hd5_plugin_path ? "export HDF5_PLUGIN_PATH=${task.ext.hd5_plugin_path}" : "export HDF5_PLUGIN_PATH=/usr/local/lib/python3.6/site-packages/ont_fast5_api/vbz_plugin"
modules/nf-core/fgbio/zipperbams/main.nf:    def compression = task.ext.compression ?: '0'
modules/nf-core/jupyternotebook/main.nf:    def parametrize = (task.ext.parametrize == null) ?  true : task.ext.parametrize
modules/nf-core/jupyternotebook/main.nf:    def implicit_params = (task.ext.implicit_params == null) ? true : task.ext.implicit_params
modules/nf-core/jupyternotebook/main.nf:    def meta_params = (task.ext.meta_params == null) ? true : task.ext.meta_params
modules/nf-core/jupyternotebook/main.nf:    def kernel   = task.ext.kernel ?: '-'
modules/nf-core/rmarkdownnotebook/main.nf:    def parametrize = (task.ext.parametrize == null) ?  true : task.ext.parametrize
modules/nf-core/rmarkdownnotebook/main.nf:    def implicit_params = (task.ext.implicit_params == null) ? true : task.ext.implicit_params
modules/nf-core/rmarkdownnotebook/main.nf:    def meta_params = (task.ext.meta_params == null) ? true : task.ext.meta_params
modules/nf-core/plink2/vcf2bgen/main.nf:    def reffirst = bgen_reffirst || task.ext.reffirst ? "ref-first" : ""
modules/nf-core/plink2/vcf2bgen/main.nf:    def dosage = task.ext.dosage_field ? task.ext.dosage_field : dosage_field
[...]

PR #11049, opened 2026-03-25
Masurca

### LINKED ISSUE #10945: new module: MASURCA

### Further Information

github: https://github.com/alekseyzimin/masurca
bioconda: https://anaconda.org/channels/bioconda/packages/masurca/overview

LiaOb21:
Masurca module [draft]

Supports assembly with Cabog assembly only. Flye and SOAPdenovo are not supported.

Supported reads:
- single end or merged reads
- paired end only
- paired end + mate pair (jump)
- paired end + nanopore
- paired end + pacbio
- paired end + nanopore + pacbio

Does not support other kinds of reads (Sanger, 454, etc) and synteny guided assembly with reference genome

## PR checklist

Closes #10945 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


chriswyatt1:
Are the value here pipeline wide, or specific to each genome being run? Is that why they need to be specific channel inpus?
    val fragment_mean
    val fragment_stdev
    val jump_mean
    val jump_stdev

chriswyatt1:
path(other_reads) doesn't appear to be used, so could be removed

chriswyatt1:
Looks like a specific build type, so I think seqera containers won't be able to handle that. 
https://github.com/alekseyzimin/masurca?tab=readme-ov-file#compileinstall-requirements
We will need to build this manually I think. I can have a go

Then we will need to do this:
https://nf-co.re/docs/contributing/guidelines/recommendations/custom_containers

LiaOb21:
> Are the value here pipeline wide, or specific to each genome being run? Is that why they need to be specific channel inpus? val fragment_mean val fragment_stdev val jump_mean val jump_stdev
 
 I think they are meant to be as realistic as possible for specific genomes. However, I ran all the tests locally and manually in a conda env using the same nf-core datasets I'm using in main.test.nf and using the same values, and they all gave me the expected outputs.
 
 > path(other_reads) doesn't appear to be used, so could be removed
 
 Yes, I was initially planning to integrate that as well, but then I decided to leave it. I forgot to remove it from the input tuple.



> Looks like a specific build type, so I think seqera containers won't be able to handle that. https://github.com/alekseyzimin/masurca?tab=readme-ov-file#compileinstall-requirements We will need to build this manually I think. I can have a go

Thank you @chriswyatt1! 😊

chriswyatt1:
OK, Just thinking to make this module more atomic, we should have these values as ext args. But if its helpful, to have this in a channel, I think it can be justified. If we think that any user would need to parse this information in for multiple genomes always, rather than set them generally in the modules.config say

chriswyatt1:
The only other thing of general comments, is about all the echo statements. Is this essential, or maybe we could move some of it into a script maybe. I will check out what other modules do in this case

LiaOb21:
> OK, Just thinking to make this module more atomic, we should have these values as ext args. But if its helpful, to have this in a channel, I think it can be justified. If we think that any user would need to parse this information in for multiple genomes always, rather than set them generally in the modules.config say

I think the ideal use would be to have a specific value per genome. Personally I'm using an average fixed value, as I often don't have information about the fragment size. Also from masurca documentation is not completely clear if they refer to the fragment size or the insert size. I don't know what would be more appropriate for the nf-core module, so I'll follow your suggestion 😊 

 

> The only other thing of general comments, is about all the echo statements. Is this essential, or maybe we could move some of it into a script maybe. I will check out what other modules do in this case

Yes, so that's the main reason why I was so confused on how setting up this module. I asked in the community and they suggested to have a look to these two modules:
- https://github.com/nf-core/modules/tree/master/modules/nf-core/controlfreec/freec
- https://github.com/nf-core/modules/tree/master/modules/nf-core/quartonotebook

controlfreec also uses echo commands and I thought that was probably the easiest way to build the config for masurca. But happy to change if there is any better way to deal with this! 😊

chriswyatt1:
I made a container, not sure it will work, but worth a try:

quay.io/ecoflowucl/masurca    so in the module you can just put ecoflowucl/masurca

test with `docker pull quay.io/ecoflowucl/masurca:v4.1.4`

Dockerfile is here: https://github.com/Eco-Flow/docker-build/tree/main/masurca

chriswyatt1:
I think the container works. But yes, there are some output files that have time stamps and specific work dir paths, which will change on each run. So we need to make the nf-test more specific. Pick files that won't change
 assemble.sh
 #!/bin/bash

# assemble.sh generated by masurca:
CONFIG_PATH="/Users/cwyatt/Downloads/modules-2/.nf-test/tests/aad480bb9b0561d1bd4bc958001b0e93/work/e4/920dca1167539f25218df125acaf84/test_masurca_config.txt"

test-masurca.log:
[Thu Mar 26 18:02:31 UTC 2026] Processing pe library reads
[Thu Mar 26 18:02:32 UTC 2026] Average PE read length 100
[Thu Mar 26 18:02:32 UTC 2026] Using kmer size of 67 for the graph
[Thu Mar 26 18:02:32 UTC 2026] MIN_Q_CHAR: 33
[Thu Mar 26 18:02:32 UTC 2026] Creating mer database for Quorum

chriswyatt1:
@LiaOb21 did you manage to test the container? I think it worked,,, but let me know if not, and I can look into it

LiaOb21:
> @LiaOb21 did you manage to test the container? I think it worked,,, but let me know if not, and I can look into it

Hey @chriswyatt1 the container works! :blush: thank you so much! 

> Looks like a specific build type, so I think seqera containers won't be able to handle that. https://github.com/alekseyzimin/masurca?tab=readme-ov-file#compileinstall-requirements We will need to build this manually I think. I can have a go
> 
> Then we will need to do this: https://nf-co.re/docs/contributing/guidelines/recommendations/custom_containers

We can proceed with this, in the meantime I'll try to set up the tests in a way that they can ignore the timestamps and so on

.vscode/extensions.json
maxulysse, line 1: why deleting this file?

.vscode/settings.json
maxulysse, line 1: why deleting this file?

.vscode/extensions.json
LiaOb21: Did I?

modules/nf-core/masurca/main.nf
```diff
@@ -0,0 +1,137 @@
+process MASURCA {
+    tag "$meta.id"
+    label 'process_high'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'oras://community.wave.seqera.io/library/coreutils_file_masurca_mummer_perl:73ce913377915362':
```
mashehu: please use the https url of the container nf-co.re/docs/tutorials/nf-core_components/using_seqera_containers

LiaOb21: I'm not going to use that container, it doesn't work. This is still work in progress. @chriswyatt1 made a container for that reason

.vscode/extensions.json
LiaOb21: I believe something overwritten this. It wasn't intentional for sure

.vscode/settings.json
LiaOb21: Same as for the other file, these changes can be ignored, I think they were overwritten

.vscode/extensions.json
LiaOb21: I added it back. Thank you for pointing that out

modules/nf-core/masurca/main.nf
```diff
@@ -0,0 +1,133 @@
+process MASURCA {
+    tag "$meta.id"
+    label 'process_high'
+
+    conda "${moduleDir}/environment.yml"
+    container "quay.io/ecoflowucl/masurca@sha256:49509d5c7d5e95e7de38127bb2d9bdd14e0d736e7397928132d05f2611aead05"
```
LiaOb21: Please ignore the container for now

LiaOb21:

LiaOb21:
Hello @mashehu @maxulysse and @chriswyatt1! :blush: 
I think this is almost ready for review.

As regards the container, here is the conda recipe: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/masurca/meta.yaml

The reason why the biocontainer obtained from conda doesn't work is because some dependencies, like for example `file`, are not in the recipe although they are used from masurca. The tests work with conda because these dependencies are normally available in a linux environment, but they fail with docker for obvious reasons. However, the container created by @chriswyatt1 works, and this is what I am using at the moment, and I already asked on slack to host it in the nf-core org.

Also, I was having some linting issues:
```
╭─ [✗] 1 Module Test Failed ───────────────────────────────────────────────────╮
│             ╷                               ╷                                │
│ Module name │ File path                     │ Test message                   │
│╶────────────┼───────────────────────────────┼───────────────────────────────╴│
│ masurca     │ modules/nf-core/masurca/main… │ main_nf_ext_key: Invalid 'ext' │
│             │                               │ keys detected:                 │
│             │                               │ ext.extend_jump_reads,         │
│             │                               │ ext.extend_jump_reads,         │
│             │                               │ ext.graph_kmer_size,           │
│             │                               │ ext.use_linking_mates,         │
│             │                               │ ext.use_linking_mates,         │
│             │                               │ ext.lhe_coverage,              │
│             │                               │ ext.mega_reads_one_pass,       │
│             │                               │ ext.mega_reads_one_pass,       │
│             │                               │ ext.limit_jump_coverage,       │
│             │                               │ ext.ca_parameters,             │
│             │                               │ ext.close_gaps,                │
│             │                               │ ext.close_gaps, ext.jf_size    │
│             ╵                               ╵                                │
╰──────────────────────────────────────────────────────────────────────────────╯
```

Then I saw this issue: https://github.com/nf-core/modules/issues/8716
And I moved the task.ext to inputs. The linting now works. I left the task.ext commented just to double check with you if this is the correct approach and go back quickly if necessary.

And, finally, the tests. These are avoiding the md5 check because masurca works with absolute paths, and these paths change for every run making the md5 inconsistent. Please let me know if I used the right approach or if there is anything else to change.

Thank you so much for your help! :blush:

modules/nf-core/masurca/main.nf
```diff
@@ -0,0 +1,141 @@
...
+    val jump_stdev
+    val extend_jump_reads
+    val graph_kmer_size
+    val use_linking_mates
+    val lhe_coverage
+    val mega_reads_one_pass
+    val limit_jump_coverage
+    val ca_parameters
+    val close_gaps
+    val jf_size
```
famosab, line 24: Are these all strictly required values?

modules/nf-core/masurca/main.nf
famosab, line 1: You can run
```bash
```

modules/nf-core/masurca/main.nf
```diff
@@ -0,0 +1,141 @@
...
+    // def graph_kmer_size = task.ext.graph_kmer_size ?: 'auto'
+    // def use_linking_mates = task.ext.use_linking_mates != null ? task.ext.use_linking_mates : 0
+    // def lhe_coverage = task.ext.lhe_coverage ?: 25
+    // def mega_reads_one_pass = task.ext.mega_reads_one_pass != null ? task.ext.mega_reads_one_pass : 0
+    // def limit_jump_coverage = task.ext.limit_jump_coverage ?: 300
+    // def ca_parameters = task.ext.ca_parameters ?: 'cgwErrorRate=0.15'
+    // def close_gaps = task.ext.close_gaps != null ? task.ext.close_gaps : 0
+    // def jf_size = task.ext.jf_size ?: 200000000
+    """
+    echo "DATA" > ${prefix}_masurca_config.txt
```
famosab: I am wondering whether it would make sense to have a sort of mascura/writeconfig and mascure/mascura (or mascura/run) division so that we can divide these two things a bit more?

modules/nf-core/masurca/main.nf
```diff
@@ -0,0 +1,141 @@
...
+    """
+
+    stub:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    mkdir -p CA
+    touch assemble.sh
+    touch ${prefix}_masurca_config.txt
+    echo "" | gzip > ${prefix}.scaffolds.fa.gz
```
famosab:
```suggestion
    echo | gzip > ${prefix}.scaffolds.fa.gz
```

modules/nf-core/masurca/main.nf
LiaOb21: Hi @famosab, they are depending on the settings, i.e. which kind of reads you want to use

modules/nf-core/masurca/main.nf
LiaOb21: masurca/writeconfig is not a command that masurca has though. I don't know what is the nf-core policy in this case. I wouldn't mind splitting if it is okay

modules/nf-core/masurca/environment.yml
```diff
@@ -0,0 +1,7 @@
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/environment-schema.json
+channels:
+  - conda-forge
+  - bioconda
+dependencies:
+  - "bioconda::masurca=4.1.4"
```
mashehu: can you add the missing dependencies here (or even better in the upstream bioconda recipe)? Then we can use this environment.yml to generate a docker and singularity container using seqera containers.

modules/nf-core/masurca/main.nf
famosab: @nf-core/maintainers I think it would make sense here to keep this a bit more organized but not 100% sure if that makes sense

modules/nf-core/masurca/environment.yml
LiaOb21: Yes I can update the bioconda recipe, but then we willl have the build number problem as for minia module with wave. If someone is willing to help with that as I didn't figure out how to do it, I can go to update bioconda. If using @chriswyatt1 container is not an option

modules/nf-core/masurca/main.nf
jonasscheid: Makes sense to me that the module expects the config file. In your pipeline you could have a default config.txt in some folder (e.g. `assets`) and use a quick groovy function to read and update the config values based on the nextlfow params given

jonasscheid: I think having a separate module might be overkill 🤔

famosab: Ah nice idea with the groovy function and a sort of template config!

There is even a module that can do template based filling, so that also might be useful for to chain with this module then https://nf-co.re/modules/yte/ (when used in a pipeline)

modules/nf-core/masurca/main.nf
famosab: Ok I see, I would go the route that jonas suggested with reasonable defaults / even using a template for the config file when in a pipeline :)

modules/nf-core/masurca/environment.yml
mashehu: it would just make future updates easier. can try to update the bioconda recipe and I will figure the wave part out

modules/nf-core/masurca/main.nf
mahesh-panchal: What I usually prefer is to have the configuration construction in groovy, and then print the config as a string using `cat` to redirect it to a file. There's an example here: https://github.com/nf-core/modules/pull/11066. 
I feel like it's easier to debug this way if there's less of the construction logic embedded in the shell script.

mahesh-panchal: Just to be explicit, the config writing line ends up looking like
```
    cat > cellranger_multi_config.csv <<-CONFIG_EOF
    ${config_content}
    CONFIG_EOF
```
and the `config_content` is made using an array that's joined into a string.

LiaOb21: @famosab and @jonasscheid, I understand how a template in assets and function to fill the config would work in a pipeline but not in a standalone module. Maybe I'm missing something.

What @mahesh-panchal suggests makes more sense, I'll try to rewrite the bash in groovy, although I doubt it will be less long/ confusing :sweat_smile: but okay

modules/nf-core/masurca/environment.yml
LiaOb21: I can update bioconda and let you know as soon as it's done. Thank you!

modules/nf-core/masurca/main.nf
jonasscheid: Sure, no groovy in the module, but you could have a test config in nf-core testdata and just fetch it from there. But @mahesh-panchal solution sounds also fine to me 👍🏼

mahesh-panchal: I don't expect the groovy to be simple. You can see that with the cell ranger module. It's still fairly complex. One of the things that still needs a clean solution is the specification of optional args. The guidelines state this should be supplied with ext.args unless they're files, but we also shouldn't hard code certain things. My tactic at this point is to supply config settings with ext.args as space separated strings, and the groovy code puts it in the correct place.

famosab: I would say this needs chaining with another module and for the tests here we just supply a test config file so no need to write a lot of code but rather require the config as an input to the module

LiaOb21: Thank you all for the suggestions! I was already working on changing the module according to @mahesh-panchal suggestions. Is that okay or do you prefer me to do it in another way? I am confused now

famosab: In the end its up to you :) I think either way is fine. I think what @jonasscheid and me proposed would put more burden on the pipeline developer but seems at least easier to me while what @mahesh-panchal proposed makes the module probably easier to integrate into the pipeline for the pipeline dev but puts more maintenance / burden on you as module developer :)

mahesh-panchal: I think for contributions to nf-core modules, the module should be as fully formed as possible. Pipeline developers have the option to `nf-core modules patch` and can change a module to reduce burden on that side. In the end, I think it just comes down to we're lacking documentation how modules should be used. If a module is separated it needs documentation that there's a module that should also be installed for upstream use. I don't want to add to confusion though.

modules/nf-core/masurca/main.nf
```diff
@@ -0,0 +1,140 @@
...
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+
+    //get input reads with absolute paths - illumina are mandatory, jump/pacbio/nanopore are optional
+    def illumina_reads = [illumina].flatten().collect { it.toRealPath() }.join(' ')
```
mahesh-panchal: If you're using the `cat` way, one trick you can do is include `$PWD` with the relative path so the actual file contains the absolute path

For example I'll add a row like
```
    if (has_ab || has_crispr) {
        fb_section << '[feature]'
        fb_section << 'reference,\$PWD/fb_reference_copy.csv'
    }
```
The path to the csv needs the absolute path. `cat` will interpolate variables in bash scope like `$PWD`. This simplifies some of the groovy code.

modules/nf-core/masurca/main.nf
```diff
@@ -41,81 +41,86 @@ process MASURCA {
...
-    echo "#if single-end, do not specify <reverse_reads>" >> ${prefix}_masurca_config.txt
-    echo "#MUST HAVE Illumina paired end reads to use MaSuRCA" >> ${prefix}_masurca_config.txt
-    echo "PE= pe ${fragment_mean} ${fragment_stdev} ${illumina_reads}" >> ${prefix}_masurca_config.txt
-
-    # Jump/mate pair reads (optional)
-    if [ -n "${jump_reads}" ]; then
-        echo "#Illumina mate pair reads supplied as <two-character prefix> <fragment mean> <fragment stdev> <forward_reads> <reverse_reads>" >> ${prefix}_masurca_config.txt
-        echo "JUMP= sh ${jump_mean} ${jump_stdev} ${jump_reads}" >> ${prefix}_masurca_config.txt
-    fi
+    // Build the config file
```
LiaOb21, line 46: @mahesh-panchal would you mind checking if I understood well and did what you were expecting?

modules/nf-core/masurca/main.nf
```diff
@@ -0,0 +1,145 @@
...
+    config_lines << "END"
+
+    def config_content = config_lines.join('\n')
+
+    """
+    # Write the config file
+cat > ${prefix}_masurca_config.txt <<-CONFIG_EOF
+${config_content}
+CONFIG_EOF
+
```
mahesh-panchal, line 123:
```suggestion
    def config_content = config_lines.join('\n    ')

    """
    # Write the config file
    cat > ${prefix}_masurca_config.txt <<-CONFIG_EOF
    ${config_content}
    CONFIG_EOF
```

modules/nf-core/masurca/main.nf
mahesh-panchal: This looks like you're on the right track. You can also use multi-line strings too btw for large sections.

mahesh-panchal: Oh, I forgot to say, if you're using multi-line strings, the `join` operation needs to reflect that.

```
def config_content = config_lines.collectMany{ cfg -> cfg.tokenize('\n') }.join('\n    ')
```

modules/nf-core/masurca/main.nf
LiaOb21: I'm worried about one thing here. If I use

`def illumina_reads = [illumina].flatten().join(' ')` + 
`config_lines << "PE= pe ${fragment_mean} ${fragment_stdev} \$PWD/${illumina_reads}"`

is not the result of ` \$PWD/${illumina_reads}"` going to be 
`path/to/reads_1.fastq.gz reads_2.fastq.gz` 
rather than `path/to/reads_1.fastq.gz path/to/reads_2.fastq.gz` ?

mahesh-panchal: If `reads` is expected to be a list, it's best to use `collect` to format it.

```groovy
"PE= pe ${fragment_mean} ${fragment_stdev} ${illumina_reads.collect{ fq -> "\$PWD/${fq}"}.join(" ") }"
```

modules/nf-core/masurca/environment.yml
LiaOb21: Hi @mashehu, PR for bioconda opened: https://github.com/bioconda/bioconda-recipes/pull/64293
I tested both the conda env and the docker with all the combinations of input reads set up here in the module and it seems all in place now

LiaOb21:
Hi @mashehu, the bioconda recipe has been update and works. I tried to obtain the docker and singularity with wave cli using both pixi:v1 and default template:

```
---
channels:
  - conda-forge
  - bioconda
dependencies:
  - "bioconda::masurca=4.1.4=ha5bb246_1"

```

```
wave --conda-file environment.yml --freeze --platform linux/amd64 -o yaml --build-template conda/pixi:v1 --await
buildId: bd-2788c9bd40859936_1
cached: true
containerImage: community.wave.seqera.io/library/masurca:4.1.4--2788c9bd40859936
duration: PT2M52.608153482S
freeze: true
mirror: false
requestId: 4a0e1f34d473
scanId: sc-e633b339ce626493_1
status: DONE
succeeded: true
targetImage: community.wave.seqera.io/library/masurca:4.1.4--2788c9bd40859936
```

```
wave --conda-file environment.yml --freeze --platform linux/amd64 -o yaml --build-template conda/pixi:v1 --singularity --await
buildId: bd-2d0ee6b415d9e77e_1
cached: false
containerImage: oras://community.wave.seqera.io/library/masurca:4.1.4--2d0ee6b415d9e77e
duration: PT3M32.779779428S
freeze: true
mirror: false
requestId: 5fbb5d604d9e
status: DONE
succeeded: true
targetImage: oras://community.wave.seqera.io/library/masurca:4.1.4--2d0ee6b415d9e77e
```

```
wave --conda-file environment.yml --freeze --platform linux/amd64 -o yaml --await
buildId: bd-6c6f779135534097_1
cached: false
containerImage: community.wave.seqera.io/library/masurca:4.1.4--6c6f779135534097
duration: PT1M59.355046634S
freeze: true
mirror: false
requestId: 7bc08416d07c
scanId: sc-680889439ee593d6_1
status: DONE
succeeded: true
targetImage: community.wave.seqera.io/library/masurca:4.1.4--6c6f779135534097
```

```
wave --conda-file environment.yml --freeze --platform linux/amd64 -o yaml --singularity --await
buildId: bd-40a5894c07082589_1
cached: false
containerImage: oras://community.wave.seqera.io/library/masurca:4.1.4--40a5894c07082589
duration: PT2M39.588376664S
freeze: true
mirror: false
requestId: f7d8cd412b66
status: DONE
succeeded: true
targetImage: oras://community.wave.seqera.io/library/masurca:4.1.4--40a5894c07082589
```

Again, the new conda recipe works, while both the docker versions (pixi and default) give me the error:
`Overlap/unitig failed, check output under CA/ and runCA1.out, try re-generating and re-running assemble.sh` for each of the tests.

Is there anything else I can try? I believe I added all the required dependencies to the bioconda recipe now.

LiaOb21:
I forgot to add that I tested the docker container from bioconda fetch artifacts with same test data and masurca config files and it works without problems.

modules/nf-core/masurca/main.nf
```diff
@@ -0,0 +1,147 @@
...
+        config_lines << "PACBIO=${pacbio_file}"
+    } else if (nanopore_file) {
+        config_lines << "#Nanopore reads must be in a single fasta or fastq file with absolute path"
+        config_lines << "NANOPORE=${nanopore_file}"
+    }
+
+    config_lines << "END"
+    config_lines << ""
+
+    // PARAMETERS section
```
mahesh-panchal, line 80: It's been a very long time since I've used Masura so I don't remember much about what's required input. Can these be supplied through `ext.args`?

The idea would be that you would have
```
config_lines.addAll(args.collect{ key, value -> "${key}=${value}"})
```
So `ext.args` is a map in this case (for example like modules VIBER or MOBSTER). 

The user would then supply lines through the `ext.args`
```
withName: 'MASURCA' {
    ext.args = [
        LHE_COVERAGE: 35,
        SOAP_ASSEMBLY: 0,
    ]
```

LiaOb21: Hi @mahesh-panchal, I moved several arguments to inputs in this commit: https://github.com/LiaOb21/modules/commit/c3c4faf5b05f48f808618bac1a66a4de4f637263 
I made this change because the linter was throwing an error, if I remember well it wanted those arguments documented in `meta.yml`, though I'm not entirely sure. I believe the map won't trigger the same issue, so we could try that approach.
However, regarding the Masurca inputs, the only guidance available on configuring it comes from comments within the config file itself. Unfortunately, those comments don't clearly specify which parameters are optional. In the past, I attempted to omit some non-required parameters based on the input files I was working with, but this caused the generation of `assemble.sh` to fail.

LiaOb21:
Hello everyone, is there anything I can do to help moving this module forward?

SPPearce:
> Hello everyone, is there anything I can do to help moving this module forward?

What do you need, a review?

SPPearce:
(and in general, keep prodding if you need help or a module reviewing)

LiaOb21:
Hi @SPPearce, there were issues with the container that @mashehu was trying to resolve. We updated the bioconda recipe, but for some reason the container that wave was producing wasn't working as expected

SPPearce:
Hmm, it is failing on docker as well

vagkaratzas:
> Hmm, it is failing on docker as well

Hmmm...I have similar fails here: https://github.com/nf-core/modules/actions/runs/25495866915/job/74815581856?pr=11458

Any chance it's not the module test faults and something else?....It's been bugging me for a while

PR #11053, opened 2026-03-26
New modules: Llamacpp-python/run and huggingface/download for allowing to run simple text workloads with local LLMs 

toniher:
This pull request, contributed jointly with @lucacozzuto , provides a simple workload for running text inference tasks using llamacpp-python against local LLMs.
This effort was worked on during the nf-core Hackathon in March 2026.

## PR checklist

Closes #XXX 

- [X] Add a resource `label`
  - For modules:
    - [X] `nf-core modules test <MODULE> --profile singularity`
    - [X] `nf-core modules test <MODULE> --profile conda`
  - For subworkflows:


modules/nf-core/huggingface/download/tests/main.nf.test
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert process.out.output.size() == 1 },
+                { assert process.out.output[0][0] == [ id:'test_model' ] },
+                { assert file(process.out.output[0][1]).name == "gemma-3-1b-it-Q4_K_M.gguf" },
+                { assert file(process.out.output[0][1]).size() > 0 },
+                { assert snapshot(process.out.findAll { key, val -> key.startsWith('versions') }).match() }
```
famosab: Can you add other files to this snapshot as well? Ideally we want all outputs to be at least present by name in the snapshot.

modules/nf-core/huggingface/download/tests/main.nf.test
```diff
@@ -0,0 +1,66 @@
...
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert process.out.output.size() == 1 },
+                { assert process.out.output[0][0] == [ id:'test_model' ] },
+                { assert file(process.out.output[0][1]).name == "gemma-3-1b-it-Q4_K_M.gguf" },
+                { assert snapshot(process.out.findAll { key, val -> key.startsWith('versions') }).match() }
```
famosab: You can use
```
```
for the stub assertion. It should not be necessary to do any other assertions in the stub test.


```suggestion
                { assert snapshot(sanitizeOutput(process.out)).match() }
```

modules/nf-core/huggingface/download/tests/nextflow.config
```diff
@@ -0,0 +1,7 @@
+nextflow.enable.moduleBinaries = true
```
famosab: I do not think this should be here

modules/nf-core/huggingface/download/tests/nextflow.config
```diff
@@ -0,0 +1,7 @@
...
+
+process {
```
famosab: We want the ext.args to be present in the main.nf.test file. That makes everything more readable in one go. See the [docs](https://nf-co.re/docs/guidelines/components/modules#configuration-of-extargs-in-tests) for more info.

modules/nf-core/huggingface/download/main.nf
```diff
@@ -0,0 +1,30 @@
+process HF_DOWNLOAD {
```
famosab: According to the [module specifications](https://nf-co.re/docs/guidelines/components/modules#naming-conventions) we want the module name to be reflected in the process name. In this case it should be `HUGGINGFACE_DOWNLOAD` or you need to change your folder name to hf but I would suggest to keep the more descriptive name.

modules/nf-core/huggingface/download/main.nf
famosab, line 1: You can run
```bash
```

modules/nf-core/huggingface/download/main.nf
```diff
@@ -0,0 +1,30 @@
...
+
+    output:
+    tuple val(meta), path(hf_file), emit: output
+    tuple val("${task.process}"), val("huggingface_hub"), eval("hf --version 2>&1 | tail -n1 | awk '{print \$NF}'"), topic: versions, emit: versions_huggingface_hub
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def hf_home_resolved = hf_home ?: "${workflow.projectDir}/hf_cache"
```
famosab: I dont think the module should write to a hard coded path in the projectDir

modules/nf-core/huggingface/download/meta.yml
```diff
@@ -0,0 +1,66 @@
+name: hf_download
+description: Tool for downloading models from HuggingFace
+keywords:
+  - llm
+  - llama
+  - ai
+tools:
+  - huggingface_hub:
+      description: "HuggingFace Hub CLI interface"
```
famosab: Can this be described a bit better?

famosab (commented):
Thank you for your contribution to nf-core! We really appreciate it. I added a few comments to your PR.

We usually recommend to have one module per PR. That makes the review process easier and its more likely that someone will review your PR. You can keep that in mind for the next PRs.

modules/nf-core/llamacpp-python/run/tests/data/stub_model.gguf
famosab, line 1: both files in this folder should be added to test-datasets and not to the modules repo

toniher:
> Thank you for your contribution to nf-core! We really appreciate it. I added a few comments to your PR.
> 
> We usually recommend to have one module per PR. That makes the review process easier and its more likely that someone will review your PR. You can keep that in mind for the next PRs.

Hi @famosab . Thanks for the feedback and I will go through your comments! I was told about the module submission, but since you need the output of one of the processes for the other, I thought it would help potential users once it could become eventually accepted. But, certainly, is more work for everyone. Sorry about this and I will avoid it in future PRs.

modules/nf-core/huggingface/download/meta.yml
toniher: I tried to provide more details here: https://github.com/nf-core/modules/pull/11053/commits/6dfff97e7fe77c2c274819cd73c19acc8748d3ba

modules/nf-core/huggingface/download/tests/nextflow.config
toniher: Moved here: https://github.com/nf-core/modules/pull/11053/changes/421603e938aa087a09af17fcb8c1e13f693501cd

modules/nf-core/huggingface/download/tests/main.nf.test
toniher: Added new model here: https://github.com/nf-core/modules/pull/11053/changes/421603e938aa087a09af17fcb8c1e13f693501cd

modules/nf-core/llamacpp-python/run/tests/data/stub_model.gguf
toniher: They are somehow dummy files (specially the model for the stub, since it was an empty file). Now I generated them on the fly, unless you think it may be worth to have a prompt JSON on test-datasets.

modules/nf-core/huggingface/download/main.nf
toniher: It is a cache directory needed for `hf` to work. It is a bit annoying when using Docker (sic). The user can choose another location. What would you suggest instead? Maybe ${workDir}, kind of similarly as singularity images? Any other approach?

toniher: I must admit that is a bit of my bias that I normally like to place `singularity` images in projectDir instead of workDir :stuck_out_tongue:

modules/nf-core/llamacpp-python/run/tests/data/stub_model.gguf
famosab: If you need them to run the nf-test then they should be in test-datasets :) if you do not need them then they should be removed

famosab:
You can always propose modules after each other :) the approval is not dependent on that its basically just whether you adhere to our guidelines :)

modules/nf-core/llamacpp-python/run/tests/data/stub_model.gguf
toniher: I removed them already and generated them on the fly. So far, the test works this way... If OK with you this way, I collapse this.

lucacozzuto:
Something to keep in mind: the new models, like Gemma 4, are not supported by the llama-cpp-python version currently in the container.

modules/nf-core/huggingface/download/tests/main.nf.test
famosab: I meant that the test assertions should all go in the snapshot, for example filename etc. You can check how other modules do that :)

modules/nf-core/huggingface/download/tests/nextflow.config
famosab: I do not think you adjusted the way its described in the docs

modules/nf-core/huggingface/download/main.nf
```diff
@@ -0,0 +1,30 @@
...
+    """
+    export HF_HOME="${hf_home_resolved}"
+    export HF_HUB_CACHE=\$HF_HOME
+    hf download ${hf_repo} ${hf_file} --local-dir \$PWD
+    """
+
+    stub:
+    """
+    touch ${hf_file}
```
famosab, line 34: how is the input file being touched here the stub test?

toniher: The inputs are  `hf_repo` and `hf_file` . The output actually is the same name as the second input.
Like a file in a web server: https://example.com/myfile.gguf -> myfile.gguf

modules/nf-core/llamacpp-python/run/main.nf
```diff
@@ -0,0 +1,36 @@
...
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    llamacpp-python.py \
+        --model ${gguf_model} \
+        --messages ${prompt_file} \
+        --output output.txt \
```
famosab:
```suggestion
        --output ${prefix}.txt \
```

modules/nf-core/llamacpp-python/run/main.nf
```diff
@@ -0,0 +1,36 @@
...
+        ${args}
+    """
+
+    stub:
+    prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    touch output.txt
+
```
famosab:
```suggestion
    touch ${prefix}.txt
```

modules/nf-core/llamacpp-python/run/main.nf
```diff
@@ -0,0 +1,36 @@
...
+    label 'process_gpu'
+
+    conda "${moduleDir}/environment.yml"
+    container "${task.ext.use_gpu ? 'quay.io/nf-core/llama-cpp-python:0.1.9' : 'community.wave.seqera.io/library/llama-cpp-python:0.3.16--b351398cd0ea7fc5'}"
+
+    input:
+    tuple val(meta), path(prompt_file), path(gguf_model)
+
+    output:
+    tuple val(meta), path("output.txt"), emit: output
```
famosab:
```suggestion
    tuple val(meta), path("${prefix}.txt"), emit: output
```

modules/nf-core/llamacpppython/run/meta.yml
```diff
@@ -0,0 +1,69 @@
+name: llamacpp_python_run
+description: Python wrapper for running locally-hosted LLM with llama.cpp
+keywords:
```
famosab, line 3: can you add more keywords :)

modules/nf-core/llamacpppython/run/meta.yml
```diff
@@ -0,0 +1,69 @@
+name: llamacpp_python_run
```
famosab: According to the [module specifications](https://nf-co.re/docs/guidelines/components/modules#naming-conventions) we want the module name to be reflected in the process name. In this case we need to take care of the proper naming, it should be something like llamacpppy/run

modules/nf-core/llamacpppython/run/main.nf
```diff
@@ -0,0 +1,36 @@
+process LLAMACPP_PYTHON_RUN {
+    tag "${meta.id}"
+    label 'process_medium'
+    label 'process_gpu'
+
+    conda "${moduleDir}/environment.yml"
+    container "${task.ext.use_gpu ? 'quay.io/nf-core/llama-cpp-python:0.1.9' : 'community.wave.seqera.io/library/llama-cpp-python:0.3.16--b351398cd0ea7fc5'}"
```
famosab: that means on GPU the conda way will not be supported? Then we should catch this in the module script

modules/nf-core/llamacpp-python/run/tests/main.nf.test
```diff
@@ -0,0 +1,119 @@
...
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert process.out.output.size() == 1 },
+                { assert process.out.output[0][0] == [ id:'test_run' ] },
+                { assert file(process.out.output[0][1]).name == "output.txt" },
+                { assert snapshot(process.out.findAll { key, val -> key.startsWith('versions') }).match() }
```
famosab: Can we put the filename in the snapshot? Also it should not be called output.txt that is too generic :) see my other comments

famosab (commented):
Few more comments (some things I only commented once but they should be followed for all occurences) :)

modules/nf-core/llamacpp-python/run/tests/main.nf.test
toniher: This should be done it gets the name from meta.id -> prefix now

modules/nf-core/huggingface/download/tests/main.nf.test
toniher: Only one file is actually the output: `gemma-3-1b-it-Q4_K_M.gguf` Are there too many assert statements for only one file, maybe?

modules/nf-core/llamacpppython/run/main.nf
toniher: My personal workstation GPU is a bit oldie, but with `nf-test` I can see that the GPU is detected with `conda`:
```
ggml_cuda_init: GGML_CUDA_FORCE_MMQ:    no
ggml_cuda_init: GGML_CUDA_FORCE_CUBLAS: no
ggml_cuda_init: found 1 CUDA devices:
  Device 0: Quadro P2200, compute capability 6.1, VMM: yes
llama_model_load_from_file_impl: using device CUDA0 (Quadro P2200) - 3921 MiB free
```
So I assume yes...

famosab: why do we then use a  different custom container if the gpu is needed?

modules/nf-core/huggingface/download/tests/main.nf.test
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert process.out.output.size() == 1 },
+                { assert process.out.output[0][0] == [ id:'test_model_smollm3' ] },
+                { assert file(process.out.output[0][1]).name == "SmolLM3-3B-UD-IQ2_XXS.gguf" },
+                { assert file(process.out.output[0][1]).size() > 0 },
+                { assert snapshot(process.out.findAll { key, val -> key.startsWith('versions') }).match() }
```
famosab: I think we can put all these assertions in the snapshot :) then its easier to see if something is different on the CI!

modules/nf-core/huggingface/download/main.nf
toniher: I changed it to workDir by default

modules/nf-core/llamacpppython/run/main.nf
toniher: For conda, gpu-enabled llama-cpp-python seems to be enabled during the build/installation / package resolution process. However, if we want to take advantage of GPU capabilities in containers, we need different ones :( 
By the way, I changed `task.ext.use_gpu` to the most generic `task.accelerator`

modules/nf-core/huggingface/download/tests/main.nf.test
toniher: This is done https://github.com/nf-core/modules/pull/11053/commits/cd3d1b96df856c972a373880cb352dfefe58e4e3

modules/nf-core/huggingface/download/tests/nextflow.config
toniher: Sorry, can you point me what you spot it is not OK?

modules/nf-core/huggingface/download/tests/main.nf.test
toniher: This is addressed below as well. All assertions of a test in the snapshot.

modules/nf-core/llamacpppython/run/main.nf
famosab: ok I see, is there a use case where people use this and do not want to use GPUs?

toniher: If their machine, e.g., my laptop, does not have GPU. On the other hand, if the machine has not a GPU and you try to use the GPU-enabled container image, it will fail, since it does not detect any GPU-device (actually CUDA-compatible one).

modules/nf-core/huggingface/download/tests/nextflow.config
toniher: @famosab I think that this is the last bit of the review. Thanks a lot for the patience. Let me know what could be missing here...

famosab: I think we are good now, although I am not entirely sure whether we would handle the containerOptions like this but I think its ok.

famosab:
I think from my side it looks good, but now I spend so much time back-and-forth with you and this module is a bit unusual :) so I would like to have another set of eyes on this PR! I will ask for another review :)

pinin4fjords (commented):
I really think we should avoid using module binaries until we're satisfied they don't limit module portability. Currently [Wave is required](https://www.nextflow.io/docs/latest/module.html#module-binaries) for Cloud scenarios without a shared file system. 

Templates are the more portable, if less pretty, approach.

modules/nf-core/llamacpppython/run/main.nf
```diff
@@ -0,0 +1,35 @@
+process LLAMACPPPYTHON_RUN {
+    tag "${meta.id}"
+    label 'process_medium'
+    label 'process_gpu'
+
+    conda "${moduleDir}/environment.yml"
+    container "${task.accelerator ? 'quay.io/nf-core/llama-cpp-python:0.1.9' : 'community.wave.seqera.io/library/llama-cpp-python:0.3.16--b351398cd0ea7fc5'}"
```
pinin4fjords: Are you sure you need this? A little bit of AI chat suggests that:

```
llama-cpp-python compiled with CUDA support does runtime GPU detection and falls back to CPU when no GPU is present
```

... which, if correct, would mean you could just supply the same container in either case and avoid the complexity.

You might be able to do a multi-stage thing like this to bring the container size down:

```
  FROM nvidia/cuda:12.4.1-devel-ubuntu22.04 AS builder                                                                         
   
  RUN apt-get update && apt-get install -y python3 python3-pip python3-dev                                                     
  RUN pip3 install --prefix=/install llama-cpp-python==0.3.16 \
      --extra-index-url https://abetlen.github.io/llama-cpp-python/whl/cu124                                                   
                  
  FROM nvidia/cuda:12.4.1-runtime-ubuntu22.04                                                                                  
                  
  RUN apt-get update && apt-get install -y --no-install-recommends python3 python3-pip \                                       
      && rm -rf /var/lib/apt/lists/*
                                                                                                                               
  COPY --from=builder /install /usr/local
```

... so it's less overhead for the non-GPU case (untested).

modules/nf-core/llamacpppython/run/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM nvidia/cuda:12.4.1-devel-ubuntu22.04
+
+RUN apt-get update && apt-get install -y python3 python3-pip
+RUN pip3 install llama-cpp-python  \
```
pinin4fjords: If you're doing this you should definitely use the same pins as the environment.yml.

modules/nf-core/llamacpppython/run/main.nf
```diff
@@ -0,0 +1,35 @@
...
+    tuple val(meta), path(prompt_file), path(gguf_model)
+
+    output:
+    tuple val(meta), path("${prefix}.txt"), emit: output
+    tuple val("${task.process}"), val("llama-cpp-python"), eval("python3 -c 'import llama_cpp; print(llama_cpp.__version__)'"), topic: versions, emit: versions_llama_cpp_python
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
```
pinin4fjords, line 19: If you can't make conda work with GPU, you should have a guard similar to how the Parbricks modules work:

```
      exit(1, "Parabricks module does not support Conda. Please use Docker / Singularity / Podman instead.")
  }    
```

modules/nf-core/llamacpppython/run/main.nf
pinin4fjords: No singularity?

modules/nf-core/huggingface/download/main.nf
```diff
@@ -0,0 +1,30 @@
...
+
+    output:
+    tuple val(meta), path(hf_file), emit: output
+    tuple val("${task.process}"), val("huggingface_hub"), eval("hf --version 2>&1 | tail -n1 | awk '{print \$NF}'"), topic: versions, emit: versions_huggingface_hub
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def hf_home_resolved = hf_home ?: "${workflow.workDir}/hf_cache"
```
pinin4fjords: Why does this need to be outside the process dir? I'm not sure if this will even work on cloud executors without a shared file system.

pinin4fjords: Think this needs to be something relative to the task. If you need to share something across tasks you need to generate it earlier in the workflow and and actually pass it to the tasks with a `path` input.

pinin4fjords (changes requested):
I think there's a few things to resolve here since we're colouring outside the lines a bit.

modules/nf-core/llamacpppython/run/Dockerfile
mashehu, line 1: since this tool is on conda-forge, you don't need to add a dockerfile, just use seqera containers https://nf-co.re/docs/developing/containers/seqera-containers

modules/nf-core/llamacpppython/run/meta.yml
```diff
@@ -0,0 +1,72 @@
+name: llamacpppython_run
+description: Python wrapper for running locally-hosted LLM with llama.cpp
+keywords:
+  - ai
```
mashehu:
```suggestion
```

modules/nf-core/huggingface/download/main.nf
toniher: In principle, once the image is generated, it should not be needed any more if other processes use that image. The point of sharing it beyond the task is for reusing the image cache if it is triggered again. We can put in the taskDir, but I think it is good that the user can tune it if needed.

modules/nf-core/llamacpppython/run/main.nf
toniher: It appears that GPU works with Conda: https://github.com/nf-core/modules/pull/11053#discussion_r3053862840 - Check messages there...

toniher:
> I really think we should avoid using module binaries until we're satisfied they don't limit module portability. Currently [Wave is required](https://www.nextflow.io/docs/latest/module.html#module-binaries) for Cloud scenarios without a shared file system.
> 
> Templates are the more portable, if less pretty, approach.

Any module example you could suggest as a model suitable for this case?

modules/nf-core/huggingface/download/main.nf
bentsherman: Having the cache in the top-level work dir is fine, but the work-dir path should be passed as a process input so that you don't have to reference `workflow.workDir` in the process definition

pinin4fjords:
> > I really think we should avoid using module binaries until we're satisfied they don't limit module portability. Currently [Wave is required](https://www.nextflow.io/docs/latest/module.html#module-binaries) for Cloud scenarios without a shared file system.
> > Templates are the more portable, if less pretty, approach.
> 
> Any module example you could suggest as a model suitable for this case?

Just look for the `template` keyword in the modules repo. But here are a couple of examples:

- https://github.com/nf-core/modules/tree/master/modules/nf-core/ea-utils/gtf2bed
- https://github.com/nf-core/modules/blob/master/modules/nf-core/custom/catadditionalfasta/main.nf

Note that these are not quite script files, and you have to escape `$` etc with `\`.

modules/nf-core/llamacpppython/run/main.nf
toniher: I just tried again running Docker in my laptop (without GPU) using `quay.io/nf-core/llama-cpp-python:0.1.9` and I cannot make it work.
```
│     RuntimeError: Failed to load shared library '/usr/local/lib/python3.10/dist-packages/llama_cpp/lib/libllama.so': libcuda.so.1: cannot open shared object file: No    │
│ such file or directory
```
So, it looks that after building the Docker image with GPU support, you cannot run it without any GPU device. It's particularly annoying because there are different GPU types and CUDA versions. :(

modules/nf-core/llamacpppython/run/meta.yml
toniher: I guess 'ai' is too generic. I removed from both meta.yml

modules/nf-core/llamacpppython/run/Dockerfile
toniher: I am actually using seqera containers for no gpu situations.

https://seqera.io/containers/?packages=conda-forge::llama-cpp-python=0.3.16

I understand I could approach it similarly as here:
https://github.com/nf-core/modules/blob/dd6396bf2d3fff18f4997613ae5ff0dc8ce52aaa/modules/nf-core/multiqc/meta.yml#L110

Until there is any better solution, I could enable when it is, at the same time containers, accelerator (and amd64?) to use `quay.io/nf-core/llama-cpp-python:0.1.9` and for the rest of situations to do as with the multiqc module.
Otherwise, I could simply remove the gpu container until there must be a better idea. It must be said that the speed gain is huge with gpu...
Any feedback is appreciated!

modules/nf-core/llamacpppython/run/main.nf
toniher: Related discussed here: https://github.com/nf-core/modules/pull/11053#discussion_r3071963438

modules/nf-core/huggingface/download/main.nf
toniher: I removed those defaults from the module.

toniher:
@pinin4fjords I tried to make it work, taking your modules as examples. It was relatively easy using:

```
if __name__ == "__main__":
    main("--model ${gguf_model} --messages ${prompt_file} --output ${prefix}.txt ${args}")
```
and a bit more code bits, but... I got another issue when running the test:

```
│   Caused by:                                                                              │
│     Process output of type 'eval' is only allowed with Bash process scripts -- Current    │
│ interpreter: /usr/bin/env python3
```
This is caused by the recent convention in the way to generate versions...
```
    tuple val("${task.process}"), val("llama-cpp-python"), eval("python3 -c 'import llama_cpp; print(llama_cpp.__version__)'"), topic: versions, emit: versions_llama_cpp_python
```

So, I guess I would need to generate `versions.yml` the old way back both inside the script and in the stub  😿

pinin4fjords:
> So, I guess I would need to generate `versions.yml` the old way back both inside the script and in the stub 😿

Yep, the eval won't work, just do exactly what the gtf2bed example I gave you does. You still output the versions.yml, but also send it to a topic. The code used by pipelines to assemble versions handles both files and strings.

modules/nf-core/llamacpppython/run/main.nf
toniher: @pinin4fjords Added Singularity equivalent https://github.com/nf-core/modules/pull/11053/commits/9069b44a190b42f39bd01c66a43a59e0be57d15c - What else would you miss so far considering the present restrictions?

toniher: Hi, any feedback? So far, the only weak point I see is the way to handle container GPU-enabled scenarios. Please let me know!

pinin4fjords:
Did a bit of AI-assisted iteration to make https://github.com/biocorecrg/nf-core-modules/pull/1 - see what you think. We're still trying to figure out the standard patterns for this- see https://github.com/nf-core/website/pull/4142.

toniher:
> Did a bit of AI-assisted iteration to make [biocorecrg#1](https://github.com/biocorecrg/nf-core-modules/pull/1) - see what you think. We're still trying to figure out the standard patterns for this- see [nf-core/website#4142](https://github.com/nf-core/website/pull/4142).

Thanks @pinin4fjords !  It took me a while (sic), but I finally merged it. It seems to pass the tests, but I want to test it locally with nf-core CLI or nf-test and ensure that is actually working with accelerator. How do you recommend doing this?
Also, let's say, I would like to upgrade the wave/GPU containers for a new version. How should I proceed? https://seqera.io/containers/ does not allow this flexibility, I think. Should I try a normal run and extract the generated container? How are you doing that? I imagine all this might be worth being added for the documentation as well.

modules/nf-core/llamacpppython/run/main.nf
```diff
@@ -0,0 +1,35 @@
...
+    template('llama-cpp-python.py')
+
+    stub:
+    prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    touch ${prefix}.txt
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        llama-cpp-python: \$(python3 -c 'import llama_cpp; print(llama_cpp.__version__)')
```
pinin4fjords, line 32:
```suggestion
        llama-cpp-python: \$(python3 -c 'import llama_cpp; print(llama_cpp.__version__)')
        cuda: no CUDA available
```

modules/nf-core/llamacpppython/run/templates/llama-cpp-python.py
```diff
@@ -0,0 +1,146 @@
...
+    )
+
+
+def write_versions():
+    versions = {"${task.process}": {"llama-cpp-python": llama_cpp.__version__}}
+    with open("versions.yml", "w", encoding="utf-8") as f:
+        for process, pkgs in versions.items():
+            f.write(f'"{process}":\\n')
+            for pkg, ver in pkgs.items():
+                f.write(f"    {pkg}: {ver}\\n")
```
pinin4fjords, line 161:
```suggestion
def get_cuda_version():
    """Return CUDA runtime version (major.minor) or 'no CUDA available'."""
    import ctypes

    for major in (13, 12, 11):
        try:
            cudart = ctypes.CDLL(f"libcudart.so.{major}")
        except OSError:
            continue
        v = ctypes.c_int()
        if cudart.cudaRuntimeGetVersion(ctypes.byref(v)) == 0:
            return f"{v.value // 1000}.{(v.value % 1000) // 10}"
    return "no CUDA available"


def write_versions():
    versions = {
        "${task.process}": {
            "llama-cpp-python": llama_cpp.__version__,
            "cuda": get_cuda_version(),
        }
    }
    with open("versions.yml", "w", encoding="utf-8") as f:
        for process, pkgs in versions.items():
            f.write(f'"{process}":\\n')
            for pkg, ver in pkgs.items():
                f.write(f"    {pkg}: {ver}\\n")
```

modules/nf-core/llamacpppython/run/tests/nextflow.gpu.config
```diff
@@ -0,0 +1,6 @@
+process {
+    withName: 'LLAMACPPPYTHON_RUN' {
+        accelerator = 1
+        ext.args    = '--seed 42'
```
pinin4fjords, line 4:
```suggestion
        ext.args    = '--seed 42'
        maxForks    = 1
```

modules/nf-core/llamacpppython/run/tests/main.nf.test.snap
```diff
@@ -0,0 +1,41 @@
...
+        "content": [
+            1,
+            {
+                "id": "test_run"
+            },
+            "test_run.txt",
+            true,
+            {
+                "versions_llama_cpp_python": [
+                    "versions.yml:md5,e4e5fd3eefb4e0b1e83d8766a8b52e7c"
```
pinin4fjords:
```suggestion
                    "versions.yml:md5,3650e9ce14ab350b69387f6631b57885"
```

modules/nf-core/llamacpppython/run/tests/main.nf.test.snap
```diff
@@ -0,0 +1,41 @@
...
+    "stub - run inference with json prompt": {
+        "content": [
+            1,
+            {
+                "id": "test_run"
+            },
+            "test_run.txt",
+            {
+                "versions_llama_cpp_python": [
+                    "versions.yml:md5,e4e5fd3eefb4e0b1e83d8766a8b52e7c"
```
pinin4fjords:
```suggestion
                    "versions.yml:md5,3650e9ce14ab350b69387f6631b57885"
```

pinin4fjords (commented):
Per the (draft) [GPU module guidelines](https://github.com/nf-core/website/pull/4142):

- Emit CUDA runtime version on the `versions` topic (CPU path falls back to `no CUDA available`)
- `maxForks = 1` on GPU tests for Singularity GPU concurrency
- Snap md5s included so you don't need to rerun

Verified locally: CPU path md5 `3650e9ce14ab350b69387f6631b57885`, real GPU path md5 `28ef3c6aa93df0de54e66c13cc9db845`.

modules/nf-core/llamacpppython/run/tests/main.gpu.nf.test.snap
```diff
@@ -0,0 +1,41 @@
...
+    "llamacpppython/run - GPU stub run inference with json prompt": {
+        "content": [
+            1,
+            {
+                "id": "test_run"
+            },
+            "test_run.txt",
+            {
+                "versions_llama_cpp_python": [
+                    "versions.yml:md5,e4e5fd3eefb4e0b1e83d8766a8b52e7c"
```
pinin4fjords: Stub path - same md5 as CPU runs.

```suggestion
                    "versions.yml:md5,3650e9ce14ab350b69387f6631b57885"
```

modules/nf-core/llamacpppython/run/tests/main.gpu.nf.test.snap
```diff
@@ -0,0 +1,41 @@
...
+        "content": [
+            1,
+            {
+                "id": "test_run"
+            },
+            "test_run.txt",
+            true,
+            {
+                "versions_llama_cpp_python": [
+                    "versions.yml:md5,e4e5fd3eefb4e0b1e83d8766a8b52e7c"
```
pinin4fjords: Real GPU run.

```suggestion
                    "versions.yml:md5,28ef3c6aa93df0de54e66c13cc9db845"
```

pinin4fjords:
Sorry for the AI language above, just thought I'd help out by fixing this up for you- all the above are tested so you should just be able to accept the suggestions. It's my fault, I amended the proposed guidelines a little in response to my own testing/ feedback thereon. Now, to your question...

pinin4fjords:
Glad it's in! On testing and Wave upgrades:

**Testing the GPU path locally**

You need three things on the host: a GPU, `nvidia-container-toolkit` (Docker) or NVIDIA drivers (Singularity), and the modules-repo `gpu` profile composed onto your container profile - that profile (in `tests/config/nf-test.config`) is what adds `--gpus all` / `--nv`. `nf-core modules test --profile X` only accepts `docker | singularity | conda` so you can't use it for the GPU run; drive `nf-test` directly:

```bash
nf-test test --tag llamacpppython/run --profile docker,gpu \
    modules/nf-core/llamacpppython/run/tests/main.gpu.nf.test
```

That dispatches the GPU container, runs real inference, and exercises the `versions.yml` shape end-to-end. The CI workflow `.github/workflows/nf-test-gpu.yml` does the same on a GPU runner.

If you don't have a GPU box to hand, a `g4dn.xlarge` on AWS (Deep Learning Base GPU AMI, Tesla T4, under $1/hr on-demand) takes ~70 seconds to run both GPU tests once the container and model are pulled. I just verified the current PR head + the open review suggestions there - both GPU tests pass and the snap md5s match (`3650e9ce…` for the stub, `28ef3c6a…` for the real run).

Without composing `gpu` in, the GPU test isn't a no-op - it fails at the `python3 -c 'import llama_cpp; …'` call inside the container, because `llama_cpp` opens `libcudart` at import time and the runtime has no GPU. So `--profile docker` on its own won't tell you anything useful about the accelerator path.

**Upgrading the Wave containers**

`seqera.io/containers` doesn't expose `--freeze` / `--build-template`, so I use the CLI directly. For this module:

```bash
# CPU
wave --conda-file environment.yml --freeze --await
wave --conda-file environment.yml --freeze --await --singularity

# GPU (pip wheel + cuda-runtime)
wave --conda-file environment.gpu.yml --freeze --await \
     --build-template conda/micromamba:v2 \
     --config-env 'LD_LIBRARY_PATH=/opt/conda/lib'
wave --conda-file environment.gpu.yml --freeze --await --singularity \
     --build-template conda/micromamba:v2 \
     --config-env 'LD_LIBRARY_PATH=/opt/conda/lib'
```

`--config-env LD_LIBRARY_PATH=/opt/conda/lib` is needed because conda's `activate.d` hooks don't fire under `docker run`, so the wheel's binary can't otherwise find the conda-provided CUDA libs at `dlopen` time.

Each invocation prints a frozen tag - `community.wave.seqera.io/library/…:<hash>` for Docker, `oras://community.wave.seqera.io/library/…:<hash>` for Singularity. Those four URLs go into the `container` ternary.

The Wave guidance is in the proposed module specs (https://github.com/nf-core/website/pull/4142), under "GPU-capable modules" and "Pip-based GPU packages" - probably worth waiting for that to land rather than duplicating it elsewhere.

pinin4fjords:
One more handy bit since you don't have a GPU on hand: if a future change shifts the GPU snap md5s, you can ask the nf-core bot to regenerate them on a CI runner with the right hardware by commenting on the PR:

```
@nf-core-bot update gpu snapshot path:modules/nf-core/llamacpppython/run
```

(works the same for any GPU-tagged module - I used it on #11258 for ribodetector). Saves spinning up an instance just to refresh a md5.

pinin4fjords:
> One more handy bit since you don't have a GPU on hand: if a future change shifts the GPU snap md5s, you can ask the nf-core bot to regenerate them on a CI runner with the right hardware by commenting on the PR:
> 
> ```
> @nf-core-bot update gpu snapshot path:modules/nf-core/llamacpppython/run
> ```
> 
> (works the same for any GPU-tagged module - I used it on #11258 for ribodetector). Saves spinning up an instance just to refresh a md5.

and of course this info triggered it!

modules/nf-core/huggingface/download/tests/main.nf.test
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(sanitizeOutput(process.out)).match() }
+            )
+        }
+    }
+}
```
pinin4fjords: Two HF internal-testing repos for variety, ~800 B each instead of ~1.5 GB.

```suggestion
    test("download config.json - tiny-random-gpt2") {

        when {
            process {
                """
                input[0] = [
                    [ id:'test_gpt2' ],
                    "hf-internal-testing/tiny-random-gpt2",
                    "config.json"
                ]
                """
            }
        }

        then {
            assertAll(
                { assert process.success },
                { assert snapshot(
                    process.out.output.size(),
                    process.out.output[0][0],
                    file(process.out.output[0][1]).name,
                    file(process.out.output[0][1]).size() > 0,
                    process.out.findAll { key, val -> key.startsWith('versions') }
                ).match() }
            )
        }
    }

    test("download config.json - tiny-random-bert") {

        when {
            process {
                """
                input[0] = [
                    [ id:'test_bert' ],
                    "hf-internal-testing/tiny-random-bert",
                    "config.json"
                ]
                """
            }
        }

        then {
            assertAll(
                { assert process.success },
                { assert snapshot(
                    process.out.output.size(),
                    process.out.output[0][0],
                    file(process.out.output[0][1]).name,
                    file(process.out.output[0][1]).size() > 0,
                    process.out.findAll { key, val -> key.startsWith('versions') }
                ).match() }
            )
        }
    }

    test("stub - download config.json - tiny-random-gpt2") {

        options "-stub"

        when {
            process {
                """
                input[0] = [
                    [ id:'test_gpt2' ],
                    "hf-internal-testing/tiny-random-gpt2",
                    "config.json"
                ]
                """
            }
        }

        then {
            assertAll(
                { assert process.success },
                { assert snapshot(sanitizeOutput(process.out)).match() }
            )
        }
    }
}
```

modules/nf-core/huggingface/download/tests/main.nf.test.snap
```diff
@@ -0,0 +1,76 @@
...
+                ]
+            }
+        ],
+        "timestamp": "2026-04-09T16:49:18.323499722",
+        "meta": {
+            "nf-test": "0.9.5",
+            "nextflow": "25.10.4"
+        }
+    }
+}
```
pinin4fjords, line 76: Snap matches the renamed tests (stub md5 unchanged - empty file from `touch` either way).

```suggestion
{
    "stub - download config.json - tiny-random-gpt2": {
        "content": [
            {
                "output": [
                    [
                        {
                            "id": "test_gpt2"
                        },
                        "config.json:md5,d41d8cd98f00b204e9800998ecf8427e"
                    ]
                ],
                "versions_huggingface_hub": [
                    [
                        "HUGGINGFACE_DOWNLOAD",
                        "huggingface_hub",
                        "1.6.0"
                    ]
                ]
            }
        ],
        "timestamp": "2026-04-04T22:00:17.896195894",
        "meta": {
            "nf-test": "0.9.5",
            "nextflow": "25.10.4"
        }
    },
    "download config.json - tiny-random-gpt2": {
        "content": [
            1,
            {
                "id": "test_gpt2"
            },
            "config.json",
            true,
            {
                "versions_huggingface_hub": [
                    [
                        "HUGGINGFACE_DOWNLOAD",
                        "huggingface_hub",
                        "1.6.0"
                    ]
                ]
            }
        ],
        "timestamp": "2026-04-09T16:48:14.073310733",
        "meta": {
            "nf-test": "0.9.5",
            "nextflow": "25.10.4"
        }
    },
    "download config.json - tiny-random-bert": {
        "content": [
            1,
            {
                "id": "test_bert"
            },
            "config.json",
            true,
            {
                "versions_huggingface_hub": [
                    [
                        "HUGGINGFACE_DOWNLOAD",
                        "huggingface_hub",
                        "1.6.0"
                    ]
                ]
            }
        ],
        "timestamp": "2026-04-09T16:49:18.323499722",
        "meta": {
            "nf-test": "0.9.5",
            "nextflow": "25.10.4"
        }
    }
}
```

pinin4fjords (commented):
Spotted while looking over the diff: the four real tests pull ~3.2 GB of GGUFs from HuggingFace per fresh CI run (Gemma-1B, SmolLM3-3B). For the inference tests in `llamacpppython/run`, swapping to `mradermacher/tiny-gemma-test-i1-GGUF` (~14 MB, same Gemma architecture, loads cleanly in `llama-cpp-python` 0.3.16 - I verified) keeps full coverage.

For `huggingface/download`, the tests only assert that download works, so they don't need a model at all - any file from any HF repo proves the path. `hf-internal-testing/tiny-random-*` repos exist for exactly this purpose; swapping to their `config.json` files (~800 B each) brings the download down to under 2 KB.

Total: ~3.2 GB → ~14 MB per run.

modules/nf-core/llamacpppython/run/tests/main.nf.test
```diff
@@ -0,0 +1,114 @@
...
+        setup {
+            run("HUGGINGFACE_DOWNLOAD") {
+                script "../../../huggingface/download/main.nf"
+                process {
+                    """
+                    input[0] = [
+                        [ id:'test_model' ],
+                        "ggml-org/gemma-3-1b-it-GGUF",
+                        "gemma-3-1b-it-Q4_K_M.gguf"
+                    ]
```
pinin4fjords, line 26: Tiny-Gemma quant from mradermacher: same architecture, ~14 MB, verified loads in `llama-cpp-python` 0.3.16. Snapshot stays valid since the test only asserts `size > 0`.

```suggestion
                    input[0] = [
                        [ id:'test_model' ],
                        "mradermacher/tiny-gemma-test-i1-GGUF",
                        "tiny-gemma-test.i1-Q4_K_M.gguf"
                    ]
```

modules/nf-core/llamacpppython/run/tests/main.gpu.nf.test
```diff
@@ -0,0 +1,115 @@
...
+
+        setup {
+            run("HUGGINGFACE_DOWNLOAD") {
+                script "../../../huggingface/download/main.nf"
+                process {
+                    """
+                    input[0] = [
+                        [ id:'test_model' ],
+                        "ggml-org/gemma-3-1b-it-GGUF",
+                        "gemma-3-1b-it-Q4_K_M.gguf"
```
pinin4fjords: Same swap as the CPU test - tiny-Gemma in place of full Gemma-1B.

```suggestion
                    input[0] = [
                        [ id:'test_model' ],
                        "mradermacher/tiny-gemma-test-i1-GGUF",
                        "tiny-gemma-test.i1-Q4_K_M.gguf"
                    ]
```

pinin4fjords:
Just looked at the conda CI failures - they're the same root cause as the size review I just posted. The conda jobs are timing out at exactly 2h on the SmolLM3 download:

```
ERROR ~ Error executing process > 'HUGGINGFACE_DOWNLOAD (test_model_smollm3)'
Caused by:
  Process exceeded running time limit (2h)
```

(from [`x64 | conda | 2`](https://github.com/nf-core/modules/actions/runs/25429023287/job/74590338563))

The `hf download` is just sitting there for 2h - likely a slower mirror or network issue under the conda env's `huggingface_hub` 1.6.0. The suggestions in the review above replace SmolLM3 (~860 MB) and Gemma-1B (~700 MB) with `hf-internal-testing/tiny-random-{gpt2,bert}/config.json` (~800 B each) and `mradermacher/tiny-gemma-test-i1-GGUF` (~14 MB), so the conda jobs would just complete in seconds.

Applying the suggestions should fix the conda red ticks too.

pinin4fjords (approved):
Approving to unblock you- I think that smaller model download thing should fix up the rest of it and get you passing CI, as long as you don't see a problem there.

pinin4fjords:
Note: the linting failure is not your fault, it's pending https://github.com/nf-core/modules/pull/11543

pinin4fjords:
@toniher can you merge in the latest master please? I don't have permission to on your fork.

toniher:
Thanks everyone for all the involved work! 🥳

PR #11064, opened 2026-03-27
tabix/tabix: update module to support region-based VCF extraction

luisas:
Add support for region-based VCF extraction

## PR checklist

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:


modules/nf-core/tabix/tabix/main.nf
```diff
@@ -8,28 +8,40 @@ process TABIX_TABIX {
         'community.wave.seqera.io/library/htslib:1.21--ff8e28a189fbecaa' }"
 
     input:
-    tuple val(meta), path(tab)
+    tuple val(meta),  path(tab)
+    tuple val(meta2), path(regions)
```
famosab: Please make this into one input tuple :)
```suggestion
    tuple val(meta),  path(tab), path(regions)
```

modules/nf-core/tabix/tabix/main.nf
```diff
@@ -8,28 +8,40 @@ process TABIX_TABIX {
...
 
     output:
-    tuple val(meta), path("*.{tbi,csi}"), emit: index
+    tuple val(meta), path("*.{tbi,csi}"),   emit: index, optional: true
+    tuple val(meta), path("${prefix}.vcf"), emit: vcf,   optional: true
```
famosab: Can we emit this bgzipped please? :)
```suggestion
    tuple val(meta), path("${prefix}.vcf.gz"), emit: vcf,   optional: true
```

famosab (commented):
Ok I started this review but then I realized that this is in my eyes a different module. Maybe @nf-core/maintainers have the same opinion but I would rather have this separate because from tabix/tabix I just expect indexing and I would otherwise call it tabix/extract or something similar.

Edit: I just saw that that was you original idea and now I am a little confused what led to the change :D

luisas:
Hi @famosab, thanks for starting the review! 

I started doing it like tabix_extract but then I had a chat with @FriederikeHanssen on how to handle these modules that have different optional outputs depending on the input flags and because of maintainance burdens we thought it would be better to have it in one module? Similar to [samtools/view](https://github.com/nf-core/modules/blob/master/modules/nf-core/samtools/view/main.nf) 

I am torn tbh, I am also liking the tabix/extract version.

modules/nf-core/tabix/tabix/main.nf
luisas: They are actually 2 different files that I believe conceptually work better in 2 tuples: the first one is the input file (with its own meta) and the second one is the subsetting instructions file (also should probably have its own meta) - do you think I could leave it separate?

adamrtalbot: One might want to do sample specific subsetting, which is much easier with 1 tuple.

It's easier to join inputs then it is to pass two inputs together in sync 😉 

This is one of the reasons we are moving to a flatter input structure.

luisas: Oh ok! Fixing it then :)

modules/nf-core/tabix/tabix/main.nf
luisas: I saw the comment after I had committed something already so I could not commit your suggestion as it now shows as outdated. But it is now fixed in the code :)

famosab:
I get the point! I am at times unsure maybe we need to discuss this also in the maintainers meeting.

If I follow your arguments I would then vote for the other submodules of tabix being deprecated (because what makes them different "enough") or add this as a separate module. But I would find it confusing that some things are excluded and some are part of the "base" module. Do you get what I mean? Maybe I am a bit strict on this matter but I want to have a structure that is understandable and sort of follows the same arguments? We could also have a vote or ask for more opinions in the slack and then adjust all tabix modules? (sorry for making this bigger than it is)

luisas:
> I get the point! I am at times unsure maybe we need to discuss this also in the maintainers meeting.
> 
> If I follow your arguments I would then vote for the other submodules of tabix being deprecated (because what makes them different "enough") or add this as a separate module. But I would find it confusing that some things are excluded and some are part of the "base" module. Do you get what I mean? Maybe I am a bit strict on this matter but I want to have a structure that is understandable and sort of follows the same arguments? We could also have a vote or ask for more opinions in the slack and then adjust all tabix modules? (sorry for making this bigger than it is)


I see the point. Doing tabix/extract was also my original intuition but I also see the other side of having something like samtools/view. Pinging  @FriederikeHanssen @adamrtalbot so we can decide how to proceed

luisas:
> > I get the point! I am at times unsure maybe we need to discuss this also in the maintainers meeting.
> > If I follow your arguments I would then vote for the other submodules of tabix being deprecated (because what makes them different "enough") or add this as a separate module. But I would find it confusing that some things are excluded and some are part of the "base" module. Do you get what I mean? Maybe I am a bit strict on this matter but I want to have a structure that is understandable and sort of follows the same arguments? We could also have a vote or ask for more opinions in the slack and then adjust all tabix modules? (sorry for making this bigger than it is)
> 
> I see the point. Doing tabix/extract was also my original intuition but I also see the other side of having something like samtools/view. Pinging @FriederikeHanssen @adamrtalbot so we can decide how to proceed

@famosab given the discussion on the team-maintainers team in slack there seems not to be the general solution for every module but there is a general agreement on trying to keep options in the same module.

Here there is also 2 other modules under tabix but they are using different command lines (bgzip, bgzip + tabix, tabix) 
This structure to me makes sense as it keeps a motivated separation (which tools or combination of tools is used) but then groups all the different command lines and outputs under each of these modules. 

What do you think?

modules/nf-core/tabix/tabix/meta.yml
```diff
@@ -23,6 +27,20 @@ input:
         description: TAB-delimited genome position file compressed with bgzip
         pattern: "*.{bed.gz,gff.gz,sam.gz,vcf.gz}"
         ontologies: []
+    - tai:
```
famosab, line 31:
```suggestion
    - index:
```

modules/nf-core/tabix/tabix/meta.yml
```diff
@@ -32,9 +50,21 @@ output:
...
           ontologies: []
+  vcf:
+    - - meta:
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. [ id:'test', single_end:false ]
+      - "${prefix}.vcf.gz":
+          type: file
+          description: bgzipped VCF of extracted regions. Emitted when a regions file is provided.
```
famosab:
```suggestion
          description: Bgzipped VCF of extracted regions. Emitted when a regions file is provided.
```

modules/nf-core/tabix/tabix/main.nf
```diff
@@ -8,28 +8,39 @@ process TABIX_TABIX {
...
 
     when:
     task.ext.when == null || task.ext.when
 
     script:
     def args = task.ext.args ?: ''
+    prefix          = task.ext.prefix ?: "${meta.id}"
+    def regions_arg = regions ? "-R ${regions}" : ""
+    def output_arg  = regions ? "| bgzip --threads ${task.cpus} > ${prefix}.vcf.gz" : ""
+
```
famosab:
```suggestion
```

modules/nf-core/tabix/tabix/main.nf
```diff
@@ -8,28 +8,39 @@ process TABIX_TABIX {
...
+        ${output_arg}
 
     """
     stub:
     def args = task.ext.args ?: ''
-    def index = args.contains("-C ") || args.contains("--csi") ? "csi" : "tbi"
+    prefix = task.ext.prefix ?: "${meta.id}"
+    def ext = args.contains("-C ") || args.contains("--csi") ? "csi" : "tbi"
+    def index = regions ? "" : "touch ${tab}.${ext}"
+    def vcf   = regions ? "touch ${prefix}.vcf.gz" : ""
```
famosab:
```suggestion
    def vcf   = regions ? "echo | gzip > ${prefix}.vcf.gz" : ""
```

modules/nf-core/tabix/tabix/tests/main.nf.test
```diff
@@ -106,6 +114,57 @@ nextflow_process {
...
+                                file(params.modules_testdata_base_path + 'genomics/sarscov2/genome/bed/test.bed', checkIfExists: true)
+                            ]
+                """
+            }
+        }
+
+        then {
+            assertAll (
+                { assert process.success },
+                { assert snapshot(sanitizeOutput(process.out, unstableKeys: ["vcf"])).match() }
```
famosab: Can we not use nft-vcf to test the md5sum of the vcf?

modules/nf-core/tabix/tabix/tests/main.nf.test
```diff
@@ -106,6 +114,57 @@ nextflow_process {
...
```
famosab:
```suggestion
                { assert snapshot(sanitizeOutput(process.out)).match() }
```
It is a stub

famosab (commented):
Few more requests, I am fine with keeping this in tabix/tabix

modules/nf-core/tabix/tabix/tests/main.nf.test
luisas: I did not know about that - thank you! Will add now

luisas:
@famosab ok now it should be all done!

Thank you a lot for the help!

muffato:
Hi. I like the extra option in the module but my problem is that tabix is for accessing data in _any_ TSV file, not just VCF. It can work with all BED files, GFF, etc, as long as the right column numbers / preset was passed.
In that sense, I wouldn't want `vcf` to be hardcoded as the output file's extension.
Can you instead extract the suffix from the `tab` parameter ?

famosab (approved):
Nice work 🚀

PR #11087, opened 2026-03-30
Add caalm

vagkaratzas:
New protein annotation software for CAZyme prediciton from amino acid sequences.
PR includes the setup module needed to download models from hugging face.

The bioconda recipe [review](https://github.com/bioconda/bioconda-recipes/pull/63904) process is slow, so for now, using a pip installation (env and containers through Seqera Containers).

This is the CPU version. Will probably create a separate GPU (dramatic speed increase) one after this gets merged.

## PR checklist

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/vcf2zarr/explode/environment.yml
+++ b/modules/nf-core/vcf2zarr/explode/environment.yml
@@ -1,0 +1,7 @@
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/environment-schema.json
+channels:
+  - conda-forge
+  - bioconda
+dependencies:
+  - "bioconda::bio2zarr=0.1.8"
--- a/modules/nf-core/vcf2zarr/explode/main.nf
+++ b/modules/nf-core/vcf2zarr/explode/main.nf
@@ -1,0 +1,40 @@
+process VCF2ZARR_EXPLODE {
+    tag "$meta.id"
+    label 'process_low'
+
+    conda "${moduleDir}/environment.yml"
+        container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/77/7713d869c8c8259c10701c95fc105bad8bcfbd6735de6941a47c9e6e26e9bb2f/data':
+        'community.wave.seqera.io/library/bio2zarr:0.1.8--c2c92dd3f64fb0f9' }"
+
+    input:
+    tuple val(meta), path(vcf)
+
+    output:
+    tuple val(meta), path("*.icf"), emit: icf
+    tuple val("${task.process}"), val('vcf2zarr'), eval('vcf2zarr --version |& sed -n "s/.* //p"'), emit: versions_vcf2zarr, topic: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    vcf2zarr \\
+        explode \\
+        ${args} \\
+        --worker-processes ${task.cpus} \\
+        ${vcf} \\
+        ${prefix}.icf
+    """
+
+    stub:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    echo ${args}
+
+    mkdir ${prefix}.icf
+    """
+}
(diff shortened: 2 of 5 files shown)


vagkaratzas:
`test_level*_embeddings`, seem to change between my local machine and GitHub runners; investigating

vagkaratzas:
> `test_level*_embeddings`, seem to change between my local machine and GitHub runners; investigating

I did some more tests and they seem to stay the same in a system, but change depending on CPU model, probably because of non-determinism in the FAISS index search or the ESM embedding computation, and not due to a container/environment difference.

Likely causes (by the bot -I excluded one that didn't make sense for this case):
```
1. CPU architecture / SIMD instructions — FAISS uses AVX2/AVX-512 on modern CPUs and falls back to SSE4  
  or scalar on older ones. The floating-point operations are reordered differently depending on the SIMD   
  path, producing subtly different embeddings due to floating-point non-associativity. GitHub runners use  
  different CPU generations than your local machine. 

2. FAISS approximate nearest-neighbour — if Level 2 uses an IVF or HNSW index, the search is approximate 
  and the results can differ when the hardware SIMD path changes, even with the same query vectors.        
   
 The fact that it's stable within the same machine across Singularity/Docker/Conda confirms it's not a    
  software version or library issue — the same binary code hits the same SIMD path on the same CPU. But
  cross-platform (your machine → GitHub runner), the CPU capabilities differ, changing the low-level       
  floating-point execution path.
```

modules/nf-core/caalm/caalm/tests/main.nf.test
```diff
@@ -0,0 +1,111 @@
...
+
+        setup {
+            run("CAALM_DOWNLOADMODELS") {
+                script "../../downloadmodels/main.nf"
+            }
+        }
+
+        when {
+            params {
+                module_args = "--save-level0-embeddings --save-level1-embeddings --save-level2-embeddings"
```
jfy133, line 24: How important are these? Are they mandatory flags somehow?

modules/nf-core/caalm/caalm/main.nf
```diff
@@ -0,0 +1,59 @@
...
+    def args = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    caalm \\
+        $args \\
+        --level0-model ${models}/level0 \\
+        --level1-model ${models}/level1 \\
+        --level2-model ${models}/level2/model.pt \\
+        --level2-faiss-dir ${models}/level2/faiss \\
+        --level2-label-tsv-dir ${models}/level2/refdb \\
```
jfy133: Can these levels be skipped somehow? E.g., what if its only level1? What happens if the previous download models step does not include the corresponding flags?

I think these seem quite hardcoded, and if all required should be separate input channels

modules/nf-core/caalm/caalm/main.nf
```diff
@@ -0,0 +1,59 @@
...
+        --output-name ${prefix} \\
+        -o . \\
```
jfy133, line 43: No multi-threading?

vagkaratzas: ofc there is...let me update...good catch, thanks :D

modules/nf-core/caalm/caalm/tests/main.nf.test
vagkaratzas: Not mandatory; it's for people who want to save their produced embeddings (per-level) for the input seqeunces

modules/nf-core/caalm/caalm/main.nf
vagkaratzas: This is how the tool expects the folder structure of the donwloaded model (I think this is also a good way visially to make sure that this will be what the user sees in the downloaded models). It is how it gets downloaded by the setup module (in the same PR).

jfy133: I would emit each one separately then and come in together (can be a single tuple ,rather than one input channel per file though)

vagkaratzas: I see your point, and then make more inputs for caalm/caalm accordingly

vagkaratzas: Done

modules/nf-core/caalm/downloadmodels/main.nf
```diff
@@ -0,0 +1,33 @@
...
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/81/81726ed58923a02e2e947be826509e0f34c3e762183e0bd5b6b109769b67ac84/data':
+        'community.wave.seqera.io/library/huggingface_hub_python:91b414233adf3eb1' }"
+
+    output:
+    path("models/level0"), emit: level0
+    path("models/level1"), emit: level1
+    path("models/level2"), emit: level2
```
jfy133: I would put these all on one tuple, they are all related and can't be used in any other way (e.g. with `.bam` and `.bai`) - that way you don't have to do any `.combine` shenanigans in this case

vagkaratzas: Also agreed. Coming up

PR #11095, opened 2026-03-31
deprecate, switch cat/cat to find/concatenate

### LINKED ISSUE #5516: Unexpected behaviour of cat/cat

### Description of the bug

```
-[ClearNote-Health/Overlord] Pipeline completed with errors-
ERROR ~ Error executing process > 'OVERLORD:CAT_CAT (1)'

Caused by:
  The name of the input file can't be the same as for the output prefix in the module CAT_CAT (currently `xxxxx.1.aaaaa.003.bed.gz`). Please choose a different one. -- Check script 'workflows/../modules/nf-core/cat/cat/main.nf' at line: 43
```

I am attempting to merge several gz BED files with `cat/cat`. For reference here is what my input channel looks like 
```
[DUMP: beds] [['id':'xxxxx.2.aaaaa', 'data_type':'fastq', 'paired_end':true, 'filter':'pass', 'genome':'R64-1-1'], [/Users/kyle/Projects/cnh-overlord/work/0d/4c52846b6116972a997695b3030925/xxxxx.2.aaaaa.004.bed.gz, /Users/kyle/Projects/cnh-overlord/work/ca/d64b77b025a6f9bb672a0341e5ccb4/xxxxx.2.aaaaa.001.bed.gz, /Users/kyle/Projects/cnh-overlord/work/75/029ecfb361c820744709ca420ef9ba/xxxxx.2.aaaaa.002.bed.gz, /Users/kyle/Projects/cnh-overlord/work/c9/00d16b1f8434d476eb2e3461e6146b/xxxxx.2.aaaaa.003.bed.gz, /Users/kyle/Projects/cnh-overlord/work/af/2863c6c4b2d7d6e45dce8a92263ce3/xxxxx.2.aaaaa.000.bed.gz]]
```

As you can see I would expect the output file to be named like `xxxxx.2.aaaaa.bed.gz`. However the `cat_cat` process is failing because it attempts to create an output file called `xxxxx.1.aaaaa.003.bed.gz` which has the same name of an input file (thankful that check exists!).

The cause of this error is the function defined in the module named `getFileSuffix`. This function incorrectly returns the suffix as`.004.bed.gz`, etc. because it attempts to find the suffix based on the first input file. I expect that it would return `.bed.gz`.

The stated intention of the function is _"for .gz files also include the second to last extension if it is present. E.g., .fasta.gz"_

The regex matcher in this function does not look right to me, it contains 3 capture groups (1 outer and 2 inner) and always grabs the outer. Also the regex looks for suffix patterns that are 1 to 5 chars long. This does not match the functions stated intention nor should there be an artificial limit.

I would also like to point out that I think the use of `$prefix` is not ideal in this module. The output of the cat command is directed to `$prefix` though `prefix` should default to the meta.id. I do not think it is good practice to manipulate it outside the context of the `task.ext.prefix`. I think it may be better to direct the output to a file that concatenates the `$prefix` and `$suffix` strings. 

_No response_

### Relevant files

_No response_

### System information

_No response_

vagkaratzas:
## PR checklist

Closes #5516 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/cat/cat/main.nf
+++ b/modules/nf-core/cat/cat/main.nf
@@ -30,6 +30,9 @@ process CAT_CAT {
     """
     assert false: deprecation_message
 
+    def file_list = files_in.collect { file -> file.toString() }
+    prefix   = task.ext.prefix ?: "${meta.id}${getFileSuffix(file_list[0])}"
+
     stub:
     def deprecation_message = """
     WARNING: This module has been deprecated. Please use nf-core/modules/find/concatenate


modules/nf-core/cat/cat/tests/main.nf.test
maxulysse, line 1: Do we need to keep ALL the tests now that we're deprecating this module?
Shouldn't 1 be enough?

vagkaratzas: There is no strict guide, but it makes sense. I will just comment all, but one, similarly to how it's done elsewhere.

vagkaratzas:
> Also, can we actually remove most of the script now it fails anyway?

I would leave `main.nf` as is (+ the deprecation warnings), for legacy/archive purposes.

mashehu:
we have the git history if people want to see the previous state, I guess.

vagkaratzas:
Ok, fair. Most of the deprecated modules do that as well.
I guess the relevant [documentation](https://nf-co.re/docs/guidelines/components/module_deprecation) would require some udpating; seems very lean, regarding what to keep/remove.

PR #11107, opened 2026-04-01
feat(busco): add bbtools memory env var

prototaxites:
See https://nfcore.slack.com/archives/CJRH30T6V/p1775037191383429

@matthdsm and I have made made a patch to the BUSCO Conda recipe:

> BUSCO calls stats.sh from bbtools, but does not provide a way to override the default fixed memory allocation of 120Mb. For larger input genomes, this can cause stats.sh to OOM, leading to crashes.
>
> BUSCO does have an option --skip_bbtools, but if you pass multiple genomes in a single call this flag only gets applied to the first genome: https://gitlab.com/ezlab/busco/-/work_items/841
>
> While we wait for a fix in BUSCO proper, this patch adds an override to allow setting the -Xmx flag to stats.sh directly by setting BUSCO_BBTOOLS_MEMORY, allowing higher fixed memory values to be passed to the subprocess.

This PR implements this new threshold in the module. Hopefully this 1Gb limit is plenty!

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/subworkflows/nf-core/fasta_gxf_busco_plot/main.nf
+++ b/subworkflows/nf-core/fasta_gxf_busco_plot/main.nf
@@ -102,7 +102,7 @@ workflow FASTA_GXF_BUSCO_PLOT {
                                                 : ch_fasta
                                                 | map { meta, fasta -> [ meta.id, meta, fasta ] }
                                                 | combine(
-                                                    ch_gxf.map { meta2, gxf -> [ meta2.id, gxf ] }
+                                                    ch_gxf.map { meta2, gxf -> [ meta2.id, gxf ] }, by: 0
                                                     // Join with matching annotation
                                                     // to allow one annotations per fasta
                                                 )


modules/nf-core/busco/busco/main.nf
```diff
@@ -4,8 +4,8 @@ process BUSCO_BUSCO {
 
     conda "${moduleDir}/environment.yml"
     container "${workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container
-        ? 'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/64/6456c1880785adefb4fc9b480bb7662479d5662c17f70d5e8715b7f2a63ee28b/data'
-        : 'community.wave.seqera.io/library/busco_numpy:b66937518a305dd7'}"
+        ? 'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/c6/c6684133dafc676a4087ca5615907bfdaaba29210f1c8eb2082c4096a009b2cd/data'
```
dialvarezs, line 7: Why the container update?

modules/nf-core/busco/busco/main.nf
```diff
@@ -58,7 +58,10 @@ process BUSCO_BUSCO {
         './*-busco/*/prodigal_output/predicted_genes/tmp/',
     ]
     def clean_cmd = clean_intermediates ? "rm -fr ${intermediate_files.join(' ')}" : ''
+    def bbtools_memory = task.memory.toGiga() > 1 ? "1g" : "120m"
```
dialvarezs: Should 1Gb be enough in all cases? Maybe could be a good idea to make it dynamic based on `task.attempt`. So, if it's not enough, on the retry it will have more.

modules/nf-core/busco/busco/main.nf
prototaxites: Needs to pull in the new Conda build!

modules/nf-core/busco/busco/main.nf
prototaxites: I have no idea what is enough, honestly... Could even just do `task.memory.toGiga() - 2`, with a fallback of the standard value?

prototaxites: Does that look more sensible? Just use most of the memory, but leave a 4Gb buffer.

dialvarezs: Maybe that could be a little overkill. By default medium has 36.GB, so it will be 34.GB.
What about `max(task.memory.toGiga() * 0.25, 120)` or something like that?

prototaxites: Oh, I like that

prototaxites: Try that? Trying to keep the units correctly formatted

modules/nf-core/busco/busco/main.nf
```diff
@@ -58,7 +58,11 @@ process BUSCO_BUSCO {
         './*-busco/*/prodigal_output/predicted_genes/tmp/',
     ]
     def clean_cmd = clean_intermediates ? "rm -fr ${intermediate_files.join(' ')}" : ''
+    def bbtools_max_memory = (task.memory * 0.25).toGiga())
```
dialvarezs:
```suggestion
    def bbtools_max_memory = (task.memory * 0.25).toGiga()
```

PR #11108, opened 2026-04-01
Add metamap for input/output in stringtie/merge

### LINKED ISSUE #8686: update module: stringtie/merge

Odulhin:
## Summary
- Add a metamap  for input, output instead of the previous harcoded ones
- Modify related subworkflow : bam_stringtie_merge
- Update snapshots 

Closes #8686


modules/nf-core/stringtie/merge/meta.yml
```diff
@@ -14,25 +14,33 @@ tools:
...
-      ontologies: []
+  - - meta:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. [ id:'test', single_end:false ]
+    - bam:
+        type: file
+        description: |
+          Stringtie transcript gtf output(s).
```
dialvarezs: Could you fix the description here?
And add pattern too.

modules/nf-core/stringtie/merge/meta.yml
```diff
@@ -14,25 +14,33 @@ tools:
...
+        ontologies: []
```
dialvarezs: Filling ontologies would be nice too

modules/nf-core/stringtie/merge/main.nf
```diff
@@ -8,30 +9,33 @@ process STRINGTIE_MERGE {
         'biocontainers/stringtie:2.2.1--hecb563c_2' }"
 
     input:
-    path stringtie_gtf
+    tuple val(meta), path(bam)
     path annotation_gtf
```
famosab: Can we put all these inputs into one tuple? That will make sure you're sure that EVERY time everything comes together in the right combination.

modules/nf-core/stringtie/merge/main.nf
famosab, line 1: You can run
```bash
```

modules/nf-core/stringtie/merge/main.nf
```diff
@@ -8,30 +9,33 @@ process STRINGTIE_MERGE {
...
+    def prefix    = task.ext.prefix ?: "${meta.id}"
     def reference = annotation_gtf ? "-G $annotation_gtf" : ""
     """
     stringtie \\
-        --merge $stringtie_gtf \\
+        --merge \\
+        $bam \\
         $reference \\
-        -o stringtie.merged.gtf \\
+        -o ${prefix}.merged.gtf \\
```
famosab: We want to let the module user decide about the output file naming in their pipeline via `ext-prefix`. So ideally the output filenames are just `${prefix}`.
```suggestion
        -o ${prefix}.gtf \\
```

modules/nf-core/stringtie/merge/main.nf
```diff
@@ -8,30 +9,33 @@ process STRINGTIE_MERGE {
         'biocontainers/stringtie:2.2.1--hecb563c_2' }"
 
     input:
-    path stringtie_gtf
+    tuple val(meta), path(bam)
     path annotation_gtf
 
     output:
-    path "stringtie.merged.gtf", emit: gtf
+    tuple val(meta), path("*.merged.gtf"), emit: gtf
```
famosab:
```suggestion
    tuple val(meta), path("${prefix}.gtf"), emit: gtf
```

modules/nf-core/stringtie/merge/main.nf
```diff
@@ -8,30 +9,33 @@ process STRINGTIE_MERGE {
...
     tuple val("${task.process}"), val('stringtie'), eval('stringtie --version'), emit: versions_stringtie, topic: versions
+
     when:
     task.ext.when == null || task.ext.when
 
     script:
     def args      = task.ext.args ?: ''
+    def prefix    = task.ext.prefix ?: "${meta.id}"
```
famosab:
```suggestion
    prefix    = task.ext.prefix ?: "${meta.id}"
```

modules/nf-core/stringtie/merge/main.nf
```diff
@@ -8,30 +9,33 @@ process STRINGTIE_MERGE {
...
         $reference \\
-        -o stringtie.merged.gtf \\
+        -o ${prefix}.merged.gtf \\
+        -p $task.cpus \\
         $args
-
     """
 
     stub:
+    def prefix = task.ext.prefix ?: "${meta.id}"
```
famosab:
```suggestion
    prefix = task.ext.prefix ?: "${meta.id}"
```

modules/nf-core/stringtie/merge/meta.yml
Odulhin: The description is correct I didn't realise that the variable name was bam instead of gtf. Will make the change and add the pattern + ontologie.

Thanks for your feedback !

modules/nf-core/stringtie/merge/tests/main.nf.test
```diff
@@ -29,7 +29,7 @@ nextflow_process {
         when {
             process {
                 """
-                input[0] = STRINGTIE_STRINGTIE.out.transcript_gtf.map { it -> it[1] }
+                input[0] = STRINGTIE_STRINGTIE.out.transcript_gtf.map { meta, gtf -> [meta, gtf] }
```
lindenb: @Odulhin useless `map{}` (same input/output)

modules/nf-core/stringtie/merge/tests/main.nf.test
```diff
@@ -49,7 +49,7 @@ nextflow_process {
...
```
lindenb: @Odulhin same here: useless map{} (same input/output)

modules/nf-core/stringtie/merge/main.nf
Odulhin: Thanks !

modules/nf-core/stringtie/merge/main.nf
Odulhin: I tried but I had issues to pass tests afterwards.

Instead I put the annotation_gtf inside a metamap which was easier and required less code modifications because in the related subworkflows it was already in this form only the metadata was discarded when calling the module.

modules/nf-core/stringtie/merge/main.nf
```diff
@@ -1,37 +1,41 @@
...
-        'biocontainers/stringtie:2.2.1--hecb563c_2' }"
+    container "${workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container
+        ? 'https://depot.galaxyproject.org/singularity/stringtie:2.2.1--hecb563c_2'
+        : 'biocontainers/stringtie:2.2.1--hecb563c_2'}"
 
     input:
-    path stringtie_gtf
-    path annotation_gtf
+    tuple val(meta), path(gtf)
+    tuple val(meta2), path(annotation_gtf)
```
famosab, line 13: Can we put all these inputs into one tuple? That will make sure you're sure that EVERY time everything comes together in the right combination.

Odulhin: I've made the necessary changes.

modules/nf-core/stringtie/merge/tests/main.nf.test
```diff
@@ -29,16 +29,17 @@ nextflow_process {
...
                 """
             }
         }
 
         then {
             assertAll(
             { assert process.success },
-            { assert snapshot(process.out.gtf).match("fs_gtf") },
+            { assert snapshot(process.out.merged_gtf).match("fs_gtf") },
             { assert snapshot(process.out.findAll { key, val -> key.startsWith('versions') }).match("fs_versions") }
```
famosab:
```suggestion
            { assert snapshot(
                process.out.merged_gtf,
                process.out.findAll { key, val -> key.startsWith('versions') }
                ).match() }
```

We only want one snapshot ideally :)

modules/nf-core/stringtie/merge/main.nf
```diff
@@ -1,37 +1,40 @@
...
-        'https://depot.galaxyproject.org/singularity/stringtie:2.2.1--hecb563c_2' :
-        'biocontainers/stringtie:2.2.1--hecb563c_2' }"
+    container "${workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container
+        ? 'https://depot.galaxyproject.org/singularity/stringtie:2.2.1--hecb563c_2'
+        : 'biocontainers/stringtie:2.2.1--hecb563c_2'}"
 
     input:
-    path stringtie_gtf
-    path annotation_gtf
+    tuple val(meta), path(gtf), val(meta2), path(annotation_gtf)
```
famosab:
```suggestion
    tuple val(meta), path(gtf), path(annotation_gtf)
```

modules/nf-core/stringtie/merge/meta.yml
```diff
@@ -14,25 +14,32 @@ tools:
...
+          Stringtie transcript gtf output(s) to be merged together.
+        pattern: "*.{gtf|gff}"
+        ontologies:
+          - edam: "http://edamontology.org/format_2306"
+          - edam: "http://edamontology.org/format_1975"
+    - meta2:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. [ id:'test', single_end:false ]
```
famosab:
```suggestion
```

modules/nf-core/stringtie/merge/main.nf
famosab: What I forgot to mention is: you can keep the two file in different input channels if you keep one file for exampl constant but to me it seemed that the module could be used with different gtf and annotation_gtf files where I would want to exchange them on a sample basis, if that is not correct then leave them separate :)

Odulhin: Ok I understand better. I think its best to keep them separate because the goal is to merge all the newly created gtf which contains the _de novo_ assembly together with the annotation_gtf which was used to make them in order to ouput a new annotation gtf which contain the novel transcripts that now can be quantified instead of being discarded.

famosab: perfect then the way it was makes complete sense sorry for not pointing this out earlier :)

modules/nf-core/stringtie/merge/tests/main.nf.test
```diff
@@ -49,16 +51,16 @@ nextflow_process {
...
                 """
             }
         }
 
         then {
             assertAll(
             { assert process.success },
-            { assert snapshot(process.out.gtf).match("rs_gtf") },
+            { assert snapshot(process.out.merged_gtf).match("rs_gtf") },
             { assert snapshot(process.out.findAll { key, val -> key.startsWith('versions') }).match("rs_versions") }
```
famosab: can you do the same change here as above please

also

You can use
```
```
to clean up the snapshot using [nft-utils](https://github.com/nf-core/nft-utils). It is used to clean process and workflow outputs by removing the numbered keys. This will create snapshots that are more easy to read by humans.

famosab: After this I think we are good to go :)

Odulhin: Thank you very much for your feedback !

subworkflows/nf-core/bam_stringtie_merge/main.nf
```diff
@@ -1,33 +1,30 @@
...
-        ch_chrgtf.map { _meta, gtf -> [ gtf ] }
+        ch_chrgtf
     )
-    ch_stringtie_gtfs = STRINGTIE_MERGE.out.gtf
+    ch_stringtie_gtfs = STRINGTIE_MERGE.out.merged_gtf
 
     emit:
     stringtie_gtf = ch_stringtie_gtfs // channel: [ meta, gtf ]
-    versions      = ch_versions                     // channel: [ path(versions.yml) ]
+    versions      = ch_versions // channel: [ path(versions.yml) ]
```
famosab: I think ch_versions is not populated anymore right? Maybe we can remove it completely from the subworkflow then?

subworkflows/nf-core/bam_stringtie_merge/main.nf
```diff
@@ -1,33 +1,30 @@
...
 workflow BAM_STRINGTIE_MERGE {
-
     take:
     bam_sorted // channel: [ meta, bam ]
-    ch_chrgtf  // channel: [ meta, gtf ]
+    ch_chrgtf // channel: [ meta, gtf ]
 
     main:
-    ch_versions       = channel.empty()
+    ch_versions = channel.empty()
```
famosab:
```suggestion
```

subworkflows/nf-core/bam_stringtie_merge/main.nf
famosab:
```suggestion
```

modules/nf-core/stringtie/merge/tests/main.nf.test
```diff
@@ -29,17 +29,20 @@ nextflow_process {
...
         then {
             assertAll(
             { assert process.success },
-            { assert snapshot(process.out.gtf).match("fs_gtf") },
-            { assert snapshot(process.out.findAll { key, val -> key.startsWith('versions') }).match("fs_versions") }
+            { assert snapshot(
+                process.out.merged_gtf,
+                process.out.findAll { key, val -> key.startsWith('versions') },
+                sanitizeOutput(process.out)
+                ).match() }
```
famosab, line 43:
```suggestion
            { assert snapshot(
                sanitizeOutput(process.out)
                ).match() }
```
this suffices (captures the whole output)

modules/nf-core/stringtie/merge/tests/main.nf.test
```diff
@@ -49,17 +52,20 @@ nextflow_process {
...
         then {
             assertAll(
             { assert process.success },
-            { assert snapshot(process.out.gtf).match("rs_gtf") },
-            { assert snapshot(process.out.findAll { key, val -> key.startsWith('versions') }).match("rs_versions") }
+            { assert snapshot(
+                process.out.merged_gtf,
+                process.out.findAll { key, val -> key.startsWith('versions') },
+                sanitizeOutput(process.out)
+                ).match() },
```
famosab:
```suggestion
            { assert snapshot(
                sanitizeOutput(process.out)
                ).match() }
```

subworkflows/nf-core/bam_stringtie_merge/main.nf
Odulhin: You're right it is not used anymore

PR #11120, opened 2026-04-05
Add module: modkit/localize

sahuno:
### Tool
[modkit](https://github.com/nanoporetech/modkit) — A bioinformatics tool for working with modified bases from Oxford Nanopore sequencing data.

### What this module does
`MODKIT_LOCALIZE` aggregates bedMethyl pileup counts around genomic features of interest (e.g. CpG islands, gene bodies, repeat elements), producing:
- A TSV of percent-modification vs. offset from feature midpoints
- An optional interactive HTML chart

This enables investigation of modification enrichment or depletion patterns relative to genomic landmarks, analogous to a methylation profile around features.

### Inputs
- bgzip-compressed + tabix-indexed bedMethyl file (from `modkit pileup`)
- Genome sizes file (e.g. `.fai` or `.sizes`)
- BED file of regions to localize around

### Outputs
- `*.tsv` — percent-modification vs. offset table
- `*.html` — interactive chart
- `*.log` — debug log

### Tests
Tests chain `MODKIT_PILEUP` → `TABIX_TABIX` → `MODKIT_LOCALIZE` using existing nf-core test-datasets (`homo_sapiens` nanopore BAM + `genome.fasta.fai`). Regions BED is created inline — no new test data upload required.

- [x] `nf-test` passes locally (singularity profile)
- [x] `nf-test` passes on SLURM compute node (job 17362918, 2 tests PASSED in 13s)
- [x] `nf-core modules lint` passes: 48/48 tests, 0 warnings, 0 failures

### Notes
- `--regions` is required by the modkit 0.6.1 CLI (enforced at the tool level)
- Container spec matches the existing `modkit/pileup` module
- A companion `modkit/localize/plot` submodule for composite multi-sample visualization is planned as a follow-up PR


famosab:
We usually recommend to have one module per PR. That makes the review process easier and it is more likely that someone will review your PR. I would suggest to split this PR up into the dorado module and the modkit module :)

sahuno:
@famosab you were right, and sorry this took so long. Split done — this PR is
now `modkit/localize` only, six files, nothing else.

- `dorado/basecaller` is gone from here entirely. It is already #11122, and the
  copy in this branch was the older revision that still committed pod5 files
  into the repo.
- `modkit/localize/plot` moves to its own module, renamed `modkit/localizeplot`
  (three-level paths fail `main_nf_module_granularity`). I have not opened that
  PR yet — see the note at the bottom.

Rebasing onto current master surfaced two things that were quietly broken:

**`TABIX_TABIX` is deprecated.** Its script block on master is now
`assert false` with a message pointing at `HTSLIB/BGZIPTABIX`, and its input
signature had grown from two elements to four. The test setup now uses
`HTSLIB_BGZIPTABIX` to build the tabix index from the `MODKIT_PILEUP` output.

**The test config was never loaded.** `main.nf.test` had no `config` directive,
so `params.module_args` never reached `ext.args` and the
`--min-coverage 1 --window 1000` values the test depends on were being silently
ignored. Added the directive.

Also aligned the container declaration with the merged modkit modules
(`workflow.containerEngine in ['singularity', 'apptainer']`, `quay.io/` prefix).

nf-core/tools 4.1.0 lint: **59 passed, 0 warnings, 0 failed**. Both nf-tests
pass under the singularity profile.

On `modkit/localizeplot`: it is ready except for its container, which currently
points at a temporary Wave build URI that is neither on the allowed registry
list nor stable. It needs a frozen `community.wave.seqera.io/library/...` image
built from its environment.yml. I would rather sort that out before opening the
PR than add another red one to the queue — but if you would prefer to see it now
regardless, say the word and I will open it.

sahuno:
Update since the split: rebased onto master, tests moved to `assert process.success` + `snapshot(sanitizeOutput(process.out))` (same pattern as #11286), snapshots regenerated, container bumped to `ont-modkit` 0.6.4. This PR is `modkit/localize` only — six files — CI is green and there are no open review threads. @famosab re-requesting review since you flagged the original bundle.

modules/nf-core/modkit/localize/main.nf
famosab, line 1: You can run
```bash
```

modules/nf-core/modkit/localize/meta.yml
```diff
@@ -0,0 +1,117 @@
...
+          Groovy Map containing regions BED file information
+          e.g. `[ id:'CpG_islands' ]`
+    - regions:
+        type: file
+        description: |
+          BED file of regions to calculate enrichment around. These serve
+          as the points from which the window number of bases is centered
+          (e.g. CpG islands, gene bodies, repeat elements).
+        pattern: "*.bed"
+        ontologies: []
```
famosab, line 56: Can you look at other meta.yml files and add ontologies to all the files where it is applicable

modules/nf-core/modkit/localize/main.nf
```diff
@@ -0,0 +1,52 @@
...
+    input:
+    tuple val(meta), path(bedmethyl), path(tbi)
+    tuple val(meta2), path(sizes)
+    tuple val(meta3), path(regions)
+
+    output:
+    tuple val(meta), path("*.tsv")  , emit: tsv  , optional: true
+    tuple val(meta), path("*.html") , emit: chart , optional: true
+    tuple val(meta), path("*.log")  , emit: log   , optional: true
+    tuple val("${task.process}"), val('modkit'), eval("modkit --version | sed 's/modkit //'"), emit: versions_modkit, topic: versions
```
famosab, line 19:
```suggestion
    tuple val(meta), path("*.tsv"), emit: tsv, optional: true
    tuple val(meta), path("*.html"), emit: chart, optional: true
    tuple val(meta), path("*.log"), emit: log, optional: true
```

PR #11128, opened 2026-04-07
Whatshap/phase: add pedigree input

tetedange13:
`whatshap/phase` can use a pedigree file to improve phasing

Relevant documentation : https://whatshap.readthedocs.io/en/latest/guide.html#phasing-pedigrees
<br>

## PR checklist

  - For modules:
    - [ ] `nf-core modules test <MODULE> --profile docker`  NO DOCKER AVAILABLE


famosab (approved):
How does the pedigree file not change anything in the output ❓ 
But i think this is good to go :)

tetedange13:
> How does the pedigree file not change anything in the output

I *think* (could not assert that from the code) that `whatshap` ignores pedigree when only 1 sample involved (which is the case in these tests)

I observed consistent behaviour locally with bigger dataset (100K variants)

famosab:
Can you then add a test with multiple samples where the pedigree is not ignored? :)

tetedange13:
@famosab I can add such test, but after inspection I do not have all required data on `nf-core/test-datatasets`

Will have to add them first -> move current PR status to "draft"

tetedange13:
Added new test with trio data, declined in 2 sub-tests where PED is provided or not
=> **They get different snapshots -> OK**

Achieved to only use available `test-datasets`, but part of them are from the `raredisease` branch, is it OK ?
=> Also do you think I should use "permalink" as URL for these data taken from `raredisease` branch ? (in case they are removed from `HEAD`)

famosab:
> Achieved to only use available test-datasets, but part of them are from the raredisease branch, is it OK ?
=> Also do you think I should use "permalink" as URL for these data taken from raredisease branch ? (in case they are removed from HEAD)

I think we prefer to then mirror them on the modules branch :)

tetedange13:
> I think we prefer to then mirror them on the modules branch :)

PR to mirror required data on `modules` branch : https://github.com/nf-core/test-datasets/pull/2021
=> **Most files are larger than Githubs's maximum**

Also added a `no_chr` version of `chr21/genome.fasta`

famosab:
> => Most files are larger than Githubs's maximum

I do not understand your comment - do you now have all files needed for your tests in the test-datasets repository or not? 🤔

tetedange13:
> do you now have all files needed for your tests in the test-datasets repository or not?

Yes, it was just a warning git raised after I pushed

But I simply downloaded required files from `raredisease` branch and copied them to `modules` branch

Maybe there is a smarter way to "mirror" data ?

modules/nf-core/whatshap/phase/meta.yml
```diff
@@ -41,6 +41,10 @@ input:
         type: file
         description: BAM index file (optional but recommended)
         pattern: "*.bai"
+    - pedigree:
+        type: file
+        description: Pedigree file (optional)
```
famosab: small nitpick maybe expand the description to actually hold information :) (when would one want to use this file for example)

famosab (approved):
I think you can merge this then 🚀

modules/nf-core/whatshap/phase/meta.yml
tetedange13: Adressed by commit d11850f

tetedange13:
Replaced need for "genome_no_chr.fasta" by a `seqkit/replace` setup

And use subsetted versions of BAMs : https://github.com/nf-core/test-datasets/pull/2048

tetedange13:
@famosab any last comment ?

I re-ran tests and they pass !

modules/nf-core/whatshap/phase/tests/main.nf.test
```diff
@@ -60,6 +62,148 @@ nextflow_process {
...
+                    ).match() }
+            )
+        }
+    }
+
+    test("whatshap - phase - trio vcf, corresponding bams, no pedigree") {
+
+        config "./nextflow.config"
+
+        setup {
```
famosab, line 140: If the setup blocks have the same setup you can just have them once at the beginning of the test file and only run them once per test run.

famosab: so just move the setup block once above all tests in the main.nf.test file

tetedange13: Hum I tried to define `SEQKIT_REPLACE + SAMTOOLS_FAIDX2` as global `setup`, but it requires to modify legacy test `whatshap - phase - vcf`, which is not very good IMO

You have to add it a `param` section, in order to provide dummy parameters for `SEQKIT_REPLACE` step (otherwise error)

Even if this step is not used by this specific test...

tetedange13: > so just move the setup block once above all tests in the main.nf.test file

Does not work either, still `whatshap - phase - vcf` not passing anymore

Do you want me to commit even so ? So that you can look for a workaround ?

famosab: no then we just leave it as is now I would say 👍

famosab:
I think you can go ahead and merge 🥳

PR #11141, opened 2026-04-09
Fix tximport gene-level crash with mismatched transcript FASTA/GTF

pinin4fjords:
## Problem

When users provide a `--transcript_fasta` that contains transcripts not present in the GTF, `tximport`'s `read_transcript_info()` function adds these unmatched transcripts as "extras" with `gene_id = transcript_id` (i.e., the transcript ID is used as a fake gene ID). After `summarizeToGene()`, these fake gene entries end up in the gene-level count matrices alongside real genes.

The downstream `SummarizedExperiment` process then receives these gene count matrices (containing a mix of real ENSG gene IDs and fake ENST "gene" IDs) along with the original `tx2gene.tsv` as row metadata. `findColumnWithAllEntries()` tries to find a single column in the tx2gene that contains ALL gene IDs from the count matrix, but no column matches: the `gene_id` column has only real gene IDs, and the `transcript_id` column has only real transcript IDs. This produces the error:

```
Error in findColumnWithAllEntries(ids, metadata) :
  No column contains all vector entries ENSG00000000003.14, ...
```

This commonly happens when users provide `--transcript_fasta` from a different source or version than their `--gtf`.

## Fix

1. **Warning**: When unmatched transcripts are detected, emit a warning with the count and first 5 IDs, so users understand the FASTA/GTF inconsistency.

2. **Filter fake genes**: After `summarizeToGene()`, remove the fake gene entries (where gene_id was set to the transcript ID) from the gene-level output. The extras are still needed as input to `summarizeToGene()` (it errors if transcripts are missing from the tx2gene), but their resulting fake gene rows are stripped before writing output files.

Existing behavior is unchanged when FASTA and GTF are consistent (no extras, no filtering).

Fixes nf-core/rnaseq#1773

## Test plan

- [x] All 8 existing tximport tests pass unchanged
- [x] New `mismatched_transcripts` test: truncates tx2gene to simulate FASTA/GTF mismatch, asserts:
  - Process succeeds (previously would crash downstream)
  - Gene count rows < transcript count rows (proves fake genes were filtered)
  - Warning message appears in `.command.err`
- [x] Verified the mismatch test **fails** against unfixed code (gene count == transcript count, no warning)


modules/nf-core/tximeta/tximport/tests/main.nf.test.snap
```diff
@@ -138,11 +138,11 @@
                 ]
```
jfy133, line 138: Unlss I'm being blined, I don't see the new test represented in this snapshot?

pinin4fjords: The new test doesn't use snapshots- it's asserting that e.g. we get the warning message we expect etc.

jfy133 (approved):

PR #11177, opened 2026-04-14
feat: switch ribodetector to GPU-capable container

pinin4fjords:
## Summary

- Replace ribodetector container with a GPU-capable build that includes `pytorch-cuda=12.4`
- The same container works for both CPU and GPU modes: the CPU binary (`ribodetector_cpu`) uses ONNX runtime (unaffected by pytorch variant), while the GPU binary (`ribodetector`) uses PyTorch with CUDA
- Fix a latent GString interpolation bug in the memory flag (`$task.memory.toGiga()` -> `${task.memory.toGiga()}`) that only manifests when the GPU path is taken
- Add GPU-tagged tests (real + stub) following the parabricks test pattern

## Container size

The GPU container is **~3.7 GB** vs **~546 MB** for the current CPU container (~7x increase). This is because CUDA pytorch requires the full CUDA toolkit, cuDNN, and GPU-specific pytorch builds. With the single-container approach, ALL ribodetector users (including CPU-only) would download this larger image. See #11178 for a dual-container alternative that avoids this cost for CPU users.

## Note on conda channels

The environment.yml adds the `pytorch` and `nvidia` conda channels, which are non-standard for nf-core modules. This follows the [official PyTorch installation instructions](https://pytorch.org/get-started/locally/) (`conda install pytorch pytorch-cuda=12.4 -c pytorch -c nvidia`). Without these channels, conda resolves pytorch to the CPU-only build from conda-forge. The only other nf-core module with GPU deps (stardist) uses pip-installed nvidia packages instead. Open to feedback on the preferred approach here.

## Alternative

See #11178 for a dual-container approach that keeps the smaller CPU container (~546 MB) for non-GPU users and only serves the CUDA container (~3.7 GB) when `task.accelerator` is set.

## Rationale

The ribodetector module already selects between GPU (`ribodetector`) and CPU (`ribodetector_cpu`) binaries based on `task.accelerator`, but the existing container only includes CPU pytorch. This PR uses a single GPU-capable container for all users for simplicity.

Related pipeline PR: https://github.com/nf-core/rnaseq/pull/1790

## Test plan

- [ ] CPU tests pass with the new container
- [ ] GPU tests run ribodetector on NVIDIA GPU hardware
- [ ] Stub tests pass for both modes

### DIFF
--- a/modules/nf-core/ribodetector/environment.yml
+++ b/modules/nf-core/ribodetector/environment.yml
@@ -2,8 +2,10 @@
 # yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/environment-schema.json
 channels:
   - pytorch
+  - nvidia
   - conda-forge
   - bioconda
 dependencies:
   - "bioconda::ribodetector=0.3.3"
+  - "pytorch::pytorch"
   - "pytorch::pytorch-cuda=12.4"
--- a/modules/nf-core/ribodetector/environment.yml
+++ b/modules/nf-core/ribodetector/environment.yml
@@ -1,7 +1,9 @@
 ---
 # yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/environment-schema.json
 channels:
+  - pytorch
   - conda-forge
   - bioconda
 dependencies:
   - "bioconda::ribodetector=0.3.3"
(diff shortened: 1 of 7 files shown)


pinin4fjords:
Closing in favor of the dual-container approach in #11178, which keeps the smaller CPU container (~546 MB) for non-GPU users.

PR #11178, opened 2026-04-14
feat: add GPU support to ribodetector (dual-container)

pinin4fjords:
## Summary

- Add GPU container switching to ribodetector: when `task.accelerator` is set, uses a CUDA-enabled container with `pytorch-gpu`; otherwise uses the existing CPU container
- CPU users are unaffected (same ~546 MB container), GPU users get a ~3.5 GB container with CUDA
- Both `conda` and `container` directives use ternaries based on `task.accelerator`
- Fix a latent GString interpolation bug in the memory flag (`$task.memory.toGiga()` -> `${task.memory.toGiga()}`)
- Fix stub gzip syntax for nf-core lint compliance
- Add GPU-tagged tests (real + stub) following the parabricks test pattern
- Improve both CPU and GPU tests: use `nft-fastq` plugin for order-independent sorted read name comparison, log content assertions for classification counts (4159 non-rRNA, 0 rRNA), ANSI color stripping

## Container size

The GPU container is **~3.5 GB** vs **~546 MB** for the CPU container. The size comes from pytorch-gpu bundling the CUDA toolkit, cuDNN, and GPU-specific pytorch builds. With the dual-container approach, CPU-only users are unaffected.

## GPU environment

The `environment.gpu.yml` uses `conda-forge::pytorch-gpu=1.11.0`. The CPU `environment.yml` is unchanged.

## ARM limitation

The GPU container is **x86_64 only**. This is a conda packaging limitation, not a hardware one (ARM + NVIDIA GPU setups like Jetson do exist):

- All conda-forge ARM `pytorch-gpu` builds (oldest available is 2.1.2) depend on the [`__cuda` virtual package](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-virtual.html), which requires CUDA drivers to be present at conda solve time. Wave's micromamba base image does not provide this, so the solve fails.
- The x86 `pytorch-gpu=1.11.0` works because it predates the `__cuda` convention. It depends on [`cudatoolkit`](https://anaconda.org/conda-forge/cudatoolkit) instead, which is a regular installable package that gets bundled into the container.
- Newer x86 builds (2.5.1+) also require `__cuda` and would fail the same way. The version pin to 1.11.0 is not just for reproducibility - it's the newest x86 version that avoids the `__cuda` gate.
- The [pytorch channel](https://anaconda.org/pytorch/pytorch) packages avoid `__cuda` entirely (they use `pytorch-cuda` and `pytorch-mutex` real packages instead), but the pytorch channel has no ARM builds at all.
- We tested `CONDA_OVERRIDE_CUDA=12.4` (which tells conda to pretend CUDA is present) but Wave does not support injecting environment variables at solve time, only at container runtime (`--config-env`).

Pipelines using this module should guard GPU ribodetector on ARM, as nf-core/rnaseq does:
```groovy
if (params.use_gpu_ribodetector && (params.arm ?: false)) {
    error("--use_gpu_ribodetector is not supported on ARM architecture.")
}
```

**Resolution paths:**
1. Wave adds support for `CONDA_OVERRIDE_CUDA` at build/solve time
2. Wave provides CUDA-aware base images
3. conda-forge drops the `__cuda` requirement for ARM pytorch-gpu (unlikely, it serves a purpose)

## Rationale

The ribodetector module already selects between GPU (`ribodetector`) and CPU (`ribodetector_cpu`) binaries based on `task.accelerator`, but the existing container only includes CPU pytorch, making the GPU binary fail with "CUDA unavailable". This PR adds the missing GPU container and wires up the selection.

Related pipeline PR: https://github.com/nf-core/rnaseq/pull/1790

## Test plan

- [x] CPU tests pass (conda + docker, verified locally and in CI)
- [x] nf-core lint passes
- [x] GPU stub tests pass (docker + singularity)
- [x] GPU real tests run ribodetector on NVIDIA GPU hardware with correct classification counts
- [x] GPU and CPU produce identical classification results (same read counts: 4159 non-rRNA, 0 rRNA)

### DIFF
--- a/.github/workflows/nf-test-gpu.yml
+++ b/.github/workflows/nf-test-gpu.yml
@@ -81,7 +81,7 @@ jobs:
         with:
           max_shards: 2
           paths: "${{ join(fromJson(steps.list.outputs.components), ' ') }}"
-          tags: "gpu"
+          tags: "gpu,gpu_highmem"
       - name: debug
         run: |
           echo ${{ steps.list.outputs.components }}
--- a/modules/nf-core/ribodetector/environment.gpu.yml
+++ b/modules/nf-core/ribodetector/environment.gpu.yml
@@ -1,0 +1,8 @@
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/environment-schema.json
+channels:
+  - conda-forge
+  - bioconda
+dependencies:
+  - "bioconda::ribodetector=0.3.3"
+  - "conda-forge::pytorch-gpu=1.11.0"
--- a/modules/nf-core/ribodetector/main.nf
+++ b/modules/nf-core/ribodetector/main.nf
@@ -2,10 +2,10 @@ process RIBODETECTOR {
 	tag "$meta.id"
 	label 'process_medium'
 
-	conda "${moduleDir}/environment.yml"
+	conda "${ task.accelerator ? "${moduleDir}/environment.gpu.yml" : "${moduleDir}/environment.yml" }"
 	container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
-        'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/46/463b8ad941e7f1f2decef20844d666c1c8ac233e166d2bc766164c4a93905a3c/data' :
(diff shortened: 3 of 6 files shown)


jfy133 (approved):

pinin4fjords:
Thanks @jfy133 !

PR #6350, opened 2024-08-26
add msisensorpro/msitumoronly module

bounlu:
This module enables tumor-only analysis for msisenspro (msisensor-pro pro)

Issue #6007, opened 2024-07-22
new module: MSISENSORPRO/MSISENSORPRO_SOMATIC_TUMORONLY


bounlu:
### Is there an existing module for this?

- [X] I have searched for the existing module

### Is there an open PR for this?

- [X] I have searched for existing PRs

### Is there an open issue for this?

- [X] I have searched for existing issues

### Are you going to work on this?

- [ ] If I'm planning to work on this module, I added myself to the `Assignees` to facilitate tracking who is working on the module

vagkaratzas:
Found a closed PR ([#6350](https://github.com/nf-core/modules/pull/6350)) with the attempt to create this module. Is anyone planning to re-open/finalise/push this?

PR #11190, opened 2026-04-14
feat: add msisensorpro/baseline module; clarify pro -d dual semantics

### LINKED ISSUE #11188: [BUG] msisensorpro/pro silently runs in hard-threshold fallback mode; baseline module missing
The `msisensorpro/pro` module passes a scan list to `pro -d` and does not expose `-i`. Per [`ProUsage()` in `cpp/distribution.cpp` L199-L206](https://github.com/xjtu-omics/msisensor-pro/blob/2063de1588cf65dd310a7b95ba5efc6f64c7515a/cpp/distribution.cpp#L199-L206), `-d` has two modes:

```
1. msisensor-pro pro -d reference.list          -i 0.1 -t tumor.bam -o out
2. msisensor-pro pro -d reference.list_baseline        -t tumor.bam -o out
```

The [wiki Best-Practices](https://github.com/xjtu-omics/msisensor-pro/wiki/Best-Practices) and the maintainer (xjtu-omics/msisensor-pro#77) recommend mode 2. The module silently runs mode 1 at the default threshold (`0.1`).

**Requested changes:**
1. Add `msisensorpro/baseline` module (missing piece between `scan` and `pro`).
2. Expose `-i` in `msisensorpro/pro` via `ext.args`.
3. Update `meta.yml` to document that `-d` accepts either a scan list or a baseline.

Happy to PR. Overlaps with #6007 and closed #6350.

nh13:
Closes #11188, overlaps with #6007 (and closed #6350).

## What

1. **New module `msisensorpro/baseline`** wrapping `msisensor-pro baseline`. Builds a trained baseline microsatellite file from a panel of normal samples, for use with `msisensor-pro pro -d` in tumor-only MSI calling. The configure file consumed by `baseline -i` is built internally from the staged `*_all` files so callers can feed a collected channel of `msisensorpro/pro` outputs directly (rather than hand-writing a configure file that references Nextflow work paths).
2. **`msisensorpro/pro/meta.yml`**: expand the `-d` input description to document both supported modes (raw scan list with hard threshold via `-i`, vs trained baseline). No behavior change; `pro`'s `main.nf` already accepts either file type and `-i` already flows through `$args`.

## Why

`msisensor-pro pro -d` accepts both a scan list and a baseline file per [`ProUsage()` in `cpp/distribution.cpp` L199-L206](https://github.com/xjtu-omics/msisensor-pro/blob/2063de1588cf65dd310a7b95ba5efc6f64c7515a/cpp/distribution.cpp#L199-L206). The [wiki Best-Practices](https://github.com/xjtu-omics/msisensor-pro/wiki/Best-Practices) documents only the baseline workflow for tumor-only, and the maintainer recommends it in xjtu-omics/msisensor-pro#77. Until now the baseline subcommand had no module, so pipelines couldn't produce a baseline within nf-core/modules and defaulted to the scan-list-with-hard-threshold fallback mode.

## TODOs before ready for review

- [ ] Verify the live test under `-profile docker` against real fixture data. The chr21 test fasta may not contain enough microsatellites for `baseline` to produce a usable output. If so, either locate richer fixtures or drop to stub-only testing and open a follow-up for fixtures.
- [ ] `nf-core modules test msisensorpro/baseline --profile {docker,singularity,conda}`
- [ ] Snapshot (`tests/main.nf.test.snap`) once live test passes
- [ ] Migrate `msisensorpro/pro`'s test to chain through the new baseline module so CI exercises mode 2 (recommended mode)
- [ ] Maintainer input on whether to rename `pro`'s second input from `list` to something like `microsat_file` (would be a breaking doc-only change)

Happy to iterate on any of the above. Specifically wanted to surface the design of the baseline module's input shape (internally-constructed configure file) before doing more work.

## PR checklist

- [ ] If you've fixed a bug or added code that should be tested, add tests! (live test present but unverified, see TODOs)
- [x] If you've added a new tool - have you followed the module conventions in the contribution docs
- [x] Add a resource `label`
- [x] Remove all TODO statements. (none in module code; TODOs above are PR-level followups)

### DIFF
--- a/modules/nf-core/msisensorpro/baseline/main.nf
+++ b/modules/nf-core/msisensorpro/baseline/main.nf
@@ -1,0 +1,48 @@
+process MSISENSORPRO_BASELINE {
+    tag "$meta.id"
+    label 'process_low'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/msisensor-pro%3A1.3.0--hfef96ef_0':
+        'biocontainers/msisensor-pro:1.3.0--hfef96ef_0' }"
+
+    input:
+    tuple val(meta), path(list)
+    tuple val(meta2), path(pro_all_files, stageAs: 'pro_all/*')
+
+    output:
+    tuple val(meta), path("${prefix}.baseline"), emit: baseline
+    tuple val("${task.process}"), val('msisensor-pro'), eval("msisensor-pro --version 2>&1 | sed -nE 's/Version:\\s*//p'") , emit: versions_msisensorpro, topic: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    # Build the configure file consumed by \`msisensor-pro baseline -i\`:
+    #   one row per normal sample, columns: <sample_name>\\t<absolute_path_to_*_all>
+    # Sample names are derived from each file's basename with the "_all" suffix stripped.
+    : > configure.txt
+    for f in pro_all/*_all; do
+        [ -e "\$f" ] || continue
+        name=\$(basename "\$f" _all)
+        printf '%s\\t%s\\n' "\$name" "\$(readlink -f "\$f")" >> configure.txt
+    done
+
+    msisensor-pro \\
+        baseline \\
+        -d $list \\
+        -i configure.txt \\
+        -o ${prefix}.baseline \\
+        $args
+    """
+
+    stub:
+    prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    touch ${prefix}.baseline
+    """
+}
--- a/modules/nf-core/msisensorpro/baseline/main.nf
+++ b/modules/nf-core/msisensorpro/baseline/main.nf
@@ -1,0 +1,48 @@
+process MSISENSORPRO_BASELINE {
+    tag "$meta.id"
+    label 'process_low'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine in ['singularity', 'apptainer'] && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/msisensor-pro%3A1.3.0--hfef96ef_0':
+        'quay.io/biocontainers/msisensor-pro:1.3.0--hfef96ef_0' }"
+
+    input:
+    tuple val(meta), path(list)
+    tuple val(meta2), path(pro_all_files, stageAs: 'pro_all/*')
+
+    output:
+    tuple val(meta), path("${prefix}.baseline"), emit: baseline
+    tuple val("${task.process}"), val('msisensor-pro'), eval("msisensor-pro --version 2>&1 | sed -nE 's/Version:\\s*v//p'") , emit: versions_msisensorpro, topic: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    # Build the configure file consumed by `msisensor-pro baseline -i`:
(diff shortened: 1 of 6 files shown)


modules/nf-core/msisensorpro/baseline/tests/main.nf.test
```diff
@@ -0,0 +1,110 @@
...
+                    file(params.modules_testdata_base_path + 'genomics/homo_sapiens/genome/chr21/sequence/genome.fasta', checkIfExists: true)
+                ]
+                input[1] = [ [ id: 'baseline' ], [] ]
+                """
+            }
+        }
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
```
SPPearce:
```suggestion
                { assert snapshot(sanitizeOutput(process.out)).match() }
```

modules/nf-core/msisensorpro/baseline/tests/main.nf.test
```diff
@@ -0,0 +1,110 @@
...
+                    .map { meta, all -> all }
+                    .collect()
+                    .map { alls -> [ [ id: 'baseline' ], alls ] }
+                """
+            }
+        }
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
```
SPPearce:
```suggestion
                { assert snapshot(sanitizeOutput(process.out)).match() }
```

modules/nf-core/msisensorpro/baseline/main.nf
```diff
@@ -0,0 +1,48 @@
...
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    # Build the configure file consumed by `msisensor-pro baseline -i`:
+    #   one row per normal sample, columns: <sample_name>\\t<absolute_path_to_*_all>
+    # Sample names are derived from each file's basename with the "_all" suffix stripped.
+    : > configure.txt
```
SPPearce: Do we need this?
At least, maybe use `touch` as that is more obvious to me.

modules/nf-core/msisensorpro/baseline/main.nf
```diff
@@ -0,0 +1,48 @@
...
+    for f in pro_all/*_all; do
+        [ -e "\$f" ] || continue
+        name=\$(basename "\$f" _all)
+        printf '%s\\t%s\\n' "\$name" "\$(readlink -f "\$f")" >> configure.txt
```
SPPearce: Does it fail if we don't use `readlink` here?

modules/nf-core/msisensorpro/baseline/tests/main.nf.test
nh13: Applied, thanks.

modules/nf-core/msisensorpro/baseline/tests/main.nf.test
nh13: Applied, thanks.

modules/nf-core/msisensorpro/baseline/main.nf
nh13: Good call — switched the `: >` truncation to `touch configure.txt`, which is clearer and does the job in a fresh work dir.

modules/nf-core/msisensorpro/baseline/main.nf
nh13: I kept `readlink -f` for now as the robust choice: `msisensor-pro baseline` reads these paths back out of the configure file, and its behaviour with paths relative to the process CWD is not documented, so absolute paths avoid any ambiguity. I could not verify a relative-path variant locally (the amd64-only biocontainer segfaults under QEMU on Apple Silicon), so I did not want to swap it blind and risk breaking the x64 CI. Happy to push a relative-path variant and let CI confirm if you would prefer the simpler form.

maxulysse: Do we have readlink for the other msisensorpro modules?

nh13: Good point, removed the readlink.

SPPearce:
Linting is complaining, but otherwise looks good

nh13:
@SPPearce fixed, thanks!

SPPearce (approved):
All good

PR #11203, opened 2026-04-16
fix: split GPU CI into separate jobs per tag to fix silent test skipping

pinin4fjords:
## Problem

GPU tests tagged `gpu` (20 modules) never run. The matrix `include` in [`1f96a6d`](https://github.com/nf-core/modules/commit/1f96a6d8f85a7fad77dbf82024ba13f4dc3021aa) added `runner`/`tag` as new keys without a matching matrix dimension, so the last entry (`gpu_highmem`) wins for all jobs. [#11179](https://github.com/nf-core/modules/pull/11179) fixed detection but not execution.

### Why not just add `tag` as a matrix dimension?

Adding `tag: [gpu, gpu_highmem]` to the matrix fixes the matching, but doubles every shard. Only 1 of 21 GPU modules is `gpu_highmem`, so most PRs would spin up g4dn.2xlarge instances that find zero tests and exit.

## Fix

Split into two independent test jobs:

- **`nf-test-gpu`**: g4dn.xlarge, `--tag gpu` (20+ modules)
- **`nf-test-gpu-highmem`**: g4dn.2xlarge, `--tag gpu_highmem` (parabricks/rnafq2bam, needs >16GB RAM)

Each has its own detection, sharding, and `if` gate. A PR touching only `gpu` modules spins up zero `gpu_highmem` runners (and vice versa).

Validated on [#11197](https://github.com/nf-core/modules/pull/11197): all 4 `gpu` jobs passed on g4dn.xlarge (docker + singularity), `gpu_highmem` correctly skipped, `confirm-pass-gpu` green.

### DIFF
--- a/.github/workflows/nf-test-gpu.yml
+++ b/.github/workflows/nf-test-gpu.yml
@@ -101,13 +101,6 @@ jobs:
           paths: "${{ join(fromJson(steps.list-gpu-highmem.outputs.components), ' ') }}"
           tags: "gpu_highmem"
 
-      - name: debug
-        run: |
-          echo "gpu paths: ${{ steps.list-gpu.outputs.components }}"
-          echo "gpu shards: ${{ steps.shards-gpu.outputs.shard }} / ${{ steps.shards-gpu.outputs.total_shards }}"
-          echo "gpu_highmem paths: ${{ steps.list-gpu-highmem.outputs.components }}"
-          echo "gpu_highmem shards: ${{ steps.shards-gpu-highmem.outputs.shard }} / ${{ steps.shards-gpu-highmem.outputs.total_shards }}"
-
   # Standard GPU tests (g4dn.xlarge - 4 vCPU, 16 GB RAM, T4 GPU)
   # Runs tests tagged "gpu" (20+ modules: parabricks, amulety, cellpose, etc.)
   nf-test-gpu:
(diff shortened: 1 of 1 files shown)


.github/workflows/nf-test-gpu.yml
```diff
@@ -57,57 +59,108 @@ jobs:
...
+      NXF_ANSI_LOG: false
+      TOTAL_SHARDS: ${{ needs.nf-test-changes.outputs.gpu_total_shards }}
+
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          fetch-depth: 0
+      - name: test cuda
+        run: |
+          nvidia-smi -L
```
mashehu, line 126:
```suggestion
```
that's old debugging code we should be able to remove by now

mashehu (approved):
don't like all this duplication, but two tags don't make it worth to add yet another composite action. Let's see what the future holds.

.github/workflows/nf-test-gpu.yml
pinin4fjords: good point

pinin4fjords:
> don't like all this duplication, but two tags don't make it worth to add yet another composite action. Let's see what the future holds.

Yeah, I did come up with an alternative, but it involved some fairly arcane-looking JSON, and I decided the duplicate was easier to look after.

PR #11206, opened 2026-04-16
Add module: vcf2circos

### LINKED ISSUE #3234: new module: vcf2circos

tetedange13:
## Informations

New module for vcf2circos tool : https://github.com/bioinfo-chru-strasbourg/vcf2circos

Tool requires a 50 Mo big annotations bundle to work -> only `stub` tests, but made 1 for each output type
<br>

## PR checklist

Closes #3234 

- [x] Add a resource `label`
  - For modules:
    - [ ] `nf-core modules test <MODULE> --profile docker` DO NOT HAVE DOCKER AVAILABLE


tetedange13:
Failed CI tests seem related to the Ubuntu security repository mirror being temporarily out of sync...

modules/nf-core/vcf2circos/main.nf
```diff
@@ -0,0 +1,40 @@
...
+        ? 'https://depot.galaxyproject.org/singularity/vcf2circos:1.2.0--pyhdfd78af_0'
+        : 'biocontainers/vcf2circos:1.2.0--pyhdfd78af_0'}"
+
+    input:
+    tuple val(meta), path(vcf), path(vcf_index), val(output_extension)
+    tuple val(meta2), path(vcf2circos_config)
+
+    output:
+    tuple val(meta), path("*.{html,png,jpg,jpeg,webp,svg,pdf,eps,json}"), emit: circos, optional: true
+    tuple val("${task.process}"), val('vcf2circos'), eval("vcf2circos --help | grep Version | sed 's/^Version: //'"), topic: versions, emit: versions_vcf2circos
```
mashehu:
```suggestion
    tuple val("${task.process}"), val('vcf2circos'), eval("vcf2circos --help | sed -n 's/^Version: //p'"), topic: versions, emit: versions_vcf2circos
```
a more compact way to do the same

modules/nf-core/vcf2circos/main.nf
```diff
@@ -0,0 +1,40 @@
...
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    vcf2circos \\
+        ${args} \\
+        --options ${vcf2circos_config} \\
+        --input ${vcf} \\
```
akaviaLab, line 28: Is vcf_index ever used? If not, should it be removed from the input?

tetedange13: This tool supports both ".vcf" and "vcf.gz" so I guess "vcf_index" will be relevant if ".vcf.gz" input ?

famosab: sometimes the index is not handed explicitely but the tool will fail if its not present in the work dir :)

akaviaLab: But the first test seems to indicate it will work even without the index. So I wanted to check that it is used if it is present, and does it have to be specified explicitly?
Also, the webpage for the tool (https://github.com/bioinfo-chru-strasbourg/vcf2circos) doesn't seem to mention index files at all, so I was wondering if it is used?

modules/nf-core/vcf2circos/main.nf
```diff
@@ -0,0 +1,40 @@
...
+    container "${workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container
+        ? 'https://depot.galaxyproject.org/singularity/vcf2circos:1.2.0--pyhdfd78af_0'
+        : 'biocontainers/vcf2circos:1.2.0--pyhdfd78af_0'}"
+
+    input:
+    tuple val(meta), path(vcf), path(vcf_index), val(output_extension)
+    tuple val(meta2), path(vcf2circos_config)
+
+    output:
+    tuple val(meta), path("*.{html,png,jpg,jpeg,webp,svg,pdf,eps,json}"), emit: circos, optional: true
```
famosab, line 15: How can the only output be optional? :)

modules/nf-core/vcf2circos/main.nf
```diff
@@ -0,0 +1,40 @@
...
+    label 'process_medium'
+
+    conda "${moduleDir}/environment.yml"
+    container "${workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container
+        ? 'https://depot.galaxyproject.org/singularity/vcf2circos:1.2.0--pyhdfd78af_0'
+        : 'biocontainers/vcf2circos:1.2.0--pyhdfd78af_0'}"
+
+    input:
+    tuple val(meta), path(vcf), path(vcf_index), val(output_extension)
+    tuple val(meta2), path(vcf2circos_config)
```
famosab, line 12: How big is this config? Is this a config "one size fits all" so defined once per pipeline run?

tetedange13: Actually this config is a "options.json" which is part of a 50 Mo big annotation cache

I named it "config" in Nextflow because this parameter expect a file

akaviaLab: It seems optional, since you have tests that run without it. Where do you get this file? Could you modify one of the tests to use a config, so people will know what it looks like.
Maybe specify a link to an example in the meta.yml?

Also, if it isn't a nextflow config, I would suggest renaming the variable as options, which will avoid confusion.

modules/nf-core/vcf2circos/main.nf
tetedange13: For this "vcf_index" input, I simply considered that a ".vcf.gz" comes with a ".vcf.gz.{tbi,csi}"

Regarding if it is used or not, [you can provide a list of chromosomes](https://github.com/bioinfo-chru-strasbourg/vcf2circos#chromosomes-section) so I guess it *could* be used (but cannot assert that)

modules/nf-core/vcf2circos/main.nf
tetedange13: All my tests are `stub` ones (as said in PR description), that is why I did not specify this config/options file

Also I can confirm this file is **mandatory** (error if absent)

modules/nf-core/vcf2circos/main.nf
tetedange13: Adressed by commit 90166b7

modules/nf-core/vcf2circos/main.nf
tetedange13: Renamed variable in commit 5f092c5

akaviaLab:
I thought the point of tests is to check that the software works, not just
as a stub. The spec implies (
https://nf-co.re/docs/specifications/components/modules/testing#scope-of-testing)
that you can only have stub tests, but this seems wrong to me. Is it
possible for you to make non-stub tests please?

tetedange13:
> I thought the point of tests is to check that the software works, not just
as a stub. The spec implies (
https://nf-co.re/docs/specifications/components/modules/testing#scope-of-testing)
that you can only have stub tests, but this seems wrong to me. Is it
possible for you to make non-stub tests please?

For me it is OK to have only stub tests if tool requires a database : https://nf-co.re/docs/specifications/test-data/modules#test-data-alternatives-for-large-datasets

Which is the case here, as this "options/cache" is required and too big ?

tetedange13:
Human version of this cache is hosted by their lab as a "tar.gz", so I could use this with `untar` module

Unfortunately you need to manually set a path inside `options.json` : https://github.com/bioinfo-chru-strasbourg/vcf2circos#configuration-folder

> Regarding where you place your configuration folder previously downloaded, **you need to specify the absolute path of the Static folder in "Static" json key (which will replace default value)**
<br>

I think I could achieve that by switching `options` input for `annotations_dir` instead and setting correct `Static` path on-the-fly with sed ?

akaviaLab:
Thank you for your effort in adding this test.
I don't think this is enough, unfortunately.

I'm thinking about a user who wants to incorporate the module in a pipeline or even just use it alone. They'd have to figure out why the module isn't working, find the configuration file, and download it, etc. Also, since the file is hosted on a webserver that isn't Github, it might be rate limited.

Would it be possible for you to modify the bioconda recipe or have the 50MB file as an asset (of the nf-core module) or something? @famosab @mashehu - can you guys please weight in? Does it make sense to include this file, or just download it for each run? I don't know what the best practice is for nf-core.

Maybe it would make sense to write a small script to download the annotation file, place it as a module asset, and then run it if the annotation file isn't provided? My goal is to have it work out of the box as much as possible.

famosab:
I think you can also use stageAs and provide the file as input for the module. For the test, the file can be hosted via test-datasets I suppose (limit is 50MB). For a pipeline it might make sense to have a module that loads the reference, depending on the complexity its either just unzip (and give it the path) or something like pcgr/getref (see https://github.com/nf-core/modules/blob/master/modules/nf-core/pcgr/getref/main.nf)

PR #11212, opened 2026-04-17
feat(custom/multiqccustombiotype): sanity-check biotype cardinality

pinin4fjords:
## Summary

- Count biotype rows in the Python template. If the number of data rows exceeds `--max_biotypes` (parsed out of `task.ext.args`, default `100`), the process fails with a clear error instead of silently producing a missing/broken plot.
- Follows the `ext.args` + `argparse` + `shlex` pattern already used by `decoupler/decoupler` and other Python templates in nf-core/modules. No bespoke `ext.max_biotypes` property.
- Error message only references what the module can see (the input count file, `--max_biotypes`, and `featureCounts -g`), so it is correct in any consuming workflow.
- `meta.yml` updated to document the new behavior.

Addresses [nf-core/rnaseq#424](https://github.com/nf-core/rnaseq/issues/424), where a misconfigured `--featurecounts_group_type` (pointing at a per-gene attribute) caused MultiQC to hang.

### Why fail rather than skip?

The original issue proposed silent skip (`wc -l` + delete the file). In practice that leaves users with a missing plot, `0 %` rRNA in general stats, and no visible signal anywhere on the Nextflow console. Failing the process surfaces the misconfiguration immediately and points the user at both the offending file and the knob to change.

## Test plan

- [x] Existing `featurecounts biotype counts` test still passes (data output md5s unchanged).
- [x] New test `featurecounts biotype counts - too many biotypes fails` generates 150 synthetic biotype rows and asserts `process.failed` + `process.errorReport.contains('Too many categories')`.
- [x] New test `featurecounts biotype counts - custom max_biotypes via ext fails` supplies `ext.args = '--max_biotypes 2'` via `tests/nextflow.config` and asserts the same.
- [x] `nf-core modules lint custom/multiqccustombiotype` passes.

### Note on pre-existing snapshot drift

The `versions.yml` md5 in the pre-existing `featurecounts biotype counts` snapshot was stale on `master` (the mutable `biocontainers/python:3.12` tag has a different patch version than when the snapshot was recorded); it reproduced on a clean checkout before any of my changes. This PR refreshes that entry.

### DIFF
--- a/modules/nf-core/custom/multiqccustombiotype/environment.yml
+++ b/modules/nf-core/custom/multiqccustombiotype/environment.yml
@@ -2,4 +2,4 @@ channels:
   - conda-forge
   - bioconda
 dependencies:
-  - conda-forge::python=3.12.2
+  - conda-forge::python=3.12.12
--- a/modules/nf-core/custom/multiqccustombiotype/main.nf
+++ b/modules/nf-core/custom/multiqccustombiotype/main.nf
@@ -4,15 +4,15 @@ process CUSTOM_MULTIQCCUSTOMBIOTYPE {
 
     conda "${moduleDir}/environment.yml"
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
-        'https://depot.galaxyproject.org/singularity/python:3.12' :
-        'biocontainers/python:3.12' }"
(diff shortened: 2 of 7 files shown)


maxulysse (approved):

pinin4fjords:
Thanks @maxulysse !

PR #676, opened 2021-08-20
Add module: bamaligncleaner

jfy133:
## PR checklist

Closes #675 

- [x] This comment contains a description of changes (with reason).
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [x] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [x] If necessary, include test data in your PR.
- [x] Remove all TODO statements.
- [x] Emit the `<SOFTWARE>.version.txt` file.
- [x] Follow the naming conventions.
- [x] Follow the parameters requirements.
- [x] Follow the input/output options guidelines.
- [x] Add a resource `label`
- [x] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
    - [ ] `PROFILE=docker pytest --tag <MODULE> --symlink --keep-workflow-wd`
    - [ ] `PROFILE=singularity pytest --tag <MODULE> --symlink --keep-workflow-wd`
    - [x] `PROFILE=conda pytest --tag <MODULE> --symlink --keep-workflow-wd`

PR #11221, opened 2026-04-17
Remove DE user params

antoniasaracco:
This PR is to close ticket https://github.com/nf-core/differentialabundance/issues/551


antoniasaracco:
Hi @suzannejin, while working on this PR, I tested the master version of the abundance differential filter subworkflow and realized some nf-tests are not running correctly before I made any modifications.

The error found is:
`Test [b0399cb7] 'deseq2 - mouse - basic' 
> N E X T F L O W  ~  version 25.10.4
> Launching `/workspaces/modules/.nf-test-b0399cb7fed9d3995405e8b46b63634a.nf` [peaceful_wiles] DSL2 - revision: 5da61da68e
> Detected join operation duplicate emission on right channel -- offending element: key=[id:test]; value=[id:treatment_mCherry_hND6_sample_number, variable:treatment, reference:mCherry, target:hND6, blocking:sample_number],treatment,mCherry,hND6,null,null
> 
> 
> Join mismatch for the following entries: 
> - key=[id:treatment_mCherry_hND6__test, variable:treatment, reference:mCherry, target:hND6, blocking:, differential_method:deseq2] values= 
> - key=[id:treatment_mCherry_hND6_sample_number_test, variable:treatment, reference:mCherry, target:hND6, blocking:sample_number, differential_method:deseq2] values=`


Is the subworkflow missing any recent updates? Thanks for your help!

suzannejin:
> Is the subworkflow missing any recent updates? Thanks for your help!

Hi @antoniasaracco, I checked the pipeline subworkflow in `dev` branch does contain the lastest update: https://github.com/nf-core/differentialabundance/blob/dev/subworkflows/nf-core/abundance_differential_filter/main.nf

What is weird though, these updates are not reflected in your [PR](https://github.com/nf-core/differentialabundance/pull/676) (if you compare the subworkflow main.nf of `dev` from the link above and the `dev` in this PR side-by-side comparison - they look different - which should not).

antoniasaracco:
Hi @suzannejin, thanks for your response!

Sorry if my previous message was unclear. What I’m seeing is that the current version of the abundance_differential_filter subworkflow in the master branch of the modules repository, which matches the version in the pipeline dev branch you shared, is not passing the nf-tests on my side.

```
git status
On branch master
Your branch is up to date with 'origin/master'.
```

nothing to commit, working tree clean`

I am seeing several test failures related to duplicate emissions in a join operation, for example:

```
Detected join operation duplicate emission on right channel -- offending element: key=[id:test]
```
and subsequent join mismatches across multiple test cases.

I also noticed that this issue was reported today, and there is already a PR in place to address it: https://github.com/nf-core/modules/pull/11240

Once that is resolved, I should be able to continue working on this ticket: https://github.com/nf-core/differentialabundance/issues/551

Thanks!

subworkflows/nf-core/abundance_differential_filter/main.nf
```diff
@@ -18,6 +18,72 @@ def mergeMaps(meta, meta2){
...
+    ])
+
+    def existing_params = meta.params ?: [:]
+
+    def missing_runtime_params = runtime_params.findAll { key, _value ->
+        !existing_params.containsKey(key)
+    }
+
+    meta + [params: existing_params + missing_runtime_params]
+}
```
pinin4fjords, line 84: Main design concern sits here.

The subworkflow uses `differential_fc_column` internally (in the filter multiMap), but `differential_pval_column`, `differential_qval_column`, `differential_foldchanges_logged` aren't used anywhere in this subworkflow - they're written to `meta.params` purely for downstream consumers (differentialabundance's reports, shinyngs app, output DSL). That makes the component reach into pipeline-specific concerns.

Suggestion: drop `addDifferentialRuntimeParams` from the subworkflow entirely. Keep `getDifferentialMethodParams` as an internal helper and have the filter multiMap read `method_params.differential_fc_column` directly - it's derived from the explicit `analysis_method` input, so no meta plumbing is needed. If differentialabundance wants these keys on meta for downstream use, its local utils (`utils_nfcore_differentialabundance_pipeline`) can stamp them in at paramset-parsing time, following Suzanne's pattern in #676.

The duplicate table across the two repos (here and in `utils_nfcore_differentialabundance_pipeline`) is the visible symptom of this - two owners each filling in the fields they care about.

subworkflows/nf-core/abundance_differential_filter/main.nf
```diff
@@ -254,32 +291,16 @@ workflow ABUNDANCE_DIFFERENTIAL_FILTER {
...
-    // plots
-    dispersion_plot            = ch_dispersion_plot
-    md_plot                    = ch_md_plot
-
-    // rdata and session info
-    rdata                      = ch_rdata
-    session_info               = ch_session_info
-
-    // versions
     versions                   = ch_versions
```
pinin4fjords, line 305: Heads up - this emit block is missing `dispersion_plot`, `md_plot`, `rdata`, `size_factors`, `session_info` compared to master.

Those were added by @suzannejin in #11024 three weeks ago, specifically for the workflow output DSL migration in nf-core/differentialabundance#662. The pipeline's `main.nf` has an explicit `output { differential { ... } }` block that maps those exact names to publish folders, fed via `ABUNDANCE_DIFFERENTIAL_FILTER.out.<name>` at `workflows/differentialabundance.nf:964-968`. There's no `publishDir` fallback in `conf/modules.config` for these outputs, so removing them here stops those files from being published at all.

Also runs against the direction of travel for upcoming Nextflow versions where anything publishable has to flow through emits.

Likely just an accidental drop during the refactor - can you restore them?

subworkflows/nf-core/abundance_differential_filter/main.nf
```diff
@@ -18,6 +18,72 @@ def mergeMaps(meta, meta2){
...
+        ],
+        'propd' : [
+            differential_fc_column         : 'LFC',
+            differential_pval_column       : 'rcDdis',
+            differential_qval_column       : 'rcDdis',
+            differential_foldchanges_logged: true,
+            fc_cardinality                 : '>=',
+            stat_column                    : 'significant',
+            stat_cardinality               : '<='
+        ],
```
pinin4fjords, line 49: Worth sanity-checking with @suzannejin: `differential_pval_column` and `differential_qval_column` are both `rcDdis` - intentional? And `stat_column: 'significant'` for propd diverges from `differential_qval_column: 'rcDdis'` here, whereas for deseq2/limma/dream the two are the same column. (If the design point above lands and `addDifferentialRuntimeParams` comes out, this largely goes away - but the internal `stat_column` value should still be right.)

antoniasaracco:
Hi @pinin4fjords @suzannejin, Thanks for your feedback! I've adressed your comments and it's ready for review. Please let me know what you think.

pinin4fjords (approved):
Seems pretty simple now, just rearranging the logic

PR #11228, opened 2026-04-20
fastq_fastqc_umitools_trimgalore: handle null trim_log in read-count map

pinin4fjords:
## Summary

Three commits, one behaviour change:

1. **`1c2b050309`** — null-log fix.
2. **`546c722afd`** — `ch_` prefix rename on outer local channels (removes the collision with closure params for `trim_log`). No behaviour change.
3. **`7af199b8ad`** — lint fix: `reads_` keeps its trailing underscore in closures because `reads` is a `take:` parameter and Nextflow lint blocks the shadow.

## The bug

When `ext.args` includes `--hardtrim5` / `--hardtrim3`, trim_galore skips the adapter / quality pass and emits no `*_trimming_report.txt`. The `.join(trim_log, remainder: true)` then carries a `null` log element, and the null guard had two scoping bugs:

```groovy
.map { meta, reads_, trim_log_ ->
    if (trim_log) {                                     // outer channel, always truthy
        def num_reads = getTrimGaloreReadsAfterFiltering(
            meta.single_end ? trim_log_ : trim_log_[-1])
        [meta, reads_, num_reads]
    } else {
        [meta, reads, min_trimmed_reads.toFloat() + 1]  // outer take param, a channel
    }
}
```

- `if (trim_log)` checked the outer channel (always truthy) → `getTrimGaloreReadsAfterFiltering(null)` → `Cannot invoke method eachLine() on null object`.
- Else branch referenced the outer `reads` take parameter instead of the closure element.

Commit 1 fixes both by targeting the `_`-suffixed closure params. Commit 2 renames outer channels to `ch_*` so the `trim_log` closure param can drop its underscore. Commit 3 reinstates `reads_` because Nextflow lint rejects shadowing the `take:` parameter.

## Reproduction

Surfaced triaging [nf-core/rnaseq#1807](https://github.com/nf-core/rnaseq/issues/1807). Running the pipeline with `--extra_trimgalore_args '--hardtrim5 8'` crashes on master at `main.nf:15`.

## Test

New case `test paired end read without UMI - hardtrim5 (no trim log)` sets `ext.args = '--hardtrim5 8'` on TRIMGALORE. Fails on master, passes here. All 9 tests pass.

## Review tip

Most of the line count lives in commit 2 (the rename). The behaviourally-important lines appear **unchanged** in the final master-vs-HEAD diff because the rename flipped which name resolves to what. Self-review pins the real fix inline. Commit-by-commit review recommended.

PR checklist
---
- [ ] `CHANGELOG.md` is updated.
- [x] Ensure the test suite passes (`nf-core subworkflows test <SUBWORKFLOW> --profile docker`).

### DIFF
--- a/subworkflows/nf-core/fastq_fastqc_umitools_trimgalore/main.nf
+++ b/subworkflows/nf-core/fastq_fastqc_umitools_trimgalore/main.nf
@@ -36,81 +36,81 @@ workflow FASTQ_FASTQC_UMITOOLS_TRIMGALORE {
     min_trimmed_reads // integer: > 0
 
     main:
-    fastqc_html = channel.empty()
-    fastqc_zip = channel.empty()
+    ch_fastqc_html = channel.empty()
+    ch_fastqc_zip = channel.empty()
     if (!skip_fastqc) {
         FASTQC(reads)
-        fastqc_html = FASTQC.out.html
-        fastqc_zip = FASTQC.out.zip
+        ch_fastqc_html = FASTQC.out.html
+        ch_fastqc_zip = FASTQC.out.zip
     }
 
-    trimmer_reads = reads
-    umi_log = channel.empty()
-    umi_reads = channel.empty()
+    ch_trimmer_reads = reads
+    ch_umi_log = channel.empty()
+    ch_umi_reads = channel.empty()
     if (with_umi && !skip_umi_extract) {
         UMITOOLS_EXTRACT(reads)
-        trimmer_reads = UMITOOLS_EXTRACT.out.reads
-        umi_reads = UMITOOLS_EXTRACT.out.reads
-        umi_log = UMITOOLS_EXTRACT.out.log
+        ch_trimmer_reads = UMITOOLS_EXTRACT.out.reads
+        ch_umi_reads = UMITOOLS_EXTRACT.out.reads
+        ch_umi_log = UMITOOLS_EXTRACT.out.log
 
         // Discard R1 / R2 if required
         if (umi_discard_read in [1, 2]) {
             UMITOOLS_EXTRACT.out.reads
-                .map { meta, reads_ ->
-                    meta.single_end ? [meta, reads_] : [meta + ['single_end': true], reads_[umi_discard_read % 2]]
+                .map { meta, reads ->
+                    meta.single_end ? [meta, reads] : [meta + ['single_end': true], reads[umi_discard_read % 2]]
                 }
-                .set { trimmer_reads }
+                .set { ch_trimmer_reads }
         }
     }
 
-    trim_reads = trimmer_reads
-    trim_unpaired = channel.empty()
-    trim_html = channel.empty()
-    trim_zip = channel.empty()
-    trim_log = channel.empty()
-    trim_read_count = channel.empty()
+    ch_trim_reads = ch_trimmer_reads
+    ch_trim_unpaired = channel.empty()
+    ch_trim_html = channel.empty()
+    ch_trim_zip = channel.empty()
+    ch_trim_log = channel.empty()
+    ch_trim_read_count = channel.empty()
     if (!skip_trimming) {
-        TRIMGALORE(trimmer_reads)
-        trim_unpaired = TRIMGALORE.out.unpaired
-        trim_html = TRIMGALORE.out.html
-        trim_zip = TRIMGALORE.out.zip
-        trim_log = TRIMGALORE.out.log
+        TRIMGALORE(ch_trimmer_reads)
+        ch_trim_unpaired = TRIMGALORE.out.unpaired
+        ch_trim_html = TRIMGALORE.out.html
+        ch_trim_zip = TRIMGALORE.out.zip
+        ch_trim_log = TRIMGALORE.out.log
 
         //
         // Filter FastQ files based on minimum trimmed read count after adapter trimming
         //
         TRIMGALORE.out.reads
-            .join(trim_log, remainder: true)
-            .map { meta, reads_, trim_log_ ->
-                if (trim_log_) {
-                    def num_reads = getTrimGaloreReadsAfterFiltering(meta.single_end ? trim_log_ : trim_log_[-1])
-                    [meta, reads_, num_reads]
+            .join(ch_trim_log, remainder: true)
+            .map { meta, reads, trim_log ->
+                if (trim_log) {
+                    def num_reads = getTrimGaloreReadsAfterFiltering(meta.single_end ? trim_log : trim_log[-1])
+                    [meta, reads, num_reads]
                 }
                 else {
-                    [meta, reads_, min_trimmed_reads.toFloat() + 1]
+                    [meta, reads, min_trimmed_reads.toFloat() + 1]
                 }
             }
             .set { ch_num_trimmed_reads }
 
         ch_num_trimmed_reads
             .filter { _meta, _reads, num_reads -> num_reads >= min_trimmed_reads.toFloat() }
-            .map { meta, reads_, _num_reads -> [meta, reads_] }
-            .set { trim_reads }
+            .map { meta, reads, _num_reads -> [meta, reads] }
+            .set { ch_trim_reads }
 
         ch_num_trimmed_reads
             .map { meta, _reads, num_reads -> [meta, num_reads] }
-            .set { trim_read_count }
+            .set { ch_trim_read_count }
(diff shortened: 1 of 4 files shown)


subworkflows/nf-core/fastq_fastqc_umitools_trimgalore/main.nf
```diff
@@ -36,55 +36,55 @@ workflow FASTQ_FASTQC_UMITOOLS_TRIMGALORE {
...
+        ch_trim_zip = TRIMGALORE.out.zip
+        ch_trim_log = TRIMGALORE.out.log
 
         //
         // Filter FastQ files based on minimum trimmed read count after adapter trimming
         //
         TRIMGALORE.out.reads
-            .join(trim_log, remainder: true)
-            .map { meta, reads_, trim_log_ ->
+            .join(ch_trim_log, remainder: true)
```
pinin4fjords, line 83: **Fix anchor.** Master had `.join(trim_log, ...)` with `trim_log_` for the element below — same base name, distinguished only by a trailing underscore. That footgun is what let bug 1 slip in. Now `ch_trim_log` = channel, `trim_log` below = element, unambiguous. (`reads_` keeps its underscore because `reads` is a `take:` parameter and Nextflow lint won't allow shadowing.)

subworkflows/nf-core/fastq_fastqc_umitools_trimgalore/main.nf
```diff
@@ -36,55 +36,55 @@ workflow FASTQ_FASTQC_UMITOOLS_TRIMGALORE {
...
+            .map { meta, reads, trim_log ->
                 if (trim_log) {
```
pinin4fjords, line 85: **Bug fix line 1.** Text unchanged vs master, semantics flipped: `trim_log` used to resolve to the outer channel (always truthy), so the null-log branch was unreachable and `getTrimGaloreReadsAfterFiltering(null)` crashed. After the rename it resolves to the closure param (the element, `null` when trim_galore emits no report). The new hardtrim5 test fails here on master.

subworkflows/nf-core/fastq_fastqc_umitools_trimgalore/tests/main.nf.test
```diff
@@ -115,6 +115,42 @@ nextflow_workflow {
         }
     }
 
+    test("test paired end read without UMI - hardtrim5 (no trim log)") {
```
pinin4fjords, line 118: Regression test for the null-log path. `--hardtrim5 8` makes trim_galore emit no report, so `.join(..., remainder: true)` carries a `null` — exactly the crash scenario on master.

pinin4fjords (commented):
Self-review: three commits, one behaviour change.

- `1c2b050309` — null-log fix. Two scoping bugs (`if (trim_log)` tested the outer channel; else branch emitted the outer `reads` take parameter). Fixed by targeting the `_`-suffixed closure params.
- `546c722afd` — `ch_` prefix rename of the outer locals, drops the underscore from `trim_log` in the null-log map. Pure rename, no behaviour change.
- `7af199b8ad` — lint fix: `reads_` stays underscored in closures because `reads` is a `take:` parameter and Nextflow lint won't permit the shadow.

The behaviourally-important lines appear unchanged in the master-vs-HEAD diff because the rename flipped which name resolves to what. Inline comments pin the real fix. Review commit-by-commit.

subworkflows/nf-core/fastq_fastqc_umitools_trimgalore/main.nf
```diff
@@ -36,81 +36,81 @@ workflow FASTQ_FASTQC_UMITOOLS_TRIMGALORE {
...
+            .join(ch_trim_log, remainder: true)
+            .map { meta, reads_, trim_log ->
                 if (trim_log) {
-                    def num_reads = getTrimGaloreReadsAfterFiltering(meta.single_end ? trim_log_ : trim_log_[-1])
+                    def num_reads = getTrimGaloreReadsAfterFiltering(meta.single_end ? trim_log : trim_log[-1])
                     [meta, reads_, num_reads]
                 }
                 else {
-                    [meta, reads, min_trimmed_reads.toFloat() + 1]
+                    [meta, reads_, min_trimmed_reads.toFloat() + 1]
```
pinin4fjords, line 90: **Bug fix line 2.** The text changed from `reads` to `reads_`, but the interesting change is the semantics: on master `reads` resolved to the outer `take:` parameter (a channel), so had this branch ever been reached (it wasn't — see bug 1) it would have emitted a channel in place of the per-sample FASTQ list. `reads_` keeps the underscore because `reads` is a take parameter and Nextflow lint blocks the shadow.

subworkflows/nf-core/fastq_fastqc_umitools_trimgalore/tests/main.nf.test
```diff
@@ -115,6 +115,42 @@ nextflow_workflow {
...
+
+        config './nextflow-hardtrim.config'
+
+        when {
+            workflow {
+                """
+                input[0] = Channel.of([
```
vagkaratzas:
```suggestion
                input[0] = channel.of([
```

subworkflows/nf-core/fastq_fastqc_umitools_trimgalore/tests/main.nf.test
```diff
@@ -115,6 +115,42 @@ nextflow_workflow {
...
+                    [ id:'test', single_end:false ], // meta map
+                    [
+                        file(params.modules_testdata_base_path + 'genomics/sarscov2/illumina/fastq/test_1.fastq.gz', checkIfExists: true),
+                        file(params.modules_testdata_base_path + 'genomics/sarscov2/illumina/fastq/test_2.fastq.gz', checkIfExists: true)
+                    ]
+                ])
+                input[1] = true  // skip_fastqc
+                input[2] = false // with_umi
+                input[3] = false // skip_umi_extract
```
vagkaratzas: shouldnt this be `true` since reads are without UMI? or are you deliberately testing this parameter combo?

subworkflows/nf-core/fastq_fastqc_umitools_trimgalore/tests/main.nf.test
```diff
@@ -115,6 +115,42 @@ nextflow_workflow {
...
+        }
+
+        then {
+            assertAll(
+                { assert workflow.success },
+                { assert snapshot(
+                    workflow.out.reads,
+                    workflow.out.trim_read_count,
+                    workflow.out.trim_unpaired,
+).match() }
```
vagkaratzas: align :D

vagkaratzas (changes requested):
Nice! Minor adjustments and a question :D

subworkflows/nf-core/fastq_fastqc_umitools_trimgalore/tests/main.nf.test
pinin4fjords: Done in b56425b6f8.

subworkflows/nf-core/fastq_fastqc_umitools_trimgalore/tests/main.nf.test
pinin4fjords: No deliberate combo — I'd mirrored the existing "without UMI" test above (`skip_umi_extract: false`, `with_umi: false`), which is behaviourally equivalent because `with_umi: false` gates the UMI block off regardless. Flipped to `true` in b56425b6f8 for semantic clarity.

subworkflows/nf-core/fastq_fastqc_umitools_trimgalore/tests/main.nf.test
pinin4fjords: Done in b56425b6f8.

subworkflows/nf-core/fastq_fastqc_umitools_trimgalore/tests/main.nf.test
vagkaratzas: Hmm tests are failing now though; could it be that the flag is not properly used? Or just needs a snapshot update?

pinin4fjords: It's just singularity having a bad day I think

vagkaratzas (approved):
LGTM!

PR #11229, opened 2026-04-20
feat(subworkflows): add per_sample_mqc_bundle emit to fastq_qc / bam_qc_rnaseq / bam_markduplicates_picard / bam_dedup_umi

pinin4fjords:
## Summary

Adds a \`per_sample_mqc_bundle\` emit to four RNA-seq subworkflows so consumer pipelines that drive per-sample MultiQC can collapse each call site from an N-way \`.join(..., remainder: true)\` chain to a single consume. Motivation: nf-core/rnaseq PR [#1803](https://github.com/nf-core/rnaseq/pull/1803), where per-sample MultiQC under \`--skip_quantification_merge\` currently rebuilds a chained join at every contributor site inside \`workflows/rnaseq/main.nf\`.

Each subworkflow now emits a single \`[val(meta), list(files)]\` channel, assembled internally by joining every MultiQC-feeding output on \`meta\` with \`remainder: true\` at each step (progressive closure preserved). A small closure strips null holes and flattens list-shaped entries.

### Subworkflows touched

- \`fastq_qc_trim_filter_setstrandedness\` — 11 MQC contributors (fastqc raw/trim/filtered zips, trim log/json, umi log, bbsplit, sortmerna, ribodetector, seqkit, bowtie2)
- \`bam_qc_rnaseq\` — 12 contributors (preseq, biotype tsv, qualimap, dupradar, plus 8 RSeQC outputs)
- \`bam_markduplicates_picard\` — 4 contributors (stats, flagstat, idxstats, metrics)
- \`bam_dedup_umi\` — 4 contributors (genomic dedup log + genome stats/flagstat/idxstats). Transcriptome stats are excluded because MultiQC can't disambiguate them from genome stats without extra processing.

### Incidental change in \`bam_dedup_umi\`

Adds \`genome_stats\` / \`genome_flagstat\` / \`genome_idxstats\` as new discrete emits (the genome-side split from the existing combined \`stats\` / \`flagstat\` / \`idxstats\`). Needed so the bundle can include them cleanly; also useful on its own for consumers that only care about genome-side stats.

### Backward compatibility

All existing discrete emits are unchanged — this PR is purely additive. Pipelines that don't drive per-sample MultiQC can ignore the new channel.

## Test plan

- [x] \`bam_markduplicates_picard\`: 4/4 local tests pass. Stub snapshots regenerated to include the new emit key; existing keys unchanged.
- [x] \`bam_dedup_umi\`: 2/2 local tests pass. No snapshot changes.
- [x] \`bam_qc_rnaseq\`: 3/4 local tests pass. Stub snapshot patched to add the new key. The one non-stub failure is a pre-existing dupRadar md5 flake on my machine — verified by rerunning the same test against baseline (without my changes) and observing the same md5 mismatch.
- [x] \`fastq_qc_trim_filter_setstrandedness\`: 4/6 local tests pass. The two failures are SORTMERNA_INDEX SIGSEGV inside its container — a platform issue on Apple Silicon Docker emulation, unrelated to this change (confirmed: the bundle emit only appears after every process completes, and the failures happen in the rRNA-index step before any bundle logic runs).
- [ ] End-to-end rnaseq run under \`--skip_quantification_merge\` on a Linux host — in progress, will report back.

Relies on CI (Linux) for the two local-only platform flakes.

### DIFF
--- a/subworkflows/nf-core/bam_dedup_umi/main.nf
+++ b/subworkflows/nf-core/bam_dedup_umi/main.nf
@@ -120,6 +120,14 @@ workflow BAM_DEDUP_UMI {
         .mix(UMI_DEDUP_GENOME.out.idxstats)
         .transpose()
 
+    // Genome-side only; transcriptome stats excluded (MultiQC can't
+    // disambiguate them from genome stats without extra work).
+    ch_per_sample_mqc_bundle = ch_genomic_dedup_log
+        .join(UMI_DEDUP_GENOME.out.stats,    remainder: true)
+        .join(UMI_DEDUP_GENOME.out.flagstat, remainder: true)
+        .join(UMI_DEDUP_GENOME.out.idxstats, remainder: true)
+        .map { row -> [row[0], row.drop(1).findAll { it != null }.collectMany { e -> e instanceof List ? e : [e] }] }
+
     emit:
     bam                            = UMI_DEDUP_GENOME.out.bam // channel: [ val(meta), path(bam) ]
     index                          = UMI_DEDUP_GENOME.out.index // channel: [ val(meta), path(bai) ]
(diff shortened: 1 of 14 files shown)


pinin4fjords:
Status update on commit \`dea6ecfe86\`:

- **Linting**: green. The earlier failure was the strict-parser choking on a bare ternary inside a nested closure (\`e instanceof List ? e : [e]\`); adding parens around \`(e instanceof List)\` — mirroring the pattern already in nf-core/rnaseq's workflow-level \`collapseAgg\` — resolves it.
- **nf-test (Docker)**: all 16 shards green.
- **nf-test (Singularity)**: all 16 shards red, but the failures are \`HTTP Error 503: Service Unavailable\` from Canonical's Launchpad during setup — transient CI infra flake, unrelated to this PR.
- **End-to-end**: ran nf-core/rnaseq against the modified subworkflows on a Linux VM under \`-profile test,docker --skip_quantification_merge\`. Pipeline completed successfully in 4m17s, 240 tasks, all 5 samples through per-sample MultiQC.

A re-run of the singularity job should clear it once Launchpad recovers.

maxulysse (approved):

PR #11233, opened 2026-04-20
fix strict syntax warning in utils subworkflow

mashehu:


SPPearce (commented):
Do we want a comment to say that this is temporary?

mashehu:
this whole subworkflow is temporary until the holy grail aka nf-core-utils plugin can replace everything

maxulysse (approved):

PR #11249, opened 2026-04-21
fix `seqkit sample` for paired-end reads

bounlu:
`seqkit sample` requires 1 file at a time, this commit fixes it when the input is a read pair (paired-end reads). still works for single end read as well.


famosab:
Can you add a test for that? :)

SPPearce:
@famosab , can you review please, I added a test

famosab (approved):

PR #11255, opened 2026-04-21
[DRAFT - do not review] llamacpppython/run GPU + huggingface/download preview

pinin4fjords:
## Not for review / will be closed

This draft is open **only to preview the combined diff** of an existing PR (#11053) plus some proposed follow-up changes on top. It will be closed without merging. Any actual changes will land via that existing PR and/or a PR back into its source branch.

No action needed on this one; leaving it as draft to avoid reviewer attention.

## What's being previewed

On top of #11053, this branch:

- Drops the hand-rolled `Dockerfile` and adds an `environment.gpu.yml` (direct cu124 wheel URL + `conda-forge::cuda-runtime` pinned to 12.4 + `python=3.11`), built via Wave freeze with `LD_LIBRARY_PATH=/opt/conda/lib` baked in through `wave --config-env` so conda-installed CUDA libs are on the loader path without a script-side export.
- Rewrites the container directive as a four-URL dual-container ternary mirroring `ribodetector`'s shape (singularity https blob URLs + Wave docker URLs, branched on `task.accelerator`).
- Drops `label 'process_gpu'` — accelerator allocation is pipeline-controlled per the draft GPU module spec (nf-core/website#4142).
- Converts the CPU singularity URL from `oras://` to `https://` blob form so both CPU and GPU paths use the same URL scheme.
- Adds `tests/main.gpu.nf.test`, `tests/nextflow.gpu.config`, and a pre-generated `main.gpu.nf.test.snap` so the existing GPU CI workflow (`nf-test-gpu.yml`) picks the tests up via the `"gpu"` tag.
- Removes `nextflow.enable.moduleBinaries` from `tests/nextflow.config` (no longer needed once fully template-based).
- Fixes a stray `- prompt_file:` fragment in both `meta.yml` input blocks.

## Validation

- Wave: CPU container rebuild from `environment.yml` reproduces the existing hash byte-for-byte. GPU container built from `environment.gpu.yml`.
- CPU `nf-test` passes locally under both `-profile docker` and `-profile singularity`.
- GPU `nf-test` passes on a Tesla T4 (g4dn.xlarge) under `-profile docker,gpu` — end-to-end Gemma-3-1B inference ran at ~143 tokens/sec, both real and stub tests produce stable snapshots.
- `nf-core modules lint` matches `ribodetector`'s warning set (known Wave-tag / version-heuristic limitation on GPU containers).


pinin4fjords:
Preview closed. Follow-up work routes through the existing PR #11053 source branch.

PR #11258, opened 2026-04-22
fix: unpin ribodetector GPU from pytorch-gpu=1.11.0 (__cuda workaround no longer needed)

pinin4fjords:
## Summary

Unpin ribodetector's GPU container from pytorch-gpu=1.11.0. That version was the last release whose conda dependencies didn't require the `__cuda` virtual package, which is absent on Wave's GPU-less build servers. [seqeralabs/wave#1027](https://github.com/seqeralabs/wave/pull/1027) (merged) removes that constraint by retrying failed solves with `CONDA_OVERRIDE_CUDA` set, so any post-1.11 pytorch-gpu can now be built.

Given ribodetector is an inference-only CNN from 2022, the goal here is not to chase newer PyTorch; it's to unfreeze from the `__cuda`-avoidance pin while keeping host compatibility as wide as possible.

## Choices

- `conda-forge::pytorch-gpu=2.1.0` - oldest post-`__cuda`-era pytorch-gpu on conda-forge that has a py<=3.10 build matrix overlap with `bioconda::ribodetector=0.3.3` (which forces py<=3.10).
- `conda-forge::cuda-version=11.2` - lowest CUDA minor that 2.1.0's py<=3.10 builds target. NVIDIA driver floor ~450 (2020), which covers essentially every current HPC GPU host. For comparison, pinning at `>=12,<13` silently resolves to cuda-version=12.9 (driver floor ~575, early 2025).
- Both pins are exact (no ranges) per nf-core policy.

## Address @mashehu review

- [x] Pin versions exactly, no ranges (`cuda-version=11.2`).
- [x] Capture the CUDA runtime version in the versions topic (emits `cpu` on the non-GPU path, actual CUDA version on the GPU path).

## Validation

Wave build of the new `environment.gpu.yml` with the `conda/micromamba:v2` template succeeded and exercises the `__cuda` retry path (proves #1027 works for a real-world env):

```
community.wave.seqera.io/library/ribodetector_pytorch-gpu_cuda-version:fa9183da731515ea
oras://community.wave.seqera.io/library/ribodetector_pytorch-gpu_cuda-version:840843c8b08b83a5
```

`nf-core modules lint ribodetector` is clean apart from the known Wave-tag-version-heuristic warning shared with all GPU modules.

### Note for Wave team

For this env, the default Wave CLI `--await PT15M` fires before the build completes (it takes ~20-25 minutes end-to-end, likely due to the `__cuda` retry plus a large solver matrix for 2.1-era builds). Tag `ribodetector_pytorch-gpu_cuda-version:fa9183da731515ea` should be enough to locate the build on the Wave backend.

## Related

- [seqeralabs/wave#1026](https://github.com/seqeralabs/wave/issues/1026) - Wave `__cuda` issue
- [seqeralabs/wave#1027](https://github.com/seqeralabs/wave/pull/1027) - Wave fix (merged)
- [nf-core/rnaseq#1788](https://github.com/nf-core/rnaseq/pull/1788) - Pipeline-side integration
- #11203 - GPU CI workflow fix (split jobs per tag)
- #11259 - RFC on documenting multiple GPU container variants in `meta.yml` (out of scope here)
- Supersedes #11197

## Test plan

- [ ] CI module tests (CPU path) pass
- [ ] GPU container runs ribodetector on a sample dataset


modules/nf-core/ribodetector/environment.gpu.yml
```diff
@@ -5,4 +5,5 @@ channels:
   - bioconda
 dependencies:
   - "bioconda::ribodetector=0.3.3"
-  - "conda-forge::pytorch-gpu=1.11.0"
+  - "conda-forge::pytorch-gpu=2.10.0"
+  - "conda-forge::cuda-version>=12,<13"
```
mashehu: you need to pin the version, no ranges allowed

mashehu: this version should also be captured in the version topics

pinin4fjords: I think this might be a special case? Forcing an exact pin on cuda could have implications https://github.com/nf-core/website/pull/4142/changes

mashehu: sorry, can't find the exact implications there. Is it the backwards compatibility?

pinin4fjords: My bad, you're right, I was confused on something. Let me find the right pin.

pinin4fjords: Improving my understanding a bit, I've actually pinned it right back

pinin4fjords: done!

modules/nf-core/ribodetector/tests/main.nf.test.snap
```diff
@@ -8330,6 +8330,13 @@
                         "ribodetector",
                         "0.3.3"
                     ]
+                ],
+                "versions_cuda": [
+                    [
+                        "RIBODETECTOR",
+                        "cuda",
+                        "cpu"
```
mashehu: that's not a version i think 🙂

pinin4fjords: So, this is what we output when there isn't CUDA (so on CPU machines): https://github.com/nf-core/modules/blob/4c116045a58ae5dec4036f40ab0935284d4bf2cf/modules/nf-core/ribodetector/main.nf#L18

Happy to use another value if you prefer.

pinin4fjords:
@nf-core-bot update gpu snapshot path:modules/nf-core/ribodetector

modules/nf-core/ribodetector/tests/main.nf.test.snap
mashehu: hmm, would it be possible to not put out any version topic for these cases?

SPPearce (approved):

modules/nf-core/ribodetector/tests/main.nf.test.snap
mashehu: otherwise something like `no CUDA available` would be a bit more informative

pinin4fjords: done!

PR #11251, opened 2026-04-21
Add back fully-qualified registry URIs

ewels:
Nextflow v25.10 introduced a new config option `docker.registryOverride`. This allows folks to use their own mirrors of nf-core module containers.

We have had a workaround for a while now where all modules don't include their host and fall back to the default `docker.registry` that is set in the pipeline config, allowing people to configure mirrors with `docker.registry`. Now that we have `docker.registryOverride` this workaround is not needed.

It's better to have modules self-sufficient (not requiring a pipeline-level config attribute), so this PR goes back to how we used to do it with fully qualified URIs for all `container` attributes.

`docker.registry` should be ignored if a fully qualified URI is found (this already happens for Seqera Container URIs), so this should not be a breaking change. In the future, we can remove `docker.registry` from the pipeline config, but it's not required.

### Non-standard containers

In doing this work, I uncovered a few modules that have non-standard containers. Thought I'd list here just in case any are worth following up.

<details>
<summary><code>nvcr.io/nvidia</code> — 1 image(s), 12 module reference(s)</summary>

| Module | Container |
| --- | --- |
| [parabricks/applybqsr](https://github.com/nf-core/modules/blob/master/modules/nf-core/parabricks/applybqsr/main.nf) | `nvcr.io/nvidia/clara/clara-parabricks:4.6.0-1` |
| [parabricks/dbsnp](https://github.com/nf-core/modules/blob/master/modules/nf-core/parabricks/dbsnp/main.nf) | `nvcr.io/nvidia/clara/clara-parabricks:4.6.0-1` |
| [parabricks/deepvariant](https://github.com/nf-core/modules/blob/master/modules/nf-core/parabricks/deepvariant/main.nf) | `nvcr.io/nvidia/clara/clara-parabricks:4.6.0-1` |
| [parabricks/fq2bam](https://github.com/nf-core/modules/blob/master/modules/nf-core/parabricks/fq2bam/main.nf) | `nvcr.io/nvidia/clara/clara-parabricks:4.6.0-1` |
| [parabricks/fq2bammeth](https://github.com/nf-core/modules/blob/master/modules/nf-core/parabricks/fq2bammeth/main.nf) | `nvcr.io/nvidia/clara/clara-parabricks:4.6.0-1` |
| [parabricks/genotypegvcf](https://github.com/nf-core/modules/blob/master/modules/nf-core/parabricks/genotypegvcf/main.nf) | `nvcr.io/nvidia/clara/clara-parabricks:4.6.0-1` |
[...]

PR #11273, opened 2026-04-23
Add module: Percolator

julianu:
This PR add a new module for Percolator

## PR checklist

- [X] Add a resource `label`
  - For modules:
    - [X] `nf-core modules test <MODULE> --profile singularity`
    - [X] `nf-core modules test <MODULE> --profile conda`
  - For subworkflows:


modules/nf-core/percolator/main.nf
```diff
@@ -0,0 +1,42 @@
+process PERCOLATOR {
+    tag "$meta.id"
+    label 'process_medium'
+
+    conda "${moduleDir}/environment.yml"
+    container "${workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container
+        ? 'https://depot.galaxyproject.org/singularity/percolator:3.7.1--h6351f2a_0'
+        : 'biocontainers/percolator:3.7.1--h6351f2a_0'}"
```
mashehu:
```suggestion
        : 'quay.io/biocontainers/percolator:3.7.1--h6351f2a_0'}"
```
was just added yesterday to the template 🙂 #11251

modules/nf-core/percolator/tests/main.nf.test
```diff
@@ -0,0 +1,87 @@
...
+                """
+                input[0] = COMET.out.pin
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
```
famosab: You can use
```
```
to clean up the snapshot using [nft-utils](https://github.com/nf-core/nft-utils). It is used to clean process and workflow outputs by removing the numbered keys. This will create snapshots that are easier to read by humans.

modules/nf-core/percolator/main.nf
```diff
@@ -0,0 +1,42 @@
...
+    script:
+    def args = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    percolator \\
+        ${args} \\
+        --num-threads ${task.cpus} \\
+        --only-psms \\
+        --post-processing-tdc \\
+        --search-input concatenated \\
```
famosab: Why are these commands hard-coded? We usually prefer to allow users to specify them via `${args}`, or if need be then via value input channels.

modules/nf-core/percolator/meta.yml
```diff
@@ -0,0 +1,77 @@
...
+  decoy_pout:
+    - - meta:
+          type: map
+          description: Groovy Map containing sample information. e.g. `[
+            id:'sample1' ]`
+      - ${prefix}.psm.decoy.pout:
+          type: file
+          description: percolator output file containing the decoys PSMs
+          pattern: "*.pout"
+          ontologies: []
```
famosab, line 126: no ontologies available ? :(

famosab (commented):
Hey nice work, I added two comments that you can ignore and one question about the hard-coded args that you should address :)

modules/nf-core/percolator/main.nf
julianu: You are right, will change this for the parameters you highlighted.
As Percolator actually allows lots of output, all specified by filenames: is there a way to set these as well by meta-variables but in the modules.config - or any other useful way?

The outputs I put in there might be the most often used. But if someone wants to percolate on peptide level, this should also be possible to export (by `--results-peptides` and `--decoy-results-peptides`, like the `--results-psms` I already impleemnted).

Or should I create all files and the user decides, which ones to use? Are there any best practices or examples for this (could not find them, yet).

modules/nf-core/percolator/meta.yml
julianu: Added some, where fitting.
Actually, most of the output files are kind of TSVs, but special ones.

modules/nf-core/percolator/main.nf
julianu: Added most/all files as output now. Some can be (de-)activated by the `${args}`param

modules/nf-core/percolator/tests/main.nf.test
julianu: Thanks, added this.

famosab:
I am not sure if this applies but if the use cases are strictly separated for specific outputs we could also do two modules depeding on the use case. Like percolator/peptide percolator/... 
> If a tool has mutually exclusive functionality controlled only by flags (rather than subcommands), the flag name can replace the subcommand in the module name. (https://nf-co.re/docs/specifications/components/modules/general)

but in theory we can also have it in one as now and then mark all files as optional that are dependent on flags being set by users. You can use val input channels to control them if just using the args seems to flaky to you.

modules/nf-core/percolator/main.nf
```diff
@@ -0,0 +1,63 @@
...
+        --xmloutput ${prefix}.pout.xml \\
+        --pepxml-output ${prefix}.pep.xml \\
+        --tab-out ${prefix}.features.pin \\
+        --weights ${prefix}.weights.tsv \\
+        --results-peptides ${prefix}.pep.target.pout \\
+        --decoy-results-peptides ${prefix}.pep.decoy.pout \\
+        --results-psms ${prefix}.psm.target.pout \\
+        --decoy-results-psms ${prefix}.psm.decoy.pout \\
+        --results-proteins ${prefix}.protein.target.pout \\
+        --decoy-results-proteins ${prefix}.protein.decoy.pout \\
```
famosab: Does it change anything in the runtime if we have all these flags activated or are the analyses specifically run only when the flag is activated?

If its the first one we can leave as is

If its the second one I would suggest val input channels that control whether the flag is added to the command 

something like

```
def res_prot_cmd = args.contains("--protein") ? "--results-proteins ${prefix}.protein.target.pout" : ""
```

julianu: It actually does not really change the runtime, only the space required to write the result files. And the flags are also not really mutually exclusive. It's rather some require some other flags (some of which I did not hard-code).

I updated the module using some helper-flags as you suggested :+1: This looks much cleaner now indeed.

famosab: I think thats a nice solution.

famosab:
Can you update the branch once? :)

modules/nf-core/percolator/tests/main.nf.test
famosab, line 1: I would prefer a bit more test coverage where some of the files are actually being produced just that we test if the custom flags work as expected (you do not need to test all of them but a few would be good).

julianu: Right, added some flags in the test

modules/nf-core/percolator/tests/nextflow.config
famosab, line 1: We want the ext.args to be present in the main.nf.test file. That makes everything more readable in one go. See the [module specifications](https://nf-co.re/docs/specifications/components/modules/testing#configuration-of-extargs-in-tests) for more info.

julianu: Ah, perfect, did not know about this yet. Thanks!

famosab (approved):
I think it looks good! Great work 🚀

PR #11278, opened 2026-04-23
bump multiqc to 1.34

mashehu:
Linting is expected to fail, because mutliqc is the only module not exporting version topics.


maxulysse (approved):

maxulysse:
@mashehu could you update multiqcsav too?

PR #11283, opened 2026-04-24
New module: modkit/entropy

sahuno:
## PR checklist

- [x] Tests added (stub + real, both `--cpg` mode).
- [x] Versions broadcast via `topic: versions`.
- [x] Resource label set (`process_high`).
- [x] \`nf-core modules lint modkit/entropy\` — 50/0/0.
- [x] \`nf-test test --profile conda\` — 2/2 passed.

## Summary

Adds a new nf-core module wrapping [\`modkit entropy\`](https://nanoporetech.github.io/modkit/), which computes methylation entropy over genomic windows from one or more mod-BAMs. Entropy is a per-window measure of the diversity of methylation patterns across reads covering the window and is complementary to mean methylation.

The module supports modkit's file-vs-directory output dichotomy: when \`--regions\` is supplied (either via \`ext.args\` or the third input channel), modkit writes per-region BED / bedgraph / TSV into a directory (emitted via \`regions_bed\`, \`bedgraph\`, \`tsv\` outputs); otherwise a single genome-wide BED is produced (emitted via \`bed\`).

Accepts multiple input BAMs (passed as repeated \`--in-bam\`) for multi-sample entropy.

## Why

\`modkit entropy\` is the canonical tool for quantifying methylation heterogeneity in ONT methylation data and is not currently in nf-core/modules. It complements the existing \`modkit/pileup\` (mean methylation) for heterogeneity-aware analyses.

## Test data

Uses the existing \`test.sorted.phased.bam\` from nf-core/test-datasets (modules branch, \`genomics/homo_sapiens/nanopore/bam/\`). No new test data required.


modules/nf-core/modkit/entropy/main.nf
```diff
@@ -0,0 +1,52 @@
...
+        --threads ${task.cpus} \\
+        --ref ${fasta} \\
+        ${out_arg} \\
+        ${bam_args}
+    """
+
+    stub:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    touch ${prefix}.bed
```
itrujnara: Can we get the stub to simulate the bigger output when `regions` is passed?

modules/nf-core/modkit/entropy/tests/main.nf.test
```diff
@@ -0,0 +1,76 @@
...
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
+            )
+        }
+    }
+
+    test("homo sapiens - nanopore modbam - cpg") {
```
itrujnara: We need a test that uses the regions input

modules/nf-core/modkit/entropy/tests/main.nf.test
```diff
@@ -0,0 +1,76 @@
...
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert process.out.bed },
+                { assert path(process.out.bed[0][1]).exists() },
+                { assert path(process.out.bed[0][1]).size() > 0 },
+                // modkit entropy's BED output isn't byte-deterministic across CPU
```
itrujnara: Is it possible to check _anything_ inside the file (e.g. header line)?

itrujnara (commented):
Hi, thanks for the high-quality submission. I have left a handful of comments.

sahuno:
Thanks for the review @itrujnara — all three points addressed.

**1. Stub now simulates the `--regions` output.** `modkit entropy --regions`
writes a *directory* containing `<prefix>_regions.bed` and
`<prefix>_windows.bedgraph` instead of the single genome-wide BED, so the stub
branches on whether `regions` is present and creates the matching layout.

While confirming the filenames against `modkit` 0.6.1 I found the `tsv` output
channel never fires — `modkit entropy --regions` emits only the `_regions.bed`
and `_windows.bedgraph` files, no TSV. Removed that emit and corrected the
`meta.yml` description, which claimed a `.tsv` was produced.

**2. Added tests that use the `regions` input** — one real and one stub. The
real one asserts the exact output filenames, that `bed` is *not* emitted when
`regions` is passed, and that the `region_name` column echoes the interval
from the input BED.

**3. Yes — now checking file contents.** `modkit entropy` has a `--header`
flag, so the test config passes `--cpg --header` and the tests assert on the
header line and the row structure:

- genome-wide BED and per-window bedgraph:
  `#chrom  start  end  entropy  strand  num_reads`
- per-region BED: `chrom  start  end  region_name  mean_entropy  ...`
- every data row has all six columns and a parseable numeric entropy value

Still snapshotting only the versions topic for the non-stub tests: the entropy
column is not byte-deterministic across CPU counts (I get `0.69360054` vs
`0.6936004` for the same window between runs with different thread counts), so
an md5 snapshot would be flaky in CI.

modules/nf-core/modkit/entropy/main.nf
sahuno: Done. `--regions` makes `modkit entropy` write a *directory* rather than a single BED, so the stub now branches on the `regions` input and mirrors that layout:

```groovy
if (regions) {
    """
    mkdir -p entropy_regions
    touch entropy_regions/${prefix}_regions.bed
    touch entropy_regions/${prefix}_windows.bedgraph
    """
} else {
    """
    touch ${prefix}.bed
    """
}
```

modules/nf-core/modkit/entropy/tests/main.nf.test
sahuno: Added — `homo sapiens - nanopore modbam - cpg - regions` passes `genome.bed` as the regions input, plus a matching `- regions - stub` test. Both assert that `bed` is empty while `regions_bed` and `bedgraph` are populated, which is the output shape that distinguishes the two modes.

modules/nf-core/modkit/entropy/tests/main.nf.test
sahuno: Yes — the outputs are now asserted on structure rather than md5, because the entropy columns are floats and are not byte-deterministic across CPU counts/architectures. The genome-wide test checks the header is exactly `#chrom\tstart\tend\tentropy\tstrand\tnum_reads`, that there is at least one data row, and that every data row has six columns with a parseable entropy value. The regions test additionally checks the `_regions.bed` header, the output filenames, and that the `region_name` column echoes the input interval (`chr22:0-40001`). Only the versions topic is snapshotted for those two.

itrujnara (approved):
Looks good now, thank you

PR #11284, opened 2026-04-24
New module: modkit/validate

sahuno:
## PR checklist

- [x] Stub test added (see note on real test below).
- [x] Versions broadcast via \`topic: versions\`.
- [x] Resource label set (\`process_medium\`).
- [x] \`nf-core modules lint modkit/validate\` — 46/0/0.
- [x] \`nf-test test --profile conda\` — 1/1 passed (stub).

## Summary

Adds a new nf-core module wrapping [\`modkit validate\`](https://nanoporetech.github.io/modkit/), which benchmarks base-modification calls in one or more mod-BAMs against paired ground-truth BED files and emits a tab-separated summary of correct / incorrect / filtered calls per class.

The module accepts paired lists of BAMs and truth-BEDs (1:1 by list index) and forwards each pair as a repeated \`--bam-and-bed <BAM> <BED>\` argument to modkit.

## Why

\`modkit validate\` is the only canonical way to benchmark ONT methylation-calling accuracy against ground truth and is not currently in nf-core/modules. It's used to compare modkit pileup thresholds, evaluate basecaller-model upgrades, and assess spike-in control performance.

## Test data — stub only

\`modkit validate\` requires a BAM paired with a **ground-truth BED** whose name column is a single-letter modified-base code (\`m\`, \`h\`, ...) or \`-\` for a canonical reference position. The public nf-core test-datasets modules branch doesn't ship such a BED. Committed test is stub-only with a placeholder \`genome.bed\` for staging; the meta.yml documents the truth-BED contract for downstream users. Happy to contribute a small curated truth BED to nf-core/test-datasets in a follow-up.


modules/nf-core/modkit/validate/tests/main.nf.test
```diff
@@ -0,0 +1,65 @@
...
+    // A real validation run needs a BAM paired with a ground-truth BED whose
+    // `name` column is the short modified-base code (m/h/a/...) or `-` for a
+    // canonical reference position.
+    //
+    // The public modules test data doesn't ship such a BED, so for the real
+    // test we pipe `MODKIT_PILEUP` output through `cut -f1-6` (a bedMethyl's
+    // first six columns already satisfy the BED6 + mod-code-in-name contract)
+    // and validate the BAM against its own pileup. This is a self-consistency
+    // smoke test — it won't produce meaningful accuracy metrics but will
+    // exercise the full CLI path end-to-end.
```
SPPearce: You should then actually do that.
Currently you are running modkit/pileup in a setup block, but then not using that output to see if it works.

SPPearce (changes requested):

sahuno:
@SPPearce you're right — the `setup` block ran `MODKIT_PILEUP` and then threw
the result away, so nothing actually exercised `modkit validate`. Fixed.

There is now a real (non-stub) test that chains
`MODKIT_PILEUP -> GUNZIP -> MODKIT_VALIDATE` and feeds the resulting bedMethyl
in as the ground-truth BED. Two things I checked against `modkit` 0.6.1 while
wiring it up:

- `modkit validate` reads only the first six columns of the truth BED and
  ignores the remaining bedMethyl columns, so no `cut -f1-6` step is needed —
  the full 18-column bedMethyl and a 6-column cut produce byte-identical
  reports.
- `MODKIT_PILEUP` emits bgzipped output, and `modkit validate` cannot read a
  compressed BED (it fails with `Error! zero valid positions parsed from BED
  file`), so `GUNZIP` sits between them.

The `setup` block now lives inside the real test, so the stub test no longer
pays for a pileup run. The report TSV is byte-deterministic across repeated
runs, so it is snapshotted, plus explicit assertions that the
`raw_accuracy` / `filtered_accuracy` / `filter_threshold` keys are present.

Note this is still a self-consistency smoke test — validating a BAM against
its own pileup can't produce meaningful accuracy numbers, but it does exercise
the full CLI path. If a ground-truth BED for one of the nanopore test BAMs
would be welcome in nf-core/test-datasets, I'm happy to contribute one.

modules/nf-core/modkit/validate/tests/main.nf.test
sahuno: You were right — the `setup` block ran `MODKIT_PILEUP` and threw the result away, so nothing exercised `modkit validate`.

The non-stub test now chains `MODKIT_PILEUP -> GUNZIP -> MODKIT_VALIDATE`: the pileup bedMethyl is decompressed (`modkit validate` cannot read a bgzipped BED) and fed in as the truth BED, since its first six columns already satisfy the BED6 + mod-code-in-`name` contract that `validate` expects. On top of the snapshot it asserts that the report contains `raw_accuracy:`, `filtered_accuracy:` and `filter_threshold:`.

It is a self-consistency smoke test — validating a BAM against its own pileup will not produce meaningful accuracy numbers — but it does exercise the full CLI path end to end, and the public test data ships no ground-truth BED to do better. The comment block at the top of the test says so.

Snapshots have since been regenerated with `sanitizeOutput`, and the container is on `ont-modkit` 0.6.4.

PR #11285, opened 2026-04-24
New module: modkit/extractfull

sahuno:
## PR checklist

- [x] This comment contains a description of changes (with reason).
- [x] Stub + real tests added.
- [x] New tool follows [module conventions](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md).
- [x] Versions broadcast via \`topic: versions\`.
- [x] Naming/parameter/I-O conventions followed.
- [x] Resource label set (\`process_high\`).
- [x] BioConda + BioContainers used.
- [x] \`nf-core modules lint modkit/extract/full\` — 49/0/0.
- [x] \`nf-test test --profile conda\` — 2/2 passed.

## Summary

Adds a new nf-core module wrapping [\`modkit extract full\`](https://nanoporetech.github.io/modkit/), which transforms the MM/ML tags in a modBAM into a tab-separated per-read-per-position probability table. Emits one row for every modified-base probability call in every read.

The module auto-detects \`--bgzf\` in \`ext.args\` and adjusts the output filename suffix accordingly (\`.tsv\` vs \`.tsv.gz\`), so users don't get a misleading extension when enabling compression.

## Why

\`modkit extract full\` is the source of truth for read-level methylation probabilities and is essential for custom downstream filtering, phased methylation plots, and ML training on raw probability distributions. Paired with \`modkit extract calls\` (companion PR) which emits thresholded categorical calls.

## Test data

Uses the existing \`test.sorted.phased.bam\` from nf-core/test-datasets (modules branch). No new test data required.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #11286, opened 2026-04-24
New module: modkit/extractcalls

sahuno:
## PR checklist

- [x] Stub + real tests added.
- [x] Versions broadcast via \`topic: versions\`.
- [x] Resource label set (\`process_high\`).
- [x] \`nf-core modules lint modkit/extract/calls\` — 49/0/0.
- [x] \`nf-test test --profile conda\` — 2/2 passed.

## Summary

Adds a new nf-core module wrapping [\`modkit extract calls\`](https://nanoporetech.github.io/modkit/), which produces a per-read per-position table of base-modification **calls** (pass / fail / filtered, with the called base) using the same thresholding algorithm as \`modkit pileup\`.

Complementary to [\`modkit/extract/full\`](https://github.com/nf-core/modules/pull/new?head=sahuno:add-modkit-extract-full): \`extract calls\` emits the thresholded categorical decision per site per read, while \`extract full\` emits the underlying probabilities.

The module auto-detects \`--bgzf\` in \`ext.args\` and adjusts the output filename suffix accordingly.

## Why

\`modkit extract calls\` is the go-to tool for per-read allele-specific methylation, methylation-aware phasing validation, and read-level QC where you want the same thresholded labels as the pileup output but at read level rather than site level.

## Test data

Uses the existing \`test.sorted.phased.bam\` from nf-core/test-datasets (modules branch). No new test data required.


SPPearce (approved):

SPPearce (changes requested):
This is named incorrectly, it is too nested in the subfolder.
So modkit/foo, but not modkit/foo/bar

SPPearce:
So modkit/extractcalls here

sahuno:
@SPPearce thanks — renamed to `modkit/extractcalls`.

`modkit/extract/calls` was three levels deep, which fails the
`main_nf_module_granularity` lint check (`<tool>` or `<tool>/<subtool>` only).
Renamed the directory, the process name (`MODKIT_EXTRACTCALLS`), the `meta.yml`
`name:` field and the nf-test tags to match.

Sibling PR #11285 (`modkit/extractfull`) has had the same rename applied.

modules/nf-core/modkit/extractcalls/tests/main.nf.test
```diff
@@ -0,0 +1,72 @@
...
+                ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
+            )
```
SPPearce, line 38: This just means that it fails immediately if the tool fails (rather than giving all the mismatches with an empty output), and `sanitizeOutput` removes the numbered output from the snapshots.
```suggestion
            assert process.success
            assertAll(
                { assert snapshot(sanitizeOutput(process.out)).match() }
            )
```

modules/nf-core/modkit/extractcalls/tests/main.nf.test
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert process.out.tsv },
+                { assert path(process.out.tsv[0][1]).exists() },
+                { assert path(process.out.tsv[0][1]).size() > 0 },
+                { assert snapshot(process.out).match() }
+            )
```
SPPearce, line 65: I don't think you need these other asserts. The snapshot will show that the tsv exists, and if it is empty then the nf-core linting will complain it has an empty md5sum.
```suggestion
            assert process.success
            assertAll(
                { assert snapshot(sanitizeOutput(process.out)).match() }
            )
```

SPPearce (commented):
Generally fine now, please just tidy up the snapshot generation.

modules/nf-core/modkit/extractcalls/meta.yml
```diff
@@ -0,0 +1,115 @@
...
+    - bam:
+        type: file
+        description: Input modBAM with MM/ML tags.
+        pattern: "*.{bam,cram}"
+        ontologies:
+          - edam: http://edamontology.org/format_2572
+    - bai:
+        type: file
+        description: BAM index (`.bai` or `.csi`).
+        pattern: "*.{bai,csi}"
```
SPPearce: Presumably this could be crai too, if cram input is supported.

modules/nf-core/modkit/extractcalls/tests/nextflow.config
SPPearce, line 1: Don't need this file if empty.

sahuno:
@SPPearce all four addressed, thanks.

First a note that explains why the diff is bigger than the four comments: the
branch was ~980 commits behind master, so `sanitizeOutput` was not available on
it at all — it comes from `nft-utils` 1.x and the branch's `nf-test.config`
still pinned `nft-utils@0.0.9`. Rebased onto current master before making the
changes.

- **Snapshot generation.** Both tests now use your form: `assert process.success`
  outside `assertAll`, then a single `snapshot(sanitizeOutput(process.out)).match()`.
  Snapshot regenerated — it drops the duplicated numbered keys and is about half
  the size.
- **Redundant asserts.** Removed from the non-stub test.
- **`.crai`.** You were right, and I checked rather than assumed: `modkit extract
  calls` reads CRAM. I built one from `test.sorted.phased.bam` against
  `genome.fasta`, indexed it, and got the same 27,813 rows and exit 0 as the BAM
  input. The index input is now `pattern: "*.{bai,csi,crai}"`, and the `bam`
  description mentions that a CRAM input also needs `fasta` supplied.
- **`tests/nextflow.config`.** Deleted, along with the `config "./nextflow.config"`
  directive that loaded it — it only set `ext.args = ''`. The non-stub tsv md5 is
  unchanged (`0ce2e5a6…`), which confirms it was a no-op rather than something the
  test quietly depended on.

Both nf-tests pass locally under the singularity profile.

One thing to flag: lint now warns `bioconda_latest: ont-modkit 0.6.1 -> 0.6.4`.
I have left it at 0.6.1 to stay consistent with the merged `modkit/pileup` and my
three sibling modkit PRs (#11283, #11284, #11285). Happy to bump all of them
together in a follow-up if you would rather they move to 0.6.4.

SPPearce:
At some point a bump to the newer version would be good, but they are better to be all on the same version as you say. Happy for a bump to occur within one of these new module PRs.

SPPearce (approved):

SPPearce:
Ok, can merge this.
If you want, you can incorporate the version bump here.
Please update the other two modules with the same sanitizeOutput changes too (if you haven't already done this)

modules/nf-core/modkit/extractcalls/meta.yml
sahuno: Good catch — updated. `bai` is now documented as "Index for \`bam\` — \`.bai\`/\`.csi\` for a BAM, \`.crai\` for a CRAM", with `pattern: "*.{bai,csi,crai}"`, and the `bam` entry notes that CRAM input also requires `fasta`.

modules/nf-core/modkit/extractcalls/tests/nextflow.config
sahuno: Removed. The module is now just `environment.yml`, `main.nf`, `meta.yml` and `tests/{main.nf.test,main.nf.test.snap}`.

sahuno:
Thanks @SPPearce. Both done:

- **Version bump is in this PR** — `ff5d631` bumps `ont-modkit` 0.6.1 → 0.6.4 in `environment.yml`, and the snapshots were regenerated against it.
- **The sibling modules already carry the same `sanitizeOutput` change** — #11285 (`modkit/extractfull`), #11284 (`modkit/validate`), #11283 (`modkit/entropy`) and #11120 (`modkit/localize`) were each rebased onto master, switched to `assert process.success` + `snapshot(sanitizeOutput(process.out))`, had their snapshots regenerated and their containers bumped to 0.6.4. (`modkit/entropy` keeps its structural assertions on the non-stub tests — its float columns are not byte-deterministic across CPU counts — but the stubs use `sanitizeOutput`.)

The two review threads above are addressed and resolved. This one is green and mergeable whenever you are ready.

PR #11298, opened 2026-04-24
feat(gffread): emit BED output when --bed is in args

pinin4fjords:
## Summary

Add a `bed` emit to `nf-core/gffread` so `gffread --bed` is a first-class output format.

## Changes

- `main.nf`: add `bed` emit; `--bed` in `ext.args` picks the `.bed` extension (takes precedence over `-T` / `-w` / `-x` / `-y`) in both script and stub.
- `meta.yml`: document the new output.
- `tests/`: new `nextflow-bed.config` and two tests (`sarscov2-gff3-bed`, `sarscov2-gff3-bed-stub`); existing snapshots regenerated to include an empty `bed` slot (no behaviour change).

## Test plan

- [x] `nf-test test modules/nf-core/gffread/tests/main.nf.test --profile docker` — 9/9 pass.
- [x] `nf-core modules lint gffread` — 58 tests pass (was 54 on master), no new warnings/failures.


modules/nf-core/gffread/tests/main.nf.test
```diff
@@ -192,6 +192,69 @@ nextflow_process {
...
     }
 
+    test("sarscov2-gff3-bed") {
+
+        config "./nextflow-bed.config"
+
+        when {
+            params {
+                outdir = "$outputDir"
+            }
```
maxulysse: we should not need this
```suggestion
```

modules/nf-core/gffread/tests/main.nf.test
```diff
@@ -192,6 +192,69 @@ nextflow_process {
...
+                ]
+                input[1] = []
+                """
+            }
+        }
+
+        then {
+            assertAll (
+                { assert process.success },
+                { assert snapshot(process.out).match() },
```
maxulysse: can you use sanitizeOutput here?

pinin4fjords:
Addressed both review comments in b2ab20a — dropped the `params { outdir }` block and wrapped `process.out` with `sanitizeOutput` (also on the existing tests' snapshots that I had to regenerate, since the process now has an extra `bed` emit). All 9 tests still pass locally.

modules/nf-core/gffread/tests/main.nf.test
```diff
@@ -192,6 +192,63 @@ nextflow_process {
...
+            }
+        }
+
+        then {
+            assertAll (
+                { assert process.success },
+                { assert snapshot(sanitizeOutput(process.out)).match() },
+                { assert process.out.gtf == [] },
+                { assert process.out.gffread_gff == [] },
+                { assert process.out.gffread_fasta == [] }
```
maxulysse: do we need these assertions since we capture the whole process.out?

maxulysse (approved):
<3

modules/nf-core/gffread/tests/main.nf.test
pinin4fjords: maybe not

PR #11315, opened 2026-04-27
fix(trimgalore): isolate --paired intermediates in a subdir

pinin4fjords:
## Summary

In `--paired` mode trim_galore writes per-mate `*_trimmed.fq.gz` intermediates and unlinks them once `validate_paired_end_files` succeeds. If that cleanup is incomplete for any reason (an interrupted run, slow async filesystem, etc.), the intermediates remain in the workdir alongside the final `*_val_*.fq.gz` outputs and the `reads` output glob then matches 3 fastqs instead of 2. Downstream consumers expecting paired arity break.

This PR runs trim_galore with `--output_dir tg_work` and moves only the final outputs back to the workdir root. Anything that doesn't get cleaned up stays in the subdir where the workdir output glob never looks. The orphan-fastq failure mode is structurally impossible regardless of how cleanup behaves.

Also drops the `cutadapt=5.2` pin from `environment.yml` (added in #11308) - it was unnecessary; the `trim-galore` conda package already pulls in a working cutadapt.

SE mode is unchanged (no per-mate intermediates, so no leak possible).

## Supersedes

- #11308 (initial cleanup hack, too broad — deleted `*.fq.gz` inputs in some workflows)
- #11309 (narrowed cleanup hack — defensive but doesn't address incomplete-cleanup of the intermediate)

Both can be considered superseded by this approach.

## Test plan

- [ ] CI green

### DIFF
--- a/modules/nf-core/trimgalore/environment.yml
+++ b/modules/nf-core/trimgalore/environment.yml
@@ -5,4 +5,3 @@ channels:
   - bioconda
 dependencies:
   - bioconda::trim-galore=0.6.10
-  - bioconda::cutadapt=5.2
--- a/modules/nf-core/trimgalore/main.nf
+++ b/modules/nf-core/trimgalore/main.nf
@@ -46,8 +46,6 @@ process TRIMGALORE {
         def args_list = args.split("\\s(?=--)").toList()
         args_list.removeAll { arg -> arg.toLowerCase().contains('_r2 ') }
         """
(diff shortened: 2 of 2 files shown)


pinin4fjords:
Closing — superseded by ongoing discussion. The trimgalore behaviour will be addressed via Felix's Rust rewrite ([nf-core/rnaseq#1789](https://github.com/nf-core/rnaseq/pull/1789)), which architecturally avoids the per-mate intermediate files entirely.

PR #11317, opened 2026-04-27
fix(trimgalore): pin output globs to finals only, exclude PE intermediates

pinin4fjords:
## Summary

Replaces the broad `*{3prime,5prime,trimmed,val}{,_1,_2}.fq.gz` reads glob with a script-block-conditional glob pinned to exactly the finals trim_galore can produce in each mode:

| Mode | Glob |
|---|---|
| SE default | `\${prefix}_trimmed.fq.gz` |
| SE hardtrim5/3 | `\${prefix}.<N>bp_<3,5>prime.fq.gz` |
| PE default | `\${prefix}_<mate>_val_<mate>.fq.gz` |
| PE hardtrim5/3 | `\${prefix}_<mate>.<N>bp_<3,5>prime.fq.gz` |

The PE intermediate `\${prefix}_<mate>_trimmed.fq.gz` (which trim_galore writes during paired-end processing and `unlink`s after `validate_paired_end_files`) is no longer matched by the reads glob. An incomplete cleanup of those intermediates can no longer leak into the `reads` channel. No script-block hacks required.

Also drops the `cutadapt=5.2` pin from `environment.yml` (added in #11308) — unnecessary; the trim-galore conda package already pulls a working cutadapt.

## Why a static glob can't do this

PE intermediate ends in `_<digit>_trimmed.fq.gz`. SE final ends in `_trimmed.fq.gz` where the prefix may itself end in a digit (`WT_REP1_trimmed.fq.gz`). Distinguishing these requires negative lookbehind, which bash/Java glob syntax doesn't support. So a static-glob solution either includes the PE intermediate (current bug) or excludes legitimate SE finals for digit-suffixed sample IDs.

## Lint state

- 56 tests pass, 1 warning (bioconda update available — irrelevant)
- 2 failures:
  - `container_links` — **pre-existing on master**, unrelated to this PR
  - `correct_meta_outputs` — caused by this PR's use of bare-variable output paths

The `correct_meta_outputs` check does naive static parsing of `path(...)` and treats `path(reads_glob)` as the literal pattern `reads_glob` rather than resolving the script-block `def`. The same pattern is used elsewhere in nf-core/modules (e.g. [`samtools/sort`](https://github.com/nf-core/modules/blob/master/modules/nf-core/samtools/sort/main.nf) uses a `def extension` in script with `path("*.${extension}")` in output). The lint check should evaluate or skip variable-only output paths to support this conditional-output pattern, which is the only path-shape solution to the trimgalore intermediate-leak problem.

## Supersedes / alternative to

- #11308, #11309 — script-block `rm` hacks
- #11315 — subdir + `mv` approach (works, passes lint, also valid)

This PR is opened as **draft** for the lint discussion. If `correct_meta_outputs` can be made to handle this, this is structurally cleaner than #11315. If not, #11315 is the fallback.

## Test plan

- [ ] CI green on the test/snapshot side
- [ ] Lint discussion resolves `correct_meta_outputs`

### DIFF
--- a/modules/nf-core/trimgalore/environment.yml
+++ b/modules/nf-core/trimgalore/environment.yml
@@ -5,4 +5,3 @@ channels:
   - bioconda
 dependencies:
   - bioconda::trim-galore=0.6.10
-  - bioconda::cutadapt=5.2
--- a/modules/nf-core/trimgalore/main.nf
+++ b/modules/nf-core/trimgalore/main.nf
@@ -11,8 +11,8 @@ process TRIMGALORE {
     tuple val(meta), path(reads)
 
     output:
-    tuple val(meta), path(reads_glob)                            , emit: reads
-    tuple val(meta), path(report_glob)                           , emit: log     , optional: true
+    tuple val(meta), path("${reads_glob}")                       , emit: reads
+    tuple val(meta), path("${report_glob}")                      , emit: log     , optional: true
     tuple val(meta), path("${prefix}_{1,2}_unpaired_{1,2}.fq.gz"), emit: unpaired, optional: true
     tuple val(meta), path("${prefix}*_fastqc.html")              , emit: html    , optional: true
     tuple val(meta), path("${prefix}*_fastqc.zip")               , emit: zip     , optional: true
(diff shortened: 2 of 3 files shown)


pinin4fjords:
Closing - superseded by ongoing discussion. The trimgalore behaviour will be addressed via Felix's Rust rewrite (nf-core/rnaseq#1789).

PR #10975, opened 2026-03-19
Clair3 version topics migration

EfraMP:
## PR checklist

Closes #10643, related with all modules topics migration #10832

+ There was a major clair3 update from v1.5 to v2.0, migrating from TensorFlow models to PyTorch
+ Updated said models at the [test repo](https://github.com/nf-core/test-datasets/pull/1917) 
+ Updated the used containers and new parameters usage
+ ⚠️ Note that he `packaged_model` test fails on **conda** because in clair3 v2.0.0 the DL models are no longer distributed with the bioconda package. Docker and Singularity tests pass successfully, with the models bundled in the Docker/Singularity container at `/opt/models/`

- [x] This comment contains a description of changes (with reason).
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [x] If necessary, include test data in your PR.
- [ ] Remove all TODO statements.
- [x] Broadcast software version numbers to `topic: versions` - [See version_topics](https://nf-co.re/blog/2025/version_topics)
- [x] Follow the naming conventions.
- [x] Follow the parameters requirements.
- [x] Follow the input/output options guidelines.
- [x] Add a resource `label`
- [x] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - For modules:
    - [x] `nf-core modules test <MODULE> --profile docker`
    - [ ] `nf-core modules test <MODULE> --profile singularity`
    - [ ] `nf-core modules test <MODULE> --profile conda`
  - For subworkflows:
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile docker`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile singularity`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile conda`

PR #11327, opened 2026-04-28
Swap clair3 back to biocontainer and bump to v2.0.1

SPPearce:
clair3 was swapped to a docker container hosted in docker.io in #10975.
This PR moves back to a biocontainer and bumps to the most recent version.

### DIFF
--- a/modules/nf-core/clair3/tests/main.nf.test
+++ b/modules/nf-core/clair3/tests/main.nf.test
@@ -220,6 +220,9 @@ nextflow_process {
         options "-stub"
 
         when {
+            params {
+                args = ""
+            }
             process {
                 """
                 def model_path = UNTAR.out.untar
--- a/modules/nf-core/clair3/tests/main.nf.test.snap
+++ b/modules/nf-core/clair3/tests/main.nf.test.snap
@@ -54,27 +54,24 @@
                     [
                         "CLAIR3",
                         "clair3",
-                        "2.0.0"
+                        "2.0.1"
                     ]
                 ]
             }
         ],
-        "timestamp": "2026-03-19T12:58:29.886268",
         "meta": {
-            "nf-test": "0.9.4",
+            "nf-test": "0.9.3",
             "nextflow": "25.10.4"
-        }
+        },
+        "timestamp": "2026-04-28T07:36:02.280127687"
     },
     "sarscov2 - bam - packaged_model": {
         "content": [
             [
-                [
-                    "testmerge_output.vcf.gz",
-                    "VcfFile [chromosomes=[MT192765.1], sampleCount=1, variantCount=9, phased=false, phasedAutodetect=false]"
-                ]
+                
             ],
             [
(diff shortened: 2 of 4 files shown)


modules/nf-core/clair3/tests/main.nf.test.snap
```diff
@@ -54,27 +54,24 @@
...
-        "timestamp": "2026-03-19T12:58:29.886268",
         "meta": {
-            "nf-test": "0.9.4",
+            "nf-test": "0.9.3",
             "nextflow": "25.10.4"
-        }
+        },
+        "timestamp": "2026-04-28T07:36:02.280127687"
     },
     "sarscov2 - bam - packaged_model": {
```
SPPearce, line 68: Hmm, why is this losing the output?

modules/nf-core/clair3/tests/main.nf.test
```diff
@@ -220,6 +220,9 @@ nextflow_process {
         options "-stub"
 
         when {
+            params {
+                args = ""
+            }
```
awgymer, line 225: Is this necessary or just convention now?

modules/nf-core/clair3/tests/main.nf.test.snap
awgymer: This doesn't look right but I have a suspicion it may be to do with the packaged model throwing a silent error (i.e. your local tests exited fine but actually didn't run the tool) because they aren't in the correct place? 

I think the package location when using `conda` (and hence bioconda based installation in containers) is different to what is used in the developer based docker image. Are you able to check the logs of the test run locally?

modules/nf-core/clair3/tests/main.nf.test
SPPearce: nf-test was whinging, as it was not defined. Wasn't an error, just a warning.

modules/nf-core/clair3/tests/main.nf.test.snap
SPPearce: Yeah, there was an error in nf-core modules test that meant it was silently failing. 
Now fixed: https://github.com/nf-core/tools/issues/4227.
But yes, this test is not finding the model at the moment.
I need to check if the container here has the model at all, otherwise we should copy to our nf-core docker account.

awgymer: Yeah I am pretty sure that the bundled models are not included since the person updated to the v2.0.0 release (I don't think the person that did that on BioConda is a Clair3 dev and perhaps the devs didn't realise because they say it is on their README). 

I'm not sure how I feel about this but I think there are a few options:
- stop supporting a packaged model as an input
- make our own docker that has packaged models
- modify the bioconda recipe to restore packaged models

SPPearce: Hmm, modifying the bioconda recipe would be the nicest fix I guess...

PR #11329, opened 2026-04-28
fix(trimgalore): split reads emit into reads_se / reads_pe (alternative)

pinin4fjords:
## Summary

Alternative to [#11317](https://github.com/nf-core/modules/pull/11317) and [#11315](https://github.com/nf-core/modules/pull/11315). Replaces the single broad `reads` emit:

```nextflow
tuple val(meta), path("*{3prime,5prime,trimmed,val}{,_1,_2}.fq.gz"), emit: reads
```

with two narrower static emits:

```nextflow
tuple val(meta), path("\${prefix}{_trimmed,.*bp_3prime,.*bp_5prime}.fq.gz")    , emit: reads_se, optional: true
tuple val(meta), path("\${prefix}_*{_val_1,_val_2,bp_3prime,bp_5prime}.fq.gz") , emit: reads_pe, optional: true
```

Each glob is positively scoped to one mode and uses only single-level brace expansion (Java PathMatcher rejects nesting). The two patterns combine to cover every trim_galore output mode:

| File shape | matched by |
|---|---|
| `\${prefix}_trimmed.fq.gz` (SE default) | `reads_se` via `_trimmed` |
| `\${prefix}.<N>bp_3prime.fq.gz` (SE --hardtrim3) | `reads_se` via `.*bp_3prime` |
| `\${prefix}.<N>bp_5prime.fq.gz` (SE --hardtrim5) | `reads_se` via `.*bp_5prime` |
| `\${prefix}_<N>_val_<N>.fq.gz` (PE default) | `reads_pe` via `_*_val_<N>` |
| `\${prefix}_<N>.<N>bp_3prime.fq.gz` (PE --hardtrim3) | `reads_pe` via `_*bp_3prime` |
| `\${prefix}_<N>.<N>bp_5prime.fq.gz` (PE --hardtrim5) | `reads_pe` via `_*bp_5prime` |
| **`\${prefix}_<N>_trimmed.fq.gz` (PE per-mate intermediate)** | **neither** — falls through unmatched |

The PE intermediate is the file that trim_galore writes between the per-mate cutadapt pass and `validate_paired_end_files`, and unlinks once validation succeeds. On any infrastructure where that unlink is asynchronous (Fusion-mounted S3, slow network FS, etc.) and doesn't always propagate before the next process scans the workdir, the orphan would otherwise be picked up by the broad reads glob and downstream consumers expecting paired arity (`fq/lint`, etc.) break.

With this PR, the orphan is structurally unmatched regardless of whether trim_galore's cleanup succeeded — neither glob has a literal segment that would catch a `_<N>_trimmed.fq.gz` file.

## Why split rather than narrow

A single static glob can't distinguish PE intermediate (`\${prefix}_<N>_trimmed.fq.gz`) from SE final (`\${prefix}_trimmed.fq.gz` where `\${prefix}` may itself end in a digit, e.g. `WT_REP1`) — the difference requires negative lookbehind, which bash/Java globs don't support. Two positive patterns sidestep that constraint without dropping support for digit-suffixed sample IDs.

## Cost

Consumers that previously took `TRIMGALORE.out.reads` need to choose `reads_se` / `reads_pe` (or `.mix(reads_se, reads_pe)` if they don't care). Affects rnaseq, atacseq, taxprofiler, smrnaseq, nascent, and any other pipeline using this module — each gets a one-line update.

## Supersedes / alternative to

- [#11308](https://github.com/nf-core/modules/pull/11308), [#11309](https://github.com/nf-core/modules/pull/11309) — script-block `rm` hacks (defensive only, didn't address root cause)
- [#11315](https://github.com/nf-core/modules/pull/11315) — subdir + `mv` (works, but adds workdir gymnastics)
- [#11317](https://github.com/nf-core/modules/pull/11317) — conditional `reads_glob` via script-block `def` (cleaner shape but blocked by `correct_meta_outputs` lint, requires [nf-core/tools#4226](https://github.com/nf-core/tools/pull/4226))

This approach has none of those costs:
- No script-block hacks
- No subdir / no `mv`
- No lint changes
- No schema changes
- Static globs that meta.yml can mirror exactly

## Test plan

- [x] `correct_meta_outputs` passes on vanilla nf-core/tools (no patch needed)
- [ ] CI green
- [ ] Confirm the consumer-side `.mix()` updates are scoped — list of pipelines using TRIMGALORE.out.reads:
  - nf-core/rnaseq (subworkflow `fastq_fastqc_umitools_trimgalore`)
  - nf-core/atacseq
  - nf-core/taxprofiler
  - nf-core/smrnaseq
  - nf-core/nascent
  - others — to enumerate before merge

### DIFF
--- a/modules/nf-core/trimgalore/main.nf
+++ b/modules/nf-core/trimgalore/main.nf
@@ -11,11 +11,16 @@ process TRIMGALORE {
     tuple val(meta), path(reads)
 
     output:
-    tuple val(meta), path("*{3prime,5prime,trimmed,val}{,_1,_2}.fq.gz"), emit: reads
-    tuple val(meta), path("*report.txt")                               , emit: log     , optional: true
-    tuple val(meta), path("*unpaired{,_1,_2}.fq.gz")                   , emit: unpaired, optional: true
-    tuple val(meta), path("*.html")                                    , emit: html    , optional: true
-    tuple val(meta), path("*.zip")                                     , emit: zip     , optional: true
+    // Two `reads_*` emits with positive-only patterns covering every trim_galore
+    // output mode (default trim, --hardtrim5, --hardtrim3). PE per-mate intermediates
+    // (`${prefix}_<N>_trimmed.fq.gz`) deliberately match neither pattern, so an
+    // incomplete unlink of those files on retry can't leak into either channel.
+    tuple val(meta), path("${prefix}{_trimmed,.*bp_3prime,.*bp_5prime}.fq.gz")    , emit: reads_se, optional: true
+    tuple val(meta), path("${prefix}_*{_val_1,_val_2,bp_3prime,bp_5prime}.fq.gz") , emit: reads_pe, optional: true
+    tuple val(meta), path("*report.txt")                                          , emit: log     , optional: true
+    tuple val(meta), path("*unpaired{,_1,_2}.fq.gz")                              , emit: unpaired, optional: true
+    tuple val(meta), path("*.html")                                               , emit: html    , optional: true
+    tuple val(meta), path("*.zip")                                                , emit: zip     , optional: true
     tuple val("${task.process}"), val("trimgalore"), eval('trim_galore --version | grep -Eo "[0-9]+(\\.[0-9]+)+"'), topic: versions, emit: versions_trimgalore
 
     when:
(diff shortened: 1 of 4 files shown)


pinin4fjords:
Closing - superseded by ongoing discussion. The trimgalore behaviour will be addressed via Felix's Rust rewrite (nf-core/rnaseq#1789).

PR #11323, opened 2026-04-27
feat: add stub blocks to 41 modules (#4570)

HReed1:
Add stub blocks to 41 nf-core modules that were missing them, as tracked in nf-core/modules#4570.

Each stub generates placeholder output files matching the module's output channel declarations:
- Plain-text outputs use `touch`
- Compressed (`.gz`) outputs use `echo | gzip >`
- Index files (`.tbi`) use `touch`
- `versions.yml` blocks are copied 1:1 from the script block

Modules with `topic: versions` emit patterns skip `versions.yml` generation, as Nextflow handles these automatically.

Stub blocks were generated programmatically using an [AST mutation script](https://github.com/HReed1/hvr-agentic-os/blob/main/docs/nextflow/tools/scripts/stub_generator.py) and validated against a [local test suite](https://github.com/HReed1/hvr-agentic-os/blob/main/docs/nextflow/tools/tests/test_nf_stubs.py) for output parity. For full documentation on the tooling, conventions, and methodology used, see the [Nextflow contribution docs](https://github.com/HReed1/hvr-agentic-os/blob/main/docs/nextflow/README.md).

<details>
<summary><b>Modules affected (41)</b></summary>

- `abricate/run`
- `bakta/bakta`
- `ectyper`
- `emmtyper`
- `gambit/query`
- `genrich`
- `hpsuissero`
- `kleborate`
- `legsta`
- `lissero`
- `meningotype`
- `mlst`
- `mobsuite/recon`
- `multiqc`
- `multiqcsav`
- `ngmaster`
- `pasty`
- `pbptyper`
- `plasmidfinder`
- `pneumocat`
- `poppunk/assign`
- `quast`
- `seqsero2`
- `seroba`
- `shigatyper`
- `shigeifinder`
- `sistr`
- `snpdists`
- `spatyper`
- `ssuissero`
- `staphopiasccmec`
- `tbprofiler/profile`
- `tetyper`
- `virsorter2/run`

_(Full list may vary slightly — see diff for exact scope)_

</details>

## PR checklist

Contributes to #4570

- [x] This comment contains a description of changes (with reason).
- [ ] ~~If you've fixed a bug or added code that should be tested, add tests!~~ _Stub-only changes — no functional logic to test._
- [x] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] ~~If necessary, include test data in your PR.~~ _No test data needed for stubs._
[...]

PR #11349, opened 2026-04-28
feat: add stub blocks for assembler modules (#4570)

HReed1:
Adds deterministic `stub:` blocks and corresponding `-stub` nf-test cases for assembler modules, as part of the ongoing effort to resolve #4570. Split from the original monolithic #11323 per reviewer feedback.

Each stub generates placeholder output files matching the module's output channel declarations:
- Plain-text outputs use `touch`
- Compressed (`.gz`) outputs use `echo | gzip >`
- Index files (`.tbi`) use `touch`
- `versions.yml` blocks are copied 1:1 from the script block

Modules with `topic: versions` emit patterns skip `versions.yml` generation, as Nextflow handles these automatically.

Stub blocks were generated programmatically using an [AST mutation script](https://github.com/HReed1/hvr-agentic-os/blob/main/docs/nextflow/tools/scripts/stub_generator.py) and validated against a [local test suite](https://github.com/HReed1/hvr-agentic-os/blob/main/docs/nextflow/tools/tests/test_nf_stubs.py) for output parity. For full documentation on the tooling, conventions, and methodology used, see the [Nextflow contribution docs](https://github.com/HReed1/hvr-agentic-os/tree/main/docs/nextflow).

**Modules affected (6)**

- `medaka`
- `racon`
- `raven`
- `salsa2`
- `shasta`
- `shovill`

## Tests

All modules include `-stub` nf-test cases with `assertAll()` + `snapshot(process.out).match()`. Snapshots generated locally via `nf-test --update-snapshot`.

## PR checklist

Contributes to #4570. Split from #11323.

- [x] Stub tests added for all modules.
- [x] Follows the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md).

### DIFF
--- a/modules/nf-core/medaka/main.nf
+++ b/modules/nf-core/medaka/main.nf
@@ -37,4 +37,15 @@ process MEDAKA {
         medaka: \$( medaka --version 2>&1 | sed 's/medaka //g' )
     END_VERSIONS
     """
+
+    stub:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    echo | gzip > ${prefix}.fa.gz
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        medaka: \$( medaka --version 2>&1 | sed 's/medaka //g' )
+    END_VERSIONS
+    """
 }
--- a/modules/nf-core/medaka/tests/main.nf.test
+++ b/modules/nf-core/medaka/tests/main.nf.test
@@ -34,4 +34,29 @@ nextflow_process {
 
     }
 
+    test("Medaka - stub") {
+
+        options "-stub"
+
+        when {
+            process {
+                """
+                input[0] = [
+                    [ id:'test', single_end:true ], // meta map
+                    file(params.modules_testdata_base_path + 'genomics/sarscov2/nanopore/fastq/test.fastq.gz', checkIfExists: true),
+                    file(params.modules_testdata_base_path + 'genomics/sarscov2/genome/genome.fasta', checkIfExists: true)
+                ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
+            )
+        }
+
+    }
+
 }
--- a/modules/nf-core/salsa2/main.nf
+++ b/modules/nf-core/salsa2/main.nf
@@ -49,4 +49,20 @@ process SALSA2 {
         SALSA2: $VERSION
     END_VERSIONS
     """
+
+    stub:
+    def prefix = task.ext.prefix ?: "${meta.id}"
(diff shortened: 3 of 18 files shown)


HReed1:
@nf-core-bot fix linting

modules/nf-core/medaka/tests/main.nf.test
```diff
@@ -34,4 +34,29 @@ nextflow_process {
...
+                    file(params.modules_testdata_base_path + 'genomics/sarscov2/genome/genome.fasta', checkIfExists: true)
+                ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
```
famosab, line 56: You can use
```
```
to clean up the snapshot using [nft-utils](https://github.com/nf-core/nft-utils). It is used to clean process and workflow outputs by removing the numbered keys. This will create snapshots that are easier to read by humans.

modules/nf-core/medaka/main.nf
```diff
@@ -37,4 +37,15 @@ process MEDAKA {
         medaka: \$( medaka --version 2>&1 | sed 's/medaka //g' )
     END_VERSIONS
     """
+
+    stub:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    echo | gzip > ${prefix}.fa.gz
```
famosab, line 44: the linting fails because of that (see https://github.com/nf-core/tools/issues/4231)

famosab: but as soon as that linting check is fixed this PR should be good to go :)

modules/nf-core/salsa2/main.nf
famosab, line 1: Can you port the modules to topics that do not have that yet? (if not the linting will not permit merging ;) )

HReed1:
Thank you for the detailed review @famosab! These are all excellent catches that apply across all of our open PRs.

**sanitizeOutput()**: Totally right — I've now incorporated `sanitizeOutput()` into all our test snapshots going forward. Updated our internal workflow documentation with this as a mandatory convention.

**Topic channel requirement**: Understood. We plan to close PRs #11349–#11358 (the stub-block-only batches) since they all lack topic channel migration and will not pass linting. We're switching to a **1-module-per-PR strategy** where each PR includes: topic channel migration + stub block + `sanitizeOutput()` tests — all in one atomic unit.

PR #11377 (emmtyper) is our pilot for this new approach, with the eval() quoting and sanitizeOutput fixes already pushed. Once that goes green, we'll submit the remaining modules individually using the same template.

Thanks for guiding us in the right direction!

famosab:
If you swap your approach can you please close the PRs that are no longer relevant? Then nobody spends time reviewing something that is already outdated :)

HReed1:
Closing legacy batched PR in favor of atomic 1-PR-per-module approach with topic channel migration, as requested by reviewers.

PR #11368, opened 2026-04-29
Make SAMTOOLS/BGZIP not FASTA specific

itrujnara:
The current implementation on SAMTOOLS/BGZIP only makes it usable for FASTA files. Per [documentation](https://www.htslib.org/doc/htsfile.html), none of the tools in the module is restricted to FASTA inputs. To make it more versatile, I propose this modification to allow it to accept arbitrary compressed and uncompressed files. The change is partially backwards compatible, since `suffix` defaults to `fasta`.

## PR checklist

  - For modules:
  - For subworkflows:


itrujnara:
@nf-core-bot fix linting

maxulysse (approved):

SPPearce (dismissed):
No prefix2 instead of suffix

modules/nf-core/samtools/bgzip/main.nf
```diff
@@ -8,10 +8,11 @@ process SAMTOOLS_BGZIP {
         : 'community.wave.seqera.io/library/htslib_samtools:1.23.1--5b6bb4ede7e612e5'}"
 
     input:
-    tuple val(meta), path(fasta)
+    tuple val(meta), path(infile)
+    val suffix
```
prototaxites:
```suggestion
    val out_ext
```

Maybe be more explicit?

prototaxites (approved):
LGTM!

charles-plessy:
Can you take the opportunity to remove

```
    // samtools-bgzip has no --version option so let's use lastal from the same suite
```
from `main.nf`?  It is cruft carried over from the `last/lastal` module that I used as a template when I wrote `samtools/bgzip` 🤦‍♂️

PR #11372, opened 2026-04-29
Add cluster metrics viz

dbaku42:
## Description

This PR adds two new modules:

- **`cluster_metrics`**: Computes standard clustering quality metrics (Silhouette score, Calinski-Harabasz index, Davies-Bouldin index) and performs a k-sweep analysis.
- **`cluster_viz`**: Generates 2D visualizations (PCA, UMAP, t-SNE) colored by cluster label and exports the coordinates as TSV files.

Both modules are designed to work together with the existing `clustering` module (which performs KMeans/DBSCAN on PLINK2 PCA results).

### Features
- Full support for `conda` via `environment.yml`
- Comprehensive `nf-test` (normal + stub tests)
- Clean output channels with proper naming
- Optional plots (PNG) for visualization

### Author
- Donald Baku (@dbaku42)

### Related modules
- Related to: [#11337](https://github.com/nf-core/modules/pull/11337) (`clustering`)

---

This completes the core components of the `snpclustering` subworkflow.


SPPearce:
Um, you need to rebase these changes on the current master, as there was a bulk update (and you have a massive conflict ;) ).
Ping me when you have done that.

SPPearce:
Hmm, this is an entirely custom set of modules?
This needs discussion as to where they live. 
They could potentially be in custom/

dbaku42:
Thanks @SPPearce, that makes sense.

I’ve moved both custom Python-based modules under `modules/nf-core/custom/`, updated the internal paths and tests, fixed the author typo, and refreshed the nf-test snapshots.

Both module tests now pass locally for `cluster_metrics` and `cluster_viz`.

SPPearce (changes requested):
I think it'd need to be custom/clustermetrics and custom/clustervisualiation (no underscores allowed).
You'll need to use `template` directly, rather than referring to `${projectDir}/modules/nf-core/custom/cluster_metrics/templates/cluster_metrics.py`.

dbaku42:
Thanks @SPPearce — I renamed the custom modules to remove underscores and updated both modules to use `${moduleDir}` instead of hardcoded `${projectDir}` paths. I also updated the tests and snapshots accordingly.

SPPearce (changes requested):
You can't use `moduleDir` at all in the main.nf, it won't work on cloud systems as the python script won't be there.
You need to use a template, [like these modules](https://github.com/search?q=repo%3Anf-core%2Fmodules+templates&type=code).

modules/nf-core/custom/clustermetrics/Dockerfile
SPPearce, line 1: You could make the docker/singularity containers via Seqera Containers rather than needing a custom image in the nf-core quay.io.

modules/nf-core/custom/clustermetrics/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM nfcore/base:2.0
```
pinin4fjords: **Severity: high** - delete this file. nf-core modules use Biocontainers or Seqera Containers built from `environment.yml`, not hand-rolled Dockerfiles, unless there is a specific reason (there is not here, the env is straightforward Python+sklearn). Same applies to `clustervisualiation/Dockerfile`.

pinin4fjords (commented):
Hi @dbaku42, thanks for the contribution. AI-assisted review (Claude, on behalf of @pinin4fjords). Some overlap with @SPPearce's earlier feedback. Beginner-friendly notes plus suggestion blocks you can apply directly.

The biggest fix is the `templates/` mechanism. Right now your `script:` block runs `python3 ${moduleDir}/templates/cluster_metrics.py`, which is just running a normal Python script that happens to live in `templates/`. **No templating is happening.** This works on your laptop because `${moduleDir}` resolves to a local path, but it breaks on cloud/HPC executors because that folder is not staged into the work directory the task runs in.

Nextflow's `template` directive does two things:
1. Stages the file into the task work directory automatically (so it works on any executor).
2. Runs the file through Groovy GString interpolation before staging, so `${features}`, `${task.ext.prefix}`, etc. get substituted into the Python source. This replaces `argparse` entirely.

Because the file goes through Groovy interpolation, a few Python source patterns need adjusting (`\n` -> `\\n`, `r"\s+"` -> `r"\\s+"`, etc., I verified this locally). I have flagged the specific spots inline. See `modules/nf-core/custom/tx2gene/templates/tx2gene.py` for the canonical shape.

**Two scope questions for discussion:**
- The `*_selected.json` "best k" output bakes a decision into the module, and the sweep refits KMeans regardless of what method produced the input labels. Could the module just emit the k-sweep table and let the consuming pipeline pick?
- `cluster_viz` does PCA + UMAP + t-SNE + plotting + TSV export, and the PCA piece is essentially a re-plot of upstream PLINK eigenvec output. Consider dropping PCA plotting from this module and tightening to the embedding methods that genuinely belong together.

The `clustervisualiation` directory name looks like a typo for `clustervisualization`; worth fixing while you're renaming things.

modules/nf-core/custom/clustermetrics/main.nf
```diff
@@ -0,0 +1,52 @@
...
+
+/*
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    PROCESS: CLUSTER_METRICS
+    Compute clustering quality metrics and k-sweep
+    Author: Donald Baku (author)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+*/
+
+process CLUSTER_METRICS {
```
pinin4fjords: **Severity: low/medium** - nf-core modules do not include the `#!/usr/bin/env nextflow`, `nextflow.enable.dsl = 2`, or banner comment, just the `process` declaration. And the `custom/` convention is `CUSTOM_<NAME>` (matches `CUSTOM_TX2GENE`, `CUSTOM_MATRIXFILTER`, etc.).

```suggestion
process CUSTOM_CLUSTERMETRICS {
```

modules/nf-core/custom/clustermetrics/main.nf
```diff
@@ -0,0 +1,52 @@
...
+    tag "$meta.id"
+    label 'process_medium'
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/clustermetrics:dev' :
+        'quay.io/nf-core/clustermetrics:dev' }"
```
pinin4fjords: **Severity: high** - the `:dev` URIs do not resolve (lint flagged 403/404). These are Seqera Containers built from your `environment.yml`. Note `in ['singularity', 'apptainer']` so apptainer users get the SIF too.

```suggestion
    container "${ workflow.containerEngine in ['singularity', 'apptainer'] && !task.ext.singularity_pull_docker_container ?
        'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/69/69a6d33f6bd1a901cad8a6914b6ad11a7db6c35005b4ff8604f20f1baba10fc3/data' :
        'community.wave.seqera.io/library/matplotlib_pandas_python_scikit-learn:b7d7028d28dc4084' }"
```

modules/nf-core/custom/clustermetrics/main.nf
```diff
@@ -0,0 +1,52 @@
...
+    input:
+    tuple val(meta), path(features), path(clusters)
+    val out_prefix
```
pinin4fjords: **Severity: medium** - drop the `val out_prefix` input. The convention is to set the output prefix via `task.ext.prefix` (configured in `modules.config` of the consuming pipeline). You are already falling through to it on L34.

```suggestion
    input:
    tuple val(meta), path(features), path(clusters)
```

modules/nf-core/custom/clustermetrics/main.nf
```diff
@@ -0,0 +1,52 @@
...
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        python: \$(python3 --version | cut -d' ' -f2)
+        pandas: \$(python3 -c "import pandas; print(pandas.__version__)")
+        scikit-learn: \$(python3 -c "import sklearn; print(sklearn.__version__)")
+        matplotlib: \$(python3 -c "import matplotlib; print(matplotlib.__version__)")
+    END_VERSIONS
+    """
+}
```
pinin4fjords, line 46: **Severity: high** - see the body for what the `template` directive does and why the current `python3 ${moduleDir}/templates/...` approach will not work on cloud executors. The Python file then needs the conversion noted in the comments on `cluster_metrics.py`.

```suggestion
    script:
    template 'cluster_metrics.py'

    stub:
    def prefix = task.ext.prefix ?: "${meta.id}"
    """
    touch ${prefix}_metrics.tsv
    touch ${prefix}_k_sweep.csv
    touch ${prefix}_selected.json

    cat <<-END_VERSIONS > versions.yml
    "${task.process}":
        python: \$(python3 --version | sed 's/Python //')
        pandas: \$(python3 -c "import pandas; print(pandas.__version__)")
        scikit-learn: \$(python3 -c "import sklearn; print(sklearn.__version__)")
        matplotlib: \$(python3 -c "import matplotlib; print(matplotlib.__version__)")
    END_VERSIONS
    """
}
```

modules/nf-core/custom/clustermetrics/meta.yml
```diff
@@ -0,0 +1,93 @@
+name: "CLUSTER_METRICS"
```
pinin4fjords: **Severity: low** - all other `custom/` modules use the lowercase `custom_<name>` form here.

```suggestion
name: "custom_clustermetrics"
```

modules/nf-core/custom/clustermetrics/templates/cluster_metrics.py
```diff
@@ -0,0 +1,285 @@
...
+
+    if "sample_id" not in df.columns:
+        raise ValueError("features file must contain a sample_id column after normalization")
+
+    sample_ids = df["sample_id"].astype(str)
+    x = df.drop(columns=["sample_id"]).apply(pd.to_numeric, errors="coerce")
+    x = x.fillna(x.mean(numeric_only=True))
+    x = x.fillna(0.0)
+
+    return x, sample_ids
```
pinin4fjords: **Severity: medium** - replace `_normalise_id_column` and `load_features` with one simpler function. The original is doing PLINK eigenvec format normalisation (FID/IID handling, header-row dedup, `#` stripping); that is pipeline glue, not module behaviour. Have the consuming pipeline pre-process its PLINK output to a clean `sample_id`-keyed TSV before passing it in. The meta.yml already advertises `sample_id` as the input format.

Note `sep="\\t"` (doubled) for template-time Groovy interpolation.

```suggestion
def load_features(path: str) -> tuple[pd.DataFrame, pd.Series]:
    df = pd.read_csv(path, sep="\\t")
    if "sample_id" not in df.columns:
        raise ValueError(f"features file must have a 'sample_id' column. Found: {list(df.columns)}")
    sample_ids = df["sample_id"].astype(str)
    x = df.drop(columns=["sample_id"]).apply(pd.to_numeric, errors="coerce").fillna(0.0)
    return x, sample_ids
```

modules/nf-core/custom/clustermetrics/templates/cluster_metrics.py
```diff
@@ -0,0 +1,285 @@
...
+            out = pd.DataFrame(
+                {
+                    "sample_id": candidate_vals,
+                    "cluster": pd.to_numeric(df[cluster_col], errors="raise").astype(int),
+                }
+            )
+            return out, "sample_id"
+
+    out = pd.DataFrame({"cluster": pd.to_numeric(df[cluster_col], errors="raise").astype(int)})
+    return out, "row_order"
```
pinin4fjords: **Severity: medium** - same simplification: drop the multi-path fallback (sample_id / IID / single-non-numeric-column / row-order). Require `sample_id` and `cluster`, period. Callers already produce that format.

```suggestion
def load_clusters(path: str) -> pd.Series:
    df = pd.read_csv(path)
    if "sample_id" not in df.columns or "cluster" not in df.columns:
        raise ValueError(f"clusters file must have 'sample_id' and 'cluster' columns. Found: {list(df.columns)}")
    return df.set_index(df["sample_id"].astype(str))["cluster"].astype(int)
```

modules/nf-core/custom/clustermetrics/templates/cluster_metrics.py
```diff
@@ -0,0 +1,285 @@
...
+        alignment_mode = "row_order"
+
+    if len(x) < 2:
+        raise ValueError("Need at least 2 samples to compute cluster metrics")
+
+    selected = safe_cluster_metrics(x, labels)
+    selected["input_clusters"] = Path(args.clusters).name
+    selected["input_features"] = Path(args.features).name
+    selected["n_samples_used"] = int(len(aligned_ids))
+    selected["alignment_mode"] = alignment_mode
```
pinin4fjords: **Severity: high** - once `main.nf` switches to `template`, the `argparse` block here goes away and the early branching collapses. Pattern (no suggestion block because args are referenced further down too, easier to do as a single edit):

```python
def main() -> None:
    features = "${features}"
    clusters_path = "${clusters}"
    prefix = "${task.ext.prefix}" if "${task.ext.prefix}" != "null" else "${meta.id}"
    k_min, k_max = 2, 12

    x_df, sample_ids = load_features(features)
    clusters = load_clusters(clusters_path)

    common = sample_ids[sample_ids.isin(clusters.index)]
    if len(common) < 2:
        raise ValueError("need at least 2 overlapping samples between features and clusters")
    x = x_df.loc[common.index].values
    labels = clusters.loc[common.values].values

    selected = safe_cluster_metrics(x, labels)
    pd.DataFrame([selected]).to_csv(f"{prefix}_metrics.tsv", sep="\\t", index=False)
    # ... rest of main, replacing args.X with the locals above and args.out_prefix with prefix
```

The `input_clusters`, `input_features`, `n_samples_used`, `alignment_mode` fields in `selected` are also redundant once `load_clusters` always returns a Series keyed by `sample_id` (paths the pipeline already knows; alignment mode no longer varies). Worth trimming during the rewrite.

Also: write `versions.yml` from inside Python at the end of `main()` like in `custom/tx2gene/templates/tx2gene.py` (the `format_yaml_like` helper there is reusable, copy it).

modules/nf-core/custom/clustermetrics/templates/cluster_metrics.py
```diff
@@ -0,0 +1,285 @@
...
+
+    metrics_tsv = f"{args.out_prefix}_metrics.tsv"
+    pd.DataFrame([selected]).to_csv(metrics_tsv, sep="\t", index=False)
```
pinin4fjords: **Severity: low** - doubled `\\t` for template-time Groovy interpolation.

```suggestion
    pd.DataFrame([selected]).to_csv(metrics_tsv, sep="\\t", index=False)
```

modules/nf-core/custom/clustermetrics/templates/cluster_metrics.py
```diff
@@ -0,0 +1,285 @@
...
+            )
+            plot_curve(
+                "calinski_harabasz",
+                "Calinski-Harabasz index (higher is better)",
+                "calinski_harabasz",
+                f"{pfx}_calinski.png",
+            )
+
+    except Exception as e:
+        Path("plot_warning.txt").write_text(f"Plotting failed: {e}\n")
```
pinin4fjords: **Severity: low** - drop the `try/except + plot_warning.txt` swallow around plotting. matplotlib is a hard dependency in `environment.yml`; if it fails to import, something is wrong and the run should fail loudly, not silently emit a warning file. Just call `plot_curve(...)` directly.

```suggestion
    import matplotlib.pyplot as plt

    def plot_curve(metric, title, ylabel, out_png):
        plt.figure(figsize=(7, 4.5))
        vals = sweep_df[metric].dropna()
        ks = sweep_df.loc[vals.index, "k"]
        plt.plot(ks, vals, marker="o")
        plt.xticks(sweep_df["k"].tolist())
        plt.title(title)
        plt.xlabel("k")
        plt.ylabel(ylabel)
        plt.tight_layout()
        plt.savefig(out_png, dpi=200)
        plt.close()

    if not sweep_df.empty:
        plot_curve("inertia", "Elbow method (KMeans inertia)", "inertia", f"{pfx}_elbow.png")
        plot_curve("silhouette", "Silhouette score (higher is better)", "silhouette", f"{pfx}_silhouette.png")
        plot_curve("davies_bouldin", "Davies-Bouldin index (lower is better)", "davies_bouldin", f"{pfx}_davies_bouldin.png")
        plot_curve("calinski_harabasz", "Calinski-Harabasz index (higher is better)", "calinski_harabasz", f"{pfx}_calinski.png")
```

modules/nf-core/custom/clustermetrics/tests/main.nf.test
```diff
@@ -0,0 +1,46 @@
...
+            }
+        }
+        then {
+            assert process.success
+            assert snapshot(
+                process.out.metrics,
+                process.out.k_sweep,
+                process.out.selected,
+                process.out.versions
+            ).match()
```
pinin4fjords: **Severity: low** - consider asserting `process.out.plots` in the snapshot too. The non-stub test currently does not capture the PNGs.

modules/nf-core/custom/clustervisualiation/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM nfcore/base:2.0
```
pinin4fjords: **Severity: high** - delete this file too, same reason as `clustermetrics/Dockerfile`.

modules/nf-core/custom/clustervisualiation/main.nf
```diff
@@ -0,0 +1,57 @@
...
+
+/*
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    PROCESS: CLUSTER_VIZ
+    Generates PCA, UMAP and t-SNE visualizations colored by cluster
+    Author: Donald Baku (author)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+*/
+
+process CLUSTER_VIZ {
```
pinin4fjords: **Severity: low/medium** - same as on `clustermetrics/main.nf`: drop the shebang/dsl/banner, and rename the process to match the `CUSTOM_<NAME>` convention. (Also worth renaming the directory from `clustervisualiation` -> `clustervisualization` to fix the typo, which then also makes this `CUSTOM_CLUSTERVISUALIZATION`.)

```suggestion
process CUSTOM_CLUSTERVISUALIATION {
```

modules/nf-core/custom/clustervisualiation/main.nf
```diff
@@ -0,0 +1,57 @@
...
+    tag "$meta.id"
+    label 'process_medium'
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+    'https://depot.galaxyproject.org/singularity/clustervisualiation:dev' :
+    'quay.io/nf-core/clustervisualiation:dev' }"
```
pinin4fjords: **Severity: high** - same fix as `clustermetrics/main.nf`. URLs are different here because the env includes `umap-learn`.

```suggestion
    container "${ workflow.containerEngine in ['singularity', 'apptainer'] && !task.ext.singularity_pull_docker_container ?
        'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/6e/6e7094e06509a04d66e8184bca83662f3867c9e3ea5675defae102f26c69446d/data' :
        'community.wave.seqera.io/library/matplotlib_pandas_python_scikit-learn_umap-learn:0f898dcbb2ca6fc8' }"
```

modules/nf-core/custom/clustervisualiation/main.nf
```diff
@@ -0,0 +1,57 @@
...
+    input:
+    tuple val(meta), path(features), path(clusters), path(pca_scores)
+    val out_prefix
```
pinin4fjords: **Severity: medium** - drop `val out_prefix`, same as for `clustermetrics`.

```suggestion
    input:
    tuple val(meta), path(features), path(clusters), path(pca_scores)
```

modules/nf-core/custom/clustervisualiation/main.nf
```diff
@@ -0,0 +1,57 @@
...
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        python: \$(python3 --version | cut -d' ' -f2)
+        pandas: \$(python3 -c "import pandas; print(pandas.__version__)")
+        scikit-learn: \$(python3 -c "import sklearn; print(sklearn.__version__)")
+        umap-learn: \$(python3 -c "import umap; print(umap.__version__)" 2>/dev/null || echo 'N/A')
+        matplotlib: \$(python3 -c "import matplotlib; print(matplotlib.__version__)")
+    END_VERSIONS
+    """
+}
```
pinin4fjords: **Severity: high** - same `template` + `stub` rewrite. Note the stub here doesn't `touch` the PCA outputs if you go with the architectural recommendation in the body to drop the PCA plotting; adjust accordingly.

```suggestion
    script:
    template 'cluster_viz.py'

    stub:
    def prefix = task.ext.prefix ?: "${meta.id}"
    """
    touch ${prefix}_umap.tsv
    touch ${prefix}_tsne.tsv
    touch ${prefix}_umap.png
    touch ${prefix}_tsne.png
    touch ${prefix}_pca.png

    cat <<-END_VERSIONS > versions.yml
    "${task.process}":
        python: \$(python3 --version | sed 's/Python //')
        pandas: \$(python3 -c "import pandas; print(pandas.__version__)")
        scikit-learn: \$(python3 -c "import sklearn; print(sklearn.__version__)")
        umap-learn: \$(python3 -c "import umap; print(umap.__version__)")
        matplotlib: \$(python3 -c "import matplotlib; print(matplotlib.__version__)")
    END_VERSIONS
    """
}
```

modules/nf-core/custom/clustervisualiation/meta.yml
```diff
@@ -0,0 +1,112 @@
+name: "CLUSTER_VIZ"
```
pinin4fjords: **Severity: low** - same naming convention as `clustermetrics`.

```suggestion
name: "custom_clustervisualiation"
```

modules/nf-core/custom/clustervisualiation/templates/cluster_viz.py
```diff
@@ -0,0 +1,222 @@
...
+
+    raise ValueError(f"Cannot find sample ID column (expected 'sample_id' or 'IID'). Found: {list(df.columns)}")
+
+
+def load_features(path: str) -> tuple[pd.DataFrame, pd.Series]:
+    df = pd.read_csv(path, sep=r"\s+", engine="python", dtype=str)
+    df = _normalise_id_column(df)
+    sample_ids = df["sample_id"].astype(str)
+    x = df.drop(columns=["sample_id"]).apply(pd.to_numeric, errors="coerce").fillna(0.0)
+    return x, sample_ids
```
pinin4fjords: **Severity: medium** - same simplification as in `cluster_metrics.py`: drop `_normalise_id_column`, require `sample_id`. Note this also fixes the `r"\s+"` Groovy interpolation issue on L63 (that becomes `r" +"` after template processing if left as-is, breaking the regex).

```suggestion
def load_features(path: str) -> tuple[pd.DataFrame, pd.Series]:
    df = pd.read_csv(path, sep="\\t")
    if "sample_id" not in df.columns:
        raise ValueError(f"features file must have a 'sample_id' column. Found: {list(df.columns)}")
    sample_ids = df["sample_id"].astype(str)
    x = df.drop(columns=["sample_id"]).apply(pd.to_numeric, errors="coerce").fillna(0.0)
    return x, sample_ids
```

modules/nf-core/custom/clustervisualiation/templates/cluster_viz.py
```diff
@@ -0,0 +1,222 @@
...
+
+
+def load_clusters(path: str) -> pd.Series:
+    df = pd.read_csv(path, sep=",", dtype=str)
+    df = _normalise_id_column(df)
+    if "cluster" not in df.columns:
+        raise ValueError("clusters CSV must have a 'cluster' column")
+    return df.set_index("sample_id")["cluster"].astype(int)
```
pinin4fjords: **Severity: medium** - simpler version that just requires `sample_id` and `cluster`.

```suggestion
def load_clusters(path: str) -> pd.Series:
    df = pd.read_csv(path)
    if "sample_id" not in df.columns or "cluster" not in df.columns:
        raise ValueError(f"clusters file must have 'sample_id' and 'cluster' columns. Found: {list(df.columns)}")
    return df.set_index(df["sample_id"].astype(str))["cluster"].astype(int)
```

modules/nf-core/custom/clustervisualiation/templates/cluster_viz.py
```diff
@@ -0,0 +1,222 @@
...
+        ).fit_transform(x)
+    except Exception as e:
+        print(f"[WARN] UMAP failed, fallback to first 2 feature columns: {e}")
+        if x.shape[1] >= 2:
+            return x[:, :2]
+        elif x.shape[1] == 1:
+            return np.column_stack([x[:, 0], np.zeros(x.shape[0])])
+        else:
+            return np.zeros((x.shape[0], 2))
+
```
pinin4fjords: **Severity: medium** - drop the silent fallback. `umap-learn` is a hard dependency in `environment.yml`; if `import umap` fails or `fit_transform` raises, the run should fail loudly, not silently use the first 2 feature columns as a pretend "embedding".

```suggestion
def compute_umap(x: np.ndarray, n_neighbors: int, min_dist: float) -> np.ndarray:
    import umap
    return umap.UMAP(
        n_components=2,
        n_neighbors=n_neighbors,
        min_dist=min_dist,
        random_state=42,
    ).fit_transform(x)
```

modules/nf-core/custom/clustervisualiation/templates/cluster_viz.py
```diff
@@ -0,0 +1,222 @@
...
+    pca_df = pd.read_csv(args.pca_scores, sep=r"\s+", engine="python", dtype=str)
+    pca_df = _normalise_id_column(pca_df)
+    comp_cols = [c for c in pca_df.columns if c != "sample_id"]
+    if len(comp_cols) < 2:
+        raise ValueError("pca_scores must have at least 2 PC columns")
+    c1, c2 = comp_cols[0], comp_cols[1]
+    for col in [c1, c2]:
+        pca_df[col] = pd.to_numeric(pca_df[col], errors="coerce")
+    merged = pca_df.merge(umap_df[["sample_id", "cluster"]], on="sample_id", how="inner")
+    plot_scatter(merged, c1, c2, args.out_pca_png, "PCA", c1, c2)
```
pinin4fjords: **Severity: medium** (architectural) - this PCA plotting block is essentially re-plotting upstream PLINK eigenvec output (it just reads the eigenvec, picks the first 2 PCs, and scatters them). Suggest removing this block from the module entirely; the consuming pipeline can render its own PCA plot from the eigenvec it already has, with one matplotlib call. That also drops the `pca_scores` input from the process, simplifies the meta.yml outputs (no `pca_png`), and removes the second `r"\s+"` Groovy gotcha on L209.

If you keep it, both `r"\s+"` instances need to become `r"\\s+"` to survive Groovy interpolation, and the `_normalise_id_column` call here re-introduces the PLINK-format coupling we are trying to drop in `load_features`.

```suggestion
```

modules/nf-core/custom/clustervisualiation/templates/cluster_viz.py
```diff
@@ -0,0 +1,222 @@
...
+    ap.add_argument("--tsne-perplexity", type=float, default=30.0)
+    ap.add_argument("--tsne-iter", type=int, default=1000)
+    ap.add_argument("--umap-neighbors", type=int, default=15)
+    ap.add_argument("--umap-min-dist", type=float, default=0.1)
+    ap.add_argument("--out-umap-tsv", required=True)
+    ap.add_argument("--out-tsne-tsv", required=True)
+    ap.add_argument("--out-umap-png", required=True)
+    ap.add_argument("--out-tsne-png", required=True)
+    ap.add_argument("--out-pca-png", required=True)
+    args = ap.parse_args()
```
pinin4fjords: **Severity: high** (paired with the `template` switch) - same as `cluster_metrics.py`: replace argparse with substitutions at the top of `main()`, then update references through the rest of the function. Sketch:

```python
def main() -> None:
    features = "${features}"
    clusters_path = "${clusters}"
    pca_scores = "${pca_scores}"
    prefix = "${task.ext.prefix}" if "${task.ext.prefix}" != "null" else "${meta.id}"
    tsne_perplexity = 30.0
    tsne_iter = 1000
    umap_neighbors = 15
    umap_min_dist = 0.1
    # ... rest of main(), with args.X replaced by the locals above and args.out_X replaced by f-strings using prefix
```

And write `versions.yml` from Python at the end of `main()`, like `custom/tx2gene/templates/tx2gene.py`.

modules/nf-core/custom/clustervisualiation/tests/main.nf.test
```diff
@@ -0,0 +1,54 @@
...
+            assert process.success
+            assert snapshot(
+                process.out.umap,
+                process.out.tsne,
+                process.out.umap_png,
+                process.out.tsne_png,
+                process.out.pca_png,
+                process.out.versions
+            ).match()
+        }
```
pinin4fjords: **Severity: low** - if you adopt the architectural recommendation to drop PCA plotting from this module, the test inputs (and `test_pca.eigenvec` data file) can drop the `pca_scores` path too, which simplifies things further.

modules/nf-core/custom/clustermetrics/templates/cluster_metrics.py
```diff
@@ -0,0 +1,314 @@
...
+    import sklearn
+
+    with open("versions.yml", "w") as f:
+        f.write(
+            f'"${task.process}":\n'
+            f'    python: {platform.python_version()}\n'
+            f'    pandas: {pandas.__version__}\n'
+            f'    scikit-learn: {sklearn.__version__}\n'
+            f'    matplotlib: {matplotlib.__version__}\n'
+        )
```
pinin4fjords: **Severity: high (CI-blocking)** - `f'"${task.process}":\n'` triggers ruff F821 (Python's f-string parser sees `{task.process}` as an expression, ruff resolves `task` as undefined) AND the `\n` becomes a literal newline after Groovy template interpolation, which would syntax-error the f-string at runtime. The fix is to switch to a regular string + `format_yaml_like` helper, the same pattern `custom/tx2gene/templates/tx2gene.py` uses. The dict key `"${task.process}"` is a regular string literal so it stays opaque to both ruff and Python.

```suggestion
    versions = {
        "${task.process}": {
            "python": platform.python_version(),
            "pandas": pandas.__version__,
            "scikit-learn": sklearn.__version__,
            "matplotlib": matplotlib.__version__,
        }
    }

    def format_yaml_like(data, indent=0):
        out = ""
        for key, value in data.items():
            spaces = "  " * indent
            if isinstance(value, dict):
                out += f"{spaces}{key}:\\n{format_yaml_like(value, indent + 1)}"
            else:
                out += f"{spaces}{key}: {value}\\n"
        return out

    with open("versions.yml", "w") as f:
        f.write(format_yaml_like(versions))
```

modules/nf-core/custom/clustermetrics/templates/cluster_metrics.py
```diff
@@ -0,0 +1,314 @@
...
+            x = x_df.values
+            labels = clusters_df["cluster"].values
+            aligned_ids = sample_ids.astype(str).tolist()
+            alignment_mode = "row_order_fallback"
+        else:
+            raise ValueError(
+                f"No overlapping sample_id between features and clusters.\n"
+                f"  features IDs (first 5): {sample_ids.head().tolist()}\n"
+                f"  clusters IDs (first 5): {list(clusters.index[:5])}"
+            )
```
pinin4fjords: **Severity: high (CI-blocking)** - `\n` inside an f-string will become a literal newline after Groovy template interpolation, breaking the f-string at runtime (newline inside an unclosed quote = `SyntaxError`). Double the backslash so it survives:

```suggestion
            raise ValueError(
                f"No overlapping sample_id between features and clusters.\\n"
                f"  features IDs (first 5): {sample_ids.head().tolist()}\\n"
                f"  clusters IDs (first 5): {list(clusters.index[:5])}"
            )
```

modules/nf-core/custom/clustermetrics/templates/cluster_metrics.py
```diff
@@ -0,0 +1,314 @@
...
+    else:
+        if len(clusters_df) != len(sample_ids):
+            raise ValueError(
+                "clusters CSV has no usable sample_id column and row counts do not match.\n"
+                f"  n_features={len(sample_ids)}\n"
+                f"  n_clusters={len(clusters_df)}"
+            )
```
pinin4fjords: **Severity: high (CI-blocking)** - same `\n`-in-f-string issue:

```suggestion
        if len(clusters_df) != len(sample_ids):
            raise ValueError(
                "clusters CSV has no usable sample_id column and row counts do not match.\\n"
                f"  n_features={len(sample_ids)}\\n"
                f"  n_clusters={len(clusters_df)}"
            )
```

modules/nf-core/custom/clustermetrics/tests/main.nf.test
```diff
@@ -0,0 +1,46 @@
+nextflow_process {
+    name "Test Process CLUSTER_METRICS"
+    script "../main.nf"
+    process "CLUSTER_METRICS"
```
pinin4fjords: **Severity: high (CI-blocking)** - process name was renamed in `main.nf` but this test still references the old name, so nf-test won't find the process.

```suggestion
    name "Test Process CUSTOM_CLUSTERMETRICS"
    script "../main.nf"
    process "CUSTOM_CLUSTERMETRICS"
```

modules/nf-core/custom/clustermetrics/tests/main.nf.test
```diff
@@ -0,0 +1,46 @@
...
+    tag "custom/clustermetrics"
+    tag "clustermetrics"
+
+    test("clustermetrics - features and clusters") {
+        when {
+            process {
+                """
+                input[0] = [ [id:'test'], file("${projectDir}/modules/nf-core/custom/clustermetrics/tests/data/test_features.tsv", checkIfExists: true), file("${projectDir}/modules/nf-core/custom/clustermetrics/tests/data/test_clusters.csv", checkIfExists: true) ]
+                input[1] = 'test'
+                """
```
pinin4fjords, line 22: **Severity: high (CI-blocking)** - the process now only takes one input (you dropped `val out_prefix`), so this `input[1]` line needs to go.

```suggestion
            process {
                """
                input[0] = [ [id:'test'], file("${projectDir}/modules/nf-core/custom/clustermetrics/tests/data/test_features.tsv", checkIfExists: true), file("${projectDir}/modules/nf-core/custom/clustermetrics/tests/data/test_clusters.csv", checkIfExists: true) ]
                """
            }
```

modules/nf-core/custom/clustermetrics/tests/main.nf.test
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    test("clustermetrics - features and clusters - stub") {
+        options "-stub"
+        when {
+            process {
+                """
+                input[0] = [ [id:'test'], file("${projectDir}/modules/nf-core/custom/clustermetrics/tests/data/test_features.tsv", checkIfExists: true), file("${projectDir}/modules/nf-core/custom/clustermetrics/tests/data/test_clusters.csv", checkIfExists: true) ]
+                input[1] = 'test'
+                """
```
pinin4fjords, line 52: **Severity: high (CI-blocking)** - same fix for the stub test (drop `input[1]`).

```suggestion
            process {
                """
                input[0] = [ [id:'test'], file("${projectDir}/modules/nf-core/custom/clustermetrics/tests/data/test_features.tsv", checkIfExists: true), file("${projectDir}/modules/nf-core/custom/clustermetrics/tests/data/test_clusters.csv", checkIfExists: true) ]
                """
            }
```

modules/nf-core/custom/clustervisualiation/main.nf
```diff
@@ -0,0 +1,26 @@
+process CUSTOM_CLUSTERVISUALIATION {
+    tag "$meta.id"
+    label 'process_medium'
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine in ['singularity', 'apptainer'] && !task.ext.singularity_pull_docker_container ?
+    'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/c0/c00b83d40a02e4ed2833ebf0d38635602231a21764eff0d30ed16885e5c02445/data' :
+    'community.wave.seqera.io/library/matplotlib_pandas_python_scikit-learn_umap-learn:2c4aaf377be5cd4a' }"
+
+    input:
+    tuple val(meta), path(features), path(clusters), path(pca_scores)
```
pinin4fjords: **Severity: high (CI-blocking)** - `cluster_viz.py main()` no longer produces a PCA plot, so the `pca_scores` input is unused.

```suggestion
    tuple val(meta), path(features), path(clusters)
```

modules/nf-core/custom/clustervisualiation/main.nf
```diff
@@ -0,0 +1,26 @@
...
+
+    output:
+    tuple val(meta), path("*_umap.tsv")   , emit: umap
+    tuple val(meta), path("*_tsne.tsv")   , emit: tsne
+    tuple val(meta), path("*_umap.png")   , emit: umap_png
+    tuple val(meta), path("*_tsne.png")   , emit: tsne_png
+    tuple val(meta), path("*_pca.png")    , emit: pca_png
```
pinin4fjords: **Severity: high (CI-blocking)** - `*_pca.png` is never created since the PCA plotting was removed from `main()`, so this emit will fail nf-test (no matching file).

```suggestion
```

modules/nf-core/custom/clustervisualiation/meta.yml
```diff
@@ -0,0 +1,109 @@
...
+        type: file
+        description: CSV/TSV file with sample_id and cluster assignment
+        pattern: "*_clusters.*"
+        ontologies: []
+    - pca_scores:
+        type: file
+        description: TSV file with PCA scores from previous step
+        pattern: "*_pca_scores.tsv"
+        ontologies:
+          - edam: http://edamontology.org/format_3475
```
pinin4fjords: **Severity: medium (CI-blocking once `main.nf` input is dropped)** - drop the `pca_scores` input doc to match `main.nf`.

```suggestion
```

modules/nf-core/custom/clustervisualiation/meta.yml
```diff
@@ -0,0 +1,109 @@
...
+          ontologies: []
+  pca_png:
+    - - meta:
+          type: map
+          description: Groovy Map containing sample information
+      - "*_pca.png":
+          type: file
+          description: PCA visualization plot
+          pattern: "*_pca.png"
+          ontologies: []
```
pinin4fjords: **Severity: medium (CI-blocking once `main.nf` emit is dropped)** - drop the `pca_png` output doc to match `main.nf`.

```suggestion
```

pinin4fjords (commented):
Hi @dbaku42, nice progress. AI-assisted follow-up review (Claude, on behalf of @pinin4fjords).

Most of what's left is mechanical, with two patterns to know:

**1. The version-write block in both wrappers is what's failing pre-commit (ruff F821) and will also `SyntaxError` when staged through the `template` directive.** Both `cluster_metrics.py` and `cluster_viz.py` wrap the version write in an f-string with `${task.process}` inside it. Two issues stack: ruff sees `{task.process}` as a Python expression and flags `task` as undefined, AND Groovy converts `\n` to a real newline at template time, breaking the f-string. The clean fix is to follow `tx2gene.py`'s pattern: a regular dict + a tiny `format_yaml_like` helper. Suggestions inline.

**2. Cluster_viz still has orphan PCA pieces.** You dropped the PCA plotting from `main()`, but `main.nf` still declares a `pca_scores` input and a `pca_png` emit, and `meta.yml` still documents both. nf-test will fail because the `pca_png` emit is never produced. Suggestions inline to drop the orphan pieces; tests need the matching trim.

**Also still pending from the previous review** (just need to be applied, no new suggestions): delete both `Dockerfile`s, add a `stub:` block to `clustervisualiation/main.nf`, and the optional Python simplifications I suggested for the PLINK-aware `_normalise_id_column` and `load_clusters` complexity.

Snapshots will need regenerating (`nf-test --update-snapshot`) after these changes land.

modules/nf-core/custom/clustervisualiation/templates/cluster_viz.py
```diff
@@ -0,0 +1,237 @@
...
+        fid_cols = [c for c in df.columns if c.upper() == "FID"]
+        if fid_cols:
+            df = df.drop(columns=fid_cols)
+        return df
+
+    raise ValueError(f"Cannot find sample ID column (expected 'sample_id' or 'IID'). Found: {list(df.columns)}")
+
+
+def load_features(path: str) -> tuple[pd.DataFrame, pd.Series]:
+    df = pd.read_csv(path, sep=r"\s+", engine="python", dtype=str)
```
pinin4fjords: **Severity: high (CI-blocking)** - `r"\s+"` becomes `r" +"` after Groovy template interpolation (`\s` is unrecognised so Groovy strips the backslash; verified locally), so the regex matches literal space, not whitespace. Since the input is documented as TSV in `meta.yml`, the simpler fix is `sep="\\t"`.

```suggestion
    df = pd.read_csv(path, sep="\\t", dtype=str)
```

modules/nf-core/custom/clustervisualiation/templates/cluster_viz.py
```diff
@@ -0,0 +1,237 @@
...
+    plt.tight_layout()
+    plt.savefig(out_png, dpi=200, bbox_inches="tight")
+    plt.close()
+
+
+def main() -> None:
+    ap = argparse.ArgumentParser(description="PCA + UMAP + t-SNE plots colored by cluster")
+    ap.add_argument("--features", required=True)
+    ap.add_argument("--clusters", required=True)
+    ap.add_argument("--pca-scores", required=True)
```
pinin4fjords: **Severity: medium (paired with PCA drop)** - drop the `--pca-scores` argparse option since the script no longer uses it.

```suggestion
```

modules/nf-core/custom/clustervisualiation/templates/cluster_viz.py
```diff
@@ -0,0 +1,237 @@
...
+    ap.add_argument("--tsne-perplexity", type=float, default=30.0)
+    ap.add_argument("--tsne-iter", type=int, default=1000)
+    ap.add_argument("--umap-neighbors", type=int, default=15)
+    ap.add_argument("--umap-min-dist", type=float, default=0.1)
+    ap.add_argument("--out-umap-tsv", required=True)
+    ap.add_argument("--out-tsne-tsv", required=True)
+    ap.add_argument("--out-umap-png", required=True)
+    ap.add_argument("--out-tsne-png", required=True)
+    ap.add_argument("--out-pca-png", required=True)
```
pinin4fjords: **Severity: medium (paired with PCA drop)** - drop the `--out-pca-png` argparse option.

```suggestion
```

modules/nf-core/custom/clustervisualiation/templates/cluster_viz.py
```diff
@@ -0,0 +1,237 @@
...
+    x_df, sample_ids = load_features(args.features)
+    clusters = load_clusters(args.clusters)
+
+    common = sample_ids[sample_ids.isin(clusters.index)]
+    if len(common) == 0:
+        raise ValueError(
+            f"No overlapping sample_id between features and clusters.\n"
+            f"  features IDs (first 5): {sample_ids.head().tolist()}\n"
+            f"  clusters IDs (first 5): {list(clusters.index[:5])}"
+        )
```
pinin4fjords: **Severity: high (CI-blocking)** - same `\n`-in-f-string issue as in `cluster_metrics.py`:

```suggestion
        raise ValueError(
            f"No overlapping sample_id between features and clusters.\\n"
            f"  features IDs (first 5): {sample_ids.head().tolist()}\\n"
            f"  clusters IDs (first 5): {list(clusters.index[:5])}"
        )
```

modules/nf-core/custom/clustervisualiation/templates/cluster_viz.py
```diff
@@ -0,0 +1,237 @@
...
+    import sys
+    import platform
+
+    prefix = "${task.ext.prefix ?: meta.id}"
+
+    sys.argv = [
+        "cluster_viz.py",
+        "--features", "$features",
+        "--clusters", "$clusters",
+        "--pca-scores", "$pca_scores",
```
pinin4fjords: **Severity: medium (paired with PCA drop)** - drop the `--pca-scores` entry in `sys.argv`.

```suggestion
```

modules/nf-core/custom/clustervisualiation/templates/cluster_viz.py
```diff
@@ -0,0 +1,237 @@
...
+        "--out-umap-tsv", f"{prefix}_umap.tsv",
+        "--out-tsne-tsv", f"{prefix}_tsne.tsv",
+        "--out-umap-png", f"{prefix}_umap.png",
+        "--out-tsne-png", f"{prefix}_tsne.png",
+        "--out-pca-png", f"{prefix}_pca.png",
```
pinin4fjords: **Severity: medium (paired with PCA drop)** - drop the `--out-pca-png` entry in `sys.argv`.

```suggestion
```

modules/nf-core/custom/clustervisualiation/templates/cluster_viz.py
```diff
@@ -0,0 +1,237 @@
...
+
+    with open("versions.yml", "w") as f:
+        f.write(
+            f'"${task.process}":\n'
+            f'    python: {platform.python_version()}\n'
+            f'    pandas: {pandas.__version__}\n'
+            f'    scikit-learn: {sklearn.__version__}\n'
+            f'    umap-learn: {umap_version}\n'
+            f'    matplotlib: {matplotlib.__version__}\n'
+        )
```
pinin4fjords: **Severity: high (CI-blocking)** - same fix as on `cluster_metrics.py`: switch the version-write block to the `tx2gene` `format_yaml_like` pattern. Resolves the ruff F821 and the `\n` rendering issue:

```suggestion
    versions = {
        "${task.process}": {
            "python": platform.python_version(),
            "pandas": pandas.__version__,
            "scikit-learn": sklearn.__version__,
            "umap-learn": umap_version,
            "matplotlib": matplotlib.__version__,
        }
    }

    def format_yaml_like(data, indent=0):
        out = ""
        for key, value in data.items():
            spaces = "  " * indent
            if isinstance(value, dict):
                out += f"{spaces}{key}:\\n{format_yaml_like(value, indent + 1)}"
            else:
                out += f"{spaces}{key}: {value}\\n"
        return out

    with open("versions.yml", "w") as f:
        f.write(format_yaml_like(versions))
```

modules/nf-core/custom/clustervisualiation/tests/main.nf.test
```diff
@@ -0,0 +1,54 @@
...
+    test("clustervisualiation - features clusters pca") {
+        when {
+            process {
+                """
+                input[0] = [ [id:'test'],
+                    file("${projectDir}/modules/nf-core/custom/clustervisualiation/tests/data/test_features.tsv", checkIfExists: true),
+                    file("${projectDir}/modules/nf-core/custom/clustervisualiation/tests/data/test_clusters.csv", checkIfExists: true),
+                    file("${projectDir}/modules/nf-core/custom/clustervisualiation/tests/data/test_pca.eigenvec", checkIfExists: true) ]
+                input[1] = 'test'
+                """
```
pinin4fjords: **Severity: high (CI-blocking)** - drop `input[1] = 'test'` (no second input now), and drop the `pca` file from `input[0]` (no longer needed once `pca_scores` is removed from `main.nf`).

```suggestion
                input[0] = [ [id:'test'],
                    file("${projectDir}/modules/nf-core/custom/clustervisualiation/tests/data/test_features.tsv", checkIfExists: true),
                    file("${projectDir}/modules/nf-core/custom/clustervisualiation/tests/data/test_clusters.csv", checkIfExists: true) ]
```

modules/nf-core/custom/clustervisualization/tests/main.nf.test
```diff
@@ -0,0 +1,54 @@
...
+        options "-stub"
+        when {
+            process {
+                """
+                input[0] = [ [id:'test'],
+                    file("${projectDir}/modules/nf-core/custom/clustervisualiation/tests/data/test_features.tsv", checkIfExists: true),
+                    file("${projectDir}/modules/nf-core/custom/clustervisualiation/tests/data/test_clusters.csv", checkIfExists: true),
+                    file("${projectDir}/modules/nf-core/custom/clustervisualiation/tests/data/test_pca.eigenvec", checkIfExists: true) ]
+                input[1] = 'test'
+                """
```
pinin4fjords: **Severity: high (CI-blocking)** - same fix for the stub test.

```suggestion
                input[0] = [ [id:'test'],
                    file("${projectDir}/modules/nf-core/custom/clustervisualiation/tests/data/test_features.tsv", checkIfExists: true),
                    file("${projectDir}/modules/nf-core/custom/clustervisualiation/tests/data/test_clusters.csv", checkIfExists: true) ]
```

modules/nf-core/custom/clustermetrics/tests/main.nf.test
pinin4fjords, line 52: **Severity: high (CI-blocking)** - same fix for the stub test (drop `input[1]`).

```suggestion
            process {
                """
                input[0] = [ [id:'test'], file("${projectDir}/modules/nf-core/custom/clustermetrics/tests/data/test_features.tsv", checkIfExists: true), file("${projectDir}/modules/nf-core/custom/clustermetrics/tests/data/test_clusters.csv", checkIfExists: true) ]
                """
            }
```

modules/nf-core/custom/clustervisualiation/main.nf
pinin4fjords: **Severity: high (CI-blocking)** - `cluster_viz.py main()` no longer produces a PCA plot, so the `pca_scores` input is unused.

```suggestion
    tuple val(meta), path(features), path(clusters)
```

modules/nf-core/custom/clustervisualiation/main.nf
pinin4fjords: **Severity: high (CI-blocking)** - `*_pca.png` is never created since the PCA plotting was removed from `main()`, so this emit will fail nf-test (no matching file).

```suggestion
```

modules/nf-core/custom/clustervisualiation/meta.yml
pinin4fjords: **Severity: medium (CI-blocking once `main.nf` input is dropped)** - drop the `pca_scores` input doc to match `main.nf`.

```suggestion
```

modules/nf-core/custom/clustervisualiation/meta.yml
pinin4fjords: **Severity: medium (CI-blocking once `main.nf` emit is dropped)** - drop the `pca_png` output doc to match `main.nf`.

```suggestion
```

modules/nf-core/custom/clustervisualiation/templates/cluster_viz.py
pinin4fjords: **Severity: high (CI-blocking)** - `r"\s+"` becomes `r" +"` after Groovy template interpolation (`\s` is unrecognised so Groovy strips the backslash; verified locally), so the regex matches literal space, not whitespace. Since the input is documented as TSV in `meta.yml`, the simpler fix is `sep="\\t"`.

```suggestion
    df = pd.read_csv(path, sep="\\t", dtype=str)
```

modules/nf-core/custom/clustervisualiation/templates/cluster_viz.py
pinin4fjords: **Severity: medium (paired with PCA drop)** - drop the `--pca-scores` argparse option since the script no longer uses it.

```suggestion
```

modules/nf-core/custom/clustervisualiation/templates/cluster_viz.py
pinin4fjords: **Severity: medium (paired with PCA drop)** - drop the `--out-pca-png` argparse option.

```suggestion
```

modules/nf-core/custom/clustervisualiation/templates/cluster_viz.py
pinin4fjords: **Severity: high (CI-blocking)** - same `\n`-in-f-string issue as in `cluster_metrics.py`:

```suggestion
        raise ValueError(
            f"No overlapping sample_id between features and clusters.\\n"
            f"  features IDs (first 5): {sample_ids.head().tolist()}\\n"
            f"  clusters IDs (first 5): {list(clusters.index[:5])}"
        )
```

modules/nf-core/custom/clustervisualiation/templates/cluster_viz.py
pinin4fjords: **Severity: medium (paired with PCA drop)** - drop the `--pca-scores` entry in `sys.argv`.

```suggestion
```

modules/nf-core/custom/clustervisualiation/templates/cluster_viz.py
pinin4fjords: **Severity: medium (paired with PCA drop)** - drop the `--out-pca-png` entry in `sys.argv`.

```suggestion
```

modules/nf-core/custom/clustervisualiation/templates/cluster_viz.py
pinin4fjords: **Severity: high (CI-blocking)** - same fix as on `cluster_metrics.py`: switch the version-write block to the `tx2gene` `format_yaml_like` pattern. Resolves the ruff F821 and the `\n` rendering issue:

```suggestion
    versions = {
        "${task.process}": {
            "python": platform.python_version(),
            "pandas": pandas.__version__,
            "scikit-learn": sklearn.__version__,
            "umap-learn": umap_version,
            "matplotlib": matplotlib.__version__,
        }
    }

    def format_yaml_like(data, indent=0):
        out = ""
        for key, value in data.items():
            spaces = "  " * indent
            if isinstance(value, dict):
                out += f"{spaces}{key}:\\n{format_yaml_like(value, indent + 1)}"
            else:
                out += f"{spaces}{key}: {value}\\n"
        return out

    with open("versions.yml", "w") as f:
        f.write(format_yaml_like(versions))
```

modules/nf-core/custom/clustervisualiation/tests/main.nf.test
pinin4fjords: **Severity: high (CI-blocking)** - drop `input[1] = 'test'` (no second input now), and drop the `pca` file from `input[0]` (no longer needed once `pca_scores` is removed from `main.nf`).

```suggestion
                input[0] = [ [id:'test'],
                    file("${projectDir}/modules/nf-core/custom/clustervisualiation/tests/data/test_features.tsv", checkIfExists: true),
                    file("${projectDir}/modules/nf-core/custom/clustervisualiation/tests/data/test_clusters.csv", checkIfExists: true) ]
```

modules/nf-core/custom/clustervisualization/tests/main.nf.test
pinin4fjords: **Severity: high (CI-blocking)** - same fix for the stub test.

```suggestion
                input[0] = [ [id:'test'],
                    file("${projectDir}/modules/nf-core/custom/clustervisualiation/tests/data/test_features.tsv", checkIfExists: true),
                    file("${projectDir}/modules/nf-core/custom/clustervisualiation/tests/data/test_clusters.csv", checkIfExists: true) ]
```

dbaku42:
Hi @pinin4fjords , I'm stuck on a snapshot mismatch issue and would appreciate some guidance. The tests for CUSTOM_CLUSTERMETRICS and CUSTOM_CLUSTERVISUALIZATION are failing because the output files (test_metrics.tsv, test_k_sweep.csv, test_selected.json) produce different MD5 hashes in CI compared to my local environment.

The root cause seems to be that my local snapshots were generated using the Wave container, while the CI runs tests using conda (as shown in the logs: Creating env using conda: .../environment.yml). The two environments resolve slightly different package versions, leading to numerically different outputs from scikit-learn/pandas.

Thanks in advance! 🙏

pinin4fjords (changes requested):
Hi @dbaku42, sorry for the volume of direct commits on your branch (AI-assisted by Claude on my behalf) - given the back-and-forth pattern of the last few rounds it felt like the fastest way to land the modules in a clean, spec-aligned state. Please review and push back on anything that doesn't sit right; happy to revert specific bits if you disagree with a choice. Highlights of what changed and why:

- **Containers rebuilt** on Seqera Containers via Wave to match the actual `environment.yml` for each module. nf-core uses [Biocontainers](https://biocontainers.pro/) or [Seqera Containers](https://seqera.io/containers/); for multi-package Python envs like these, Seqera Containers via the [`wave` CLI](https://github.com/seqeralabs/wave-cli) is the right call. Singularity branch is the `https://community-cr-prod.seqera.io/.../data` blob URL (not `docker://`).
- **`${...}` substitutions** in `clustering.py`: ruff lints the source as Python, so `KMeans(n_clusters=${n_clusters}, ...)` fails. Moved substitutions to locals at the top of `main()` and use the locals from then on.
- **`versions.yml`** writes via `yaml.dump` (with `pyyaml` added to the envs), since adding deps is cheap now that we're on Wave. Stub heredocs use `importlib.metadata.version(...)` to avoid import side effects (`import umap` triggers numba JIT cache writes that fail under read-only singularity).
- **Test portability**: KMeans inertia wobbles a few ULPs across CPUs (BLAS reduction order), so the clustering test rounds it in the assertion only. Added a DBSCAN test case so both algorithm branches are exercised.
- **`ext.args` plumbing** in `cluster_metrics.py` (`--k-min`, `--k-max`) and `cluster_viz.py` (`--umap-neighbors`, `--tsne-perplexity`), parsed via `argparse` + `shlex.split("$task.ext.args")` inside the template - matches the [spec](https://nf-co.re/docs/specifications/components/modules/general#3-optional-command-arguments) that optional non-file args go via `ext.args`. Same approach as `custom/matrixfilter` uses for R.
- **Output filenames** unified to `${prefix}.<descriptor>.<ext>` across all three modules (was mixed dot/underscore), matching the spec example and the existing `custom/tx2gene` / `custom/matrixfilter` pattern.
- **Simplifications**: removed the PLINK-aware `_normalise_id_column` / multi-mode `load_clusters` from `cluster_metrics.py` and `cluster_viz.py` (both already require `sample_id` + `cluster` and the `clustering` module emits exactly that). Replaced the positional-index alignment with `load_features().join(load_clusters(), how="inner")`. Replaced the 45-line manual eigenvec parser with `pd.read_csv`. Reverted a stray `.gitignore` change unrelated to this PR.

**One blocker before this can land**: `tests/data/` files can't live here. nf-core stores all module test data in [nf-core/test-datasets](https://github.com/nf-core/test-datasets) on the `modules` branch, referenced via `params.modules_testdata_base_path`. See `custom/basicpy` for an example. To unblock:

1. PR your five test files to the `modules` branch of `nf-core/test-datasets` under something like `genomics/popgen/clustering/`.
2. Once it merges, update each `tests/main.nf.test` to use `params.modules_testdata_base_path + ...` and delete the `tests/data/` directories.
3. Re-record snapshots and push.

Marking as Changes Requested for that, and resolving the previous threads since they're addressed by the recent commits. Apologies again for the heavy hand - shout if anything needs reverting.

Local workflow to catch the easy stuff before pushing in future:

```bash
prek run --all-files
nf-test test --profile docker --update-snapshot modules/nf-core/custom/clustering/tests/main.nf.test
nf-test test --profile docker --update-snapshot modules/nf-core/custom/clustermetrics/tests/main.nf.test
nf-test test --profile docker --update-snapshot modules/nf-core/custom/clustervisualization/tests/main.nf.test
```

dbaku42:
Hi @pinin4fjords,

thank you so much for all the work you did directly on the branch (and for the super clear explanations)!  
I've reviewed all the commits and they look excellent: the Wave/Seqera containers, correct use of the `template` directive, `task.ext.args`, unified file naming, `versions.yml` with `pyyaml` + `importlib.metadata`, the Singularity/Numba fixes, etc. I have no objections at all — on the contrary, thank you for speeding up the process so much.

Regarding the remaining blocker:
- I'm preparing right now a PR on **nf-core/test-datasets** (branch `modules`) to upload the 5 test files under `genomics/popgen/clustering/` (as you suggested).
- As soon as that PR is merged, I'll update the `tests/main.nf.test` files for both modules, delete the `tests/data/` folders, re-snapshot and push the changes here.

I'll let you know as soon as the test-datasets PR is opened (and I'll tag you).  
If you'd like me to use a slightly different path or rename anything in the test files, just let me know!

Thank you again for the huge help ❤️  
(By the way, the modules look really great now)

@nf-core/modules

dbaku42:
Hi @pinin4fjords,

I've just opened the PR for the test data on nf-core/test-datasets:  
https://github.com/nf-core/test-datasets/pull/2051

As soon as it's merged I'll update the two `main.nf.test` files, remove the local `tests/data/` folders, re-snapshot and push here.

Thanks again for all the help!

dbaku42:
Hi @pinin4fjords, 
I’ve now moved the test data to nf-core/test-datasets and updated the module tests to use params.modules_testdata_base_path.
I also re-recorded the snapshots as suggested. Most tests completed successfully, but unfortunately the snapshot update still fails on 3 jobs.
I’m currently investigating whether this is related to the execution environment or remaining module-specific issues.
The rest of the requested changes have been addressed and pushed.

pinin4fjords:
Hi @dbaku42 - I've pushed `f5a2ba7` to your branch with the corrected snapshot for `clustervisualization`. AI-assisted (Claude, on behalf of @pinin4fjords).

**What broke:** the md5s for `test.umap.tsv` / `test.tsne.tsv` you committed in `c840e845` / `05cbfc98` / `a2617a13` were generated on your local machine, but they don't match what the Linux x86 CI runners produce. UMAP and t-SNE outputs are not bit-identical across CPU architectures (Apple Silicon vs Linux x86, different BLAS implementations, numba JIT) even with `random_state=42` set. CI was deterministic across all three profiles (conda/docker/singularity all produced the same md5s) - the mismatch was purely between your machine and CI.

**Fix:** I re-ran `nf-test test modules/nf-core/custom/clustervisualization/tests/main.nf.test --profile docker --update-snapshot` on a Linux x86 box and committed the resulting snap. The diff is just the two md5s plus the timestamp/nextflow-version metadata.

**Going forward:** when nf-test snapshots fail in CI but pass locally, the snap usually needs regenerating in an environment that matches CI (Linux x86 + docker). Either run nf-test in a Linux VM/container, or wait for CI to run and copy the expected hashes out of the failure diff.

CI should go green now. Sorry for the direct push, but the fix was trivial.

pinin4fjords:
@dbaku42 - apologies for the late-breaking change: I've pushed `776f3d5` renaming `custom/clustering` → `custom/pcaclustering` (AI-assisted, Claude on behalf of @pinin4fjords).

**Reason this should have been caught earlier:** the module parses PLINK2 `.eigenvec` specifically and only supports KMeans/DBSCAN, so the original name claimed generality the module didn't have. We should have flagged this in one of the earlier review rounds rather than at the finish line - sorry it landed this late. `clustermetrics` and `clustervisualization` are genuinely generic (any features TSV + clusters CSV), so those names stay as-is.

**What's in the commit:**
- `git mv` of the module directory
- `CUSTOM_CLUSTERING` → `CUSTOM_PCACLUSTERING` in `main.nf`, `meta.yml`, and the test file
- `custom/clustering` tag → `custom/pcaclustering` in the test
- Snapshot regenerated (the versions.yml md5 changed because `${task.process}` is now `CUSTOM_PCACLUSTERING`)

I verified locally that the rename builds and the tests pass under docker; CI on the new head will confirm. Nothing else needed from you for the rename - if you disagree with the new name push back and we can revisit.

dbaku42:
Thanks for handling this @pinin4fjords (and for the clear explanation)! No objections on the rename — `pcaclustering` is definitely more accurate and specific than the original name. Appreciated!

pinin4fjords (approved):
Fine by me now

modules/nf-core/custom/clustermetrics/templates/cluster_metrics.py
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/env python3
+
```
SPPearce, line 2: I think we should have a license here.
https://nf-co.re/docs/specifications/pipelines/requirements/mit_license

modules/nf-core/custom/pcaclustering/meta.yml
```diff
@@ -0,0 +1,77 @@
+name: "CUSTOM_PCACLUSTERING"
+description: "Performs KMeans or DBSCAN clustering on principal components from PLINK2
```
SPPearce: Is this just plink2 output?

modules/nf-core/custom/pcaclustering/meta.yml
```diff
@@ -0,0 +1,77 @@
...
+      identifier: ""
+input:
+  - - meta:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. `[ id:'sample1' ]`
+    - eigenvec:
+        type: file
+        description: PLINK2 .eigenvec file generated by --pca
```
SPPearce: So only relevant for plink2?

SPPearce (commented):
I think we need a licence on all the python scripts.

modules/nf-core/custom/pcaclustering/meta.yml
dbaku42: Yes, currently the module expects the .eigenvec output generated by PLINK2 --pca, so the input format is PLINK2-specific.

modules/nf-core/custom/clustermetrics/templates/cluster_metrics.py
dbaku42: thanks @SPPearce,  I’ve added the required MIT license header (SPDX: MIT) to the custom script files

modules/nf-core/custom/pcaclustering/meta.yml
SPPearce: this isn't really pcaclustering then either, as I think it would be reasonable to expect that to work on a generic input file

dbaku42: I’ll rename it to plink2pcaclustering, which better reflects both the PLINK2 dependency and the PCA-based workflow, let me know if you have any suggestion regarding the name

SPPearce (commented):
You seem to still have pcaclustering as well as plink2pcaclustering.
Actually, I'm wondering if this should just go into plink2/pcaclustering

pinin4fjords:
@SPPearce @dbaku42 - while looking at the `pcaclustering` vs `plink2pcaclustering` duplication, I sketched an alternative: drop the PLINK-aware parser entirely and accept the same generic `sample_id` + numeric-features TSV that `clustermetrics` already takes. PLINK eigenvec then becomes a 1-line `awk` upstream.

Verified locally that the existing snapshot stays byte-identical (proof in the PR body): https://github.com/dbaku42/modules/pull/1

Targeted at @dbaku42's branch as a proposal, not a request - if you'd rather go the other way and nest under `plink2/pcaclustering`, happy to close it. Just wanted to put the diff on the table.

AI-assisted (Claude, on behalf of @pinin4fjords).

dbaku42:
Thanks @pinin4fjords! Keeping it general is  a better approach. I've already merged your proposal.

pinin4fjords:
> Thanks @pinin4fjords! Keeping it general is a better approach. I've already merged your proposal.

OK, lemme fix what's broken

pinin4fjords:
Pushed `50d2883f` to fix the CI failure on `custom/pcaclustering` - the test fixture in `nf-core/test-datasets` is named `test_features.tsv`, not `test.tsv`. My mistake in the earlier refactor PR (#11372 / dbaku42#1). One-line path fix on the three test cases. AI-assisted (Claude, on behalf of @pinin4fjords).

SPPearce (commented):
That seems much better.
Just to check, does it work if the first column (the sample id's) are all numeric?

pinin4fjords:
@SPPearce: computer says yes. Plain integer IDs work fine, but I pushed `f5edb634` to also preserve zero-padded ones (`0001` etc.) - pandas was type-inferring the first column as int and stripping the padding. Pinning `dtype={0: str}` at read time fixes it without changing snapshot md5s against the existing fixture.

dbaku42:
Thanks @pinin4fjords for your help! Regarding the test.tsv file, I already opened a PR https://github.com/nf-core/test-datasets/pull/2073

pinin4fjords:
Well, tests are passing now. FYI I'm about to go on leave. You have my approval, I'll let @SPPearce give his final approval when he's ready.

SPPearce (approved):

Issue #10832, opened 2026-03-13
Adopt topic channels in all nf-core modules [PART 4]


atrigila:
Parent issue and instructions here: https://github.com/nf-core/modules/issues/9978

- [ ] Migration to topic channels: eagle2  
- [ ] Migration to topic channels: ectyper  
- [ ] Migration to topic channels: eigenstratdatabasetools/eigenstratsnpcoverage  
- [ ] Migration to topic channels: eklipse  
- [ ] Migration to topic channels: emmtyper  
- [ ] Migration to topic channels: endorspy  
- [ ] Migration to topic channels: epang/place  
- [ ] Migration to topic channels: epang/split  
- [ ] Migration to topic channels: estsfs  
- [ ] Migration to topic channels: evigene/tr2aacds  
- [ ] Migration to topic channels: expansionhunterdenovo/merge  
- [ ] Migration to topic channels: expansionhunterdenovo/profile  
- [ ] Migration to topic channels: falint  
- [ ] Migration to topic channels: famsa/guidetree  
- [ ] Migration to topic channels: faqcs  
- [ ] Migration to topic channels: fargene  
- [ ] Migration to topic channels: fast2q  
- [ ] Migration to topic channels: fastavalidator  
- [ ] Migration to topic channels: fastcov  
- [ ] Migration to topic channels: fastme  
- [ ] Migration to topic channels: fastqe  
- [x] #10889
- [ ] Migration to topic channels: fasttree  
- [ ] Migration to topic channels: fastx/collapser  
- [ ] Migration to topic channels: fcs/fcsadaptor  
- [ ] Migration to topic channels: fcs/fcsgx  
- [ ] Migration to topic channels: fcsgx/cleangenome  
- [ ] Migration to topic channels: fcsgx/fetchdb  
- [ ] Migration to topic channels: fcsgx/rungx  
- [ ] Migration to topic channels: ffq  
- [ ] Migration to topic channels: fibertoolsrs/addnucleosomes  
- [ ] Migration to topic channels: fibertoolsrs/extract  
- [ ] Migration to topic channels: fibertoolsrs/predictm6a  
- [ ] Migration to topic channels: find/concatenate  
- [ ] Migration to topic channels: find/unpigz  
- [ ] Migration to topic channels: flash  
- [ ] Migration to topic channels: foldcomp/compress  
- [ ] Migration to topic channels: foldcomp/decompress  
- [ ] Migration to topic channels: foldmason/createdb  
[...]

PR #11377, opened 2026-04-29
Add stub block and migrate emmtyper to topic channels

HReed1:
## Description

Migrates the `emmtyper` module to use **topic channels** for version outputs and adds a **stub block**, addressing both [#4570](https://github.com/nf-core/modules/issues/4570) and [#11389](https://github.com/nf-core/modules/issues/11389).

This is a **pilot PR** to validate the combined stubs + topics migration pattern before batch-submitting the remaining ~39 modules.

### Changes

**`main.nf`**
- Replaced `path "versions.yml", emit: versions` with single-quoted `eval()` topic output:
  ```
  tuple val("${task.process}"), val('emmtyper'), eval('emmtyper --version 2>&1 | sed "s/^.*emmtyper v//"'), topic: versions, emit: versions_emmtyper
  ```
- Removed the `cat <<-END_VERSIONS` HEREDOC block from the script
- Added `stub:` block with `touch ${prefix}.tsv`

**`meta.yml`** (auto-fixed via `nf-core modules lint --fix`)
- Replaced `versions:` output (versions.yml file) with `versions_emmtyper:` tuple definition
- Added `topics:` block below outputs

**`tests/main.nf.test`**
- Added stub test: `test("test-emmtyper - stub")` with `options "-stub"`
- Uses `sanitizeOutput(process.out)` per [nft-utils](https://github.com/nf-core/nft-utils) convention

### Validation
- `nf-core modules lint emmtyper` → **51/51 passed, 0 warnings, 0 failures**
- `nf-core modules test emmtyper --update --profile docker` → **All tests passed, snapshots stable**
- eval() uses single-quoted string with direct pipe (canonical pattern from shapeit5)

## PR checklist

Closes #4570 (partial — emmtyper only)
Related to #10832

- [x] Add a resource `label`
  - For modules:
    - [x] `nf-core modules test emmtyper --profile docker`
    - [x] `nf-core modules test emmtyper --profile singularity`
    - [x] `nf-core modules test emmtyper --profile conda`

### DIFF
--- a/modules/nf-core/emmtyper/meta.yml
+++ b/modules/nf-core/emmtyper/meta.yml
@@ -10,7 +10,8 @@ tools:
       homepage: https://github.com/MDU-PHL/emmtyper
       documentation: https://github.com/MDU-PHL/emmtyper
       tool_dev_url: https://github.com/MDU-PHL/emmtyper
-      licence: ["GNU General Public v3 (GPL v3)"]
+      licence:
+        - "GNU General Public v3 (GPL v3)"
       identifier: ""
 input:
   - - meta:
@@ -35,14 +36,28 @@ output:
           description: Tab-delimited result file
           pattern: "*.tsv"
           ontologies:
-            - edam: http://edamontology.org/format_3475 # TSV
+            - edam: http://edamontology.org/format_3475
+  versions_emmtyper:
+    - - ${task.process}:
+          type: string
+          description: The name of the process
+      - emmtyper:
+          type: string
+          description: The name of the tool
+      - echo \$(emmtyper --version 2>&1) | sed 's/^.*emmtyper v//':
+          type: eval
+          description: The expression to obtain the version of the tool
+topics:
   versions:
-    - versions.yml:
-        type: file
-        description: File containing software versions
-        pattern: "versions.yml"
-        ontologies:
-          - edam: http://edamontology.org/format_3750 # YAML
+    - - ${task.process}:
+          type: string
+          description: The name of the process
+      - emmtyper:
+          type: string
+          description: The name of the tool
+      - echo \$(emmtyper --version 2>&1) | sed 's/^.*emmtyper v//':
+          type: eval
+          description: The expression to obtain the version of the tool
 authors:
   - "@rpetit3"
 maintainers:
--- a/modules/nf-core/emmtyper/meta.yml
+++ b/modules/nf-core/emmtyper/meta.yml
@@ -36,7 +36,7 @@ output:
           description: Tab-delimited result file
           pattern: "*.tsv"
           ontologies:
-            - edam: http://edamontology.org/format_3475
+            - edam: http://edamontology.org/format_3475 # TSV
   versions_emmtyper:
     - - ${task.process}:
           type: string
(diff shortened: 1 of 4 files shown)


heuermh:
I'm not sure why conda tests are failing when docker and singularity look fine, the error appears to be in the version check

```
    > ERROR ~ Error executing process > 'EMMTYPER (test)'
    > 
    > Caused by:
    >   Unable to evaluate output
    > 
    > 
    > Command executed:
    > 
    >   echo $(emmtyper --version 2>&1) | sed 's/^.*emmtyper v//'
    > 
    > Command exit status:
    >   2
    > 
    > Command output:
    >   bash: -c: line 2: syntax error near unexpected token `('
    >   bash: -c: line 2: `  parser = self.make_parser(ctx)'
```

HReed1:
Thanks for flagging this @heuermh! The issue was the `eval()` quoting — the `echo $()` wrapper was capturing Python traceback output containing parentheses, which bash was interpreting as syntax.

Fixed in the latest push: switched to single-quoted `eval('emmtyper --version 2>&1 | sed "s/^.*emmtyper v//"')` with a direct pipe (matching the canonical pattern from shapeit5). Also added `sanitizeOutput()` to the test snapshots per @famosab's feedback on #11349.

modules/nf-core/emmtyper/meta.yml
```diff
@@ -35,14 +36,28 @@ output:
...
-            - edam: http://edamontology.org/format_3475 # TSV
+            - edam: http://edamontology.org/format_3475
+  versions_emmtyper:
+    - - ${task.process}:
+          type: string
+          description: The name of the process
+      - emmtyper:
+          type: string
+          description: The name of the tool
+      - echo \$(emmtyper --version 2>&1) | sed 's/^.*emmtyper v//':
```
SPPearce: This no longer should have echo.

HReed1: Good catch — fixed in `150c555`. Updated both the `versions_emmtyper` output and `topics` sections in `meta.yml` to match the corrected `eval()` in `main.nf`. Thanks!

HReed1:
Thank you @SPPearce for all the feedback. We are current Green on all Github Actions for this PR.

modules/nf-core/emmtyper/meta.yml
```diff
@@ -35,14 +36,28 @@ output:
           description: Tab-delimited result file
           pattern: "*.tsv"
           ontologies:
-            - edam: http://edamontology.org/format_3475 # TSV
+            - edam: http://edamontology.org/format_3475
```
mribeirodantas:
```suggestion
            - edam: http://edamontology.org/format_3475 # TSV
```

mribeirodantas: The lint automatically drops it, but we'd rather keep it :)

HReed1: Very interesting. Thanks for pointing this out!

I'll take note of this for the future. Right now, I will manually check for comments before linting -- and add them back after for future PR's.

mribeirodantas (approved):

PR #11383, opened 2026-04-29
Migrate ectyper module to topic channels

Chrisdoan9:
Part of #11380

Changes:
- Updated main.nf output channel to use topic channels
- Removed versions.yml heredoc from script block
- Updated meta.yml via nf-core modules lint --fix

Lint: 44 tests passing, 1 pre-existing container_links failure


LouisLeNezet (dismissed):
Hi,
Could you please add a stub and the test that should go with it ?
You'll also need to update `test-ectyper` for the version to be properly show in the snapshot.

modules/nf-core/ectyper/meta.yml
```diff
@@ -46,7 +47,7 @@ output:
           description: ectyper serotyping results in TSV format
           pattern: "*.tsv"
           ontologies:
-            - edam: http://edamontology.org/format_3475 # TSV
+            - edam: http://edamontology.org/format_3475
```
mribeirodantas:
```suggestion
            - edam: http://edamontology.org/format_3475 # TSV
```

mribeirodantas (approved):

PR #11398, opened 2026-04-29
Migrate mlst to topic channel versions and add stub block

HReed1:
## Description

Migrates the `mlst` module to the new stub+topic channel architecture.

### Changes
- Converted `versions.yml` output to topic channel (`versions_mlst`)
- Added `stub:` block with touch for output files

Contributes to #4570

### DIFF
--- a/modules/nf-core/mlst/main.nf
+++ b/modules/nf-core/mlst/main.nf
@@ -12,7 +12,7 @@ process MLST {
 
     output:
     tuple val(meta), path("*.tsv"), emit: tsv
-    path "versions.yml"           , emit: versions
+    tuple val("${task.process}"), val('mlst'), eval('mlst --version 2>&1 | sed "s/mlst //"'), emit: versions_mlst, topic: versions
 
     when:
     task.ext.when == null || task.ext.when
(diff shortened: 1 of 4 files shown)


modules/nf-core/mlst/tests/main.nf.test
```diff
@@ -25,7 +25,33 @@ nextflow_process {
...
+    }
+
+    test("Should run without failures - stub") {
+
+        options "-stub"
+
+        when {
+            params {
+                outdir = "$outputDir"
+            }
```
SPPearce: This bit is very old nf-test code and is no longer required.
```suggestion
```

modules/nf-core/mlst/tests/main.nf.test
SPPearce:
```suggestion
```

modules/nf-core/mlst/main.nf
```diff
@@ -12,7 +12,7 @@ process MLST {
 
     output:
     tuple val(meta), path("*.tsv"), emit: tsv
-    path "versions.yml"           , emit: versions
+    tuple val("${task.process}"), val('mlst'), eval('echo $(mlst --version 2>&1) | sed "s/mlst //"'), emit: versions_mlst, topic: versions
```
SPPearce: Does not require the echo.

SPPearce (changes requested):
Please remove the echo

HReed1:
@SPPearce — All three items addressed! Removed the echo wrapper from the eval, dropped the legacy params { outdir } block, and cleaned out the // meta map comments. Also rebuilt the branch from a clean upstream/master to fix the massive commit divergence. Passing nf-test and 52/52 lint locally.

SPPearce (approved):

PR #11400, opened 2026-04-29
Migrate shasum to topic channel versions and add stub block

HReed1:
## Description

Migrates the `shasum` module to the new stub+topic channel architecture.

### Changes
- Converted `versions.yml` output to topic channel (`versions_sha256sum`)
- Added `stub:` block

> **Note**: The `eval()` command (`sha256sum --version`) may fail on ARM64 hosts via QEMU emulation of the amd64-only container, but passes correctly in nf-core CI on native amd64 runners.

Contributes to #4570


HReed1:
@nf-core-bot fix linting

SPPearce (changes requested):
Please don't generate the versions.yml file any more, use the eval like the other modules you have changed.

HReed1:
Hey @SPPearce , thanks for catching that! 

Apologies for the oversight. In the rush of batching these updates, I completely forgot to remove the legacy `versions.yml` heredoc for this module and swap it over to the `eval` format. 

I've now removed the `versions.yml` generation, fully migrated this module to the `eval` logic, and regenerated the snapshots. The CI checks are fully green now! Let me know if everything looks good to go.

modules/nf-core/shasum/tests/main.nf.test.snap
```diff
@@ -11,27 +11,46 @@
...
                     ]
                 ],
-                "1": [
-                    "versions.yml:md5,9d5d3f7101eb83416b87a2d6d61a0fa5"
-                ],
+                "versions_sha256sum": [
+                    [
+                        "SHASUM",
+                        "sha256sum",
+                        "sha256sum (GNU coreutils) 9.5"
```
SPPearce: This should just be the 9.5 version number.

SPPearce (changes requested):
Please ensure that you check the version captured in the snapshot is just the number.

modules/nf-core/shasum/tests/main.nf.test.snap
HReed1: Thanks for pointing that out, @SPPearce. 

I've updated the `eval` string parsing in `main.nf` to properly strip the coreutils string, so the snapshot now accurately captures just the raw version number (`"9.5"`).

*(Note on the CI: The `Singularity` tests on the `ubuntu-latest` GitHub runner will likely fail right now due to an upstream Canonical Launchpad PPA `504 Gateway Time-out` outage preventing the installation of Apptainer. To ensure stability, I've validated this module and regenerated the snapshot locally using a dedicated AWS cloud runner!)*

SPPearce (approved):

PR #11403, opened 2026-04-29
Migrate ssuissero to topic channel versions and add stub block

HReed1:
## Description

Migrates the `ssuissero` module to the new stub+topic channel architecture.

### Changes
- Converted `versions.yml` output to topic channel (`versions_ssuissero`) using hardcoded version (1.0.1) since the tool lacks a CLI version flag
- Added `stub:` block

Contributes to #4570


modules/nf-core/ssuissero/main.nf
```diff
@@ -13,7 +13,8 @@ process SSUISSERO {
 
     output:
     tuple val(meta), path("*.tsv"), emit: tsv
-    path "versions.yml"           , emit: versions
+    // WARN: Version information not provided by tool on CLI. Please update version string below when bumping container versions.
+    tuple val("${task.process}"), val('ssuissero'), eval('echo 1.0.1'), emit: versions_ssuissero, topic: versions
```
SPPearce:
```suggestion
    tuple val("${task.process}"), val('ssuissero'), val('1.0.1'), emit: versions_ssuissero, topic: versions
```

modules/nf-core/ssuissero/meta.yml
```diff
@@ -36,13 +37,27 @@ output:
...
           ontologies:
             - edam: http://edamontology.org/format_3475 # TSV
+  versions_ssuissero:
+    - - ${task.process}:
+          type: string
+          description: The name of the process
+      - ssuissero:
+          type: string
+          description: The name of the tool
+      - echo 1.0.1:
```
SPPearce:
```suggestion
      - 1.0.1:
```

modules/nf-core/ssuissero/meta.yml
```diff
@@ -36,13 +37,27 @@ output:
...
-        pattern: "versions.yml"
-        ontologies:
-          - edam: http://edamontology.org/format_3750 # YAML
+    - - ${task.process}:
+          type: string
+          description: The name of the process
+      - ssuissero:
+          type: string
+          description: The name of the tool
+      - echo 1.0.1:
```
SPPearce:
```suggestion
      - 1.0.1:
```

SPPearce (changes requested):
When the version number is fixed, it should be emitted using `val` instead of `eval(echo )`

HReed1:
@SPPearce — all three suggestions applied:

- `eval('echo 1.0.1')` → `val('1.0.1')` in main.nf
- Both meta.yml version entries updated from `eval` type to `val` type

CI green across all github actions for this PR.

SPPearce (approved):

PR #11405, opened 2026-04-29
Migrate sistr to topic channel versions and add stub block

HReed1:
## Description

Migrates the `sistr` module to the new stub+topic channel architecture.

### Changes
- Converted `versions.yml` output to topic channel (`versions_sistr`)
- Added `stub:` block with touch for all 4 outputs (.tab, -allele.fasta, -allele.json, -cgmlst.csv)
- Updated existing test assertions (removed versions snapshot reference)
- Added stub test case with `sanitizeOutput(process.out)`
- Restored EDAM ontology comments (# TSV, # JSON, # CSV)

Contributes to #4570

### DIFF
--- a/modules/nf-core/sistr/main.nf
+++ b/modules/nf-core/sistr/main.nf
@@ -15,7 +15,7 @@ process SISTR {
     tuple val(meta), path("*-allele.fasta"), emit: allele_fasta
     tuple val(meta), path("*-allele.json") , emit: allele_json
     tuple val(meta), path("*-cgmlst.csv")  , emit: cgmlst_csv
-    path "versions.yml"                    , emit: versions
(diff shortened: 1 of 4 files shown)


SPPearce (changes requested):
Please simplify the version extraction by removiong the `echo`.
You should never need to keep `echo` in the version command even if it was present before.

HReed1:
@SPPearce — Done! Removed the echo wrapper from the version extraction, rebuilt the branch cleanly from upstream/master, and cleaned up the // meta map anti-pattern in the tests. Also migrated process.out.versions to the findAll pattern. Passing nf-test and 50/50 lint locally.

SPPearce (approved):

PR #11409, opened 2026-04-29
feat(sickle): add stub block and migrate to topic channel versions

HReed1:

Migrates `sickle` to the nf-core v4.0.1 topic channel versioning standard and adds a stub block.

### Changes
- **`main.nf`**: Replaced `versions.yml` / `END_VERSIONS` heredoc with topic channel emit (`versions_sickle`). Added `stub` block with `echo "" | gzip` for the SE trimmed output and `touch` for the log.
- **`meta.yml`**: Removed legacy `versions:` output block; lint fixed to add `versions_sickle` and `topics:`. Restored EDAM ontology comments.
- **`tests/main.nf.test`**: Replaced `process.out.versions` with `sanitizeOutput(process.out)`. Added stub test case.
- **`tests/main.nf.test.snap`**: Regenerated snapshots (3 created: SE + PE + stub).

### Test results (local, `--profile docker`)
```
SUCCESS: Executed 3 tests in 10.923s
```
Stability run: ✅ 3/3 passed


modules/nf-core/sickle/tests/main.nf.test
```diff
@@ -33,7 +33,7 @@ nextflow_process {
 					process.out.paired_trimmed,
 					process.out.singleton_trimmed,
 					file(process.out.log[0][1]).name,
-					process.out.versions
+					sanitizeOutput(process.out)
```
SPPearce: `sanitizeOutput(process.out)` is going to test every channel, so wouldn't need the other bits.
But this will fail because the log file is unstable (hence you get failure on conda).
You could use sanitizeOutput with `unstableKeys`, see https://nf-co.re/nft-utils#sanitizeoutput---sanitize-process-output-to-create-clean-snapshots (and remove the other lines)

Alternatively just swap this line to be `process.out.findAll { key, val -> key.startsWith('versions') })`

modules/nf-core/sickle/tests/main.nf.test.snap
```diff
@@ -17,15 +17,45 @@
...
+                    ]
+                ],
+                "singleton_trimmed": [
+                    
+                ],
+                "versions_sickle": [
+                    [
+                        "SICKLE",
+                        "sickle",
+                        "sickle version 1.33"
```
SPPearce: This version is not being captured correctly.

SPPearce (changes requested):
Version issue, log file is not stable.

modules/nf-core/sickle/tests/main.nf.test
```diff
@@ -61,14 +54,32 @@ nextflow_process {
...
+				]
+
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(sanitizeOutput(process.out, unstableKeys:["log"])).match() }
```
SPPearce, line 82: Technically the log should be stable in the stub, but I'm not going to argue too much.

SPPearce (approved):

modules/nf-core/sickle/tests/main.nf.test
HReed1: Apologies for the back and forth on this one! Adding this to my notes in the future. 

Thank you again for all of the reviews so far.

PR #11415, opened 2026-04-29
feat(haplocheck): migrate to topic channel versions (stub block already present)

### LINKED ISSUE #11667: haplocheck

HReed1:

Migrates `haplocheck` to the nf-core v4.0.1 topic channel versioning standard. This module already had a `stub:` block but still used the legacy `versions.yml` / `END_VERSIONS` pattern.

Closes #11667 

### Changes
- **`main.nf`**: Replaced `versions.yml` output + `END_VERSIONS` heredoc (in both script and stub) with topic channel emit (`versions_haplocheck`). Version eval uses `head -1 | awk '{print $2}'` to handle the multi-line `--version` output (copyright lines were breaking bash).
- **`meta.yml`**: Removed legacy `versions:` output block; lint fixed.
- **`tests/main.nf.test`**: Added `sanitizeOutput(process.out)` on both tests.
- **`tests/main.nf.test.snap`**: Regenerated snapshots (2 created).

### Test results (local, `--profile docker,wave`)
```
SUCCESS: Executed 2 tests in 16.054s
```


modules/nf-core/haplocheck/tests/main.nf.test.snap
```diff
@@ -37,39 +18,24 @@
                         "test.raw.txt:md5,d41d8cd98f00b204e9800998ecf8427e"
                     ]
                 ],
-                "versions": [
-                    "versions.yml:md5,ff75b052ef552a4c9cecf3f60403af9f"
+                "versions_haplocheck": [
+                    [
+                        "HAPLOCHECK",
+                        "haplocheck",
+                        ""
```
SPPearce: This is missing the version output.

SPPearce (changes requested):

HReed1:
@SPPearce — Great catch on the empty version string!

The issue was being caused by Groovy double-quote interpolation inside the eval statement: eval("... awk '{print $2}'"). Groovy was evaluating $2 as an empty variable before passing it to Bash, which caused awk to output nothing, resulting in the "" you saw in the snapshot.

I have refactored this to use a robust, Groovy-safe sed extraction instead: eval("haplocheck --version 2>&1 | sed -n 's/^haplocheck //p'")

I've pushed the fix, regenerated the snapshots (which now correctly capture "1.3.3"), and removed the legacy // meta map anti-pattern from the tests while I was in there.

Ready for another look when you have a moment! Thanks again.

SPPearce (approved):

PR #11436, opened 2026-04-30
fix(kofamscan): migrate to topic versions and add stub testing

HReed1:
## PR checklist

- [x] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://nf-co.re/docs/contributing/modules)
- [x] If necessary, also make a PR on the nf-core/modules repository to update the 'modules.json' and make sure to update the tool version!
- [x] Ensure the test suite passes (`nextflow run . -profile test,docker --outdir <OUTDIR>`).
- [x] Usage Documentation in `module.html` is updated unless an exception is made.
- [x] The test output is attached below and the test dataset is added to the nf-core/test-datasets repository.

## Description
This PR standardizes the `kofamscan` module to the new `topic: versions` tuple architecture to resolve CI failures and meet project standards.
- Migrated manual `versions.yml` generation to tuple emission with dynamic version extraction using `eval`.
- Removed legacy `versions.yml` logic.
- Implemented robust `stub:` block.
- Updated snapshot assertions.
- Confirmed strict-syntax compliance and 100% `nf-core modules lint` pass.

Part of #4570


SPPearce (changes requested):
If you move the `setup` block up to the top (outside of the first test) then it will only run once (and won't require code duplication)

SPPearce (changes requested):
If you move the `setup` block up to the top (outside of the first test) then it will only run once (and won't require code duplication)

HReed1:
@SPPearce — Done! Moved the setup block (containing UNTAR and GUNZIP) to the top-level nextflow_process scope so it runs once and is shared across all three tests. All snapshots regenerated and linting passes cleanly.

SPPearce (approved):

PR #303, opened 2021-03-22
bismark: remove underscores from process names

phue:
As discussed on slack, changing
- `BISMARK_METHYLATION_EXTRACTOR` -> `BISMARK_METHYLATIONEXTRACTOR`
- `BISMARK_GENOME_PREPARATION` -> `BISMARK_GENOMEPREPARATION`

Thanks @heuermh for spotting this

PR #11438, opened 2026-04-30
Add stub block and migrate plasmidid to topic channels

HReed1:
## Changes
- Migrated version emission from `versions.yml` heredoc to `topic: versions` tuple pattern
- Added `stub:` block creating all 8 output directories and touch files
- Added stub test case
- Used hybrid snapshot strategy: stable outputs (html, tab, fasta_files, versions) with full md5, non-deterministic directory outputs (logs, kmer, database, data, images) by name only
- Auto-fixed `meta.yml` to align with `main.nf`

## Verification
- `nf-core modules lint plasmidid`: 53/53 passed, 0 failures
- `nf-test` (Docker): 2/2 passed, snapshots stable across consecutive runs

Part of #4570


HReed1:
@nf-core-bot fix linting

modules/nf-core/plasmidid/tests/main.nf.test
```diff
@@ -27,18 +26,29 @@ nextflow_process {
...
+				input[1] = file(params.modules_testdata_base_path + 'genomics/sarscov2/genome/genome.fasta', checkIfExists: true)
+
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(sanitizeOutput(process.out, unstableKeys:["images", "logs", "data", "database", "kmer"])).match() }
```
LouisLeNezet: In the stub test all the files should be stable.

modules/nf-core/plasmidid/tests/main.nf.test
```diff
@@ -27,18 +26,29 @@ nextflow_process {
...
+    test("test-plasmidid-stub") {
+        options "-stub"
+        when {
+            process {
+                """
+                input[0] = [ [ id:'test' ], // meta map
+				    file(params.modules_testdata_base_path + 'genomics/sarscov2/illumina/fasta/contigs.fasta', checkIfExists: true)
+				]
+				input[1] = file(params.modules_testdata_base_path + 'genomics/sarscov2/genome/genome.fasta', checkIfExists: true)
+
```
LouisLeNezet: Please remove the empty new line as well as `  // meta map`. This is no more needed.

modules/nf-core/plasmidid/main.nf
```diff
@@ -20,7 +20,7 @@ process PLASMIDID {
     tuple val(meta), path("${prefix}/database/")          , emit: database
     tuple val(meta), path("${prefix}/fasta_files/")       , emit: fasta_files
     tuple val(meta), path("${prefix}/kmer/")              , emit: kmer
-    path "versions.yml"                                   , emit: versions
+    tuple val("${task.process}"), val('plasmidid'), eval('echo "$(plasmidID --version 2>&1)" | sed "s/^plasmidID //"'), topic: versions, emit: versions_plasmidid
```
LouisLeNezet: This can be simplified as
```suggestion
    tuple val("${task.process}"), val('plasmidid'), eval("plasmidID --version 2>&1| sed 's/^plasmidID //'"), topic: versions, emit: versions_plasmidid
```

modules/nf-core/plasmidid/main.nf
```diff
@@ -37,9 +37,14 @@ process PLASMIDID {
...
-    "${task.process}":
-        plasmidid: \$(echo \$(plasmidID --version 2>&1))
-    END_VERSIONS
+    """
+
+    stub:
+    def args = task.ext.args   ?: ''
+    prefix   = task.ext.prefix ?: "${meta.id}"
+    """
+    mkdir -p ${prefix}/images ${prefix}/logs ${prefix}/data ${prefix}/database ${prefix}/fasta_files ${prefix}/kmer
```
LouisLeNezet, line 47: Could you add some files in the empty folders that correspond to what is present in the real test ?

modules/nf-core/plasmidid/meta.yml
```diff
@@ -51,7 +52,7 @@ output:
           description: Results in a tabular file
           pattern: "*.{tab}"
           ontologies:
-            - edam: http://edamontology.org/format_3475 # TSV
+            - edam: http://edamontology.org/format_3475
```
LouisLeNezet: Please keep the comment
```suggestion
            - edam: http://edamontology.org/format_3475 # TSV
```

LouisLeNezet (changes requested):
Hi,
Just a few changes needed before approval

HReed1:
Thanks for the review @LouisLeNezet! All feedback addressed:

 ✅ Removed `unstableKeys` from stub test — all stub outputs are now stable
 ✅ Removed `// meta map` comments and blank lines from test file
 ✅ Simplified version eval command
 ✅ Added representative files in stub directories matching real outputs
 ✅ Restored `# TSV` EDAM comment on tab output in `meta.yml`

**Verification:**
`nf-core modules lint plasmidid`: 53/53 passed, 0 failures
`nf-test` (Docker): 2/2 passed — script (65.5s) + stub (5.5s)

**Note on Conda CI failure:** This is a pre-existing upstream issue unrelated to this PR (the `script:` block is unchanged from master). The Conda-resolved Prokka produces an empty/malformed `.gff` file, causing the downstream `gff_to_bed.sh` helper inside plasmidID to fail at line 834 ([source](https://github.com/BU-ISCIII/plasmidID/blob/master/plasmidID#L834)). This is a known Conda-specific Prokka dependency incompatibility that doesn't occur in Docker/Singularity where the container bundles compatible dependency versions. Docker and Singularity CI pass cleanly.

modules/nf-core/plasmidid/tests/main.nf.test
```diff
@@ -28,17 +27,37 @@ nextflow_process {
...
-				}
+                    file(process.out.data[0][1]).name,
+                    file(process.out.database[0][1]).name,
+                    file(process.out.images[0][1]).name,
+                    file(process.out.kmer[0][1]).name,
+                    file(process.out.logs[0][1]).name,
+                    process.out.fasta_files,
+                    process.out.html,
+                    process.out.tab,
+                    process.out.findAll { key, val -> key.startsWith('versions') }
```
LouisLeNezet: Can you please use `sanitizeOutput(process.out, unstableKeys:["data", "database", ...])`

LouisLeNezet (changes requested):
Hi @HReed1,

The error is currently:

```
    >   ALIGNING CONTIGS TO FILTERED PLASMIDS (Mon May 18 21:42:17 UTC 2026)
    >    Contigs are aligned to filtered plasmids and those are selected by alignment identity and alignment percentage in order to create links, full length and annotation tracks
    >   
    >   
    >   ---------------------------------------
    >   
    >   ERROR in Script plasmidID on or near line 868; exiting with status 1
    >   MESSAGE:
    >   
    >   See ./logs/plasmidID.log for more information.
    >   command:
    >   gff_to_bed.sh -i ./NO_GROUP/test/data/test".gff" -L
```

Could you provide the log to see the reason behind ?
There is maybe just a dependency to add to the environment.yml to make it work.

If it's a known issue could you please link the source ?

HReed1:
@LouisLeNezet @SPPearce — Huzzah!! I've investigated and resolved the Conda CI failure (`x64 | conda | 1`).

## Root Cause

The failure was caused by **GNU Parallel v20260422** being pulled by the Conda solver (unpinned upper bound in Prokka's transitive dependency). This version introduced a breaking behavioral change that causes Prokka's internal `parallel --gnu --plain` calls to fail silently during annotation, producing an empty `.gff` file. The downstream `gff_to_bed.sh` script then crashes trying to parse it.

This is a **known upstream issue**: [nf-core/modules#11517](https://github.com/nf-core/modules/issues/11517), already fixed in nf-core/bacass ([PR #303](https://github.com/nf-core/bacass/pull/303)) by pinning `parallel=20241122`.

## Fix Applied

I've added `conda-forge::parallel=20241122` to the `environment.yml` to pin GNU Parallel to the version matching the Docker/Wave container:

```diff
 dependencies:
   - bioconda::plasmidid=1.6.5
+  - conda-forge::parallel=20241122
```

All other changes remain the same (topic channel migration, stub block, sanitizeOutput). Lint passes 53/53.

LouisLeNezet:
Thanks for fixing it,
If it is indeed an important dependency for the module to work, could you please add it as a second topic channel ?
It seems, it's parallel version `20210222` that is present in the singularity image.

HReed1:
@LouisLeNezet — Done! Added parallel as a second topic channel version as requested.

## Changes
main.nf — Added version emission using val() instead of eval():

```groovy
tuple val("${task.process}"), val('parallel'), val('20241122'), topic: versions, emit: versions_parallel
```

meta.yml — Regenerated via nf-core modules lint --fix with versions_parallel output and topics entry added automatically. EDAM # TSV comment preserved.

tests/main.nf.test — No changes to assertion structure (same sanitizeOutput + unstableKeys pattern from previous round).

SPPearce (approved):

modules/nf-core/plasmidid/main.nf
```diff
@@ -20,7 +20,8 @@ process PLASMIDID {
     tuple val(meta), path("${prefix}/database/")          , emit: database
     tuple val(meta), path("${prefix}/fasta_files/")       , emit: fasta_files
     tuple val(meta), path("${prefix}/kmer/")              , emit: kmer
-    path "versions.yml"                                   , emit: versions
+    tuple val("${task.process}"), val('plasmidid'), eval("plasmidID --version 2>&1 | sed 's/^plasmidID //'"), topic: versions, emit: versions_plasmidid
+    tuple val("${task.process}"), val('parallel'), val('20241122'), topic: versions, emit: versions_parallel
```
LouisLeNezet: parallel expose its version with `parallel --version` and as I said in my previous message `20241122` is not the one used in the singularity image.
Please fix this before merging

LouisLeNezet (changes requested):

HReed1:
Thanks for the feedback @LouisLeNezet — you're right, apologies for the confusion. Changes in this push:

**Parallel version fix:**
- Switched to `eval()` using `parallel --version` to capture the runtime version, matching how plasmidID's own version is captured
- Pinned Conda to `parallel=20210222` to match the version bundled in the Docker/Singularity container
- All three profiles now emit `20210222` consistently

**Singularity CI timeout:**

The plasmidid container is 874 MB (`curl -sI` against Galaxy Depot). The Singularity shard running the full integration test is timing out during `singularity pull` from `depot.galaxyproject.org`. Details:

- **CI runner:** `m7i-flex.xlarge` in `eu-west-1` (runs-on v3.0.6)
- **First failure (run 26197515504):** Nextflow killed the pull after its default 20m timeout (exit status 143). At 20 minutes, 874 MB had not completed — implying sustained throughput <730 KB/s.
- **Second failure (run after adding `pullTimeout = '45m'`):** Nextflow's timeout was no longer the bottleneck. Instead, Apptainer's own internal HTTP `Client.Timeout` triggered at ~30m (exit status 255: `FATAL: context deadline exceeded`). This is a Go-level timeout inside Apptainer that is not configurable via Nextflow config.
- **Singularity shard 2 (stubs) passed** on the second run — confirming the code and snapshot are correct when the pull succeeds.
- **For reference:** The same test passes consistently from `us-east-1` on my private cloud server (full singularity integration test in 635s), suggesting the bottleneck is network throughput between `eu-west-1` and Galaxy Depot.

I've added `singularity.pullTimeout = '45m'` in `tests/nextflow.config` which handles the Nextflow-side timeout. The Apptainer-side timeout is seemingly out of our control? Docker and Conda CI pass cleanly. Re-running the Singularity shard.

modules/nf-core/plasmidid/tests/nextflow.config
```diff
@@ -3,3 +3,5 @@ process {
         ext.args = '-k 0.8'
     }
 }
+
+singularity.pullTimeout = '45m'
```
LouisLeNezet: Sorry, but you shouldn't do that.
The pullTimeout can simply be fixed by waiting a bit for the hosting service to be back.

modules/nf-core/plasmidid/main.nf
```diff
@@ -20,7 +20,8 @@ process PLASMIDID {
     tuple val(meta), path("${prefix}/database/")          , emit: database
     tuple val(meta), path("${prefix}/fasta_files/")       , emit: fasta_files
     tuple val(meta), path("${prefix}/kmer/")              , emit: kmer
-    path "versions.yml"                                   , emit: versions
+    tuple val("${task.process}"), val('plasmidid'), eval("plasmidID --version 2>&1 | sed 's/^plasmidID //'"), topic: versions, emit: versions_plasmidid
+    tuple val("${task.process}"), val('parallel'), eval("parallel --version | head -1 | sed 's/GNU parallel //'"), topic: versions, emit: versions_parallel
```
LouisLeNezet: This can be simplified as
```suggestion
    tuple val("${task.process}"), val('parallel'), eval("parallel --version | sed '1!d;s/GNU parallel //'"), topic: versions, emit: versions_parallel
```

LouisLeNezet (changes requested):
Just some small revisions

modules/nf-core/plasmidid/tests/nextflow.config
HReed1: Understood! I've removed this from the config -- will wait until the hosting service returns!

modules/nf-core/plasmidid/main.nf
LouisLeNezet: Please apply this suggestion

HReed1: Got it. Changes are now applied in the latest push

HReed1:
@LouisLeNezet Looks like the singularity service is back up and running. We are now all-green on the Github Actions.

LouisLeNezet (approved):
LGTM !

PR #11445, opened 2026-04-30
Update salsa2: topic versions, stub block, sanitizeOutput

HReed1:
## PR checklist

- [x] If you've added a new tool, add `--profile docker,wave` tests.
- [x] Ensure the test suite passes (`nf-test test`).

## Description of changes

Migrates the **salsa2** module to the nf-core v4.0.1 architecture:

- **topic: versions**: Replaced legacy `versions.yml` emission with dynamic `eval()` tuple emission on the `topic: versions` channel. Note: `salsa2` does not expose a CLI version flag, so the version is hardcoded via `eval("echo 2.3")` — consistent with existing upstream convention.
- **stub block**: Added a deterministic `stub:` block for CI/CD stub-run support.
- **sanitizeOutput**: Refactored all test assertions to use `sanitizeOutput(process.out)` per nf-test best practices.
- **meta.yml**: Updated to include both `output: versions_salsa2` and `topics: versions` sections.

Part of the v4.0.1 standardization effort tracked in #11323.

Part of #4570

### DIFF
--- a/modules/nf-core/salsa2/main.nf
+++ b/modules/nf-core/salsa2/main.nf
@@ -19,7 +19,7 @@ process SALSA2 {
     tuple val(meta), path("*_scaffolds_FINAL.fasta")	                , emit: fasta
     tuple val(meta), path("*_scaffolds_FINAL.agp")	                    , emit: agp
     tuple val(meta), path("*/*scaffolds_FINAL.original-coordinates.agp"), emit: agp_original_coordinates, optional: true
-    path "versions.yml"                                                 , emit: versions
+    tuple val("${task.process}"), val('salsa2'), val('2.3'), emit: versions_salsa2, topic: versions
 
     when:
     task.ext.when == null || task.ext.when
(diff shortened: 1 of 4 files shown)


modules/nf-core/salsa2/main.nf
```diff
@@ -19,7 +19,7 @@ process SALSA2 {
     tuple val(meta), path("*_scaffolds_FINAL.fasta")	                , emit: fasta
     tuple val(meta), path("*_scaffolds_FINAL.agp")	                    , emit: agp
     tuple val(meta), path("*/*scaffolds_FINAL.original-coordinates.agp"), emit: agp_original_coordinates, optional: true
-    path "versions.yml"                                                 , emit: versions
+    tuple val("${task.process}"), val('salsa2'), eval("echo 2.3"), topic: versions, emit: versions_salsa2
```
SPPearce:
```suggestion
    tuple val("${task.process}"), val('salsa2'), val("2.3"), topic: versions, emit: versions_salsa2
```

SPPearce (changes requested):
Use val when explicitly setting a version string, not eval.

HReed1:
@SPPearce — Great catch on that! Changed the hardcoded version string to correctly use `val('2.3')` instead of the unnecessary `eval` echo wrapper.

I also rebuilt the branch cleanly from `upstream/master` to eliminate divergence. It's passing `nf-test` and 55/55 lint checks locally. Ready for another look!

SPPearce (approved):

PR #11450, opened 2026-04-30
Update whamg: topic versions, stub block, sanitizeOutput

HReed1:
## PR checklist

- [x] If you've added a new tool, add `--profile docker,wave` tests.
- [x] Ensure the test suite passes (`nf-test test`).

## Description of changes

Migrates the **whamg** module to the nf-core v4.0.1 architecture:

- **topic: versions**: Replaced legacy `versions.yml` file emission with dynamic `eval()` tuple emission on the `topic: versions` channel.
- **stub block**: Added a deterministic `stub:` block for CI/CD stub-run support.
- **sanitizeOutput**: Refactored all test assertions to use `sanitizeOutput(process.out)` per nf-test best practices.
- **meta.yml**: Updated to include both `output: versions_whamg` and `topics: versions` sections.

Part of the v4.0.1 standardization effort tracked in #11323.

Part of #4570


modules/nf-core/whamg/tests/main.nf.test
```diff
@@ -30,13 +30,34 @@ nextflow_process {
         then {
             assertAll(
                 { assert process.success },
-                { assert snapshot(
-					path(process.out.vcf[0][1]).vcf.summary,
-					file(process.out.tbi[0][1]).name,
```
SPPearce, line 35: vcf and tbi files are typically unstable due to differences between conda/docker/singularity.
The vcf summary command comes from the [nft-vcf](https://github.com/seppinho/nft-vcf) plugin.
Any time you see a snapshot split out like this it means you need to take special care and not just snapshot the whole process.out.
You can use unstableKeys for the tbi channel, but the vcf summary is useful to keep for showing the vcf file is not changing.

modules/nf-core/whamg/main.nf
```diff
@@ -16,7 +16,7 @@ process WHAMG {
     tuple val(meta), path("*.vcf.gz")       , emit: vcf
     tuple val(meta), path("*.vcf.gz.tbi")   , emit: tbi
     tuple val(meta), path("*.txt")          , emit: graph, optional: true
-    path "versions.yml"                     , emit: versions
+    tuple val("${task.process}"), val('whamg'), eval('whamg 2>&1 | grep Version | sed \'s/^Version: v//; s/-.*$//\''), topic: versions, emit: versions_whamg
```
SPPearce: Can you make this one sed command please.

SPPearce (changes requested):

modules/nf-core/whamg/tests/main.nf.test
```diff
@@ -30,13 +30,34 @@ nextflow_process {
...
-					process.out.graph,
-					process.out.versions
-					).match()
-				}
+                { assert snapshot(sanitizeOutput(process.out)).match() }
```
SPPearce: Still need to swap back to use nft-vcf for the vcf file, and the tbi is unstable

HReed1:
@SPPearce — Thanks for the follow-up. Both points have been addressed and pushed!

1. **Version Command:** 
   Simplified the extraction from `grep | sed` into a single `sed -n 's/^Version: v\([^-]*\).*/\1/p'` command as requested. `meta.yml` has been updated to match.

2. **VCF / TBI Snapshot Stability:** 
   I swapped the blanket `sanitizeOutput` back to targeted channel assertions to properly handle the unstable VCF headers and `.tbi` hashes across runtimes. We are now using a hybrid approach that explicitly checks both:
   * **VCF:** Safely asserted using the `nft-vcf` plugin (`path(it[1]).vcf.summary`) to ignore header drift.
   * **TBI:** Explicitly asserted by capturing just its filename (`file(it[1]).name`). This ensures the test still strictly verifies the index was generated without failing on the non-deterministic MD5 hash.

SPPearce (approved):

PR #11513, opened 2026-05-04
Add module to fetch SDRF files from PRIDE Archive

jonasscheid:
## PR checklist

- [x] This comment contains a description of changes (with reason).
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [x] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [x] If necessary, include test data in your PR.
- [x] Remove all TODO statements.
- [x] Broadcast software version numbers to `topic: versions` - [See version_topics](https://nf-co.re/blog/2025/version_topics)
- [x] Follow the naming conventions.
- [x] Follow the parameters requirements.
- [x] Follow the input/output options guidelines.
- [x] Add a resource `label`
- [x] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - For modules:
    - [x] `nf-core modules test <MODULE> --profile docker`
    - [x] `nf-core modules test <MODULE> --profile singularity`
    - [x] `nf-core modules test <MODULE> --profile conda`

## Notes

Version is currently captured via `python -c "from importlib.metadata import version; ..."` because the bundled `pridepy` 0.0.15 CLI does not expose a `--version` flag. Upstream PR [PRIDE-Archive/pridepy#96](https://github.com/PRIDE-Archive/pridepy/pull/96) adds `--version` and has been merged; once a new `pridepy` release is published on bioconda, version parsing in this module will be simplified to `pridepy --version` in a follow-up.

PR #11514, opened 2026-05-04
New module: parsesdrf/convert

jonasscheid:
Adds a new module `sdrfpipelines/parsesdrf` wrapping the `parse_sdrf` CLI from [bigbio/sdrf-pipelines](https://github.com/bigbio/sdrf-pipelines) (bioconda `sdrf-pipelines=0.1.3`). The module converts an SDRF (Sample and Data Relationship Format) file into pipeline-specific samplesheets and search-preset files.
The conversion subcommand (e.g. `convert-mhcquant`, `convert-openms`) and its output flags are injected via `task.ext.args` rather than hardcoded, so the same module can serve different downstream proteomics pipelines. Pairs naturally with #11513

## PR checklist

- [x] Add a resource `label`
  - For modules:


modules/nf-core/sdrfpipelines/parsesdrf/main.nf
LouisLeNezet: As both output are expected they should be hardcoded in the script block.
Could you also use prefix = meta.id ?

modules/nf-core/sdrfpipelines/parsesdrf/tests/nextflow.config
LouisLeNezet: Please pass convert-mhcquant as params.module_args

LouisLeNezet (changes requested):
Hi,

Could you follow the nf-core convention with the meta map ?
This will allow to better control how the output tsv are named.

LouisLeNezet:
Also there is other files that can be passed to [`parse-sdrf`](https://github.com/bigbio/sdrf-pipelines/blob/main/COMMANDS.md) could you add them as input ?
But after further reading it might be better to name the tool `PARSESDRF_CONVERTMHCQUANT` ?

jonasscheid:
Thanks @LouisLeNezet for the hint, I initially wasn't sure how exactly to do it since the tool is very versatile, but I think you are right to rename to PARSESDRF_CONVERTMHCQUANT.

Edit: Should it be SDRFPIPELINES_CONVERTMHCQUANT or PARSESDRF_CONVERTMHCQUANT @LouisLeNezet 🤔

LouisLeNezet:
> Edit: Should it be SDRFPIPELINES_CONVERTMHCQUANT or PARSESDRF_CONVERTMHCQUANT

Good question :thinking:

jonasscheid:
Ok suggestion: Since CONVERTMHCQUANT is basically a pipeline specific module (local module), I guess it makes more sense to have a more generic nf-core module. Maybe makes more sense to cover e.g. the "convert-" subcommands (convert-openms, convert-mhcquant etc.) What do you think about SDRPIPELINES_CONVERT?

LouisLeNezet:
> What do you think about SDRPIPELINES_CONVERT?

I would agree at first thought, by adding a `val(format)`, but I prefer to ask first other maintainers to be sure !

LouisLeNezet:
After discussing with other maintainers, `PARSESDRF_CONVERT` might be best.
I have to admit that the people naming the package and tool could have choose only one :smile:

jonasscheid:
>  I have to admit that the people naming the package and tool could have choose only one 😄

💯 😄 

Thanks for the effort! Really appreciated

jonasscheid:
@LouisLeNezet could you have another look again if that was the idea you had in mind?

modules/nf-core/parsesdrf/convert/environment.yml
```diff
@@ -0,0 +1,7 @@
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/environment-schema.json
+channels:
+  - conda-forge
+  - bioconda
+dependencies:
+  - "bioconda::sdrf-pipelines=0.1.4"
```
LouisLeNezet:
```suggestion
  - bioconda::sdrf-pipelines=0.1.4
```

modules/nf-core/parsesdrf/convert/main.nf
```diff
@@ -0,0 +1,81 @@
...
+        'https://depot.galaxyproject.org/singularity/sdrf-pipelines:0.1.4--pyhdfd78af_0':
+        'quay.io/biocontainers/sdrf-pipelines:0.1.4--pyhdfd78af_0' }"
+
+    input:
+    tuple val(meta), path(sdrf)
+    val format
+
+    output:
+    tuple val(meta), path("${prefix}_samplesheet.tsv"), path("${prefix}_presets.tsv"),             emit: mhcquant,     optional: true
+    tuple val(meta), path("${prefix}_samplesheet.tsv"), path("${prefix}_experimental_design.tsv"), emit: openms,       optional: true
```
LouisLeNezet, line 16: I didn't knew that both files was need for the channel to be emitted.
This is a nice feature in this case.

modules/nf-core/parsesdrf/convert/tests/main.nf.test
```diff
@@ -0,0 +1,338 @@
...
+            )
+        }
+
+    }
+
+    test("sdrf - maxquant - PXD009752") {
+
+        when {
+            params {
+                module_args = '-f PXD009752.sdrf.tsv -r .'
```
LouisLeNezet: Why is the path passed through external arguments here ?

modules/nf-core/parsesdrf/convert/meta.yml
```diff
@@ -0,0 +1,168 @@
...
+        pattern: "*.{tsv,sdrf.tsv}"
+        ontologies:
+          - edam: http://edamontology.org/format_3475
+  - format:
+      type: string
+      description: |
+        Target converter. One of: `mhcquant`, `openms`, `maxquant`,
+        `msstats`, `normalyzerde`, `diann`. Selects the
+        `parse_sdrf convert-<format>` subcommand. For `maxquant`, the
+        required `--fastafilepath` and `--raw_folder` arguments must be
```
LouisLeNezet: This shouldn't be the case, all files and path should be passed as input channel.
They could be empty for all format except maxquant, where they are expected with a small error.

LouisLeNezet (changes requested):
Just a few comments but otherwise the modules seems solid !
Great work

jonasscheid:
Thanks again! Updated per your comments

LouisLeNezet (approved):
This is a complex modules, nice job figuring it out 👍 
LGTM

jonasscheid:
Thanks for the thorough review!

PR #11526, opened 2026-05-05
fix(tximeta/tximport, quant_tximport_summarizedexperiment, quantify_pseudo_alignment): publish augmented tx2gene actually used by tximport

pinin4fjords:
## Summary

`TXIMETA_TXIMPORT` silently augments the user-supplied tx2gene table with self-mappings for orphan transcripts (transcripts present in `quant.sf` / `abundance.tsv` / `*.isoforms.results` but missing from the input tx2gene), then uses that augmented table to compute the gene-level outputs. Only the input tx2gene was retained downstream, so users feeding it back into `tximport()` got different gene-level numbers - orphan transcripts were dropped on their side, perturbing the per-sample library-size totals that feed `countsFromAbundance` scaling.

This PR:

- **`tximeta/tximport`**
  - Adds a `tx2gene_augmented` output with the full mapping table actually used by tximport. Input column headers are preserved, so the file plugs straight back into `tximport()`.
  - Adds a new `mismatched_transcripts` test that truncates the input tx2gene and asserts (a) gene rows < transcript rows (filter still works), (b) augmented tx2gene rows == transcript count (one mapping per quantified transcript), (c) the warning is emitted.
  - Fixes a `lengths_transcript` `meta.yml` `pattern` typo (`*gene_lengths.tsv` -> `*transcript_lengths.tsv`).
  - Switches the new test's tx2gene-truncation expression to `lines.size().intdiv(2)` so it passes under Nextflow 26 strict mode (the older `(int)(... / 2)` cast triggers `Unknown method invocation 'call' on int type`).
- **`quant_tximport_summarizedexperiment`** - propagates the new `tx2gene_augmented` channel through the subworkflow's `emit:` block, plus matching test/snapshot updates.
- **`quantify_pseudo_alignment`** - propagates `tx2gene_augmented` one level further so pipelines consuming the top-level subworkflow can access it.

Mirrors nf-core/rnaseq#1831 - rnaseq has been carrying a local copy of all three components with these edits while waiting on this update.

## PR checklist

- [x] Your changes pass the existing tests on the testing dataset.
- [x] Bump versions when components have changed.
- [ ] Have you followed the [pipeline conventions](https://nf-co.re/docs/contributing/pipelines/pipeline_conventions) in the [contribution docs](https://nf-co.re/docs/contributing/modules)
- [x] If necessary, also make a PR on the nf-core/modules _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets/tree/modules) repository. (n/a)

### DIFF
--- a/modules/nf-core/tximeta/tximport/main.nf
+++ b/modules/nf-core/tximeta/tximport/main.nf
@@ -21,6 +21,7 @@ process TXIMETA_TXIMPORT {
     tuple val(meta), path("*transcript_tpm.tsv")           , emit: tpm_transcript
     tuple val(meta), path("*transcript_counts.tsv")        , emit: counts_transcript
     tuple val(meta), path("*transcript_lengths.tsv")       , emit: lengths_transcript
+    tuple val(meta), path("*tx2gene_augmented.tsv")        , emit: tx2gene_augmented
     path "versions.yml"                                    , emit: versions, topic: versions
 
     when:
@@ -39,6 +40,7 @@ process TXIMETA_TXIMPORT {
     touch ${meta.id}.transcript_tpm.tsv
     touch ${meta.id}.transcript_counts.tsv
     touch ${meta.id}.transcript_lengths.tsv
+    touch ${meta.id}.tx2gene_augmented.tsv
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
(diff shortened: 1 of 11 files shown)


FriederikeHanssen (approved):

pinin4fjords:
Thanks @FriederikeHanssen !

PR #11532, opened 2026-05-05
Fix dumpParametersToJson to be compatible with parameter typing

nvnieuwk:
This PR fixes the `dumpParametersToJson` to be compatible with the new parameter typing introduced in the new strict syntax parser. 

Main changes:
- Removed usage of `File()` and fully moved to the built-in nextflow `file()` function. Used `.resolve` instead of String additions as this can cause unexpected issues on cloud systems
- Added type conversions for all possible parameter types that aren't compatible with JSON generation

### DIFF
--- a/subworkflows/nf-core/utils_nextflow_pipeline/main.nf
+++ b/subworkflows/nf-core/utils_nextflow_pipeline/main.nf
@@ -73,7 +73,7 @@ def getWorkflowVersion() {
 def dumpParametersToJSON(outdir) {
     def timestamp = new java.util.Date().format('yyyy-MM-dd_HH-mm-ss')
     def filename  = "params_${timestamp}.json"
-    def temp_pf   = workflow.launchDir.resolve(".${filename}")
+    def temp_pf       = workflow.launchDir.resolve(".${filename}")
     def jsonGenerator = new groovy.json.JsonGenerator.Options()
         .excludeNulls()
         .addConverter(Path) { Path path -> path.toUriString() }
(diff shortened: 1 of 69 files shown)


subworkflows/nf-core/utils_nextflow_pipeline/main.nf
```diff
@@ -73,11 +73,23 @@ def getWorkflowVersion() {
 def dumpParametersToJSON(outdir) {
     def timestamp = new java.util.Date().format('yyyy-MM-dd_HH-mm-ss')
     def filename  = "params_${timestamp}.json"
-    def temp_pf   = new File(workflow.launchDir.toString(), ".${filename}")
-    def jsonStr   = groovy.json.JsonOutput.toJson(params)
+    def temp_pf   = workflow.launchDir.resolve(".${filename}")
+    def jsonGenerator = new groovy.json.JsonGenerator.Options()
```
mashehu:
```suggestion
    def temp_pf       = workflow.launchDir.resolve(".${filename}")
    def jsonGenerator = new groovy.json.JsonGenerator.Options()
```

mashehu (commented):
I guess similar things need to be done in  https://github.com/nf-core/nf-core-utils/tree/main

mashehu (approved):

nvnieuwk:
@nf-core-bot fix linting

mashehu:
this is a nextflow lint error, which the bot can't fix atm

nvnieuwk:
Allright I'll see if it breaks anything otherwise I'll just force merge it

nvnieuwk:
Nextflow 26.04.0 is needed for this, is this okay to merge?

mashehu:
hmm, that should be documented in the subworkflow. 
actually not sure if we should merge this like this. People will get a lint message that it is outdated for them, but then they would also need to update the template version...

mashehu:
what if we only update this in the plugin and point people to use the plugin if they are using 26.04?

nvnieuwk:
Never mind, I got it working with the currently used version :)

nvnieuwk:
> what if we only update this in the plugin and point people to use the plugin if they are using 26.04?

I'd rather have both until the plugin is in the template, I'll open a PR with the same changes to the plugin though :)

PR #11539, opened 2026-05-06
Add cli_typecasting option to the nfschema subwf

nvnieuwk:
This adds an option introduced in the recent nf-schema versions. Older versions will ignore this option so this remains backwards compatible


mashehu (approved):

nvnieuwk:
Lint failure is solved by bumping to 26.04.0 so it isn't an issue now

PR #11546, opened 2026-05-06
Add `fgumi/extract`

atrigila:
For: https://github.com/nf-core/smrnaseq/issues/535


modules/nf-core/fgumi/extract/main.nf
```diff
@@ -0,0 +1,39 @@
...
+    output:
+    tuple val(meta), path("*.bam"), emit: bam
+    tuple val("${task.process}"), val('fgumi'), eval('fgumi --version | sed "s/^fgumi //"'), topic: versions, emit: versions_fgumi
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
```
SPPearce, line 22: Can we please set a default prefix, so that the input and output files don't automatically clash.
You should also add the disambiguation check too.

atrigila: I can do that, however, the input and output files are different types, so they shouldn't theoretically clash.

SPPearce (commented):
Oh, ok, my apologies, I thought it was taking ubam.
You can remove the default prefix then, if it isn't going to collide within the module.
You shouldn't need to set prefix in the test at that point though.

modules/nf-core/fgumi/extract/main.nf
```diff
@@ -0,0 +1,39 @@
...
+    if (!args.contains('--sample') || !args.contains('--library')) {
+        error("fgumi extract requires both --sample and --library to be supplied via task.ext.args")
+    }
```
SPPearce: Hmm, if these are required arguments, they should probably be part of the input tuple.

SPPearce: https://nf-co.re/docs/specifications/components/modules/general#non-file-mandatory-command-arguments

atrigila: Thank you, I've incorporated this.

apeltzer (approved):

PR #11550, opened 2026-05-07
PROOF OF CONCEPT: samtools/multicommand

prototaxites:
See discussion on the nf-core slack: https://nfcore.slack.com/archives/C043UU89KKQ/p1777986494144269

Here is a working example of a "samtools/multicommand" module that constructs a chain of samtools commands that pipe into each other following a user-specified input describing which tools to use.

Currently the supported commands are only those which write *AM or fasta/fastq - I think that limits the scope helpfully. If you want to run samtools stats, for example, pass the output from this into that dedicated module.

```
# valid commands:
['view', 'sort', 'markdup', 'fixmate', 'merge', 'cat', 'collate', 'fastq', 'fasta']
```

The exact order of operations is specified by the pipeline input, which takes a list such as [view, sort, markdup]. This is then interpolated to a script where the output of each command is streamed to the next before being written to disk. The script can handle references being passed, as well as writing indexes, outputting in CRAM format, and conversion to FASTA and FASTQ.

By default, the threads for each tool are set to the total number of available cpus. However, it may be desirable to configure the number of threads used for each step, for example to give more threads to `samtools sort` than to `samtools view`. This can be achieved by specifying a number of threads in `$argsX` with `-@`, which will override the number of threads for that command. There is no sanity check that the number of threads is less than the total number available when doing this.

## PR checklist

Closes #XXX 

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:


modules/nf-core/samtools/multicommand/main.nf
```diff
@@ -0,0 +1,169 @@
...
+        }
+        if (idx == n_commands - 1) {
+            if (output_reference) {
+                cmd_parts << output_reference
+            }
+            cmd_parts << output_string
+        }
+
+        return cmd_parts.join(" ")
+    }.join(" |\\\n")
```
muffato: You just pipe the commands directly. Don't you need to add some `-` or `-o -` to force stdin/stdout ? I did that in my version https://github.com/nf-core/modules/pull/4571/changes#diff-34320bc32138ad0bc2d6a277f60183a5c4fc988b01314cab05ac6ff6245a9c94R38 , but perhaps that was unnecessary.

modules/nf-core/samtools/multicommand/main.nf
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    // Build the pipeline command
+    def pipeline_command = pipeline.withIndex().collect { subcommand, idx ->
+        def argsKey = idx == 0 ? "args" : "args${idx + 1}"
+        def taskArgs = task.ext[argsKey] ?: ""
+
+        def cmd_parts = ["samtools", subcommand]
+        if (taskArgs) {
+            cmd_parts << taskArgs
```
muffato: Add the number of CPUs ([except for the `reheader` command](https://github.com/nf-core/modules/pull/4571/changes#diff-34320bc32138ad0bc2d6a277f60183a5c4fc988b01314cab05ac6ff6245a9c94R48))

modules/nf-core/samtools/multicommand/main.nf
```diff
@@ -0,0 +1,169 @@
...
+            }
+            cmd_parts << (input instanceof List ? input.join(" ") : input)
+        }
+        if (idx == n_commands - 1) {
+            if (output_reference) {
+                cmd_parts << output_reference
+            }
+            cmd_parts << output_string
+        }
+
```
muffato, line 77: I thought `-u` was the recommended setting when piping commands, cf https://www.htslib.org/algorithms/duplicate.html#workflow

modules/nf-core/samtools/multicommand/main.nf
```diff
@@ -0,0 +1,169 @@
...
+        }
+    } else if (final_command == "fastq") {
+        // fastq produces multiple files with special output flags
+        output_string = "-0 ${prefix}_other.fastq.gz"
+        if (!meta.single_end) {
+            output_string = output_string + " -1 ${prefix}_1.fastq.gz -2 ${prefix}_2.fastq.gz -s ${prefix}_singleton.fastq.gz"
+        } else {
+            output_string = output_string + " -1 ${prefix}_1.fastq.gz -s ${prefix}_singleton.fastq.gz"
+        }
+    }
```
muffato: I think these specific fasta/q deserve dedicated output channels (you can keep the existing `fasta` and `fastq` as catchall)

modules/nf-core/samtools/multicommand/main.nf
```diff
@@ -0,0 +1,169 @@
...
+
+    def valid_options = ['view', 'sort', 'markdup', 'fixmate', 'merge', 'cat', 'collate', 'fastq', 'fasta']
+    pipeline.collect { tool ->
+        if (!(tool in valid_options)) {
+            error("Error: ${tool} not a valid pipeline argument for SAMTOOLS_PIPELINE! Valid options are: ${valid_options.join(", ")}")
+        }
+    }
+
+    def n_commands = pipeline.size()
+    def final_command = pipeline[n_commands - 1]
```
muffato: Check that there is at least 1 command (maybe even 2)

modules/nf-core/samtools/multicommand/main.nf
prototaxites: AFAIK it's not mandatory in samtools?

modules/nf-core/samtools/multicommand/main.nf
prototaxites: sorted

modules/nf-core/samtools/multicommand/main.nf
prototaxites: added to all the commands except cat, which doesn't support `-u`

modules/nf-core/samtools/multicommand/main.nf
prototaxites: oops, I meant to do that. added now!

modules/nf-core/samtools/multicommand/main.nf
muffato: I don't know :) I don't use samtools often enough

modules/nf-core/samtools/multicommand/main.nf
```diff
@@ -0,0 +1,185 @@
...
+        if (meta.single_end) {
+            stub_outputs << "echo | gzip > ${prefix}_1.fastq.gz"
+            stub_outputs << "echo | gzip > ${prefix}_singleton.fastq.gz"
+        } else {
+            stub_outputs << "echo | gzip > ${prefix}_1.fastq.gz"
+            stub_outputs << "echo | gzip > ${prefix}_2.fastq.gz"
+            stub_outputs << "echo | gzip > ${prefix}_singleton.fastq.gz"
+        }
+        stub_outputs << "echo | gzip > ${prefix}_other.fastq.gz"
+    }
```
muffato: Switch to `bgzip` to mimic samtools ?

prototaxites: I think that would foul expectations from linting

modules/nf-core/samtools/multicommand/main.nf
prototaxites: This was in fact the case, but not consistently...!

subworkflows/nf-core/bam_markduplicates_samtools_module_replacement/meta.yml
```diff
@@ -0,0 +1,38 @@
...
+        Groovy Map containing sample information
+        e.g. `[ id:'test' ]`
+  - bam:
+      type: file
+      description: Sorted and markduplicate BAM/CRAM/SAM file
+      pattern: "*.{bam,cram,sam}"
+authors:
+  - "@priyanka-surana"
+maintainers:
+  - "@priyanka-surana"
```
muffato: Give yourself more credit, @prototaxites ;)

modules/nf-core/samtools/multicommand/main.nf
```diff
@@ -223,3 +140,112 @@ process SAMTOOLS_MULTICOMMAND {
...
+        parts << reference
+    }
+    if (output) {
+        parts << output
+    }
+    if (uncompressed) {
+        parts << uncompressed
+    }
+
+    return parts.findAll().join(" ")
```
muffato: Since each element added to `parts` is controlled by its own `if`, can `findAll()` return anything but `parts` itself ?

modules/nf-core/samtools/multicommand/main.nf
```diff
@@ -223,3 +140,112 @@ process SAMTOOLS_MULTICOMMAND {
...
+            return "-O"
+        }
+        else if (subcommand in ["view", "sort", "cat"]) {
+            return ""
+        }
+        return "-"
+    }
+
+    def argsKey = n_commands == 1 ? "args" : "args${n_commands}"
+    def argsLast = task.ext[argsKey] ?: ""
```
muffato: Feels redundant with `get_args`

modules/nf-core/samtools/multicommand/main.nf
```diff
@@ -223,3 +140,112 @@ process SAMTOOLS_MULTICOMMAND {
...
+// ============================================================================
+
+def get_input(is_first, input) {
+    if (is_first) {
+        return input.join(" ")
+    }
+    return "-"
+}
+
+def get_output(subcommand, is_last, final_command, task, n_commands, prefix, single_end) {
```
muffato: This function interface confuses me.

- When called on the last sub-command, `is_last` is _true_ and `subcommand`==`final_command` ?
- `task` and `n_commands` seem to be only used for the last sub-command. Maybe make the caller call `get_args` itself and pass the args ?
- Maybe split it out in two functions, `get_intermediate_output` and `get_last_output`, since they have distinct implementations.

prototaxites: I got Copilot to do most of the above, it didn't really understand the brief. With a fresh brain I've refactored it more myself, hopefully the logic is much clearer.

modules/nf-core/samtools/multicommand/main.nf
prototaxites: It actually doesn't! I have changed it to `bgzip`.

modules/nf-core/samtools/multicommand/main.nf
```diff
@@ -0,0 +1,287 @@
+process SAMTOOLS_MULTICOMMAND {
+    tag "${meta.id}"
+    label 'process_medium'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine in ['singularity', 'apptainer'] && !task.ext.singularity_pull_docker_container
+?         'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/c2/c2f02ceb8cbe5aa37930dcbc0de6f1d07dfb292178c7e780c44ab2f556bf829f/data'
+:         'community.wave.seqera.io/library/htslib_samtools:1.24--84fe6b04a60dfab2' }"
```
muffato: The placement of `?` and `:` seems off

modules/nf-core/samtools/multicommand/main.nf
```diff
@@ -0,0 +1,287 @@
...
+}
+
+// ============================================================================
+// Helper Functions
+// ============================================================================
+
+
+// ============================================================================
+// Helper Functions
+// ============================================================================
```
muffato:
```suggestion
```

modules/nf-core/samtools/multicommand/main.nf
```diff
@@ -0,0 +1,287 @@
...
+    def valid_options = ['view', 'sort', 'markdup', 'fixmate', 'merge', 'cat', 'collate', 'fastq', 'fasta']
+    pipeline.collect { tool ->
+        if (!(tool in valid_options)) {
+            error("Error: ${tool} not a valid pipeline argument for SAMTOOLS_MULTICOMMAND! Valid options are: ${valid_options.join(", ")}")
+        }
+    }
+
+    def n_commands = pipeline.size()
+    def is_cram_input = fasta && input.collect { f -> f.getExtension() == "cram" }.any()
+    def is_cram_output = fasta && get_args(task, n_commands - 1)
```
muffato: calling `get_args` but not doing anything with it 🤔 ? Did you mean something like
```suggestion
    def is_cram_output = fasta && get_output_extension(get_args(task, n_commands - 1)) == "cram"
```

prototaxites: Yes 🤦‍♂️

modules/nf-core/samtools/multicommand/main.nf
prototaxites: it's nf-core tools's new container tool - currently writes bad formatting

muffato (approved):
Good for me !

prototaxites:
I will see if anyone else wants to weigh in before I merge!

prototaxites:
This doesn’t catch the case that we have CRAM inputs or outputs but no reference supplied. It should gracefully error.

muffato:
> This doesn’t catch the case that we have CRAM inputs or outputs but no reference supplied. It should gracefully error.

How / where exactly ? CRAM can work without an explicit reference too (htslib downloads from the ENA server – wait: has this default behaviour been removed in a recent version ?). There are reference-free, unaligned CRAMs as well.

prototaxites:
Ah, that’s a fair point - I forgot about that functionality.

SPPearce:
> This doesn’t catch the case that we have CRAM inputs or outputs but no reference supplied. It should gracefully error.

This should error with a relatively clear error message from samtools I think.

mahesh-panchal:
*Review by Claude, requested by @mahesh-panchal.*

**`get_reference` never fires for CRAM output.**

```groovy
def is_cram_output = fasta && (get_output_extension(get_args(task, n_commands - 1)) == "cram")
```

`get_args(ext, idx)` expects `task.ext`; this passes `task`. `task["argsN"]` doesn't reach the nested `ext` map, so `get_args` always returns `""`, `get_output_extension` always reads `"bam"`, and `is_cram_output` is always `false`. `--reference` never gets added when the last command writes CRAM. Doesn't fail because samtools accepts CRAM output without a reference, so it's silent. Compare the correct call at L71:

```diff
-def is_cram_output = fasta && (get_output_extension(get_args(task, n_commands - 1)) == "cram")
+def is_cram_output = fasta && (get_output_extension(get_args(task.ext, n_commands - 1)) == "cram")
```

**Per-tool `-@` override doesn't work.**

```groovy
def cmd_args = get_args(task.ext, idx)
def cmd_threads = get_threads(subcommand, task.cpus, args)
```

`args` isn't `cmd_args` — it's never assigned, so this reads whatever `args` resolves to outside this scope, not the tool's actual arguments. `get_threads` checks that value for `-@` to decide whether to skip appending `-@ ${cpus}`; since it's checking the wrong variable, `-@ ${cpus}` gets appended unconditionally, contradicting the PR description ("specifying `-@` in `$argsX`... will override the number of threads"). No test sets a `pipeline_argsN` containing `-@`, so this path is untested.

```diff
-def cmd_threads = get_threads(subcommand, task.cpus, args)
+def cmd_threads = get_threads(subcommand, task.cpus, cmd_args)
```

**`fasta_interleaved`/`fastq_interleaved` outputs are unreachable.**

`get_file_output()` never emits an interleaved-mode samtools flag for `fasta`/`fastq` — only `_1`/`_2`/singleton/other. Both interleaved output channels can never populate. `fastq_interleaved`'s glob is also missing `.gz` (L30: `*_interleaved.fastq`) unlike `fasta_interleaved`'s `*_interleaved.fasta.gz` (L26). Either wire up `-i` support or drop the two output declarations.

mahesh-panchal:
So a visual inspection for me on this module is way too complex. I definitely didn't pick up what Claude did. It needs to be easier for a person to review correctness.

prototaxites:
Thanks Mahesh!

I have fixed the first two issues.

> So a visual inspection for me on this module is way too complex. I definitely didn't pick up what Claude did. It needs to be easier for a person to review correctness.

I think a module like this is inherently going to be complex, as there are a lot of edge-cases that need to be caught. That's why I switched it to a command-building approach that makes liberal use of small helper functions to construct each command, which hopefully makes the logic more clean and configurable. 

I think there's one clear way in general to simplify this more, which would be to drop fastX outputs entirely, and only write SAM/BAM/CRAM. That would take a bunch of possibilities/complexities out of the module without compromising IMO its core function which is to process *AM files with a pipeline. That would also fix the interleaved problem, which arises because I cloned some of the code from the samtools/fast{a,q} modules without peering too hard and which actually needs an extra process input to work.

SPPearce:
> I think there's one clear way in general to simplify this more, which would be to drop fastX outputs entirely, and only write SAM/BAM/CRAM. That would take a bunch of possibilities/complexities out of the module without compromising IMO its core function which is to process *AM files with a pipeline. That would also fix the interleaved problem, which arises because I cloned some of the code from the samtools/fast{a,q} modules without peering too hard and which actually needs an extra process input to work.

This seems like a reasonable plan to me. The collate + fastq is served by samtools/collatefastq already.

mahesh-panchal:
I get that these kind of modules are going to be complex, but I think we somehow need to make it easier to read. Part of my issue is the function calls moving logic to lower down, and then having to hop around to figure out how things are connected. It may be a necessity to keep things clean though. 

I think testing needs to be more complete to compensate. It should cover every tool invoked, and cover every function invoked. At least one test for the minimal case too.

mahesh-panchal:
Just linking the older PR for reference too: https://github.com/nf-core/modules/pull/4571

prototaxites:
> I get that these kind of modules are going to be complex, but I think we somehow need to make it easier to read. Part of my issue is the function calls moving logic to lower down, and then having to hop around to figure out how things are connected. It may be a necessity to keep things clean though.
>
> Just linking the older PR for reference too: https://github.com/nf-core/modules/pull/4571

@muffato's version uses a switch on each potential tool (we could use a big if/elif/else given that switch is disallowed) to construct a command for each command explicitly. Do you think that's a more explicitly readable approach? It would lead to a bunch of duplicated code to catch a handful of exceptions. 

Personally I like functional approach. And once this is well-tested presumably it's likely to be as stable as samtools is and not require much maintenance.

> I think testing needs to be more complete to compensate. It should cover every tool invoked, and cover every function invoked. 

This is definitely true. The tests should also explicitly catch every exceptional case (e.g. `cat` not taking `-u` or `markdup` not taking a `-o` flag to write to disk).

> At least one test for the minimal case too.

Which minimal case?

mahesh-panchal:
Looking back at the code, I don't find it that easy to read either because it's difficult to see what's being thrown into the arrays (and the irony is I'm pretty sure that part was my suggestion). I'm honestly not sure how to handle it. 

My worry is not so much stability. It's adaptation. If someone wants to add a new feature, how easy would it be to get over the module complexity. Hence, why readability is part of my issue.

By minimal case, I meant the single tool use cases.

prototaxites:
> My worry is not so much stability. It's adaptation. If someone wants to add a new feature, how easy would it be to get over the module complexity. Hence, why readability is part of my issue.

Ultimately, I think this is just a limitation of writing Groovy inside a Nextflow process. There are a lot of programming features that are not very available here that could be useful (a command constructor class, perhaps), not to mention the fact that functions simply cannot be defined in the process section where they would be most discoverable to a casual reader.

In the face of this, I tried to be as explicit as possible with commenting and documenting the code. Perhaps I could do more on this front.

What say you to dropping the FASTX outputs?

> By minimal case, I meant the single tool use cases.

By design these fail at the moment, you need at least 2 commands. You should use the appropriate module.

mahesh-panchal:
> What say you to dropping the FASTX outputs?

I'm ok with leaving it out, but I would defer to heavier users of this.

> 
> > By minimal case, I meant the single tool use cases.
> 
> By design these fail at the moment, you need at least 2 commands. You should use the appropriate module.

Ah, yeah. True. So then any 2+ common chains, covering all subtools at least once.

mahesh-panchal:
What about using closures to move the functions in the process body?

```nextflow
    script:

    if (pipeline.size() <= 1) {
        error("Error: SAMTOOLS_MULTICOMMAND requires at least two commands!")
    }

    def valid_options = ['view', 'sort', 'markdup', 'fixmate', 'merge', 'cat', 'collate', 'fastq', 'fasta']
    pipeline.collect { tool ->
        if (!(tool in valid_options)) {
            error("Error: ${tool} not a valid pipeline argument for SAMTOOLS_MULTICOMMAND! Valid options are: ${valid_options.join(", ")}")
        }
    }

    def get_stdout = { subcommand ->
        if (subcommand == "collate") {
            return "-O"
        }
        else if (subcommand in ["view", "sort", "cat"]) {
            return ""
        }
        return "-"
    }

    def get_output_extension = { args ->
        if (args.contains("--output-fmt sam")) {
            return "sam"
        }
        else if (args.contains("--output-fmt cram")) {
            return "cram"
        }
        return "bam"
    }

    def get_file_output = { subcommand, args, single_end ->
        def extension = get_output_extension(args)

        if (subcommand in ['view', 'sort', 'merge', 'cat', 'fixmate', 'collate']) {
            return "-o ${prefix}.${extension}"
        }
        else if (subcommand == "markdup") {
            return "${prefix}.${extension}"
        }
        else if (subcommand in ['fasta', 'fastq']) {
            def seq_ext = subcommand == 'fasta' ? 'fasta.gz' : 'fastq.gz'
            def output = "-0 ${prefix}_other.${seq_ext} -1 ${prefix}_1.${seq_ext}"
            if (!single_end) {
                output += " -2 ${prefix}_2.${seq_ext}"
            }
            return "${output} -s ${prefix}_singleton.${seq_ext}"
        }

        error("Trying to get output for unsupported subcommand: ${subcommand}")
    }

    def get_args = { idx ->
        def argsKey = idx == 0 ? "args" : "args${idx + 1}"
        return task.ext[argsKey] ?: ""
    }

    def get_threads = { subcommand, args ->
        if (subcommand != "cat") {
            if (!args.contains("-@")) {
                return "-@ ${task.cpus}"
            }
        }
        return ""
    }

    def get_uncompressed_flag = { subcommand ->
        if (subcommand != "cat") {
            return "-u"
        }
        return ""
    }

    def get_reference = { is_first_command, is_last_command, is_cram_input, is_cram_output ->
        if ((is_cram_input && is_first_command) || (is_cram_output && is_last_command)) {
            return "--reference ${fasta}"
        }
        return ""
    }

    def n_commands = pipeline.size()
    def is_cram_input = fasta && input.collect { f -> f.getExtension() == "cram" }.any()
    def is_cram_output = fasta && (get_output_extension(get_args(n_commands - 1)) == "cram")

    def pipeline_command = pipeline
        .withIndex()
        .collect { subcommand, idx ->
            def is_first_command = (idx == 0)
            def is_last_command = (idx == n_commands - 1)

            def cmd_args = get_args(idx)
            def cmd_threads = get_threads(subcommand, cmd_args)
            def cmd_input = is_first_command ? "${input}" : "-"
            def cmd_output = is_last_command ? get_file_output(subcommand, cmd_args, meta?.single_end ?: false) : get_stdout(subcommand)
            def uncompressed = !is_last_command ? get_uncompressed_flag(subcommand) : ""
            def reference = get_reference(is_first_command, is_last_command, is_cram_input, is_cram_output)

            return ["samtools", subcommand, cmd_threads, uncompressed, cmd_args, reference, cmd_input, cmd_output].findAll().join(" ")
        }
        .join(" |\\\n")

    """
    ${pipeline_command}
    """


stub:
def final_command = pipeline[-1]
def final_args = get_args(task.ext, pipeline.size() - 1)

def stub_outputs = []

if (final_command in ['view', 'sort', 'merge', 'cat', 'markdup', 'fixmate', 'collate']) {
    stub_outputs << "touch ${prefix}.${get_output_extension(final_args)}"
}
else if (final_command in ['fasta', 'fastq']) {
    def ext = final_command == 'fasta' ? 'fasta.gz' : 'fastq.gz'
    stub_outputs << "echo | bgzip > ${prefix}_1.${ext}"
    if (!meta.single_end) {
        stub_outputs << "echo | bgzip > ${prefix}_2.${ext}"
    }
    stub_outputs << "echo | bgzip > ${prefix}_singleton.${ext}"
    stub_outputs << "echo | bgzip > ${prefix}_other.${ext}"
}

"""
${stub_outputs.join("\n")}
"""
```

Edit: Or maybe not. One just has to try to wrap their head around the complexities of the module.

prototaxites:
I do like that more, actually. It puts the functions in the scope of the script as you read it so you know what's coming before the command is constructed.

But you're right that you do just have to wrap your head around the module a bit. That's why I also made sure to write a good description in the meta.yaml, so that less code-inclined/more busy users would be able to just use it.

Edit: the linter doesn't like the closures, it says they can't be found within the `.collect {}` closure.

awgymer:
I think I agree with Mahesh that the actual implementation of this kind of module seems overly complex, to the point that I question how this is better than simply writing your own hardcoded local module that explicitly pipes tools together?

I'm not particularly convinced that this belongs as a directly accessible module in nf-core - it feels more like something which should be template which users could build on via something like `nf-core modules from-template piped/samtools` or something. 

From an end-user perspective I would agree that the code/logic is hard to parse and thus hard to debug, and so again I would likely reach for a local custom module instead. 

I don't particularly like the handling of threads here either. My issue is that in many cases with piped commands you do not want each command call to have the same threads, or for any one command to use all available threads, but equally setting these things for a user is not normally something we think about in `nf-core`. Ideally I think the default should be for commands to get `task.cpus/n_tasks` threads each, but again it depends on what the threads for each command are actually used for; some tools will take a threads args which is actually only used for i/o compression, some tools use it for compute, some tools have a soft cap (performance is static or decreases beyond e.g. 4 threads). This may not all apply to samtools specifically but I think if this is intended to be a scoping exercise for this sort of module more generally it is worth noting.

modules/nf-core/samtools/multicommand/main.nf
```diff
@@ -0,0 +1,265 @@
...
+                error("Error: ${subcommand} not a valid pipeline argument for SAMTOOLS_MULTICOMMAND! Valid options are: ${valid_options.join(", ")}")
+            }
+            if (subcommand in ["fastq", "fasta"] && idx != pipeline.size() - 1) {
+                error("Error: ${subcommand} must be the last command for SAMTOOLS_MULTICOMMAND!")
+            }
+        }
+
+    // Input/output flags
+    def is_cram_input = fasta && input.collect { f -> f.getExtension() == "cram" }.any()
+    def is_cram_output = fasta && (get_output_extension(get_args(task.ext, n_commands - 1)) == "cram")
```
muffato, line 64: Should this be overridden by the last subcommand being `fastq`/`fasta` ?

prototaxites: It's just ignored in this case currently

mahesh-panchal:
If you want to use the closures, you need to add in the `.call()`. I should have checked that. 

```nextflow
    def pipeline_command = pipeline
        .withIndex()
        .collect { subcommand, idx ->
            def is_first_command = (idx == 0)
            def is_last_command = (idx == n_commands - 1)

            def cmd_args = get_args.call(idx)
            def cmd_threads = get_threads.call(subcommand, cmd_args)
            def cmd_input = is_first_command ? "${input}" : "-"
            def cmd_output = is_last_command ? get_file_output.call(subcommand, cmd_args, meta?.single_end ?: false) : get_stdout.call(subcommand)
            def uncompressed = !is_last_command ? get_uncompressed_flag.call(subcommand) : ""
            def reference = get_reference.call(is_first_command, is_last_command, is_cram_input, is_cram_output)

            return ["samtools", subcommand, cmd_threads, uncompressed, cmd_args, reference, cmd_input, cmd_output].findAll().join(" ")
        }
        .join(" |\\\n")
```

prototaxites:
> I think I agree with Mahesh that the actual implementation of this kind of module seems overly complex, to the point that I question how this is better than simply writing your own hardcoded local module that explicitly pipes tools together?

> From an end-user perspective I would agree that the code/logic is hard to parse and thus hard to debug, and so again I would likely reach for a local custom module instead.

Obviously I'm biased as I wrote it, but I really don't think the code is that complex, or that complex code doesn't belong in nf-core. I suspect that whether you prefer a hard-coded local module over this would be a matter of personal preference - I certainly think I'd be reaching for this as a first point of call and only going local if there were additional complexities that needed implementing.

> I'm not particularly convinced that this belongs as a directly accessible module in nf-core - it feels more like something which should be template which users could build on via something like nf-core modules from-template piped/samtools or something.

I think that would potentially be worse from a code complexity standpoint - the complexity just gets moved behind the scenes in nf-core tools.

> I don't particularly like the handling of threads here either. My issue is that in many cases with piped commands you do not want each command call to have the same threads, or for any one command to use all available threads, but equally setting these things for a user is not normally something we think about in nf-core. Ideally I think the default should be for commands to get task.cpus/n_tasks threads each, but again it depends on what the threads for each command are actually used for; some tools will take a threads args which is actually only used for i/o compression, some tools use it for compute, some tools have a soft cap (performance is static or decreases beyond e.g. 4 threads). This may not all apply to samtools specifically but I think if this is intended to be a scoping exercise for this sort of module more generally it is worth noting.

What I've written is IMO the compromise position here. I would note that existing nf-core modules that implement piped commands (e.g. sormadup: https://github.com/nf-core/modules/blob/master/modules/nf-core/samtools/sormadup/main.nf) also just supply `task.cpus`to each command. I think this is better than dividing the number of cpus by the number of jobs. It has its downsides (if you bump the number of cpus in a config, and the pipeline has customised `args` to set CPUs for each process, you will have to change those too) but I don't think there's an easy way to avoid this.

awgymer:
> the complexity just gets moved behind the scenes in nf-core tools

But importantly, not at the point that a user has to interact with it. I think that ultimately my preference in nf-core is to avoid potential footguns as much as possible. 

For the CPUs issue, this is not necessarily really a criticism of this specific module, rather I'd argue the precedent in nf-core is wrong/problematic and this is just my latest attempt to stop it continuing.

prototaxites:
So there were suggestions for changes to:

1. set an "input command" option, that takes one of [null, "cat", "merge"] and sets the input command accordingly (cat and merge both take multiple input files)
2. set an "output format" option that takes one of [sam, bam, cram, fasta, fastq] - with the idea being that if a FASTX option is selected, that command is appended to the end of the pipeline, and otherwise provides a value for `--output-fmt`.

While I see how these might help to simplify the interface, I think on the whole these would make specifying args for the module a lot more complicated. At the moment, the first command gets `args`, and the last command gets `argsN`, and it's easy to know which is which because the list of tools is specified as a single string. By modifying the list of commands with other arguments, this will make understanding which tool takes which arg more complicated. This would also preclude adding formatting options to --output-fmt.

I also received some push-back that FASTX outputs would be very useful - so maybe they're not actually worth dropping from the module!

PR #11615, opened 2026-05-12
Add primerprospector_analyzeprimers

vagkaratzas:
Workaround shim in main:
```
Primer Prospector 1.0.1 (this is an old codebase) is a Python 2 package and the BioContainer entrypoint uses Python 2.7.
The bug is specifically Python 2 behavior: range() requires integer arguments, and the package passes a
numpy.float64 during plotting.
```

## PR checklist

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:


modules/nf-core/primerprospector/analyzeprimers/main.nf
```diff
@@ -0,0 +1,79 @@
...
+    else:
+        raise RuntimeError('Could not find analyze_primers.py on PATH')
+    PY
+
+    python analyze_primers_compat.py \\
+        $args \\
+        -f "${fasta_arg}" \\
+        ${primer_arg}
+
+    mv *_hits.txt "${prefix}_hits.txt"
```
timrozday-mgnify: For multiple primers in the input primer file this will cause an error or overwrite some results. Rather all the files should be emitted in the hits output channel. If prefix renaming is desired then it needs to be performed on each of the produced hits files.

timrozday-mgnify (changes requested):
Thanks @vagkaratzas ! This looks useful.

I have one major issue with this module relating to incorrect handling of output hit files when there are multiple primers in the input primer file (a common case). The tests should be amended with this case.

I think it would be useful to support gzipped fasta inputs, but I don't know what nf-core good practice says about that.

modules/nf-core/primerprospector/analyzeprimers/main.nf
vagkaratzas: Good catch, updated!

vagkaratzas:
> I think it would be useful to support gzipped fasta inputs, but I don't know what nf-core good practice says about that.

Since the module allows a list of fasta files (and not just a single fasta one), it is a bit more complicated to do that in the code (but definitely not impossilbe). But for now I guess running `GUNZIP` as an intermediate step can solve that case. The module is already a bit more complicated than usually (due to being old and possibly unreachable authors to update), so I'd leave the decompressing out of the current scope.

modules/nf-core/primerprospector/analyzeprimers/main.nf
```diff
@@ -0,0 +1,76 @@
...
+        -f "${fasta_arg}" \\
+        ${primer_arg}
+    """
+
+    stub:
+    def args = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "${meta.id}"
+    def arg_tokens = args.tokenize()
+    if (arg_tokens.any { arg -> arg == '-f' || arg.startsWith('--fasta_seqs') || arg == '-P' || arg.startsWith('--primers_filepath') || arg == '-o' || arg.startsWith('--output_dir') }) {
+        error "'-f/--fasta_seqs', '-P/--primers_filepath' and '-o/--output_dir' are reserved by this module. Use input files for fasta/primers and task.ext.prefix to control output file prefixes."
```
timrozday-mgnify:
```suggestion
        error "'-f/--fasta_seqs', '-P/--primers_filepath' and '-o/--output_dir' are reserved by this module. Use input files for fasta/primers."
```

modules/nf-core/primerprospector/analyzeprimers/main.nf
```diff
@@ -0,0 +1,76 @@
...
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "${meta.id}"
+    def fasta_arg = fasta instanceof List ? fasta.join(':') : fasta
+    def primer_arg = primers ? "-P \"${primers}\"" : ''
+    def arg_tokens = args.tokenize()
+    if (arg_tokens.any { arg -> arg == '-f' || arg.startsWith('--fasta_seqs') || arg == '-P' || arg.startsWith('--primers_filepath') || arg == '-o' || arg.startsWith('--output_dir') }) {
+        error "'-f/--fasta_seqs', '-P/--primers_filepath' and '-o/--output_dir' are reserved by this module. Use input files for fasta/primers and task.ext.prefix to control output file prefixes."
```
timrozday-mgnify:
```suggestion
        error "'-f/--fasta_seqs', '-P/--primers_filepath' and '-o/--output_dir' are reserved by this module. Use input files for fasta/primers."
```

timrozday-mgnify (approved):
Looks great, thank you @vagkaratzas!

Just two edits to the error messages since prefix is no longer used to rename output files.

I've approved so you can crack on with merging when you choose.

modules/nf-core/primerprospector/analyzeprimers/main.nf
vagkaratzas: Good catch, even though it's always nice to have the prefix to overwrite output names, I guess in this scenario, a name combination of sequence_name + primer_name makes more sense.

PR #11628, opened 2026-05-12
fix(shinyngs/validatefomcomponents): emit valid TSV in stub block

pinin4fjords:
## Summary

The `SHINYNGS_VALIDATEFOMCOMPONENTS` stub block emits empty files via `touch`. Pipelines that pipe its `contrasts_file.tsv` output through `splitCsv(header:true)` then fail with `IllegalStateException: Missing 'header' in CSV file`, taking the whole `-stub-run` down with them.

This PR replaces each `touch` with a `printf` that writes a minimal header-bearing TSV. Column names match what `validate_fom_components.R` actually emits in real runs, so downstream consumers see the same schema in stub and live mode.

Snapshot refreshed for the existing `test("stub")` case.

## Motivation

Reported in [nf-core/differentialabundance#717](https://github.com/nf-core/differentialabundance/issues/717): `-stub-run` is currently unusable for that pipeline, which is a real productivity hit during development.

## PR checklist

- [x] If you've fixed a bug or added code that should be tested, add tests! (existing stub test still exercises this path; snapshot updated.)
- [x] Ensure the test suite passes (`nf-core modules test shinyngs/validatefomcomponents --profile docker`).
- [x] Used the `nf-core modules test <module>` command to make sure the tests pass.
- [x] `main.nf` only changes the stub block; the `script` block is untouched.
- [x] Module follows the [DSL2 process guidelines](https://nf-co.re/docs/contributing/modules#dsl2-process-guidelines).

### DIFF
--- a/modules/nf-core/shinyngs/validatefomcomponents/main.nf
+++ b/modules/nf-core/shinyngs/validatefomcomponents/main.nf
@@ -42,10 +42,18 @@ process SHINYNGS_VALIDATEFOMCOMPONENTS {
     stub:
     def prefix = task.ext.prefix ?: meta.id
     """
-    mkdir $prefix
-    touch $prefix/${prefix}.sample_metadata.tsv
-    touch $prefix/${prefix}.feature_metadata.tsv
-    touch $prefix/${prefix}.assay.tsv
-    touch $prefix/${prefix}.contrasts_file.tsv
+    mkdir -p $prefix
+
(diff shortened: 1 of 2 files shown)


LouisLeNezet (approved):
Seems fine to me !

pinin4fjords:
Thanks @LouisLeNezet !

PR #11641, opened 2026-05-13
fix: propr/propd write properly headed genewise TSV

pinin4fjords:
## Summary

`propr/propd` writes the genewise results table using `write.table(..., row.names=TRUE, col.names=TRUE)`, which produces a header with N column names but data rows with N+1 fields (the row name + the N columns):

```
Nsig	LFC	lrmD	rcDdis	significant
ENSMUSG00000076613	981	-10.375	-20.868	0.001	0
```

This makes the file malformed for any downstream consumer that does positional/named column parsing (e.g. `csvtk join`, `read.delim` without explicit `row.names = 1`). This PR moves the gene IDs from rownames into an explicit `features_id_col` column and writes with `row.names = FALSE`.

The rewrite is guarded against the "no significant theta found" branch, which already constructs `results_genewise` with `features_id_col` as a regular column - applying the cbind there would produce a TSV header with two columns sharing that name.

The module test snapshot is re-recorded against nf-test 0.9.5 / Nextflow 26.04.1 to pick up the new TSV md5s.

## Why this matters

This bug only surfaced because the existing module test only verified the file by md5 - it never opens the TSV with anything that cares about header alignment. nf-core/differentialabundance is the (currently only) downstream consumer of `propr/propd`, and we discovered it while plumbing propd as a `--differential_method` option (see [nf-core/differentialabundance#724](https://github.com/nf-core/differentialabundance/issues/724) and its associated PR). With this fix the differentialabundance pipeline can drop a local patch and consume `propr/propd` from upstream directly.

## Test plan

- [x] `nf-core modules test propr/propd --profile docker` passes after re-recording the snapshot.
- [x] Header of `*.propd.genewise.tsv` has the same number of fields as data rows.
- [x] The empty-results branch (`theta_cutoff == FALSE`) also emits a properly aligned single-header TSV (guarded by `if (!(features_id_col %in% colnames(...)))`).
- [ ] `csvtk join --fields gene_id ...` against the genewise TSV finds the gene_id column (verified end-to-end via nf-core/differentialabundance#727 CI once that PR drops its local patch).

### DIFF
--- a/modules/nf-core/propr/propd/main.nf
+++ b/modules/nf-core/propr/propd/main.nf
@@ -20,7 +20,7 @@ process PROPR_PROPD {
     tuple val(meta), path("*.propd.adjacency.csv")        , emit: adjacency                , optional:true
     tuple val(meta), path("*.propd.fdr.tsv")              , emit: fdr                      , optional:true
     path "*.R_sessionInfo.log"                            , emit: session_info
-    path "versions.yml"                                   , emit: versions
+    path "versions.yml"                                   , emit: versions, topic: versions
 
     when:
     task.ext.when == null || task.ext.when
--- a/modules/nf-core/propr/propd/meta.yml
+++ b/modules/nf-core/propr/propd/meta.yml
@@ -14,7 +14,8 @@ tools:
       documentation: "https://rdrr.io/cran/propr/man/propr.html"
       tool_dev_url: "https://github.com/tpq/propr"
       doi: "10.1038/s41598-017-16520-0"
-      licence: ["GPL-2"]
+      licence:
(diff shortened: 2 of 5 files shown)


suzannejin:
Hi @caraiz2001 can you have a look?

atrigila (approved):
LGTM. I understand this change works in the pipeline as it was tested in the nf-test "Test rnaseq propd + grea profile" in this PR https://github.com/nf-core/differentialabundance/pull/727

pinin4fjords:
Thanks @atrigila ! @suzannejin @caraiz2001 we can revise later if you have changes

PR #11643, opened 2026-05-14
Add mifaser

vinisalazar:
##  Description

  Adds mifaser for functional annotation of metagenomic reads via EC number assignment (DOI: [10.1093/nar/gkx1209](http://doi.org/10.1093/nar/gkx1209)).

Work towards nf-core/funcprofiler#52, nf-core/funcprofiler#58

##  Notes on container

  The bioconda package (mifaser=1.60) and the tool's own container (ghcr.io/vdblab/mifaser:1.64d) have incompatible --databasefolder semantics:

  - 1.60 (bioconda/bromberglab): resolves --databasefolder relative to the Python package install directory (`os.path.join(pkg_dir, 'database', args.databasefolder)`), making it unusable with a user-provided database path unless an absolute path is passed.
  - 1.64d (vdblab): resolves the flag as a direct filesystem path, which is the expected behaviour.

  This module uses ghcr.io/vdblab/mifaser:1.64d to work around the 1.60 bug. This fails the container_links lint check. A proper fix would require either:
  1. A biocontainers-compatible release of mifaser ≥ 1.64, or
  2. A patch to the 1.60 bioconda package.

Currently the mifaser code is available on BitBucket: https://bitbucket.org/bromberglab/mifaser/ and version 1.64 is available, but the [bioconda recipe](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/mifaser/meta.yaml) points to the [PyPI release](https://pypi.org/project/mifaser/) which is still 1.60

@nickp60 do you think we could set the bioconda recipe to get the tarball from a bitbucket tag endpoint instead? For example, a link from https://bitbucket.org/bromberglab/mifaser/downloads/?tab=tags


nickp60:
Thanks @vinisalazar!  I emailed the authors of Mifaser to ask them to push a new version to pypi.  The Bioconda folks want stuff coming direct from Pypi so their recipe updater bots can keep things current.

In the meantime, could you maybe set the command to prepend the database with `$PWD` in the `script` section instead so it has an absolute path to work with?

And just FYI their is an official docker container https://hub.docker.com/r/bromberglab/mifaser/tags.  I built the ghcr.io/vdblab/mifaser:1.64d container because I needed version tagged versions rather than the "latest" or "dev" tags.

vinisalazar:
Thanks @nickp60, I ended up going with 1.60 and Biocontainers (instead of the official Docker image you linked) as that is a requirement from the linter. Biocontainers consumes from Bioconda so we will need an update to the recipe in order to get 1.64. I added the $PWD fix you suggested.

vinisalazar:
Also, on the PyPI requirement for Bioconda: completely reasonable, but while there isn't an automated routine for the mifaser BitBucket repo to push new tags to PyPI, I'd argue that in the meantime, using the BitBucket tags would make more sense, even if it requires manually updating the recipe (the lack of the automated routine I mentioned already implies the need for manual update step)

modules/nf-core/mifaser/main.nf
```diff
@@ -0,0 +1,50 @@
...
+    label 'process_medium'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/mifaser:1.60--pyh106432d_0' :
+        'quay.io/biocontainers/mifaser:1.60--pyh106432d_0' }"
+
+    input:
+    tuple val(meta), path(reads)
+    path db
```
CaroAMN: If you need to use the database for every incoming sample ( sample 1, sample 2, etc.), you need to use `each` : https://docs.seqera.io/nextflow/process#input-repeaters-each 
Otherwise, the process will only run for the first incoming sample and not for potentially more

CaroAMN:
```suggestion
    each path(db)
```

nickp60:
> Also, on the PyPI requirement for Bioconda: completely reasonable, but while there isn't an automated routine for the mifaser BitBucket repo to push new tags to PyPI, I'd argue that in the meantime, using the BitBucket tags would make more sense, even if it requires manually updating the recipe (the lack of the automated routine I mentioned already implies the need for manual update step)

They already pushed the version so you should be good to go with using official pypi/conda/biocontainers 1.64!

vinisalazar:
Thank you for the review @CaroAMN.

Will be applying changes shortly.

And fantastic news about the update @nickp60.

vinisalazar:
@nickp60 looks like the Singularity depot hasn't updated for 1.64 yet, will wait a couple of days to see if that changes

modules/nf-core/mifaser/main.nf
```diff
@@ -0,0 +1,50 @@
...
+    label 'process_medium'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/mifaser:1.64--pyh106432d_0' :
+        'quay.io/biocontainers/mifaser:1.64--pyh106432d_0' }"
+
+    input:
+    tuple val(meta), path(reads)
+    each path(db)
```
CaroAMN:
```suggestion
   path(db)
```

CaroAMN: hello its me again. Sorry for the confusion, I just learned that people dont use each for this purpose so i suggest changing it back. Sorry again

vinisalazar:
Hi @CaroAMN,

The singularity container we were waiting on is now online, so I believe this one is good to go.

Thanks again for your review.

Vini

CaroAMN (approved):
LGTM :)

PR #11648, opened 2026-05-15
Update scvi-tools version to 1.4.3

Zethson:
It looks like scvi-tools 1.3.3 doesn't support Blackwell GPUs because it installs old torch versions. Therefore, we upgrade to 1.4.3


nictru (approved):
LGTM

PR #11651, opened 2026-05-15
update scar scvi-tools version to 1.4.3

Zethson:
Follow up to https://github.com/nf-core/modules/pull/11648

Also updates scvi-tools for scar.
Further sets required environment variables. Otherwise, scvi-tools:py3.13-cu12-1.4.3- crashes inside `torch._dynamo` initialization when run with -u <uid>:<gid> and the UID isn't in /etc/passwd.
PyTorch 2.7's inductor cache-dir helper calls `getpass.getuser()`, which falls through to pwd.getpwuid() and raises KeyError for any UID not present in the container's passwd file. This is the default Docker/Singularity behaviour on most institutional setups (LDAP/AD-assigned UIDs).

We set the 
```python
os.environ.setdefault("TORCHINDUCTOR_CACHE_DIR", os.path.join(os.getcwd(), "torch_cache"))
```

to fix this.

### DIFF
--- a/modules/nf-core/scvitools/scar/environment.yml
+++ b/modules/nf-core/scvitools/scar/environment.yml
@@ -4,4 +4,4 @@ channels:
   - conda-forge
   - bioconda
 dependencies:
-  - conda-forge::scvi-tools=1.3.3
+  - conda-forge::scvi-tools=1.4.3
--- a/modules/nf-core/scvitools/scar/main.nf
+++ b/modules/nf-core/scvitools/scar/main.nf
@@ -3,9 +3,10 @@ process SCVITOOLS_SCAR {
     label 'process_medium'
 
     conda "${moduleDir}/environment.yml"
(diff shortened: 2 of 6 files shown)


nictru:
I think it would be more elegant to set [`TORCHINDUCTOR_CACHE_DIR`](https://docs.pytorch.org/tutorials/recipes/torch_compile_caching_configuration_tutorial.html#torchinductor-cache-dir) so that we do not have to alter `USER` and `HOME` which could affect other things that we do not think about currently

Zethson:
> I think it would be more elegant to set [TORCHINDUCTOR_CACHE_DIR](https://docs.pytorch.org/tutorials/recipes/torch_compile_caching_configuration_tutorial.html#torchinductor-cache-dir) so that we do not have to alter USER and HOME which could affect other things that we do not think about currently

Right, I was also worried about that but in the end thought "whatever". I'll change it.

nictru (approved):

PR #11684, opened 2026-05-18
feat(ribocode,ribotish): pyfasta indexes, prefix-scoped outputs, optional ribotish -a

pinin4fjords:
Bundles three in-place module changes carried in nf-core/riboseq#174. Each is self-contained and addresses a different pain point we hit running RiboCode / Ribo-TISH at scale.

### ribocode/prepare

Pre-build the pyfasta `.gdx`/`.flat` indexes for `annotation/transcripts_sequence.fa` immediately after `prepare_transcripts`, using the same `key_fn` RiboCode applies internally (split on first space, otherwise split on `|`). The stub also touches the two new sidecars.

Why: downstream RiboCode steps open the FASTA with pyfasta, which lazily writes `.gdx`/`.flat` next to the input on first read. Under Fusion staging those writes land back at the upstream task's S3 prefix and silently corrupt the staged copy on retries. Building the indexes inside the producing task fixes it.

### ribocode/ribocode

Switch the `orf_txt` and `orf_txt_collapsed` output globs from `*.txt` / `*_collapsed.txt` to `${prefix}.txt` / `${prefix}_collapsed.txt` so multi-instance publication is unambiguous (`*.txt` previously matched both the all-ORFs and collapsed files into the same emit). The `prefix` binding is promoted out of `def` in both `script:` and `stub:` so it resolves at the output-glob stage; the Nextflow 26 strict parser rejects re-declaring the same local with `def` across the two blocks.

The existing stub assertion that indexed `process.out.orf_txt[0][1][0]` is corrected to the new single-file shape (`process.out.orf_txt[0][1]`).

### ribotish/predict

**Breaking signature change.** The third input tuple gains an optional fourth element, `reference_gtf`, plumbed through to `ribotish predict` as `-a <gtf>` when populated:

```groovy
tuple val(meta3), path(fasta), path(gtf), path(reference_gtf, stageAs: 'secondary.gtf')
```

Callers must supply a fourth element on every emit. Pass `[]` for the no-op case (no secondary annotation). The existing test cases in this PR are migrated that way; positive-coverage tests for the populated path will land in a follow-up.

Why: Ribo-TISH's `-a` argument is the documented hook for layering a secondary annotation (e.g. MANE/RefSeq) on top of the primary GTF, and we want to expose it from the module without a second optional input tuple.

### Test plan

All three modules pass under Docker on a `c5.9xlarge` VM with `nf-core 4.0.2` / `nextflow 26.04.1` / `nf-test 0.9.5`:

```
nf-core modules test --profile docker ribocode/prepare
nf-core modules test --profile docker ribocode/ribocode
nf-core modules test --profile docker ribotish/predict
```

Snapshot deltas:

- `ribocode/prepare`: non-stub snapshot gains the two new file md5s (`transcripts_sequence.fa.flat`, `transcripts_sequence.fa.gdx`); existing files' md5s unchanged.
- `ribocode/ribocode`: `orf_outputs` snapshot drops the duplicate `test_collapsed.txt` entry that the old `*.txt` glob had pulled into `orf_txt`; everything else unchanged.
- `ribotish/predict`: no snapshot change (the `[]` migration is a no-op at runtime).

Source: nf-core/riboseq#174

### DIFF
--- a/modules/nf-core/ribocode/prepare/main.nf
+++ b/modules/nf-core/ribocode/prepare/main.nf
@@ -27,6 +27,18 @@ process RIBOCODE_PREPARE {
         -f ${fasta} \\
         -o annotation \\
         $args
+
+    # Pre-build pyfasta .gdx/.flat with RiboCode's key_fn so consumers don't write to staged inputs.
+    python - <<'PYTHON'
+from pyfasta import Fasta
+def key_fn(name):
+    if ' ' in name:
+        return name.split()[0]
+    if '|' in name:
+        return name.split('|')
+    return name
+Fasta('annotation/transcripts_sequence.fa', key_fn=key_fn)
+PYTHON
     """
 
     stub:
@@ -36,6 +48,8 @@ process RIBOCODE_PREPARE {
 
     touch annotation/transcripts_cds.txt
     touch annotation/transcripts_sequence.fa
+    touch annotation/transcripts_sequence.fa.gdx
+    touch annotation/transcripts_sequence.fa.flat
     touch annotation/transcripts.pickle
     """
 }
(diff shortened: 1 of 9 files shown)


pinin4fjords:
Superseded by the per-module splits:

- ribocode (prepare + ribocode): #11685
- ribotish/predict: #11686

Closing this bundled draft. Branch `ribocode-ribotish-bundled-fixes` will be pruned once the splits land.

PR #11685, opened 2026-05-18
feat(ribocode): pre-build pyfasta indexes + prefix-scoped outputs

pinin4fjords:
Two related changes carried in nf-core/riboseq#174.

### ribocode/prepare

After `prepare_transcripts` runs, pre-build the pyfasta `.gdx`/`.flat` sidecars for `annotation/transcripts_sequence.fa` by instantiating `RiboCode.prepare_transcripts.GenomeSeq` directly.

Why: RiboCode's downstream `detectORF.py` opens `transcripts_sequence.fa` via pyfasta, which lazily writes `.gdx`/`.flat` next to the FASTA on first read. Under Nextflow's default symlink staging those writes go through the symlink and land in the upstream `RIBOCODE_PREPARE` task's work dir; parallel consumers then race on the same sidecar paths, with last-writer-wins behaviour on shared/network storage. Building the sidecars in the producing task makes them part of the published annotation directory and removes the lazy-write path entirely - same pattern as `samtools/faidx` shipping `.fai` alongside its FASTA.

### ribocode/ribocode

Switch `orf_txt` and `orf_txt_collapsed` from `*.txt` / `*_collapsed.txt` to `${prefix}.txt` / `${prefix}_collapsed.txt`. The previous globs matched both files into the same emit. The `prefix` binding is promoted out of `def` in both `script:` and `stub:` so it resolves at the output-glob stage; the Nextflow 26 strict parser rejects re-declaring the same local across the two blocks. Existing stub assertion adjusted from `process.out.orf_txt[0][1][0]` to `process.out.orf_txt[0][1]`.

### Snapshot deltas

- `ribocode/prepare`: gains two new file md5s (`.flat`, `.gdx`); existing md5s unchanged.
- `ribocode/ribocode`: drops the duplicate `test_collapsed.txt` entry the old `*.txt` glob double-counted.

Source: nf-core/riboseq#174. Supersedes the ribocode half of nf-core/modules#11684.

### DIFF
--- a/modules/nf-core/ribocode/prepare/main.nf
+++ b/modules/nf-core/ribocode/prepare/main.nf
@@ -27,6 +27,18 @@ process RIBOCODE_PREPARE {
         -f ${fasta} \\
         -o annotation \\
         $args
+
+    # Pre-build pyfasta .gdx/.flat with RiboCode's key_fn so consumers don't write to staged inputs.
+    python - <<'PYTHON'
+from pyfasta import Fasta
+def key_fn(name):
+    if ' ' in name:
+        return name.split()[0]
+    if '|' in name:
+        return name.split('|')
+    return name
+Fasta('annotation/transcripts_sequence.fa', key_fn=key_fn)
+PYTHON
     """
 
     stub:
@@ -36,6 +48,8 @@ process RIBOCODE_PREPARE {
 
     touch annotation/transcripts_cds.txt
     touch annotation/transcripts_sequence.fa
+    touch annotation/transcripts_sequence.fa.gdx
+    touch annotation/transcripts_sequence.fa.flat
     touch annotation/transcripts.pickle
     """
 }
--- a/modules/nf-core/ribocode/prepare/main.nf
+++ b/modules/nf-core/ribocode/prepare/main.nf
@@ -28,7 +28,9 @@ process RIBOCODE_PREPARE {
         -o annotation \\
         $args
 
-    # Pre-build pyfasta .gdx/.flat with RiboCode's key_fn so consumers don't write to staged inputs.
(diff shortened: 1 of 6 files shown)


modules/nf-core/ribocode/prepare/main.nf
```diff
@@ -27,6 +27,11 @@ process RIBOCODE_PREPARE {
         -f ${fasta} \\
         -o annotation \\
         $args
+
+    # Pre-build the pyfasta .gdx/.flat sidecars by instantiating RiboCode's own GenomeSeq -
```
jonasscheid, line 31: Could this be an additional flag in `prepare_transcripts` in the future? Maybe worth to file an issue to their github repo

jonasscheid (approved):

modules/nf-core/ribocode/prepare/main.nf
pinin4fjords: Thanks! It's kind of something that comes out of the workflow use case specifically, but I can at least flag it with them.

pinin4fjords:
@jonasscheid good shout - opened [xryanglab/RiboCode#70](https://github.com/xryanglab/RiboCode/issues/70) upstream proposing the eager-build patch. If they take it we can drop the inline pre-build entirely. I'll add a link in the module's inline comment so future maintainers can find it.

pinin4fjords:
Update: opened the upstream PR too - [xryanglab/RiboCode#71](https://github.com/xryanglab/RiboCode/pull/71). If they accept it, the next ribocode container build picks up the eager index and we can drop the inline pre-build from this module entirely. For now this PR remains the workaround.

jonasscheid:
Great stuff 👍🏼

PR #11686, opened 2026-05-18
feat(ribotish/predict): optional reference_gtf input + topics versions + 0.2.8

pinin4fjords:
Adds an optional secondary GTF input to `ribotish predict` (so callers can supply a MANE/RefSeq-style overlay via `-a` while keeping the primary GTF on `-g`), and brings the module up to current nf-core conventions.

### Changes

- **New optional input**: `tuple val(meta7), path(reference_gtf, stageAs: 'secondary.gtf')`. When populated, plumbed through as `-a <reference_gtf>` on the ribotish command line. Lives in its own tuple - matches the pattern this module already uses for `bam_ti`, `candidate_orfs`, `para_ribo`, `para_ti`. Existing `(meta3, fasta, gtf)` signature is preserved, so consumers add a single new input channel rather than restructuring an existing tuple.
- **Topic-based versions**: legacy `path "versions.yml"` emit + `cat <<-END_VERSIONS` heredoc replaced with `tuple val("${task.process}"), val('ribotish'), eval("ribotish --version | sed 's/ribotish //'"), topic: versions, emit: versions_ribotish`. `meta.yml` regenerated via `nf-core modules lint --fix` (adds the `topics:` block and reshapes the `versions_ribotish` output entry).
- **Bump ribotish 0.2.7 → 0.2.8** (bioconda; same build hash, container URLs and `environment.yml` updated).

### Test plan

`nf-core modules test --profile docker ribotish/predict` runs all four cases green (single/multi ribo bam × main/stub). Snapshot regenerated:

- `versions_*` keys renamed to `versions_ribotish_*` (matching the new emit name).
- Version string `0.2.7` → `0.2.8`.
- Prediction-table assertions unchanged - 0.2.8 is a patch release.

### Caller migration

Existing callers must add a 7th input slot. Pass `Channel.of([[], []])` (or `[[id:'reference'], []]`) for the no-op case; pass a populated `[meta, gtf_path]` to use `-a`.

Source: nf-core/riboseq#174. Supersedes the ribotish half of nf-core/modules#11684.

### DIFF
--- a/modules/nf-core/ribotish/predict/environment.yml
+++ b/modules/nf-core/ribotish/predict/environment.yml
@@ -4,4 +4,4 @@ channels:
   - conda-forge
   - bioconda
 dependencies:
-  - bioconda::ribotish=0.2.7
+  - bioconda::ribotish=0.2.8
--- a/modules/nf-core/ribotish/predict/main.nf
+++ b/modules/nf-core/ribotish/predict/main.nf
@@ -4,22 +4,23 @@ process RIBOTISH_PREDICT {
 
     conda "${moduleDir}/environment.yml"
     container "${ workflow.containerEngine in ['singularity', 'apptainer'] && !task.ext.singularity_pull_docker_container ?
-        'https://depot.galaxyproject.org/singularity/ribotish:0.2.7--pyhdfd78af_0':
-        'quay.io/biocontainers/ribotish:0.2.7--pyhdfd78af_0' }"
+        'https://depot.galaxyproject.org/singularity/ribotish:0.2.8--pyhdfd78af_0':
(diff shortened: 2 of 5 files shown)


SPPearce:
Dear Claude, can you update the nf-test structure while you are at it, so that each test only has one snapshot (which doesn't need a name)

pinin4fjords:
@SPPearce thanks - done in 5516eb154. Each test now has a single anonymous `snapshot()`; non-stub tests keep their `contains()` row checks separately (they pin specific known-good rows independently of the snapshot). Versions use the canonical `process.out.findAll { key, val -> key.startsWith('versions') }` pattern.

SPPearce (approved):

PR #11695, opened 2026-05-19
feat(rpbp): add 8 rpbp modules + fasta_gtf_bam_rpbp subworkflow

pinin4fjords:
Adds 8 `rpbp/*` modules and the `fasta_gtf_bam_rpbp` subworkflow that wrap the Rp-Bp ribosome-profiling ORF predictor (Malone et al. 2017, [doi:10.1093/nar/gkw1350](https://doi.org/10.1093/nar/gkw1350)).

### What Rp-Bp does

Ribosome profiling (Ribo-seq) sequences the short mRNA fragments protected by translating ribosomes. Rp-Bp uses those reads to score, for every candidate open reading frame (ORF) in the annotation, whether it shows the 3-nucleotide periodicity that marks active translation. The output is a per-sample BED of predicted translated ORFs plus matching DNA / protein FASTAs.

### Why not call the umbrella CLI

Rp-Bp ships two top-level scripts (`prepare-rpbp-genome`, `predict-translated-orfs`) that chain the internal steps together. Both also do work the modules don't need:

- `prepare-rpbp-genome` builds a bowtie2 rRNA-filter index and a STAR alignment index alongside the BED prep. These modules consume pre-aligned BAMs supplied by the caller, so those indexes are unnecessary and would force `STAR` and `bowtie2` into the module container.
- `predict-translated-orfs` internally re-runs flexbar + bowtie + STAR on raw FASTQs, duplicating any upstream alignment the caller already produced, before getting to the actual ORF prediction logic.

Splitting the underlying steps into 8 single-purpose modules:

- skips the alignment/index work,
- lets the modules ship a smaller Rp-Bp-only container,
- gives each step independent `-resume` caching.

### Modules

Reference prep (run once per genome):

- `rpbp/preparegenome` - enumerates candidate ORFs from the GTF and writes the BEDs the per-sample steps consume.

Per-sample chain:

- `rpbp/extractmetageneprofiles` - per-read-length read coverage around annotated start codons (a "metagene profile").
- `rpbp/estimatemetagenebayesfactors` - Bayes-factor score per read length for "this length shows 3-nt periodicity vs not".
- `rpbp/selectperiodicoffsets` - chooses one P-site offset per read length (the canonical ribosome-position adjustment).
- `rpbp/getperiodiclengthsoffsets` - threshold-filters down to the read lengths that will drive ORF scoring. Thresholds are configurable via `ext.args` (4 space-separated tokens: min count, min BF mean, max BF var, min BF likelihood; defaults mirror `rpbp.defaults.metagene_options`).
- `rpbp/extractorfprofiles` - builds per-ORF P-site count vectors using the surviving read lengths and offsets.
- `rpbp/estimateorfbayesfactors` - Bayes-factor score per ORF for "translated vs untranslated", via Stan models bundled in the Rp-Bp Python package.
- `rpbp/selectfinalpredictionset` - applies score / length / overlap rules and emits the final predicted-ORF BED + DNA / protein FASTAs.

### Subworkflow

`fasta_gtf_bam_rpbp` wires the modules together: `rpbp/preparegenome` once per genome, the 7-step per-sample chain for each input BAM. Designed to be consumed by any caller that has pre-aligned Ribo-seq BAMs plus the matching genome FASTA / annotation GTF.

### Container

All `rpbp/*` modules share a Wave-built container built directly from each module's `environment.yml` (`bioconda::rpbp=4.0.1`):

- docker: `community.wave.seqera.io/library/rpbp:4.0.1--71297b462026e13b`
- singularity: `https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/14/146c3f15abf184a5ec13531d2a040ba7b9235c1091723aa37c7a119817411367/data`

Versions emitted via `topic: versions`.

### Test plan

Real + stub tests pass for each module and the subworkflow under `nf-core modules test --profile docker` / `nf-core subworkflows test --profile docker`.

Module-level tests consume per-stage static fixtures from `nf-core/test-datasets:modules` under `data/genomics/homo_sapiens/riboseq_expression/rpbp/` (added in [nf-core/test-datasets#2064](https://github.com/nf-core/test-datasets/pull/2064)) - one immediate-upstream artefact per module rather than chaining six upstream stages in setup. Each module test runs in well under a minute (the per-ORF Bayes-factor fit is the only multi-minute step). The full chain still runs end-to-end in the subworkflow test.


SPPearce:
Not reviewing this, make 8 individual module PRs.

PR #3656, opened 2023-07-20
Hisat2 align splicesitesoptional

JackCurragh:
Updated Hisat2 align to allow splice sites to actually be optional 
Changes involve the specification of the splice sites paramater being defined outside the command so that if it is empty (' ') then it will not throw an error. 

Closes #3653 
## PR checklist

- [x] This comment contains a description of changes (with reason).
- [x] If you've fixed a bug or added code that should be tested, add tests!

PR #3179, opened 2023-03-28
Collectinsertsizemetrics

FerriolCalvet:
## PR checklist

Closes #XXX 

- [ ] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [ ] Remove all TODO statements.
- [ ] Emit the `versions.yml` file.
- [ ] Follow the naming conventions.
- [ ] Follow the parameters requirements.
- [ ] Follow the input/output options guidelines.
- [ ] Add a resource `label`
- [ ] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - [ ] `PROFILE=docker pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [ ] `PROFILE=singularity pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`
  - [ ] `PROFILE=conda pytest --tag <MODULE> --symlink --keep-workflow-wd --git-aware`

PR #11696, opened 2026-05-19
feat(simpleaf): bump to 0.25.0 with Seqera Wave container and topic-channel versions

an-altosian:
## Summary

- Updates **simpleaf/index** and **simpleaf/quant** modules to simpleaf 0.24.1 using a Seqera Wave / Community container.
- Migrates the versions output from the legacy `versions.yml` file to the new **topic-channel** convention (matches `t1k/build`).
- Realigns `environment.yml` pins with what the container actually ships, including removing the no-longer-required salmon dependency.

## Container

| Type | URI |
| --- | --- |
| Docker | `community.wave.seqera.io/library/simpleaf:0.24.1--41e05fd46d838d92` |
| Singularity / Apptainer | `https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/83/83ab36a4aba7f26860f96dd73bb428a4a4eb05450c4cfc9488cf77adefad1adb/data` |

### Versions / dependencies

- simpleaf `0.19.5` → **`0.24.1`**
- alevin-fry `0.11.2` → **`0.14.0`** (matches container)
- piscem `0.12.2` → **`0.20.0`** (matches container)
- salmon **removed** — simpleaf 0.24.x's piscem-only pipeline does not depend on salmon, and the Wave container does not ship it. The previous emit silently produced an empty version string because `$(salmon --version)` errored silently inside the `cat <<-END_VERSIONS` heredoc.

### Topic-channel versions

Both `main.nf` files now emit each tool's version as a separate topic-channel tuple instead of writing `versions.yml`:

\`\`\`groovy
tuple val(\"\${task.process}\"), val('alevin-fry'), eval(\"alevin-fry --version | sed 's/alevin-fry //'\"), topic: versions, emit: versions_alevin_fry
tuple val(\"\${task.process}\"), val('piscem'),     eval(\"piscem --version | sed 's/piscem //'\"),         topic: versions, emit: versions_piscem
tuple val(\"\${task.process}\"), val('simpleaf'),   eval(\"ALEVIN_FRY_HOME=. simpleaf --version | sed 's/simpleaf //'\"), topic: versions, emit: versions_simpleaf
\`\`\`

The `ALEVIN_FRY_HOME=.` prefix is needed because the `eval()` directive runs in a separate shell context from the script block where the env var is exported.

### Test updates

- piscem 0.20 produces a new on-disk index format. The legacy `.sshash` / `.json` files were replaced by `.ssi` / `.ssi.mphf` / `_ver.json` / `.sigs.json` / `.tct` / `.tdct`. Assertions in `main.nf.test` and the corresponding stub `touch` lines have been updated.
- Snapshot assertions migrated from `snapshot(process.out.versions).match()` to `snapshot(process.out.findAll { key, val -> key.startsWith(\"versions\") }).match()` to capture the new per-tool topic emits.
- `meta.yml` regenerated via `nf-core modules lint --fix`; obsolete `versions: - versions.yml` stanza removed by hand.

## Test plan

- [x] `nf-test test modules/nf-core/simpleaf/index/tests/main.nf.test --profile docker` — 5/5 PASS
- [x] `nf-test test modules/nf-core/simpleaf/quant/tests/main.nf.test --profile docker` — 2/2 PASS
- [x] `nf-core modules lint simpleaf/index` — **0 failures**
- [x] `nf-core modules lint simpleaf/quant` — **0 failures**

## Note on residual lint warnings

A separate PR comment below explains the residual `main_nf_container` and `container_links` warnings, which are upstream nf-core/tools quirks that surface on every Seqera Wave module (e.g. `t1k/build`, `ngsbits/*`) and are not actionable at the module level.


an-altosian:
### Note for reviewers: residual `nf-core modules lint` warnings

This PR adopts a Seqera Wave / Community container per the [nf-core Seqera Containers docs](https://nf-co.re/docs/developing/containers/seqera-containers). Two lint warnings remain after this PR and are **not actionable at the module level** — they're known artifacts of nf-core/tools predating the Wave URL formats. Including them here so reviewers don't have to re-investigate.

#### 1. `main_nf_container: Container versions do not match`

**Cause:** The lint check at [`nf_core/modules/lint/main_nf.py:360-384`](https://github.com/nf-core/tools/blob/main/nf_core/modules/lint/main_nf.py) parses the Singularity tag with the regex `(?:[:.])?([A-Za-z\d\-_.]+?)(?:\.img)?(?:\.sif)?$`. For our Wave Singularity URL `https://community-cr-prod.seqera.io/.../sha256/<hash>/data`, that regex captures the literal string `\"data\"`. The Docker tag regex captures `\"0.24.1--41e05fd46d838d92\"`. Comparison `\"data\" == \"0.24.1--41e05fd46d838d92\"` → False → warning.

The check was designed for biocontainers-style URLs (`…/fastqc:0.11.9--0`) where both URLs end with the same `version--build` suffix. Wave's content-addressable blob URLs have no version substring at all.

#### 2. `container_links` (HTTP 404)

**Cause:** Lint does a naked HTTP HEAD against the Docker URL. Wave's community registry doesn't answer with 200 on that endpoint — it requires an OCI manifest GET with `Accept: application/vnd.oci.image.manifest.v1+json` plus a token negotiation. `docker pull` / `singularity pull` / `nextflow run` all handle that handshake correctly. nf-test in this PR successfully pulled and ran the container end-to-end.

#### Reproducibility

Both warnings reproduce on every existing Wave-container module in the repo:
- [`modules/nf-core/t1k/build`](https://github.com/nf-core/modules/blob/master/modules/nf-core/t1k/build/main.nf) (the gold-standard topic-channel example)
- [`modules/nf-core/ngsbits/bedannotategc`](https://github.com/nf-core/modules/blob/master/modules/nf-core/ngsbits/bedannotategc/main.nf), and other ngsbits modules
- … and approximately 70 other modules already on Wave

#### Related upstream issues

- [nf-core/tools#3179 — Download: Fix support for Seqera Community Containers](https://github.com/nf-core/tools/issues/3179) (CLOSED, but documents that Wave Docker and Singularity URLs legitimately have different SHA values)
- [nf-core/tools#3656 — Enable configurations for module linting](https://github.com/nf-core/tools/issues/3656) (OPEN — proposes granular per-check disabling, which would let modules opt out of these specific Wave-related warnings)
- [nf-core/tools#1377 — Modules linting 'Container versions do not match'](https://github.com/nf-core/tools/issues/1377) (CLOSED 2021, addressed an older biocontainers-format mismatch; the fix didn't anticipate Wave URLs)

#### What this PR does NOT include

I did not file a new upstream issue specifically for the Wave version-mismatch case — happy to do so if reviewers want one to track. The structural cause is documented in #3179 and the workaround would naturally land in #3656.

mashehu (approved):

PR #11717, opened 2026-05-20
feat(custom/orfcountmatrix): add ORF x sample P-site count matrix builder

pinin4fjords:
## Summary

Adds `custom/orfcountmatrix`, a stdlib-only Python module that pivots per-sample ORF P-site count TSVs into a single ORF x sample matrix.

- Input shape: `tuple val(meta), path(per_sample_counts, stageAs: 'counts/*'), path(orf_catalogue_bed12)`. Per-sample TSVs are `sample_id<TAB>orf_id<TAB>count` (every row of a given file shares the same sample_id). The BED12 catalogue's 4th column defines the canonical ORF id list.
- Output: a single TSV with header `orf_id<TAB>sample1<TAB>sample2...`, one row per catalogue ORF in catalogue order, columns in lexicographic sample-id order.
- Zero-fill semantics: ORFs absent from a sample get `0` in that column; ORFs in the catalogue but absent from every sample appear as a row of zeros. This guarantees the matrix is keyed on the catalogue rather than just the union of observed ORFs.

Under `custom/` rather than as a tool-specific module because the pivot is a generic transform over the `(sample_id, orf_id, count)` input shape produced upstream by P-site counters such as Plastid, Ribotricer, or riboWaltz.

Source: extracted from `modules/local/orf/count_matrix/` in nf-core/riboseq#174.

## Paired test-datasets PR

nf-core/test-datasets#2065 adds the hand-crafted synthetic fixtures used by the test. The test currently fetches them from `raw.githubusercontent.com/pinin4fjords/test-datasets/add-orfcountmatrix-test-data/...` - **the URLs will be flipped to `nf-core/test-datasets/modules` once #2065 merges.**

## Test plan

- [x] Real test (non-stub) runs and produces the expected zero-filled, catalogue-ordered matrix.
- [x] Stub test passes.
- [x] Snapshot stable across two consecutive `nf-core modules test` invocations.
- [x] `nf-core modules lint`: 51 passed, 0 warnings, 0 failures.

### DIFF
--- a/modules/nf-core/custom/orfcountmatrix/environment.yml
+++ b/modules/nf-core/custom/orfcountmatrix/environment.yml
@@ -1,0 +1,7 @@
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/environment-schema.json
+channels:
+  - conda-forge
+  - bioconda
+dependencies:
+  - python=3.11
--- a/modules/nf-core/custom/orfcountmatrix/environment.yml
+++ b/modules/nf-core/custom/orfcountmatrix/environment.yml
@@ -4,4 +4,6 @@ channels:
   - conda-forge
   - bioconda
 dependencies:
-  - python=3.11
+  - pandas=2.2.3
+  - python=3.11.15
+  - pyyaml=6.0.2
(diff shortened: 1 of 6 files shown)


pinin4fjords:
Closing — on reflection the use case is too pipeline-specific for nf-core/modules (it's a thin pandas pivot of riboseq-flavoured ORF count TSVs against a BED12 catalogue from this pipeline's own ORF catalogue stage). The cleaned-up version (pandas pivot + template + Wave container) has been carried over to the local module in [nf-core/riboseq#174](https://github.com/nf-core/riboseq/pull/174) at commit `f63267c`. Paired test-datasets PR nf-core/test-datasets#2065 will be closed too.

PR #11724, opened 2026-05-20
feat(simpleaf): add simpleaf/multiplexquant module

an-altosian:
## Summary

Adds a new `simpleaf/multiplexquant` module wrapping [`simpleaf multiplex-quant`](https://simpleaf.readthedocs.io/en/latest/flex-quant-command.html) for sample-multiplexed single-cell libraries (10x Chromium Fixed RNA Profiling / Flex). End-to-end: auto-resolves probe set + sample-BC TSV from a chemistry preset, optionally builds a piscem probe index, maps reads with `piscem map-sc`, performs hierarchical cell-BC + sample-BC correction, collates, and quantifies — emitting barcodes prefixed by demultiplexed sample name.

- Module follows current nf-core conventions: Seqera Wave container, topic-channel versions output (`alevin-fry`, `piscem`, `simpleaf`), `process_medium` label, stub block, meta-merge across input channels.
- Four input channels: `(meta, chemistry, reads)`, `(meta2, index, t2g_map)`, `(meta3, probe_set, sample_bc_list, cell_bc_list)`, `resolution`. The middle two channels are fully optional (`[]` for any element) — empty inputs fall back to chemistry-preset auto-resolution.
- `simpleaf 0.25.0` / `piscem 0.20.0` / `alevin-fry 0.15.0` via `community.wave.seqera.io/library/simpleaf:0.25.0--b9f96d8b71a01864`.

## Validation

- `nf-test`: `flex - auto` (end-to-end via Wave container) and `flex - auto - stub` pass locally on Docker.
- Real-data validation: ran against an internal 10x GEM-X Flex v2 mouse library (466M reads → 98.10% mapping → 36,859 cells across the expected sample-BC wells), matching cellranger output.

## Test plan

- [ ] `nf-test test modules/nf-core/simpleaf/multiplexquant/tests/main.nf.test --profile docker` passes
- [ ] `nf-core modules lint simpleaf/multiplexquant` passes (expected Wave-container lint warnings as in #11696)
- [ ] CI green


modules/nf-core/simpleaf/multiplexquant/main.nf
```diff
@@ -0,0 +1,101 @@
...
+
+    output:
+    tuple val(meta), path("${prefix}/af_map")                       , emit: map
+    tuple val(meta), path("${prefix}/af_quant")                     , emit: quant
+    tuple val(meta), path("${prefix}/af_quant/alevin/quants.h5ad")  , emit: h5ad,        optional: true
+    tuple val(meta), path("${prefix}/probe_t2g.tsv")                , emit: t2g,         optional: true
+    tuple val(meta), path("${prefix}/probe_index/index")            , emit: probe_index, optional: true
+    tuple val("${task.process}"), val('alevin-fry'), eval("alevin-fry --version | sed 's/alevin-fry //'"),                    topic: versions, emit: versions_alevin_fry
+    tuple val("${task.process}"), val('piscem'),     eval("piscem --version | sed 's/piscem //'"),                            topic: versions, emit: versions_piscem
+    tuple val("${task.process}"), val('simpleaf'),   eval("ALEVIN_FRY_HOME=. simpleaf --version | sed 's/simpleaf //'"),      topic: versions, emit: versions_simpleaf
```
SPPearce: I think that the `ALEVIN_FRY_HOME` environment variable should be inherited by this part as well. 
```suggestion
    tuple val("${task.process}"), val('simpleaf'),   eval("simpleaf --version | sed 's/simpleaf //'"),      topic: versions, emit: versions_simpleaf
```

modules/nf-core/simpleaf/multiplexquant/main.nf
```diff
@@ -0,0 +1,101 @@
...
+// `--organism`. Any combination of overrides is allowed.
+def referenceArgs(index, probe_set, sample_bc_list, cell_bc_list, t2g_map) {
+    def parts = []
+    if (index)          parts << "--index ${index}"
+    if (probe_set)      parts << "--probe-set ${probe_set}"
+    if (sample_bc_list) parts << "--sample-bc-list ${sample_bc_list}"
+    if (cell_bc_list)   parts << "--cell-bc-list ${cell_bc_list}"
+    if (t2g_map)        parts << "--t2g-map ${t2g_map}"
+    return parts.join(' \\\n        ')
+}
```
SPPearce: This is very non-standard, compared to just doing them one at a time.

modules/nf-core/simpleaf/multiplexquant/main.nf
```diff
@@ -0,0 +1,101 @@
...
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "${meta.id}"
+
+    def mapping_args   = mappingArgs(chemistry, reads)
+    def reference_args = referenceArgs(index, probe_set, sample_bc_list, cell_bc_list, t2g_map)
+
+    meta = meta2 + meta3 + meta
```
SPPearce: Why are you doing this? We just take the first meta. I think this is pipeline-specific.

SPPearce (changes requested):

modules/nf-core/simpleaf/multiplexquant/main.nf
an-altosian: Thank you very much for the comment. My thought is that quantification depends on all these inputs, so merging the meta from all channels provides a comprehensive meta for downstream tasks. The first meta is added last, meaning on key collision the first meta's fields are kept. If this breaks the guidance, I can just use the first meta as the output.

modules/nf-core/simpleaf/multiplexquant/main.nf
an-altosian: Thanks for pointing this out. I was just trying to implement it in a clearer way so that parameters are grouped by their functions.

PR #11728, opened 2026-05-21
Swap vclust to topics

SPPearce:


subworkflows/nf-core/fasta_vclust_prefilter_align_cluster/tests/main.nf.test
```diff
@@ -32,12 +32,9 @@ nextflow_workflow {
         }
 
         then {
+            assert workflow.success
```
atrigila, line 35: I am curious as to why this is now outside the assertAll

atrigila (approved):
Looks good, just a minor question

subworkflows/nf-core/fasta_vclust_prefilter_align_cluster/tests/main.nf.test
SPPearce: It means that if the workflow fails, the nf-test output is just that it failed, without also reporting that every individual md5sum is different to expected.

SPPearce: It is now that way in the module and subworkflow templates: https://github.com/nf-core/tools/blob/main/nf_core/subworkflow-template/tests/main.nf.test.j2#L37

atrigila: Makes sense, thanks!

PR #11729, opened 2026-05-21
feat(subworkflows): gtf_hybridmerge_gffcompare

pinin4fjords:
## Summary

Adds a new subworkflow `gtf_hybridmerge_gffcompare` that builds a hybrid GTF from a novel-transcript GTF and a reference annotation:

1. `gffcompare` classifies every novel transcript against the reference (assigns class codes).
2. `gawk` keeps only transcripts whose class code is in a user-supplied set (e.g. `"u"` for novel intergenic, `"u,j,i"` for a wider net).
3. `bedtools intersect -v` (optional) subtracts survivors that overlap a user-supplied blacklist BED, e.g. rRNA or repeat regions. Skipped by passing `channel.empty()`.
4. `gawk` concatenates the survivors with an annotation backbone GTF, stripping comment lines and synthesising any `gene` row whose `gene_id` is referenced by a novel transcript but absent from the backbone. Synth gene rows span the union of their child transcripts' coordinates so the gene span is scientifically correct, not just the first transcript's footprint.

The output is a self-consistent hybrid GTF suitable as a `--sjdbGTFfile` for a second STAR pass, as input to gffread for transcriptome FASTA generation, or as the annotation any other downstream tool consumes.

The subworkflow takes a separate `ch_reference_gtf` (used by gffcompare for classification) and `ch_backbone_gtf` (the annotation skeleton the survivors are grafted onto). For most callers these are the same file - pass the same channel twice. Callers that want classification done against a fuller annotation but the novel transcripts grafted onto a reduced backbone (e.g. canonical-only) pass two different channels.

The emitted `hybrid_gtf` meta is the backbone's meta, not the novel input's. The hybrid GTF is logically the backbone annotation with novel transcripts grafted on, so its identity (and any downstream `.join()` that wants to align it with star-index, transcriptome-fasta or other backbone-keyed channels) should follow the backbone.

## Name

`gtf_hybridmerge_gffcompare` follows the spec's "name by purpose" branch (the subworkflow has conditional logic - the optional blacklist intersect) plus the dominant tool. The compound verb "hybridmerge" matches the `subsampledepth`/`markduplicates` shape used by existing subworkflows; "build" and "annotation" were dropped from an earlier draft as redundant (every subworkflow produces output; a GTF is an annotation). Happy to revisit if the reviewer prefers a different form.

## Components

- `gffcompare` (existing)
- `gawk` (existing, aliased as `GAWK_FILTER` and `GAWK_CONCAT`)
- `bedtools/intersect` (existing)

No new modules are introduced. The two `gawk` programs (class-code filter and backbone-plus-novel concat) are buffer-then-emit so the order of transcript / exon rows in their inputs doesn't matter; the awk source lives in `awk/{filter,concat}.awk` next to the subworkflow `main.nf` and is passed to GAWK via the `program_file` input (`${moduleDir}/awk/...`, so the subworkflow stays location-independent). Consumers `includeConfig` the bundled `nextflow.config` or pass it with `-c` for the prefix / suffix / closure-form args.

## Caller-policy bits

The bundled config sets `ext.args = '-v'` on `BEDTOOLS_INTERSECT` so the channel name `ch_blacklist_bed` actually subtracts. Consumers that want strand-aware subtraction (`-v -s`) or positive overlap can override in their own `modules.config`.

The class-codes value flows through the subworkflow's `val_class_codes` take onto `meta.class_codes` and is read at task launch via a closure-form `ext.args` on `GAWK_FILTER`. The key is dropped from meta before the hybrid_gtf is emitted (carried by the backbone-meta swap at the concat step) so the value cannot leak downstream and break consumer joins.

## Scope

Input format: standard GTF2 with `gene_id` / `transcript_id` attributes and `gene` / `transcript` feature types. GFF3 dialects and prokaryotic GTFs without transcript rows are out of scope. Surveyed nf-core/rnaseq's issue queue around attribute / feature-type edge cases: the only recurring concern is prokaryotic annotations, which don't apply to hybrid-transcriptome building. Parameterising attribute / feature-type names would add surface area for no real-world benefit in the consumers that need this subworkflow.

Resource sizing: GAWK steps run under `process_single` and hold the full input in awk arrays. Fine for annotation-scale GTFs; consumers feeding >~500 MB GTFs should override `memory` / `cpus` directly via `withName`.

## Test plan

- [x] `nf-test test subworkflows/nf-core/gtf_hybridmerge_gffcompare/tests/main.nf.test --profile docker` passes locally (2 tests covering both the no-blacklist passthrough branch and the with-blacklist intersect branch).
- [x] `nf-core subworkflows lint gtf_hybridmerge_gffcompare` clean (28 passed, 0 warned, 0 failed).
- [x] `pre-commit run` clean.
- [x] CI green.

Tests use existing `nf-core/test-datasets:modules` fixtures: `chr21_gencode.gtf` as reference + backbone, `genome.gtf` (chr22 subset) as the novel input, `genome.blacklist_intervals.bed` (chr22) as the blacklist. The cross-contig setup gives every chr22 novel transcript a `class_code "u"` against the chr21 reference, exercising the class-code filter, the blacklist subtraction, and the synth-gene-row code path in one chain. No new test fixtures required.

Beyond the snapshot, the tests assert:
- every non-comment line in the hybrid GTF carries a well-formed `gene_id "..."` attribute (catches gawk/Groovy string-escaping regressions),
- the hybrid_gtf meta equals the backbone meta exactly (catches both meta-source mistakes and key-leakage),
- at least one synthesised `gene` row appears on the novel contig (confirms the synth code path actually fired),
- every synthesised gene row's span equals the min(start) / max(end) of its child transcripts (catches a regression to first-transcript coords).


subworkflows/nf-core/gtf_hybridmerge_gffcompare/nextflow.config
```diff
@@ -0,0 +1,126 @@
...
+     *
+     * The buffer-then-emit shape removes the otherwise-fragile assumption that
+     * gffcompare emits each transcript row before its children: if a future
+     * release reorders, a single-pass print-as-you-see implementation would
+     * drop exons silently. Here, every record waits for the END pass after the
+     * full set of kept transcript_ids is known.
+     */
+    withName: 'GTF_HYBRIDMERGE_GFFCOMPARE:GAWK_FILTER' {
+        ext.args   = { "-v codes_csv='${meta.class_codes}'" }
+        ext.args2  = '\'BEGIN { n = split(codes_csv, arr, /,[[:space:]]*/); for (i = 1; i <= n; i++) keep_code[arr[i]] = 1 } /^#/ { next } $7 == "." { next } { if (match($0, /transcript_id "([^"]+)"/, t)) { tid = t[1]; if (match($0, /class_code "([^"]+)"/, c)) { if (c[1] in keep_code) keep_tx[tid] = 1 } lines[NR] = $0; line_tid[NR] = tid } } END { PROCINFO["sorted_in"] = "@ind_num_asc"; for (i in lines) { if (line_tid[i] in keep_tx) print lines[i] } }\''
```
jonasscheid: This is very complex. Any way to reduce complexity?

subworkflows/nf-core/gtf_hybridmerge_gffcompare/nextflow.config
```diff
@@ -0,0 +1,126 @@
...
+     *     gene_id only gets one parent row, regardless of how many child
+     *     transcripts it has.
+     *
+     * Why union spans: GTF readers that take gene-row coords at face value
+     * (genome browsers, custom span filters) would otherwise see the gene
+     * limited to the first novel transcript's footprint instead of the full
+     * locus. Tools that ignore gene-row coords and re-derive from children
+     * (STAR, Salmon, featureCounts, gffread) are unaffected either way.
+     */
+    withName: 'GTF_HYBRIDMERGE_GFFCOMPARE:GAWK_CONCAT' {
```
jonasscheid, line 40: Same here..

subworkflows/nf-core/gtf_hybridmerge_gffcompare/nextflow.config
jonasscheid: Maybe by giving this code as `program_file`?

jonasscheid (changes requested):

subworkflows/nf-core/gtf_hybridmerge_gffcompare/nextflow.config
pinin4fjords: I thought about this during development but couldn't think how to package the programs alongside the subworkflow- but now done with `moduleDir`.

jonasscheid (approved):
LGTM, good idea with outsourcing the awk programms

PR #11733, opened 2026-05-21
new module: custom/bed12codonpositions

pinin4fjords:
## Why

A pipeline that needs codon-level genomic positions along a spliced transcript (ribo-seq P-site counts per codon, frame / periodicity QC, novel-ORF tiling) has nothing off-the-shelf to reach for. `bedtools makewindows` takes BED3 and ignores BED12 blocks; chaining `bedtools bed12tobed6 | bedtools makewindows` re-anchors each exon at offset 0, so codons crossing an intron land in the wrong frame. No existing nf-core module emits per-codon BED from a BED12.

## What it does

For each BED12 record, walks the blocks in mRNA order (5'→3'), emits every `--step`-th mRNA position starting at `--frame`, and projects each back to genomic coordinates. With `--width N > 1` it emits an N-nt span per position, splitting at block boundaries so a codon that crosses an intron becomes two rows whose union still maps to a contiguous mRNA region.

### Worked example

Two-exon `-` strand record with one intron (spliced mRNA length 20):
```
chr1    100    210    geneX    500    -    100    210    0    2    10,10    0,100
```

Default args (`--step 3 --width 1`, one row per codon at its 5'-most nt) — i.e. the BED you'd intersect with an offset-corrected ribo-seq BAM to count P-sites per codon:
```
chr1    209    210    geneX    500    -
chr1    206    207    geneX    500    -
chr1    203    204    geneX    500    -
chr1    200    201    geneX    500    -
chr1    107    108    geneX    500    -
chr1    104    105    geneX    500    -
chr1    101    102    geneX    500    -
```

Rows are in 5'→3' mRNA order, so on the `-` strand the genomic coordinates count down. The jump from `200` back to `107` is the spliced intron.

With `--width 3` (full 3-nt codon spans), the codon at mRNA position 9 straddles the intron — its 5' nucleotide is in the upstream exon and the other two are in the downstream exon, so it's split into two BED rows (4th and 5th below):
```
chr1    207    210    geneX    500    -
chr1    204    207    geneX    500    -
chr1    201    204    geneX    500    -
chr1    200    201    geneX    500    -
chr1    108    110    geneX    500    -
chr1    105    108    geneX    500    -
chr1    102    105    geneX    500    -
```

Score (column 5) is preserved. Frame is taken from the record's own start, so the module works regardless of any GTF `phase` annotation — running it three times with `--frame 0/1/2` gives you the three per-frame BEDs needed for ribo-seq periodicity QC.

## I/O

- Input: `tuple val(meta), path(bed12)`
- Output: `tuple val(meta), path("${prefix}.bed")` (BED6) + versions topic
- `ext.args`: `--frame INT` (0), `--step INT` (3), `--width INT` (1), `--keep-duplicates`

## Container

Wave-built `community.wave.seqera.io/library/python_pandas_pyyaml:75514f9f977be607` (with matching `community-cr-prod.seqera.io` singularity blob).

## Test plan

- [x] nf-test (default, width 3, frame 1, intron fixture at width 1 and 3, keep-duplicates, stub)
- [x] nf-core modules lint clean (2 Wave-container false positives)
- [x] pre-commit clean
- [x] CI green


SPPearce (dismissed):
Can you swap to just one nextflow.config and use `module_args` please.

pinin4fjords:
Done!

modules/nf-core/custom/bed12codonpositions/templates/bed12codonpositions.py
SPPearce, line 1: Could do with a license for the script.

modules/nf-core/custom/bed12codonpositions/tests/main.nf.test
```diff
@@ -0,0 +1,210 @@
...
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(
+                    process.out.bed,
+                    process.out.findAll { key, val -> key.startsWith('versions') }
+                ).match() }
```
SPPearce, line 35:
```suggestion
                { assert snapshot(sanitizeOutput(process.out)).match() }
```

modules/nf-core/custom/bed12codonpositions/tests/main.nf.test
```diff
@@ -0,0 +1,210 @@
...
```
SPPearce:
```suggestion
                { assert snapshot(sanitizeOutput(process.out)).match() }
```

SPPearce (dismissed):
Can just be sanitizeOutput(process.out) everywhere here.
Also, can we add the yaml parsing of the versions.yml file at least once, so it is captured directly in the snapshot, so we can see that it is being created correctly.

SPPearce (approved):

PR #11740, opened 2026-05-21
feat(custom): orfnormalise + orfmerge modules + orftable_fasta_gtf_buildorfcatalogue subworkflow

pinin4fjords:
## What

Ribo-seq experiments measure which open reading frames (ORFs) on the transcriptome are actively translated by ribosomes. Five widely-used callers - **RiboCode, Ribo-TISH, Ribotricer, Rp-Bp, PRICE** - each emit ORF predictions in a different table format with different score semantics and classification vocabularies. There's no upstream way to combine them into one cohort catalogue.

This PR adds three components that fill that gap:

- **`custom/orfnormalise`** - parses one caller's output into a unified BED12 + sidecar TSV. The caller (one of ribocode, ribotish, ribotricer, rpbp, price) is supplied as a `val` input.
- **`custom/orfmerge`** - class-aware clustering of normalised calls across callers and samples into one cohort catalogue.
- **`orftable_fasta_gtf_buildorfcatalogue`** subworkflow - composes the two with `bedtools/getfasta` + `seqkit/translate` to produce a catalogue AA FASTA.

Bundled because the merger's schema is defined by the normaliser, and the subworkflow only makes sense once both modules exist.

## Harmonised schema

`orf_class` is the union vocabulary across the five callers:

| Class | Meaning |
|---|---|
| `canonical_cds` | Maps to an annotated CDS (or truncated / extended variant) |
| `uORF` / `dORF` | ORF in the 5'UTR / 3'UTR of an annotated transcript |
| `novel_u` | Novel / intergenic ORF, no annotated host CDS |
| `smORF` | `aa_length <= 100`, regardless of location |
| `other` | Internal / overlap / frame variants |

Score direction is per-caller (p-values lower-better, Bayes factors / phase scores higher-better); the merger keeps the direction-appropriate best per cluster.

Merger clustering is class-aware:
- `canonical_cds`: by `(transcript_id, strand)` (one CDS per transcript).
- `uORF` / `dORF` / `other`: by `(transcript_id, strand, start, end)` (a transcript can host multiple distinct positional ORFs).
- `novel_u` / `smORF`: greedy reciprocal-overlap (default 0.8).

## Configurable source columns + provenance

Each caller's output exposes multiple candidate columns for `score` (Ribo-TISH alone has FisherPvalue / RiboPvalue / TISPvalue plus Q-value variants), `orf_type`, and length. Defaults are in `meta.yml`; override per-call via `ext.args`:

```groovy
withName: 'CUSTOM_ORFNORMALISE_RIBOTISH' {
    ext.args = '--score-field FisherQvalue'  // FDR-adjusted, not raw p-value
}
```

Every normalised TSV starts with a `# parser_columns: ...` line recording exactly which source column was read for each derived field, so downstream consumers (and reviewers) can verify provenance. Standard CSV / pandas readers skip `#`-prefixed lines.

## Example output

Cohort catalogue TSV (`called_by_*` / `score_*` track which callers contributed):

```
orf_id        chrom  start     end       strand  gene_id          orf_class      aa_length  called_by_ribotish  called_by_ribocode  score_ribotish  score_ribocode
orf_00000001  20     16731702  16741005  +       ENSG00000125870  canonical_cds  224        0                   1                                   0.0101442
orf_00000002  20     17570208  17607146  +       ENSG00000125868  canonical_cds  164        0                   1                                   0.00112689
```

Subworkflow also emits multi-block BED12 (intron-skipping), `orf_to_gene.tsv` (many-to-many ORF → host lookup), catalogue AA FASTA, and a MultiQC custom-content per-class-count sidecar.

## Test data

Paired with nf-core/test-datasets#2070 (five real-tool outputs sliced to <13 KB each from existing module-test outputs; full provenance in the test-data README). Tests reference the fork branch inline; will swap to `params.modules_testdata_base_path` once #2070 merges.

## Test plan

- [x] All 5 callers' fixtures round-trip through `custom/orfnormalise` with content-level assertions (non-zero `aa_length`, populated `score`, non-collapsed `orf_class` distribution, multi-block BED12 for multi-exon callers, provenance header present).
- [x] `custom/orfmerge` chains two normaliser invocations and asserts coherent `called_by_*` / `score_*` columns.
- [x] Subworkflow chains GUNZIP + normaliser × 2 + merger + getfasta + translate end-to-end.
- [x] `--score-field` override exercised; provenance header confirms it.
- [x] `nf-core modules lint` + `nf-core subworkflows lint` clean (only the two cosmetic Wave container-version warnings shared with custom/bed12codonpositions remain).


iraiosub (approved):
From my side, this implementation looks good.

pinin4fjords:
Thanks @iraiosub !

PR #11752, opened 2026-05-22
simplify meta-schema and try out prek parallization

mashehu:
Schema:
- update to draft 2020-12
- reduce redundancies in schema

pre-commit-config:
- update outdated ruff hook id
- add `priority` fields making test that touch different files in parallel (prek only feature, but pre-commit just throws a warning)


JoseEspinosa (approved):
LGTM 🎸

PR #11755, opened 2026-05-24
Improve resource usage in `glimpse2`

atrigila:
The `test_full` of phaseimpute show that these two modules receive a lot of unnecessary resources:
- `GLIMPSE2_CHUNK` shows only ~200–400 MB of RAM used and moderate CPU utilisation. 
- `GLIMPSE_PHASE` the job peaks at only one or two cores with <1 GB RAM.
Therefore, I am updating the labels to `process_single`. 

Additionally, the subworkflow `BAM_VCF_IMPUTE_GLIMPSE2` is improved by using BCF intermediate outputs. The ligation process uses less resources if the input files are BCF: 
- With `vcf.gz` `GLIMPSE2_LIGATE` used 301.2 MB peak RSS, 527.3 MB peak vmem, duration 1.6s.
- With `bcf` `GLIMPSE2_LIGATE` used 13.3 MB peak RSS, 23.3 MB peak vmem, duration 626ms.

Also, this fixes a linting error: "Invalid 'ext' keys detected: ext.suffix ".

### DIFF
--- a/modules/nf-core/glimpse2/chunk/main.nf
+++ b/modules/nf-core/glimpse2/chunk/main.nf
@@ -1,6 +1,6 @@
 process GLIMPSE2_CHUNK {
     tag "${meta.id}"
-    label 'process_low'
+    label 'process_single'
 
     beforeScript """
         if cat /proc/cpuinfo | grep avx2 -q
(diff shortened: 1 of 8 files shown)


LouisLeNezet (approved):
Just a question, how did you check for this resource usage ?

atrigila:
I checked the test full results and resource usage per task

Issue #11835, opened 2026-06-03
new module: BWA-FASTALIGN


peterpru:
### Is there an existing module for this?

- [x] I have searched for the existing module

### Is there an open PR for this?

- [x] I have searched for existing PRs

### Is there an open issue for this?

- [x] I have searched for existing issues

### Further Information

This issue aims to add the BWA fastalign tool to nf-core modules.

https://github.com/zzhofict/BWA-FastAlign

### Are you going to work on this?

- [x] If I'm planning to work on this module, I added myself to the `Assignees` to facilitate tracking who is working on the module

### Software requirements (conda)

```yaml

```

peterpru:
bwa-fastalign will need a new reference genome indexing, different from e.g. bwa-mem, so this has to be two modules.

```
bwa-fastalign/mem
bwa-fastalign/index
```

PR #11840, opened 2026-06-03
New module bwafastalign-mem

peterpru:
This PR add the bwa-fastalign mem module.

This modules uses the fastalign index subtool in the test, so that one has to be merged first.

There is a linked issue here for the index and mem subtools: https://github.com/nf-core/modules/issues/11835

## PR checklist

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


peterpru:
Need to add bioconda/containers, test snapshots and merge bwa-fastalign index.

modules/nf-core/bwafastalign/mem/main.nf
```diff
@@ -0,0 +1,87 @@
...
+    def args = task.ext.args ?: ''
+    def args2 = task.ext.args2 ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def samtools_command = sort_bam ? 'sort' : 'view'
+    if (!mbuffer) {
+        log.info '[bwafastalign-mbuffer] Memory for mbuffer is not set - defaulting to 3GB for mbuffer.'
+        mbuffer_mem = 3072
+    } else {
+        mbuffer_mem = mbuffer
+    }
```
maxulysse: Let's use something like we have in gatk4 modules to handle memory:

https://github.com/nf-core/modules/blob/master/modules/nf-core/gatk4/combinegvcfs/main.nf#L28-L34

maxulysse: and I think we can do something similar to handle the threads

ramprasadn: Hmmm… I would say this is a special case. In bwa-fastalign, the memory footprint is determined entirely by the reference size and isn’t configurable. That means any formula for configuring memory and cpus  would be based on assumptions we can’t validate now. So it is best set by user.

maxulysse: We can so something like https://github.com/nf-core/modules/blob/master/modules/nf-core/bwamem2/index/main.nf#L5

FriederikeHanssen: can't we default to machine size? What's the reason not to default to task.memory?

ramprasadn: Just to provide some context on why mbuffer is there, bwa-fastalign is high-throughput and produces alignments faster than samtools sort can handle during phase 1 (which runs concurrently with the aligner), so we use mbuffer as a middleman to store alignments until samtools is ready to process them.

Making assumptions about memory is hard here because the issue is the internal split. Of course, bwa-fastalign consumes most of the process memory for the index, but the remaining memory allocated to samtools + mbuffer depends on factors outside the module's knowledge (reference size, disk I/O speed, and whether the time spent in sort phase 2, which is executed after alignment finishes, matters to the user). Any percentage we pick would be based on unvalidated assumptions and might not work on all machines, as we could end up over- or under-allocating resources. So I think it's best left to the user. What do you guys think?

FriederikeHanssen: Got it. I think where I am still stuck is, why does task.memory not fulfill this, which the user also should specify and right size? it doesn't look to me the mbuffer is computed in any way, but just taken at face value.

ramprasadn: Based on our tests, it looks like mbuffer memory is not shared with other processes. Therefore, if we allocate all of task.memory to it, there would be no memory left for samtools sort.

modules/nf-core/bwafastalign/mem/main.nf
```diff
@@ -0,0 +1,87 @@
...
+
+    input:
+    tuple val(meta), path(reads)
+    tuple val(meta2), path(index)
+    tuple val(meta3), path(fasta)
+    val   sort_bam
+    val   mbuffer
+    val   samtools_threads
+
+    output:
```
SPPearce, line 16: Can we please just emit one channel with the actual aligned file and one channel with the index? So you don't have to have lots of really awkward joins 🥺

ramprasadn: Done 👍🏻

modules/nf-core/bwafastalign/mem/main.nf
jfy133: Naive response while skim reading this:

>  bwa-fastalign is high-throughput and produces alignments faster than samtools sort can handle during phase 1

> so we use mbuffer as a middleman to store alignments until samtools is ready to process them

If that is the case, why not just drop the pipe into samtools and have a separate process (greatly simplifying the module and better adhereing to the specs?)

ramprasadn: The benefit of using bwa-fastalign is that it produces alignments faster than other aligners. The mbuffer in the middle exists specifically to absorb that output while samtools is flushing its sort buffer to disk. Splitting into two processes would force writing an unsorted SAM to disk first, which defeats the purpose of the tool and removes the benefit mbuffer provides. Keeping them piped also means this module can serve as a drop-in replacement for other bwa modules that produce sorted BAMs directly, which makes it easier to use in existing pipelines.

jfy133: Fair enough!

pontus: So, I'm thinking is that this adds knobs for behaviour many users (and pipeline authors) will probably not know how to tune correctly (meaning we should have good defaults). But having them as inputs adds some form of cognitive load/stress to the pipeline author to pass good values (which is very likely to depend on the users system, so not an easy task) or expose them to the user (meaning more cruft).

I also may be forgetting something, but assuming the pipeline author hardcodes these values, won't that be very difficult for the user to adjust in the case they need to?

I haven't tried formulating it exactly (with good defaults), but it seems to me this would better be addressed by some `ext.args`-thingy.

SPPearce: Yes, but we're not allowed to add extra `ext.` thingys, because :shrug:

modules/nf-core/bwafastalign/mem/main.nf
```diff
@@ -0,0 +1,84 @@
...
+    tuple val(meta), path("${prefix}.{sam,bam,cram}"), emit: output, optional: true
```
SPPearce: Why is this optional? ;)

modules/nf-core/bwafastalign/mem/main.nf
ramprasadn: Thanks for the feedback @pontus ! We do have sensible defaults built into the module (3GB for mbuffer, 2 threads for samtools) so pipeline authors who don't expose these get reasonable behaviour out of the box. For users who need to tune them, I think it's better if pipeline devs expose them as pipeline parameters which I think is cleaner than ext.args, users get simple CLI flags without needing to know how to write config overrides. And as CLI flags they are more prominent and discoverable for users compared to buried config entries. What do you think?

pontus: This may just be me being pessimistic, but having pipeline developers expose and provide some short explanation and users wanting to fill everything in does not always lead to optimal results in my experience (to the degree that I'm thinking it /might/ be better to hardcode this non-configurable than offering that). If there's no koscher way of dealing with it through `ext.args`, that's just how it is, but if that's possible I'd have strong preference for that.

(If it's in any way unclear; I don't think more prominent and discoverable for users is a good thing here - most often seeing the code and understand it is enough to make good choices, but here they also need to understand a fair bit about the machines where they'll run and so on.)

ramprasadn: That's a fair point. The reason I kept these as inputs rather than hardcoding them or moving them to ext.args is that there isn't really a good universal default. The appropriate values depend on factors like the reference genome size and available hardware, which the module itself has no visibility into. For example, a 3 GB mbuffer might be fine for a viral genome on a small machine, but completely inappropriate for human genome alignment on an HPC node. Pipeline developers generally have that context, so exposing these as inputs allows them to choose sensible defaults for their specific use case. That said, I'm open to suggestions if there's a cleaner way to handle it.

modules/nf-core/bwafastalign/mem/main.nf
ramprasadn: 🤦🏻‍♂️  Good catch! I will fix it.

modules/nf-core/bwafastalign/mem/main.nf
SPPearce: Feel free to weigh in on (and vote on) this PR for a `task.ext.mem_multiplier` that was designed as a way to specify with a config this kind of override: https://github.com/nf-core/proposals/issues/65.

ramprasadn: On second thought, here's what I'll do. I'll remove `mbuffer` and `samtools_threads` as module inputs and make them configurable via ext.args2 (mbuffer) and ext.args3 (samtools) instead. I'll also include a README explaining how to set these, similar to what we have for the DeepVariant module. If this turns out to be problematic in practice, we can always revisit it. What do you guys think?

maxulysse (approved):

maxulysse:
I'd say we're good to go

PR #11859, opened 2026-06-04
Update DEEPVARIANT/* modules to support phasing. 

awgymer:
Update the split processes for DEEPVARIANT to support using phased read inputs passed between processes. 

Also fix the capture pattern for small model call records output in MAKE_EXAMPLES for prefixes containing a `.`

See [google/deepvariant#1078](https://github.com/google/deepvariant/issues/1078) for more details on the root cause of these issues and need to fix. 

## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


SPPearce:
The tests and linting are sad.

awgymer:
Yeah, setting the phasing to false fixes the tests but I am waiting for some feedback on whether a hard failure like this is expected with the `WGS` mode (some of their docs suggest it only intended for ONT/Pacbio)

awgymer:
Ok. They got back to me:

> The phasing here is implemented for long-read applications (PacBio and ONT). It's not implemented in a way where you could get meaningful results with short reads.

I am wondering whether it is OK to just accept that the old behaviour is unaffected and that we don't have tests for the long-read profiles which would leverage this new option (beyond stubs)?

awgymer:
I've added a couple of tests which use the `--model PACBIO` setting which allows the phasing logic to be properly tested. The biological correctness is of course rubbish since the inputs are actually short-reads, but I think this is a satisfactory situation for testing that the module code works?

SPPearce:
Sounds fine, biological correctness is not required

awgymer:
The lint failures are due to an issue with the container prefix (but the container hasn't changed at all - it's just a `ghcr` one because that's what we've always used) so could you possibly review and we can try and force merge

modules/nf-core/deepvariant/makeexamples/tests/pacbio_profile.config
```diff
@@ -0,0 +1,14 @@
+process {
+    withName: "DEEPVARIANT_MAKEEXAMPLES" {
+        ext.args = '--checkpoint "/opt/models/pacbio" --call_small_model_examples --small_model_indel_gq_threshold "30" --small_model_snp_gq_threshold "25" --small_model_vaf_context_window_size "51" --trained_small_model_path "/opt/smallmodels/pacbio/model.keras"'
+    }
+    withName: "DEEPVARIANT_CALLVARIANTS" {
```
maxulysse, line 5:
```suggestion
    withName: DEEPVARIANT_MAKEEXAMPLES {
        ext.args = '--checkpoint "/opt/models/pacbio" --call_small_model_examples --small_model_indel_gq_threshold "30" --small_model_snp_gq_threshold "25" --small_model_vaf_context_window_size "51" --trained_small_model_path "/opt/smallmodels/pacbio/model.keras"'
    }
    withName: DEEPVARIANT_CALLVARIANTS {
```

maxulysse: let's keep the config file consistent

subworkflows/nf-core/deepvariant/tests/pacbio_profile.config
```diff
@@ -0,0 +1,14 @@
+process {
+    withName: "DEEPVARIANT_MAKEEXAMPLES" {
+        ext.args = '--checkpoint "/opt/models/pacbio" --call_small_model_examples --small_model_indel_gq_threshold "30" --small_model_snp_gq_threshold "25" --small_model_vaf_context_window_size "51" --trained_small_model_path "/opt/smallmodels/pacbio/model.keras"'
+    }
+    withName: "DEEPVARIANT_CALLVARIANTS" {
+        ext.args = '--checkpoint "/opt/models/pacbio"'
```
maxulysse, line 6: I am slightly confused by the paths to models and such

maxulysse (approved):
I think we can make the assertions in the tests better, but fine with me as is.
I would just like to keep the way we write config consistent

subworkflows/nf-core/deepvariant/tests/pacbio_profile.config
awgymer: The paths specifically? This is a new style of passing the args/checkpoint which the DeepVariant guys implemented, partly to help make it easier to use the split-command style call in automated situations such as nextflow (as opposed to the earlier recommended solution of running `rundeepvariant --dry-run` and copy-pasting the commands). 

I used the `pacbio` models here so that the phasing can be tested (it only works with long-reads model types)

modules/nf-core/deepvariant/makeexamples/tests/pacbio_profile.config
awgymer: If you look at the existing `nextflow.config` you'll see this is copied directly from there, so removing the quotes would be *inconsistent* with the existing code?

PR #11059, opened 2026-03-26
Add snpclustering subworkflow

dbaku42:
## Description

This PR adds the `snpclustering` subworkflow for end-to-end unsupervised clustering of genomic samples directly from multi-sample VCF files.

### Features
- Variant filtering (MAF + missingness) with `bcftools/filter`
- LD pruning with `plink2/indeppairwise`
- Export pruned VCF with `plink2/recodevcf`
- PCA with `flashpca2`

The subworkflow was developed in relation to the accepted nf-core proposal for the `consepopgen` pipeline.

**Related to:**  
- https://github.com/nf-core/proposals/issues/57 (nf-core/consepopgen)

### Checklist
- [x] `nf-core subworkflows lint snpclustering` passed
- [x] `nf-core subworkflows test snpclustering` passed
- [x] Follows nf-core subworkflow conventions

Closes # (no specific issue)

PR #11879, opened 2026-06-04
Add snpclustering subworkflow

dbaku42:
## Description
 
This PR adds the `snpclustering` subworkflow for end-to-end unsupervised clustering of genomic
samples from genotyped multi-sample VCF files.
 
This is a revised and extended version of the subworkflow originally proposed in #11059, which was
based on FlashPCA2 (not available as an nf-core module). The current implementation replaces it
with official nf-core modules throughout and extends the pipeline with imputation, PCA-based
clustering, quality metrics, and visualization steps.
 
> **Note:** This subworkflow depends on the modules added in #11372
> (`custom/pcaclustering`, `custom/clustermetrics`, `custom/clustervisualization`),
> which should be merged first.
 
### Pipeline
 
```
VCF → BEAGLE5_BEAGLE → PLINK2_VCF → PLINK2_PCA → EIGENVEC_TO_TSV
                                                         ↓
                                          CUSTOM_PCACLUSTERING
                                         ↙                    ↘
                          CUSTOM_CLUSTERMETRICS    CUSTOM_CLUSTERVISUALIZATION
```
 
### Features
 
- Genotype imputation and phasing with `beagle5/beagle` (reference panel and genetic map optional)
- Conversion to PLINK2 pgen format with `plink2/vcf`
- PCA with `plink2/pca`
- Inline `.eigenvec` → TSV conversion via an embedded `EIGENVEC_TO_TSV` process (gawk)
- Unsupervised clustering (k-means, DBSCAN) with `custom/pcaclustering`
- Cluster quality metrics with `custom/clustermetrics`
- UMAP/t-SNE visualization with `custom/clustervisualization`
### Related PRs
 
- #11059 — previous version of this subworkflow (FlashPCA2-based, superseded by this PR)
- #11372 — adds `custom/pcaclustering`, `custom/clustermetrics`, `custom/clustervisualization


subworkflows/nf-core/snpclustering/tests/nextflow.config
```diff
@@ -0,0 +1,44 @@
...
+    use_approx = true
+    algorithm = "kmeans"
+    n_clusters = 3
+    dbscan_eps = 0.5
+    dbscan_min_samples = 5
+}
+
+process {
+    withName: 'BEAGLE5_BEAGLE' {
+        ext.prefix = { "${meta.id}" }
```
famosab: That should be the defacto standard for all modules so I do not think you need these

subworkflows/nf-core/snpclustering/main.nf
```diff
@@ -0,0 +1,156 @@
+include { BEAGLE5_BEAGLE } from '../../../modules/nf-core/beagle5/beagle/main'
+include { PLINK2_VCF } from '../../../modules/nf-core/plink2/vcf/main'
+include { PLINK2_PCA } from '../../../modules/nf-core/plink2/pca/main'
+include { CUSTOM_PCACLUSTERING } from '../../../modules/nf-core/custom/pcaclustering/main'
+include { CUSTOM_CLUSTERMETRICS } from '../../../modules/nf-core/custom/clustermetrics/main'
+include { CUSTOM_CLUSTERVISUALIZATION } from '../../../modules/nf-core/custom/clustervisualization/main'
+
```
famosab:
```suggestion
include { BEAGLE5_BEAGLE } from '../../../modules/nf-core/beagle5/beagle'
include { PLINK2_VCF } from '../../../modules/nf-core/plink2/vcf/main'
include { PLINK2_PCA } from '../../../modules/nf-core/plink2/pca/main'
include { CUSTOM_PCACLUSTERING } from '../../../modules/nf-core/custom/pcaclustering/main'
include { CUSTOM_CLUSTERMETRICS } from '../../../modules/nf-core/custom/clustermetrics/main'
include { CUSTOM_CLUSTERVISUALIZATION } from '../../../modules/nf-core/custom/clustervisualization/main'

```

trailing main can be removed
please align the } nicely

subworkflows/nf-core/snpclustering/tests/nextflow.config
dbaku42: Thanks @famosab
I removed them

dbaku42:
thank you @famosab,
Is anything else to fix?

subworkflows/nf-core/snpclustering/tests/main.nf.test.snap
famosab, line 1: It seems like all of the snaps are empty?

dbaku42: Hi @famosab, thanks for catching this — you're right, the snapshots are empty.

I've been trying to figure out why: running the subworkflow directly with nextflow run (outside nf-test) does show populated channels and the files do exist in the nf-test work directories, so I don't think it's a dataflow issue in the subworkflow itself — but inside nf-test, workflow.out comes back empty no matter how I access it (direct .collect{} on named channels, sanitizeOutput(workflow.out), size checks). I also tried removing the version-topic channel mixing in case that was interfering, with no change.

I haven't been able to pin down the actual cause, so I don't want to guess at a fix. Could you point me in the right direction — is there a pattern I should be using for subworkflows with this many emit channels, or something specific to check in how the test is written? Happy to make whatever change is needed.

dbaku42: I’ve now resolved the empty snapshots issue. The root cause was that the test nextflow.config was setting ext.when = false for all processes in the subworkflow, so nothing was actually executed during the nf-test runs.
I removed those lines, re-ran the tests with --profile docker --update-snapshot, and the snapshots are now properly populated with the expected outputs (meta + filenames + versions).
Could you please take another look when you have a moment? Thanks!

subworkflows/nf-core/snpclustering/tests/main.nf.test
```diff
@@ -0,0 +1,162 @@
...
+                """
+            }
+        }
+
+        then {
+            assert workflow.success
+            assert snapshot(
+                workflow.out.clusters.collect { [
+                    it[0],
+                    path(it[1]).getFileName().toString()
```
famosab: Are all the md5sums unstable? Can we maybe at least assert parts of the output? Only the filename is just the last resort. You can check other subworkflows how to get to the md5sums.

subworkflows/nf-core/snpclustering/main.nf
```diff
@@ -0,0 +1,163 @@
...
+            [],
+            [],
+            [],
+            [],
+            region
+        )
+    }
+
+    BEAGLE5_BEAGLE(ch_beagle_input)
+    ch_versions = ch_versions.mix(BEAGLE5_BEAGLE.out.versions_beagle)
```
famosab: I think its maybe a bit much to ask but as it only relates to three modules: Would you open a separate PR and swap these three modules to topics? Then we can have this subworkflow clean and it does not need to be touched another time after the modules are swapped :)

subworkflows/nf-core/snpclustering/main.nf
```diff
@@ -0,0 +1,163 @@
...
+
+    umap_tsv      = CUSTOM_CLUSTERVISUALIZATION.out.umap_tsv
+    tsne_tsv      = CUSTOM_CLUSTERVISUALIZATION.out.tsne_tsv
+    umap_png      = CUSTOM_CLUSTERVISUALIZATION.out.umap_png
+    tsne_png      = CUSTOM_CLUSTERVISUALIZATION.out.tsne_png
+
+    versions      = ch_versions
+}
+
+process EIGENVEC_TO_TSV {
```
famosab: Can you move this process to its own module ( you can use https://github.com/nf-core/modules/tree/master/modules/nf-core/custom ) to put it somewhere

dbaku42:
@famosab I'm running into unstable md5 sums in the nf-test snapshots here — likely from the UMAP/t-SNE PNG plots and BEAGLE5/PLINK2 log files, which aren't deterministic across runs.

How would you suggest I proceed?

dbaku42:
Hi @famosab, 
I tried to pin down the unstable md5s from the subworkflow test config. Snapshots now hash the TSV, cluster assignments, clustering_info.json and metrics.tsv. 
Does this config look acceptable to you?

subworkflows/nf-core/snpclustering/main.nf
dbaku42: Hi @famosab,
I’m still a bit new to the topics migration — when you say “swap these three modules to topics”, do you mean only the versions channel, or also the other outputs (clusters, metrics, plots, …)?
The three custom/* modules already emit
path "versions.yml", emit: versions, topic: versions on master. I just want to make sure I change the right thing in #11879 and don’t touch more than needed. 
Thanks so much!

dbaku42:
@famosab a short update on the test/config cleanup, then a question on merge timing.

`tests/nextflow.config` now only sets what the spec allows plus one required `ext` key:
- `ext.args` seeds on BEAGLE5 / PLINK2_VCF / PLINK2_PCA so hashed PCA/cluster files stay stable
- `ext.suffix = 'tsv'` on `GAWK_EIGENVEC_TO_TSV` so the gawk output does not collide with `${meta.id}.eigenvec` (the module asserts this)

Removed from that file: `env {}`, `cpus`/`memory`, and `ext.args` that only repeated Python defaults.

The eigenvec → TSV conversion (`OFS`, `#` strip, `IID` → `sample_id`) lives in `awk/eigenvec_to_tsv.awk` so it is not test-only.

Snapshots no longer md5 non-deterministic files. We still assert content where it is stable, following the `custom/pcaclustering` module test:
- `clusters.csv`, `metrics.tsv`, `selected.json`, eigenvec/eigenval/tsv hashed
- `clustering_info.json` parsed; kmeans `inertia` rounded to 4 decimals
- UMAP/t-SNE TSV: header + n_rows + n_cols (coordinates differ conda vs docker, as in #12771)
- logs / VCF / pgen / pvar / PNG / k_sweep: filename only

On topics: the three custom modules already emit `path "versions.yml", emit: versions, topic: versions` on master. This subworkflow does not `mix` `.out.versions`, so it is born on topics. #12771 was closed for that reason.

Happy to change anything you still want touched. When you have a moment, is there anything left before this can be merged.

subworkflows/nf-core/snpclustering/main.nf
```diff
@@ -0,0 +1,125 @@
...
+include { CUSTOM_CLUSTERMETRICS         } from '../../../modules/nf-core/custom/clustermetrics'
+include { CUSTOM_CLUSTERVISUALIZATION   } from '../../../modules/nf-core/custom/clustervisualization'
+
+workflow SNPCLUSTERING {
+    take:
+    vcf_ch
+    refpanel_ch
+    genmap_ch
+    region
+    npcs
```
nschan:
```suggestion
    n_pcs
```

subworkflows/nf-core/snpclustering/main.nf
```diff
@@ -0,0 +1,125 @@
...
+
+    /*
+     * PLINK2_PCA expects:
+     * tuple val(meta), val(npcs), val(use_approx), path(pgen), path(psam), path(pvar)
+     */
+    ch_plink_pca_input = PLINK2_VCF.out.pgen
+        .join(PLINK2_VCF.out.pvar)
+        .join(PLINK2_VCF.out.psam)
+        .map { meta, pgen, pvar, psam ->
+            tuple(meta, npcs, use_approx, pgen, psam, pvar)
```
nschan:
```suggestion
            tuple(meta, n_pcs, use_approx, pgen, psam, pvar)
```

subworkflows/nf-core/snpclustering/meta.yml
```diff
@@ -0,0 +1,258 @@
...
+            description: Index for the VCF file
+            pattern: "*.{tbi,csi}"
+  - refpanel_ch:
+      description: Optional BEAGLE reference panel channel
+  - genmap_ch:
+      description: Optional BEAGLE genetic map channel
+  - region:
+      type: string
+      description: Optional genomic region passed to BEAGLE
+  - npcs:
```
nschan, line 46:
```suggestion
  - n_pcs:
```

subworkflows/nf-core/snpclustering/main.nf
nschan: It bothers me that it is `n_clusters` but not `n_pcs`

subworkflows/nf-core/snpclustering/main.nf
```diff
@@ -0,0 +1,125 @@
...
+
+    PLINK2_PCA(ch_plink_pca_input)
+
+    /*
+     * Convert .eigenvec to a tab-separated file for downstream clustering
+     * modules: strips the leading '#' from the header and renames the IID
+     * column to sample_id. Output keeps its default naming (<meta.id>.eigenvec)
+     * — only the tab-separated content matters downstream, not the extension.
+     */
+    GAWK_EIGENVEC_TO_TSV(
```
nschan, line 74: Personally, I wonder if having a dedicated process to remove `#` and rename `IID` to `sample_id` is necessary, especially considering that the recipient of this is a custom module, which could handle this trivial conversion if it is required?

dbaku42: Fair question. I left it as a dedicated step so custom/pcaclustering stays generic (any TSV of sample × features) and does not absorb PLINK-specific quirks (# header, IID/FID). The rename to sample_id also matches what clustermetrics / clustervisualization already expect. I’d rather keep that adapter at the subworkflow boundary than special-case PLINK inside the clustering module.

nschan: I think it would be better to special-case PLINK format in your clustering module, since it seems to commonly consume files that were generated in PLINK format. Which in my view would actually make the module more generic?

I understand that this is a custom module, but to me it would facilitate wider adoption if this module was tailored to the inputs, and it is a very simple operation (e.g. `sed 's/#//; s/IID/sample_id/' "${plinkfile}" > "${tsvfile}"` should do it), and I assume that you could simply test if it is necessary based on extension of the input file.

My main concern here is that I assume spawning all of these jobs creates some overhead, and the jobs do very little. Since @famosab has been involved more than I am, maybe there are considerations here that I am not aware of.

dbaku42:
@nschan thanks — this is now in place.

`custom/pcaclustering` (#12904) and `custom/clustermetrics` /
`custom/clustervisualization` (#12906) read PLINK2 `.eigenvec`
(`#FID IID` / `#IID`) directly. This subworkflow no longer runs
`GAWK_EIGENVEC_TO_TSV`; `PLINK2_PCA.out.evecfile` goes straight
into clustering, metrics and visualization.

nschan:
> `custom/pcaclustering` (#12904) and `custom/clustermetrics` / `custom/clustervisualization` (#12906) read PLINK2 `.eigenvec` (`#FID IID` / `#IID`) directly. This subworkflow no longer runs `GAWK_EIGENVEC_TO_TSV`; `PLINK2_PCA.out.evecfile` goes straight into clustering, metrics and visualization.

Great, thanks for taking up that comment and incorporating it. In my view it improves the modules, and streamlines this workflow. I have no further comments. Since I was not involved in this from the beginning, maybe it would be good if @SPPearce and/or @famosab could take another look?

dbaku42:
Hi @famosab, @SPPearce,
Is there anything else to improve?

famosab (approved):
Looks good from my side :)

PR #11950, opened 2026-06-10
Add fgumi/merge

SPPearce:


mashehu (approved):

matthdsm (approved):

modules/nf-core/fgumi/merge/main.nf
```diff
@@ -0,0 +1,37 @@
...
+    tag "${meta.id}"
+    label 'process_medium'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine in ['singularity', 'apptainer'] && !task.ext.singularity_pull_docker_container ?
+        'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/a5/a510706f3481fae12ff6100d6e4ad298b8bf464a2d93a6afe35e9cf26542d080/data':
+        'community.wave.seqera.io/library/fgumi:0.2.0--fe028e7a64e5da27' }"
+
+    input:
+    tuple val(meta), path(bams, stageAs: "?/*")
```
matthdsm, line 11: same question as before, what about indices?

SPPearce: Hmm, not sure if it cares. Template coordinate sorted files don't have one, and it won't generate an index for coordinate sorted files. Index is usually used for direct access, and here it is merging all the reads so needs to go through all the reads anyway.

SPPearce: Claude agrees, saying the code is just streaming through all the files from start to finish

PR #11954, opened 2026-06-10
feat(rpbp/preparegenome): add module

pinin4fjords:
Part of a series splitting #11695 (closed at reviewer request: one PR per module).

Adds `rpbp/preparegenome`, which enumerates all candidate ORFs from a GTF annotation and writes the BED files consumed by every downstream per-sample rpbp step. Runs once per genome/annotation pair.

### Series

| Order | Module / subworkflow | PR |
|-------|----------------------|----|
| genome prep | `rpbp/preparegenome` | #11954 |
| 1 | `rpbp/extractmetageneprofiles` | #11955 |
| 2 | `rpbp/estimatemetagenebayesfactors` | #11956 |
| 3 | `rpbp/selectperiodicoffsets` | #11957 |
| 4 | `rpbp/getperiodiclengthsoffsets` | #11958 |
| 5 | `rpbp/extractorfprofiles` | #11959 |
| 6 | `rpbp/estimateorfbayesfactors` | #11960 |
| 7 | `rpbp/selectfinalpredictionset` | #11961 |
| — | `fasta_gtf_bam_rpbp` subworkflow | #11962 |

PR #11955, opened 2026-06-10
feat(rpbp/extractmetageneprofiles): add module

pinin4fjords:

Adds `rpbp/extractmetageneprofiles`, which computes per-read-length read coverage around annotated start codons (metagene profiles) for a single Ribo-seq BAM. This is step 1 of the 7-step per-sample rpbp chain.

### Series

| Order | Module / subworkflow | PR |
|-------|----------------------|----|
| genome prep | `rpbp/preparegenome` | #11954 |
| 1 | `rpbp/extractmetageneprofiles` | #11955 |
| 2 | `rpbp/estimatemetagenebayesfactors` | #11956 |
| 3 | `rpbp/selectperiodicoffsets` | #11957 |
| 4 | `rpbp/getperiodiclengthsoffsets` | #11958 |
| 5 | `rpbp/extractorfprofiles` | #11959 |
| 6 | `rpbp/estimateorfbayesfactors` | #11960 |
| 7 | `rpbp/selectfinalpredictionset` | #11961 |
| — | `fasta_gtf_bam_rpbp` subworkflow | #11962 |


modules/nf-core/rpbp/extractmetageneprofiles/main.nf
```diff
@@ -0,0 +1,38 @@
...
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    extract-metagene-profiles \\
+        ${bam} \\
+        ${transcript_bed} \\
+        ${prefix}.metagene-profile.csv.gz \\
```
SPPearce: Does the output file need this full name?
Could set a default prefix to include this   
`prefix = task.ext.prefix ?: "${meta.id}.metagene-profile"` otherwise (or shorten the name)

pinin4fjords: I much prefer the descriptor in the filename so it's clear what the file type is

pinin4fjords: ... but maybe I can shorten it if you insist

SPPearce: Put it in the default prefix then, then other people can change it

pinin4fjords: Can you point me at the docs where we say we only allow standard extensions outside the prefix?

pinin4fjords: I really want to retain that additional suffix. It gives me guarantee that I'm not picking up any rogue files with similar extensions, and it guarantees that this type will be recognisable in all pipelines that use it. I've used this pattern lots before without issue, I'm not sure why it's a blocker now.

SPPearce: https://nf-co.re/docs/specifications/components/modules/naming-conventions#command-file-output-naming

> Output file (and/or directory) names SHOULD consist of only ${prefix} and the file-format suffix (for example, ${prefix}.fq.gz or ${prefix}.bam).

modules/nf-core/rpbp/extractmetageneprofiles/main.nf
```diff
@@ -0,0 +1,38 @@
...
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    extract-metagene-profiles \\
+        ${bam} \\
+        ${transcript_bed} \\
+        ${prefix}.metagene.csv.gz \\
```
SPPearce: This will give you what you want, but lets other people override it:
```suggestion
    prefix = task.ext.prefix ?: "${meta.id}.metagene"
    """
    extract-metagene-profiles \\
        ${bam} \\
        ${transcript_bed} \\
        ${prefix}.csv.gz \\
```

pinin4fjords: Honestly I'd rather prescribe something guaranteed not to clash with something else in the dir, but if you insist...

pinin4fjords: Concrete case: set `ext.prefix = "sample"` on both `extractmetageneprofiles` and `estimatemetagenebayesfactors`. The first module outputs `sample.csv.gz`. Nextflow stages that file into the second module's work directory, also as `sample.csv.gz`. The second module's output declaration is `path("${prefix}.csv.gz")` - which now matches the staged input file, not just the tool's actual output.

With the descriptor in the glob instead:

```
# extractmetageneprofiles
ext.prefix = "sample"
→ output: sample.metagene.csv.gz   (staged into next process as sample.metagene.csv.gz)

# estimatemetagenebayesfactors
ext.prefix = "sample"
→ output glob: path("sample.metagene-bayes.csv.gz")   ← no possible collision
```

The descriptor makes the glob specific enough that no plausible input file name can accidentally match it, regardless of what prefix the user chooses. With just `path("${prefix}.csv.gz")`, all it takes is a consistent prefix setting across the chain and you have an ambiguous glob pointing at both the input and the output.

pinin4fjords: But to be clear, the change is applied, I'll start a separate discussion on this if I have the energy

SPPearce: If it is going to cause later collisions, then fixed it into the name.
I don't think you need the `metagene` though.

pinin4fjords: That's fine, this is the standard so I'll play ball and just make sure riboseq keeps it safe. Can I get approval?

FriederikeHanssen (approved):

SPPearce (commented):
Do you not want to set the default to be `profiles` or similar, rather than `metagene`?

pinin4fjords:
> Do you not want to set the default to be `profiles` or similar, rather than `metagene`?

I think this is fine

PR #11956, opened 2026-06-10
feat(rpbp/estimatemetagenebayesfactors): add module

pinin4fjords:

Adds `rpbp/estimatemetagenebayesfactors`, which scores each read length for whether it shows the 3-nucleotide periodicity characteristic of active translation (vs background noise). This is step 2 of the 7-step per-sample rpbp chain.

### Series

| Order | Module / subworkflow | PR | Status |
|-------|----------------------|----|--------|
| genome prep | `rpbp/preparegenome` | #11954 | ✅ merged |
| 1 | `rpbp/extractmetageneprofiles` | #11955 | ✅ merged |
| 2 | `rpbp/estimatemetagenebayesfactors` | #11956 | open |
| 3 | `rpbp/selectperiodicoffsets` | #11957 | ✅ merged |
| 4 | `rpbp/getperiodiclengthsoffsets` | #11958 | ✅ merged |
| 5 | `rpbp/extractorfprofiles` | #11959 | open |
| 6 | `rpbp/estimateorfbayesfactors` | #11960 | open |
| 7 | `rpbp/selectfinalpredictionset` | #11961 | ✅ merged |
| — | `fasta_gtf_bam_rpbp` subworkflow | #11962 | open |

### Implementation notes

A few things in this module are deliberate and may look unusual at first glance:

- **Stan models are resolved from the installed rpbp package, not taken as inputs.** The script discovers them at runtime (`RPBP_MODELS_BASE=$(python3 -c "... inspect.getfile(rpbp) ... 'models'")`, then globs `periodic/*.stan` and `nonperiodic/*.stan`). They are intentionally *not* exposed as `path` inputs: rpbp loads a model via `CmdStanModel(exe_file=Path(model).with_suffix(""))`, which requires the **pre-compiled CmdStan binary to sit next to the `.stan` file**. That binary only exists inside the package install (compiled at container-build time), so a staged user-supplied `.stan` could not be used. The models are version-locked to the pinned rpbp container, which is the correct coupling.

- **Resource label is `process_medium`.** The step fits models for a fixed, library-determined set of (read length, offset) pairs, so its footprint does not scale with input size. Profiling on full-depth data: <1 GiB peak RSS (770 MiB at 12 cpus, 82 MiB at 1 cpu) and clean cpu scaling (162 s → 24 s from 1 → 12 cpus). `process_medium` matches this with large headroom.

- **Output is MCMC-derived, so the test asserts only the deterministic columns.** Inference uses a fixed seed and is reproducible across cpu counts within a toolchain, but the Bayes-factor floats (`p_periodic_*`, `p_nonperiodic_*`, `bayes_factor_*`) differ between the conda and container toolchains. The test therefore snapshots the row count, header, and an md5 of the input-derived columns (`offset`, `profile_sum`, `profile_peak`, `length`) - byte-identical across conda/docker/singularity - rather than a full content snapshot that would fail on the conda shard.


pinin4fjords (commented):
(withdrawn)

modules/nf-core/rpbp/estimatemetagenebayesfactors/main.nf
```diff
@@ -0,0 +1,42 @@
...
+    tuple val("${task.process}"), val('rpbp'), eval('python -c "import rpbp; print(rpbp.__version__)"'), emit: versions_rpbp, topic: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "${meta.id}.metagene-bayes"
+    """
+    RPBP_MODELS_BASE=\$(python3 -c "import os, inspect, rpbp; print(os.path.join(os.path.dirname(inspect.getfile(rpbp)), 'models'))")
```
FriederikeHanssen, line 24: I mean I trust, but this looks a bit unhinged.

FriederikeHanssen (approved):

modules/nf-core/rpbp/estimatemetagenebayesfactors/main.nf
pinin4fjords: Yeah, I know it's a bit mad looking, but it was the best I could come up with to work around the upstream tool's internals. They wrote it as a scrappy workflow-like thing, I'm implementing as a more formal workflow so it's nicer, but there are some rough edges like that.

It asks the installed rpbp package where it lives and grabs the models/*.stan bundled inside it, rather than taking the models as path inputs. I did try making them inputs, but had to revert. rpbp loads each model with CmdStanModel(exe_file=...) - it needs the pre-compiled CmdStan binary sitting next to the .stan, which only exists inside the package (compiled at container-build time). A user-staged bare .stan has no binary beside it, so the "override" input couldn't actually work.

So the modules always use the bundled models, which is correct: they're version-locked to the pinned container, and it's what upstream rpbp does. Asking the package where it lives is more robust than hardcoding a path that'd break on any layout change.

PR #11957, opened 2026-06-10
feat(rpbp/selectperiodicoffsets): add module

pinin4fjords:
Part of a series splitting #11695 (closed at reviewer request: one PR per module).

Adds `rpbp/selectperiodicoffsets`, which selects the canonical P-site offset for each read length that passed the periodicity threshold. This is step 3 of the 7-step per-sample rpbp chain.

### Series

| Order | Module / subworkflow | PR |
|-------|----------------------|----|
| genome prep | `rpbp/preparegenome` | #11954 |
| 1 | `rpbp/extractmetageneprofiles` | #11955 |
| 2 | `rpbp/estimatemetagenebayesfactors` | #11956 |
| 3 | `rpbp/selectperiodicoffsets` | #11957 |
| 4 | `rpbp/getperiodiclengthsoffsets` | #11958 |
| 5 | `rpbp/extractorfprofiles` | #11959 |
| 6 | `rpbp/estimateorfbayesfactors` | #11960 |
| 7 | `rpbp/selectfinalpredictionset` | #11961 |
| — | `fasta_gtf_bam_rpbp` subworkflow | #11962 |

PR #11960, opened 2026-06-10
feat(rpbp/estimateorfbayesfactors): add module

pinin4fjords:

Adds `rpbp/estimateorfbayesfactors`, which scores each candidate ORF for whether it is actively translated vs untranslated using Stan models bundled in the rpbp Python package. This is step 6 of the 7-step per-sample rpbp chain.

### Series

| Order | Module / subworkflow | PR | Status |
|-------|----------------------|----|--------|
| genome prep | `rpbp/preparegenome` | #11954 | ✅ merged |
| 1 | `rpbp/extractmetageneprofiles` | #11955 | ✅ merged |
| 2 | `rpbp/estimatemetagenebayesfactors` | #11956 | open |
| 3 | `rpbp/selectperiodicoffsets` | #11957 | ✅ merged |
| 4 | `rpbp/getperiodiclengthsoffsets` | #11958 | ✅ merged |
| 5 | `rpbp/extractorfprofiles` | #11959 | open |
| 6 | `rpbp/estimateorfbayesfactors` | #11960 | open |
| 7 | `rpbp/selectfinalpredictionset` | #11961 | ✅ merged |
| — | `fasta_gtf_bam_rpbp` subworkflow | #11962 | open |


modules/nf-core/rpbp/estimateorfbayesfactors/main.nf
```diff
@@ -0,0 +1,44 @@
...
+    tuple val("${task.process}"), val('rpbp'), eval('python -c "import rpbp; print(rpbp.__version__)"'), emit: versions_rpbp, topic: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "${meta.id}.bayes-factors"
+    """
+    RPBP_MODELS_BASE=\$(python3 -c "import os, inspect, rpbp; print(os.path.join(os.path.dirname(inspect.getfile(rpbp)), 'models'))")
```
FriederikeHanssen, line 25: same here.

FriederikeHanssen (approved):

modules/nf-core/rpbp/estimateorfbayesfactors/main.nf
pinin4fjords: Yeah, I know it's a bit mad looking, but it was the best I could come up with to work around the upstream tool's internals. They wrote it as a scrappy workflow-like thing, I'm implementing as a more formal workflow so it's nicer, but there are some rough edges like that.

It asks the installed rpbp package where it lives and grabs the models/*.stan bundled inside it, rather than taking the models as path inputs. I did try making them inputs, but had to revert. rpbp loads each model with CmdStanModel(exe_file=...) — it needs the pre-compiled CmdStan binary sitting next to the .stan, which only exists inside the package (compiled at container-build time). A user-staged bare .stan has no binary beside it, so the "override" input couldn't actually work.

So the modules always use the bundled models, which is correct: they're version-locked to the pinned container, and it's what upstream rpbp does. Asking the package where it lives is more robust than hardcoding a path that'd break on any layout change.

PR #11990, opened 2026-06-12
Fix centrifuger output naming to avoid filename collisions in taxprof…

haris18s:
Updates the output filenames of centrifuger/centrifuger and centrifuger/quantification to be distinct, so they don't collide when used in the same pipeline (e.g., nf-core/taxprofiler).

Changes:
centrifuger/centrifuger: output pattern changed from *.tsv to *.classification.tsv (script and stub updated accordingly)
centrifuger/quantification: Output pattern changed from hardcoded ${meta.id}.tsv to glob *.report.tsv so custom ext.prefix works correctly.

Both modules previously emitted ${prefix}.tsv, causing filename collisions when both were used together in a pipeline (e.g., centrifuger classification + quantification in nf-core/taxprofiler).

### DIFF
--- a/modules/nf-core/centrifuger/centrifuger/meta.yml
+++ b/modules/nf-core/centrifuger/centrifuger/meta.yml
@@ -60,7 +60,7 @@ output:
           type: file
           description: |
             File containing classification results
-          pattern: "${prefix}.tsv"
+          pattern: "*.tsv"
           ontologies:
             - edam: http://edamontology.org/format_3475 #TSV
   fastq_classified:
--- a/modules/nf-core/centrifuger/quantification/main.nf
+++ b/modules/nf-core/centrifuger/quantification/main.nf
@@ -15,7 +15,7 @@ process CENTRIFUGER_QUANTIFICATION {
     path size_table
 
     output:
-    tuple val(meta), path("${meta.id}.tsv"), emit: report_file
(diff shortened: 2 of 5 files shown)


jfy133:
Actually this should not be necessary here - we can fix this in the pipelien with `modules.conf` :) (i.e. add `.classification`. via the `ext.prefix` bit :)

jfy133:
(e.g. maybe a differnet pipelline dev doesn't want `.classificaiton` in the name, but something else)

haris18s:
> Actually this should not be necessary here - we can fix this in the pipelien with `modules.conf` :) (i.e. add `.classification`. via the `ext.prefix` bit :)

ok thanks for letting me know, i thought would be more clear to have it in module-level,as centrfiuge has different names. 
I reverted the "classification" / "report" suffix cchanges, but i replaced meta.id with *tsv. for flexibility

SPPearce:
> > Actually this should not be necessary here - we can fix this in the pipelien with `modules.conf` :) (i.e. add `.classification`. via the `ext.prefix` bit :)
> 
> ok thanks for letting me know, i thought would be more clear to have it in module-level,as centrfiuge has different names. I reverted the "classification" / "report" suffix cchanges, but i replaced meta.id with *tsv. for flexibility

Personally I am happy to do this here, if there is a likely collision in a later module.

jfy133 (approved):

PR #12007, opened 2026-06-15
fix(pypolca): support gzipped and uncompressed assemblies

Harshita-sriv:
## Description

PyPOLCA currently assumes the assembly input is gzipped and always runs:

gzip -cdf $contigs > contigs_uncompressed

This fails when an uncompressed assembly is provided.

This PR updates the module to:
- decompress gzipped assemblies when the input ends with .gz
- copy uncompressed assemblies directly otherwise

This allows the module to support both compressed and uncompressed assembly inputs.


dialvarezs (approved):
Looks good!

PR #12010, opened 2026-06-16
feat(orftable_fasta_gtf_buildorfcatalogue): collapse small ORFs by amino-acid identity

pinin4fjords:
## Summary

Adds peptide-level deduplication to the Ribo-seq ORF catalogue and makes the catalogue build deterministic and byte-stable. Three coupled changes plus the subworkflow wiring that ties them together.

### `custom/orfcollapse` (new module)

The coordinate merge in `custom/orfmerge` only collapses ORFs that overlap on the genome, so the same micropeptide encoded at several distinct, non-overlapping loci (typically repetitive regions) survives as separate catalogue rows. This module clusters catalogue peptides by amino-acid identity (`mmseqs/easycluster`, `--min-seq-id 0.9 -c 0.8`) and folds each multi-member small-ORF (`orf_class == smORF`, aa_length ≤ 100) cluster down to one representative, unioning the cross-caller (`called_by_*` / `score_*`), cross-sample (`n_samples` / `samples`) and gene-mapping evidence. Larger and transcript-anchored ORFs pass through untouched. The representative is chosen here (longest aa_length, ties by orf_id) so the result is independent of which sequence MMseqs2 labelled the cluster representative. This adopts the peptide-level deduplication and 0.9 amino-acid-similarity threshold of the GENCODE Ribo-seq ORF consolidation (Mudge et al. 2022, Nat Biotechnol, doi:10.1038/s41587-022-01369-0; gencode-riboseqORFs collapse_cutoff 0.9), implemented here with MMseqs2 sequence-identity clustering rather than that tool's longest-shared-string / P-site-overlap metric.

### `custom/orfmerge`

`orfmerge` assigned sequential `orf_id`s in cluster-build order, which followed the nondeterministic order in which per-caller inputs were collected (`.collect()`). Catalogue content was identical run-to-run but row order / `orf_id` numbering were not, making downstream snapshots flaky. Clusters are now sorted by genomic coordinate (`chrom, start, end, strand, gene_id, transcript_id, orf_class`) before `orf_id` assignment.

### `orftable_fasta_gtf_buildorfcatalogue` (subworkflow)

Lifts the merged BED12 to an amino-acid FASTA (`bedtools/getfasta -split -s -nameOnly` then `seqkit/translate --trim`, keyed by `orf_id`), clusters it with `mmseqs/easycluster`, and feeds `custom/orfcollapse`. The collapse pass is gated by a `val_collapse` boolean take input and routed with `branch`/`mix`: when false the merged catalogue is emitted unchanged and the clustering modules do not run.

### Output naming

`custom/orfmerge` and `custom/orfcollapse` move the `catalogue` descriptor out of the hard `path(...)` outputs and into the overridable default prefix (`${meta.id}.catalogue`), so output declarations are `${prefix}.<format>`. The emitted `catalogue_aa_fasta` is `<prefix>.fasta` regardless of the collapse toggle.

## Validation

- `custom/orfcollapse` module test: collapses identical-peptide smORFs on opposite strands, unions evidence, passes through non-smORF / unique ORFs.
- `custom/orfmerge` module test and the subworkflow test (both `val_collapse` routes) pass end-to-end on x86 Linux; snapshots regenerated in the modules-repo test context.
- Test fixtures live in `nf-core/test-datasets` (`modules` branch) and are referenced via `params.modules_testdata_base_path`.

### DIFF
--- a/modules/nf-core/custom/orfcollapse/templates/orfcollapse.py
+++ b/modules/nf-core/custom/orfcollapse/templates/orfcollapse.py
@@ -192,6 +192,7 @@ def main():
                 "${task.process}": {
                     "python": platform.python_version(),
                     "pandas": pd.__version__,
+                    "pyyaml": yaml.__version__,
                 }
             },
             fh,
--- a/modules/nf-core/custom/orfcollapse/templates/orfcollapse.py
+++ b/modules/nf-core/custom/orfcollapse/templates/orfcollapse.py
@@ -6,12 +6,15 @@
 The coordinate-based merge in custom/orfmerge groups ORFs that overlap on the
 genome, but the same micropeptide is frequently encoded at several distinct,
 non-overlapping genomic loci (typically repetitive regions), and those copies
-survive as separate catalogue rows. Following the GENCODE Ribo-seq ORF catalogue
-convention (Mudge et al. 2022, Nat Biotechnol, doi:10.1038/s41587-022-01369-0;
-gencode-riboseqORFs collapse_cutoff 0.9), small ORFs (orf_class == "smORF", i.e.
-aa_length <= 100) are clustered by amino-acid sequence identity upstream
-(mmseqs/easycluster) and this module folds each multi-member cluster down to one
-representative.
+survive as separate catalogue rows. This adopts the peptide-level deduplication
+and 0.9 amino-acid-similarity threshold of the GENCODE Ribo-seq ORF
+consolidation (Mudge et al. 2022, Nat Biotechnol,
+doi:10.1038/s41587-022-01369-0; gencode-riboseqORFs collapse_cutoff 0.9),
+implemented here with MMseqs2 sequence-identity clustering (--min-seq-id 0.9)
+rather than that tool's longest-shared-string / P-site-overlap metric. Small
+ORFs (orf_class == "smORF", i.e. aa_length <= 100) are clustered by amino-acid
+identity upstream (mmseqs/easycluster) and this module folds each multi-member
+cluster down to one representative.
 
 Only smORF rows are collapsed; larger ORFs and transcript-anchored classes pass
 through untouched, preserving the deterministic coordinate/transcript merge from
--- a/modules/nf-core/custom/orfcollapse/templates/orfcollapse.py
+++ b/modules/nf-core/custom/orfcollapse/templates/orfcollapse.py
@@ -1,0 +1,201 @@
+#!/usr/bin/env python3
+# Written by Jonathan Manning (@pinin4fjords). Released under the MIT license.
+
+"""Collapse small ORFs sharing an amino-acid cluster into a single catalogue entry.
+
+The coordinate-based merge in custom/orfmerge groups ORFs that overlap on the
+genome, but the same micropeptide is frequently encoded at several distinct,
+non-overlapping genomic loci (typically repetitive regions), and those copies
+survive as separate catalogue rows. Following the GENCODE Ribo-seq ORF catalogue
+convention (Mudge et al. 2022, Nat Biotechnol, doi:10.1038/s41587-022-01369-0;
+gencode-riboseqORFs collapse_cutoff 0.9), small ORFs (orf_class == "smORF", i.e.
+aa_length <= 100) are clustered by amino-acid sequence identity upstream
+(mmseqs/easycluster) and this module folds each multi-member cluster down to one
+representative.
+
+Only smORF rows are collapsed; larger ORFs and transcript-anchored classes pass
+through untouched, preserving the deterministic coordinate/transcript merge from
+upstream. Among the smORF members of a cluster the representative is chosen here
+(longest aa_length, ties broken by orf_id) so the result is independent of which
+sequence MMseqs2 labelled the cluster representative. Catalogue row order is
+preserved; dropped members fold their cross-caller / cross-sample evidence and
+gene mappings into the survivor.
+"""
+
+import argparse
+import platform
+import re
+import shlex
+import sys
+from collections import defaultdict
+
+import pandas as pd
+import yaml
+
+CALLERS = ("ribotish", "ribocode", "ribotricer", "rpbp", "price")
+# `bedtools getfasta -nameOnly -s` appends the strand as "(+)"/"(-)" to each
+# sequence name, so FASTA headers and MMseqs2 cluster ids arrive as
+# "<orf_id>(+)". Strip it to recover the bare orf_id used in the catalogue.
+STRAND_SUFFIX = re.compile(r"\\([+-]\\)\$")
+SCORE_DIRECTIONS = {
+    "ribotish": "min",
+    "ribocode": "min",
+    "ribotricer": "max",
+    "rpbp": "max",
+    "price": "min",
+}
+CLASS_ORDER = ("canonical_cds", "uORF", "dORF", "novel_u", "smORF", "other")
+SMORF_CLASS = "smORF"
+
+
+def read_fasta(path):
+    seqs = {}
+    name, chunks = None, []
+    with open(path) as fh:
+        for line in fh:
+            line = line.rstrip("\\n")
+            if line.startswith(">"):
+                if name is not None:
+                    seqs[name] = "".join(chunks)
+                header = line[1:].split()[0] if len(line) > 1 else ""
+                name = STRAND_SUFFIX.sub("", header)
+                chunks = []
+            else:
+                chunks.append(line)
+    if name is not None:
+        seqs[name] = "".join(chunks)
+    return seqs
+
+
+def read_clusters(path):
+    """Map member orf_id -> cluster key (representative id) from an mmseqs cluster TSV."""
+    cluster_of = {}
+    with open(path) as fh:
+        for line in fh:
+            fields = line.rstrip("\\n").split("\\t")
+            if len(fields) >= 2:
+                member = STRAND_SUFFIX.sub("", fields[1])
+                rep = STRAND_SUFFIX.sub("", fields[0])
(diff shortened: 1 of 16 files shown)


subworkflows/nf-core/orftable_fasta_gtf_buildorfcatalogue/nextflow.config
```diff
@@ -4,7 +4,10 @@ process {
         ext.args   = '-split -s -nameOnly'
     }
     withName: 'ORFTABLE_FASTA_GTF_BUILDORFCATALOGUE:SEQKIT_TRANSLATE' {
-        ext.prefix = { "${meta.id}.catalogue.aa" }
+        ext.prefix = { "${meta.id}.catalogue" }
         ext.args   = '--trim'
     }
+    withName: 'ORFTABLE_FASTA_GTF_BUILDORFCATALOGUE:MMSEQS_EASYCLUSTER' {
+        ext.args = '--min-seq-id 0.9 -c 0.8 --cov-mode 0'
+    }
```
luisas, line 12: They are a bit buried here to be overwriting the defaults, should they be maybe params? Not strongly opinionated on this one, just flagging

modules/nf-core/custom/orfmerge/main.nf
```diff
@@ -11,27 +11,27 @@ process CUSTOM_ORFMERGE {
...
+    tuple val(meta), path("${prefix}.orf_to_gene.tsv") , emit: orf_to_gene_tsv
+    tuple val(meta), path("${prefix}.mqc.tsv")         , emit: multiqc
+    path "versions.yml"                                , emit: versions, topic: versions
 
     when:
     task.ext.when == null || task.ext.when
 
     script:
-    prefix = task.ext.prefix ?: "${meta.id}"
+    prefix = task.ext.prefix ?: "${meta.id}.catalogue"
```
luisas, line 24: **Severity: medium — breaking change (default output filename)**

The default prefix changes from `${meta.id}` to `${meta.id}.catalogue`. For `bed12`, `tsv`, and `mqc.tsv` the on-disk name is unchanged because `.catalogue.` was already embedded in the extension pattern. `orf_to_gene.tsv` is the odd one out: it had no `.catalogue.` infix before, so it silently renames from `{id}.orf_to_gene.tsv` → `{id}.catalogue.orf_to_gene.tsv`. Any pipeline consuming this module with the default prefix (e.g. a `publishDir` glob or a downstream process that references the file by name) will break. Worth flagging explicitly in the PR description as a breaking change.

subworkflows/nf-core/orftable_fasta_gtf_buildorfcatalogue/nextflow.config
```diff
@@ -4,7 +4,10 @@ process {
         ext.args   = '-split -s -nameOnly'
     }
     withName: 'ORFTABLE_FASTA_GTF_BUILDORFCATALOGUE:SEQKIT_TRANSLATE' {
-        ext.prefix = { "${meta.id}.catalogue.aa" }
+        ext.prefix = { "${meta.id}.catalogue" }
```
luisas, line 7: **Severity: medium — breaking change (subworkflow emitted filename)**

`ext.prefix` for `SEQKIT_TRANSLATE` changes from `${meta.id}.catalogue.aa` to `${meta.id}.catalogue`, renaming the emitted `catalogue_aa_fasta` file from `{id}.catalogue.aa.fasta` → `{id}.catalogue.fasta`. The rename is consistent with how `custom/orfcollapse` names its AA FASTA output (both paths now emit `.fasta`), but it is still a breaking change for any downstream pipeline that references the file by name or has a `publishDir` pattern like `*.aa.fasta`.

modules/nf-core/custom/orfcollapse/templates/orfcollapse.py
```diff
@@ -0,0 +1,204 @@
...
+        )
+        for cls in CLASS_ORDER:
+            mh.write(f"{cls}\\t{int(counts.get(cls, 0))}\\n")
+
+    with open("versions.yml", "w") as fh:
+        yaml.safe_dump(
+            {
+                "${task.process}": {
+                    "python": platform.python_version(),
+                    "pandas": pd.__version__,
```
luisas, line 194: **Severity: low — pyyaml version not recorded in versions.yml**

`environment.yml` pins `pyyaml=6.0.2` and the script imports `yaml`, but only `python` and `pandas` appear in the `versions.yml` output. pyyaml exposes `yaml.__version__`, so it can be included alongside pandas:
```suggestion
                    "pandas": pd.__version__,
                    "pyyaml": yaml.__version__,
```

modules/nf-core/custom/orfcollapse/tests/main.nf.test
```diff
@@ -0,0 +1,61 @@
...
+                // The dropped member's gene mapping is folded onto the survivor.
+                { assert o2g.any { it == 'orf_00000001\tg5\tt5' } },
+                // AA FASTA loses the dropped member and is keyed by bare orf_id.
+                { assert aa.contains('>orf_00000001') },
+                { assert !aa.contains('>orf_00000002') }
+            )
+        }
+    }
+
+}
```
luisas, line 89: **Severity: low — stub test missing**

The module defines a `stub` block but there is no corresponding nf-test stub test. nf-core convention requires a stub test for new modules. A minimal addition before the closing brace:

```groovy
    test("stub") {
        options "-stub"
        when {
            process {
                """
                def dir = params.modules_testdata_base_path + "genomics/homo_sapiens/riboseq_expression/orf_catalogue"
                input[0] = [
                    [ id: 'cohort' ],
                    file("\${dir}/cohort.catalogue.bed12",    checkIfExists: true),
                    file("\${dir}/cohort.catalogue.tsv",      checkIfExists: true),
                    file("\${dir}/cohort.orf_to_gene.tsv",    checkIfExists: true),
                    file("\${dir}/cohort.catalogue.aa.fasta", checkIfExists: true),
                    file("\${dir}/cohort_cluster.tsv",        checkIfExists: true)
                ]
                """
            }
        }
        then {
            assertAll(
                { assert process.success },
            )
        }
    }
```

luisas (commented):
AI-assisted review (Claude, on behalf of @luisa.santus)

modules/nf-core/custom/orfmerge/main.nf
luisas: Human comment: not really medium severity but i think it can be added in the PR comment so Claudio is happy too :)

subworkflows/nf-core/orftable_fasta_gtf_buildorfcatalogue/nextflow.config
luisas: Same as above

modules/nf-core/custom/orfcollapse/tests/main.nf.test
luisas: The stub test would be nice

subworkflows/nf-core/orftable_fasta_gtf_buildorfcatalogue/main.nf
```diff
@@ -56,16 +63,45 @@ workflow ORFTABLE_FASTA_GTF_BUILDORFCATALOGUE {
...
+    //    emits the merged catalogue untouched.
+    ch_routed = CUSTOM_ORFMERGE.out.bed12
+        .join(CUSTOM_ORFMERGE.out.catalogue_tsv)
+        .join(CUSTOM_ORFMERGE.out.orf_to_gene_tsv)
+        .join(CUSTOM_ORFMERGE.out.multiqc)
+        .join(SEQKIT_TRANSLATE.out.fastx)
+        .branch { _meta, _bed, _tsv, _o2g, _mqc, _aa ->
+            collapse: val_collapse
+            keep    : true
+        }
```
luisas, line 89: nice

modules/nf-core/custom/orfcollapse/tests/main.nf.test
```diff
@@ -0,0 +1,61 @@
...
+            assertAll(
+                { assert process.success },
+                { assert snapshot(sanitizeOutput(process.out)).match() },
+                // The two identical smORFs on different strands/loci collapse to one row.
+                { assert rows.size() == 3 },
+                { assert byId.containsKey('orf_00000001') },
+                { assert !byId.containsKey('orf_00000002') },
+                // The survivor unions both callers' evidence and both samples.
+                { assert byId['orf_00000001'][cols.indexOf('called_by_ribotish')] == '1' },
+                { assert byId['orf_00000001'][cols.indexOf('called_by_ribocode')] == '1' },
```
luisas, line 46: naive question, why having all these manual checks instead of just a snapshot check?

luisas (approved):
LGTM! I left few comments/questions but leaving on approved

subworkflows/nf-core/orftable_fasta_gtf_buildorfcatalogue/nextflow.config
pinin4fjords: Could possibly be subworkflow inputs, I'll evaluate

modules/nf-core/custom/orfmerge/main.nf
pinin4fjords: There are no consuming pipelines yet (I'm writing it), so we're good :-)

modules/nf-core/custom/orfcollapse/tests/main.nf.test
pinin4fjords: Some folks object to this because yes, the snapshot captures it. But you can blindly re-generate a snapshot and not notice something has broken- I like having assertions with logical meaning alongside.

PR #12011, opened 2026-06-16
fix(custom/orfmerge): deterministic orf_id ordering

pinin4fjords:
## Summary

`custom/orfmerge` assigned sequential `orf_id`s in the order clusters were built, which follows the order in which per-caller normalised inputs were collected by the calling workflow. Nextflow `.collect()` does not guarantee a stable order, so although the catalogue **content** was identical run-to-run, the row order and `orf_id` numbering were not - making downstream snapshots flaky.

This sorts clusters by genomic coordinate (`chrom, start, end, strand, gene_id, transcript_id, orf_class`) before assigning `orf_id`s, so the merged catalogue is byte-deterministic.

## Validation
Verified against two independent pipeline runs: pre-fix the catalogue differed only by row order / `orf_id` (content identical when sorted); the sort makes it byte-identical.

## Draft - pending
- [ ] Regenerate the `custom/orfmerge` test snapshot (orf_id ordering changes).
- [ ] (test data migration tracked with the catalogue subworkflow PR)

### DIFF
--- a/modules/nf-core/custom/orfmerge/templates/orfmerge.py
+++ b/modules/nf-core/custom/orfmerge/templates/orfmerge.py
@@ -197,6 +197,22 @@ def write_catalogue(prefix, clusters, bed_index):
 
     per_class_counts = defaultdict(int)
 
+    # Stable genomic ordering so orf_id assignment is deterministic regardless of
+    # the order in which per-caller inputs were collected upstream.
+    def _sort_key(cluster):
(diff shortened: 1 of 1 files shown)


pinin4fjords:
Folded into #12010 - custom/orfmerge is a component of the orftable_fasta_gtf_buildorfcatalogue subworkflow being added there, so the determinism fix belongs in the same PR.

PR #12029, opened 2026-06-17
fix(pypolca): normalize version string

Harshita-sriv:
PyPOLCA reports its version as:

pypolca, version 0.4.0

Normalize the reported version to emit only the version number:

0.4.0

This follows the version reporting style used by other nf-core modules and avoids including extra text in the versions channel output.

Updates the module test snapshot accordingly.


dialvarezs (approved):
LGTM!

PR #12039, opened 2026-06-18
refactor(pypolca): use assembly metadata as primary input

Harshita-sriv:
Make assembly metadata the primary metadata object for pypolca/run
Swap input order to contigs then reads
Preserve assembly metadata in outputs without relying on secondary metadata fields
Update module tests and metadata specification accordingly

### DIFF
--- a/modules/nf-core/pypolca/run/main.nf
+++ b/modules/nf-core/pypolca/run/main.nf
@@ -8,8 +8,8 @@ process PYPOLCA_RUN {
         'quay.io/biocontainers/pypolca:0.4.0--pyhdfd78af_0' }"
 
     input:
-    tuple val(meta) , path(reads)
-    tuple val(meta2), path(contigs)
+    tuple val(meta) , path(contigs)
+    tuple val(meta2), path(reads)
 
     output:
     tuple val(meta), path("${prefix}/*_corrected.fasta"), emit: polished
--- a/modules/nf-core/pypolca/run/main.nf
+++ b/modules/nf-core/pypolca/run/main.nf
@@ -8,8 +8,7 @@ process PYPOLCA_RUN {
         'quay.io/biocontainers/pypolca:0.4.0--pyhdfd78af_0' }"
 
     input:
-    tuple val(meta) , path(contigs)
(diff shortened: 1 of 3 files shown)


modules/nf-core/pypolca/run/main.nf
```diff
@@ -8,8 +8,8 @@ process PYPOLCA_RUN {
         'quay.io/biocontainers/pypolca:0.4.0--pyhdfd78af_0' }"
 
     input:
-    tuple val(meta) , path(reads)
-    tuple val(meta2), path(contigs)
+    tuple val(meta) , path(contigs)
+    tuple val(meta2), path(reads)
```
mashehu: why this swap?
This change is a bit difficult to track during a module update, so should be avoided if possible

Harshita-sriv: I made this change following feedback from the MAG PR review. The issue there was that the module emits the metadata from the reads input, which meant the assembly metadata (e.g. assembler information) was lost downstream and required additional joining logic that could become problematic when multiple assemblers are enabled.

My thinking was that since the primary output is a polished assembly, making the assembly metadata the primary metadata object would allow it to propagate naturally through the workflow.

dialvarezs: Hi @mashehu! I was the one suggesting this change. I agree that this could lead to problematic silent behavior change, but personally I think that this is how this module should be designed in the first place (it's an assembly polishing tool, so the primary meta should be the assembly one).

And alternative would be merging the inputs, i.e. [meta, contigs, reads]. That should get rid of the issue we have (assembler metadata lost) and intentionally breaks how the module works, so it doesn't go unnoticed on a module update. If you prefer that, we can go that route.

dialvarezs: Let's go with option B @Harshita-sriv: single input `[meta, contigs, reads]`.

Harshita-sriv: @dialvarezs is this fine now!

dialvarezs (approved):
Thanks @Harshita-sriv!

PR #12105, opened 2026-06-19
`jellyfish/count`: add support for gzipped input and test

nschan:
Allow gzipped input, and test if it works.

### DIFF
--- a/modules/nf-core/jellyfish/count/main.nf
+++ b/modules/nf-core/jellyfish/count/main.nf
@@ -20,17 +20,13 @@ process JELLYFISH_COUNT {
     task.ext.when == null || task.ext.when
 
     script:
+    def is_compressed = fasta.getName().endsWith(".gz") ? true : false
+    def fasta_name    = fasta.getName().replace(".gz", "")
     def args = task.ext.args ?: ''
     prefix = task.ext.prefix ?: "${meta.id}"
     """
-    if [[ ${fasta} == *.gz ]]; then
-        zcat ${fasta} > ${fasta.baseName}.fasta
-    fi
-    if [[ ${fasta} == *.fa ]]; then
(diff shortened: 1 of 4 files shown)


modules/nf-core/jellyfish/count/main.nf
```diff
@@ -23,14 +23,23 @@ process JELLYFISH_COUNT {
...
     """
+    if [[ ${fasta} == *.gz ]]; then
+        zcat ${fasta} > ${fasta.baseName}.fasta
+    fi
+    if [[ ${fasta} == *.fa ]]; then
+        cp ${fasta} ${fasta.baseName}.fasta
+    fi
+    if [[ ${fasta} == *.fastq ]]; then
+        cp ${fasta} ${fasta.baseName}.fasta
+    fi
```
vagkaratzas: cant these be merged into one OR clause? they do the same things

vagkaratzas: Can't something like this (used in a lot of nf-core modules) work? 
````
def is_compressed = fasta.getName().endsWith(".gz") ? true : false
def fasta_name    = fasta.getName().replace(".gz", "")
"""
    gzip -c -d ${fasta} > ${fasta_name}
fi
````

And then use ${fasta_name} in the command?
Is there a specfic reason fastq is renamed to fasta?

vagkaratzas (changes requested):
Just a thing to decide on the conditions, but other than that, ready to shoot

modules/nf-core/jellyfish/count/main.nf
nschan: This is much cleaner 👍

vagkaratzas (approved):
LGTM!

PR #12114, opened 2026-06-19
Add `round_digits` to `decoupler` and avoid sorting by `topTable` in `dream`

atrigila:
`decoupler` was not using the `round_digits` feature that we have in nf-core/differentialabundance. The changes here fix that. Additionally, in `dream`, we do not rely on topTable to automatically sort rows, which was causing also some instability. 
In the dream test there is also a removal of an obsolete test. 

For: https://github.com/nf-core/differentialabundance/pull/745


dialvarezs (approved):
Looks good!

PR #12123, opened 2026-06-22
Add hmftools/amber module

### LINKED ISSUE #12121: new module: wisecondorx/convert

### Further Information

https://github.com/CenterForMedicalGeneticsGhent/WisecondorX

```yaml
bioconda::wisecondorx
```

JP-Bro:
Closes #12121

Add module for AMBER (Allele Balance for Major chromosomal Events 
via Regression), part of the Hartwig Medical Foundation hmftools suite.

AMBER generates a tumor BAF file required for downstream tumor purity 
and ploidy estimation with PURPLE.

- main.nf
- meta.yml  
- environment.yml
- tests/main.nf.test (stub test)

### DIFF
--- a/modules/nf-core/hmftools/amber/environment.yml
+++ b/modules/nf-core/hmftools/amber/environment.yml
@@ -1,0 +1,5 @@
+channels:
+  - conda-forge
+  - bioconda
+dependencies:
+  - bioconda::hmftools-amber=4.2
--- a/modules/nf-core/hmftools/amber/main.nf
+++ b/modules/nf-core/hmftools/amber/main.nf
@@ -1,0 +1,60 @@
+process HMFTOOLS_AMBER {
+    tag "$meta.id"
+    label 'process_medium'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine in ['singularity', 'apptainer'] && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/hmftools-amber:4.2--hdfd78af_0' :
+        'biocontainers/hmftools-amber:4.2--hdfd78af_0' }"
+
+    input:
+    tuple val(meta), path(tumor_bam), path(tumor_bai), path(normal_bam), path(normal_bai)
+    path fasta
+    path fasta_fai
+    path loci
+
+    output:
+    tuple val(meta), path("${prefix}/"), emit: amber_dir
+    path "versions.yml",                emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "${meta.id}"
+    def tumor_id  = meta.tumor_id  ?: "${meta.id}_tumor"
+    def normal_id = meta.normal_id ?: "${meta.id}_normal"
+    """
+    amber \\
+        -tumor ${tumor_id} \\
+        -tumor_bam ${tumor_bam} \\
+        -reference ${normal_id} \\
+        -reference_bam ${normal_bam} \\
+        -output_dir ${prefix} \\
+        -loci ${loci} \\
+        -ref_genome ${fasta} \\
+        -threads ${task.cpus} \\
+        ${args}
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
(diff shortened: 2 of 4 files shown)


awgymer:
Given that there exists [implementations](https://github.com/nf-core/oncoanalyser/blob/master/modules/local/amber/main.nf) of the `hmftools` in `nf-core/oncoanalyser` which are maintained by people working at HMF I have to wonder:

- is there a good reason these have not been ported to `nf-core/modules` already?
- should the implementation in `nf-core/modules` aim to match those implementations as closely as possible

I wonder if @scwatts might be able to add some context here.

SPPearce:
> Given that there exists [implementations](https://github.com/nf-core/oncoanalyser/blob/master/modules/local/amber/main.nf) of the `hmftools` in `nf-core/oncoanalyser` which are maintained by people working at HMF I have to wonder:
> 
> * is there a good reason these have not been ported to `nf-core/modules` already?
> * should the implementation in `nf-core/modules` aim to match those implementations as closely as possible
> 
> I wonder if @scwatts might be able to add some context here.

The oncoanalyser people don't seem to be particularly interested in putting the modules into here.

scwatts:
Had a chat with the Hartwig folks on the topic this morning

Not having the tools in `nf-core/modules` is more a product of the hmftools / wigits developers wanting to discourage unsupported use of their tooling due to potential for misconfiguration or misuse

This is a concern that they see being addressed through standardisation with oncoanalyser and contributing recommended configuration / parameterisation / etc, and their perspective is that these tools should only be run via oncoanalyser

SPPearce:
> Had a chat with the Hartwig folks on the topic this morning
> 
> Not having the tools in `nf-core/modules` is more a product of the hmftools / wigits developers wanting to discourage unsupported use of their tooling due to potential for misconfiguration or misuse
> 
> This is a concern that they see being addressed through standardisation with oncoanalyser and contributing recommended configuration / parameterisation / etc, and their perspective is that these tools should only be run via oncoanalyser

We don't ask permission from the tool authors for tools to be added to nf-core/modules, anyone can do it. Ideally it'd be someone who actually knows the tool well, but if not then anyone can do it.

ohofmann:
> We don't ask permission from the tool authors for tools to be added to nf-core/modules, anyone can do it. Ideally it'd be someone who actually knows the tool well, but if not then anyone can do it.

Maybe to provide some additional context here, and noting that I cannot speak for the methods developers. One of the main source of issues over at https://github.com/hartwigmedical/hmftools stem from users trying to get the various methods to run with different inputs -- slightly different parameters than expected, with VCFs from variant callers we've never seen outputs from before, or with unexpected reference data. Even if we can get, say, PURPLE to work with a VCF that is missing some key field the results are then untested and might result in a degradation of quality without anyone being aware of it. 

_One_ of the motivations for nf-core/oncoanalyser was to alleviate that problem. To have a well-defined way for these tools to work together, to be able to (reliably) leverage information from upstream methods. That has allowed the (surprisingly small!) team to expand in the other direction -- to support not just WGS T/N but panel-based methods, or MRD, and so on. 

So I don't think the concern here is not whether the tools are available to others, at all. They are open source and meant to be used (and re-mixed!); it's just that it pushes the support burden upstream to the methods developers, and maybe worse, running it outside of the defined sequence of methods can result in user frustration, potentially poor results, and a belief that some of these just don't work.

The point of users the just grabbing the code is well taken, though, and something worth considering.

awgymer:
> Even if we can get, say, PURPLE to work with a VCF that is missing some key field the results are then untested and might result in a degradation of quality without anyone being aware of it.

This was something I did in a previous role (actually we converted our VCFs first to look close to expected, but of course they are still not truly the same thing). 

I think the reality is that sort of adaptation of tools to formats outside the prior tested ones is **very** common in bioinformatics and there's not much that can be done about it. 

> pushes the support burden upstream to the methods developers, and maybe worse, running it outside of the defined sequence of methods can result in user frustration, potentially poor results, and a belief that some of these just don't work

I think this is a very fair concern but really it's an issue of users not behaving in the right way and understanding the implied contracts between developers and users. 

I do actually also think that providing the tools within nf-core as modules might help reduce the burden in some ways too, because hopefully people who have an issue with running the tool via nextflow often come to `nf-core` first (via Slack or GitHub) and the community is quite good at filtering what should be referred upstream and what can be solved here. If users implement the tools themselves in nextflow then they are more likely to just go to the original tool GitHub I think.

PR #11066, opened 2026-03-27
Update cellranger multi module to comply to nf-core guidelines

mahesh-panchal:
## Description

- Removes `task.ext.force_include_frna` :- Part of #8716 - Revert #10922 
- Refactors python script to Groovy/Bash since the python script wraps a command-line call.
- Renames meta_* to meta\d? to conform to nf-core guidelines
- Updates topic emission to use `eval`
- Makes a copy of fb_reference to fix bug in https://github.com/nf-core/scrnaseq/issues/529#issuecomment-4126097217
- Removes empty file from tests and uses `[]` for optional files.
- Removes missing file from test.
- Updates test config to conform to nf-core guidelines.

## TODO

- Refactor out meta?.options to conform to nf-core guidelines
- Check file staging logic. Currently matches test input. 

## PR checklist

Closes #XXX 

- [x] This comment contains a description of changes (with reason).
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [ ] Remove all TODO statements.
- [ ] Broadcast software version numbers to `topic: versions` - [See version_topics](https://nf-co.re/blog/2025/version_topics)
- [ ] Follow the naming conventions.
- [ ] Follow the parameters requirements.
- [ ] Follow the input/output options guidelines.
- [ ] Add a resource `label`
- [ ] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - For modules:
    - [ ] `nf-core modules test <MODULE> --profile docker`
    - [ ] `nf-core modules test <MODULE> --profile singularity`
    - [ ] `nf-core modules test <MODULE> --profile conda`
  - For subworkflows:
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile docker`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile singularity`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile conda`

PR #12152, opened 2026-06-23
Fix: Update Cellranger/multi to reinclude functionality from previous PR update

mahesh-panchal:
## Description

Related to #11066 . Adds back in functionality from the python template that was accidentally dropped.

- Adds options for all input types
- Explicitly tests for certain keys.

## PR checklist

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


mahesh-panchal:
@leandrotiburske @fmalmeida The PR is ready and I think I have the flags in, but please check. From today I won't be near a computer to make updates so I made this branch on nf-core modules where you can both push/pull your own changes onto this branch. 

The tests on this module look like they really need updating as well, but that's perhaps outside the scope. I leave this up to you both now.

mahesh-panchal:
This PR should remove the old python template too since it's redundant.

fmalmeida:
Hi @leandrotiburske,

Do you think you can take over from here to do the final amendments?

As @mahesh-panchal mentioned, the try is to move all the logic into the module itself.

So, could you double check if all the functionality from before is now there so to remove the template script?

As Mahesh, I am also not fluent in this tool, I had only helped out on the code but someone else was assessing the results ... *so I believe it would be really good if someone that knows how the results should be could take a closer look*.

I am on holidays until next Wednesday and I only have a phone, so I cannot help with the code until there but I can help with comments and review.

fmalmeida:
@nictru ;

You think you can help with this one? Not sure how much understanding of the data itself and how it should you have, but if you can, it would be lovely to have cellrangermulti fixed to bring it back to scrnaseq updated :)

It relates to #11066

modules/nf-core/cellranger/multi/main.nf
```diff
@@ -66,28 +66,40 @@ process CELLRANGER_MULTI {
...
+        // GEX options forwarded from the gex_options input map
+        ['filter-probes', 'r1-length', 'r2-length', 'chemistry', 'expect-cells', 'force-cells',
+         'no-secondary', 'check-library-compatibility', 'no-target-umi-filter', 'include-introns'].each { key ->
+            if (gex_options?.containsKey(key)) gex_section << "${key},${gex_options[key]}"
         }
 
-        def create_bam = gex_has_opts.containsKey("create-bam") ? gex_has_opts["create-bam"] : "true"
-        gex_section << "create-bam,${create_bam}"
+        // create-bam defaults to true if not specified
+        gex_section << "create-bam,${gex_options?.get('create-bam') ?: 'true'}"
```
nictru, line 76: If `gex_options?.get('create-bam')` returns boolean `false`, then string `true` will be used here. Is this something that could become a real problem? Or can one be sure that `gex_options?.get('create-bam')` will always return a string `'false'`?

modules/nf-core/cellranger/multi/main.nf
```diff
@@ -129,25 +143,35 @@ process CELLRANGER_MULTI {
...
-            r2=\$(find "\${r2_dir}" -maxdepth 1 -name "*_R2_*.fastq.gz" | head -1)
-
-            [ -z "\${r1}" ] || [ -z "\${r2}" ] && continue
-
-            ln -sf "\$(readlink -f "\${r1}")" "fastq_all/\${modality}/${prefix}_S1_L\$(printf %03d \${lane})_R1_001.fastq.gz"
-            ln -sf "\$(readlink -f "\${r2}")" "fastq_all/\${modality}/${prefix}_S1_L\$(printf %03d \${lane})_R2_001.fastq.gz"
+            if [ "${skip_renaming}" = "true" ]; then
+                r1=\$(find "\${r1_dir}" -maxdepth 1 -name "*.fastq.gz" | head -1)
+                r2=\$(find "\${r2_dir}" -maxdepth 1 -name "*.fastq.gz" | head -1)
+                [ -z "\${r1}" ] || [ -z "\${r2}" ] && continue
```
nictru, line 170: From what I can see the old version raised an AssertionError on mismatch, here we skip silently

nictru (commented):
Some minor comments
Also: The python script is not being removed by this PR so far

fmalmeida:
> Some minor comments
> Also: The python script is not being removed by this PR so far


@nictru

Yeah. I don't think it is ready yet. It still has things to be done, as with the script you mentioned.

Tagged you because it would be really good to have a review of someone that understands the underlying data better than me.

Once I am back from holidays I can give it a check to try to finish it if not progressed until then.

modules/nf-core/cellranger/multi/main.nf
fmalmeida: Hi,
This one should be fine.
Since the `gex_section` variable, in the end, must be evalued as a string because it is written to the cellranger multi config file.
And here, the idea is that, if not specified, we create the BAM file by default.

modules/nf-core/cellranger/multi/main.nf
fmalmeida: Hi @nictru ;
Re-added in commit:
https://github.com/nf-core/modules/pull/12152/commits/00d72a0ab887785bd670136512b8f9c22b39b074

fmalmeida:
Hi @mahesh-panchal , @nictru and @leandrotiburske,

I got AI to help assess the change from the Python template to the current main.nf, and it seems were are there ...

**Assessment:**
- The PR goal was to move `cellranger/multi` from the old Python template to native main.nf while preserving the same functional behavior ... with all the old options.
- The assessment shows that the current main.nf still performs the key Python-template behaviors:
  - beam antigen panel is still written as `reference,$PWD/<beam_antigen_panel>` under `[feature]`
  - `fb_reference` is still copied to `fb_reference_copy.csv`
  - `skip_renaming` logic is preserved, with equivalent symlink and rename behavior
  - modality option forwarding for gex/vdj/ab/cmo is preserved
  - sample/antigen blocks are generated for CMO/FRNA/OCM/BEAM inputs

**So we can close:**

Can you guys help with confirming the generated config output and staged reference/sample files for all modalities, and if the goal of the PR is achieved? @leandrotiburske, I think your input here will be extra valuable.

<details>
  <summary><b>AI assessment reference</b></summary>

**Assessment — cellranger/multi: Python template vs native `main.nf`**

Summary
-------
- Files compared: `templates/cellranger_multi.py` (old Python template) and `main.nf` (current Groovy/native implementation).
- Goal: list differences, identify regressions from the Python template move, and recommend fixes.

Comparison
----------

| Item | Python template (old) | `main.nf` (new) | Status / Recommendation |
|---|---|---|---|
| versions output | Writes `versions.yml` with `cellranger` version. | Emits a `versions_cellranger` tuple via `eval(...)` and publishes it with `topic: versions`. | Intended change: output now uses topic channels instead of writing `versions.yml`. |
| Beam antigen panel handling | Reads `beam_antigen_csv` and emits `reference,$PWD/<beam_antigen_panel>` under `[feature]`. | Adds `reference,$PWD/<beam_antigen_panel>` under `[feature]`. | Behavior is consistent; no change required unless `cellranger multi` specifically needs inline CSV contents instead of a file reference. |
| R1/R2 filename validation | Strict regex-based validation ensuring R1->R2 mapping and asserts even-numbered list (raises on mismatch). | Bash symlink logic pairs `*_R1_*` and `*_R2_*` within directories and now validates that the R2 basename matches the R1 basename after `_R1_` → `_R2_` replacement. | Implemented in `main.nf`; review only if additional filename patterns or pairing rules should be supported. |
| skip_renaming behavior | Python renames or preserves filenames according to `${skip_renaming}`; also canonicalises paths for references. | `main.nf` implements `skip_renaming` symlink logic in bash, and uses `$PWD/${...name}` for references. | Behavior is preserved; path canonicalisation differs but is acceptable. |
| fb_reference copy | Python copies `fb_reference` to `fb_reference_copy.csv` to avoid symlink corruption. | `main.nf` also copies `fb_reference` to `fb_reference_copy.csv`. | Preserved. |
| Options forwarded (gex/vdj/ab/cmo) | Forwards many gex/vdj/ab/cmo options (e.g., `r1-length`, `r2-length`, `chemistry`, `min-assignment-confidence`, and `create-bam` default). | Builds the same sections and forwards the same option keys into the config lines. | Preserved. |
| Extra sample/antigen blocks | Python conditionally appends CMO/FRNA/OCM/BEAM CSV contents inline when present. | `main.nf` appends `[samples]` blocks using the `.text` of inputs and adds `[antigen-specificity]` for beam control panel. | Functionally similar for sample blocks; beam antigen CSV inline vs reference needs confirmation (see beam row). |
| Generated config formatting | Python builds config by string templating and removes blank lines. | `main.nf` builds `config_content` from joined lists and writes via heredoc. | Equivalent final result in most cases. |

Other notes
-----------
- `main.nf` adds `when: task.ext.when == null || task.ext.when` gating — a common Nextflow pattern; not present in the Python-only template but not a regression.
- `main.nf` uses `$PWD/${...name}` which yields absolute (within container) paths in the config; Python adjusted relative paths to the working directory. Both are acceptable but different.

PR #12152 verification
----------------------
- I examined PR #12152 (title: "Fix: Update Cellranger/multi to reinclude functionality from previous PR update"). The PR explicitly reintroduced per-modality options and test updates that had been lost when moving away from the Python template.
- Changes confirmed present in the branch:
	- Input tuple signatures in `main.nf` now include per-modality options: `gex_options`, `vdj_options`, `ab_options`, `beam_options`, `cmo_options`, `crispr_options`.
	- `meta.yml` updated to declare these `*_options` maps and document supported keys (e.g., `filter-probes`, `r1-length`, `r2-length`, `chemistry`, `min-assignment-confidence`, etc.).
	- Tests updated (`tests/main.nf.test`) so channels include the third options map element (e.g., `ch_gex_fastqs = [ [:], [], [:] ]`) and `.map` calls produce the extra `[:]` options map to match the new tuple shapes.
- Conclusion: PR #12152 addressed the complaint that per-modality option variables were missed. Those variables are now present in `main.nf`, `meta.yml`, and tests.

Recommendations (prioritised)
----------------------------
1. Confirm generated config output and staged reference/sample files for all modalities.

</details>

fmalmeida:
Hi @leandrotiburske and @nictru ,
Do you guys have any extra inputs? Otherwise, I would say it is ready for final check and closing.

fmalmeida (approved):
Hi @mahesh-panchal ,

I am going to approve the PR since I understand that the goal of the PR was achieved and the tests are passing.

@leandrotiburske @nictru if you guys or other people still experience issues or have something to correct, then a new focused PR can be made.

mahesh-panchal:
I fixed some tests and removed dead code in the tests. I also added the paired read guard back.

PR #12166, opened 2026-06-25
fix(quant_tximport_summarizedexperiment): deterministic tx2gene sample selection

pinin4fjords:
## Description

In `QUANT_TXIMPORT_SUMMARIZEDEXPERIMENT`, the quant results fed to `CUSTOM_TX2GENE` were picked with `.first()`:

```groovy
ch_tx2gene_quants = quant_results
    .first()
    .map { _meta, results -> [ [:], results ] }
```

tx2gene only needs one sample's quant files (it reads them to discover which GTF attribute holds the transcript IDs; all samples share a transcriptome), so picking a single sample is correct. But `.first()` selects whichever per-sample quant emission *arrives first*, and queue-channel arrival order equals task-completion order, which is non-deterministic across runs.

The result is a non-deterministic input to `CUSTOM_TX2GENE`: a different sample's `quant` directory is staged on different runs, the task hash changes, and `-resume` re-runs tx2gene even when nothing has changed. Because `TXIMETA_TXIMPORT` (and the downstream `SummarizedExperiment` builds) depend on `CUSTOM_TX2GENE.out.tx2gene`, that cache miss cascades through the whole subworkflow.

## Fix

Select the sample deterministically by sorting on the staged results name and taking one:

```groovy
ch_tx2gene_quants = quant_results
    .toSortedList { a, b -> a[1].name <=> b[1].name }
    .map { sorted -> [ [:], sorted.first()[1] ] }
```

The same sample is now chosen on every run, so the tx2gene cache key is stable and `-resume` works. Sorting keys on the results path (guaranteed by the channel contract) rather than `meta.id`, so it doesn't assume the meta map carries an `id`.

The selected sample is still arbitrary biologically (the tx2gene output is identical for any sample sharing the transcriptome), so module outputs and the test snapshot are unchanged - this only makes the selection reproducible.

## PR checklist

- [ ] Ensure the test passes locally - relying on CI; the change does not affect any output, so snapshots are unchanged.
- [x] `README.md` is updated (including new tool citations and authors/contributors).

### DIFF
--- a/subworkflows/nf-core/quant_tximport_summarizedexperiment/main.nf
+++ b/subworkflows/nf-core/quant_tximport_summarizedexperiment/main.nf
@@ -28,9 +28,14 @@ workflow QUANT_TXIMPORT_SUMMARIZEDEXPERIMENT {
     // tx2gene independently per sample. If a use case arises requiring mixed
     // transcriptomes, this will need to be revisited.
     //
+    // The sample is selected by sorting on the staged results name so the same
+    // one is chosen on every run, keeping the CUSTOM_TX2GENE cache key stable
+    // across -resume. Picking by arrival order would vary between runs and
+    // invalidate the cache for tx2gene and everything downstream of it.
+    //
     ch_tx2gene_quants = quant_results
-        .first()
-        .map { _meta, results -> [ [:], results ] }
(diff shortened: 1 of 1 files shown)


mashehu (approved):

pinin4fjords:
Thanks for quick review @mashehu !

PR #12167, opened 2026-06-25
feat(custom/orfmerge): cross-caller/sample consensus catalogue view (+ orftable subworkflow emit)

pinin4fjords:
## Changes

Adds a configurable **consensus view** to the cross-sample ORF catalogue, alongside the existing full catalogue (which is always emitted, unchanged at the defaults).

- `custom/orfmerge` and `custom/orfcollapse` both gain `--min-callers N` / `--min-samples M` (default 1) and emit `*.consensus.{bed12,tsv,orf_to_gene.tsv}` filtered to ORFs supported by at least that many distinct callers and recurring in at least that many samples.
- The filter is applied to the **final** catalogue: when the amino-acid collapse runs, the consensus is the high-confidence subset of the de-redundified catalogue (so a folded micropeptide is judged on its combined cross-caller / cross-sample evidence); when collapse is disabled, it comes straight from the merger.
- At the defaults (`1`/`1`) nothing is filtered, so the consensus view equals the full catalogue and existing behaviour is unchanged. Raising either threshold (e.g. `--min-callers 2`) yields a higher-confidence catalogue that tames downstream ORF-level multiple testing, without discarding the full union.

The filter operates on the per-ORF evidence the catalogue already records (`called_by_<caller>` indicators and the distinct-sample set), so no new computation is introduced beyond writing the filtered files.

## Subworkflow

`orftable_fasta_gtf_buildorfcatalogue` exposes the consensus outputs (`consensus_bed12` / `consensus_tsv` / `consensus_orf_to_gene_tsv`) through its emit interface, sourced from `CUSTOM_ORFCOLLAPSE` when the collapse runs (`val_collapse`) and from `CUSTOM_ORFMERGE` otherwise.

## Other

- `SEQKIT_TRANSLATE` output prefix aligned to `.catalogue.aa` (matching its `.catalogue.nt` `BEDTOOLS_GETFASTA` sibling).
- `custom/orfcollapse` docstring corrected: the smORF-only restriction and locus-agnostic global-identity clustering are this module's design choices, not properties of the GENCODE consolidation it is inspired by; `--min-seq-id 0.9` approximates rather than reproduces `collapse_cutoff 0.9`.
- Replaced the modules' parenthesized/`ExitStack` multi-file writes with plain sequential `with` blocks, valid on the repo's `py38` target.

## Tests

- `custom/orfmerge`: added a `--min-samples 2` case asserting the consensus view is a non-empty strict subset of the full catalogue.
- Module and subworkflow snapshots regenerated.

### DIFF
--- a/subworkflows/nf-core/orftable_fasta_gtf_buildorfcatalogue/main.nf
+++ b/subworkflows/nf-core/orftable_fasta_gtf_buildorfcatalogue/main.nf
@@ -105,11 +105,12 @@ workflow ORFTABLE_FASTA_GTF_BUILDORFCATALOGUE {
     catalogue_aa_fasta = CUSTOM_ORFCOLLAPSE.out.aa_fasta.mix(ch_routed.keep.map { meta, _bed, _tsv, _o2g, _mqc, aa -> [ meta, aa ] })
     multiqc            = CUSTOM_ORFCOLLAPSE.out.multiqc.mix(ch_routed.keep.map { meta, _bed, _tsv, _o2g, mqc, _aa -> [ meta, mqc ] })
 
-    // Consensus view of the merged catalogue (ORFs meeting the merger's
-    // --min-callers / --min-samples thresholds). It is a cross-caller /
-    // cross-sample evidence filter applied at merge time, so it precedes the
-    // optional amino-acid collapse and is emitted straight from the merger.
-    consensus_bed12           = CUSTOM_ORFMERGE.out.consensus_bed12
-    consensus_tsv             = CUSTOM_ORFMERGE.out.consensus_tsv
-    consensus_orf_to_gene_tsv = CUSTOM_ORFMERGE.out.consensus_orf_to_gene_tsv
+    // Consensus view: ORFs meeting the --min-callers / --min-samples thresholds.
+    // The filter is applied to the final catalogue, so when the amino-acid
+    // collapse runs the consensus is the high-confidence subset of the
+    // de-redundified catalogue (from CUSTOM_ORFCOLLAPSE); otherwise it comes
+    // straight from the merger.
+    consensus_bed12           = val_collapse ? CUSTOM_ORFCOLLAPSE.out.consensus_bed12 : CUSTOM_ORFMERGE.out.consensus_bed12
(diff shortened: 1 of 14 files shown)


subworkflows/nf-core/orftable_fasta_gtf_buildorfcatalogue/main.nf
```diff
@@ -104,4 +104,12 @@ workflow ORFTABLE_FASTA_GTF_BUILDORFCATALOGUE {
     orf_to_gene_tsv    = CUSTOM_ORFCOLLAPSE.out.orf_to_gene_tsv.mix(ch_routed.keep.map { meta, _bed, _tsv, o2g, _mqc, _aa -> [ meta, o2g ] })
     catalogue_aa_fasta = CUSTOM_ORFCOLLAPSE.out.aa_fasta.mix(ch_routed.keep.map { meta, _bed, _tsv, _o2g, _mqc, aa -> [ meta, aa ] })
     multiqc            = CUSTOM_ORFCOLLAPSE.out.multiqc.mix(ch_routed.keep.map { meta, _bed, _tsv, _o2g, mqc, _aa -> [ meta, mqc ] })
+
+    // Consensus view of the merged catalogue (ORFs meeting the merger's
+    // --min-callers / --min-samples thresholds). It is a cross-caller /
+    // cross-sample evidence filter applied at merge time, so it precedes the
+    // optional amino-acid collapse and is emitted straight from the merger.
```
suhrig: Can you explain a bit more the rationale behind skipping the amino-acid collapse for the consensus filter? To my understanding, these two filters serve different purposes: the consensus selects high-confidence calls, the amino-acid collapse removes redundancy. Wouldn't it make sense to remove this redundancy for consensus calls, too?

pinin4fjords: Good call, you're right. I've moved the filter so it runs after the collapse rather than before, so the consensus is the high-confidence subset of the de-redundified catalogue instead of the raw merge. `custom/orfcollapse` now takes the same `--min-callers`/`--min-samples` and emits the `consensus_*` outputs, and the subworkflow sources the consensus from the collapse when it runs (falling back to the merger when collapse is disabled).

Applying it post-collapse also has a nice side effect: a micropeptide folded across several loci is judged on its combined cross-caller/cross-sample evidence rather than per-copy, so recurrent smORFs aren't penalised for being split. Pushed.

suhrig (approved):

PR #12168, opened 2026-06-25
fix(quant_tximport_summarizedexperiment): guard tx2gene against empty quant_results

pinin4fjords:
## Description

Follow-up to #12166, which switched the tx2gene sample selection from `.first()` to `toSortedList(...).map { sorted -> sorted.first()[1] }` for deterministic, cache-stable selection.

That introduced a regression for the **empty** `quant_results` case. `.first()` emits *nothing* on an empty channel, so `CUSTOM_TX2GENE` simply doesn't run. `toSortedList` instead *always* emits - an empty list `[]` for an empty channel - so `sorted.first()` throws:

```
ERROR ~ Cannot access first() element from an empty List
```

This aborts any pipeline that invokes the subworkflow with an empty `quant_results` (e.g. nf-core/rnaseq, where the unused alignment-based quant path passes an empty channel when only pseudo-alignment runs). It surfaced as broad nf-core/rnaseq CI failures on the `--skip_alignment` pseudo-aligner tests.

## Fix

Drop the empty-list emission before the map:

```groovy
ch_tx2gene_quants = quant_results
    .toSortedList { a, b -> a[1].name <=> b[1].name }
    .filter { sorted -> sorted.size() > 0 }
    .map { sorted -> [ [:], sorted.first()[1] ] }
```

Empty input → `CUSTOM_TX2GENE` is skipped (matching the original `.first()` behaviour); non-empty input → deterministic selection as in #12166. No output or snapshot changes.

Verified against nf-core/rnaseq's `tests/kallisto.nf.test` (`--pseudo_aligner kallisto --skip_qc --skip_alignment`), which reproduced the abort and now passes with snapshots unchanged.

## PR checklist

- [x] Make sure your code lints (`nf-core subworkflows lint`).
- [x] Verified downstream (nf-core/rnaseq pipeline test) passes; no output/snapshot change.
- [x] `README.md` is updated (including new tool citations and authors/contributors).

### DIFF
--- a/subworkflows/nf-core/quant_tximport_summarizedexperiment/main.nf
+++ b/subworkflows/nf-core/quant_tximport_summarizedexperiment/main.nf
@@ -31,10 +31,13 @@ workflow QUANT_TXIMPORT_SUMMARIZEDEXPERIMENT {
     // The sample is selected by sorting on the staged results name so the same
     // one is chosen on every run, keeping the CUSTOM_TX2GENE cache key stable
     // across -resume. Picking by arrival order would vary between runs and
-    // invalidate the cache for tx2gene and everything downstream of it.
+    // invalidate the cache for tx2gene and everything downstream of it. The
+    // empty-list guard keeps tx2gene from running when no quant results are
+    // supplied, since toSortedList still emits an empty list in that case.
     //
     ch_tx2gene_quants = quant_results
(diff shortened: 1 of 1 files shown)


dialvarezs (approved):
Good job finding the issue!
I hate when this happens :(

PR #12173, opened 2026-06-26
fix(custom/orfnormalise): treat ribotish GenomePos start as 0-based

pinin4fjords:
## Summary

`custom/orfnormalise` mis-parses ribotish `predict` coordinates. ribotish emits **0-based half-open** `GenomePos` (and `Start`/`Stop`) coordinates, but `_parse_ribotish_genpos` subtracted 1 from the start, treating it as 1-based. This shifts every ribotish-derived ORF 5' by one base:

- **+ strand**: a full reading-frame shift. The normalised BED12 `thickStart`/blocks no longer start on the ATG, so any downstream codon-aware consumer (e.g. expanding the catalogue into in-frame codon positions for P-site quantification) walks frame 2, and the catalogue AA FASTA translates with premature in-frame stops.
- **− strand**: a 1-nt 3' overhang (ORF span no longer a whole number of codons).

### Verification

Against the chr20 reference for a worked ribotish ORF (`GenomePos=20:326098-326872:+`, `AALen=257`):

- corrected 0-based start `326098` → codon `ATG`, clean translation to 257 aa then stop at codon 258 (matches `AALen`);
- the old `-1` start `326097` → codon `CAT`, premature stop at codon 72.

After the fix, all five caller fixtures (ribocode, ribotish, ribotricer, rpbp, price) emit ORFs whose blockSizes sum is a multiple of 3, and every chr20 ORF starts on ATG with no internal stop codons. Only ribotish was affected; the other callers were already correct.

## Changes

- **Fix**: `_parse_ribotish_genpos` uses the `GenomePos` start as-is (the end was already treated as 0-based-exclusive).
- **Test data**: migrate the caller fixtures off the deleted `pinin4fjords/test-datasets@add-orf-prediction-fixtures` branch to `params.modules_testdata_base_path` (nf-core/test-datasets). The test could not run as-is (dead URLs → 404).
- **Regression guard**: each caller test now asserts every emitted ORF spans a whole number of codons (`sum(blockSizes) % 3 == 0`), which catches this class of coordinate/frame bug for all callers.
- **Snapshots**: regenerate `custom/orfnormalise` and the dependent `orftable_fasta_gtf_buildorfcatalogue` subworkflow snapshots. Only the ribotish-derived records change.

## PR checklist

- [x] Tests pass (`nf-test test modules/nf-core/custom/orfnormalise/... subworkflows/nf-core/orftable_fasta_gtf_buildorfcatalogue/...`, NXF 25.04.8, `--profile docker`).
- [x] Snapshots regenerated.


modules/nf-core/custom/orfnormalise/templates/orfnormalise.py
suhrig, line 210: Do we need to omit `- 1` here as well? I haven't confirmed empirically, but I'm assuming the `blocks` column of Ribotish is 0-based, too.

Caution: This function is also used for parsing of Ribotricer output, which may be 1-based.

suhrig (commented):
Thank you for spotting/fixing this!

The %3 assertion is a great idea!

modules/nf-core/custom/orfnormalise/templates/orfnormalise.py
pinin4fjords: Good catch, and you're right to be wary of `parse_intervals` being shared.

Checked it empirically: ribotish `predict` emits 19 columns (`Gid, Tid, Symbol, GeneType, GenomePos, StartCodon, Start, Stop, TisType, TISGroup, TISCounts, TISPvalue, RiboPvalue, RiboPStatus, FisherPvalue, TISQvalue, FrameQvalue, FisherQvalue, AALen`) and no `Blocks` column, so `parse_intervals(row.get("Blocks", ...))` never actually fired - exon structure is always recovered from the GenomePos span intersected with the transcript exons. A 4-exon ribotish ORF in a full pipeline run reconstructs and translates cleanly through that fallback after the fix, so the GenomePos correction is the whole story for ribotish. I've dropped the dead `Blocks` branch so ribotish always uses that verified path (output unchanged, snapshots still match).

And yes, `parse_intervals` stays 1-based on purpose: ribotricer is 1-based (its fixtures translate to clean ATG-start, no-internal-stop, `len % 3 == 0` ORFs through that `-1`), so removing it there would have broken ribotricer. Leaving it alone was deliberate.

suhrig (approved):

PR #12187, opened 2026-06-29
Add new component: `quarto/prerender`

fasterius:
This PR adds a new module: `quarto/partial`, which aims to add new functionality for Quarto within nf-core. The `quartonotebook` already exists, which can render Quarto/Jupyter/RMarkdown reports into HTML, but it only creates a single, monolithic report and can't be extended. Quarto has various ways of working with sub-reports or extending reports, which is what this new module is all about. The major use case is allowing for more granular reporting flow, dividing work up in arbitrary steps to reduce re-computation (e.g. when iterating over some analysis step or re-running a pipeline with a different set of parameters for only some of the reporting steps) while still allowing for a single output report.

There are two major ways of working with sub-reports or extended reports in Quarto: Quarto Projects, and Quarto Partials. Projects do not play well within a workflow manager context, as Quarto then needs to be the arbiter of what gets executed at what time, which depends on the working directory state (which will always be fresh in a Nextflow context). Partials, however, fits more nicely into the workflow manager context: a sub-report is first rendered to markdown (_not_ HTML), which is then referred to in the final report using the `{{< include partial.md >}}` notation, and the final report is rendered only once.

While the `quartonotebook` module is already quite flexible, there are several reasons why incorporating partial rendering into it doesn't work well:

 - Separation of concerns: the `quartonotebook` module renders a single Quarto document to HTML, where partials needs to render intermediate markdown documents.
 - While some input/output is shared, there is some non-overlapping parts: partials do not use extensions, while the main output differs (an HTML versus a markdown + related files, e.g. images). Having a single module would mean that almost all output would be optional due to having to account for the different render modes, which feels brittle to me.
 - Rendering a partial needs to explicitly use `--to markdown`, which overwrites whatever has been specified in the report. The `quartonotebook` module doesn't use `--to` at all, which allows a user to specify the format in the `.qmd` file themselves, which might include a custom format in e.g. an extension. Keeping the modules separate here makes it easier to work with both, circumventing having to use `ext.args` for every single module use (which would also be more opaque).
 - Partials are more of a pure Quarto thing, and while it's probably possible to make it work with the other input types that `quartonotebook` can handle, I had some issues there. If somebody is working with partials, it is reasonable to assume a Quarto-centric system, thus removing support of the other formats (`.ipynb` and `.rmd`) and for the test suite.

A lot of the `script:` part of the modules remain similar, though, but I believe it's worth it due to the above mentioned things.

Should this module be accepted, the existing `quartonotebook` module would need to be renamed to e.g. `quarto/notebook`. I'm not sure what the procedure for renaming modules is, but I'm happy to make a PR both to the modules repository for this, as well as for all of the pipelines currently using the module (which is four, at time of writing) after its been updated.

## PR checklist

- [X] Add a resource `label`
  - For modules:


modules/nf-core/quarto/partial/main.nf
```diff
@@ -0,0 +1,102 @@
...
+
+    script:
+    def args = task.ext.args ?: ''
+    // Partial is meant to be run once (not per sample), hence the naming scheme
+    prefix = task.ext.prefix ?: "${notebook.baseName}"
+    // Implicit parameters can be overwritten by supplying a value with parameters
+    notebook_parameters = [
+        meta: meta,
+        cpus: task.cpus,
+        artifact_dir: "artifacts",
```
SPPearce: When would we want to be able to change the artifact directory? Could it not just be hardcoded entirely?

modules/nf-core/quarto/partial/main.nf
```diff
@@ -0,0 +1,102 @@
...
+    tuple val(meta), path("${notebook_parameters.artifact_dir}/*")                             , emit: artifacts, optional: true
+    tuple val("${task.process}"), val('quarto')   , eval('quarto -v')                          , emit: versions_quarto   , topic: versions
+    tuple val("${task.process}"), val('papermill'), eval('papermill --version | cut -f1 -d" "'), emit: versions_papermill, topic: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    // Partial is meant to be run once (not per sample), hence the naming scheme
```
SPPearce: Why?

modules/nf-core/quarto/partial/main.nf
fasterius: This is something I've thought about changing myself, and would love some more discussions around - I kept it for this PR to mirror how the original QUARTONOTEBOOK (and RMARKDOWNNOTEBOOK / JUPYTERNOTEBOOK) module works. In my experience, most of the time the default is used without changing anything (which is cheap in terms of added boilerplate code, as the input parameter can just be `[]` when using default values). However, I've been experimenting with changing the artifact dir to better mirror the various contents that different module invocations yield - for example, the way I run my projects I would divide up into separate steps, maybe one for quality controls and one for analyses. Being able to change the artifact directory means being able to publish outputs into e.g. `results/qc` and `results/analysis` instead of `results/artifacts` (with the possibility to add subdirectories for the different output types, of course).

modules/nf-core/quarto/partial/main.nf
fasterius: There were several iterations on the general module structure before I came to the conclusion that one run per sample was the best tradeoff overall. The problem is that while it's easy on the Nextflow side to name output partials along the lines of `${notebook}-${meta.id}`, partials have the added complication of needing the be explicitly named in the final HTML rendering (which would be done with the QUARTONOTEBOOK module) in `{{< include >}}` blocks. This would mean that any pipeline would have to hard-code multiple of these partial names into the final notebook, which seems non-ideal to me from a workflow standpoint. 

The way I currently use partials (and the envisioned use-case for this module) is sequential reports where each subsequent report uses the output of the previous one, handling all of the samples at the same time (which is usually at the end of a per-sample pipeline, where one wants to analyse all the samples in aggregate). Any pipeline wanting to use the one-partial-per-sample route could change `prefix` to e.g. `${notebook.baseName}-${meta.id}` using `ext.prefix` though, but since the way I use it currently is one-partial-per-notebook I put this as the default.

modules/nf-core/quarto/partial/main.nf
mahesh-panchal: The rmarkdownnotebook and jupyternotebook modules have this directory hardcoded. We have both solutions. The published directory name can also be controlled with publishing mechanisms. However the important question is if the artifacts need to be staged in the receiving process in which case they cannot have the same name.

modules/nf-core/quarto/partial/main.nf
mahesh-panchal: I don't find the comment clear enough to warrant it's inclusion here. I feel it's just a confusing comment.

modules/nf-core/quarto/partial/main.nf
```diff
@@ -0,0 +1,102 @@
...
+    val parameters
+    path input_files
+
+    output:
+    tuple val(meta), path("${prefix}{.md,_files}")                                             , emit: partial
+    tuple val(meta), path(notebook)                                                            , emit: notebook
+    tuple val(meta), path("params.yml")                                                        , emit: params_yaml
+    tuple val(meta), path("${notebook_parameters.artifact_dir}/*")                             , emit: artifacts, optional: true
+    tuple val("${task.process}"), val('quarto')   , eval('quarto -v')                          , emit: versions_quarto   , topic: versions
+    tuple val("${task.process}"), val('papermill'), eval('papermill --version | cut -f1 -d" "'), emit: versions_papermill, topic: versions
```
mahesh-panchal, line 31: Is papermill still a necessity for this module?
It's just a side question though since this module is intended to be used after patching.

modules/nf-core/quarto/prerender/main.nf
```diff
@@ -0,0 +1,102 @@
+// NB: You'll likely want to override this with a container containing all
+// required dependencies for your analyses, or use wave to build the container
+// for you from the environment.yml. You'll at least need Quarto itself,
+// Papermill and whatever language you are running your analyses on; you can see
+// an example in this module's environment file.
```
mahesh-panchal: Can you update this note to remind users to also edit the output to emit versions for the updated packages and use `nf-core modules patch` to save the changes.

mahesh-panchal (commented):
So, generally, I'm in favour of this. My issue currently is the terminology. If you look for partials in the documentation, the search focuses on template partials which is not what this is. As I understand it, one can precompute parts and then render the final report separately as long as the precomputed stuff is also staged. 

Or is this something completely different to that? This is an old implementation of what I mean: https://github.com/mahesh-panchal/nextflow-quarto-website/blob/main/main.nf where each part is rendered in different environments and then rendered again to make the final report. 

Can you link the specific doc section at the top of the module in the note?

Since the inclusion mechanism is `{{< include ... >}}`, why does this need to specifically be md and not say qmd? https://quarto.org/docs/authoring/includes.html

As for the module renaming, it's done through the deprecation mechanism normally https://nf-co.re/docs/developing/components/deprecating-components

SPPearce (commented):
Thanks Mahesh, you seem to understand these report generating modules more than me

modules/nf-core/quarto/prerender/main.nf
mahesh-panchal: Perhaps include as TODO's so it's picked up by pipeline linting. I don't remember if module linting will flag it though.

modules/nf-core/quarto/partial/main.nf
fasterius: I was not aware that the RMARKDOWNNOTEBOOK/JUPYTERNOTEBOOK modules now uses a hard-coded artifact directory - when I originally created the QUARTONOTEBOOK module they were not hard-coded, so I stuck with that. Good point about publishing. In regards to staging, the way I envision the module to work is indeed that some of the artifacts would need to be staged in subsequent modules, be it another instance of this module or the final QUARTONOTEBOOK module that renders the final HTML. However, since I made it with the notion of one instance per partial, there wouldn't be any identical names in-between instances, so that's (probably?) a moot point in regards to hard-coding or not.

modules/nf-core/quarto/partial/main.nf
fasterius: Fair enough. Should it be removed entirely or expanded to more clearly explain the overall idea?

modules/nf-core/quarto/partial/main.nf
fasterius: Yes, it's still necessary to have Papermill in order for parametrisation to work.

modules/nf-core/quarto/prerender/main.nf
fasterius: For version exporting, there are two alternatives: patching the module as you described, or to export a `versions.yml` file from the partial `.qmd` itself (currently not listed in the output of the module). What would be the favoured solution for these types of modules? I know that the Python/R modules used in e.g. `nf-core/scdownstream` and `nf-core/spatialvi` uses the second solution, but these are local modules. My spontaneous feeling is that exporting a `versions.yml` would be cleaner, as the user would just add a few lines of code to their QMD file, rather than having to patch the module, but maybe there's other things I'm missing here.

fasterius:
> So, generally, I'm in favour of this. My issue currently is the terminology. If you look for partials in the documentation, the search focuses on template partials which is not what this is. As I understand it, one can precompute parts and then render the final report separately as long as the precomputed stuff is also staged.
>
> Or is this something completely different to that? This is an old implementation of what I mean: https://github.com/mahesh-panchal/nextflow-quarto-website/blob/main/main.nf where each part is rendered in different environments and then rendered again to make the final report.

That is indeed what the goal of this module is: pre-computing some things and including them in a final report without re-computing anything. I agree that the naming is indeed wrong, and something like `quarto/includes` is probably more appropriate.

> Since the inclusion mechanism is `{{< include ... >}}`, why does this need to specifically be md and not say qmd? https://quarto.org/docs/authoring/includes.html

Because including a QMD would just paste the code itself into the final document, meaning that everything would be run (invalidating the whole point of this module). Including a markdown document that has already been "rendered" (so to speak) means that the results get included, not the code itself.

modules/nf-core/quarto/prerender/main.nf
mahesh-panchal: Either way requires a patch here, currently. The output isn't set to look for a versions.yml. Emitting the versions.yml would be the more flexible option as you're not bound to a container definition then. You could alias the module multiple times for different parts and containers.

modules/nf-core/quarto/partial/main.nf
mahesh-panchal: I would say remove it.

modules/nf-core/quarto/partial/main.nf
mahesh-panchal: The artifacts dir doesn't include the prefix by default so it would likely collide with the artifacts dir of another if not changed by parameters.

mahesh-panchal (commented):
> Because including a QMD would just paste the code itself into the final document, meaning that everything would be run (invalidating the whole point of this module). Including a markdown document that has already been "rendered" (so to speak) means that the results get included, not the code itself.

Even if the cache is propagated? I haven't tried it with includes, but wouldn't using [freeze = true](https://quarto.org/docs/projects/code-execution.html#freeze) mean the content wouldn't be rerun? The freeze cache needs to be propagated in that case though like in my example.

Maybe for naming use something like quarto_subdocument or quarto_documentpart, rather than include.

modules/nf-core/quarto/prerender/main.nf
fasterius: I would then just add an optional `versions.yml` output, so it wouldn't need a patch.

modules/nf-core/quarto/partial/main.nf
fasterius: Will do.

modules/nf-core/quarto/partial/main.nf
fasterius: Ah, right. So, where do you lean on this, hard-coded or not?

fasterius:
> Even if the cache is propagated? I haven't tried it with includes, but wouldn't using [freeze = true](https://quarto.org/docs/projects/code-execution.html#freeze) mean the content wouldn't be rerun? The freeze cache needs to be propagated in that case though like in my example.

I haven't tried using freeze at all, but wouldn't that require the use of a Quarto project? I have only experimented a little bit with projects inside Nextflow, and I generally felt that they were an overall burden for Nextflow-centric work - I kind of went away with the notion that Quarto projects are great outside of Nextflow, and that with Nextflow I wanted as lean of a solution as possible.

> Maybe for naming use something like quarto_subdocument or quarto_documentpart, rather than include.

Wouldn't using `quarto/includes` make it clear where to find it in the documentation though, and be more explicit in what the module uses in terms of functionality? I'd probably agree that `quarto/subdocument` or similar is clearer intent on the face of it, but since this module is sort of "deep" into Quarto's capabilities there's an argument that mentioning "includes" has merit, as it is a common pattern of [shortcodes](https://quarto.org/docs/authoring/shortcodes.html) in Quarto. Another similar option could be `quarto/prerender`.

mahesh-panchal:
I had a play around, and freeze is at the page level so it doesn't work here. 

I guess this also depends on what kind of output we want. For a single page report, this workflow (rendering portions to md) is the only one that works. However, if you style your report as a webpage, then, freeze is also an option. 

For naming, let's go with `quarto/prerender` then. 

My remaining comments are then to name the artifacts and params.yml outputs with the prefix as default.

modules/nf-core/quarto/partial/main.nf
mahesh-panchal: not hard coded, and prefix with `prefix`.

modules/nf-core/quarto/prerender/tests/nextflow.config
mahesh-panchal, line 1: Why does the cpus need to be set here?

modules/nf-core/quarto/prerender/main.nf
mahesh-panchal: This is tricky. I would not be comfortable leaving `versions.yml` as an optional file. I feel like it should an either or thing. Either a single versions.yml or multiple `eval`s.

@SPPearce could you provide a second opinion?

modules/nf-core/quarto/prerender/tests/nextflow.config
fasterius: Because I was getting differing snapshots between the GitHub CI and locally without explicitly setting the number of cpus to use. Maybe that wasn't the right way to go about it?

modules/nf-core/quarto/prerender/main.nf
fasterius: Well, it's a bit complicated, since there are several components that would need to be version checked: the version of Quarto, the version of Papermill, and the version of whatever the user/pipeline is using inside the notebook. The former two are easily checked and exported into the topic, while the latter needs to be done inside the notebook by the user/pipeline. I feel like the former two controls how rendering has happened, while the latter controls whatever analysis is performed in the notebook.

If we look at the `quartonotebook` module from which this originates, we don't even ask for versions of packages used inside the notebook - mostly because I couldn't think of a good way to do it when that module was created. I guess making it non-optional would force the user/pipeline to properly report versions, although it would also force the addition of boilerplate code inside the notebooks.

modules/nf-core/quarto/prerender/tests/nextflow.config
mahesh-panchal: By differing snapshots you mean the `params.yml` output is changing? Can you include a comment then that this is assigned to keep `params.yml` (or whichever files are affected) snapshot fixed. Otherwise it will just be queried again in future.

fasterius: Done.

modules/nf-core/quarto/prerender/main.nf
mahesh-panchal: I think the versions.yml should be not optional. 
The module itself should have a message that the qmd should write a versions.yml in addition.

Also the test dataset qmd should be updated to also write a versions.yml then.

modules/nf-core/quarto/prerender/tests/nextflow.config
mahesh-panchal: Include all affected files please in the comment.

modules/nf-core/quarto/prerender/tests/nextflow.config
```diff
@@ -1,4 +1,5 @@
-// Explicitly set number of CPUs so that `params.yml` is stable in CI testing
+// Explicitly set number of CPUs so that `params.yml` is stable in CI testing,
+// which affects the snapshot of `params.yml` and the final rendered markdown.
```
mahesh-panchal:
```suggestion
// Pinned: task.cpus is embedded in params.yml and printed in the rendered
// markdown, so leaving it unset breaks snapshot reproducibility across machines.
```

modules/nf-core/quarto/prerender/main.nf
fasterius: I've started to work in this, but quickly realised it's going to affect a lot of things. The test-dataset notebooks needs to add writing of a versions YAML, which will also require the addition of a new `task_process` parameter in order to correctly format the resulting YAML, which means also changing the module code for the three affected modules (`rmarkdownnotebook`, `jupyternotebook`, `quartonotebook` plus the new `quarto/notebook`).  The `rmarkdownnotebook` currently exports a session info file, which I assume is its way of storing software versions, while `jupyternotebook` doesn't seem to have anything like this, similar to `quartonotebook`.

It's a fairly large deviation from how these modules have worked in the past. I'm fine with doing all this, but would like some confirmation that this is the way we want to go for all of these modules before spending time on it.

mahesh-panchal: Is this a suitable intermediate solution?
```nextflow
if [ ! -f notebook_versions.csv ]; then
    echo "ERROR: notebook_versions.csv not found — the .qmd script must write out tool,version pairs used within the notebook." >&2
    exit 1
fi

"${task.process}":
    quarto: \$(quarto --version)
    papermill: \$(papermill --version | cut -f1 -d' ')
\$(awk -F',' '{printf "    %s: %s\n", \$1, \$2}' notebook_versions.csv)
END_VERSIONS
```

fasterius: Sure, I guess that'd be fine. I'd prefer just going for a "final" solution straight away though, if this general pattern of writing software versions inside of notebooks is something we want to have for nf-core in general. @grst is listed as maintainer for the Rmarkdown/Jupyter notebooks, what are your thoughts on this?

mahesh-panchal: By intermediate, I meant one that didn't involve passing the task.process as a parameter. The qmd only needs to write out it's environment in a common format. This is my suggestion for what should be implemented, but I don't see any issues having not tested it myself. 

I also wonder if it's worth separating the module environment from the quarto execution environment. Since the other two are already in a topic, one could write the versions.yml as a kind of execution environment yaml instead.

```nextflow
"${task.process}_${prefix}":
```
and it should appear as a separate entry.

grst: The module might have evolved quite a bit since I initially developed it. Some thoughts nonetheless: 
 * a full notebook environment can easily have 100s of packages - do we really want to dump this into a version.yml? 
 * If we want that, for R, sessionInfo is the natural way of getting all loaded packages. For Python, there's now [session-info2](https://session-info2.readthedocs.io/en/latest/index.html#) (maintained by scverse). 
 * alternatively, one could just instruct the user to manually export the versions of selected packages into a file.

fasterius: I generally don't feel the need to dump _all_ versions into a file, just the "main" ones, _i.e._ those that would be specified in a Conda `environment.yml` file from _e.g._ Seqera Containers or however the packages are installed.

I generally think separating the module environment from the notebook environment is a good thing. That way, we keep the former in a topic (current implementation) and add however we want the notebook environment specified.

I'm not sure what the best way to export notebook versions is. Adding `process_name` as a parameter mirrors how python/R modules do it in _e.g._ scdownstream/spatialvi templates, which is good for consistency, but it adds boilerplate code in that the user needs to add another default parameter to their notebook (in addition to `meta` and `cpus` which already exists). The `notebook_versions.csv` solution doesn't need this, but is less consistent with other modules. Both solutions probably add comparable amount of code in terms of writing a YAML or CSV. Which solution do you guys prefer?

mahesh-panchal: The notebooks could be using languages other than R/Python. Getting the author to write a simple csv with the relevant package and versions seems like a good compromise then.

How would having the `task.process` as a parameter work when the notebook is used outside of the workflow? That seems to be part of the how to use these notebooks too, which is why I think not adding it as a notebook parameter works well and just having a list. Outside the Nextflow workflow, the user doesn't have to worry about the Nextflow context it was compiled in just to write a version file only relevant to the pipeline.

> alternatively, one could just instruct the user to manually export the versions of selected packages into a file.

So we've basically converged on the same thing with the csv.

fasterius: Okay, I've implemented this now and opened a PR to the test-datasets repo with changed to the notebooks: https://github.com/nf-core/test-datasets/pull/2216. I'll add tests once that PR is accepted and we agree on the implementation.

Regarding parameters outside of the Nextflow context: one can just supply default values for parameters, which are overridden by values provided by Nextflow. I usually just set default values to _e.g._ `None` or `NULL` in Python and R, respectively, or `1` for the `cpus` parameter.

mahesh-panchal (approved):
Thank you.

PR #11571, opened 2026-05-08
Add HTSLIB/BGZIPTABIX and deprecate redundant modules

itrujnara:
⚠️ **Heads up:** This PR contains numerous significant changes. Proper review might take time.

## Changes
* Added HTSLIB/BGZIPTABIX module for bgzip compression, decompression, and indexing
* Deprecated SAMTOOLS/BGZIP
* Deprecated TABIX/BGZIP
* Deprecated TABIX/BGZIPTABIX
* Deprecated TABIX/TABIX
* Replaced all deprecated modules with HTSLIB/BGZIPTABIX in subworkflows and CI setup blocks
 
### ⚠️ Breaking changes
* HTSLIB/BGZIPTABIX is not an exact drop-in replacement for any of the deprecated modules; some channel tweaks and/or additional inputs are required
* Automatic renaming from TABIX/TABIX was not ported; pipelines must control the naming of its outputs through configs
 
## Rationale
This change was initially proposed by Maxime. bgzip-related operations are currently spread across 4 different modules with largely overlapping functionality. The modules are not consistent in their API nor behavior. This makes it hard for developers to quickly distinguish their purposes and choose the right tools. Additionally, it creates an unnecessary maintenance burden. Finally, none of the tools is placed in a correctly named subdirectory. A single, multi-purpose module will guarantee a predictable and consistent behavior, and facilitate maintenance in the future.

### Purpose
The new module can be used for:
* compressing files to bgzip
* converting other compression types to bgzip (currently supported: gz, bz2, xz)
* decompressing bgzip and several other formats
* indexing compressed tabular files with tabix 

### Inputs
The module accepts the following input channels:
* `meta, infile, index, regions`: meta map, arbitrary input file, optional tabix index
* `action`: explicit action input: `compress` or `decompress`
* `make_index`: boolean, whether to generate a tabix index
* `out_ext`: output file extension (Mahesh style, as discussed on Slack)

### Intended behavior
If `action = "compress"`, the module should:
* compress an uncompressed file with bgzip
* decompress a non-bgzip compressed file and recompress with bgzip
* return a bgzip-compressed file as-is
* if requested, index the resulting bgzip with tabix
* if regions are passed, performs reindexing with tabix 
If a tabix index is passed, it is returned as-is and no new index is generated (for consistency with prior modules).

If `action = "decompress"`, the module should:
* decompress a bgzip or other compressed file
* return an uncompressed file as-is
* not create any index
 
Any other action is invalid.

### Error handling
[...]

PR #12197, opened 2026-06-30
htslib-rebgzip module

nschan:
This module takes either unzipped, xz, bz2 or gz input, and returns a bgzipped file. This can be helpful since some tools refuse to take regular gzip. 
I have kept the input / output as two element tuples, which should be compatible with most tool outputs.


itrujnara:
How is the behaviour here different from `htslib/bgziptabix` with `compress` and no index? You might be about to reintroduce the very issue that led to the creation of the `bgziptabix` module

itrujnara (changes requested):
Looks duplicative, blocking until a compelling justification is provided

nschan:
The difference is largely that this is convenience module that does not require `.map {}` on common output tuples in the form of `tuple val(meta) file(maybe bgzipped or not)`.

nschan:
> You might be about to reintroduce the very issue that led to the creation of the bgziptabix module

I am not aware of this history of this module, which issue are you referring to?

itrujnara:
I see your point, I will need to discuss it with others before deciding how to solve this.

itrujnara:
> > You might be about to reintroduce the very issue that led to the creation of the bgziptabix module
> 
> I am not aware of this history of this module, which issue are you referring to?

Originally we had `bgzip/bgzip`, `bgzip/tabix`, `bgzip/bgziptabix`, and `samtools/bgzip`, each of which was slightly different. The setup created a lot of maintenance effort, since changes had to propagate to all of them. In [a recent PR](https://github.com/nf-core/modules/pull/11571), we combined them into the current `htslib/bgziptabix` to have all the functionality in one place.

maxulysse:
Not convinced at all by adding such a duplicate of the already existing htslib module

famosab:
Maybe we can think more if we can ease the use of htslib/bgziptabix ?

nschan:
For me the main motivation is that I would prefer an "unsupervised" module, that gives me `bzipped` file, without me having to do anything extra, or having to care too much about the input. As I understand the current `bgziptabix` module, it is possible to do re-compression, but would require me to provide `out_ext` to name the output correctly, in addition to the input-channel wrangling mentioned above. The module proposed here is also able to have identical in and output filenames, which may be useful for `.gz` files that should be turned into bgzipped files.

A simplification of `bgziptabix` would also be good, to me the issue is that `bgziptabix` actually does more than `bgzip` only, which makes it clumsy to use for simple `bgzip` tasks. Given that all 'pure' bgzip modules were deprecated in favor of `bgziptabix`, to me this turns a routine task unnecessarily convoluted.

nschan:
After reading through #11571, it seems that what I feel is inconvenient about the `bgziptabix` module was a deliberate choice.

PR #11738, opened 2026-05-21
feat(vcontact3): add vcontact3/prepareddatabases module

olawal1001:
## PR checklist

Closes #XXX 

- [ ] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [ ] Remove all TODO statements.
- [ ] Broadcast software version numbers to `topic: versions` - [See version_topics](https://nf-co.re/blog/2025/version_topics)
- [ ] Follow the naming conventions.
- [ ] Follow the parameters requirements.
- [ ] Follow the input/output options guidelines.
- [ ] Add a resource `label`
- [ ] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - For modules:
    - [ ] `nf-core modules test <MODULE> --profile docker`
    - [ ] `nf-core modules test <MODULE> --profile singularity`
    - [ ] `nf-core modules test <MODULE> --profile conda`
  - For subworkflows:
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile docker`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile singularity`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile conda`

PR #12198, opened 2026-06-30
feat: add vcontact3/run module

### LINKED ISSUE #11569: new module: VCONTACT3/RUN

### Further Information

https://bitbucket.org/MAVERICLab/vcontact3/src/master/
https://vcontact3.readthedocs.io/en/latest/

```yaml
---
channels:
  - conda-forge
  - bioconda
dependencies:
  - bioconda::vcontact3=3.1.6
```

JP-Bro:
## Description
Implements the VCONTACT3/RUN module for viral genome clustering using protein-level homology comparison.

## Related Issues
Closes #11569

## Module Details
- **Tool**: VCONTACT3 v3.1.6
- **Purpose**: Cluster and compare viral genomes
- **Inputs**: FASTA/GenBank genome files
- **Outputs**: Cluster assignments, comparison networks, cluster profiles

## Checklist
- [x] Module follows nf-core standards
- [x] Process definition in main.nf includes proper inputs/outputs
- [x] Metadata (meta.yml) is complete
- [x] Conda environment (environment.yml) specified
- [x] Test structure created
- [x] Code is clean and documented

## Testing
Module created with proper Nextflow DSL2 structure. Ready for integration testing with nf-core CI/CD pipeline.


modules/nf-core/vcontact3/run/tests/tags.yml
mashehu, line 1: we don't use this file anymore 🙂

mashehu (commented):
it looks like you started this module an outdated template, so I would recommend to rerun `nf-core moduels create` with nf-core/tools version 4.0.2. (and then copy over the code)

modules/nf-core/vcontact3/run/main.nf
```diff
@@ -0,0 +1,47 @@
+process VCONTACT3_RUN {
+    tag "$meta.id"
+    label 'process_medium'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularityOptions ?
```
mashehu:
```suggestion
    container "${ workflow.containerEngine in ['singularity', 'apptainer'] && !task.ext.singularityOptions ?
```
is the current template form

modules/nf-core/vcontact3/run/main.nf
```diff
@@ -0,0 +1,47 @@
...
+    vcontact3 run \\
+        -i ${genomes.join(' ')} \\
+        -o vcontact3_output/ \\
+        --threads ${task.cpus} \\
+        ${args}
+
+    cat > versions.yml <<-EOF_VERSIONS
+        "${task.process}":
+            vcontact3: \$( vcontact3 --version 2>&1 | grep -oP 'vcontact3, version \\K[^\\s]+' )
+    EOF_VERSIONS
```
mashehu: use version topics please https://nf-co.re/docs/tutorials/migrate_to_topics/update_modules

mashehu (commented):
looks like you missed  the template changes iin the main.nf 🙂

modules/nf-core/vcontact3/run/tests/main.nf.test
```diff
@@ -0,0 +1,32 @@
...
+                    file(params.modules_testdata_base_path + 'genomics/sarscov2/fasta/test_1.fasta', checkIfExists: true)
+                ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out.versions).match("versions") }
```
mashehu:
```suggestion
                { assert snapshot(process.out)}
```

modules/nf-core/vcontact3/run/tests/main.nf.test
```diff
@@ -0,0 +1,32 @@
...
+                    file(params.modules_testdata_base_path + 'genomics/sarscov2/genome/genome.fasta', checkIfExists: true)
+                ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out.results, process.out.versions).match() }
```
mashehu:
```suggestion
                { assert snapshot(process.out).match() }
```

modules/nf-core/vcontact3/run/tests/main.nf.test
```diff
@@ -0,0 +1,32 @@
...
+    name "Test Process VCONTACT3_RUN"
+    script "../main.nf"
+    process "VCONTACT3_RUN"
+
+    tag "modules"
+    tag "modules_nfcore"
+    tag "vcontact3"
+    tag "vcontact3/run"
+
+    test("vcontact3_run - stub") {
```
mashehu, line 9: why not a full test?

modules/nf-core/vcontact3/run/main.nf
```diff
@@ -0,0 +1,40 @@
...
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularityOptions ?
+        'docker://quay.io/biocontainers/vcontact3:3.1.6--pyhdfd78af_0' :
+        'quay.io/biocontainers/vcontact3:3.1.6--pyhdfd78af_0' }"
+
+    input:
+    tuple val(meta), path(genomes)
+
+    output:
+    tuple val(meta), path("vcontact3_output/"), emit: results
+    tuple val("${task.process}"), val('vcontact3'), val(versions), topic: versions, emit: versions_vcontact3
```
mashehu:
```suggestion
     tuple val("${task.process}"), val('vcontact3'), eval("vcontact3 --version 2>&1 | sed -n 's/vcontact3, version \([^ ]*\)/\1/p'"), emit: versions_vcontact3, topic: versions
```

modules/nf-core/vcontact3/run/meta.yml
```diff
@@ -0,0 +1,56 @@
...
+
+tools:
+  - vcontact3:
+      description: Automated host-to-phage gene-level similarity searching
+      homepage: https://bitbucket.org/MAVERICLab/vcontact3
+      documentation: https://vcontact3.readthedocs.io/en/latest/
+      tool_dev_url: https://bitbucket.org/MAVERICLab/vcontact3
+      licence:
+        - GPL-3.0-or-later
+      publication_doi: 10.1186/s13059-019-1879-9
```
mashehu:
```suggestion
	  doi: 10.1186/s13059-019-1879-9
```

modules/nf-core/vcontact3/run/meta.yml
```diff
@@ -0,0 +1,56 @@
...
+
+output:
+  results:
+    - - meta:
+          type: map
+          description: Groovy map with sample metadata
+      - vcontact3_output/:
+          type: directory
+          description: VCONTACT3 output directory containing clustering results
+          pattern: "vcontact3_output/"
```
mashehu: you also need to add the versions_vcontact3 channel here.
This should be automatically resolved with`nf-core modules lint --fix`

modules/nf-core/vcontact3/run/main.nf
```diff
@@ -12,7 +12,7 @@ process VCONTACT3_RUN {
 
     output:
     tuple val(meta), path("vcontact3_output/"), emit: results
-    tuple val("${task.process}"), val('vcontact3'), eval("vcontact3 --version 2>&1 | sed -n 's/vcontact3, version \\([^ ]*\\)/\\1/p'"), emit: versions_vcontact3, topic: versions
+    tuple val("${task.process}"), val('vcontact3'), val('3.1.6'), emit: versions_vcontact3, topic: versions
```
mashehu: please keep the eval. that's the correct way to do it. it willl also work for the stub then

modules/nf-core/vcontact3/run/main.nf
```diff
@@ -0,0 +1,37 @@
...
+        -o vcontact3_output/ \\
+        --threads ${task.cpus} \\
+        ${args}
+    """
+
+    stub:
+    """
+    mkdir -p vcontact3_output/
+    touch vcontact3_output/clusters.csv
+    touch vcontact3_output/merged_df.csv
```
mashehu: based on this, the output section should capture the csv files, not the whole directory

mashehu:
are you writing the snapshots by hand? they don't follow the nf-test format. you can generate them using `nf-test test --tag vcontact3 --update-snapshot`

JP-Bro:
Fixed the crash and CI is now green ✅

Root cause: VCONTACT3_RUN was calling vcontact3 run with -d (database path) unset — the process crashed every time since db_path was None.

Fix:

Added database as a declared path input in main.nf, wired into -d ${database}
Updated meta.yml to document the new input (ran nf-core modules lint --fix to match main.nf exactly)

modules/nf-core/vcontact3/run/tests/main.nf.test
famosab, line 1: You need to generate the snapshot with nf-core modules test ... :)

JP-Bro:
added a proper assert snapshot(process.out).match() and generated the snapshot via nf-core modules test vcontact3/run --update (Docker profile)

modules/nf-core/vcontact3/run/tests/main.nf.test
SPPearce: Echoing this, why do you not have an actual test?

modules/nf-core/vcontact3/run/main.nf
```diff
@@ -0,0 +1,40 @@
...
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+
+    """
+    vcontact3 run \\
+        -n ${genomes.join(' ')} \\
+        -o vcontact3_output/ \\
```
SPPearce: Can the tool not just make these files in the current directory?

SPPearce: Alternatively, make the name of the output folder customizable using ext.prefix

JP-Bro:
@SPPearce Went with task.ext.prefix — added prefix = task.ext.prefix ?: "${meta.id}" to both the script and stub blocks, and the output path now uses ${prefix}/ instead of the hardcoded vcontact3_output/. Defaults to meta.id when no prefix is set, matching the standard nf-core pattern

Updated meta.yml to reflect the new output pattern and verified locally — stub test passes, output correctly lands under test/ for the default case, and nf-core modules lint vcontact3/run is clean

Let me know if any other changes are needed

modules/nf-core/vcontact3/run/tests/main.nf.test
```diff
@@ -0,0 +1,27 @@
...
+                    [ id:'test' ],
+                    file(params.modules_testdata_base_path + 'genomics/sarscov2/genome/genome.fasta', checkIfExists: true)
+                ]
+                input[1] = file(params.modules_testdata_base_path + 'genomics/sarscov2/genome/genome.fasta.fai', checkIfExists: true)
+                """
+            }
+        }
+        then {
+            assert process.success
+            assert snapshot(process.out).match()
```
SPPearce:
```suggestion
            assert snapshot(sanitizeOutput(process.out)).match()
```

modules/nf-core/vcontact3/run/main.nf
SPPearce: Why are you hardcoding the version, when [the tool](https://vcontact3.readthedocs.io/en/latest/cli/version.html#vcontact3-version) suggests it has a version command.

SPPearce (commented):
You still need an actual non-stub test. And use the versions.
I wonder if this tool should actually be called vcontact, not vcontact3.

modules/nf-core/vcontact3/run/main.nf
```diff
@@ -0,0 +1,42 @@
...
+        'docker://quay.io/biocontainers/vcontact3:3.1.6--pyhdfd78af_0' :
+        'quay.io/biocontainers/vcontact3:3.1.6--pyhdfd78af_0' }"
+
+    input:
+    tuple val(meta), path(genomes)
+    path database
+
+    output:
+    tuple val(meta), path("${prefix}/"), emit: results
+    tuple val("${task.process}"), val('vcontact3'), val('3.1.6'), emit: versions_vcontact3, topic: versions
```
erikrikarddaniel: The already-merged sibling module `vcontact3/prepareddatabases` (#11738) has this solved:

```suggestion
    tuple val("${task.process}"), val('vcontact3'), eval('vcontact3 version'), emit: versions_vcontact3, topic: versions
```

For what it's worth, I think the earlier suggestion of `vcontact3 --version 2>&1 | sed ...` is why this kept failing. There is no `--version` flag — `build_parser()` in `vcontact3/cli.py` registers only the subcommands `version`, `prepare_databases` and `run`, so `--version` exits with *unrecognized arguments*. The `version` subcommand does `print(__version__)` and nothing else, so no `sed` is needed.

Two follow-ups: `-stub` is not an obstacle here — I checked on Nextflow 26.04.6 and `eval` outputs do resolve under `-stub`, reaching both the named channel and the `versions` topic. And `meta.yml` lines 50 and 61 currently use `3.1.6:` as the key where the eval command string belongs; `nf-core modules lint --fix` should sort that out once this line changes.

modules/nf-core/vcontact3/run/main.nf
```diff
@@ -0,0 +1,42 @@
+process VCONTACT3_RUN {
+    tag "$meta.id"
+    label 'process_medium'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularityOptions ?
+        'docker://quay.io/biocontainers/vcontact3:3.1.6--pyhdfd78af_0' :
+        'quay.io/biocontainers/vcontact3:3.1.6--pyhdfd78af_0' }"
```
erikrikarddaniel: This differs from the current `nf-core/tools` module template in three ways, and `vcontact3/prepareddatabases` on master already has the right form:

```suggestion
    container "${ workflow.containerEngine in ['singularity', 'apptainer'] && !task.ext.singularity_pull_docker_container ?
        'https://depot.galaxyproject.org/singularity/vcontact3:3.1.6--pyhdfd78af_1':
        'quay.io/biocontainers/vcontact3:3.1.6--pyhdfd78af_1' }"
```

- `== 'singularity'` misses apptainer.
- `task.ext.singularityOptions` appears in **0** modules on master; the template and 1928 modules use `task.ext.singularity_pull_docker_container`. (An earlier review suggestion here used the `singularityOptions` spelling, so applying that one verbatim would still leave this wrong — worth flagging.)
- `docker://quay.io/biocontainers/...` makes Singularity convert a Docker image at runtime on every host. 996 modules use the pre-built `https://depot.galaxyproject.org/singularity/` image instead.

Note I've also changed the build to `pyhdfd78af_1` to match the sibling module — right now the two vcontact3 modules pull different builds of the same tool version. Whichever is current, they should agree.

modules/nf-core/vcontact3/run/main.nf
```diff
@@ -0,0 +1,42 @@
...
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "${meta.id}"
+
+    """
+    vcontact3 run \\
+        -n ${genomes.join(' ')} \\
```
erikrikarddaniel: I think this is a real bug that hasn't come up yet. From `vcontact3/cli.py`:

```python
inputs.add_argument('-n', '--nucleotide', type=Path, dest='nucleotides',
                    help='FASTA-formatted nucleotide file. ...')
```

`type=Path` with no `nargs` — it takes exactly **one** file. But the input is declared `path(genomes)` (plural) and joined with spaces, which advertises multi-file support the tool doesn't have. With two files argparse consumes the first and errors on the rest.

It looks correct today only because the stub test passes a single file. I checked the Groovy side on Nextflow 26.04.6: one staged file gives a `UnixPath` whose `.join(' ')` happens to render `a.fasta`, while two give a `BlankSeparatedList` rendering `a.fasta b.fasta`. So the string is only accidentally right in the one case that is currently exercised — which is the same argument the maintainers have been making for a non-stub test.

Simplest fix is to make the interface honest:

```groovy
tuple val(meta), path(genome)
...
    -n ${genome} \\
```

If multi-genome runs are wanted, they need concatenating into one FASTA first, which is a different design.

modules/nf-core/vcontact3/run/main.nf
```diff
@@ -0,0 +1,42 @@
...
+    label 'process_medium'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularityOptions ?
+        'docker://quay.io/biocontainers/vcontact3:3.1.6--pyhdfd78af_0' :
+        'quay.io/biocontainers/vcontact3:3.1.6--pyhdfd78af_0' }"
+
+    input:
+    tuple val(meta), path(genomes)
+    path database
```
erikrikarddaniel, line 12: Two things about this input that I don't think have been raised.

**It doesn't compose with its own producer.** `VCONTACT3_PREPAREDDATABASES` emits `tuple val(meta), path("${prefix}/")` — a meta-wrapped *directory*. This takes a bare `path database` and `meta.yml` documents it as `type: file`. Chaining the two currently needs a `.map { _meta, db -> db }` in between. A bare `path` for a reference is fine, but the documented type should say directory.

**File-valued options can't be reached through `task.ext.args`.** `--len-nucleotide`, `--gene2genome` and `--proteins` are all `type=Path` in the CLI, so anyone needing them has to patch the module — `ext.args` can pass the flag but nothing stages the file. Worth deciding deliberately now rather than in a follow-up PR; adding them as optional inputs (`path len_nucleotide` with `[]` for absent) is the usual pattern.

modules/nf-core/vcontact3/run/environment.yml
```diff
@@ -0,0 +1,5 @@
+channels:
+  - conda-forge
+  - bioconda
+dependencies:
+  - bioconda::vcontact3=3.1.6=pyhdfd78af_0
```
erikrikarddaniel: Matching `vcontact3/prepareddatabases` on master:

```suggestion
---
# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/environment-schema.json
channels:
  - conda-forge
  - bioconda
dependencies:
  - bioconda::vcontact3=3.1.6
```

The build string matters: the module template says explicitly that *"for Conda, the build (i.e. `h9402c20_2`) must be EXCLUDED to support installation on different operating systems"*. Exactly one module on master pins one. The schema header is on 1812 of them.

modules/nf-core/vcontact3/run/meta.yml
```diff
@@ -0,0 +1,67 @@
...
+  - - meta:
+        type: map
+        description: Groovy map containing sample metadata
+    - genomes:
+        type: file
+        description: Input FASTA or GenBank files with viral sequences
+        pattern: "*.{fasta,fa,faa,gbk,gb}"
+        ontologies: []
+  - database:
+      type: file
```
erikrikarddaniel: Its only producer, `VCONTACT3_PREPAREDDATABASES`, emits a directory, and the CLI help for `-d/--db-path` says *"Path to database file or directory"*:

```suggestion
      type: directory
```

modules/nf-core/vcontact3/run/tests/main.nf.test
```diff
@@ -0,0 +1,27 @@
...
+    test("vcontact3_run - stub") {
+        options "-stub"
+        when {
+            process {
+                """
+                input[0] = [
+                    [ id:'test' ],
+                    file(params.modules_testdata_base_path + 'genomics/sarscov2/genome/genome.fasta', checkIfExists: true)
+                ]
+                input[1] = file(params.modules_testdata_base_path + 'genomics/sarscov2/genome/genome.fasta.fai', checkIfExists: true)
```
erikrikarddaniel, line 18: Concretely on the non-stub-test point others have raised: this passes a FASTA **index** (`.fai`) as the vContact3 reference database. It only gets through because `-stub` never runs the command, so nothing here exercises the argument construction at all — including the `-n` multi-file problem I've flagged in `main.nf`.

`VCONTACT3_PREPAREDDATABASES` is on master now, so a real test could chain the two, or use its stub output shape as the fixture.

erikrikarddaniel (dismissed):
See my comment above and in the code.

SPPearce (changes requested):
Just marking as request changes to make it clear there are things that need to be fixed.

JP-Bro:
Fixed:

Single-file input bug (@erikrikarddaniel ): path(genomes) + .join(' ') → path(genome), matching vcontact3's -n flag which only accepts one file (confirmed no nargs in cli.py)
Hardcoded version (@SPPearce , @erikrikarddaniel ): now uses eval('vcontact3 version') instead of val('3.1.6'), matching the sibling module
Container block (@erikrikarddaniel ): fixed the singularity/apptainer condition, corrected task.ext.singularity_pull_docker_container var name, switched to the prebuilt depot.galaxyproject.org image, bumped to pyhdfd78af_1 to match vcontact3/prepareddatabases
environment.yml: dropped the build hash pin, added the schema header, matching module template conventions
database input type: file → directory in meta.yml, matching what VCONTACT3_PREPAREDDATABASES actually emits
Snapshot sanitization (@SPPearce ): assert snapshot(sanitizeOutput(process.out)).match()

On the non-stub test: I attempted exactly what was suggested — chained VCONTACT3_PREPAREDDATABASES via nf-test's setup block to produce a real database, then ran VCONTACT3_RUN against it for real. The database download (2.1GB) completed successfully, but the actual vcontact3 run command crashes inside vcontact3's own code:

File "vcontact3/profiles.py", line 381, in __init__
    merged_overview_df.loc[merged_overview_df['Reference'] == True, 'shares_clusters'] = ...
TypeError: Invalid value 'ref only' for dtype 'float64'

I initially suspected a --db-domain mismatch (test genome is SARS-CoV-2/eukaryotic host, default domain is prokaryotes), so I tried --db-domain eukaryotes explicitly — same crash, different value ('withUser' instead of 'ref only'), same code path. This confirms it's not a domain issue: vcontact3 3.1.6 is writing string placeholder values into a pandas column that gets inferred as float64, which fails under pandas>=2.1.1's stricter setitem dtype coercion (per the bioconda recipe's pinned pandas version). This looks like a genuine upstream compatibility bug between vcontact3 3.1.6 and modern pandas, independent of anything in this module.

Given that, a working non-stub test isn't currently achievable regardless of test data or configuration — I've kept the stub-only test (with sanitizeOutput) rather than ship a test that can't pass. Happy to open an issue upstream with reproduction steps if that's useful, or revisit if there's a known workaround I'm missing.

erikrikarddaniel (approved):
I'm approving to not further block this, but it's a pity that not all CLI file arguments are reachable, could be made non-mandatory arguments to the module.

SPPearce:
Err, I think we need to understand why the tool doesn't work with the output of its own prep tool

JP-Bro:
@SPPearce 
Exact command that ran (container: quay.io/biocontainers/vcontact3:3.1.6--pyhdfd78af_1):

```
vcontact3 run \
    -n genome.fasta \
    -o test/ \
    --threads 4 \
    -d test \
    --db-domain eukaryotes
```

test/ here is the real output of VCONTACT3_PREPAREDDATABASES run for real (--get-version latest, database version 230, downloaded via vcontact3 prepare_databases itself — not a stub, not a hand-crafted fixture).

Full log up to the crash:

```
[INFO] vcontact3.profiles: Unique genes identified in user data: 9
[INFO] vcontact3.profiles: Unique genomes identified in user data: 1
[INFO] vcontact3.profiles: After merging user + reference data: 124496 unique genes, 7707 unique genomes
[INFO] vcontact3.profiles: Processing identity: 0.3
[INFO] vcontact3.profiles: Merging and updating PCs associated with this identity
[INFO] vcontact3.profiles: Rectifying new cluster IDs during merger of reference database and user sequences
[INFO] vcontact3.profiles: There were 0 clusters added, 9 clusters extended from merge
Traceback (most recent call last):
  File ".../pandas/core/internals/blocks.py", line 1115, in setitem
    casted = np_can_hold_element(values.dtype, value)
  File ".../pandas/core/dtypes/cast.py", line 1792, in np_can_hold_element
    raise LossySetitemError
pandas.errors.LossySetitemError
```

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  ```
File "/usr/local/bin/vcontact3", line 8, in <module>
    sys.exit(main())
  File ".../vcontact3/cli.py", line 43, in run_vcontact3_entry
    run_vcontact3(**kwargs)
  File ".../vcontact3/vcontact3.py", line 213, in run_vcontact3
    profile = profiles.Profiles(
  File ".../vcontact3/profiles.py", line 377, in __init__
    merged_overview_df.loc[list(ref_shared_with_user), 'shares_clusters'] = 'withUser'
  File ".../pandas/core/internals/blocks.py", line 468, in coerce_to_target_dtype
    raise TypeError(f"Invalid value '{other}' for dtype '{self.values.dtype}'")
TypeError: Invalid value 'withUser' for dtype 'float64'

```
Gets through database sourcing, gene prediction, and the actual profile merge (7707 genomes merged successfully) — dies at profiles.py:377, assigning the string 'withUser' into merged_overview_df['shares_clusters'], a column pandas has inferred as float64 (presumably because it starts out all-NaN). Under pandas>=2.1.1 (pinned in vcontact3's own bioconda recipe), that's a hard TypeError instead of the silent upcast-to-object that older pandas did.

I first ran this with default --db-domain (prokaryotes, since the test genome is SARS-CoV-2) and got the identical crash at the identical line — just 'ref only' instead of 'withUser'. So it's not a domain-mismatch artifact; it's this exact line failing regardless of which branch of the merge populates shares_clusters first.

Given that, three ways forward:

1. Report + patch upstream. I'll open an issue on vcontact3's Bitbucket repo with this repro, and take a look at whether a fix is straightforward (e.g. explicitly setting dtype=object on merged_overview_df['shares_clusters'] at creation, so it can safely hold both floats and strings).
2. pinning pandas<2.1.1 in vcontact3's bioconda recipe would unblock this without waiting on a code fix — a workaround rather than a real fix, but it'd let container rebuilds pick up a working pandas version immediately.
3. Ship this PR stub-only in the meantime. Since neither of the above is fixable from the module side, I'd suggest merging with the stub test as-is now, and following up with the real chained test once there's an upstream fix (either the code patch or the recipe pin) to test against. Keeps this PR unblocked without hiding the issue — it's fully documented here.

SPPearce:
Is the issue that the test dataset is missing something? Does the tool work on a bigger dataset?
You definitely should file an issue upstream I think.

PR #12207, opened 2026-06-30
gpu: caalm enable gpu mode

vagkaratzas:
Based on the guidelines [here](https://nf-co.re/docs/developing/components/gpu-modules)

For `conda-forge::mkl=2024.* # pins the MKL ABI for FAISS (mkl 2026 renames libmkl_intel_lp64.so.2 -> .so.3, breaking faiss import)` -> needed to be pinned because latest pulled changes in the conda env are breaking the faiss import due to a rename.

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:


pinin4fjords (commented):
Nice to see the GPU guidelines actually get used end to end here - dual-container ternary, `task.accelerator` detection, `versions_cuda` topic, `gpu`-tagged tests are all present and CI is green.

One real issue: the dual-container split doesn't shrink the CPU path here, which is the whole point of doing dual containers. Compare against `ribodetector` (the guidelines' own reference module): its CPU container is 0.57 GB vs 6.0 GB for GPU, a ~10x gap. Here it's the other way round:
- caalm CPU container: **6.59 GB** compressed
- caalm GPU container: **5.27 GB** compressed

Cause: `environment.yml` pip-installs plain `torch==2.6.0`, and PyPI's default Linux torch wheel bundles the full CUDA runtime (nvidia-cublas, nvidia-cudnn, etc) whether or not a GPU is used, and one layer in that image alone is 6.5 GB. `ribodetector` avoids this entirely by getting torch through conda (`bioconda::ribodetector` pulls it in on the CPU side, `conda-forge::pytorch-gpu` on the GPU side), never through pip. You can see the symptom in your own snapshot too: `main.nf.test.snap` (the plain CPU test) reports `"versions_cuda": "12.4"`, a real CUDA version, when the guidelines say that should read `no CUDA available` on the non-GPU path.

Left a suggestion below switching the CPU side to `conda-forge::pytorch` (mirroring your own `conda-forge::pytorch-gpu` on the GPU side, and the `ribodetector` precedent). That needs three follow-ups I can't do via suggestion:
1. delete the now-redundant `torch==2.6.0` line from the `pip:` block a couple lines below
2. rebuild the CPU container with Wave against the updated `environment.yml` and swap the new hash into `main.nf`
3. re-run the CPU test and confirm `versions_cuda` flips to `no CUDA available`, and check the image actually shrunk

Small aside, not blocking: adding the mkl pin to `environment.yml` technically deviates from "the CPU environment.yml MUST remain unchanged" in the spec. Looks like a legit, unrelated bug fix rather than an accident, might be worth a one-line callout in the PR description so it doesn't read as scope creep.

modules/nf-core/caalm/caalm/environment.yml
```diff
@@ -5,6 +5,7 @@ channels:
   - bioconda
 dependencies:
   - conda-forge::faiss-cpu=1.10.0
+  - conda-forge::mkl=2024.*       # pins the MKL ABI for FAISS (mkl 2026 renames libmkl_intel_lp64.so.2 -> .so.3, breaking faiss import)
```
pinin4fjords: See the size comparison against `ribodetector` in the top-level comment - this alone won't fix it, you'd still need to drop `torch==2.6.0` from the `pip:` block below and rebuild the container. `conda-forge::pytorch` is the CPU-only build, same approach `ribodetector` uses for its own CPU container.

```suggestion
  - conda-forge::mkl=2024.*       # pins the MKL ABI for FAISS (mkl 2026 renames libmkl_intel_lp64.so.2 -> .so.3, breaking faiss import)
  - conda-forge::pytorch=2.6.0
```

modules/nf-core/caalm/caalm/tests/nextflow.gpu.config
```diff
@@ -0,0 +1,7 @@
+process {
+    withName: CAALM_CAALM {
+        ext.args = params.module_args
+        accelerator = 1
```
pinin4fjords, line 4: The GPU guidelines flag Singularity GPU concurrency as a deadlock risk when multiple GPU tasks share one GPU (e.g. CI). Not hit here since this passed, but cheap insurance.

```suggestion
        accelerator = 1
        maxForks    = 1
```

modules/nf-core/caalm/caalm/environment.yml
vagkaratzas: Nice, singularity image 147MB now

vagkaratzas:
Thank you @pinin4fjords! Created the updated CPU containers through wave.

pinin4fjords (approved):
Verified the fix. New CPU container is 0.55 GB (was 6.59 GB), GPU stays at 5.28 GB, so the size relationship is now correct and matches ribodetector's CPU/GPU ratio closely. `versions_cuda` on the CPU path now reports `no CUDA available` as expected. `maxForks = 1` added too. All CI green. Looks good, thanks for the quick turnaround.

PR #12218, opened 2026-07-01
update telescope version to 1.0.4

hanalysis:
this update just updates the version of telescope so that is includes the argument --stranded_mode, no other changes.


modules/nf-core/telescope/assign/main.nf
```diff
@@ -3,9 +3,7 @@ process TELESCOPE_ASSIGN {
     label 'process_single'
 
     conda "${moduleDir}/environment.yml"
-    container "${ workflow.containerEngine in ['singularity', 'apptainer'] && !task.ext.singularity_pull_docker_container ?
-        'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/ea/eaa68e02f66957e51ecba656dfeaa8576bada172780d95592105bc316a01a65c/data':
-        'community.wave.seqera.io/library/telescope:1.0.3_fix--d176f12022b914cf' }"
+    container 'community.wave.seqera.io/library/telescope:1.0.4--a061d59a7493d996'
```
mashehu: you need the ternary expression with the singularity image as well.

hanalysis: ah thank you ! Hopefully this is correct now.

hanalysis:
I think everything is resolved now, should be ready to merge, thanks for your help @mashehu

mashehu:
something is off, because a change in the version should mean you would also need to update the snapshot 🤔

mashehu:
I am checking the container now, what the version output says there.

hanalysis:
Do I just need to run nf-core modules test on it to generate a new snapshot? If so I forgot to do that first, that's my fault! Can do now

mashehu:
no the tests should fail, if you need to update the snapshots.

mashehu:
okay, is it possible that they just updated the bioconda recipe and the version there (instead of the build number)? because the repo still shows 1.0.3 and the tool also reports that version (that's why the snapshot didn't need to be changed.

hanalysis:
Yes! It was me who updated the recipe & I didn't update the build number, sorry, didn't realise that was the reason

mashehu:
since you own the fork with the fix. do you mind bumping the version in your fork https://github.com/hanalysis/telescope/blob/main/telescope/_version.py and then making a new bioconda build?

mashehu (approved):

PR #12228, opened 2026-07-02
gstama_polyacleanup: Fixed regression - Fixing output glob

sguizard:
The globs modification from the last update breaks the content of the output channels. 
Without the `_tama` prefix, `*.fa.gz` grab `_tama.fa.gz` AND `_tails.fa.gz`.


modules/nf-core/gstama/polyacleanup/main.nf
```diff
@@ -11,7 +11,7 @@ process GSTAMA_POLYACLEANUP {
     tuple val(meta), path(fasta)
 
     output:
-    tuple val(meta), path("*.fa.gz")                   , emit: fasta
+    tuple val(meta), path("*_seq.fa.gz")               , emit: fasta
     tuple val(meta), path("*_polya_flnc_report.txt.gz"), emit: report
     tuple val(meta), path("*_tails.fa.gz")             , emit: tails
```
atrigila: What if you do:
```
output:
tuple val(meta), path("${prefix}.fa.gz")                   , emit: fasta
tuple val(meta), path("${prefix}_polya_flnc_report.txt.gz"), emit: report
tuple val(meta), path("${prefix}_tails.fa.gz")             , emit: tails
```
This way, you do not rename the tool output at all. Just make the fasta output specific.

modules/nf-core/gstama/polyacleanup/main.nf
```diff
@@ -28,15 +28,18 @@ process GSTAMA_POLYACLEANUP {
         -f $fasta \\
         -p ${prefix} \\
         $args
-    gzip ${prefix}.fa
+
+    mv ${prefix}.fa ${prefix}_seq.fa
```
atrigila: Here you could remove the mv this and then do:

```
gzip ${prefix}.fa
gzip ${prefix}_polya_flnc_report.txt
gzip ${prefix}_tails.fa
```

modules/nf-core/gstama/polyacleanup/main.nf
sguizard: I should have correct my first message once I understood the real problem. 
I forgot, my apologies for that.
Let's rewind to the original problem and fold back to the solution. 

The last update of the module changed the outputs to make them more generic by removing the `_tama` prefix. This was a good idea, but it introduced a regression. 

For your information, the `_tama` prefix is an artefact from when I developed the isoseq pipeline. Maybe not the smartest idea at the time, but I learned a lot since then. 

So what was the idea behind those three channels and why the last update introduced a regression? 
In order to not mixing two kind of sequence information generated by the script, the Full Length Non Chimeric reads one side and the poly A on the other side, I have set up a `fasta` output channel and a `tails` output channel. 
The tails channel I suppose to grab the polyA sequences and the `fasta` is supose to grab FLNC sequences. 
In its original code, the channel globs included a `_tama` prefix. The reason of this is simply because the nf-core/isoseq pipeline generated files have this prefix. Again, maybe not the wisest choice, but I was young and nf-core naive. 

Now, fast forward to the last module update. In order, make the module code more generic, `_tama` has been removed. A wise choice, but with unexpected consequences. Changing the `fasta` channel glob from `*_tama.fa.gz` to `*.fa.gz` change which files are grabbed by the channel. Instead of collecting only the FLNCs, it grabs the FLNCs and the polyA tails. Indeed, those two files shares the same extension.

Even if this small modification has nothing dramatic, it has the annoying effect to mix the two information into one channel. After, I can understand that the channel name can be confusing.`fasta` can be changed to `seq` to avoid this confusion. 

To fix this regression, this simplest solution is to make FLNC fasta file name unique by renaming it and adapt  the glob to this renamed version. This is the point of this patch. 

Writing this made me realize how the `fasta` channel name can be confusing. I'm going to push a commit to rename it into `seq`. It will be more inline with its purpose.

sguizard:
@atrigila Hi, can you have a second look when you have time please?

modules/nf-core/gstama/polyacleanup/main.nf
atrigila: Thanks for explaining the original regression. I agree that *.fa.gz incorrectly captures both files. 

However, I do not think renaming the generated file or the emit is necessary. 

`path("${prefix}.fa.gz")` would match only the FLNC output and exclude `${prefix}_tails.fa.gz`, while preserving the existing fasta channel. 

Is there a case where the exact prefix-based pattern would not work?

SPPearce: I agree with @atrigila , I think you should be using the prefix explicitly.

SPPearce (approved):

modules/nf-core/gstama/polyacleanup/main.nf
sguizard: An other old artefact from past me. 
I applied your suggestions. 
Thanks for the help.

atrigila (approved):
Thanks for addressing all the comments :)

sguizard:
> Thanks for addressing all the comments :)

You're welcome! 
The * pattern was so printed in brain I didn't see you added the prefix at first sight. 
Thanks for the reviewing.

PR #12230, opened 2026-07-02
fix(dotseq/dotseq): make plotDOT() heatmap actually render

pinin4fjords:
## Summary

`DOTSEQ_DOTSEQ`'s `*.heatmap.png` output has never actually been produced, for any annotation:

- The template calls `plotDOT()` with `id_mapping = FALSE`. DOTSeq's `plot_volcano()` normalises `isFALSE(id_mapping)` to `NULL`, but the heatmap's internal `cistronic_data()` helper has no such guard, so it evaluates `FALSE$ensembl_gene_id` and throws `$ operator is invalid for atomic vectors`. DOTSeq catches this internally and just logs a message, so no R error ever reaches the module's own `tryCatch`, and since nothing gets drawn between `png()` and `dev.off()`, no file is written at all - not even an empty one. The module's `safe_plot_dot()` only checked for a thrown exception, so it reported success regardless, silently swallowing the missing plot (the Nextflow output is `optional: true`).
- Fix: pass `id_mapping = NULL` (handled correctly by every `plotDOT()` plot type) and check for a non-empty output file rather than the absence of an exception. This also means the existing uORF/dORF heatmap fallback logic actually runs for the first time.

Separately, `cistronic_data()` hardcodes the literal string `"mORF"` for a gene's main ORF with no way to configure it, so an annotation using a different main-ORF label (e.g. `canonical_cds`, as nf-core/riboseq's `orf_class` column does) still produces no heatmap once the above is fixed, because there are no `"mORF"`-labelled rows to pair against. Added `--orf_type_main_value` (default `mORF`) so the template remaps whatever label the caller uses to `mORF` before building the annotation `GRanges`. `meta.yml` documents the `mORF`/`uORF`/`dORF` vocabulary constraint.

## Test plan

- [x] `nf-test test modules/nf-core/dotseq/dotseq/tests/main.nf.test --profile +docker --update-snapshot` - all 4 tests pass, snapshot updated
- [x] Added `heatmap_plot.size() > 0` to the existing test - now produces a real heatmap PNG that never existed before
- [x] Added a new test ("non-mORF main label") against a fixture on `pinin4fjords/test-datasets` (`add-dotseq-testdata` branch) with `orf_type` relabelled `mORF` → `canonical_cds`, using `--orf_type_main_value canonical_cds` - produces a heatmap
- [x] `nf-core modules lint dotseq/dotseq` - clean (2 pre-existing, unrelated container-registry-connectivity warnings only)

### DIFF
--- a/modules/nf-core/dotseq/dotseq/meta.yml
+++ b/modules/nf-core/dotseq/dotseq/meta.yml
@@ -65,12 +65,19 @@ input:
           Per-ORF annotation table (one row per ORF). Required columns:
           `orf_id` (matches the count matrix) and `gene_id` (parent gene;
           DOTSeq's DOU model groups child ORFs by gene). Optional columns:
-          `orf_type` (mORF / uORF / dORF; used by plotDOT()'s heatmap and
-          defaults to mORF when absent) and `chrom`, `start`, `end`,
-          `strand` (used only for downstream inspection - dummy ranges
-          are generated when absent because DOTSeq's fit does not depend
-          on genomic coordinates). Column names can be overridden via
-          task.ext.args (`--gene_id_col`, `--orf_type_col`, etc.).
+          `orf_type` (used by plotDOT()'s heatmap to pair a gene's main ORF
+          against its short ORFs; defaults to the main-ORF value when absent)
+          and `chrom`, `start`, `end`, `strand` (used only for downstream
+          inspection - dummy ranges are generated when absent because
+          DOTSeq's fit does not depend on genomic coordinates). Column names
+          can be overridden via task.ext.args (`--gene_id_col`,
+          `--orf_type_col`, etc.). plotDOT()'s heatmap hardcodes the literal
+          label `mORF` for a gene's main ORF and only accepts `uORF` or
+          `dORF` for the short-ORF bucket it is paired against; whichever
+          value in `orf_type` denotes the main ORF is remapped to `mORF`
+          internally (override the source value via `--orf_type_main_value`,
+          default `mORF`), but short-ORF rows must already be labelled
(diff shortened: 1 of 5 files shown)


FelixKrueger:
**Possible edge case in the `--orf_type_main_value` remap** ([`templates/dotseq.R:254-256`](https://github.com/nf-core/modules/blob/fix/dotseq-nonmorf-heatmap/modules/nf-core/dotseq/dotseq/templates/dotseq.R#L254-L256)):

```r
orf_type <- if (opt$orf_type_col %in% names(ann)) as.character(ann[[opt$orf_type_col]]) else rep(opt$orf_type_main_value, nrow(ann))
orf_type[is.na(orf_type) | !nzchar(trimws(orf_type))] <- opt$orf_type_main_value
orf_type[orf_type == opt$orf_type_main_value] <- "mORF"
```

The remap only rewrites rows matching `--orf_type_main_value`; it never excludes rows that are already the literal string `"mORF"`. If a user sets `--orf_type_main_value canonical_cds` but the annotation still has some rows pre-labelled `"mORF"` (stale data, or a coincidentally-named non-main biotype), those rows pass through unchanged on line 3 and get grouped into the heatmap's main-ORF bucket alongside the real `canonical_cds` rows — even though they were never intended as main ORFs.

Probably a narrow case in practice (would need a mixed-vocabulary annotation), but worth a guard — e.g. relabel any pre-existing `"mORF"` rows to a sentinel before the remap when `orf_type_main_value != "mORF"` — so the hardcoded target string can't collide with real input data.

FelixKrueger (approved):
Maybe look at the edge case, other than that the bugfix looks good

pinin4fjords:
Thanks @FelixKrueger ! All sorted.

PR #12256, opened 2026-07-07
Add meta maps to kallistobustools ref and count modules

nictru:
It's always annoying in pipelines if some modules expect reference files with meta maps, and other ones without. Leads to a lot of `map` operators to ensure the right channel shape. This PR adds meta map inputs to kallistobustools modules, that were previously lacking them


modules/nf-core/kallistobustools/ref/tests/main.nf.test
```diff
@@ -14,18 +14,24 @@ nextflow_process {
...
             }
         }
 
         then {
+            assert process.success
             assertAll(
-                { assert process.success },
                 { assert snapshot(sanitizeOutput(process.out, unstableKeys: ["index"])).match() },
-                { assert file(process.out.index.get(0)).exists() }
+                { assert file(process.out.index.get(0).get(1)).exists() }
```
LouisLeNezet: I don't think this line is necessary as it is present in the snapshot.

modules/nf-core/kallistobustools/ref/tests/main.nf.test
```diff
@@ -35,18 +41,24 @@ nextflow_process {
...
```
LouisLeNezet: Same

modules/nf-core/kallistobustools/ref/tests/main.nf.test
```diff
@@ -80,18 +98,24 @@ nextflow_process {
...
```
LouisLeNezet: Same

modules/nf-core/kallistobustools/ref/main.nf
```diff
@@ -8,17 +8,17 @@ process KALLISTOBUSTOOLS_REF {
...
-    path "cdna.fa"        , emit: cdna
-    path "intron.fa"      , optional:true, emit: intron
-    path "cdna_t2c.txt"   , optional:true, emit: cdna_t2c
-    path "intron_t2c.txt" , optional:true, emit: intron_t2c
+    tuple val(meta), path("kb_ref_out.idx") , emit: index
+    path "t2g.txt"                          , emit: t2g
+    path "cdna.fa"                          , emit: cdna
+    path "intron.fa"                        , optional:true, emit: intron
+    path "cdna_t2c.txt"                     , optional:true, emit: cdna_t2c
+    path "intron_t2c.txt"                   , optional:true, emit: intron_t2c
```
LouisLeNezet: Could you put the optional after the emit and align them ? It's cleaner this way.
```suggestion
    path "intron.fa"                        , emit: intron, optional:true
    path "cdna_t2c.txt"                     , emit: cdna_t2c, optional:true
    path "intron_t2c.txt"                  , emit: intron_t2c , optional:true
```

modules/nf-core/kallistobustools/ref/main.nf
```diff
@@ -8,17 +8,17 @@ process KALLISTOBUSTOOLS_REF {
...
     tuple val("${task.process}"), val('kallistobustools'), eval("kb --version 2>&1 | grep -Eo '[0-9]+\\.[0-9]+\\.[0-9]+' | head -n1"), emit: versions_kallistobustools, topic: versions
```
LouisLeNezet: This should be simplified as 
```suggestion
    tuple val("${task.process}"), val('kallistobustools'), eval("kb --version 2>&1 | sed -n 's/kb_python //p'"), emit: versions_kallistobustools, topic: versions
```

modules/nf-core/kallistobustools/count/main.nf
LouisLeNezet: Same as below

LouisLeNezet (commented):
Thanks for cleaning this up.
I just have a few comments to clean it a bit more and align it last practices.

LouisLeNezet (approved):
LGTM

PR #12266, opened 2026-07-07
Swap trycycler to topic

SPPearce:


modules/nf-core/trycycler/subsample/environment.yml
SPPearce, line 1: This is already pinned in the other module.

atrigila (approved):
Looks good

PR #12268, opened 2026-07-07
Remove versions block from trgt and fix tests

SPPearce:
Two of these submodules had a topic emit but still generated the versions.yml. They also were failing on tests due to upstream changes to samtools/sort and bcftools/sort


modules/nf-core/trgt/genotype/tests/main.nf.test
```diff
@@ -118,16 +124,17 @@ nextflow_process {
...
+                { assert snapshot(sanitizeOutput(process.out)).match() }
             )
         }
 
+
     }
 
     test("homo sapiens - [bam,bai,[]], [fa,fai], bed - stub") {
-
+        options "-stub"
```
SPPearce, line 137: Got to love stub tests without the `stub` option.

atrigila (approved):

PR #12269, opened 2026-07-07
Add `rclone check` and `checksum`

atrigila:
I created two new modules for the `RCLONE` suite: `rclone check` and `rclone checksum`. I used the new functionality in nf-core/tools (dev) to create the container for `rclone` (which includes conda-lock files) from `environment.yml` files. 

Notes:
- The software would produce an error if, for example, the files are not available in destination. As we find this is something informative that we want to get a report on, I added the `|| true`, so it won't fail, but we will get the files with the report. 

Related to: https://github.com/nf-core/datasync/issues/44


modules/nf-core/rclone/check/tests/main.nf.test
```diff
@@ -0,0 +1,85 @@
...
+                input[0] = [[id:'test'], file(params.modules_testdata_base_path + 'genomics/sarscov2/illumina/fastq/test_1.fastq.gz', checkIfExists: true)]
+                input[1] = [[id:'test_dst'], file(params.modules_testdata_base_path + 'genomics/sarscov2/illumina/fastq/test_1.fastq.gz', checkIfExists: true)]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot([match: process.out.match, versions_rclone: process.out.versions_rclone]).match() }
```
SPPearce, line 25: why not sanitizeOutput(process.out)?

modules/nf-core/rclone/checksum/main.nf
```diff
@@ -0,0 +1,59 @@
...
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine in ['singularity', 'apptainer'] && !task.ext.singularity_pull_docker_container
+            ? 'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/5d/5dfd28fd0090c69f57c9bd93ea3235d8df194e8f269b5cb3027b6b59bff567d5/data'
+            : 'community.wave.seqera.io/library/rclone:1.74.3--2ef33c5b9132aa97' }"
+
+    input:
+    tuple val(meta), path(sumfile)
+    val hash
+    tuple val(meta2), path(destination, stageAs: 'destination/*')
```
SPPearce: I would swap these over
```suggestion
    tuple val(meta2), path(destination, stageAs: 'destination/*')
    val hash
```

SPPearce (commented):
Why are you doing something unusual in the snapshots?

modules/nf-core/rclone/check/tests/main.nf.test
atrigila: I should have clarified this in the PR description. I tried that. If you snapshot all results it will produce empty files for the optional output processes (and that would make the linter fail). This is the "cleanest" I could come up with but I am open to hear if you have any other suggestion.

atrigila:
@SPPearce Do you have any other idea on how to approach this? Thanks!

modules/nf-core/rclone/check/tests/main.nf.test
SPPearce: sanitizeOutput(process.out, unstableKeys: ["combined", "differ", "missing_on_dst", "missing_on_src","error"]) ?

PR #523, opened 2021-05-25
Pairtools select

jianhong:
## PR checklist

- [ x ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ x ] If necessary, include test data in your PR.
- [ x ] Remove all TODO statements.
- [ x ] Emit the `<SOFTWARE>.version.txt` file.
- [ x ] Follow the naming conventions.
- [ x ] Follow the parameters requirements.
- [ x ] Follow the input/output options guidelines.
- [ x ] Add a resource `label`
- [ x ] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
    - [ x ] `PROFILE=docker pytest --tag <MODULE> --symlink --keep-workflow-wd`
    - [ ] `PROFILE=singularity pytest --tag <MODULE> --symlink --keep-workflow-wd`
    - [ x ] `PROFILE=conda pytest --tag <MODULE> --symlink --keep-workflow-wd`

PR #12283, opened 2026-07-08
implemented readgroup addition in fastq_align_chromap, since chromap has no option to do it

Kevin-Brockers:
## PR checklist

Closes #XXX 

  - For modules:
  - For subworkflows:

### DIFF
--- a/subworkflows/nf-core/fastq_align_chromap/main.nf
+++ b/subworkflows/nf-core/fastq_align_chromap/main.nf
@@ -2,8 +2,9 @@
  * Map reads, sort, index BAM file and run samtools stats, flagstat and idxstats
  */
 
-include { CHROMAP_CHROMAP         } from '../../../modules/nf-core/chromap/chromap/main'
-include { BAM_SORT_STATS_SAMTOOLS } from '../bam_sort_stats_samtools/main'
+include { CHROMAP_CHROMAP               } from '../../../modules/nf-core/chromap/chromap/main'
+include { PICARD_ADDORREPLACEREADGROUPS } from '../../../modules/nf-core/picard/addorreplacereadgroups/main'
+include { BAM_SORT_STATS_SAMTOOLS       } from '../bam_sort_stats_samtools/main'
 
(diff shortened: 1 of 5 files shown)


subworkflows/nf-core/fastq_align_chromap/main.nf
```diff
@@ -14,9 +15,11 @@ workflow FASTQ_ALIGN_CHROMAP {
     ch_whitelist // channel (optional):  [ whitelist ]
     ch_chr_order // channel (optional):  [ chr_order ]
     ch_pairs_chr_order // channel (optional):  [ pairs_chr_order ]
+    update_readgroups // boolean (mandatory): true or false
```
JoseEspinosa: Does this not need to be provided when calling the subworkflow now?

JoseEspinosa (changes requested):
Just a small question but otherwise LGTM!

JoseEspinosa:
Ah, I see this is done in #523 I guess

subworkflows/nf-core/fastq_align_chromap/main.nf
Kevin-Brockers: Yes indeed. Though to have it explicit since bwa or bowtie2 can add readgroups while aligning.

But if you prefer, I can switch to a more flexible approach? Disable by default and control via args.

JoseEspinosa: But wait, this has to be to be changed in the modules repo, now it isn't, is it?

Kevin-Brockers: yes, this pr is to update the subworkflow in the modules repo. I will then update the subworkflow in the chipseq later.

Don't get confused with: https://github.com/nf-core/chipseq/pull/523

Kevin-Brockers: Set it as an optional parameter and a default to false

JoseEspinosa (approved):
Thanks! LGTM 🚀

PR #12285, opened 2026-07-08
Swap taxpasta to topic channels

SPPearce:


modules/nf-core/taxpasta/merge/environment.yml
atrigila, line 1: Is this change to solve the `Needs to be sorted (differences conda vs container)` or does it have any other purpose?

SPPearce: nf-core modules lint automatically sorts the dependencies

SPPearce: (i.e. it wasn't an intentional change, it just did it without me asking)

atrigila (approved):
LGTM

PR #12325, opened 2026-07-08
Bump custom/gtffilter to Python 3.14.5 with Wave containers

nictru:
Currently this module does not work, because it declares python 3.9 as a dependency, but recently the ruff config update introduced syntax that require python 3.10 or later. Not sure how the changes made it through the CI. This PR upgrades to py3.14.


mashehu (approved):
oh, I guess the python script changes didn't trigger anything in nf-test. weird. thanks for the fix. will check the other changed modules

mashehu:
all the other ones are fine

PR #12329, opened 2026-07-09
Fix strict syntax in utils_* subworkflows

itrujnara:
Removed names of single outputs in utils_* subworkflows to comply with Nextflow strict syntax. Edited metas to match. This should remove strict syntax warnings in pipelines.

## PR checklist
  - For subworkflows:


subworkflows/nf-core/utils_nextflow_pipeline/meta.yml
```diff
@@ -25,7 +25,7 @@ input:
       description: |
         Check if the conda channel priority is correct.
 output:
-  - dummy_emit:
+  - "true":
       type: boolean
       description: |
         Dummy emit to make nf-core subworkflows lint happy
```
mashehu, line 31: does `nf-core s lint` still complain about this?

itrujnara: It does not as long as it is passed as a string, but it must be there (zero outputs is an illegal pattern)

subworkflows/nf-core/utils_nfcore_pipeline/main.nf
```diff
@@ -17,7 +17,7 @@ workflow UTILS_NFCORE_PIPELINE {
     checkProfileProvided(nextflow_cli_args)
 
     emit:
-    valid_config = valid_config
+    valid_config
```
mashehu, line 20: I thought we needed this for strict syntax https://github.com/nf-core/modules/commit/a3fb7351b1fdb2b1de282b765816bbea190e86a8

itrujnara: I can't find the documentation for this, but `nextflow lint` shows a warning if a workflow has a single output and it is named. I have sent the warning message on Slack. I agree with Maxime that this rule should not be there, but that's the way Ben designed it.

mashehu: but in the current way i get the following with nextflow 26.04.4:
```
Warn  subworkflows/nf-core/utils_nfcore_pipeline/main.nf:16:5: Variable was declared but not used
│  16 |     valid_config = checkConfigProvided()
╰     |     ^^^^^^^^^^^^

Warn  subworkflows/nf-core/utils_nfcore_pipeline/main.nf:20:5: Emit name should be omitted when there is only one emit
│  20 |     valid_config
╰     |     ^^^^^^^^^^^^
```
that's why I had the previous version

itrujnara: I must have tested with a wrong version, with 26.04.4 it indeed returns these warnings. According to [the docs](https://docs.seqera.io/nextflow/workflow#takes-and-emits) the version without the extra name is correct. I am puzzled as to how I should write the emit section to avoid complaints from the language server.

SPPearce:
Has this been resolved yet?

PR #12333, opened 2026-07-09
12306 adding request db module to the nf coregapseq

### LINKED ISSUE #12306: Adding request db module to the nf-core/gapseq

Gapseq automatically downloads the database when running gapseq the first time after installation. But since nextflow initializes several hundreds+ reconstructions at once, I need to specificaly load it once and point other modules to the existing db

_No response_

_No response_

### Additional context

_No response_

brovolia:
New submodule should provide a one-off download of db for gapseq, that the parallel processes can share it afterwards

### DIFF
--- a/modules/nf-core/gapseq/doall/main.nf
+++ b/modules/nf-core/gapseq/doall/main.nf
@@ -9,6 +9,7 @@ process GAPSEQ_DOALL {
 
     input:
     tuple val(meta), path(fasta), path(medium)
+    path(db)
 
     output:
     tuple val(meta), path("*.RDS")  , emit: model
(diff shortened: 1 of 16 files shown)


dialvarezs (commented):
Is it possible to integrate the output form the new `GAPSEQ_REQUESTDB` module into the other modules instead of just leaving the parameter empty?
That would help to test the wiring.

brovolia:
I did this in the stub tests, so the wiring is validated with zero network cost. The real-data tests keep `input[1] = []`, so both the "shared pre-downloaded DB" and "self-download" paths are covered.

dialvarezs (approved):
Looks good to me then!

PR #12337, opened 2026-07-09
Add rclone copy

antoniasaracco:
I added a new module for the RCLONE suite: rclone copy.

This module uses rclone copy to transfer files from the source path to the destination path, while allowing extra arguments to be passed through task.ext.args. This makes it possible to configure options such as logging, stats, certificate handling, chunk size, HTTP URL support and dry-run behavior from the pipeline config.

Notes:

For the nf-test, I am using the same input and output location together with --dry-run. The module would not have write permissions to perform the actual copy in the test environment, so running it this way allows us to mock how the command would be executed without requiring remote write access.

This is related to https://github.com/nf-core/datasync/issues/45


modules/nf-core/rclone/copy/tests/main.nf.test
```diff
@@ -0,0 +1,66 @@
...
+                input[1] = []
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
+            )
```
SPPearce, line 58:
```suggestion
            assert process.success
            assertAll(
                { assert snapshot(sanitizeOutput(process.out)).match() }
            )
```

modules/nf-core/rclone/copy/tests/main.nf.test
```diff
@@ -0,0 +1,66 @@
...
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert process.out.log[0][1].endsWith("rclone-copy.log") },
+                { assert snapshot(process.out.versions_rclone).match("versions") }
+            )
```
SPPearce, line 33:
```suggestion
            assert process.success
            assertAll(
                { assert snapshot(sanitizeOutput(process.out), unstableKeys: ["log"]).match() }
            )
```

modules/nf-core/rclone/copy/main.nf
```diff
@@ -0,0 +1,40 @@
...
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def configArg = rclone_config ? "--config '${rclone_config}'" : ''
+    def transfers = task.cpus
+    def checkers = task.cpus
+
+    """
+    rclone ${configArg} copy ${args} \\
```
SPPearce:
```suggestion
    rclone ${configArg} copy \\
        ${args} \\
```

modules/nf-core/rclone/copy/main.nf
```diff
@@ -0,0 +1,40 @@
...
+    tuple val(meta), path("rclone-copy.log"), emit: log
+    tuple val("${task.process}"), val('rclone'), eval("rclone --version | sed -n '1s/^rclone v//p'"), topic: versions, emit: versions_rclone
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def configArg = rclone_config ? "--config '${rclone_config}'" : ''
+    def transfers = task.cpus
+    def checkers = task.cpus
```
SPPearce: Are these supposed to both be the same?

modules/nf-core/rclone/copy/main.nf
```diff
@@ -0,0 +1,40 @@
+process RCLONE_COPY {
+
```
SPPearce:
```suggestion
```

modules/nf-core/rclone/copy/tests/main.nf.test
```diff
@@ -0,0 +1,66 @@
...
+        config "./nextflow.config"
+
+        when {
+            process {
+                """
+                def test_file_path = params.modules_testdata_base_path.replaceFirst(/^https?:\\/\\/[^\\/]+\\//, '') + 'genomics/homo_sapiens/illumina/gvcf/test.genome.vcf.gz'
+
+                input[0] = [
+                    [ id:'test' ],
+                    ':http:' + test_file_path,
```
SPPearce: This is a little ugly, can it not directly take the https: path?

antoniasaracco: Thanks for the comment! This is what I have found to make it work with https but please let me know if you have any other solution.

image
https://rclone.org/http/

SPPearce: Fine. 
One thought would be whether you could add the `http:` inside the module...

modules/nf-core/rclone/copy/main.nf
antoniasaracco: you are correct. I've modified now to respect the ratio between transfers and checkers set by default in the tool.
image
https://rclone.org/flags/

modules/nf-core/rclone/copy/tests/main.nf.test
antoniasaracco: okay, I thought it would be best for each pipeline to handle this outside but makes sense. What do you think now?

modules/nf-core/rclone/copy/tests/nextflow.config
```diff
@@ -0,0 +1,8 @@
+process {
+    withName: 'RCLONE_COPY' {
+        ext.args = {
+            def base_url = (params.modules_testdata_base_path =~ /^(https?:\/\/[^\/]+)/)[0][1]
+            "--dry-run --http-url '${base_url}' --no-check-certificate"
```
SPPearce: Will this not be setting --http-url twice now?

modules/nf-core/rclone/copy/tests/main.nf.test
SPPearce: I like it, means that people don't have to do that bit manually, we have taken care of it for them.

modules/nf-core/rclone/copy/tests/nextflow.config
antoniasaracco: you are right, thanks! solved now

SPPearce (approved):

PR #12344, opened 2026-07-10
Bump bismark to 3.0.0

pinin4fjords:
## Summary
- Bumps Bismark from 0.25.1 to 3.0.0 across all 7 bismark modules (align, coverage2cytosine, deduplicate, genomepreparation, methylationextractor, report, summary). 3.0.0 is Bismark's Rust rewrite, confirmed available on bioconda (`bioconda::bismark=3.0.0`).
- Drops the separate `bioconda::samtools` dependency from `deduplicate` and `methylationextractor`'s `environment.yml`. The 3.0.0 bioconda recipe no longer lists samtools as a runtime requirement, since the Rust rewrite does its own BAM/SAM/CRAM I/O internally; the module scripts never invoked samtools directly, so this was only ever a transitive dependency of the old Perl Bismark.
- Fixes the version-string `eval` command in all 7 modules (`main.nf`) and their `meta.yml` docs: the old `bismark -v 2>&1 | sed -n 's/^.*Bismark Version: v//p'` matched Perl Bismark's output format and returned an empty string against 3.0.0. Replaced with `bismark --version 2>&1 | grep -Eo '[0-9]+\.[0-9]+\.[0-9]+'`, which works against the new output format (and is consistent with the pattern used elsewhere in this repo).
- Rebuilds Docker/Singularity containers via Wave for the new environment.yml.
- Regenerates the version string in each module's `tests/main.nf.test.snap`.

Verified that all other test-suite content (BAM/report/coverage/methylation-call md5s) is unchanged between 0.25.1 and 3.0.0 - the only diff in every snapshot is the version string, matching Bismark's changelog claim that 3.0.0 output is byte-identical to 0.25.1 on default settings.

## Test plan
- [x] `nf-test test modules/nf-core/bismark/*/tests/main.nf.test --profile docker` - all 17 tests pass across all 7 modules


FriederikeHanssen (approved):

pinin4fjords:
Thanks @FriederikeHanssen !

PR #12352, opened 2026-07-14
Add mm2plus/align

### LINKED ISSUE #12350: new module: mm2-fast/align

### Further Information

_No response_

```yaml
---
channels:
  - conda-forge
  - bioconda
dependencies:
  - bioconda::mm2-plus=1.3
  - bioconda::samtools=1.24
```

fellen31:
## PR checklist

Closes #12350. Drop in replacement for minimap2/align so more or less just copying that module. 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

### DIFF
--- a/modules/nf-core/mm2plus/align/main.nf
+++ b/modules/nf-core/mm2plus/align/main.nf
@@ -1,0 +1,67 @@
+process MM2PLUS_ALIGN {
+    tag "$meta.id"
+    label 'process_high'
+
+    // Note: the versions here need to match the versions used in the mulled container below and mm2plus/index
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine in ['singularity', 'apptainer'] && !task.ext.singularity_pull_docker_container ?
+        'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/72/7224ad076c5513325c2ef76f8788249af1d791dacaf9ee378d5b7512019e3a32/data' :
+        'community.wave.seqera.io/library/mm2plus_samtools:fe581c94b0a4dc10' }"
+
+    input:
+    tuple val(meta), path(reads)
+    tuple val(meta2), path(reference)
+    val bam_format
+    val bam_index_extension
+    val cigar_paf_format
+    val cigar_bam
+
+    output:
+    tuple val(meta), path("*.paf")                       , optional: true, emit: paf
+    tuple val(meta), path("*.bam")                       , optional: true, emit: bam
+    tuple val(meta), path("*.bam.${bam_index_extension}"), optional: true, emit: index
+    tuple val("${task.process}"), val("mm2plus"), eval("mm2plus --version"), topic: versions, emit: versions_mm2plus
+    tuple val("${task.process}"), val("samtools"), eval("samtools version | sed '1!d;s/.* //'"), topic: versions, emit: versions_samtools
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args  = task.ext.args ?: ''
+    def args2 = task.ext.args2 ?: ''
+    def args3 = task.ext.args3 ?: ''
+    def args4 = task.ext.args4 ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def bam_index = bam_index_extension ? "${prefix}.bam##idx##${prefix}.bam.${bam_index_extension} --write-index" : "${prefix}.bam"
+    def bam_output = bam_format ? "-a | samtools sort -@ ${task.cpus-1} -o ${bam_index} ${args2}" : "-o ${prefix}.paf"
+    def cigar_paf = cigar_paf_format && !bam_format ? "-c" : ''
+    def set_cigar_bam = cigar_bam && bam_format ? "-L" : ''
+    def bam_input = "${reads.extension}".matches('sam|bam|cram')
+    def samtools_reset_fastq = bam_input ? "samtools reset --threads ${task.cpus-1} $args3 $reads | samtools fastq --threads ${task.cpus-1} $args4 |" : ''
+    def query = bam_input ? "-" : reads
+    def target = reference ?: (bam_input ? error("Error: mm2-plus/align BAM input mode requires reference") : reads)
+    """
+    $samtools_reset_fastq \\
+    mm2plus \\
+        ${args} \\
+        -t ${task.cpus} \\
+        ${target} \\
+        ${query} \\
+        ${cigar_paf} \\
+        ${set_cigar_bam} \\
+        ${bam_output}
+    """
+
+    stub:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def output_file = bam_format ? "${prefix}.bam" : "${prefix}.paf"
+    def bam_index = bam_index_extension ? "touch ${prefix}.bam.${bam_index_extension}" : ""
+    def bam_input = "${reads.extension}".matches('sam|bam|cram')
+    if(bam_input && !reference) {
+        error("Error: mm2-plus/align BAM input mode requires reference!")
+	}
+    """
+    touch $output_file
+    ${bam_index}
+    """
+}
--- a/modules/nf-core/mm2plus/align/meta.yml
+++ b/modules/nf-core/mm2plus/align/meta.yml
@@ -24,9 +24,12 @@ input:
     - reads:
         type: file
         description: |
-          List of input FASTA or FASTQ files of size 1 and 2 for single-end
+          List of input BAM, or input FASTA or FASTQ files of size 1 and 2 for single-end
           and paired-end data, respectively.
-        ontologies: []
+        ontologies:
+          - edam: http://edamontology.org/format_2572 # BAM
+          - edam: http://edamontology.org/format_1929 # FASTA
+          - edam: http://edamontology.org/format_1930 # FASTQ
   - - meta2:
         type: map
         description: |
@@ -36,7 +39,8 @@ input:
         type: file
         description: |
           Reference database in FASTA format.
-        ontologies: []
+        ontologies:
+          - edam: http://edamontology.org/format_1929 # FASTA
   - bam_format:
       type: boolean
       description: Specify that output should be in BAM format
@@ -73,7 +77,8 @@ output:
           type: file
           description: Alignment in BAM format
           pattern: "*.bam"
-          ontologies: []
+          ontologies:
+            - edam: http://edamontology.org/format_2572 # BAM
   index:
     - - meta:
           type: map
@@ -84,7 +89,8 @@ output:
           type: file
           description: BAM alignment index
           pattern: "*.bam.*"
-          ontologies: []
+          ontologies:
+            - edam: http://edamontology.org/format_3327 # BAM index
   versions_mm2plus:
     - - ${task.process}:
           type: string
--- a/modules/nf-core/mm2plus/align/meta.yml
+++ b/modules/nf-core/mm2plus/align/meta.yml
@@ -1,0 +1,131 @@
+name: mm2plus_align
+description: A versatile pairwise aligner for genomic and spliced nucleotide sequences
+keywords:
+  - align
+  - fasta
+  - fastq
+  - genome
+  - paf
+  - reference
+tools:
+  - mm2plus:
+      description: |
+        A versatile pairwise aligner for genomic and spliced nucleotide sequences.
+      homepage: https://github.com/at-cg/mm2-plus
+      documentation: https://github.com/at-cg/mm2-plus#readme_mm2
+      licence: ["MIT"]
+      identifier: ""
+input:
+  - - meta:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. [ id:'test', single_end:false ]
+    - reads:
+        type: file
+        description: |
+          List of input FASTA or FASTQ files of size 1 and 2 for single-end
+          and paired-end data, respectively.
+        ontologies: []
+  - - meta2:
+        type: map
+        description: |
+          Groovy Map containing reference information
+          e.g. [ id:'test_ref']
+    - reference:
+        type: file
+        description: |
+          Reference database in FASTA format.
+        ontologies: []
+  - bam_format:
+      type: boolean
+      description: Specify that output should be in BAM format
+  - bam_index_extension:
+      type: string
+      description: BAM alignment index extension (e.g. "bai")
+  - cigar_paf_format:
+      type: boolean
+      description: Specify that output CIGAR should be in PAF format
+  - cigar_bam:
+      type: boolean
+      description: |
+        Write CIGAR with >65535 ops at the CG tag. This is recommended when
+        doing XYZ (https://github.com/lh3/minimap2#working-with-65535-cigar-operations)
+output:
+  paf:
+    - - meta:
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. [ id:'test', single_end:false ]
+      - "*.paf":
+          type: file
+          description: Alignment in PAF format
+          pattern: "*.paf"
+          ontologies: []
+  bam:
+    - - meta:
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. [ id:'test', single_end:false ]
+      - "*.bam":
+          type: file
+          description: Alignment in BAM format
+          pattern: "*.bam"
+          ontologies: []
+  index:
+    - - meta:
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. [ id:'test', single_end:false ]
+      - "*.bam.${bam_index_extension}":
+          type: file
+          description: BAM alignment index
+          pattern: "*.bam.*"
+          ontologies: []
+  versions_mm2plus:
+    - - ${task.process}:
+          type: string
+          description: The process name
+      - mm2plus:
+          type: string
+          description: The tool name
+      - mm2plus --version:
+          type: eval
+          description: The tool version
+  versions_samtools:
+    - - ${task.process}:
+          type: string
+          description: The process name
+      - samtools:
+          type: string
+          description: The tool name
+      - samtools version | sed '1!d;s/.* //':
+          type: eval
+          description: The tool version
+topics:
+  versions:
+    - - ${task.process}:
+          type: string
+          description: The process name
+      - mm2plus:
+          type: string
+          description: The tool name
+      - mm2plus --version:
+          type: eval
+          description: The tool version
+    - - ${task.process}:
+          type: string
+          description: The process name
+      - samtools:
+          type: string
+          description: The tool name
+      - samtools version | sed '1!d;s/.* //':
+          type: eval
+          description: The tool version
+authors:
+  - "@fellen31"
+maintainers:
+  - "@fellen31"
--- a/modules/nf-core/mm2plus/align/tests/main.nf.test
+++ b/modules/nf-core/mm2plus/align/tests/main.nf.test
@@ -1,0 +1,441 @@
+nextflow_process {
+
+    name "Test Process MM2PLUS_ALIGN"
+    script "../main.nf"
+    process "MM2PLUS_ALIGN"
+
+    tag "modules"
+    tag "modules_nfcore"
+    tag "mm2plus"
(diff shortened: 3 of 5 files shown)


modules/nf-core/mm2plus/align/tests/main.nf.test
```diff
@@ -0,0 +1,441 @@
...
+    name "Test Process MM2PLUS_ALIGN"
+    script "../main.nf"
+    process "MM2PLUS_ALIGN"
+
+    tag "modules"
+    tag "modules_nfcore"
+    tag "mm2plus"
+    tag "mm2plus/align"
+
+    test("sarscov2 - fastq, fasta, true, [], false, false") {
```
vagkaratzas: I'd prefer more descriptive tag names, i.e. `sarscov2 - single_end - bam`. The current one is both hard to remember to call on its own, as well as doesnt make any sense at a glance on what it is testing

modules/nf-core/mm2plus/align/tests/main.nf.test
```diff
@@ -0,0 +1,441 @@
...
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(
+                    bam(process.out.bam[0][1]).getHeader(),
+                    bam(process.out.bam[0][1]).getReadsMD5(),
+                    process.out.findAll { key, val -> key.startsWith("versions_") }
+                ).match() }
```
vagkaratzas: Look into other modules that are using:
```
assertAll(
                { assert process.success },
                { assert snapshot(sanitizeOutput(process.out, unstableKeys: ["bam"])).match() }
            )
```
and then test your unstable keys as you already do.
that way `process.out.findAll { key, val -> key.startsWith("versions_") }` can be removed as well

modules/nf-core/mm2plus/align/main.nf
```diff
@@ -0,0 +1,67 @@
...
+        'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/72/7224ad076c5513325c2ef76f8788249af1d791dacaf9ee378d5b7512019e3a32/data' :
+        'community.wave.seqera.io/library/mm2plus_samtools:fe581c94b0a4dc10' }"
+
+    input:
+    tuple val(meta), path(reads)
+    tuple val(meta2), path(reference)
+    val bam_format
+    val bam_index_extension
+    val cigar_paf_format
+    val cigar_bam
```
vagkaratzas, line 17: Can this be made in a single tuple? are there any cases where reference should not be merged with reads on same meta before calling this?

modules/nf-core/mm2plus/align/main.nf
```diff
@@ -0,0 +1,67 @@
...
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args  = task.ext.args ?: ''
+    def args2 = task.ext.args2 ?: ''
+    def args3 = task.ext.args3 ?: ''
+    def args4 = task.ext.args4 ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def bam_index = bam_index_extension ? "${prefix}.bam##idx##${prefix}.bam.${bam_index_extension} --write-index" : "${prefix}.bam"
+    def bam_output = bam_format ? "-a | samtools sort -@ ${task.cpus-1} -o ${bam_index} ${args2}" : "-o ${prefix}.paf"
```
vagkaratzas, line 36: why ${task.cpus-1} ?

modules/nf-core/mm2plus/align/main.nf
```diff
@@ -0,0 +1,67 @@
...
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args  = task.ext.args ?: ''
+    def args2 = task.ext.args2 ?: ''
+    def args3 = task.ext.args3 ?: ''
+    def args4 = task.ext.args4 ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def bam_index = bam_index_extension ? "${prefix}.bam##idx##${prefix}.bam.${bam_index_extension} --write-index" : "${prefix}.bam"
```
vagkaratzas, line 35: is this weird convention required somewhere? why not just `${prefix}.bam.${bam_index_extension}`

modules/nf-core/mm2plus/align/main.nf
```diff
@@ -0,0 +1,67 @@
...
+    tuple val("${task.process}"), val("samtools"), eval("samtools version | sed '1!d;s/.* //'"), topic: versions, emit: versions_samtools
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args  = task.ext.args ?: ''
+    def args2 = task.ext.args2 ?: ''
+    def args3 = task.ext.args3 ?: ''
+    def args4 = task.ext.args4 ?: ''
```
vagkaratzas, line 33: we'll need nf-tests with nextflow.configs that pass all these 4 args to be convinced that they are required

modules/nf-core/mm2plus/align/meta.yml
```diff
@@ -0,0 +1,131 @@
...
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. [ id:'test', single_end:false ]
+    - reads:
+        type: file
+        description: |
+          List of input FASTA or FASTQ files of size 1 and 2 for single-end
+          and paired-end data, respectively.
+        ontologies: []
```
vagkaratzas: add ontologies everywhere

vagkaratzas (changes requested):
This seems like a tough one. Left some comments around to consider

modules/nf-core/mm2plus/align/tests/main.nf.test
fellen31: https://nf-co.re/docs/specifications/components/modules/testing#test-names 👀

fellen31: But, I agree with you.

modules/nf-core/mm2plus/align/tests/main.nf.test
fellen31: Thanks!

modules/nf-core/mm2plus/align/tests/main.nf.test
fellen31: Is https://github.com/nf-core/website where to add suggestions to update the documentation? Maybe we can make an issue there so we don't have to repeat this.

modules/nf-core/mm2plus/align/main.nf
fellen31: It could, but I don't see the advantage. I'd prefer to keep this as similar as possible to the minimap2/align module, so users can use this as a drop-in replacement. 

Perhaps one would like to align the same set of reads to multiple references? Then it could be nice to keep the meta of the reference, in case you want to use it in the tag or prefix.

modules/nf-core/mm2plus/align/main.nf
fellen31: Because it's the number of _additional_ threads.

modules/nf-core/mm2plus/align/main.nf
fellen31: Yes, see the samtools documentation: https://www.htslib.org/doc/samtools.html.

modules/nf-core/mm2plus/align/main.nf
fellen31: Sorry, I don't understand. Is it you that need to be convinced? All tools in this module can take additional arguments, and therefore needs task.ext: https://nf-co.re/docs/specifications/components/modules/general#optional-command-arguments, https://nf-co.re/docs/specifications/components/modules/general#naming-of-args-variables.

> "All non-mandatory command-line tool non-file arguments MUST be provided as a string via the $task.ext.args variable": 

```
Usage: samtools reset [options]
  -o FILE      Output file
  -x, --remove-tag STR
               Aux tags to be removed
      --keep-tag STR
               Aux tags to be retained. Equivalent to -x ^STR
      --reject-PG ID
               Removes PG line with ID matching to input and succeeding PG lines
      --no-RG  To have RG lines or not
      --no-PG  To have PG entry or not for reset operation
      --dupflag
               Keeps the duplicate flag as it is
  -O, --output-fmt FORMAT[,OPT[=VAL]]...
               Specify output format (SAM, BAM, CRAM)
  -@, --threads INT
               Number of additional threads to use [0]
```

```
samtools fastq
Usage: samtools fastq [options...] <in.bam>

Description:
Converts a SAM, BAM or CRAM to FASTQ format.

Options:
  -0 FILE      write reads designated READ_OTHER to FILE
  -1 FILE      write reads designated READ1 to FILE
  -2 FILE      write reads designated READ2 to FILE
  -o FILE      write reads designated READ1 or READ2 to FILE
               note: if a singleton file is specified with -s, only
               paired reads will be written to the -1 and -2 files.
  -d, --tag TAG[:VAL]
               only include reads containing TAG, optionally with value VAL
  -D, --tag-file STR:FILE
               only include reads containing TAG, with a value listed in FILE
  -f, --require-flags INT
               only include reads with all  of the FLAGs in INT present [0]
  -F, --excl[ude]-flags INT
               only include reads with none of the FLAGs in INT present [0x900]
      --rf, --incl[ude]-flags INT
               only include reads with any  of the FLAGs in INT present [0]
  -G INT       only EXCLUDE reads with all  of the FLAGs in INT present [0]
  -n           don't append /1 and /2 to the read name
  -N           always append /1 and /2 to the read name
  -O           output quality in the OQ tag if present
  -s FILE      write singleton reads designated READ1 or READ2 to FILE
  -t           copy RG, BC and QT tags to the FASTQ header line
  -T TAGLIST   copy arbitrary tags to the FASTQ header line, '*' for all
  -v INT       default quality score if not given in file [1]
  -i           add Illumina Casava 1.8 format entry to header (eg 1:N:0:ATCACG)
  -c INT       compression level [0..9] to use when writing bgzf files [1]
  --i1 FILE    write first index reads to FILE
  --i2 FILE    write second index reads to FILE
  --barcode-tag TAG
               Barcode tag [BC]
  --quality-tag TAG
               Quality tag [QT]
  --index-format STR
               How to parse barcode and quality tags

      --input-fmt-option OPT[=VAL]
               Specify a single input file format option in the form
               of OPTION or OPTION=VALUE
      --reference FILE
               Reference sequence FASTA FILE [null]
  -@, --threads INT
               Number of additional threads to use [0]
      --verbosity INT
               Set level of verbosity

The files will be automatically compressed if the file names have a .gz
or .bgzf extension.  The input to this program must be collated by name.
Run 'samtools collate' or 'samtools sort -n' to achieve this.

Reads are designated READ1 if FLAG READ1 is set and READ2 is not set.
Reads are designated READ2 if FLAG READ1 is not set and READ2 is set.
Otherwise reads are designated READ_OTHER (both flags set or both flags unset).
Run 'samtools flags' for more information on flag codes and meanings.

The index-format string describes how to parse the barcode and quality tags.
It is made up of 'i' or 'n' followed by a length or '*'.  For example:
   i14i8       The first 14 characters are index 1, the next 8 are index 2
   n8i14       Ignore the first 8 characters, and use the next 14 for index 1

If the tag contains a separator, then the numeric part can be replaced with
'*' to mean 'read until the separator or end of tag', for example:
   i*i*        Break the tag at the separator into index 1 and index 2
   n*i*        Ignore the left part of the tag until the separator,
               then use the second part of the tag as index 1

Examples:
To get just the paired reads in separate files, use:
   samtools fastq -1 pair1.fq -2 pair2.fq -0 /dev/null -s /dev/null -n in.bam

To get all non-supplementary/secondary reads in a single file, redirect
the output:
   samtools fastq in.bam > all_reads.fq
```

```
samtools sort
Usage: samtools sort [options...] [in.bam]
Options:
  -l INT     Set compression level, from 0 (uncompressed) to 9 (best)
  -u         Output uncompressed data (equivalent to -l 0)
  -m INT     Set maximum memory per thread; suffix K/M/G recognized [768M]
  -M         Use minimiser for clustering unaligned/unplaced reads
  -R         Do not use reverse strand (only compatible with -M)
  -K INT     Kmer size to use for minimiser [20]
  -I FILE    Order minimisers by their position in FILE FASTA
  -w INT     Window size for minimiser indexing via -I ref.fa [100]
  -H         Squash homopolymers when computing minimiser
  -n         Sort by read name (natural): cannot be used with samtools index
  -N         Sort by read name (ASCII): cannot be used with samtools index
  -t TAG     Sort by value of TAG. Uses position as secondary index (or read name if -n is set)
  -o FILE    Write final output to FILE rather than standard output
  -T PREFIX  Write temporary files to PREFIX.nnnn.bam
      --no-PG
               Do not add a PG line
      --template-coordinate
               Sort by template-coordinate
      --input-fmt-option OPT[=VAL]
               Specify a single input file format option in the form
               of OPTION or OPTION=VALUE
  -O, --output-fmt FORMAT[,OPT[=VAL]]...
               Specify output format (SAM, BAM, CRAM)
      --output-fmt-option OPT[=VAL]
               Specify a single output file format option in the form
               of OPTION or OPTION=VALUE
      --reference FILE
               Reference sequence FASTA FILE [null]
  -@, --threads INT
               Number of additional threads to use [0]
      --write-index
               Automatically index the output files [off]
      --verbosity INT
               Set level of verbosity
```

```
minimap2
Usage: minimap2 [options] <target.fa>|<target.idx> [query.fa] [...]
Options:
  Indexing:
    -H           use homopolymer-compressed k-mer (preferrable for PacBio)
    -k INT       k-mer size (no larger than 28) [15]
    -w INT       minimizer window size [10]
    -I NUM       split index for every ~NUM input bases [8G]
    -d FILE      dump index to FILE []
  Mapping:
    -f FLOAT     filter out top FLOAT fraction of repetitive minimizers [0.0002]
    -g NUM       stop chain enlongation if there are no minimizers in INT-bp [5000]
    -G NUM       max intron length (effective with -xsplice; changing -r) [200k]
    -F NUM       max fragment length (effective with -xsr or in the fragment mode) [800]
    -r NUM[,NUM] chaining/alignment bandwidth and long-join bandwidth [500,20000]
    -n INT       minimal number of minimizers on a chain [3]
    -m INT       minimal chaining score (matching bases minus log gap penalty) [40]
    -X           skip self and dual mappings (for the all-vs-all mode)
    -p FLOAT     min secondary-to-primary score ratio [0.8]
    -N INT       retain at most INT secondary alignments [5]
  Alignment:
    -A INT       matching score [2]
    -B INT       mismatch penalty (larger value for lower divergence) [4]
    -O INT[,INT] gap open penalty [4,24]
    -E INT[,INT] gap extension penalty; a k-long gap costs min{O1+k*E1,O2+k*E2} [2,1]
    -z INT[,INT] Z-drop score and inversion Z-drop score [400,200]
    -s INT       minimal peak DP alignment score [80]
    -u CHAR      how to find GT-AG. f:transcript strand, b:both strands, n:don't match GT-AG [n]
    -J INT       splice mode. 0: original minimap2 model; 1: miniprot model [1]
    -j FILE      junctions in BED12 to extend *short* RNA-seq alignment []
  Input/Output:
    -a           output in the SAM format (PAF by default)
    -o FILE      output alignments to FILE [stdout]
    -L           write CIGAR with >65535 ops at the CG tag
    -R STR       SAM read group line in a format like '@RG\tID:foo\tSM:bar' []
    -c           output CIGAR in PAF
    --cs[=STR]   output the cs tag; STR is 'short' (if absent) or 'long' [none]
    --ds         output the ds tag, which is an extension to cs
    --MD         output the MD tag
    --eqx        write =/X CIGAR operators
    -Y           use soft clipping for supplementary alignments
    -y           copy FASTA/Q comments to output SAM
    -t INT       number of threads [3]
    -K NUM       minibatch size for mapping [500M]
    --version    show version number
  Preset:
    -x STR       preset (always applied before other options; see minimap2.1 for details) []
                 - lr:hq - accurate long reads (error rate <1%) against a reference genome
                 - splice/splice:hq - spliced alignment for long reads/accurate long reads
                 - splice:sr - spliced alignment for short RNA-seq reads
                 - asm5/asm10/asm20 - asm-to-ref mapping, for ~0.1/1/5% sequence divergence
                 - sr - short reads against a reference
                 - map-pb/map-hifi/map-ont/map-iclr - CLR/HiFi/Nanopore/ICLR vs reference mapping
                 - ava-pb/ava-ont - PacBio CLR/Nanopore read overlap

See `man ./minimap2.1' for detailed description of these and other advanced command-line options.
```

modules/nf-core/mm2plus/align/meta.yml
fellen31: Added everywhere, except PAF which I couldn't find!

modules/nf-core/mm2plus/align/tests/main.nf.test
vagkaratzas: These are fine and standard:
```
test("homo_sapiens - [fastq1, fastq2] - bam")
test("sarscov2 - [ cram, crai ] - fasta - fai")
```
It starts to get weird when you add boolean flags and/or too many file types and can't tell the input from the output

modules/nf-core/mm2plus/align/tests/main.nf.test
```diff
@@ -0,0 +1,414 @@
...
+    name "Test Process MM2PLUS_ALIGN"
+    script "../main.nf"
+    process "MM2PLUS_ALIGN"
+
+    tag "modules"
+    tag "modules_nfcore"
+    tag "mm2plus"
+    tag "mm2plus/align"
+
+    test("FASTQ reads (single-end) and FASTA reference as input, BAM output, no index") {
```
vagkaratzas: better! maybe just put back `sarscov - ` in the start

modules/nf-core/mm2plus/align/main.nf
vagkaratzas: Alright, not a dealbreaker for me!

modules/nf-core/mm2plus/align/main.nf
vagkaratzas: Will probably need some documentation for this, maybe in the comments and/or meta.yml file. It isnt clear how they are used:
`samtools_reset_fastq` command uses all but one, before the main command
`-t ${task.cpus}` uses all cpus, but then in the same command `bam_output` is called with all but one --does this mean one is kept for the main command while the rest are passed in an internal process?

modules/nf-core/mm2plus/align/main.nf
vagkaratzas: Thanks!

modules/nf-core/mm2plus/align/main.nf
vagkaratzas: Now it makes sense. Will it be doable to have an extra test case (with a nextflow.conifg) that provides all or some of these ext.args?

vagkaratzas (changes requested):
Awesome, getting there :D
Some more things to consider before merging

fellen31:
> Now it makes sense. Will it be doable to have an extra test case (with a nextflow.conifg) that provides all or some of these ext.args?

I could add it, but there are so many modules that doesn't test if ext.args works. Same as there are many modules that doesn't check if ext.prefix works. Are there requirements for it written in the documentation? I think it makes sense to either require it for all modules, or no modules. 

> Will probably need some documentation for this, maybe in the comments and/or meta.yml file. It isnt clear how they are used:
samtools_reset_fastq command uses all but one, before the main command
-t ${task.cpus} uses all cpus, but then in the same command bam_output is called with all but one --does this mean one is kept for the main command while the rest are passed in an internal process?

Added a note in the module. The samtools `--threads` parameter specifies the number of additional threads to use. The default is 0 (1 thread + 0 additional threads). Most tools (including minimap2), specifies the total number of threads.

Another question is if mm2plus uses e.g 16 threads, it is most optimal for (edit:) the samtools steps then to also use 16 threads? Probably not, but it's a bit tricky to set reasonable defaults (see e.g. https://github.com/nf-core/modules/pull/11840#discussion_r3394937891), and decide how and if you want the user to be able to specify these threads separately. The current solution has worked so far in the minimap2 module.

modules/nf-core/mm2plus/align/tests/main.nf.test
fellen31: Done!

vagkaratzas (approved):
LGTM!

PR #12355, opened 2026-07-14
Swap snippy to topic channels

SPPearce:
Swap snippy to topics


mashehu (approved):

modules/nf-core/snippy/run/tests/main.nf.test
mashehu:
```suggestion
					process.out.findAll { key, val -> key.startsWith('versions') }
```

PR #12356, opened 2026-07-14
bismark: bump to 3.1.0

FelixKrueger:
## Bumps all seven `bismark` modules from 3.0.0 → 3.1.0

`align`, `coverage2cytosine`, `deduplicate`, `genomepreparation`, `methylationextractor`, `report`, `summary`.

Bismark 3.1.0 is released and on bioconda (latest). The CLI is a drop-in for 3.0.0 (no interface changes), so this is a straight version + container bump.

### How
- `environment.yml` → `bioconda::bismark=3.1.0` for each module.
- Containers regenerated with the new **`nf-core modules containers create`** (`@dev`) — Seqera Containers via Wave, docker + singularity, `linux/amd64` + `linux/arm64` — which added the per-arch `.conda-lock/` files and the `meta.yml` `containers:` block.

### ⚠️ Draft: nf-test snapshots not yet regenerated
The `versions.yml` in each module's snapshot still reads `3.0.0`, so CI's nf-test will fail until the snapshots are regenerated on **native Docker**. I couldn't regenerate them locally — bismark's alignment fails under my machine's amd64-on-arm64 emulation (colima). Regenerating on a native-amd64 runner (or letting CI/@nf-core-bot handle it) will finish this off. Happy to push the updated snapshots as soon as I can run them natively.

### Testing the new `containers create` subcommand (cc @maxulysse — you asked for feedback 🙏)
Exercised it on all 7 modules; two notes:
1. Without `TOWER_ACCESS_TOKEN` set it fails with a raw `HTTP 401` from `api.cloud.seqera.io/user-info` — a hint like "set `TOWER_ACCESS_TOKEN`" would be friendlier.
2. The generated `container` directive is written with the ternary `?`/`:` at column 0 (leading operators) rather than the canonical operator-trailing `"${ workflow… ?` style — I reformatted each block to match the existing modules. Might be worth aligning the emitted style with `nf-core modules lint`/prettier expectations.

cc @pinin4fjords — follow-up to the 3.0.0 bump in nf-core/methylseq#610.


mashehu:
> Without TOWER_ACCESS_TOKEN set it fails with a raw HTTP 401 from api.cloud.seqera.io/user-info — a hint like "set TOWER_ACCESS_TOKEN" would be friendlier.

if there is no TOWER_ACCESS_TOKEN set, we print already:
```console
WARNING  TOWER_ACCESS_TOKEN is not set. Wave API requests will be subject to stricter rate limits.
         To increase your quota, set the TOWER_ACCESS_TOKEN environment variable with your Seqera Platform token.
         See https://docs.seqera.io/wave/api#api-limits for more information.
```

> The generated container directive is written with the ternary ?/: at column 0 (leading operators) rather than the canonical operator-trailing "${ workflow… ? style — I reformatted each block to match the existing modules. Might be worth aligning the emitted style with nf-core modules lint/prettier expectations.

your LLM needs an update on nextflow formatting 😉.  `nextflow lint -format` formats ternaries in this way (would have also preferred a different way https://github.com/nextflow-io/language-server/issues/143 🙂)

SPPearce:
If you don't expect much change in the output files, just find/replace 3.0.0 with 3.1.0 in the snapshots.

FelixKrueger:
It seems I need to teach my Claude some manners first - thanks for your comments! Felix

Thanks both — pushed the fixes (`02f436f`):

@SPPearce spot on — bismark 3.1.0 is a byte-identical drop-in on the standard path, so I just find/replaced the `"3.0.0"` → `"3.1.0"` version strings in the 7 snapshots (outputs unchanged, no re-run). 🙏

@mashehu:
- **Formatting:** fair cop 😄 — ran `nextflow lint -format` on all 7 `main.nf`; it produces the leading-operator ternary exactly as you said, and I've reverted my (outdated) operator-trailing reformat. Thanks for the pointer.
- **Token:** one clarification in case it's useful — with no `TOWER_ACCESS_TOKEN`, `containers create` didn't just warn-and-throttle for me; every build **hard-failed** with `HTTP 401` (`Unauthorized access to Tower resource: …/user-info`) and produced no containers (`✗ Some container builds failed`). Setting the token fixed it immediately. So at least for the frozen community build it seemed *required* rather than only rate-limited — possibly worth a clearer "token required" message. Happy to share the full log.

FelixKrueger:
CI is green now ✅ (lint + nf-test across all 7 modules; snapshots matched since 3.1.0 is byte-identical to 3.0.0 on the standard path).

One heads-up from the back-and-forth, in case it's useful for reviewers and the containers-create rollout: I first ran `nextflow lint -format` on the `main.nf`s, but that rewrites the scripts into new strict-syntax (`catch (e: Exception)`, `log.warn(...)`, etc.) which passes `nf-core lint` (new parser) yet **fails nf-test's runtime compilation** (classic parser, nextflow 25.10.2 — `Unexpected input '{'` / `Module compilation error`). So I reverted to a minimal **container-URL-only** change, keeping the classic-compatible syntax the merged modules currently use.

@mashehu — flagging for the `nextflow lint -format` / containers-create work: on these modules the formatter's output isn't nf-test-runnable yet (the runtime still parses classic syntax), so a full reformat breaks the test job even though lint is happy.

Ready for review — cc @pinin4fjords 🙏

pinin4fjords (approved):

PR #12362, opened 2026-07-15
fix(tximeta/tximport): add jq dep as workaround for yq missing its jq runtime dep

pditommaso:
## Summary

Adds `conda-forge::jq` to the `tximeta/tximport` conda environment as a **temporary workaround** for a missing upstream dependency.

## Problem

`bioconductor-tximeta` pulls in `bioconductor-genomeinfodbdata`, whose payload is installed by a conda **post-link** script (`installBiocDataPackage.sh`). That script parses the download URL with `yq`, and `yq` (the Python `yq`, a jq wrapper) shells out to the `jq` binary.

The conda-forge `yq` package **does not declare `jq` as a runtime dependency**. On minimal build images that don't ship `jq` (e.g. Wave's `micromamba:2-amazon2023` build stage), the post-link fails silently, `GenomeInfoDbData` is never installed, and the module then fails at runtime with:

```
Error: package or namespace load failed for 'GenomeInfoDb':
 there is no package called 'GenomeInfoDbData'
```

This was hit building the container from this module's `environment.yml` (e.g. via Wave); the prebuilt biocontainer is unaffected because it is built in bioconda CI where `jq` is present.

## Fix

Pin `conda-forge::jq` in `environment.yml` so `jq` is guaranteed present at build time, letting the bioconductor post-link resolve the URL, download the tarball, and `R CMD INSTALL GenomeInfoDbData`.

Verified with a real container build on `micromamba:2-amazon2023`: without `jq`, `GenomeInfoDbData` is missing; with it, the package installs and the module runs.

## Upstream

The real root cause is the missing `jq` run dependency in the conda-forge `yq` recipe:
- https://github.com/conda-forge/yq-feedstock/issues/41

This module-level pin should be **reverted once the upstream `yq` recipe declares `jq`**.

## Notes

- Comment in `environment.yml` documents why the dep exists and links the upstream issue so it can be cleaned up later.
- Container references (`main.nf`) will need the multi-package image rebuilt by nf-core tooling to include `jq`.


modules/nf-core/tximeta/tximport/environment.yml
```diff
@@ -5,3 +5,12 @@ channels:
   - bioconda
 dependencies:
   - bioconda::bioconductor-tximeta=1.20.1
+  # Temporary workaround: bioconductor data-package post-link scripts
```
pinin4fjords: Can we remove the narrative please?

pinin4fjords (changes requested):
Makes sense to add this.

Can you strip the narrative, and also update the container URIs to replace the biocontainers ones please? Or I can

modules/nf-core/tximeta/tximport/environment.yml
pditommaso: You can 😄, just make a suggestion to this PR please

modules/nf-core/tximeta/tximport/environment.yml
```diff
@@ -5,3 +5,12 @@ channels:
...
+  # (installBiocDataPackage.sh) parse download URLs with `yq`, which shells out
+  # to `jq`. The conda-forge `yq` package does not declare `jq` as a runtime
+  # dependency, so on minimal build images `jq` is absent, the post-link fails
+  # silently, and GenomeInfoDbData is not installed (runtime error:
+  # "there is no package called 'GenomeInfoDbData'"). Pinning jq here guarantees
+  # it is present at build time. Remove once the upstream yq recipe is fixed.
+  # Upstream issue: https://github.com/conda-forge/yq-feedstock/issues/41
+  - conda-forge::jq
```
mashehu: needs to be pinned to a version

pinin4fjords:
Pushed a follow-up commit to this branch:

- Dropped the inline explanation from `environment.yml` (kept it here in the description instead) — just `conda-forge::jq=1.8.2` now.
- Rebuilt the container via Wave from the updated `environment.yml` and pointed `main.nf` at the multi-package image, so `-profile docker`/`-profile singularity` actually get `jq` too (the previous commit only fixed the `-profile conda` path, since the container reference still pointed at the old single-package biocontainer).

New container references:
- Docker: `community.wave.seqera.io/library/bioconductor-tximeta_jq:78bccd386c46a07c`
- Singularity: `https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/bd/bdba33f8ad1b2df156f8f6775279fb217ce0f8233a3dc637337245de9ad2f29f/data`

Verified directly against the Docker image:
```
$ docker run --rm community.wave.seqera.io/library/bioconductor-tximeta_jq:78bccd386c46a07c bash -c "jq --version && Rscript -e 'library(GenomeInfoDb); library(tximeta)'"
jq-1.8.2
OK: GenomeInfoDb + tximeta loaded
```

Also worth noting for reviewers: CI on the prior commit was green on all profiles, but that's not meaningful signal for this bug — GitHub-hosted runners ship `jq` by default, so `-profile conda` never reproduced the failure in CI either way. The bug is specific to minimal build images (like Wave's build stage) that don't have `jq` preinstalled.

modules/nf-core/tximeta/tximport/tests/main.nf.test.snap
```diff
@@ -167,23 +167,23 @@
                     {
                         "id": "test"
                     },
-                    "test.gene_counts.tsv:md5,c14cab7e15cfac73ec0602dc2c404551"
+                    "test.gene_counts.tsv:md5,28b312144d974387ec28f057f8c9d747"
```
mashehu, line 170: how come the tsv's changed?

pditommaso: Oops, no idea. let me fix this

pinin4fjords: I'm just working this out now

pinin4fjords: Locale issue affecting sort on the output files

pinin4fjords (approved):

PR #12370, opened 2026-07-15
Remove unused ch_versions from subworkflows

SPPearce:
Three subworkflows had ch_versions but no actual versions going into it.


subworkflows/nf-core/fastq_trim_fastp_fastqc/tests/main.nf.test.snap
```diff
@@ -19,15 +19,13 @@
             [
                 
             ],
-            [
-                
-            ]
+            null
```
mashehu: almost 🙂

mashehu (approved):

subworkflows/nf-core/fastq_trim_fastp_fastqc/tests/main.nf.test.snap
SPPearce: Bah! I thought I'd got them all.

PR #12372, opened 2026-07-15
Propr grea topics

SPPearce:
Swap propr/grea to topics


modules/nf-core/propr/grea/meta.yml
```diff
@@ -77,6 +77,11 @@ output:
         description: dump of R SessionInfo
         pattern: "*.R_sessionInfo.log"
         ontologies: []
+topics:
```
FriederikeHanssen, line 80: is this right given that the template doesn't allow for topics being used atm?

subworkflows/nf-core/differential_functional_enrichment/main.nf
```diff
@@ -149,9 +146,6 @@ workflow DIFFERENTIAL_FUNCTIONAL_ENRICHMENT {
 
     ch_session_info = GPROFILER2_GOST.out.session_info
 
-    ch_versions = ch_versions
```
FriederikeHanssen, line 152: same here

modules/nf-core/propr/grea/meta.yml
SPPearce: https://nf-co.re/docs/specifications/components/modules/general#topic-qualifiers-should-be-used-for-emitting-versions
For template, we still make the versions.yml.

modules/nf-core/propr/grea/main.nf
FriederikeHanssen, line 16:
```suggestion
    path "versions.yml",                 emit: versions, topic: versions
```

SPPearce: Oh, awkward.

SPPearce: Not quite sure where that went, it must have been present when nf-core modules lint --fix updated the meta.yml

PR #12385, opened 2026-07-19
bismark/align: intra-instance -p threading (replace --multicore fork model for Bowtie 2 + HISAT2)

FelixKrueger:
## What & why

`bismark/align` auto-sets `--multicore N`. For Bowtie 2 that forks N workers, each loading the 2 (directional/pbat) or 4 (non-directional) converted-strand indexes — so peak RAM scales with N, and the `cpus/3`,`cpus/5` divisor leaves cores idle.

This replaces the fork model with intra-instance threading: `-p = task.cpus / n_instances` (n = 2 directional/pbat, 4 non-directional, 1 for `--combined_index`). One index load per strand, all cpus used. Applies to Bowtie 2 and HISAT2; minimap2/rammap keep the legacy `--multicore` path (bismark can't route `-p` to them yet). An explicit `--multicore`/`-p`/`--parallel` in `ext.args` still short-circuits the auto-compute.

Bowtie 2 output stays **byte-identical** — `-p N --reorder` is thread-invariant (verified 3-way against a single-core oracle). Benchmark (GRCh38, 2 M PE): wall −38 to −46 %, peak RAM −44 to −82 %.

## Testing

`nf-test` green. The 5 existing tests are untouched — at the default `cpus=2` no `-p` is emitted, so no snapshots regenerate; 4 new tests assert the emitted `-p` (from the report's options echo) at the 4-cpu CI ceiling.

## Checklist
- [x] Description of changes with reason (above)
- [x] Tests added; `nf-test` passes in CI
- [x] No version bump / `modules.json` change


modules/nf-core/bismark/align/tests/main.nf.test
```diff
@@ -205,4 +205,142 @@ nextflow_process {
...
+                input[2] = BISMARK_GENOMEPREPARATION.out.index
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                // directional Bowtie 2 = 2 instances; 4 cpus / 2 = -p 2
+                { assert file(process.out.report[0][1]).text.contains('-p 2 --reorder') }
```
mashehu, line 238:
```suggestion
                { assert file(process.out.report[0][1]).text.contains('--threads 2 --reorder') }
```
we use long names in nf-core

modules/nf-core/bismark/align/main.nf
```diff
@@ -28,32 +28,64 @@ process BISMARK_ALIGN {
...
+            n_instances = (args.contains('--non_directional') && args.contains('--combined_index_parallel')) ? 2 : 1
+        } else if(args.contains('--non_directional')){
+            n_instances = 4
+        }
+
+        // -p threads per instance; emit only when it lands >= 2 (bismark requires -p >= 2),
+        // otherwise omit it (faithful single-thread-per-instance default).
+        def pthreads = ((task.cpus as int) / n_instances) as int
+        if(pthreads >= 2){
+            args += " -p ${pthreads}"
```
mashehu, line 80:
```suggestion
            args += " --threads ${pthreads}"
```

FelixKrueger:
Thanks @mashehu! Two notes on these:

1. **Long name** — good call. The catch is that the module invokes `bismark` (not bowtie2 directly), and bismark's CLI currently defines `-p` as short-only — no `--threads` alias, unlike bowtie2 itself — so `bismark --threads N` would fail to parse against the shipped `3.1.0` container. I've added a `--threads` alias to bismark (FelixKrueger/Bismark#1075); once a bismark release carrying it lands and this module bumps to that container, we can switch the flag here. Keeping `-p` for now so the module works against the current `bismark:3.1.0`.

2. **Test assertion** — I've kept `-p 2 --reorder` rather than `--threads 2`: the assertion checks the alignment *report*, which echoes the underlying **bowtie2** command. bismark always constructs that with `-p` internally regardless of how threads are passed to bismark, so `--threads` never appears in the report — the assertion has to stay `-p`.

modules/nf-core/bismark/align/main.nf
FelixKrueger: see below, this change can't be applied until a new Bismark release adds `--threads`.

modules/nf-core/bismark/align/tests/main.nf.test
FelixKrueger: same here, see below

modules/nf-core/bismark/align/tests/nextflow_pthreads.config
```diff
@@ -0,0 +1,17 @@
+process {
+    // 4 cpus = the nf-core CI runner ceiling (strict mode hard-errors if a process
+    // requests more than is available). At 4 cpus the intra-instance `-p` divisor gives:
+    // directional -> -p 2, HISAT2 directional -> -p 2, combined (n=1) -> -p 4, and
+    // non-directional (n=4) -> 4/4=1 -> NO -p (exercises the I2 omit-below-2 boundary).
+    withName: BISMARK_ALIGN {
+        cpus     = 4
```
mashehu: why is this needed?

FelixKrueger: The module's default test config runs at `cpus = 2`, and at 2 cpus the `-p = cpus / n_instances` divisor this PR adds emits **no** `-p` (directional: `2/2 = 1 < 2`) — so without a higher cpu count, none of the new tests would actually exercise the `-p` behaviour they're here to verify. `cpus = 4` is the minimum that produces `-p` (directional → `-p 2`), and it's the nf-core CI runner ceiling (strict mode hard-errors if a process requests more cpus than available).

It's a *separate* config (rather than bumping the global) on purpose: the existing tests must stay at `cpus = 2`, where the new divisor emits no `-p` — that's exactly what keeps their output byte-identical, so this PR needs **zero snapshot regeneration**. Raising the global cpu count would change their behaviour and force snapshot churn.

mashehu: ah, we should probably bump the global config as well.
If you need a fixed number of cpus for tests to be stable, you should specify them.

FelixKrueger: Agreed that tests should specify cpus rather than lean on the runner default. Two things to flag before bumping the global config:

1. Raising the global cpus to 4 will re-baseline the **hisat2** snapshots — hisat2 isn't thread-invariant, so `-p 2` shifts its output. (bowtie2 and minimap2 stay put: their snapshots are reads-only / thread-invariant.) Happy to regenerate those.
2. The align tests also need genome preparation to receive **only** its aligner-selection flag — `bismark_genome_preparation` rejects `--non_directional` / `--combined_index` — so a small `withName: BISMARK_GENOMEPREPARATION` arg override stays either way.

Want me to set a fixed `cpus` on `BISMARK_ALIGN` in the module's global test config (and regen the hisat2 snapshots), or keep it scoped to the align tests?

modules/nf-core/bismark/align/tests/nextflow_pthreads.config
```diff
@@ -0,0 +1,12 @@
+process {
+    // 4 cpus = nf-core CI ceiling; enough to exercise the -p divisor across paths.
+    withName: BISMARK_ALIGN {
+        cpus     = 4
```
mashehu: this is outdated now after https://github.com/nf-core/modules/pull/12402

FelixKrueger: Good call — dropped it in 4b4b594. Tests now inherit the global `cpus=4` from #12402, and I folded the genome-prep arg separation into the default test config. CI's green. Thanks!

modules/nf-core/bismark/align/main.nf
```diff
@@ -28,32 +28,54 @@ process BISMARK_ALIGN {
...
-            log.warn "Not able to define bismark align multicore based on available memory"
+            // Check that we have enough memory
+            try {
+                def tmem = (task.memory as MemoryUnit).toBytes()
+                def mcore = (tmem / mem_per_multicore) as int
+                ccore = Math.min(ccore, mcore)
+            } catch (Exception e) {
+                log.warn "Error catched: ${e}"
+                log.warn "Not able to define bismark align multicore based on available memory"
+            }
```
mashehu:
```suggestion
            if (task.memory) {
			    def mcore = (task.memory.toBytes() / mem_per_multicore) as int
			    ccore = Math.min(ccore, mcore)
			} else {
			    log.warn "Not able to define bismark align multicore based on available memory"
			}
```
no need to be so wordy

mashehu (approved):

modules/nf-core/bismark/align/main.nf
FelixKrueger: Done in d317f1f — dropped the try/catch for a plain `task.memory` guard (formatter tidied the braces). Thanks!

PR #12386, opened 2026-07-20
added new module: hackgap/count

LeonHornich:
This PR adds a module supporting [Hackgap](https://gitlab.com/rahmannlab/hackgap), a k-mer counter supporting gapped k-mers.

Note: the tool outputs 2 types of files:
- `.hash`: hash table resulting from Bucketed Cuckoo. Since this is a probabilistic method, the files content (and with this the nf-test snapshots) are not reproducible between runs. That's why we only check for the files existence during testing.
- `.info`: metadata of hashing process. 

## PR checklist

- [x] Add a resource `label`
  - For modules:


modules/nf-core/hackgap/count/main.nf
```diff
@@ -0,0 +1,44 @@
...
+    tuple val(meta), path("*.info"), emit: info
+    tuple val("${task.process}"), val('hackgap'), eval("hackgap --version"), topic: versions, emit: versions_hackgap
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def split_threads = Math.max(1, (task.cpus / 4) as int)
```
piplus2: better passing task.cpus only as this may become hidden from the user

modules/nf-core/hackgap/count/main.nf
```diff
@@ -0,0 +1,44 @@
...
+    def read_threads = Math.max(1, (task.cpus / 4) as int)
+    def subtables_threads = Math.max(1, task.cpus - split_threads - read_threads)
+    """
+    hackgap count \\
+        $args \\
+        --threads-split ${split_threads} \\
+        --threads-read ${read_threads} \\
+        --subtables ${subtables_threads} \\
+        --files ${reads} \\
+        -o ${prefix}
```
piplus2:
```suggestion
        --output ${prefix}
```

piplus2 (commented):
tiny stuff

modules/nf-core/hackgap/count/main.nf
SPPearce: What is the issue with this?

piplus2: ~~this hides the cpu settings from the user. I don't get why one should set task.cpus and then it gets into 1/4th. Passing the right number makes it transparent.~~

piplus2: I get that they must sum to task.cpus, but 1/4 seems arbitrary. Any reason why that fraction?

LeonHornich: No particular reason. I used a similar logic on the last [module](https://github.com/nf-core/modules/blob/master/modules/nf-core/cleanifier/filter/main.nf) i added and it wasn't commented on. Therefore I assumed it was good to go. But please if you have a suggestion for a better cpu thread handling let me know, I"ll gladly adapt it :)

piplus2: https://gitlab.com/rahmannlab/hackgap/-/blob/master/hackgap/hackgap/hackgap_count.py

```
def check_threads(args):
    adjust = (args.subtables is None) or (args.threads_split is None) or (args.threads_read is None)
    # 1. Define number of subtables
    cpus = os.cpu_count()
    subtables = args.subtables
    if subtables is None:
        subtables = max(min(cpus // 2 - 1, cpus - 3, 19), 1)
        if (subtables % 2) == 0:
            subtables += 1
    if subtables < 1:
        debugprint0(f"- Error: At least one subtable is required, but {subtables=}")
        sys.exit(1)
    if subtables % 2 == 0:
        debugprint0(f"- Error: Number of subtables must be odd, but {subtables=}")
        sys.exit(1)
    # 2. Define threads for reading files
    threads_read = args.threads_read
    if threads_read is None:
        threads_read = int(ceil(subtables / 10))  # who knows?
    # 3. Define threads for
    threads_split = args.threads_split
    if threads_split is None:
        threads_split = 2 * threads_read
    if adjust and (subtables + threads_read + threads_split >= cpus):
        threads_read = 1
        threads_split = 2
    # 4. Return results
    assert threads_read >= 1
    assert threads_split >= 1
    return (subtables, threads_read, threads_split)
```

from the source code it looks like that `threads_split` is suggested to be twice `threads_read`. I'd suggest to follow the same logic. Also please be sure that the sum `subtables + threads_read + threads_split` is never >= task.cpus, or it drops to 1 and 2 as from the source. Thanks!

modules/nf-core/hackgap/count/main.nf
```diff
@@ -0,0 +1,44 @@
...
+    def read_threads = Math.max(1, (task.cpus / 4) as int)
+    def subtables_threads = Math.max(1, task.cpus - split_threads - read_threads)
+    """
+    hackgap count \\
+        $args \\
+        --threads-split ${split_threads} \\
+        --threads-read ${read_threads} \\
+        --subtables ${subtables_threads} \\
+        --files ${reads} \\
+        --out ${prefix}
```
piplus2, line 46: shouldn't it be --output?

[https://gitlab.com/rahmannlab/hackgap/-/blob/master/README.md](https://gitlab.com/rahmannlab/hackgap/-/blob/master/README.md?ref_type=heads#usage-guide)

`-o` or `--output`: This parameter specifies the index name and the path at which it is stored.

LeonHornich: Yes, i noticed that as well. My tests crash though when i use `--output` instead of `--out`. I checked the source code and it looks like it expects `--out`:

```
p.add_argument("-o", "--out", metavar="OUTPUT_PREFIX", required=True,
        help="path/prefix of output file with k-mer count hash (required; use '/dev/null' or 'null' or 'none' to avoid output)")
```

Seems like this is a small inconsistency within the documentation. I will bring this up with the maintainer.

piplus2: ok, thanks!

piplus2 (approved):
LGTM once threads are fixed

PR #12387, opened 2026-07-20
Propr topics

### LINKED ISSUE #12535: Migrate to topic propr/grea

### LINKED ISSUE #12536: Migrate to topic propr/propr

### LINKED ISSUE #12537: Migrate to topic propr/logratio

SPPearce:
Swap propr to topics

Close #12537
Close #12536 
Close #12535


modules/nf-core/propr/propd/tests/main.nf.test
Joon-Klaps, line 1: Feel like this snapshot test might also need to use sanitizeOutput with unstable keys & then one of them having the version explicitly in there as well ot be consistent with the previous module?

Joon-Klaps (commented):
Notice you didn't rename the versions to `module_versions`, you'll probably now the guidelines better but was that mandatory? I mean, it won't break users pipelines now if when they patch.

SPPearce:
> Notice you didn't rename the versions to `module_versions`, you'll probably now the guidelines better but was that mandatory? I mean, it won't break users pipelines now if when they patch.

The versions channel? I thought we were just using emit: versions, topic: versions, but will check.

SPPearce:
> Notice you didn't rename the versions to `module_versions`, you'll probably now the guidelines better but was that mandatory? I mean, it won't break users pipelines now if when they patch.

https://nf-co.re/docs/specifications/components/modules/general#topic-qualifiers-should-be-used-for-emitting-versions says to keep with `path "versions.yml", emit: versions, topic: versions` (as is used by ~50 modules currently).

modules/nf-core/propr/propd/tests/main.nf.test
SPPearce: Updated

Joon-Klaps (approved):

PR #12390, opened 2026-07-20
Update shinyngs modules to r-shinyngs 3.1.1

pinin4fjords:
## Summary

- Bumps all 4 `shinyngs` modules (`app`, `staticdifferential`, `staticexploratory`, `validatefomcomponents`) from `r-shinyngs` 2.4.0 to 3.1.1.
- Built multi-arch (linux/amd64, linux/arm64) Wave containers for the new environment using the new `nf-core/tools` dev `modules containers create` command, and added the resulting `containers:` metadata block plus `.conda-lock/` files to each module's `meta.yml`, following the existing `fastqc`/`multiqc` precedent.
- Fixed the `staticdifferential` test's `--unlog_foldchanges` invocation: it's a boolean `store_true` flag and doesn't take a value, so it's now only passed as a bare flag when the corresponding param is true.
- Updated snapshots: every diff is the `2.4.0` -> `3.1.1` version string (plus the `app.R` content hash for `shinyngs/app`, since `make_app_from_files.R`'s generated app bundle differs slightly between versions).

## Test plan

- [x] `nf-core modules lint` on all 4 modules: 0 failed (the 2 pre-existing warnings, `container_links` 404 and `main_nf_container` version mismatch, are stable-linter limitations with Wave containers - also present on the already-merged `multiqc` module)
- [x] `nf-test --profile docker` for all 4 modules: all pass
- [x] Verified the actual regression this bump depended on - a crash in the MAD/outlier plot whenever `--write_html` is used - is fixed upstream in [pinin4fjords/shinyngs#268](https://github.com/pinin4fjords/shinyngs/pull/268) (released as 3.1.1)

### DIFF
--- a/modules/nf-core/shinyngs/staticexploratory/.conda-lock/linux_amd64-bd-b5595b301bfbfd4d_2.txt
+++ b/modules/nf-core/shinyngs/staticexploratory/.conda-lock/linux_amd64-bd-b5595b301bfbfd4d_2.txt
@@ -1,0 +1,3302 @@
+
+version: 6
+environments:
+default:
+channels:
+- url: https://conda.anaconda.org/conda-forge/
+- url: https://conda.anaconda.org/bioconda/
+- url: https://conda.anaconda.org/bioconda/
+options:
+pypi-prerelease-mode: if-necessary-or-explicit
+packages:
+linux-64:
+- conda: https://conda.anaconda.org/conda-forge/linux-64/_openmp_mutex-4.5-20_gnu.conda
+- conda: https://conda.anaconda.org/conda-forge/noarch/_r-mutex-1.0.1-anacondar_1.tar.bz2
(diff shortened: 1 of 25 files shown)


modules/nf-core/shinyngs/staticexploratory/.conda-lock/linux_amd64-bd-b5595b301bfbfd4d_2.txt
luisas, line 1: are these meant to be uploaded?

luisas (approved):
The only thing I am unsure about is the conda lock files, are they now supposed to be uploaded?

modules/nf-core/shinyngs/staticexploratory/.conda-lock/linux_amd64-bd-b5595b301bfbfd4d_2.txt
pinin4fjords: Thanks for review! Yep, for the record these are the product of the newest nf-core tools version and a few modules are starting to adopt them

PR #12402, opened 2026-07-22
bump default CPU and memory to use the full runner resources

mashehu:
This was set back when GitHub runners only had 2 CPUs, I guess.

Discovered in #12385


itrujnara (approved):
Looks reasonable

PR #12414, opened 2026-07-23
Migrate ngmaster versions to topic channel

### LINKED ISSUE #12322: Migrate to topic ngmaster

xHaMMaDy:
## Summary

- migrate NGMASTER version reporting from `versions.yml` to the `versions` topic
- document the version tuple in `meta.yml`
- update nf-test snapshots
- handle ngmaster warnings consistently across execution profiles

Closes #12322

## Tests

- [x] Docker
- [x] Singularity
- [x] Conda
- [x] Nextflow lint


modules/nf-core/ngmaster/main.nf
```diff
@@ -12,7 +12,7 @@ process NGMASTER {
 
     output:
     tuple val(meta), path("*.tsv"), emit: tsv
-    path "versions.yml"           , emit: versions
+    tuple val("${task.process}"), val('ngmaster'), eval("ngmaster --version 2>&1 | sed -n 's/^ngmaster //p' | tail -n 1"), emit: versions_ngmaster, topic: versions
```
mashehu:
```suggestion
    tuple val("${task.process}"), val('ngmaster'), eval("ngmaster --version 2>&1 | sed -n '\$s/^ngmaster //p'"), emit: versions_ngmaster, topic: versions
```
a bit more compact

mashehu (commented):
Please join the github org via the slack channel so the CI runs for you.

modules/nf-core/ngmaster/main.nf
mashehu: what was wrong with this way to extract the version?

xHaMMaDy: Nothing was wrong with the original expression. After I applied the compact suggestion, CI failed on Nextflow 25.10.2 with No such variable: s. The $s was interpreted as variable interpolation inside the double-quoted eval string, and the suggested command was also missing the pipe before sed. I switched to the equivalent tail-first form to avoid the interpolation. It now passes lint and all Docker, Conda, and Singularity jobs. I’m happy to restore the original expression if you prefer.

mashehu: sorry for the buggy suggestion than. Will suggest a better one.

mashehu: see my fixed suggestion above.

xHaMMaDy: Thanks. I tested the edited suggestion locally with Nextflow 25.10.2. It passes parsing, but nf-test still fails with `sed: unsupported command /` because the generated `.command.sh` wraps the expression in another double-quoted `bash -c`, where `$s` is expanded again.

Using `\\\$s` in the Nextflow source preserves it through both layers:

```groovy
eval("ngmaster --version 2>&1 | sed -n '\\\$s/^ngmaster //p'")
```

This version passes Nextflow lint and Docker nf-test 2/2 locally. Would you prefer this form or the current green tail-first expression?

mashehu: yes

xHaMMaDy: Applied, thanks!

SPPearce (approved):

PR #12415, opened 2026-07-23
Change arity for tumor-only mode: `nanomonsv/get` module

manascripts:
This PR makes a small change to the `main.nf` of `nanomonsv/get` to allow it to run in tumor-only mode without matched control files from `nanomonsv/parse`.


piplus2 (approved):
LGTM

PR #12420, opened 2026-07-24
update missing defs in main and conditional gzip outputs, meta yml missing fields

vagkaratzas:
### Changes

**Scope leak.** `output`, `alignment`, `target_summary` and `domain_summary` were
assigned in the `script:` block without `def`, leaking them into the enclosing
process scope. Added `def` to each. Safe here because the `output:` block matches
on globs (`path('*.txt.gz')`), never on these variables.

**Explicit gzip file list.** The trailing `gzip` call built its arguments from
globs plus inline ternaries:

Also brought test naming and snapshots to latest standards.

```groovy
gzip --no-name *.txt \
    ${write_align ? '*.sto' : ''} \
    ${write_target ? '*.tbl' : ''} \
    ${write_domain ? '*.domtbl' : ''}
```
This now builds the list in Groovy and names the files it actually created:
```
def to_gzip = ["${prefix}.txt"]
if (write_align)  to_gzip << "${prefix}.sto"
if (write_target) to_gzip << "${prefix}.tbl"
if (write_domain) to_gzip << "${prefix}.domtbl"
```
Stub rebuilt from the same list. The stub previously emitted dangling line
continuations when the write flags were false:
```
touch "${prefix}.txt"
${write_align ? "touch ${prefix}.sto" : ''} \
${write_target ? "touch ${prefix}.tbl" : ''} \
${write_domain ? "touch ${prefix}.domtbl" : ''}
```
With all three false this renders a touch followed by backslashes joining blank
lines. It works today by luck. The stub now uses the same explicit list as the
script block.

No functional change

For every input combination the module accepts, the generated command line is
equivalent to before. Only the script's construction changed.

Why the ${to_gzip ? ... : ''} guard

to_gzip is unconditionally seeded with ${prefix}.txt, so it can never be empty
and the guard is always true today. It is deliberate insurance: a bare
gzip --no-name with no file arguments reads stdin, so the task would hang
forever rather than fail. That becomes reachable the moment anyone makes the main
output conditional — e.g. routing -o to /dev/null for some flag combination,
which would leave the list empty whenever write_align/write_target/
write_domain are all false. The guard costs nothing and closes the footgun in
advance.

## PR checklist

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:


modules/nf-core/hmmer/hmmsearch/main.nf
```diff
@@ -38,23 +42,18 @@ process HMMER_HMMSEARCH {
         $hmmfile \\
         $seqdb
 
-    gzip --no-name *.txt \\
-        ${write_align ? '*.sto' : ''} \\
-        ${write_target ? '*.tbl' : ''} \\
-        ${write_domain ? '*.domtbl' : ''}
+    ${to_gzip ? "gzip --no-name ${to_gzip.join(' ')}" : ''}
```
mashehu, line 45: isn't that always true?

vagkaratzas: For now yes. but if somebody (like me), decides to route -o to /dev/null to save storage in the future, that would be empty and fail. Just a free extra guard

modules/nf-core/hmmer/hmmsearch/tests/main.nf.test
```diff
@@ -59,21 +56,15 @@ nextflow_process {
...
-                { assert path(process.out.target_summary.get(0).get(1)).linesGzip.toString().contains('[ok]') },
-                { assert snapshot(
-                    process.out.alignments,
-                    path(process.out.output[0][1]).linesGzip.size(),
-                    path(process.out.target_summary[0][1]).linesGzip.size(),
-                    path(process.out.domain_summary[0][1]).linesGzip.size(),
-                    process.out.findAll { key, val -> key.startsWith("versions") }
-                ).match() }
+                // "output" carries the CPU-time footer and "target_summary"/"domain_summary" carry a
+                // "Current dir"/"Date" comment block, all run-dependent; "alignments" is stable, no key needed
```
mashehu:
```suggestion
                // "Current dir"/"Date" comment block, all run-dependent;
```

mashehu (approved):

PR #12435, opened 2026-07-27
fix(gtf_hybridmerge_gffcompare): anchor withName selectors

pinin4fjords:
## Problem

`gtf_hybridmerge_gffcompare/nextflow.config` uses unanchored `withName` selectors of the form `GTF_HYBRIDMERGE_GFFCOMPARE:GAWK_FILTER`. Nextflow full-string matches `withName` against the fully qualified process name, so these match when the subworkflow is invoked directly, as in its own nf-test, but never inside a consumer pipeline where the name is prefixed by the calling chain. The config block then silently never applies, with no warning or error.

Because the bundled tests invoke the subworkflow at the top level, this class of bug is invisible to nf-core/modules CI: lint and nf-test both pass with or without the fix.

## Impact

In nf-core/riboseq, which nests this several levels deep, `GAWK_FILTER` never received `-v codes_csv=...`, so the awk filter ran with an empty keep-set and emitted nothing. The hybrid GTF silently contained zero novel transcripts while the pipeline completed cleanly.

Confirmed with a three-level probe pipeline on Nextflow 26.04.6. Before the fix the merged GTF had an identical md5 across three runs seeded with three different novel-transcript inputs; after, all three differ.

## Fix

Anchor all three selectors (`GAWK_FILTER`, `BEDTOOLS_INTERSECT`, `GAWK_CONCAT`) with a leading `.*` so they match standalone and nested.

`nf-core subworkflows lint`: 28/28 passed. `nf-test`: both tests still pass, confirming standalone invocation is unaffected.

## Scale of the pattern

A sweep of every `subworkflows/nf-core/*/nextflow.config` found one other component with the identical defect, `orftable_fasta_gtf_buildorfcatalogue`, addressed separately in #12434. Everything else is either already anchored or selects on unqualified simple names, which match at any nesting depth.

nf-core/riboseq carries this as a local patch pending this landing.

## PR checklist

- [x] `nf-core subworkflows test gtf_hybridmerge_gffcompare --profile docker`
- [ ] `nf-core subworkflows test gtf_hybridmerge_gffcompare --profile singularity`
- [ ] `nf-core subworkflows test gtf_hybridmerge_gffcompare --profile conda`

### DIFF
--- a/subworkflows/nf-core/gtf_hybridmerge_gffcompare/nextflow.config
+++ b/subworkflows/nf-core/gtf_hybridmerge_gffcompare/nextflow.config
@@ -10,6 +10,10 @@
  * `process_single`. Fine for annotation-scale GTFs; for >~500 MB inputs
  * override `memory` / `cpus` directly via `withName`.
  *
+ * Selector form: Nextflow matches `withName` as a full-string regex against
+ * the fully qualified process name, so the leading `.*` is required for these
+ * blocks to apply inside a consumer pipeline as well as standalone.
+ *
  * Annotation format: the bundled awks assume standard GTF2 with `gene_id`
  * and `transcript_id` attributes in column 9 and `gene` / `transcript` in
  * column 3. GFF3 inputs (Parent= / ID= attributes), prokaryotic GTFs
@@ -20,7 +24,7 @@ process {
     // GAWK_FILTER (awk/filter.awk): keeps the records belonging to
     // transcripts whose gffcompare class_code is in the caller-supplied set
     // passed through `meta.class_codes`.
-    withName: 'GTF_HYBRIDMERGE_GFFCOMPARE:GAWK_FILTER' {
+    withName: '.*GTF_HYBRIDMERGE_GFFCOMPARE:GAWK_FILTER' {
         ext.args   = { "-v codes_csv='${meta.class_codes}'" }
         ext.prefix = { "${meta.id}_filtered" }
         ext.suffix = 'gtf'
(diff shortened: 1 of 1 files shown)


subworkflows/nf-core/gtf_hybridmerge_gffcompare/nextflow.config
```diff
@@ -10,6 +10,10 @@
  * `process_single`. Fine for annotation-scale GTFs; for >~500 MB inputs
  * override `memory` / `cpus` directly via `withName`.
  *
+ * Selector form: Nextflow matches `withName` as a full-string regex against
+ * the fully qualified process name, so the leading `.*` is required for these
+ * blocks to apply inside a consumer pipeline as well as standalone.
+ *
```
mashehu:
```suggestion
```

mashehu (approved):

subworkflows/nf-core/gtf_hybridmerge_gffcompare/nextflow.config
mashehu, line 1: also why does this file exist? these things should be handled inside the module imo

mashehu (changes requested):

subworkflows/nf-core/gtf_hybridmerge_gffcompare/nextflow.config
pinin4fjords: I would have removed this before un-drafting :-)

subworkflows/nf-core/gtf_hybridmerge_gffcompare/nextflow.config
pinin4fjords: This config exists to show pipelines consuming this subworkflow how they need to configure the modules (which could be configured differently in different pipelines). So I'm not sure what you mean.

mashehu: non-standard config files are bad documentation. You can be either more specific in the meta.yml or use the tests for this. Also, without knowing the tools too well, these setting look quite common nf-core style, no?

pinin4fjords: Test configurations are a distinct thing, and not necessarily the same as the 'this is how you should set the config up under normal usage'. I don't agree that meta.yml files would a better way of documenting how subworkflow authors intend that subworkflow to be configured. 

Perhaps we can discuss the pattern of using configs like this in wider discussion, for now I'd just like to make sure this one isn't broken.

mashehu: meta.yml's get rendered on the website, so have a larger discovery surface then a config.
I will add it to the agenda for maintainer meetings.

mashehu (approved):

PR #12436, opened 2026-07-27
fix(custom/orfnormalise): resolve PRICE gene ids from the annotation

pinin4fjords:
`parse_price` took `gene_id` from PRICE's `Gene` column first and only consulted the annotation when that column was empty. PRICE concatenates every gene overlapping an ORF's genomic span with an underscore, so multi-gene ORFs came out with ids like `GENEA_GENEB`, which are meaningless downstream and will not join against a single-gene count matrix. This flips the precedence: resolve from the GTF transcript-to-gene mapping first, fall back to the `Gene` column only when the transcript is absent from the annotation.

It also fixes the `Id` parse. PRICE builds `Id` as `<transcript_id>_<orf_type>_<n>`, and `split("_", 1)[0]` assumed the transcript id has no underscore, which holds for Ensembl but not for every annotation. A new `_price_transcript_id` helper anchors the split on the row's own `orf_type` value.

### Verified

- The existing `homo_sapiens [chr19+chr22] - price` case runs against the real `cohort.price.orfs.tsv` in nf-core/test-datasets and passes an empty annotation, so every row takes the `Gene`-column fallback and its snapshot entry is byte identical after the change. That path is worth keeping: PRICE without a GTF is a supported mode, and 9 of that fixture's 15 rows carry a concatenated `Gene` value, so it pins the fallback behaviour precisely.
- Because that case supplies no annotation, it cannot reach the transcript-first resolution this PR adds. The new case is the one that exercises it: a row whose `Gene` column holds a concatenated id and whose transcript, itself underscore-containing, is present in a small GTF, asserting the plain gene id from the GTF wins. Its inputs are built with `collectFile` rather than shipped fixture files, matching how other modules handle small synthetic inputs; no module in this repo keeps data fixtures in its `tests/` directory.
- Full module suite passes on x86_64 with nf-test 0.9.5 and Docker, 7 tests. The snap diff is 49 insertions and no deletions.
- `nf-core modules lint custom/orfnormalise`: 59 passed, 0 failed. The two warnings (container registry 404, container version mismatch) are pre-existing and untouched.
- `ruff check`, `ruff format`, prettier and the meta.yml schema check are clean.

### Other parsers

I checked the other four parsers in the same template for the same precedence mistake. `parse_ribotish` has the same shape (tool column first, annotation fallback), but Ribo-TISH's `Gid` is a single gene id, so there is nothing to fix. `parse_ribocode` and `parse_ribotricer` read a single-gene `gene_id` column with no annotation fallback, and `parse_rpbp` resolves from the annotation only. PRICE is the only caller that emits a concatenated multi-gene value.

Consumed by nf-core/riboseq, which currently carries this as a local patch.

### DIFF
--- a/modules/nf-core/custom/orfnormalise/meta.yml
+++ b/modules/nf-core/custom/orfnormalise/meta.yml
@@ -46,6 +46,10 @@ description: |
 
   Per-caller mapping notes (lossy collapses):
 
+    - PRICE's `Gene` column concatenates every gene overlapping an ORF's
+      genomic span, so `gene_id` is resolved from the GTF via the ORF's
+      transcript id; the `Gene` column is used only for transcripts absent
+      from the annotation, and can then carry a multi-gene value.
     - PRICE `iORF` (internal ORF), `intronic`, and `orphan` map to `other`.
       Cross-caller catalogue tracking still flags these via
       `called_by_price`, but the specific PRICE sub-type is not preserved.
--- a/modules/nf-core/custom/orfnormalise/templates/orfnormalise.py
+++ b/modules/nf-core/custom/orfnormalise/templates/orfnormalise.py
@@ -766,6 +766,24 @@ def _price_score_from_p(p):
     return clamp1000(min(-math.log10(p) * 100, 1000))
 
 
+def _price_transcript_id(orf_id_raw, orf_type):
+    """Recover the transcript id from PRICE's `Id` column.
+
+    PRICE builds `Id` as `<transcript_id>_<orf_type>_<n>` (an increasing
+    counter within the transcript). The split anchors on the row's own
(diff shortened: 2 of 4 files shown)


adamrtalbot (approved):
Transcript -> gene generally safer so this is a wise move.

Do we still need the old test? It's not causing any harm but we have a dual mode test where one of the two tests requires test-datasets and one of them doesn't.

pinin4fjords:
Yeah, they cover different things. The test-datasets one passes an empty annotation, so every row falls back to the Gene column. Worth keeping, since PRICE without a GTF is a supported mode and 9 of those 15 rows have a concatenated Gene. But it can't reach the transcript lookup this PR adds, which is why the second test brings its own small GTF.

PR #12438, opened 2026-07-27
fix(ribotish/predict): pin python=3.13.9 to avoid Python 3.14 forkserver default

pinin4fjords:
Two changes to `ribotish/predict`.

**Pin `conda-forge::python=3.13.9`**, matching the biocontainers image. `predict.py` sets per-run state in module-level globals inside `run()` and reads them back in `_pred_gene`, mapped across a `multiprocessing.Pool` when `-p` is above 1. That relies on `fork` inheriting the parent's globals. Python 3.14 switched the default start method on non-macOS POSIX to `forkserver`, whose workers re-import the module and never see them, so predict fails with `NameError: name 'genomefapath' is not defined`. Only conda is exposed: `ribotish=0.2.8` requires just `python >=3`, while the container is fixed at 3.13.9. Exact pin rather than `<3.14` because `environment-schema.json` rejects `<` and `>`. The underlying bug is Ribo-TISH's own use of globals across a process pool, which nothing on their tracker covers; the pin only avoids triggering it.

**Hash `transprofile` over sorted lines.** ribotish emits rows through `imap_unordered`, so the raw file md5 differs between runs while the content does not. The snapshot value is a hash of the sorted lines rather than of the file as written. Not caused by the pin: it fails on docker and singularity too.

`ribotish/quality` has no `multiprocessing.Pool` usage and is untouched.

Verified: `nf-core modules lint` clean, all four tests pass on repeated runs, container confirmed at Python 3.13.9.

Consumed by nf-core/riboseq.

### DIFF
--- a/modules/nf-core/ribotish/predict/tests/main.nf.test.snap
+++ b/modules/nf-core/ribotish/predict/tests/main.nf.test.snap
@@ -28,7 +28,8 @@
                         "single_end": true,
                         "strandedness": "forward"
                     },
-                    "test_transprofile.py:md5,d41d8cd98f00b204e9800998ecf8427e"
+                    "test_transprofile.py",
+                    "d41d8cd98f00b204e9800998ecf8427e"
                 ]
             ],
             {
@@ -41,10 +42,10 @@
                 ]
             }
         ],
-        "timestamp": "2026-05-19T10:03:51.7008198",
+        "timestamp": "2026-07-27T12:46:49.440056573",
         "meta": {
             "nf-test": "0.9.5",
-            "nextflow": "26.04.1"
+            "nextflow": "26.04.6"
         }
     },
     "sarscov2 [bam] - single_end - multi ribo bam - stub": {
@@ -76,7 +77,8 @@
                         "single_end": true,
                         "strandedness": "forward"
                     },
-                    "test_transprofile.py:md5,d41d8cd98f00b204e9800998ecf8427e"
+                    "test_transprofile.py",
+                    "d41d8cd98f00b204e9800998ecf8427e"
                 ]
             ],
             {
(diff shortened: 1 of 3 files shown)


modules/nf-core/ribotish/predict/tests/main.nf.test.snap
```diff
@@ -28,7 +28,8 @@
                         "single_end": true,
                         "strandedness": "forward"
                     },
-                    "test_transprofile.py:md5,d41d8cd98f00b204e9800998ecf8427e"
+                    "test_transprofile.py",
+                    "d41d8cd98f00b204e9800998ecf8427e"
```
mashehu: what's the benefit of splitting it into two lines?

adamrtalbot: This is nf-test behaviour where it's gone from one string to a collection.

I don't think joining them back together to a string will add anything.

adamrtalbot: It's gone to a collection because a `sort` has been added (explained in PR description).

adamrtalbot (approved):
How frustrating. 

> The underlying bug is Ribo-TISH's own unsafe use of globals across a process pool

Offending code is [here](https://github.com/zhpn1024/ribotish/blob/d721b7a742e144df2a3d3ce92e3114e685f8a8a1/src/run/predict.py#L73-L76) and [here](https://github.com/zhpn1024/ribotish/blob/d721b7a742e144df2a3d3ce92e3114e685f8a8a1/src/run/predict.py#L228-L232). Last updated 9 months ago, you could open a bug report, it may be ignored but it's the good citizen thing to do.

For now this will fix the issue.

modules/nf-core/ribotish/predict/tests/main.nf.test.snap
mashehu: would prefer if it stayed one line like the rest.

mashehu (approved):

PR #12439, opened 2026-07-27
fix(fasta_gtf_bam_rpbp): require value channel for reference input

pinin4fjords:
## Summary

`ch_fasta_gtf` is paired inside this subworkflow with the per-sample BAM channel (`ch_bam`), so it must be a value channel. If a caller instead passes a single-item queue channel, Nextflow consumes it once and the fan-out silently runs for only one sample, with no error and no warning. This subworkflow's own nf-test can't catch that, since it only exercises a single sample, so the hazard survives until a caller with multiple samples hits it.

The `take:` declaration now states the value-channel requirement explicitly. The subworkflow previously worked around ambiguity here by calling `.first()` on channels derived from `ch_fasta_gtf` inside `main:` (`RPBP_PREPAREGENOME.out.transcript_bed.first()` and friends). Those calls are removed: once the input is genuinely a value channel, Nextflow already makes single-value-input process outputs value channels, so the internal `.first()` calls were redundant, and they mask rather than prevent a caller violation of the contract.

The nf-test now passes `.first()` on the channel supplied for `ch_fasta_gtf`, so the test itself supplies a real value channel per the new contract.

## Context

This is being upstreamed from a local patch in [nf-core/riboseq](https://github.com/nf-core/riboseq), where this class of defect (a reference channel silently treated as single-shot) caused a per-sample alignment step to emit results for only one of six samples, with content varying between runs.

## Test plan

- [x] `nf-core subworkflows lint fasta_gtf_bam_rpbp` - clean (37/37 passed, 0 warnings)
- [x] `nf-test test subworkflows/nf-core/fasta_gtf_bam_rpbp/tests/main.nf.test --profile docker` on a Linux x86_64 VM - both tests pass, including the full (non-stub) chr20 end-to-end Rp-Bp run, with no snapshot changes

### DIFF
--- a/subworkflows/nf-core/fasta_gtf_bam_rpbp/main.nf
+++ b/subworkflows/nf-core/fasta_gtf_bam_rpbp/main.nf
@@ -11,16 +11,18 @@ workflow FASTA_GTF_BAM_RPBP {
 
     take:
     ch_bam        // channel: [ val(meta), path(bam), path(bai) ] - Ribo-seq BAMs
-    ch_fasta_gtf  // channel (single value): [ val(meta), path(fasta), path(gtf) ]
+    ch_fasta_gtf  // value channel: [ val(meta), path(fasta), path(gtf) ]. Must be a value channel:
+                  //                the annotation products derived from it are paired with the
+                  //                per-sample BAM channel below.
 
     main:
 
     RPBP_PREPAREGENOME(ch_fasta_gtf)
 
-    ch_transcript_bed   = RPBP_PREPAREGENOME.out.transcript_bed.first()
-    ch_orfs_genomic_bed = RPBP_PREPAREGENOME.out.orfs_genomic_bed.first()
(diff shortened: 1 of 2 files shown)


pinin4fjords:
Closing this. On review the change was a net loss of robustness rather than a fix.

The `.first()` calls it removed on `RPBP_PREPAREGENOME.out.*` and the `ch_genome_fasta` derivation were what let the subworkflow tolerate a caller passing a single-item queue channel for `ch_fasta_gtf`. Replacing them with a `take:` comment enforces nothing: a caller that gets it wrong would silently process one sample and drop the rest, which is exactly the failure the change claimed to guard against.

The tell was in the test. It already passed a queue channel and worked because of those `.first()` calls; the PR had to add `.first()` on the caller side to compensate, which is adjusting the test to fit a weakened contract.

`meta.yml` already documents `ch_fasta_gtf` as a single-value channel, so with the code reverted there is nothing left to upstream here.

PR #12451, opened 2026-07-28
fix(gedi): make the genome index self-contained and relocatable

pinin4fjords:
## Description

`gedi/indexgenome` wrote an index that only resolves in the directory it was built in, so PRICE fails on it under AWS Batch with Fusion.

GEDI records the `-s` path in the `.fi` it writes: a bare relative name when the FASTA is co-located with the `.fi`, an absolute path otherwise. Passing a staged input at the task work directory root therefore recorded an absolute path to a file outside the emitted index, which Fusion never materialises, and PRICE died in `PriceCodonInference` with `Could not read sequence from file`. Indexing from a copy inside the index directory makes GEDI record a relative name that travels with the index.

That leaves the `.oml`, whose absolute member paths are fixed at build time, so `gedi/price` now rewrites them onto the staged index, into the task directory since writing through a staged directory lands back at the upstream task's storage prefix under Fusion. Together the index becomes relocatable: verified by running PRICE against a copied-out index whose indexing work directory had been deleted.

The `indexgenome` half is @FelixKrueger's fix from nf-core/riboseq#217, verified there before being brought up here.

### One thing for reviewers

`gedi/price`'s snapshot pinned the `orfs.tsv` row count, and that count is not reproducible across environments: this PR's first CI run gave 370 rows under docker, 374 under conda and 379 under singularity, against the committed 377. It also tracks thread count (379 at 1 thread, 377 at 2, 370 at 4, 381 at 8), and is stable only within a fixed machine, engine and thread count, which is how the pinned value passed when the module was added. Running single-threaded to force determinism takes 20 minutes against 5 and still leaves the value machine-dependent.

So the last commit swaps that assertion for the `orfs.tsv` header columns plus a row count between 300 and 450. Happy to tighten the band or drop it if you'd rather.

Separately, and not something this PR fixes: PRICE is flaky on this fixture. One of eight runs here died in `PriceOrfInference` with `Exception in parallel iterator thread! / by zero`, having written its codon output first. Worth knowing that this test will fail occasionally for reasons unrelated to whatever is being changed.

## PR checklist

- [x] Add a resource `label`
- [x] `nf-core modules lint` passes for both modules (0 failures, warning counts unchanged from master).


FelixKrueger (approved):
LGTM

FelixKrueger (approved):
Nice fix, Jon — the two-pronged approach (make indexgenome write a relative `-s`, rewrite the `.oml` paths at read time in price) is exactly right for the Fusion path-baking problem. Went through both changes; happy to approve.

Two non-blocking notes, nothing that should hold this up:
- `indexgenome` now copies the FASTA into the index dir, so the published index carries a full FASTA copy — doubles its size for big genomes. Fair cost for relocatability, just flagging it.
- The `price` sed (`[^"]*/`) flattens every `.oml` member to a basename under `$PWD/${index}/`. Works because the members are flat in the index dir today; it'd silently mis-point if an index ever nested them. Not worth changing now.

Cheers!

modules/nf-core/gedi/price/tests/main.nf.test
```diff
@@ -78,9 +78,17 @@ nextflow_process {
...
                     process.out.signal.collect        { meta, f -> [meta, file(f).name] },
                     process.out.param.collect         { meta, f -> [meta, file(f).name] },
-                    process.out.orfs_tsv.collect      { meta, f -> [meta, file(f).name, path(f).readLines().size()] },
+                    process.out.orfs_tsv.collect      { meta, f -> [meta, file(f).name] },
                     process.out.findAll { key, val -> key.startsWith('versions') }
-                ).match() }
+                ).match() },
+                // PRICE's ORF count is not reproducible across machines: the same fixture
+                // gives 370 under docker, 374 under conda and 379 under singularity in one
+                // CI run. Assert the format and the order of magnitude instead.
```
mashehu:
```suggestion
                ).match() },
                // PRICE's ORF count is not reproducible
```
please add to your claude.md to be more concise in the comments

modules/nf-core/gedi/price/main.nf
```diff
@@ -33,9 +33,14 @@ process GEDI_PRICE {
     ls -1 bams/*.bam > price_input.bamlist
     bamlist2cit -n ${task.cpus} -p price_input.bamlist
 
+    # The .oml holds absolute paths to its members, fixed when the index was built, so an
+    # index read from anywhere other than where it was built needs them pointed at the
+    # staged copy. Rewrite into the task directory: the staged index must not be written to.
```
mashehu: make this less wordy

mashehu (changes requested):

modules/nf-core/gedi/price/tests/main.nf.test
FelixKrueger: I have added instructions to reduce overly wordy comments while still making thinking visible

modules/nf-core/gedi/price/main.nf
FelixKrueger:
```suggestion
    # .oml member paths are absolute; repoint them at the staged index
```
Dropped the not-in-place note: the `>` redirect already shows it, and 39505626 explains the Fusion reason in full.

modules/nf-core/gedi/indexgenome/main.nf
```diff
@@ -23,8 +23,14 @@ process GEDI_INDEXGENOME {
     prefix   = task.ext.prefix ?: "${meta.id}"
     """
     mkdir -p ${prefix}
+
+    # GEDI records the -s path in the .fi it writes: a bare relative name when the FASTA
+    # is co-located with the .fi, an absolute path otherwise. Index from a copy inside the
+    # index directory so the reference resolves wherever the index is later read.
```
FelixKrueger:
```suggestion
    # GEDI records -s relative only when the FASTA sits beside the .fi
```
Not flagged, but the same three-lines-paraphrasing-the-commit-message shape as the other two — worth trimming in the same pass.

FelixKrueger (commented):
Concrete trims for the two remaining wordy comments — @mashehu's suggestion already covers the test file. In both cases the detail stays in the commit message (39505626, ba122237), so `git log -L` still recovers the full reasoning.

mashehu (approved):

PR #12453, opened 2026-07-28
docs(orfnormalise): correct the GTF contract for rpbp and price

pinin4fjords:
Doc-only. `custom/orfnormalise`'s `gtf` input is documented as unused for `rpbp` and `price`, with `[ [:], [] ]` offered as acceptable when the caller subset excludes `ribocode` and `ribotish`. Neither holds:

- `parse_rpbp` resolves `gene_id` as `transcripts[tid].gene_id`, so with no GTF every rpbp row gets an empty `gene_id`.
- `parse_price` resolves `gene_id` from the transcript id in the `Id` column (nf-core/modules#12436), falling back to PRICE's `Gene` column, which concatenates every gene the ORF's genomic span overlaps. With no GTF, every price row degrades to that compound value, which joins against nothing downstream.

There is also an unstated requirement worth writing down: the annotation has to be a superset of the one each caller was run against. A transcript absent from it cannot be resolved, and a pipeline that hands the normaliser a narrower annotation than its callers used (a one-transcript-per-gene backbone, say) silently loses gene ids for ORFs on the isoforms it dropped.

Corrected in the module's `meta.yml` and in the `orftable_fasta_gtf_buildorfcatalogue` subworkflow that passes the GTF through (`meta.yml` and the `take:` comment), which carried the same claim.

No behaviour change. `nf-core modules lint custom/orfnormalise` and `nf-core subworkflows lint orftable_fasta_gtf_buildorfcatalogue` pass (the two module warnings are pre-existing container-version notes).


modules/nf-core/custom/orfnormalise/meta.yml
```diff
@@ -145,13 +145,16 @@ input:
...
+          Annotation GTF. Every caller draws on it: ribocode requires it
+          (transcript-relative coordinates → genomic lift), ribotish and
+          ribotricer use it for multi-exon block recovery (from GenomePos
+          and the ORF_ID span respectively), and rpbp and price resolve
+          `gene_id` from the ORF's transcript id through it. Supply the
+          annotation the caller was run against, or a superset of it: a
+          transcript missing here cannot be resolved, and price then falls
+          back to its `Gene` column, which concatenates every gene the
+          ORF's genomic span overlaps. Passing `[ [:], [] ]` loses
+          multi-exon structure and, for rpbp and price, `gene_id`.
```
mashehu: can you please rewrite this in your own words, feels like marketing speak

mashehu (commented):
since it's docs only, can you please try to write them in your own words?

modules/nf-core/custom/orfnormalise/meta.yml
pinin4fjords: Made more concise

SPPearce (approved):

PR #12459, opened 2026-07-29
untarfiles to topics

SPPearce:
Swap untarfiles to topics.
Note this is deprecated, so no changes to the tests.


modules/nf-core/untarfiles/main.nf
```diff
@@ -12,8 +12,7 @@ process UNTARFILES {
 
     output:
     tuple val(meta), path("${prefix}/**") , emit: files
-    path "versions.yml"                   , emit: versions
-
+    tuple val("${task.process}"), val('tar'), eval("tar --version 2>&1 | sed 's/^.*(GNU tar) //; s/ Copyright.*\$//'"), topic: versions, emit: versions_tar
```
mashehu:
```suggestion
    tuple val("${task.process}"), val('tar'), eval("tar --version 2>&1 | sed '1s/^.*(GNU tar) //;q'"), topic: versions, emit: versions_tar
```
quits after the first match.

mashehu (approved):

PR #12467, opened 2026-07-30
New module: transrate

erikrikarddaniel:
## Description

Adds `transrate`, a reference-free and reference-based quality assessment tool for de novo
assemblies. It's currently used as a hand-written local module in
nf-core/metatdenovo. The [nf-core/metatdenovo#445](https://github.com/nf-core/metatdenovo/issues/445) issue asks for several local modules there to be swapped for official nf-core/modules
equivalents. This PR is the upstream half of that migration so metatdenovo can later switch
to the official module.

Notes on the implementation:

- Uses the most recent biocontainers build (`1.0.3--h87e0c26_7`); `1.0.3` remains the only
  bioconda release of transrate.
- `transrate` embeds the resolved absolute path of its input assembly file into its own CSV
  report, which made the output non-reproducible across task work directories and broke
  snapshot testing on repeated runs. The module post-processes the report with `sed` to
  replace that column with the sample prefix instead.
- Output renamed to a `multiqc`/`*_mqc.csv` emit, matching the existing nf-core convention for
  MultiQC custom-content outputs (following `dupradar`/`endorspy` precedent) rather than
  metatdenovo's original local naming.

Tested with the `docker` profile (both the main and stub test pass, stable across repeated
runs); not yet tested with `singularity` or `conda` profiles.

Generated by Claude Sonnet 5

## PR checklist

- [x] Add a resource `label`
  - For modules:

### DIFF
--- a/modules/nf-core/transrate/main.nf
+++ b/modules/nf-core/transrate/main.nf
@@ -31,12 +31,15 @@ process TRANSRATE {
         --output .
 
     mv assemblies.csv ${prefix}.assemblies.csv
-    mv */contigs.csv ${prefix}.contigs.csv
+    # transrate always writes contigs.csv into a subdirectory named after the assembly's
+    # own basename (not configurable via --output).
+    mv ${assembly.baseName}/contigs.csv ${prefix}.contigs.csv
 
     # transrate writes the resolved absolute path of the input assembly into the first
-    # column of the report; replace it with the sample prefix so the report content is
-    # reproducible across environments and task work directories.
-    sed -i "2s@^[^,]*@${prefix}@" ${prefix}.assemblies.csv
+    # column of the report; that path is task-work-directory-specific, so replace it with
+    # just the file's own name to keep the column meaningful without embedding a
+    # non-portable, non-reproducible absolute path.
+    sed -i "2s@^[^,]*@${assembly.name}@" ${prefix}.assemblies.csv
     """
 
     stub:
--- a/modules/nf-core/transrate/main.nf
+++ b/modules/nf-core/transrate/main.nf
@@ -3,15 +3,18 @@ process TRANSRATE {
     label 'process_low'
 
     conda "${moduleDir}/environment.yml"
-    container "${ workflow.containerEngine in ['singularity', 'apptainer'] && !task.ext.singularity_pull_docker_container ?
-        'https://depot.galaxyproject.org/singularity/transrate:1.0.3--h87e0c26_7':
-        'quay.io/biocontainers/transrate:1.0.3--h87e0c26_7' }"
+    container "${ workflow.containerEngine in ['singularity', 'apptainer'] && !task.ext.singularity_pull_docker_container
+?         'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/c3/c3f7b24e30eab0cde846c0d31556293afc7818492d016e6770a1fef561042405/data'
+:         'community.wave.seqera.io/library/transrate_blast:092d9b5f119b9733' }"
 
     input:
     tuple val(meta), path(assembly)
+    path reference
 
     output:
-    tuple val(meta), path("*_mqc.csv")                                              , emit: multiqc
+    tuple val(meta), path("*.assemblies.csv")                                       , emit: assemblies
+    tuple val(meta), path("*.contigs.csv")                                          , emit: contigs
+    tuple val(meta), path("*_mqc.csv")                                              , emit: multiqc, optional: true
     tuple val("${task.process}"), val('transrate'), eval("transrate --version"), topic: versions, emit: versions_transrate
 
     when:
@@ -20,24 +23,35 @@ process TRANSRATE {
     script:
     def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
+    // Reference-based comparative metrics (blast+, packageable). Read-based metrics
+    // (--left/--right) are deliberately not supported: they hard-require the `bam-read`
+    // binary, which isn't published on any conda channel (only as a prebuilt GitHub
+    // release from Blahah/transrate-tools), and their aligner (snap-aligner, pinned to
+    // an old dev build) segfaults building an index for at least small test genomes.
+    def reference_arg = reference ? "--reference ${reference}" : ''
     """
     transrate \\
         $args \\
         --threads $task.cpus \\
         --assembly $assembly \\
-        --output ${prefix}_transrate
+        ${reference_arg} \\
+        --output .
 
-    mv ${prefix}_transrate/assemblies.csv ${prefix}_mqc.csv
+    mv assemblies.csv ${prefix}.assemblies.csv
+    mv */contigs.csv ${prefix}.contigs.csv
 
     # transrate writes the resolved absolute path of the input assembly into the first
     # column of the report; replace it with the sample prefix so the report content is
     # reproducible across environments and task work directories.
-    sed -i "2s@^[^,]*@${prefix}@" ${prefix}_mqc.csv
+    sed -i "2s@^[^,]*@${prefix}@" ${prefix}.assemblies.csv
+
+    cp ${prefix}.assemblies.csv ${prefix}_mqc.csv
     """
 
     stub:
     def prefix = task.ext.prefix ?: "${meta.id}"
     """
-    touch ${prefix}_mqc.csv
+    touch ${prefix}.assemblies.csv
+    touch ${prefix}.contigs.csv
     """
 }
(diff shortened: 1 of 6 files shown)


vagkaratzas:
I had attempted to port this some time ago: https://github.com/nf-core/modules/tree/transrate

There are a bit more inputs than just the assembly fasta (--left, --right reads, --reference). So those inputs also need to be there.

I don't remember why I dropped it, but there was something that was giving me pain. But that was before the bots, so better luck now I hope!

SPPearce (commented):
I don't understand why this is making a file called `_mqc.csv`, when it should just be emiting the file as is. If you want, you can make that file as well, but I think we should be emiting the actual file.

Also can you set the output directory to be the work directory, instead of nested?
Does it make more files that the one?

erikrikarddaniel:
> I had attempted to port this some time ago: https://github.com/nf-core/modules/tree/transrate
> 
> There are a bit more inputs than just the assembly fasta (--left, --right reads, --reference). So those inputs also need to be there.
> 
There's apparently no package that includes the `bam-read` binary -- only available as a pre-built binary -- that's needed for `--left` and `--right`, so we're not building support for that. 

> I don't remember why I dropped it, but there was something that was giving me pain. But that was before the bots, so better luck now I hope!

I hope too! :-)

modules/nf-core/transrate/meta.yml
```diff
@@ -0,0 +1,120 @@
...
+            - edam: "http://edamontology.org/format_3752" # CSV
+  multiqc:
+    - - meta:
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. `[ id:'sample1' ]`
+      - "*_mqc.csv":
+          type: file
+          description: Optional copy of the assembly summary, named for MultiQC custom-content auto-detection. Only produced when `task.ext.make_mqc` is truthy.
```
vagkaratzas: Shouldn't this `task.ext.make_mqc` be defined somewhere? 
It's not a flag of the actual tool, is it?

modules/nf-core/transrate/main.nf
```diff
@@ -0,0 +1,57 @@
...
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    // Reference-based comparative metrics (blast+, packageable). Read-based metrics
+    // (--left/--right) are deliberately not supported: they hard-require the `bam-read`
+    // binary, which isn't published on any conda channel (only as a prebuilt GitHub
+    // release from Blahah/transrate-tools), and their aligner (snap-aligner, pinned to
+    // an old dev build) segfaults building an index for at least small test genomes.
```
vagkaratzas: meta.yaml description should be enough to explain this. No need to haev here as well

modules/nf-core/transrate/main.nf
```diff
@@ -0,0 +1,57 @@
...
+    label 'process_low'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine in ['singularity', 'apptainer'] && !task.ext.singularity_pull_docker_container
+?         'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/c3/c3f7b24e30eab0cde846c0d31556293afc7818492d016e6770a1fef561042405/data'
+:         'community.wave.seqera.io/library/transrate_blast:092d9b5f119b9733' }"
+
+    input:
+    tuple val(meta), path(assembly)
+    path reference
```
vagkaratzas: deliberately as second element instead of in the tuple above?

modules/nf-core/transrate/main.nf
```diff
@@ -0,0 +1,57 @@
...
+
+    mv assemblies.csv ${prefix}.assemblies.csv
+    mv */contigs.csv ${prefix}.contigs.csv
+
+    # transrate writes the resolved absolute path of the input assembly into the first
+    # column of the report; replace it with the sample prefix so the report content is
+    # reproducible across environments and task work directories.
+    sed -i "2s@^[^,]*@${prefix}@" ${prefix}.assemblies.csv
+
+    cp ${prefix}.assemblies.csv ${prefix}_mqc.csv
```
vagkaratzas: I would prefer this to happen in your pipeline, as a `mv` instead. The counter argument would be, we should have _mqc support for all output CSVs across all modules.

vagkaratzas (changes requested):
Left some comments

modules/nf-core/transrate/meta.yml
erikrikarddaniel: Good catch — that was stale documentation describing a flag that was never actually implemented; the `_mqc.csv` copy was unconditional in the script. Removed the whole `multiqc` output/emit entirely (see the other thread on line 48). Renaming to `*_mqc.csv` for MultiQC custom-content pickup is now left to the consuming pipeline instead.

Comment added by Claude Sonnet 5

modules/nf-core/transrate/main.nf
erikrikarddaniel: Agreed, removed — the top-level `meta.yml` `description` already covers this.

Comment added by Claude Sonnet 5

modules/nf-core/transrate/main.nf
erikrikarddaniel: Fair point, folded `reference` into the main tuple (`tuple val(meta), path(assembly), path(reference)`) as suggested.

Comment added by Claude Sonnet 5

modules/nf-core/transrate/main.nf
erikrikarddaniel: Removed the `_mqc.csv` output/copy entirely rather than gating it — agreed this is better handled in the consuming pipeline than baked into the module.

Comment added by Claude Sonnet 5

erikrikarddaniel:
Thanks for your comments @vagkaratzas! As you can see, I let my programming partner reply. I'm still learning how to work with it. So far I'm happy with its performance, but not always with my own. I should have seen the `*_mqc.csv` thing myself (a left-over from the local module) but didn't. So, thanks for your patience.

modules/nf-core/transrate/main.nf
```diff
@@ -0,0 +1,48 @@
...
+    """
+    transrate \\
+        $args \\
+        --threads $task.cpus \\
+        --assembly $assembly \\
+        ${reference_arg} \\
+        --output .
+
+    mv assemblies.csv ${prefix}.assemblies.csv
+    mv */contigs.csv ${prefix}.contigs.csv
```
vagkaratzas: are the contigs under folder(s)? Make sure the `*/` here makes sense

modules/nf-core/transrate/main.nf
```diff
@@ -0,0 +1,48 @@
...
+
+    # transrate writes the resolved absolute path of the input assembly into the first
+    # column of the report; replace it with the sample prefix so the report content is
+    # reproducible across environments and task work directories.
+    sed -i "2s@^[^,]*@${prefix}@" ${prefix}.assemblies.csv
```
vagkaratzas: Also make sure that this is the wanted behaviour and not another agent artifact.
Shouldn't the user prefer the path, as was intended by the original software?

vagkaratzas (commented):
No worries! :D
Two more clarifying questions and should be good to go

modules/nf-core/transrate/main.nf
erikrikarddaniel: Verified empirically (docker run against the real container, both with and without --reference): transrate always writes contigs.csv into exactly one subdirectory, named after the assembly's own basename — not configurable via --output, and unaffected by whether a reference/blast db is used (those land in the same subdirectory too, not separate ones). So the wildcard wasn't actually ambiguous in practice, but I agree it was needlessly implicit — replaced with an explicit `${assembly.baseName}/contigs.csv`.

Comment added by Claude Sonnet 5

modules/nf-core/transrate/main.nf
erikrikarddaniel: Fair challenge. Checked, and the previous version did lose real information — it replaced the whole path with the pipeline-assigned prefix, discarding the actual filename. Changed it to replace the path with just `${assembly.name}` (the file's own name) instead: this still strips the non-portable, task-work-directory-specific absolute path (the actual reproducibility problem for snapshot testing), while keeping the meaningful value transrate itself intended to report.

Comment added by Claude Sonnet 5

erikrikarddaniel: I have thought about this and agree with Claude here. One could perhaps argue that it would be useful to know the work directory, but if this file is published and the work directory is removed, I think it's just confusing while the file path should be clear.

vagkaratzas (approved):
LGTM!

PR #12340, opened 2026-07-09
Bump Samtools to v1.24

matthdsm:
## PR checklist

Closes #XXX 

- [ ] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, include test data in your PR.
- [ ] Remove all TODO statements.
- [ ] Broadcast software version numbers to `topic: versions` - [See version_topics](https://nf-co.re/blog/2025/version_topics)
- [ ] Follow the naming conventions.
- [ ] Follow the parameters requirements.
- [ ] Follow the input/output options guidelines.
- [ ] Add a resource `label`
- [ ] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - For modules:
    - [ ] `nf-core modules test <MODULE> --profile docker`
    - [ ] `nf-core modules test <MODULE> --profile singularity`
    - [ ] `nf-core modules test <MODULE> --profile conda`
  - For subworkflows:
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile docker`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile singularity`
    - [ ] `nf-core subworkflows test <SUBWORKFLOW> --profile conda`

PR #12468, opened 2026-07-30
Fix contract comments in orftable_fasta_gtf_buildorfcatalogue and bam_dedup_umi

FelixKrueger:
Two subworkflow contract comments say the opposite of what the code does, and both have already caused bugs downstream in nf-core/riboseq. `orftable_fasta_gtf_buildorfcatalogue` documents the GTF as ignored by rpbp and price, but both resolve gene ids from it and rpbp has no fallback. `bam_dedup_umi` documents `multiqc_files` as bare files, but it emits `[ meta, file ]`. Comments and `meta.yml` only, no functional change.

<details>
<summary>Detail (AI-assisted analysis)</summary>

## `orftable_fasta_gtf_buildorfcatalogue`

Which parsers in `custom/orfnormalise` read the GTF to resolve `gene_id`:

| parser | reads the GTF? | line |
| --- | --- | --- |
| `parse_ribocode` | yes | 473 |
| `parse_ribotish` | yes, after the `Gid` column | 533 |
| `parse_ribotricer` | no — takes `transcripts`, never reads it | 618 |
| `parse_rpbp` | yes, and it is the only source | 715 |
| `parse_price` | yes, `Gene` column as fallback | 807-809 |

`parse_rpbp` is `gid = transcripts[tid].gene_id if tid in transcripts else ""`, so without a GTF every rpbp ORF gets an empty `gene_id`. `parse_price` started preferring the annotation in #12436.

That also made the old `[ [:], [] ]` advice unsafe: with rpbp enabled it gives empty gene ids, and with price enabled it gives PRICE's `Gene` column, which concatenates every gene an ORF's genomic span overlaps.

Downstream: nf-core/riboseq fed the catalogue its one-transcript-per-gene backbone on the strength of the old wording, dropping PRICE ORFs on non-representative isoforms out of the ORF-level DTE join (nf-core/riboseq#203, fixed in nf-core/riboseq#207).

## `bam_dedup_umi`

`ch_multiqc_files` mixes `[ meta, path ]` module outputs and ends with `.transpose()` (L117-121), so every emission is a two-element tuple. Mixing it straight into a MultiQC input channel passes the meta map where a path is expected:

```
Not a valid path value type: java.util.LinkedHashMap
```

That broke every `--with_umi` run in nf-core/riboseq#220, fixed there with `.map { _meta, file -> file }` at the call site. `genomic_dedup_log` from the same source channel is annotated correctly two lines up, so this looks like an oversight rather than an intended contract. Fixing the comment rather than the emission avoids breaking existing consumers.

</details>

### DIFF
--- a/subworkflows/nf-core/bam_dedup_umi/main.nf
+++ b/subworkflows/nf-core/bam_dedup_umi/main.nf
@@ -143,7 +143,7 @@ workflow BAM_DEDUP_UMI {
     tsv_edit_distance              = ch_tsv_edit_distance // channel: [ val(meta), path(tsv) ]
     tsv_per_umi                    = ch_tsv_per_umi // channel: [ val(meta), path(tsv) ]
     tsv_umi_per_position           = ch_tsv_umi_per_position // channel: [ val(meta), path(tsv) ]
-    multiqc_files                  = ch_multiqc_files // channel: file
+    multiqc_files                  = ch_multiqc_files // channel: [ val(meta), path(file) ]
     transcriptome_bam              = ch_dedup_transcriptome_bam // channel: [ val(meta), path(bam) ] - final output
     transcriptome_dedup_bam        = UMI_DEDUP_TRANSCRIPTOME.out.bam // channel: [ val(meta), path(bam) ] - after dedup, before name sort
     transcriptome_sorted_bam       = SAMTOOLS_SORT.out.bam // channel: [ val(meta), path(bam) ] - name-sorted
--- a/subworkflows/nf-core/bam_dedup_umi/meta.yml
+++ b/subworkflows/nf-core/bam_dedup_umi/meta.yml
@@ -213,6 +213,9 @@ output:
   - multiqc_files:
       description: Channel containing files for MultiQC
       structure:
+        - meta:
+            type: map
+            description: Metadata map
         - file:
             type: file
             description: File for MultiQC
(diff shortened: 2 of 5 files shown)


SPPearce:
Your meta changes look fine, but there is something unstable about the subworkflow 😦

FelixKrueger:
Thanks — and agreed it isn't the meta changes, this diff is comments and `meta.yml` only so it can't reach samtools output. But I don't think it's unstable either; it looks deterministic:

- docker, conda and singularity all produce the **same** md5 (`1dd820a3…`) for `test.umi_dedup.genome.sorted.bam.stats`, and all three disagree with the committed `5268f5f2…`.
- `tests/config/nf-test.config` pins `cpus = 4`, so the `--threads` value that `samtools stats` writes into its header is constant — which rules out the usual environment-dependent-header cause.
- Only the two `.stats` files move. The BAMs, flagstat and idxstats all match, so it's the stats file rather than the alignment data.

The snapshot was last regenerated in #12340 (samtools 1.24), so my guess is the version string in the stats header has shifted since, but I can't confirm that from the logs alone.

Happy to regenerate the snapshot on this branch if you want it unblocked, or to drop the `.stats` md5s for content assertions if that header is going to keep moving. Say which and I'll push it.

FelixKrueger:
Pushed a fix for it rather than leaving it as a diagnosis — the three `.stats` md5s were just out of date.

Evidence they're reproducible rather than flaky: docker, conda and singularity all produced the same values, and so did the run after you merged master, which is a separate run on a different commit. I lifted the md5s from CI's output instead of regenerating locally, so they're the runners' values and not my machine's.

The earlier commit is untouched and still comments and `meta.yml` only.

SPPearce (approved):
Maybe it just wasn't updated properly, but it could possibly be a date or something.
Whatever, we'll take a look if it comes up again, thanks. ;)

PR #12472, opened 2026-07-31
Deep mased imp modules

SkyLexS:
## PR checklist

Closes #XXX 

- [x] Made DeepMAsED opt-in by default, merged the redundant skip flags into one (skip_deepmased), and renamed deepmased_cpu_only to deepmased_use_gpu (also opt-in). Fixed the assembly filter to select by assembler instead of sr_platform, correctly excluding hybrid/long-read assemblies, and decoupled DeepMAsED from skip_binning like ALE. On the modules: added the process_gpu label to DEEPMASED_PREDICT and fixed DEEPMASED_FEATURES' output (plus stub) to accept gzipped files. Docs updated accordingly.
- [x] Add a resource `label`
  - For modules:
  - For subworkflows:


modules/nf-core/deepmased/features/main.nf
```diff
@@ -13,7 +13,7 @@ process DEEPMASED_FEATURES {
 
     output:
     tuple val(meta), path("*_feature_file_paths.tsv"), emit: feature_table
-    tuple val(meta), path("*_feats.tsv"), emit: feature_files
+    tuple val(meta), path("*_feats.tsv{,.gz}"), emit: feature_files
```
dialvarezs, line 16: Need to update the meta output block to match this change.

SkyLexS: it was an unexpected conflict and I resolved it poorly

dialvarezs:
Hi @SkyLexS,
I did a few additional tests and came to the conclusion that supporting GPU in this module would be basically impossible without forking the tool and updating it.

First, the Biocontainers container doesn't include CUDA (TensorFlow is compiled for CPU), so we can't use that one for GPU execution.

I tried to create a new container that includes CUDA-enabled TensorFlow, but I didn't succeed. The main problem is that DeepMAsED depends on Python 3.7 and TensorFlow 2.0, both of which are quite old. Both dependencies need to be matched with CUDA 10, so there's no support for GPUs from roughly the last 7 years. And we can't use a newer TensorFlow because DeepMAsED relies on old APIs. There have been no new developments on the tool in all these years (in fact, the developers created a new tool in the meantime: https://github.com/leylabmpi/resmico).

For these reasons, I suggest we just run it in CPU mode and don't add GPU options at all.

dialvarezs (approved):
LGTM!

PR #12473, opened 2026-07-31
Swap plink2 modules to topics

SPPearce:


modules/nf-core/plink2/pmerge/tests/main.nf.test
```diff
@@ -64,7 +64,7 @@ nextflow_process {
         }
         then {
             assertAll(
-                { assert process.success },
+                { assert process.success},
```
erikrikarddaniel: The lack of trailing white space before the curly bracket is not consistent with the rest of the file. (And maybe in other files as well. Caught my eye here.)

modules/nf-core/plink2/vcf2bgen/tests/main.nf.test
erikrikarddaniel, line 1: I suppose you have a reason for removing the single end param, but it looks a bit odd since it has nothing to do with the rest of the PR.

erikrikarddaniel (approved):
LGTM. Two tiny comments.

modules/nf-core/plink2/pmerge/tests/main.nf.test
SPPearce: Yes, I used find and replace, but this one wasn't changed as this moduel was already topics.

modules/nf-core/plink2/vcf2bgen/tests/main.nf.test
SPPearce: The single_end meta value isn't doing anything here, and is just a hangover from the old template.

PR #12474, opened 2026-07-31
Swap quantmsutils to topics

SPPearce:


subworkflows/nf-core/dia_proteomics_analysis/tests/main.nf.test.snap
erikrikarddaniel, line 1: I can't see the version of the tool here.

erikrikarddaniel (approved):
I'm missing the tool version in one snap diff -- see comment -- or am I blind? Approving in case the latter is true.

subworkflows/nf-core/dia_proteomics_analysis/tests/main.nf.test.snap
SPPearce: This is intended. Subworkflows no longer process the versions, because they are passed directly to the topic channel instead.

PR #12491, opened 2026-08-01
Bump trgt/genotype

bagnacan:
This PR bumps trgt/genotype from version `5.0.0` to version `5.1.0`.

Note that, differently from the new conda environment and [docker container build](https://wave.seqera.io/view/builds/bd-66e3b26326e566e7_1), the [singularity container build](https://wave.seqera.io/view/builds/bd-50ffb29e555ef626_1) must explicitly install gzip. This is done to preserve the `stub` block as it is.

In detail: when the singularity container build does not install gzip, the `--profile singularity` stub tests fail with `.command.sh: line 3: gzip: command not found`; this _could_ be avoided by changing [this line](https://github.com/nf-core/modules/blob/826a9b28d6da61097d60e1923bb8a2e51de3c5b4/modules/nf-core/trgt/genotype/main.nf#L43) of the `stub` block with `touch ${prefix}.vcf.gz`, however, the resulting file would be an empty file instead of an (empty) _gzip compressed_ file.

## PR checklist
  - For modules:


mashehu (dismissed):
Add gzip as a dependency in environment.yml please. Makes it easier to do these updates in the future.

SPPearce (approved):

PR #12494, opened 2026-08-02
updated module: xengsort/index

LeonHornich:
Updated existing `xengsort/index` module with the following changes:
- tool version: now latest version -> 2.2.1: this adds support for `--bisulfite` mode
- adjusted input and output handling to be more in-line with nf-core standards
- added thread parameters as according to tool [handling](https://gitlab.com/genomeinformatics/xengsort/-/blob/master/xengsort/xengsort/xengsort_index.py?ref_type=heads)
- extended testing cases: default, bisulfite, stub

## PR checklist

Closes #XXX 

- [x] Add a resource `label`
  - For modules:


modules/nf-core/xengsort/index/tests/main.nf.test
```diff
@@ -3,32 +3,114 @@ nextflow_process {
...
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert path(process.out.hash[0][1]).exists() },
+                { assert path(process.out.info[0][1]).exists() },
+                { assert snapshot(
+                    path(process.out.hash[0][1]).getFileName().toString(),
+                    path(process.out.info[0][1]).getFileName().toString(),
+                    process.out.findAll { key, val -> key.startsWith("versions") }
```
SPPearce: This could probably all be sanitizeOutput(process.out, unstableKeys: ["hash", "info"])

modules/nf-core/xengsort/index/main.nf
```diff
@@ -3,41 +3,57 @@ process XENGSORT_INDEX {
...
+    }
+
+    def read_threads = Math.ceil(subtables / 10) as int
+
+    def split_threads = 2 * read_threads
+
+    if ((subtables + read_threads + split_threads) >= cpus) {
+        read_threads = 1
+        split_threads = 2
+    }
```
SPPearce, line 42: Where is all this coming from?

modules/nf-core/xengsort/index/meta.yml
```diff
@@ -76,7 +87,8 @@ topics:
       - xengsort --version:
           type: eval
           description: The expression to obtain the version of the tool
+
 authors:
-  - "@diegomscoelho"
```
SPPearce, line 80: Still was the original author. You can add yourself though.

modules/nf-core/xengsort/index/main.nf
LeonHornich: I have had a similar discussion [here](https://github.com/nf-core/modules/pull/12386). The values and logic is taken from the tools source [code](https://gitlab.com/genomeinformatics/xengsort/-/blob/master/xengsort/xengsort/xengsort_index.py?ref_type=heads).

LeonHornich: @SPPearce any thoughts on this? Or should i omit this approach entirely?

SPPearce: If it is in the source code, does that mean it does it automatically?

SPPearce: I don't mind you doing this if that is suggested by the tool, just add a comment pointing to that effect

LeonHornich: > If it is in the source code, does that mean it does it automatically?

This is the tools default threading behavior if none of `--threads-split`, `--threads-read` or `--subtables` is specified when executing the tool. Meaning these could be overwritten by a `nextflow.config` file. 

Since the tool doesn't explicitly support a single `--threads ${task.cpus}` (or similar) logic, i figured this approach was still in line with nf-core modules guidelines and the support for multi-threading?

modules/nf-core/xengsort/index/tests/main.nf.test
```diff
@@ -3,32 +3,100 @@ nextflow_process {
...
+                input[2] = 10000
                 """
             }
         }
 
         then {
             assertAll(
                 { assert process.success },
-                { assert snapshot(sanitizeOutput(process.out)).match() }
+                { assert snapshot(process.out).match() }
```
SPPearce:
```suggestion
                { assert snapshot(sanitizeOutput(process.out)).match() }
```

SPPearce (commented):
Just the snapshot that is missing sanitizeOutput and then good.

modules/nf-core/xengsort/index/tests/main.nf.test
LeonHornich: thought this wasn't necessary for the stub tests, since it's just empty files? Is there a reason for this aside from universality?

SPPearce: without this, the snapshot has the named channels AND the numbered channels, both showing the same thing.

modules/nf-core/xengsort/index/tests/main.nf.test
```diff
@@ -3,32 +3,100 @@ nextflow_process {
...
+                input[2] = 10000
                 """
             }
         }
 
         then {
             assertAll(
                 { assert process.success },
-                { assert snapshot(sanitizeOutput(process.out)).match() }
+                { assert snapshot(sanitizeOutput(process.out, unstableKeys: ["hash", "info"])).match() }
```
SPPearce:
```suggestion
                { assert snapshot(sanitizeOutput(process.out)).match() }
```

modules/nf-core/xengsort/index/tests/main.nf.test
```diff
@@ -3,32 +3,100 @@ nextflow_process {
...
+                ]
+                input[2] = 2000000
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(sanitizeOutput(process.out)).match() }
```
SPPearce:
```suggestion
                { assert snapshot(sanitizeOutput(process.out, unstableKeys: ["hash", "info"])).match() }
```

modules/nf-core/xengsort/index/tests/main.nf.test
LeonHornich: that makes sense thank you :)

SPPearce (commented):
What happened to `mask`?

SPPearce (approved):
I think this is fine, provided the mask removal is expected.

LeonHornich:
> What happened to `mask`?

For the tool to run you have to specify either a k-mer length `-k` or a k-mer mask (see module version previous tests). These two are mutually exclusive and are therefore technically not always required by the tool. This is why i do not include them in the module as an input channel. They'll have to be specified in the config.

PR #12495, opened 2026-08-02
Add cross-caller merge tests to custom/orfmerge

FelixKrueger:
`custom/orfmerge` exists to collapse the same ORF called by different callers into one catalogue row, but its real-data test can't demonstrate that: Ribo-TISH and RiboCode share exactly one `transcript_id` and zero `(transcript_id, strand, start, end)` keys on the chr20 fixture, so no row ever carries both callers. This adds two synthetic tests, one per clustering strategy — transcript-anchored ORFs collapsing on an exact span, and non-transcript-anchored ORFs collapsing by reciprocal overlap. Both build their input inline with `collectFile`, so no test-data changes are needed and nothing about the module changes.

<details>
<summary>AI-assisted detail — keep or bin as you see fit</summary>

The existing assertion looks like a merge check but is satisfied by zero merging:

```groovy
{ assert rows.every { it[called_ribotish] == '1' || it[called_ribocode] == '1' } },
```

Every row coming from exactly one caller passes it. That is the actual state on this fixture — 20 rows, 15 from RiboCode and 5 from Ribo-TISH, none from both. Cross-*sample* merging is genuinely covered (`n_samples == 2` on five rows, since Ribo-TISH runs twice on the same input), so only the cross-caller half is missing.

Why the fixture can't cover it: the two callers overlap on one transcript, `ENST00000618693`, and even there they describe different features — Ribo-TISH reports a ~200 aa CDS spanning the whole 177 kb locus, RiboCode a 70 aa out-of-frame internal ORF inside part of it. Nothing should merge those, and nothing does.

The two added tests:

| test | strategy exercised | asserts |
| --- | --- | --- |
| transcript-anchored | collapse by `(transcript_id, strand, start, end)` | two callers at identical coordinates give one row with both `called_by_*` set; a second uORF on the same transcript at a different span stays a separate row |
| non-transcript-anchored | `cluster_by_reciprocal_overlap` | two callers 10 nt apart (0.97 reciprocal) give one row with both `called_by_*` set |

They use `uORF` and `novel_u`, and assert on `called_by_*` and `orf_class` only — no snapshot — so they stay valid across changes to the `orf_class` vocabulary and need no `.snap` entries.

Verified against this commit's parent by running nf-test locally: both new tests pass. I could not reproduce the container in my environment, so the two pre-existing tests' `snapshot(process.out).match()` assertions fail locally on a `versions.yml` Python difference — they fail identically with the pristine test file, so that is unrelated to this change, and CI is the authority.

Found while adding an `is_smorf` / positional-`orf_class` change to nf-core/riboseq, where this gap meant a merge regression would have gone undetected. That work is separate and not yet submitted; this stands on its own.

</details>


FelixKrueger:
The failures are a stale snapshot on `master`, not something this PR changed.

All six shards (docker, conda, singularity × 2) disagree with the committed snapshot on the same four md5s, and agree with each other. I reproduced those exact values locally against **pristine** upstream code with this PR's test additions reverted, so the mismatch predates the PR — it just wasn't visible, because nf-core only runs tests for changed modules and nothing had touched `custom/orfmerge` since the snapshot was recorded in #12167.

Re-recorded the four values in 3b78904. Only the catalogue and consensus outputs moved; `versions.yml`, the ORF-to-gene maps and the MultiQC sidecar are unchanged, and every behavioural assertion in both pre-existing tests was already passing (`1 of 14` and `1 of 6` failed, both the snapshot).

<details>
<summary>AI-assisted detail — evidence and what I ruled out</summary>

| output | committed | produced (CI ×6 and locally) |
| --- | --- | --- |
| `cohort.catalogue.bed12` | `59b9b0c4…` | `00ba4985…` |
| `cohort.catalogue.tsv` | `d6af31aa…` | `87f40563…` |
| `cohort.catalogue.consensus.bed12` (min-samples 2) | `b3da4afa…` | `b3c71662…` |
| `cohort.catalogue.consensus.tsv` (min-samples 2) | `70d77817…` | `43d98f76…` |

Ruled out as causes:

- **This PR.** Reverting the added tests and re-running pristine upstream code gives the same four values CI produced. The additions are separate `test` blocks with no snapshot assertions, so they cannot alter another test's output.
- **Test-data drift.** `sample1.ribotish.pred.txt` and `sample1.ribocode.txt` last changed 2026-05-21, before the snapshot was recorded on 2026-06-25; the chr20 GTF dates from 2024.
- **The `custom/orfnormalise` change in #12436**, whose output this module's `setup` consumes. That patch only touches `parse_price`, and this test's setup runs Ribo-TISH and RiboCode; its own snapshot diff was `+49/-0`, so no existing entry moved.
- **Environment.** `versions.yml` is identical in the diff, and the four values reproduce across three container engines plus a container-free local run.

I could not identify what originally invalidated it between 2026-06-25 and now, so I've re-recorded rather than claimed a root cause. If you would rather this landed as its own PR ahead of the test additions, happy to split it.

</details>

erikrikarddaniel (approved):
There's no CHANGELOG entry. Not sure if it's needed, but an "Improved tests" could be useful.

FelixKrueger:
Thanks for the review!

I went looking and couldn't find one to add — a repo-wide search for `filename:CHANGELOG.md` in nf-core/modules returns 0 hits, and the PR checklist in `.github/PULL_REQUEST_TEMPLATE.md` doesn't mention a changelog either (unlike the pipeline template, where it is a required item). So unless there's a convention I've missed, there's nothing here to update and the commit messages plus this description are the record.

Happy to add an entry if you can point me at the right file.

erikrikarddaniel:
Sorry, my mistake. I was thinking of pipeline PRs.

PR #12512, opened 2026-08-03
Update icountmini suite to v4.0.1

hanalysis:
Updating all the icountmini commands to v4.0.1 from v3.0.1

### DIFF
--- a/modules/nf-core/icountmini/metagene/tests/main.nf.test.snap
+++ b/modules/nf-core/icountmini/metagene/tests/main.nf.test.snap
@@ -14,16 +14,16 @@
                     [
                         "ICOUNTMINI_METAGENE",
                         "iCount-Mini",
-                        "3.0.1"
+                        "4.0.0"
                     ]
                 ]
             }
         ],
-        "timestamp": "2026-06-02T09:44:12.242625359",
         "meta": {
-            "nf-test": "0.9.4",
-            "nextflow": "26.03.0"
-        }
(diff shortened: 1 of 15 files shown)


modules/nf-core/icountmini/metagene/tests/main.nf.test.snap
```diff
@@ -14,16 +14,16 @@
                     [
                         "ICOUNTMINI_METAGENE",
                         "iCount-Mini",
-                        "3.0.1"
+                        "4.0.0"
```
erikrikarddaniel, line 17: The PR title says 4.0.1. In more places.

erikrikarddaniel (approved):
Looks good except that the tool seem to report 4.0.0 but the package is pinned at 4.0.1

modules/nf-core/icountmini/metagene/tests/main.nf.test.snap
hanalysis: This is my fault - I updated icountmini but forgot to change the version number in `__about__.py` on the initial release, I did then change it and re-release (https://github.com/ulelab/iCount-Mini/commit/d93829a7ccf695af25e6e2906f858af11fbed1b7) but it looks like the tag points to the commit prior to the `__about__.py` update (as I used the same tag name for both releases).

The only difference is the number in `__about__.py` - all changes for v4.0.1 were included in the initial commit that the tag points to.

PR #12516, opened 2026-08-03
Swap trust4 to topics

SPPearce:


modules/nf-core/trust4/tests/main.nf.test.snap
```diff
@@ -1,119 +1,339 @@
...
+                            "test_airr_align.tsv",
+                            "test_report.tsv"
+                        ]
+                    ]
+                ],
+                "versions_trust4": [
+                    [
+                        "TRUST4",
+                        "trust4",
+                        "TRUST4 using v1.1.5-r573"
```
LouisLeNezet: Could you update the version emission to keep only the version number ?

modules/nf-core/trust4/main.nf
```diff
@@ -26,7 +25,7 @@ process TRUST4 {
     tuple val(meta), path("*.out")                  , emit: out
     tuple val(meta), path("*.fq")                   , emit: fq
     tuple val(meta), path("**")                     , emit: outs
-    path "versions.yml"                             , emit: versions
+    tuple val("${task.process}"), val('trust4'), eval("run-trust4 2>&1 | grep -o 'v[0-9.]*-r[0-9]*' | sed 's/^/TRUST4 using /'"), emit: versions_trust4, topic: versions
```
LouisLeNezet: Something like the following maybe ?
```suggestion
    tuple val("${task.process}"), val('trust4'), eval("run-trust4 2>&1 | sed -n 's/.*TRUST4 using v/p'"), emit: versions_trust4, topic: versions
```

LouisLeNezet (changes requested):
A small issue with the output

LouisLeNezet (approved):

PR #12574, opened 2026-08-04
Add subworkflow for gridss somatic structural variant calling

### LINKED ISSUE #4498: new subworkflow: distributed computing GRIDSS subworkflow
### Is there an existing subworkflow for this?

- [X] I have searched for the existing subworkflow

- [X] If I'm planning to work on this subworkflow, I added myself to the `Assignees` to facilitate tracking who is working on the subworkflow

imsarath:
This PR adds `bam_tumor_normal_somatic_structural_variant_calling_gridss`, a subworkflow that calls somatic structural variants from a tumor/normal BAM pair. It chains the full GRIDSS workflow - optional targeted fragment extraction, per-sample preprocess, joint assembly and call, then `gridss_somatic_filter`. It emits the unfiltered joint VCF plus the all-somatic and high-confidence somatic call sets.

## PR checklist

Closes #4498 

- [x] Add a resource `label`
  - For modules:
  - For subworkflows:


subworkflows/nf-core/bam_somatic_sv_calling_gridss/tests/main.nf.test
```diff
@@ -0,0 +1,162 @@
+nextflow_workflow {
+
+    name "Test Subworkflow BAM_TUMOR_NORMAL_SOMATIC_STRUCTURAL_VARIANT_CALLING_GRIDSS"
```
SPPearce: Maybe we could shorten this? Seems a little excessive.

subworkflows/nf-core/bam_somatic_sv_calling_gridss/tests/main.nf.test
```diff
@@ -0,0 +1,162 @@
...
+            }
+        }
+    }
+
+    test("human - bam - tumor_normal - untargeted") {
+
+        when {
+            workflow {
+                """
+                input[0] = channel.of([
```
SPPearce, line 40: Shouldn't need the channel.of or channel.value at all.

subworkflows/nf-core/bam_somatic_sv_calling_gridss/main.nf
```diff
@@ -0,0 +1,67 @@
...
+
+
+workflow BAM_TUMOR_NORMAL_SOMATIC_STRUCTURAL_VARIANT_CALLING_GRIDSS {
+
+    take:
+    ch_input_bam            // channel: [ val(meta), [ tumor_bam, normal_bam ], [tumor_bai, normal_bai] ]
+    ch_fasta_fai_bwaindex   // channel: [ val(meta2), fasta, fai, bwa_index ]
+    ch_pondir               // channel: [ val(meta3), [ pondir ] ]
+    ch_gridss_config        // channel: [ val(meta4), [ gridss_config ] ]
+    val_target_bed          // string:  target BED for panel/exome data, null to run untargeted
```
SPPearce: This isn't a value?

subworkflows/nf-core/bam_somatic_sv_calling_gridss/tests/main.nf.test
imsarath: By shortening, you mean renaming the subworkflow. I just followed  `bam_tumor_normal_somatic_variant_calling_{gatk,strelka}` and added  `structural_`, which pushed it over. I did feel it was getting long as I was writing it. 

How about `bam_somatic_sv_calling_gridss`or `bam_tumor_normal_sv_calling_gridss`? Happy to use something else if you have any suggestions; just let me know, and I'll update it.

SPPearce (commented):
Can you format the main.nf with each module argument on its own line too please

subworkflows/nf-core/bam_somatic_sv_calling_gridss/tests/main.nf.test
imsarath: Modified the subworkflow name - `bam_somatic_sv_calling_gridss`

subworkflows/nf-core/bam_somatic_sv_calling_gridss/tests/main.nf.test
imsarath: I tried removing it completely, but `input[0]` and `input[1]` still require it.

subworkflows/nf-core/bam_somatic_sv_calling_gridss/main.nf
imsarath: Yes, it's a groovy path value for target_bed. I realized we could implement this in a cleaner way, so I've updated the code a bit. Let me know what you think of this approach!

subworkflows/nf-core/bam_somatic_sv_calling_gridss/tests/main.nf.test
SPPearce: hmm, dunno then, but fine.

subworkflows/nf-core/bam_somatic_sv_calling_gridss/tests/main.nf.test
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        then {
+            assertAll(
+                { assert workflow.success },
+                { assert snapshot(
+                    path(workflow.out.gridss_vcf[0][1]).vcf.summary,
+                    workflow.out.all_somatic_vcf.collect { meta, vcf -> [ meta, file(vcf).name ] },
+                    workflow.out.high_confidence_somatic_vcf.collect { meta, vcf -> [ meta, file(vcf).name ] }
+                ).match() }
```
SPPearce, line 74: Try this (untested):
```suggestion
                { assert snapshot(sanitizeOutput(workflow.out, variantsMD5Keys: ["gridss_vcf", "all_somatic_vcf", "high_confidence_somatic_vcf"]).match() }
```

SPPearce (commented):
A couple of comments.
You should be able to calculate `is_targeted` based on whether ch_target_bed exists?
Should the input tuple be flatter, rather than [tumour_bam, normal_bam] as a nested tuple.
Should the target bed be part of the input tuple? I could see scenarios where you are only interested in certain variants per sample.
What happens if you have more than one tumour sample per normal sample, can the subworkflow deal with that?

imsarath:
Sorry for the slow reply, and thanks for the detailed review!

1. `is_targeted` from `ch_target_bed` - I did try that first, but I couldn't get it to work. The problem is that `is_targeted` drives an if block that decides which processes get included in the DAG, and that decision happens at workflow-composition time - the channel contents aren't available yet, so an empty vs. non-empty ch_target_bed can't be inspected there. A boolean was the simplest thing that worked. Happy to switch if there's a better pattern.
2. Flatter input tuple -  Is there a specific advantage to the flat form? I used the nested one because GRIDSS_ASSEMBLE/CALL take the paired BAMs as a list anyway. Easy to change if flat is the convention.
3. Target BED in the input tuple - Agreed, makes sense. Note that `GRIDSS_EXTRACTOVERLAPPINGFRAGMENTS` currently takes the BED as a separate channel, so that module would need updating too.
4. Multiple tumours per normal - I hadn't considered it; the focus was one tumour/one normal, so it won't handle that today. We can add it, but it'd need a nested structure for the tumour BAMs/BAIs, which conflicts with point 2.

PR #11962, opened 2026-06-10
feat(fasta_gtf_bam_rpbp): add subworkflow

pinin4fjords:
Part of a series splitting #11695 (closed at reviewer request: one PR per module).

Adds the `fasta_gtf_bam_rpbp` subworkflow, which wires all 8 rpbp modules together: `rpbp/preparegenome` once per genome, then the 7-step per-sample chain for each input BAM. Designed for callers that have pre-aligned Ribo-seq BAMs plus a matching genome FASTA and GTF.

Depends on all 8 module PRs in this series being merged first.

### Series

| Order | Module / subworkflow | PR | Status |
|-------|----------------------|----|--------|
| genome prep | `rpbp/preparegenome` | #11954 | ✅ merged |
| 1 | `rpbp/extractmetageneprofiles` | #11955 | ✅ merged |
| 2 | `rpbp/estimatemetagenebayesfactors` | #11956 | ✅ merged |
| 3 | `rpbp/selectperiodicoffsets` | #11957 | ✅ merged |
| 4 | `rpbp/getperiodiclengthsoffsets` | #11958 | ✅ merged |
| 5 | `rpbp/extractorfprofiles` | #11959 | ✅ merged |
| 6 | `rpbp/estimateorfbayesfactors` | #11960 | ✅ merged |
| 7 | `rpbp/selectfinalpredictionset` | #11961 | ✅ merged |
| — | `fasta_gtf_bam_rpbp` subworkflow | #11962 | open |

PR #12599, opened 2026-08-05
feat(rpbp/preparegenome): emit Rp-Bp's per-ORF labels file

FelixKrueger:
`get_orfs` already writes `<name>.orfs-labels.annotated.tab.gz` into `transcript-index/`, holding the ORF type that `label-orfs` assigned to every candidate ORF. The module moved three files out of the reference layout and then removed the directory, so this one was discarded. This adds it to the move list and emits it. No existing output changes.

It matters because consumers currently have no access to Rp-Bp's positional call at all: `custom/orfnormalise` reads the predicted-orfs BED, which has no ORF-type column, so every Rp-Bp ORF is classified as a canonical CDS regardless of where it actually sits.

**One unrelated line came along, because CI needed it.** `environment.yml` pinned only `rpbp`, while `versions.yml` reports `platform.python_version()` — so conda resolved a different interpreter than the container was built with and produced a different `versions.yml`. A snapshot holds one md5, so the conda shards failed while docker and singularity passed, on `master` as much as here. Pinning `python=3.13.13` (the version the container carries — it reproduces the recorded md5 exactly) makes all three engines agree. The `custom/orf*` modules already pin python for this reason. **Heads up: the other seven `rpbp/*` modules pin only `rpbp` and will hit the same thing the next time they are touched** — happy to send a follow-up covering them if you'd like.

<details>
<summary>Detail (AI-assisted)</summary>

**Where the file comes from.** `get_orfs` builds the path via `filenames.get_labels(...)` and runs `label-orfs` with it as the sole entry in `out_files`, so it is written unconditionally on the annotated path this module uses (`is_annotated=True, is_de_novo=False`). It is a BED12+ file with a `#`-prefixed header and columns `orf_num`, `id`, `orf_type`, `transcripts` — the same shape as the predicted-orfs BED, and joinable to it on `id`.

**Labels are display names.** `label_orfs` copies `defaults.orf_type_name_map` and assigns its *values*, so the file carries `CDS`, `altCDS`, `intORF`, `uORF`, `uoORF`, `dORF`, `doORF`, `ncORF`, `Overlap`, `Suspect`. The `novel_*` variants need `--label-prefix novel_`, which only the de-novo path sets, so they cannot appear here. `Novel` itself needs `--nonoverlapping-label`, which this path also does not pass.

**Scope.** One entry in the move list, one `emit:`, one stub line, one `meta.yml` output block, plus the python pin. `shutil.move` still raises if the file is absent, which is the desired behaviour — a missing labels file means `label-orfs` failed. The rendered `path(...)` is one character shorter than `orfs_genomic_bed`'s, so the existing `, emit:` alignment is unchanged.

**Snapshot.** The three `orfs_labels` md5s are CI's own values. Worth noting the three pre-existing md5s came back unchanged, which is the real check here: `label-orfs` overwrites `orfs_genomic` in place during the same step, so adding a fourth file to the move list disturbing nothing else was not a given.

**Not affected.** `fasta_gtf_bam_rpbp` installs this module, but its `emit:` block is unchanged, so its snapshot should be stable. That snapshot dates from #11962, and the only constituent module touched since is `rpbp/getperiodiclengthsoffsets` in #12221, which removed a blank line from a template — so no drift is expected there either.

</details>

### DIFF
--- a/modules/nf-core/rpbp/preparegenome/environment.yml
+++ b/modules/nf-core/rpbp/preparegenome/environment.yml
@@ -3,3 +3,4 @@ channels:
   - bioconda
 dependencies:
   - bioconda::rpbp=4.0.1
+  - conda-forge::python=3.13.13
--- a/modules/nf-core/rpbp/preparegenome/main.nf
+++ b/modules/nf-core/rpbp/preparegenome/main.nf
@@ -14,6 +14,7 @@ process RPBP_PREPAREGENOME {
     tuple val(meta), path("${prefix}.annotated.bed.gz")             , emit: transcript_bed
     tuple val(meta), path("${prefix}.orfs-genomic.annotated.bed.gz"), emit: orfs_genomic_bed
     tuple val(meta), path("${prefix}.orfs-exons.annotated.bed.gz")  , emit: orfs_exons_bed
+    tuple val(meta), path("${prefix}.orfs-labels.annotated.tab.gz"), emit: orfs_labels
     path "versions.yml"                                             , emit: versions_rpbp, topic: versions
 
     when:
@@ -30,6 +31,7 @@ process RPBP_PREPAREGENOME {
     echo "" | gzip > ${prefix}.annotated.bed.gz
     echo "" | gzip > ${prefix}.orfs-genomic.annotated.bed.gz
     echo "" | gzip > ${prefix}.orfs-exons.annotated.bed.gz
+    echo "" | gzip > ${prefix}.orfs-labels.annotated.tab.gz
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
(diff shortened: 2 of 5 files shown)


pinin4fjords (approved):
[low] The detail section describes the labels file as BED12+ and says it has the same shape as the predicted-ORFs BED. Rp-Bp writes a four-column table containing `orf_num`, `id`, `orf_type`, and `transcripts`, without BED coordinates. Worth correcting that sentence so downstream consumers do not expect geometry in this file.

Otherwise, this looks good. The new output is consistently represented in the process, stub, metadata, and snapshots, and the current checks are green.

AI-assisted review (Codex, on behalf of @pinin4fjords).

Comment added by Codex

PR #12601, opened 2026-08-06
fix probe-set inclusion assertion

### LINKED ISSUE #12600: Cellranger Multi fails in scrnaseq when running CMO

### Description of the bug

When running CMO samples in `nf-core/scrnaseq` pipeline, the module fails with the following:

```
cat > cellranger_multi_config.csv <<-CONFIG_EOF
[gene-expression]
reference,$PWD/references/gex/GRCh38
probe-set,$PWD/references/gex/probeset/frna_probeset_subset.csv
chemistry,auto
create-bam,true
[libraries]
fastq_id,fastqs,lanes,feature_types
PBMC_10K_CMO,$PWD/fastq_all/gex,,Gene Expression
PBMC_10K_CMO,$PWD/fastq_all/cmo,,Multiplexing Capture
[samples]
$(cat references/cmo/barcodes/PBMC_10K_CMO_cmo.csv)
CONFIG_EOF

cellranger multi 

--id=PBMC_10K_CMO 

--csv=cellranger_multi_config.csv 

--localcores=4 

--localmem=10 \

Command exit status:
1

Command output:
Martian Runtime - v4.0.14
Serving UI at [http://10707875421d:36439?auth=7OJWhXpSxQvRatJNYv0SjocGgtUNWPRpaXYAV_c5v6s](http://10707875421d:36439/?auth=7OJWhXpSxQvRatJNYv0SjocGgtUNWPRpaXYAV_c5v6s)

Running preflight checks (please wait)...

[error] Pipestance failed. Error log at:
PBMC_10K_CMO/SC_MULTI_CS/MULTI_PREFLIGHT/fork0/chnk0-u00b97425c3/_errors

Log message:
These library types are not compatible with Flex: Multiplexing Capture
```

The problem is that, when using Multiplexing Capture, the probe-set reference should not be added in the config.

**The solution**

The required change in the module is very simple, changing this line from:

```
if (gex_frna_probeset) gex_section << "probe-set,\$PWD/${gex_frna_probeset.name}"
```

to

```
// still allow frna probe-set for flex, but avoid adding when CMO or OCM barcodes are present, since those are mutually exclusive with frna
if (gex_frna_probeset && !has_cmo && !has_ocm) gex_section << "probe-set,\$PWD/${gex_frna_probeset.name}"
```

```console

```

### Relevant files

_No response_

### System information

_No response_

fmalmeida:
Resolves #12600 

Updating the `frna` probes-set reference assertion.
It was seen when running the `nf-core/scrnaseq` pipeline test dataset, that the probe-set was still being included even though data was `cmo`.


mahesh-panchal (commented):
Can you add nf-tests for this change please?

fmalmeida:
Hi @mahesh-panchal ;
The change should be still checked by the already present `Flex` test ... where `frna` should still be using.

But also, added a new test for mimicking the pipeline scenario where we might have multiple references loaded for samples with different library types, and thus, avoid loading the reference for the mutually exclusive types.

Example added is CMO with probeset frna present ... but then, module should not load it in the execution.

modules/nf-core/cellranger/multi/tests/main.nf.test
```diff
@@ -731,4 +731,120 @@ nextflow_process {
...
+        then {
+            assert process.success
+            assert snapshot(
+                process.out.findAll { key, val -> key.startsWith("versions") },
+                process.out.outs[0][1].findAll { file(it).name == 'assignment_confidence_table.csv' },
+                process.out.outs[0][1].findAll { file(it).name == 'tag_calls_summary.csv' },
+                process.out.outs[0][1].findAll { file(it).name == 'cells_per_tag.json' },
+                process.out.outs[0][1].findAll { file(it).name.contains('metrics_summary.csv') },
+                process.out.outs[0][1].findAll { file(it).name.contains('sample_filtered_feature_bc_matrix.h5') }
+            ).match()
```
mahesh-panchal: Can you add asserts on the config please to ensure they contain certain key words?
```
{ assert !path(process.out.config[0][1]).text.contains('probe-set,') }
```
This would also help the other tests too.

fmalmeida: Hi @mahesh-panchal ,
Sure, I can do that.
Sorry for the late reply, I was on holidays.

Will start it tomorrow.

fmalmeida: Hi @mahesh-panchal ;
I have added a list of "key assertions" to each test.
Currently I added them outside of the snaps.

Should they be inside it?

mahesh-panchal: They are fine to keep as independent asserts outside snapshot().

One thing to fix though: all the asserts should be wrapped in assertAll() 

```
then {
    assertAll(
        { assert process.success },
        { assert snapshot(
            process.out.findAll { key, val -> key.startsWith("versions") },
            ...
        ).match() },
        { assert path(process.out.config[0][1]).text.contains('[gene-expression]') },
        { assert !path(process.out.config[0][1]).text.contains('probe-set,') },
        // ...
    )
}
```

fmalmeida: Hi @mahesh-panchal ;
I have migrated them to inside an `assertAll()` function.

mahesh-panchal (approved):

PR #12613, opened 2026-08-06
Map Ribo-TISH's colon-qualified ORF types on their location

FelixKrueger:
Ribo-TISH qualifies a positional ORF-type label after a colon rather than replacing it — `Novel:CDSFrameOverlap`, `3'UTR:CDSFrameOverlap`, `Internal:CDSFrameOverlap`, `Novel:Known`. `CLASS_TOKENS` holds only the bare locations and `classify()` matched the whole token, so all of those fell through unmatched; since #12498 made an unmatched token fatal, real Ribo-TISH output now aborts the process. This matches on the part before the first colon instead. `orf_class` is positional and `CDSFrameOverlap` describes a frame relationship rather than a position, so the qualifier does not belong in the class — and `orf_type_native` already carries the full token verbatim, so nothing is lost. No new vocabulary entries are needed.

Found by four nf-core/riboseq pipeline tests failing on real chr20 data, with 3–33 offending rows each.

<details>
<summary>Verification notes (AI-assisted)</summary>

**Why the existing tests didn't catch it.** The module fixtures use bare tokens; only genuine Ribo-TISH output produces the composite form, so #12498 was 63/63 green while this was already broken for real data. The added test uses the four tokens observed in production and **fails without this change with the same message the pipeline produced**:

```
orfnormalise: unmapped ORF type value(s) for caller 'ribotish':
  "3'UTR:CDSFrameOverlap" (1), 'Internal:CDSFrameOverlap' (1),
  'Novel:CDSFrameOverlap' (1), 'Novel:Known' (1)
```

**Why location-only, rather than enumerating the composites.** Taking the location keeps `orf_class` purely positional, which is the invariant #12498 established when it split length out into `is_smorf`. Enumerating `<location>:<qualifier>` pairs would need a decision per combination and would reintroduce the brittleness that exact-token matching removed. The four observed tokens resolve through the **existing** table with no additions:

| token | location | → `orf_class` |
|---|---|---|
| `Novel:CDSFrameOverlap` | `Novel` | `novel_u` |
| `Novel:Known` | `Novel` | `novel_u` |
| `3'UTR:CDSFrameOverlap` | `3'UTR` | `dORF` |
| `Internal:CDSFrameOverlap` | `Internal` | `intORF` |

**Why the split is safe for every caller.** Checked all 52 `CLASS_TOKENS` keys across the five callers: none contains a colon, so splitting cannot shorten a token that was meant to match whole. A comment on the table records that constraint for anyone adding keys later.

**A location that matches nothing still fails**, so the guard #12498 added keeps its value — only the qualifier is permitted to be unknown, which is what makes this robust to Ribo-TISH adding qualifiers.

`nf-test test modules/nf-core/custom/orfnormalise` — 11/11 pass locally (10 pre-existing plus the new one).

</details>

### DIFF
--- a/modules/nf-core/custom/orfnormalise/templates/orfnormalise.py
+++ b/modules/nf-core/custom/orfnormalise/templates/orfnormalise.py
@@ -414,8 +414,7 @@ def write_outputs(bed_path, tsv_path, bed_lines, tsv_rows, parser_columns, unmap
 # exactly (casefolded) rather than by substring. Substring matching mis-fired
 # on the overlap forms, because "uorf" is a substring of "overlap_uorf".
 #
-# Keys hold the positional part only. Ribo-TISH qualifies it after a colon
-# (`Novel:CDSFrameOverlap`); classify() splits there, so no key may contain one.
+# No key may contain a colon: classify() matches only the part before one.
 # ----------------------------------------------------------------------------
 
 CLASS_TOKENS = {
@@ -501,12 +500,8 @@ CLASS_TOKENS = {
 def classify(caller, orf_type):
     """Map a caller's native ORF-type token to the harmonised class.
 
-    Callers may qualify a positional label rather than replace it: Ribo-TISH
-    emits `Novel:CDSFrameOverlap`, `3'UTR:CDSFrameOverlap` and similar. Only the
-    part before the first colon describes where the ORF sits, and `orf_class`
-    is positional, so the qualifier is dropped here and preserved verbatim in
-    `orf_type_native`. No CLASS_TOKENS key contains a colon, so this cannot
-    shorten a token that was meant to match whole.
+    Only the part before the first colon is matched; the full token survives in
+    `orf_type_native`.
 
     Returns (orf_class, matched); `matched` is False when a non-empty token's
     location matched no entry, so unmapped labels can be counted rather than
--- a/modules/nf-core/custom/orfnormalise/templates/orfnormalise.py
+++ b/modules/nf-core/custom/orfnormalise/templates/orfnormalise.py
@@ -413,6 +413,9 @@ def write_outputs(bed_path, tsv_path, bed_lines, tsv_rows, parser_columns, unmap
 # Every caller's ORF-type vocabulary is a closed enum, so tokens are matched
 # exactly (casefolded) rather than by substring. Substring matching mis-fired
 # on the overlap forms, because "uorf" is a substring of "overlap_uorf".
+#
+# Keys hold the positional part only. Ribo-TISH qualifies it after a colon
+# (`Novel:CDSFrameOverlap`); classify() splits there, so no key may contain one.
 # ----------------------------------------------------------------------------
 
 CLASS_TOKENS = {
@@ -498,13 +501,21 @@ CLASS_TOKENS = {
 def classify(caller, orf_type):
     """Map a caller's native ORF-type token to the harmonised class.
 
-    Returns (orf_class, matched); `matched` is False when a non-empty token
-    matched no entry, so unmapped labels can be counted rather than silently
-    absorbed into `other`.
+    Callers may qualify a positional label rather than replace it: Ribo-TISH
+    emits `Novel:CDSFrameOverlap`, `3'UTR:CDSFrameOverlap` and similar. Only the
+    part before the first colon describes where the ORF sits, and `orf_class`
+    is positional, so the qualifier is dropped here and preserved verbatim in
+    `orf_type_native`. No CLASS_TOKENS key contains a colon, so this cannot
+    shorten a token that was meant to match whole.
+
+    Returns (orf_class, matched); `matched` is False when a non-empty token's
+    location matched no entry, so unmapped labels can be counted rather than
+    silently absorbed into `other`.
     """
     if not orf_type:
         return "other", True
-    cls = CLASS_TOKENS[caller].get(orf_type.strip().casefold())
+    location = orf_type.split(":", 1)[0]
+    cls = CLASS_TOKENS[caller].get(location.strip().casefold())
     if cls is None:
         return "other", False
     return cls, True
--- a/modules/nf-core/custom/orfnormalise/tests/main.nf.test
+++ b/modules/nf-core/custom/orfnormalise/tests/main.nf.test
@@ -431,4 +431,51 @@ nextflow_process {
         }
     }
 
+    test("ribotish composite ORF types map on the location, keeping the qualifier native") {
+
+        when {
+            process {
+                """
+                input[0] = channel
+                    .of(
+                        'Tid\tGid\tGenomePos\tTisType\tAALen\tFisherPvalue',
+                        't1\tg1\tchr1:1000-1100:+\tNovel:CDSFrameOverlap\t30\t0.001',
+                        "t1\tg1\tchr1:1100-1200:+\t3'UTR:CDSFrameOverlap\t30\t0.001",
+                        't1\tg1\tchr1:1200-1300:+\tInternal:CDSFrameOverlap\t30\t0.001',
+                        't1\tg1\tchr1:1300-1400:+\tNovel:Known\t30\t0.001'
+                    )
+                    .collectFile(name: 'composite.ribotish.pred.txt', newLine: true, sort: false)
+                    .map { orfs -> [[id: 'sample1'], orfs, 'ribotish'] }
+                input[1] = channel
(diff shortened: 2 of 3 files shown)


FelixKrueger:
CI is red here but none of it belongs to this PR: all four failures are `PIRATE`, and they split by container engine — `PIRATE` **passes on docker** in this same run and fails only on conda and singularity. This PR changes Python inside a module template, which is engine-agnostic, so a failure that appears on two engines and not the third cannot be coming from it. On singularity the log says `Failed to pull singularity image`; on conda `PIRATE` runs but writes its own `fail_test.txt` instead of its expected outputs. Meanwhile `CUSTOM_ORFNORMALISE` ran in 33 shards and passed in every one, including the new test `ribotish composite ORF types map on the location, keeping the qualifier native`, and `BWAMEM3_MEM` and `STRINGTIE_STRINGTIE` passed alongside `PIRATE` in the same failing shards, so the runners are healthy. One other thing worth flagging: the head is `4a6b135f` rather than `d4d18fce` because the original checks wedged during the 2026-08-06 Actions incident — that commit is empty and the tree is byte-identical. Happy to wait if you'd rather not merge past an unrelated red; nothing here needs a change on my side.

<details>
<summary>Evidence (AI-assisted)</summary>

**The four failing checks**, out of 72 check runs across 7 suites:

| Check | Cause |
|---|---|
| `x64 \| conda \| 5` | `PIRATE` snapshot mismatch |
| `x64 \| singularity \| 5` | `PIRATE` snapshot mismatch **+** `Failed to pull singularity image` |
| `x64 \| singularity \| 6` | same pair |
| `confirm-pass-nf-test` | downstream aggregate, fails because the above did |

**`x64 | docker | 5` and `x64 | docker | 6` both passed** — same shard numbers, so the same tests, on the third engine. That is the load-bearing observation.

**PIRATE's diff shows the tool deciding it failed**, not snapshot drift — expected outputs absent, failure marker present:

```
-    "PIRATE.gene_families.ordered.tsv",
-    "PIRATE.gene_families.tsv",
-    "PIRATE.pangenome_summary.txt",
-    "cluster_alleles.tab",
     "PIRATE.log",
+    "fail_test.txt",
     "genome2loci.tab",
```

**This module's tests.** `Test Process CUSTOM_ORFNORMALISE` appears in 33 shard jobs, every one with conclusion `success` and no `FAILED` lines. Confirmed by test name rather than by aggregate, since this repo shards at roughly one test per shard:

```
'ribotish composite ORF types map on the location, keeping the qualifier native'
'homo_sapiens [chr20] - ribotish'
'homo_sapiens [chr20] - ribotish + ribocode'
'homo_sapiens [chr20] - ribotish + ribocode - collapse disabled'
'homo_sapiens [chr20] - ribotish - score-field override'
```

**Why the head moved.** The five original checks on `d4d18fce` wedged during the Actions incident: `queued` with zero jobs created, refusing both recovery verbs with contradictory errors — `cancel` reported "already completed", `rerun` reported "already running" (403). With no API route to revive them, a new head was the only way to get CI. `4a6b135f` re-parents the same tree (`276fd8f25`), so the content under test is unchanged and the local verification still describes it. Those five will read `queued` on `d4d18fce` indefinitely and gate nothing.

**Local verification, unchanged from the PR description:** 11/11 including a negative control — reverting only `classify()` makes the new test fail with the exact production error message.

</details>

FelixKrueger:
Merging `master` in `a563217c` cleared the `PIRATE` failures without touching `orfnormalise` — now green over 56 checks (50 passed, 6 skipped, none failed). So the attribution above is settled empirically too, and this is ready for review.

modules/nf-core/custom/orfnormalise/tests/main.nf.test
```diff
@@ -431,4 +431,51 @@ nextflow_process {
...
     }
 
+    test("ribotish composite ORF types map on the location, keeping the qualifier native") {
+
+        when {
+            process {
+                """
+                input[0] = channel
+                    .of(
+                        'Tid\tGid\tGenomePos\tTisType\tAALen\tFisherPvalue',
```
pinin4fjords: Do you have a real input we could use as a proper fixture in `test-datasets`? The existing one (`genomics/homo_sapiens/riboseq_expression/orf_predictions/sample1.ribotish.pred.txt`) obviously isn't representative enough, so we should consider adding another one rather than embedding dummy data here.

modules/nf-core/custom/orfnormalise/templates/orfnormalise.py
```diff
@@ -498,13 +501,21 @@ def write_outputs(bed_path, tsv_path, bed_lines, tsv_rows, parser_columns, unmap
...
 
-    Returns (orf_class, matched); `matched` is False when a non-empty token
-    matched no entry, so unmapped labels can be counted rather than silently
-    absorbed into `other`.
+    Callers may qualify a positional label rather than replace it: Ribo-TISH
+    emits `Novel:CDSFrameOverlap`, `3'UTR:CDSFrameOverlap` and similar. Only the
+    part before the first colon describes where the ORF sits, and `orf_class`
+    is positional, so the qualifier is dropped here and preserved verbatim in
+    `orf_type_native`. No CLASS_TOKENS key contains a colon, so this cannot
+    shorten a token that was meant to match whole.
```
pinin4fjords: Can we reduce the AI historical narrative in the comments? I get this problem a lot, but comments should be stand-alone and represent the current state, not reference a previous state you're fixing.

FelixKrueger: Trimmed in `45d5122` — comment-only across the three files, and no executable change: `orfnormalise.py`'s token stream is identical apart from the docstring literal itself.

- `classify()` docstring: the paragraph describing what the previous code did is gone. It now states only that the part before the first colon is matched and that the full token survives in `orf_type_native`.
- `CLASS_TOKENS` header: the Ribo-TISH narrative is replaced by the single constraint a future editor of that table actually needs — `No key may contain a colon: classify() matches only the part before one.` That invariant is load-bearing (a colon-bearing key would be silently truncated by the split), so I kept it, but moved it onto the table it constrains rather than leaving it as a justification in the docstring.
- `meta.yml` and the new test assertion: same treatment.

One thing I left alone: there is a pre-existing instance of exactly this pattern a few lines above, from #12498 — `Substring matching mis-fired on the overlap forms, because "uorf" is a substring of "overlap_uorf"`. Happy to fold that in here if you'd like it gone now; I left it out so this PR stays scoped to comments it introduced itself.

modules/nf-core/custom/orfnormalise/tests/main.nf.test
FelixKrueger: Yes — real output with these tokens exists, so this needn't stay synthetic. The composite labels are what made a pipeline-level snapshot-recording run fail on genuine chr20 Ribo-TISH output, with this distribution:

```
Novel:CDSFrameOverlap     29
Novel:Known                4
3'UTR:CDSFrameOverlap      1
Internal:CDSFrameOverlap   1
```

That is also why the module tests missed the gap — the existing fixture only contains bare tokens, exactly as you say.

Two choices I would rather you made before I open the test-datasets PR, since both affect other modules:

1. **A second file, or make the existing one representative?** My instinct is to add a new file (`sample1.ribotish.composite.pred.txt` or similar) rather than modify `sample1.ribotish.pred.txt` in place. Editing a shared fixture is what briefly broke `custom/orfcollapse` when nf-core/test-datasets#2193 merged — there is no version gate, so every consumer picks the change up immediately and only notices when something next touches it. But if you would rather the existing fixture simply became representative, that is fine too and I will sequence the module PRs behind it.
2. **Region.** The existing `sample1.*` family is chr20, and real composite tokens are plentiful there, so the family can stay on one region unless you want otherwise.

Once you have picked, the order is test-datasets first, then this PR swaps the inline rows for the fixture. Happy to do both.

pinin4fjords: We try not to overwrite files in test-datasets, they should really be write-once (though it's patchily followed). Add a new one.

FelixKrueger: Added as a new file in nf-core/test-datasets#2208, generated by running the pipeline on its own chr20 data in extended-ORF mode. Worth flagging: real output has seven colon-qualified forms, not the four we knew about -- 5'UTR:Known, 5'UTR:CDSFrameOverlap and Truncated:Known also occur, and all seven already resolve through the existing table. I will point the test at the fixture and drop the inline rows once that merges.

pinin4fjords (approved):

PR #12617, opened 2026-08-07
Icountmini metagene png output

hanalysis:
Just adding .png as an optional output for icount/metagene, the command has always produced this (dependent on input), just updating the module to support it.


modules/nf-core/icountmini/metagene/tests/main.nf.test.snap
```diff
@@ -2,6 +2,9 @@
     "homo_sapiens - bed - gtf": {
         "content": [
             {
+                "png": [
```
nschan, line 5: It seems like the test does not produce this output, in which cases is the png actually produced?

nschan (approved):
I have 1 question, but generally seems good.

modules/nf-core/icountmini/metagene/tests/main.nf.test.snap
hanalysis: You need specific types of features (e.g. intron->exon boundary of a specific amount of nucleotides) and the graph builds a histogram of each feature - I think the test bed file is too small to produce this output.

PR #12641, opened 2026-08-09
refactor: parameterise optitype module and update to 1.5.0

SpikyClip:
The following PR updates the optitype module to the newest version, and adds better parameterisation. Needed to close this [PR](https://github.com/nf-core/hlatyping/pull/229).

- Refactors optitype module by adding named parameters
- Updates optitype to version `1.5.0`
- Adds `cplex` package to container for completeness so that all three optitype solvers are supported
- Enable multithreading for `cbc/cplex` solvers

## PR checklist

- [x] Add a resource `label`
  - For modules:


modules/nf-core/optitype/main.nf
```diff
@@ -4,42 +4,48 @@ process OPTITYPE {
...
     echo "[behavior]" >> config.ini
     echo "deletebam=true" >> config.ini
-    echo "unpaired_weight=0" >> config.ini
-    echo "use_discordant=false" >> config.ini
+    echo "unpaired_weight=${unpaired_weight}" >> config.ini
+    echo "use_discordant=${use_discordant}" >> config.ini
 
-    # Run the actual OptiType typing with args
-    OptiTypePipeline.py -i ${bam} -c config.ini $args --prefix $prefix --outdir $prefix
+    optitype run --${meta['seq_type']} --config config.ini ${args} --prefix "${prefix}" --outdir "${prefix}" --input ${bam_str}
```
jonasscheid, line 48: meta.seq_type should come via ext.args. If someone would use optitype in another pipeline than hlatyping and there is no meta.seq_type defined, this would throw an error. We wanna outsource those pipeline-specific definitions into ext.args such that each pipeline can customise it.

heuermh (approved):

modules/nf-core/optitype/main.nf
```diff
@@ -4,42 +4,48 @@ process OPTITYPE {
 
     conda "${moduleDir}/environment.yml"
     container "${ workflow.containerEngine in ['singularity', 'apptainer'] && !task.ext.singularity_pull_docker_container ?
-        'https://depot.galaxyproject.org/singularity/mulled-v2-f3579993d544920cc7c09e8ddf77569e95ea8950:d3ee2331cb4615915b1eda5d23f231da6b0741d7-0' :
-        'quay.io/biocontainers/mulled-v2-f3579993d544920cc7c09e8ddf77569e95ea8950:d3ee2331cb4615915b1eda5d23f231da6b0741d7-0' }"
+        'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/9f/9ff14340c6c903c1c9d926fb7ffa60460986852515c76673cc54f6c958ff35bd/data' :
```
jonasscheid, line 7: Does this container have optitype cplex and coincbc?

modules/nf-core/optitype/main.nf
```diff
@@ -4,42 +4,48 @@ process OPTITYPE {
...
-    def args2     = task.ext.args2  ?: ''
-    prefix        = task.ext.prefix ?: "${meta.id}"
+    def args            = task.ext.args ?: ''
+    def solver          = task.ext.args2?.getAt("solver") ? "${task.ext.args2["solver"]}" : 'glpk'
+    // GLPK is single threaded only
+    def solver_threads  = "${solver}" == 'glpk' ? 1 : "${task.cpus}"
+    def unpaired_weight = task.ext.args2?.getAt("unpaired_weight") ? "${task.ext.args2["unpaired_weight"]}" : 0
+    def use_discordant  = task.ext.args2?.getAt("use_discordant") ? "${task.ext.args2["use_discordant"]}" : 'false'
+    // Each bam has to be prefixed with `--input`
+    def bam_str         = bam.join(" --input ")
```
jonasscheid, line 29: This line is not needed

modules/nf-core/optitype/main.nf
```diff
@@ -4,42 +4,48 @@ process OPTITYPE {
...
 
     when:
     task.ext.when == null || task.ext.when
 
     script:
-    def args      = task.ext.args   ?: ''
-    def args2     = task.ext.args2  ?: ''
-    prefix        = task.ext.prefix ?: "${meta.id}"
+    def args            = task.ext.args ?: ''
+    def solver          = task.ext.args2?.getAt("solver") ? "${task.ext.args2["solver"]}" : 'glpk'
```
jonasscheid, line 23: I don't know why args2 was used initially here, but it is not needed afaik. One can put the solver params in args. ArgsN should only be used for positional params that cannot be part of ext.args(1).

modules/nf-core/optitype/main.nf
```diff
@@ -4,42 +4,48 @@ process OPTITYPE {
...
+    def solver_threads  = "${solver}" == 'glpk' ? 1 : "${task.cpus}"
+    def unpaired_weight = task.ext.args2?.getAt("unpaired_weight") ? "${task.ext.args2["unpaired_weight"]}" : 0
+    def use_discordant  = task.ext.args2?.getAt("use_discordant") ? "${task.ext.args2["use_discordant"]}" : 'false'
+    // Each bam has to be prefixed with `--input`
+    def bam_str         = bam.join(" --input ")
+    prefix              = task.ext.prefix ?: "${meta.id}"
 
     """
-    # Create a config for OptiType on a per sample basis with task.ext.args2
+    touch config.txt
```
jonasscheid, line 33: Dead code?

modules/nf-core/optitype/main.nf
```diff
@@ -4,42 +4,48 @@ process OPTITYPE {
...
 
     script:
-    def args      = task.ext.args   ?: ''
-    def args2     = task.ext.args2  ?: ''
-    prefix        = task.ext.prefix ?: "${meta.id}"
+    def args            = task.ext.args ?: ''
+    def solver          = task.ext.args2?.getAt("solver") ? "${task.ext.args2["solver"]}" : 'glpk'
+    // GLPK is single threaded only
+    def solver_threads  = "${solver}" == 'glpk' ? 1 : "${task.cpus}"
+    def unpaired_weight = task.ext.args2?.getAt("unpaired_weight") ? "${task.ext.args2["unpaired_weight"]}" : 0
```
jonasscheid, line 26: Does this even work since ext.args is normally a string not a map? Maybe a simple 'is unpaired_weight in ext.args?' Does the job here

jonasscheid: Same for the solver logic

modules/nf-core/optitype/tests/nextflow.config
```diff
@@ -1,6 +1,12 @@
 process {
     withName: OPTITYPE {
-        ext.args = '--dna -e 1 -b 0.009'
-        ext.args2 = 'solver=glpk'
+        ext.args = { '--enumerate 1 --beta 0.009' }
```
jonasscheid, line 3: Change back to str please and drop args2

jonasscheid (commented):
Could you fix those changes in a new PR? Sorry for delay

modules/nf-core/optitype/main.nf
SpikyClip: So I was on the fence about what is the appropriate strategy here, I took a look at the [module guidelines](https://nf-co.re/docs/specifications/components/modules/input-output-options#required-val-channel-inputs), specifically:

> Input channel val declarations SHOULD be defined for all mandatory non-file inputs that are essential for the functioning of the tool (e.g., parameters, flags etc).

>- Mandatory non-file inputs are options that the tool MUST have to be able to be run.
>- These non-file inputs are typically booleans or strings, and must be documented as such in the corresponding entry in the meta.yaml.
>- Options, flags, parameters that are not required by the tool to function should NOT be included - rather these can be passed via ext.args.

I figured the rna/dna args were mandatory so they ought to go through the meta channel. There was a lot of arguments for and against this position, and I'm happy to switch back to having mandatory args in `ext.args` if you like. Are they any other pipelines that use the optitype module?

modules/nf-core/optitype/main.nf
SpikyClip: Yes, I used [Seqera containers](https://seqera.io/containers/) to generate the containers from the `environment.yml` file, and added tests using the two different solvers.

modules/nf-core/optitype/main.nf
SpikyClip: Line 26 and 27 are needed to set the defaults, otherwise you might end up with:

```bash
    echo "unpaired_weight=" >> config.ini
    echo "use_discordant=" >> config.ini
```

Which might cause issues.

Line 29 is necessary to meet the [requirements](https://github.com/FRED-2/Optitype#command-line) of OptiType 1.5.0, where `-i` is specified for every input.

Without line 29 joining the bams with ` --input `, if there is more than 1 bam in the bam stream, as in the case of [test.config](https://github.com/nf-core/hlatyping/blob/master/conf/test.config) in `nf-core/hlatyping` with paired bams, you get the following error (see attached for full log):

```
 Error: Got unexpected extra argument (SAMPLE_PAIRED_END_BAM_2.mapped.bam)
```

There isn't a test for this in the module, since the module only uses a [single bam](https://github.com/nf-core/modules/blob/fcad955c9c9529407951b09583707290c9220c94/modules/nf-core/optitype/tests/main.nf.test#L17-L18) in its tests.

Anyway, I'm not entirely sure why yara produces two bams or if optitype is even meant to process `BAM_2` (was it silently only processing the first bam before the `1.5.0` update?). If `BAM_2` should be ignored then it should be removed before generating the call.

[err.nextflow.log](https://github.com/user-attachments/files/31211434/err.nextflow.log)

modules/nf-core/optitype/main.nf
SpikyClip: I thought it would be neater to keep the `config.ini` args separate from the true command line args. I can shift solver over if you want me to since it now can be supplied via cli? But the other args are still CLI only.

modules/nf-core/optitype/main.nf
SpikyClip: Yes, not strictly necessary but thought it would be a bit more verbose. I'll remove it.

modules/nf-core/optitype/main.nf
SpikyClip: The main idea of the PR was to shift to using a closure rather than a plain string for `ext.args` as it is done for other modules such as [controlfreec](https://github.com/nf-core/modules/blob/master/modules/nf-core/controlfreec/freec/tests/nextflow.config), as it can be brittle supply more complex sets of args as a single string and attempting to parse it downstream, rather than just supplying it in a structured form to begin with.

Suppose, for example, you wanted to an `enumerate` of `2`, `unpaired_weight` of `0.2`, a `solver` of `cbc`, a `beta` of `0.01`, turning on the `use_discordant` flag. Using my approach, the module config might look like:

```
process {
    withName: OPTITYPE {
        ext.args = { '--enumerate 2 --beta 0.01' }
        ext.args2 = {
            [
                'solver': 'cbc',
                'unpaired_weight': 0.2,
                'use_discordant': false
            ]
        }
    }
}
```

Which works (tests won't pass if it didn't). How would these args be supplied in the current form, where args are a single string in ext.args?

modules/nf-core/optitype/tests/nextflow.config
SpikyClip: I kept `ext.args2` in the test configs as a reference to how it should be structured when used in a pipeline.

modules/nf-core/optitype/main.nf
heuermh: My interpretation of this doc is that mandatory non-file inputs should be `input`s, i.e.

```nextflow
    input:
    tuple val(meta), path(bam), path(bai)
    val seq_type
```

An example of this is `index_format` in `samtools sort`
https://github.com/nf-core/modules/blob/master/modules/nf-core/samtools/sort/main.nf#L13

SpikyClip: That's how I would approach it if the `seq_type` was a fixed flag for the whole pipeline run, i.e. it only processes rna or dna samples. As it is currently implemented, it can take a mix of dna and rna samples, so seq_type appropriately rides with the meta val I think.

With regard to whether `seq_type` should go into the `ext.args`, I think it should not as it can be a breaking change with an opaque error, which is what I encountered when I switched to using the `dev` branch to test as I had a pre-existing `ext.args` (That being said, this PR is a breaking change regardless since I've convert `ext.arg2` into a mapping in order to expose the other args, and renamed the parameters for optitype to be in line with the other tools).

The downside is flexibility of course, if you had a pipeline that only processed dna or rna samples its easier to force the flag in the configs than inject it into `meta.seq_type`. But in that case you wouldn't want users to be able to mess with that either by overriding the configs.

In any case, I'm not fussed if it gets switched again, there will be other breaking changes that will push the pipeline to a major revision.

jonasscheid: You are right, lets keep `val seq_type` !

PR #12654, opened 2026-08-11
Add custom/collectstats module

erikrikarddaniel:
## PR checklist

Related to nf-core/magmap#237 (not "Closes", it's phase 1 of a larger consolidation effort)

- [x] Add a resource `label`
  - For modules:
    - [x] `nf-core modules test custom/collectstats --profile docker`
    - [ ] `nf-core modules test custom/collectstats --profile singularity`
    - [ ] `nf-core modules test custom/collectstats --profile conda`

## Description

New shared component to join per-sample read-processing statistics (trimming, decontamination, alignment, feature counting, and optionally taxonomy/function summaries) from several upstream tools into a single overall-stats table.

This is phase 1 of the consolidation plan discussed in nf-core/magmap#237: `nf-core/magmap` and `nf-core/metatdenovo` each independently maintain a local `COLLECT_STATS` module that turned out to be the same script, evolved separately from a common ancestor. Ported from magmap's version, the more hardened of the two (avoids shelling out via `pipe(sprintf(...))` with sample-derived filenames in favour of `Sys.glob()`+`readLines()`; named-column `read_tsv` types instead of positional). Once this merges, both pipelines will adopt it in follow-up PRs to their own repos, removing their local duplicates.

It's a hand-rolled R aggregation script rather than a tool wrapper -- following the precedent already set by other `custom/*` modules (`custom/rsemmergecounts`, `custom/matrixfilter`, etc.).

Test data (small two-sample fixtures: Trim Galore reports, BBDuk logs, samtools idxstats, featureCounts-derived tables, an optional merge table) is in a separate PR to `nf-core/test-datasets`' `modules` branch: nf-core/test-datasets#2210. **This PR depends on that one merging first** -- CI here will fail to resolve the fixture paths until it does. Verified locally against both PRs' branches directly (all 3 nf-test cases pass with real Docker execution, including the optional-inputs-absent path), and separately verified the R aggregation logic by hand against the fixtures before either PR was opened.

### DIFF
--- a/modules/nf-core/custom/collectstats/main.nf
+++ b/modules/nf-core/custom/collectstats/main.nf
@@ -16,7 +16,7 @@ process CUSTOM_COLLECTSTATS {
 :         'community.wave.seqera.io/library/custom_collectstats:c1f477e6251c36bb' }"
 
     input:
-    tuple val(meta), val(samples), path(trimlogs), path(bblogs), path(idxstats), path(fcs), path(mergetab)
+    tuple val(meta), val(samples_meta), path(trimlogs), path(bblogs), path(idxstats), path(fcs), path(mergetab)
 
     output:
     tuple val(meta), path("${outfile}"), emit: overall_stats
@@ -31,10 +31,10 @@ process CUSTOM_COLLECTSTATS {
 
     def read_trimlogs = ""
     if ( trimlogs ) {
-        def se_trimlogs = samples
+        def se_trimlogs = samples_meta
             .findAll { s -> s.single_end }
             .collect { s -> "${rq(s.id)}, ${rq(s.id + '.*_trimming_report.txt')}, 1," }
-        def pe_trimlogs = samples
+        def pe_trimlogs = samples_meta
             .findAll { s -> ! s.single_end }
             .collect { s -> "${rq(s.id)}, ${rq(s.id + '_1.*_trimming_report.txt')}, 2," }
 
@@ -93,7 +93,7 @@ process CUSTOM_COLLECTSTATS {
     library(tidyr)
     library(stringr)
 
-    start    <- tibble(sample = c(${samples.collect { s -> rq(s.id) }.join(', ')}))
+    start    <- tibble(sample = c(${samples_meta.collect { s -> rq(s.id) }.join(', ')}))
 
     ${read_trimlogs}
 
--- a/modules/nf-core/custom/collectstats/main.nf
+++ b/modules/nf-core/custom/collectstats/main.nf
@@ -1,0 +1,188 @@
+// Safely quote a Groovy value as a single-quoted R string literal. Used everywhere
+// below that a sample ID or other user-controlled value gets embedded into generated
+// R source -- without this, a value containing a quote could break the generated R
+// syntax, or worse.
+def rq(v) {
+    return "'" + v.toString().replace('\\', '\\\\').replace("'", "\\'") + "'"
+}
+
+process CUSTOM_COLLECTSTATS {
+    tag "$meta.id"
+    label 'process_medium'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine in ['singularity', 'apptainer'] && !task.ext.singularity_pull_docker_container
+?         'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/9d/9d1a8065dcebe35c513db5eb4a5237795aa4bae4756086f3c4319a820a8a647c/data'
+:         'community.wave.seqera.io/library/custom_collectstats:c1f477e6251c36bb' }"
+
+    input:
+    tuple val(meta), val(samples), path(trimlogs), path(bblogs), path(idxstats), path(fcs), path(mergetab)
+
+    output:
+    tuple val(meta), path("${outfile}"), emit: overall_stats
+    path "versions.yml"                , emit: versions, topic: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    outfile    = "${prefix}.overall_stats.tsv.gz"
+
+    def read_trimlogs = ""
+    if ( trimlogs ) {
+        def se_trimlogs = samples
+            .findAll { s -> s.single_end }
+            .collect { s -> "${rq(s.id)}, ${rq(s.id + '.*_trimming_report.txt')}, 1," }
+        def pe_trimlogs = samples
+            .findAll { s -> ! s.single_end }
+            .collect { s -> "${rq(s.id)}, ${rq(s.id + '_1.*_trimming_report.txt')}, 2," }
+
+        read_trimlogs = """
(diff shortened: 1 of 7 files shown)


modules/nf-core/custom/collectstats/main.nf
```diff
@@ -0,0 +1,188 @@
...
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        R: 4.1.0
+        dplyr: 1.0.7
+        readr: 2.0.0
+        purrr: 0.3.4
+        tidyr: 1.1.3
+        stringr: 1.4.0
+    END_VERSIONS
+    """
```
sofiademmou, line 187: Should this be hardcoded ? Should it not change like in the non stub test ?

modules/nf-core/custom/collectstats/meta.yml
```diff
@@ -0,0 +1,123 @@
+name: "custom_collectstats"
+description: |
+  Collects per-sample read-processing statistics (trimming, decontamination, alignment,
+  feature counting, and optionally taxonomy/function summaries) from a set of
+  heterogeneous log/table files into one overall-stats table per run.
```
sofiademmou:
```suggestion
  Collects per-sample read-processing statistics (trimming, decontamination, alignment,
  feature counting, and optionally taxonomy/function summaries) from a set of
  heterogeneous log/table files into a single overall-stats table per run.
```

modules/nf-core/custom/collectstats/meta.yml
```diff
@@ -0,0 +1,123 @@
...
+        several upstream tools into a single wide table.
+      tool_dev_url: "https://github.com/nf-core/modules/blob/master/modules/nf-core/custom/collectstats/main.nf"
+      identifier: ""
+input:
+  - - meta:
+        type: map
+        description: |
+          Groovy Map containing run-level information
+          e.g. `[ id:'run' ]`
+    - samples:
```
sofiademmou: What do you think about calling it `samples_meta` instead of `samples` to make it even clearer ? :)

```suggestion
    - samples_meta:
```

modules/nf-core/custom/collectstats/tests/main.nf.test
```diff
@@ -0,0 +1,170 @@
...
+                { assert snapshot(
+                    process.out.overall_stats,
+                    path(process.out.overall_stats.get(0).get(1)).linesGzip
+                ).match() }
+            )
+        }
+
+    }
+
+    test("two samples - optional inputs absent") {
```
sofiademmou, line 68: Could you add a stub test for this one as well ?

modules/nf-core/custom/collectstats/tests/main.nf.test
```diff
@@ -0,0 +1,170 @@
...
+                    file(params.modules_testdata_base_path + 'generic/tsv/mergetab.tsv', checkIfExists: true)
+                ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
```
sofiademmou: Could you try this if it works ? :)
```suggestion
                { assert snapshot(sanitizeOutput(process.out)).match() }
```

modules/nf-core/custom/collectstats/tests/main.nf.test
```diff
@@ -0,0 +1,170 @@
...
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(
+                    process.out.overall_stats,
+                    path(process.out.overall_stats.get(0).get(1)).linesGzip
+                ).match() }
```
sofiademmou: Could you add the version to the snapshot as well ?
```suggestion
                    path(process.out.overall_stats.get(0).get(1)).linesGzip
                )
                process.out.findAll { key, val -> key.startsWith("versions")}
                .match() }
```

sofiademmou (changes requested):
Nice ! ⭐ 

Just a few comments/questions. I mainly looked at the nf-core module aspect of this since i am not really an expert in R. You might want to ask someone more experienced in R and custom modules to be sure !

modules/nf-core/custom/collectstats/main.nf
erikrikarddaniel: ~Yes it should, good point.~ My mistake, since it's in the stub this should be fine.

modules/nf-core/custom/collectstats/tests/main.nf.test
erikrikarddaniel: Good point to check, but I don't think this would add real coverage: the stub script (`cat /dev/null | gzip -c > ${outfile}`) never reads or branches on any input content -- unlike the real script (which does branch on whether `trimlogs`/`mergetab` are present), the stub would produce byte-identical output regardless of which inputs are given. So a second stub test here would just duplicate the existing one. Happy to add it if you'd still like it for symmetry, just flagging it wouldn't exercise anything new. (Checked this with some help from Claude Code.)

modules/nf-core/custom/collectstats/tests/main.nf.test
erikrikarddaniel: Confirmed, you understood it exactly right: yes, we do generate versions dynamically via packageVersion() etc., and yes, that's precisely why we've kept them out of the snapshot for these two tests -- if the underlying container ever gets rebuilt with a newer dplyr/readr/purrr, the snapshot would fail on a version bump that has nothing to do with any actual behavior change here. We hit exactly this problem before in our own pipeline's test suite, which is why we settled on this pattern there too. The stub test is different and can safely include versions in its snapshot, since those are hardcoded literals in the stub script, not queried at runtime -- they never drift. So I'd rather leave these two as-is. (Replied with some help from Claude Code.)

modules/nf-core/custom/collectstats/meta.yml
erikrikarddaniel: Done.

modules/nf-core/custom/collectstats/tests/main.nf.test
sofiademmou: Right ! It doesn't depend on the input so it should be fine like this.

modules/nf-core/custom/collectstats/tests/main.nf.test
sofiademmou: Do you mean that the snapshot would fail if we update the container to a new version for a tool ? I think that should probably be reflected in the snapshot and we should update them then.

I also found this in the [guidelines](https://nf-co.re/docs/specifications/components/modules/testing?utm_source=chatgpt.com)
> Tests for modules MUST, at a minimum, run on the GitHub repository CI with a stub test that replicates the generation of (empty) output files and a versions file.

So I think the version should be included right ?

erikrikarddaniel: You're right. Added.

sofiademmou (approved):
Looks good to me ! 👍

PR #12655, opened 2026-08-11
Bump cnvkit to 0.9.14

### LINKED ISSUE #12653: update module: cnvkit

paddyharker1005:
As mentioned in #12653 , the CNVkit modules were running a mixture of version 0.9.12 and 0.9.13. This bumps CNVkit to the latest version (0.9.14) in all modules so that they are consistent.

For `cnvkit/batch`, one CRAM-based test was also using the wrong reference FASTA for a chromosome-21 CRAM input. The test in main.nf.test was corrected to use the matching chr21 reference and FAI.

## PR checklist

Closes #12653  

  - For modules:


paddyharker1005:
@SPPearce having some issues with the test data in one of the cnvkit/batch tests `'cnvkit batch tumouronly mode - cram'` so haven't update the snaps for this yet

SPPearce:
Looks like the .call.cns and .cns files aren't stable at the moment.
What is the issue with the test data?

paddyharker1005:
> Looks like the .call.cns and .cns files aren't stable at the moment. What is the issue with the test data?

Seemed to be an issue with the reference file for one of the tests but this might need checking. Tests are passing now

paddyharker1005:
> > Looks like the .call.cns and .cns files aren't stable at the moment. What is the issue with the test data?
> 
> Seemed to be an issue with the reference file for one of the tests but this might need checking. Tests are passing now

were passing*

modules/nf-core/cnvkit/batch/tests/main.nf.test
SPPearce:
```suggestion
            assert snapshot(sanitizeOutput(process.out, unstableKeys: ["cns"])).match()
```

SPPearce (commented):
Something like this will not snapshot the md5sum for the cns file

modules/nf-core/cnvkit/batch/tests/main.nf.test
paddyharker1005: Ok thanks - looks like it might be an issue with CNVkit/batch using parallelisation which causes non-deterministic floating-point calculations. Will go with this for now

SPPearce:
you do need to update the snapshots too

paddyharker1005:
> you do need to update the snapshots too

Yep doing that now but also need to apply this change to all the tests with cns output

modules/nf-core/cnvkit/batch/tests/main.nf.test.snap
```diff
@@ -228,6 +228,17 @@
                 "genome.antitarget.bed"
             ],
             {
+                "bed": [
```
paddyharker1005, line 225: @SPPearce any understanding of why the bed files are still included here again even though I'm using `ignoreKeys: ["bed"]`?

paddyharker1005: (it's currently trying to snapshot the md5sum from an empty bed file which is what I'm trying to avoid)

SPPearce: Don't know why that doesn't work, but you are probably better to put it into `unstableKeys` anyway, then it'll check it exists. The ignoreKeys is more for files which don't even have a stable name (such as dated log files)

paddyharker1005: I think the issue here is that the bed file content isn't necessarily unstable but occasionally it's empty which causes the linting tests to fail. If it's easier just to put it in `unstableKeys` and ignore the content then I'll go with that

paddyharker1005: Annoyingly everything was fine until I tried to bump cnvkit for `cnvkit/batch`... 🤦

.vscode/settings.json
SPPearce, line 1: Shouldn't be editing this I guess

SPPearce (approved):

.vscode/settings.json
```diff
@@ -4,5 +4,6 @@
         "./modules/meta-schema.json": ["modules/nf-core/**/meta.yml"],
         "./modules/environment-schema.json": ["modules/nf-core/**/environment.yml"],
         "./subworkflows/yaml-schema.json": ["subworkflows/nf-core/**/meta.yml"]
-    }
+    },
+    "nextflow.telemetry.enabled": false
```
SPPearce:
```suggestion
    }
```

PR #12662, opened 2026-08-12
Remove fixed UMI-tools seeds

pinin4fjords:
## Summary

- stop injecting `--random-seed=100` in `umitools/dedup` and `umitools/group`
- stop setting `PYTHONHASHSEED=0` in production commands
- keep module and dependent subworkflow tests deterministic through test configuration

## Rationale

UMI-tools chooses a representative read randomly when duplicate candidates tie. Hard-coding a production seed makes that arbitrary choice repeat without making it robust, and prevents fresh runs from exposing stochastic variability. Users can still request reproducibility by setting `--random-seed` through `task.ext.args`.

## Tests

- x86_64 Docker: all 13 affected UMI-tools module and subworkflow nf-tests passed
- targeted pre-commit hooks and `ruff format --check` passed
- `nf-core modules lint umitools/dedup`: 61 passed, 3 existing metadata/container warnings
- `nf-core modules lint umitools/group`: 57 passed, 3 existing metadata/container warnings
- `nf-core subworkflows lint bam_dedup_stats_samtools_umitools`: 32 passed
- `nf-core subworkflows lint bam_dedup_umi`: 33 passed


SPPearce (changes requested):
No, I disagree entirely, I think we should be hardcoding random seeds if not specified.

pinin4fjords:
> No, I disagree entirely, I think we should be hardcoding random seeds if not specified.

This confuses me. If a tool doesn't set a random seed by default, why should we? The variant you force by setting the seed isn't any more valid than a variant you'd get by setting another one.

modules/nf-core/umicollapse/main.nf
```diff
@@ -36,16 +36,14 @@ process UMICOLLAPSE {
...
-    # Getting the umicollapse jar file like this because `umicollapse` is a Python wrapper script generated
-    # by conda that allows to set the heap size (Xmx), but not the stack size (Xss).
-    # `which` allows us to get the directory that contains `umicollapse`, independent of whether we
-    # are in a container or conda environment.
-    # WARN: Please update this string when bumping container versions.
-    UMICOLLAPSE_JAR=\$(dirname \$(which umicollapse))/../share/umicollapse-1.1.0-0/umicollapse.jar
+    # The generated launcher allows configuring heap size, but not stack size.
+    UMICOLLAPSE_JAR=\$(find "\$(dirname "\$(command -v umicollapse)")/../share" -maxdepth 2 -name umicollapse.jar -print -quit)
+
+    set -o pipefail
```
maxulysse: don't we have that in pipelines' nextflow.config already?

matthdsm:
Not sure about the contentious part here. Test are stable through ext.args, the rest is up to the pipeline dev. 

Also, fixed random seeds should be 42 and not some sensible number like 100

EDIT: fixed butterfingers on mobile keyboard

pinin4fjords:
> Not sure about the contentious part here. Test are stable through ext.args, the rest is ip to tje pipeline dev.

Agreed!

PR #12663, opened 2026-08-12
Add basic subworkflow for unzipping and indexing fasta

DLBPointon:
A pretty simple pattern i've seen in alot of places is unzipping fasta, cleaning and then index/sizes generation.

Thought I'd make it a subworkflow seeing as it's so common.

Linked to: https://github.com/sanger-tol/nf-core-modules/pull/274

### Further dev:
~~- Could potentially add something to clean headers too, alot of the tools we use don't like headers with any special characters or even spaces in a couple of cases.~~

~~EDIT: Just realised that it's set up for Sanger-tol data. I'll get that sorted.~~


subworkflows/nf-core/gunzip_fasta_index/main.nf
```diff
@@ -0,0 +1,70 @@
+include { GUNZIP                        } from "../../../modules/nf-core/gunzip/main"
+include { GAWK as GAWK_UPPER_SEQUENCE   } from '../../../modules/nf-core/gawk/main'
+include { SAMTOOLS_FAIDX                } from "../../../modules/nf-core/samtools/faidx/main"
+
+workflow GUNZIP_FASTA_INDEX {
```
prototaxites:
```suggestion
workflow FASTA_UNZIP_UPPER_GAWK {
```

Following the naming guidelines: https://nf-co.re/docs/specifications/components/subworkflows/naming-conventions#name-format-of-subworkflow-files

Anyone else have a better suggestion? I'm not used to the subworkflow naming conventions

DLBPointon:
Just spoke with Matthieu

Adding a header cleaner is likely a good idea, we already have a tool to do this in ASCC. Could also include some basic information about the fasta too (max_seq_length, total_length, n_seqs) in json format so that we don't pollute the meta.

subworkflows/nf-core/gunzip_fasta_index/main.nf
SPPearce: There is a #subworkflows channel on the slack, good to ask there.
I think maybe something more general like FASTA_PREPARE

SPPearce:
Maybe we could generate the dict file as well, for the tools that require that.

subworkflows/nf-core/gunzip_fasta_index/main.nf
```diff
@@ -0,0 +1,70 @@
...
+        /^>/ { print; next }
+        !/^>/ { print toupper(\$0) }
+        '''.stripIndent()
+    ).collectFile(name: "uppercase_sequence.awk", cache: true)
+    .collect()
+
+    GAWK_UPPER_SEQUENCE(
+        unzipped_reference,
+        ch_upper_sequence,
+        false,
```
prototaxites: Is it worth using `seqkit seq -u` for this?

https://bioinf.shenwei.me/seqkit/usage/#seq

Could add in some extra options - `-v` to validate the bases, there's an option to print sequence headers only, and it will also reformat all lines to a standard width.

subworkflows/nf-core/gunzip_fasta_index/main.nf
prototaxites: as I suggested on Slack: FASTA_REFORMAT_INDEX or FASTA_CLEAN_INDEX

subworkflows/nf-core/fasta_clean_faidx/main.nf
```diff
@@ -0,0 +1,116 @@
...
+
+
+    //
+    // MODULE: TRIM HEADERS AFTER THE FIRST SPACE OR SPECIAL CHARACTER
+    //         EITHER OF WHICH CAN CAUSE ERRORS IN DOWNSTREAM ANALYSIS
+    //
+    SEQKIT_TRIM (
+        SEQKIT_DOTS.out.fastx,
+        "fasta"
+    )
```
prototaxites: `seqkit seq` has the `--name` option which just retains the FASTA identifiers - could remove a separate operation

DLBPointon: Do you mean `--only-id` to get the identifier? `--name` returns the full header doesn't it?

prototaxites: Ah, yes, I misread the option

DLBPointon: Cool, yeah that makes sense. Updated that now and looks like everything is working.

subworkflows/nf-core/fasta_clean_faidx/main.nf
```diff
@@ -0,0 +1,106 @@
...
+                n_sequences : lengths.size(),
+                total_length: lengths.sum() ?: 0L,
+            ]
+            if (lengths) {
+                sequence_map.max_length = lengths.max()
+            }
+
+            def outfile = file("${params.outdir}/genome_stats/${meta.id}.genome_stats.json")
+            outfile.parent.mkdirs()
+            outfile.text = groovy.json.JsonOutput.prettyPrint(groovy.json.JsonOutput.toJson(sequence_map))
```
prototaxites: Don't like this, this should be configurable by the subworkflow user and publication may not even be desired.

Perhaps you could instead define a local native process and use that?

```
process GENOME_STATS {    
    executor "local"
    
    input: 
    tuple val(meta), path(fai)

    output:
    tuple val(meta), path("*.json")

    exec:

    <code...>

   def outfile = task.workDir.resolve("${prefix}.samplesheet.csv"))
   outfile.text = groovy.json.JsonOutput.prettyPrint(groovy.json.JsonOutput.toJson(sequence_map))
}
```

This can then be published with publishDir or the new workflow outputs if desired.

DLBPointon: I agree, but I went with the risky move because I could actually get it working.

However, after much headache, it is done.

subworkflows/nf-core/fasta_clean_faidx/main.nf
```diff
@@ -0,0 +1,127 @@
...
+    input:
+    // NOTE: Due to how the staging of files works, inputting the fai as a path
+    //       results in the process trying to seach in the wrong place.
+    tuple val(meta), val(fai)
+
+    output:
+    tuple val(meta), path("*.json"), emit: json
+
+    exec:
+    def outfile = task.workDir.resolve("${meta.id}.genome_stats.json")
```
prototaxites:
```suggestion
    def prefix = task.ext.prefix ?: "${meta.id}"
    def outfile = task.workDir.resolve("${prefix}.json")
```

prototaxites (commented):
Looking really good! One last small thought :)

subworkflows/nf-core/fasta_clean_faidx/main.nf
DLBPointon: Why did i even get rid of the prefix def? Re-added now.

Now, we should be done.

prototaxites (approved):
LGTM!

PR #12742, opened 2026-08-18
feat: add debreak module for long-read SV calling

### LINKED ISSUE #12740: new module: DeBreak

### Further Information

[DeBreak](https://github.com/Maggi-Chen/DeBreak) is a structural variant caller for long-read sequencing data (PacBio CLR, PacBio HiFi, Oxford Nanopore). It takes a sorted BAM + reference FASTA and outputs a standard VCF. Key features include breakpoint refinement via partial-order alignment, large insertion detection via local assembly, and duplication/insertion disambiguation.

Published in [Nature Communications (Chen et al. 2023)](https://doi.org/10.1038/s41467-023-37619-3).

Bioconda package: https://anaconda.org/bioconda/debreak (v1.3)
Biocontainers image: biocontainers/debreak:1.3--h9ee0642_0

```yaml
---
channels:
  - conda-forge
  - bioconda
dependencies:
  - bioconda::debreak=1.3
```

aksenia:
## Description

Adds a new module for [DeBreak](https://github.com/Maggi-Chen/DeBreak) (v1.3), a structural variant caller for long-read sequencing data (PacBio HiFi, PacBio CLR, Oxford Nanopore).

Closes #12740

## Module details

- **Input**: sorted BAM + BAI, optional reference FASTA
- **Output**: VCF with detected SVs
- **Container**: `quay.io/biocontainers/debreak:1.3--h9ee0642_0`
- **Label**: `process_high`
- Reference FASTA is optional but enables breakpoint refinement (`--poa`) and large insertion detection (`--rescue_large_ins`) when passed via `ext.args`

## Tests

- `homo_sapiens - [bam, bai] - no_fasta`: runs without reference
- `homo_sapiens - [bam, bai], fasta`: runs with reference FASTA
- `homo_sapiens - [bam, bai] - no_fasta - stub`: stub mode

All tests pass, lint 62/62 clean.

## Checklist

- [x] Module follows nf-core guidelines
- [x] `nf-core modules lint debreak` passes (62/62)
- [x] Tests run with Docker profile
- [x] Snapshot generated
- [x] `environment.yml` pinned to bioconda::debreak=1.3


modules/nf-core/debreak/main.nf
```diff
@@ -0,0 +1,40 @@
...
+        'https://depot.galaxyproject.org/singularity/debreak:1.3--h9ee0642_0':
+        'quay.io/biocontainers/debreak:1.3--h9ee0642_0' }"
+
+    input:
+    tuple val(meta), path(bam), path(bai)
+    tuple val(meta2), path(fasta)
+
+    output:
+    tuple val(meta), path("${prefix}/*.vcf"), emit: vcf
+    tuple val("${task.process}"), val('debreak'), eval('debreak --version | sed "s/debreak //"'), emit: versions_debreak, topic: versions
```
nourmahfel: This module reports version `1.0.2`, although the package is pinned to `1.3`. The incorrect version is also recorded in `main.nf.test.snap`.

It looks like DeBreak hardcodes its version string and it was not updated for the `v1.3` release. The [DeBreak v1.3 source used by the Bioconda recipe](https://github.com/ChongLab/DeBreak/blob/v1.3/debreak#L20) contains:

```python
parser.add_argument(
    '-v',
    '--version',
    action='version',
    version='DeBreak_v1.0.2'
)
```

Also, `sed "s/debreak //"` does not match `DeBreak_v1.0.2` because the case and format differ, so the substitution is a no-op.

Since the CLI cannot provide a reliable version, could you use the hardcoded-version pattern?

```nextflow
tuple val("${task.process}"), val('debreak'), val('1.3'), emit: versions_debreak, topic: versions
```

Please also add the standard warning above the `conda` directive:

```nextflow
// WARN: Version information not provided by tool on CLI. Please update version string below when bumping container versions.
```

modules/nf-core/debreak/meta.yml
```diff
@@ -0,0 +1,85 @@
...
+
+topics:
+  versions:
+    - - ${task.process}:
+          type: string
+          description: The name of the process
+      - debreak:
+          type: string
+          description: The name of the tool
+      - 'debreak --version | sed "s/debreak //"':
```
nourmahfel: If the version output is changed in `main.nf`, both the `output:` and `topics:` blocks need to describe the new value.

Could you update both entries to match the format used in `crabz/decompress/meta.yml`?

```yaml
- 1.3:
    type: string
    description: The name of the tool version
```

modules/nf-core/debreak/main.nf
```diff
@@ -0,0 +1,40 @@
...
+    container "${ workflow.containerEngine in ['singularity', 'apptainer'] && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/debreak:1.3--h9ee0642_0':
+        'quay.io/biocontainers/debreak:1.3--h9ee0642_0' }"
+
+    input:
+    tuple val(meta), path(bam), path(bai)
+    tuple val(meta2), path(fasta)
+
+    output:
+    tuple val(meta), path("${prefix}/*.vcf"), emit: vcf
```
nourmahfel: Every sample currently produces a file named `debreak.vcf`, so outputs will collide when two or more samples are published to the same directory.

The filename is [hardcoded upstream in `debreak_writevcf.py`](https://github.com/ChongLab/DeBreak/blob/v1.3/debreak_writevcf.py#L66):

```python
f = pysam.VariantFile(outpath + 'debreak.vcf', 'w', header=vcfheader)
```

This also means `${prefix}/*.vcf` can only match that fixed filename. The `svim/alignment` module handles a similar limitation by renaming the output with `mv`. Could you use the same approach here?

```nextflow
output:
tuple val(meta), path("*.vcf"), emit: vcf

script:
def prefix  = task.ext.prefix ?: "${meta.id}"
def args    = task.ext.args ?: ''
def ref_arg = fasta ? "-r ${fasta}" : ""
"""
debreak \\
    --bam ${bam} \\
    -o ${prefix} \\
    -t ${task.cpus} \\
    ${ref_arg} \\
    ${args}

mv ${prefix}/debreak.vcf ${prefix}.vcf
"""

stub:
"""
touch ${prefix}.vcf
"""
```

Once `prefix` is no longer referenced in the `output:` block, it can be changed back to `def prefix`.

The VCF pattern in `meta.yml` line 54 should also be changed to:

```yaml
pattern: "*.vcf"
```

The snapshot will then need to be regenerated.

modules/nf-core/debreak/tests/main.nf.test
```diff
@@ -0,0 +1,94 @@
+nextflow_process {
+
+    name "Test Process DEBREAK"
+    script "../main.nf"
+    process "DEBREAK"
```
nourmahfel, line 5: `tests/nextflow.config` is not currently loaded, so the configured `--min_support 1` argument is never applied.

The root `nf-test.config` points to the repository-level configuration:

```groovy
configFile = "tests/config/nf-test.config"
```

Module-level `tests/nextflow.config` files are only applied when the test declares them. Could you add the following to the process block?

```groovy
process "DEBREAK"
config "./nextflow.config"
```

Alternatively, the module-level configuration file can be removed if the additional argument is unnecessary. It is worth checking which behaviour is intended because, with test-sized data, the support threshold may determine whether DeBreak produces any calls.

nourmahfel (changes requested):
Thanks for this contribution. Only minor changes I would like to see addressed before merge. 

1. The module reports `DeBreak_v1.0.2`, although it is pinned to `debreak=1.3`.
2. Every sample produces an output named `debreak.vcf`, which can cause collisions when outputs are published.
3. `tests/nextflow.config` is not loaded, so `--min_support 1` is never applied.

I have added the implementation details and suggested changes inline :)

modules/nf-core/debreak/main.nf
aksenia: nice catch! I have implemented your recipe and filed an issue in DeBreak repo: https://github.com/Maggi-Chen/DeBreak/issues/12

modules/nf-core/debreak/main.nf
aksenia: thank you! Done. :)

nourmahfel (approved):
Very useful module to add! Thank you for implementing it into nf-core :)

modules/nf-core/debreak/tests/main.nf.test
```diff
@@ -0,0 +1,95 @@
...
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(
+                    process.out.findAll { key, val -> key.startsWith("versions") },
+                    process.out.vcf.collect { meta, vcf -> file(vcf).name }
+                ).match() }
```
maxulysse: I think something like this should simplify the logic and the snapshot
```suggestion
                { assert snapshot(sanitizeOutput(process.out, unstableKeys:['vcf'])).match() } ,
```

modules/nf-core/debreak/tests/main.nf.test
```diff
@@ -0,0 +1,92 @@
...
```
maxulysse:
```suggestion
                { assert snapshot(sanitizeOutput(process.out, unstableKeys:['vcf'])).match() } ,
```

modules/nf-core/debreak/tests/main.nf.test
```diff
@@ -0,0 +1,92 @@
...
+                ]
+                input[1] = [[], []]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
```
maxulysse:
```suggestion
                { assert snapshot(sanitizeOutput(process.out)).match() } ,
```

SPPearce:
@aksenia , please go to the nf-core Slack and post on github-invitations to be invited to the organisation, so the tests run.

PR #12744, opened 2026-08-18
Isolate bot PAT from untrusted PR code in comment-triggered workflows

ewels:
### What changed

`fix-linting.yml`, `update-gpu-snapshot.yml`, and `update-sentieon-snapshot.yml` are triggered by public PR comments (`issue_comment`) and previously checked out and executed pull-request-controlled code (`prek` hooks, `nf-test`) in the same job that held the `nf-core-bot` PAT. Because `actions/checkout` persists credentials into the local git config by default, PR-controlled code running in that job could read the bot's push-capable token — a classic "pwn request" pattern.

Each workflow is now split into:
- an **untrusted prepare/test job** — no bot PAT, `persist-credentials: false` — that runs the untrusted code and produces a patch artifact (never a full checkout token)
- a separate **privileged push job** — re-verifies the PR head SHA hasn't moved, applies the patch (data only, no code execution), and only then uses the bot PAT to commit and push with git hooks disabled

Also:
- `permissions: {}` set at the workflow level, with minimal per-job grants
- User-supplied snapshot test paths (from the comment body) are now parsed with an anchored allowlist regex and checked for symlinks/traversal before being passed to `nf-test`
- Sentieon secrets are now only used for pull requests whose head branch is in `nf-core/modules` (fork PRs no longer get Sentieon credentials)

No change to the contributor-facing commands or reactions — `@nf-core-bot fix linting` / `update gpu snapshot path: ...` / `update sentieon snapshot path: ...` behave the same from a comment author's perspective.

### Testing

- YAML/workflow syntax validated
- Verified job `permissions:`/`needs:` graphs match the intended trust boundaries (untrusted jobs never hold the bot PAT; the push jobs never execute PR-controlled code)
- Could not exercise the `issue_comment` trigger itself outside GitHub, so please review the job structure carefully before merging — happy to iterate based on feedback.

### DIFF
--- a/.github/workflows/update-gpu-snapshot.yml
+++ b/.github/workflows/update-gpu-snapshot.yml
@@ -4,8 +4,9 @@ on:
   issue_comment:
     types: [created]
 
+permissions: {}
+
 env:
-  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   # renovate: datasource=github-releases depName=askimed/nf-test versioning=semver
   NFT_VER: "0.9.5"
   NXF_ANSI_LOG: false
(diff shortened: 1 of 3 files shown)


.github/workflows/update-gpu-snapshot.yml
```diff
@@ -4,56 +4,125 @@ on:
...
+      issues: write
+      pull-requests: read
+    runs-on: ubuntu-latest
+    outputs:
+      head_repo: ${{ steps.pr.outputs.head_repo }}
+      head_sha: ${{ steps.pr.outputs.head_sha }}
+      test_path: ${{ steps.path.outputs.test_path }}
+      valid_path: ${{ steps.path.outputs.valid_path }}
     steps:
-      # indication that the snapshot is being updated
```
mashehu, line 24: this comments are useful imo

.github/workflows/update-gpu-snapshot.yml
```diff
@@ -4,56 +4,125 @@ on:
...
   NXF_VER: "25.10.2"
 
 jobs:
-  update-gpu-snapshot:
-    # Only run if comment is on a PR with the main repo, and if it contains the magic keywords
+  prepare-gpu-update:
     if: >
       contains(github.event.comment.html_url, '/pull/') &&
       contains(github.event.comment.body, '@nf-core-bot update gpu snapshot') &&
       github.repository == 'nf-core/modules'
```
mashehu:
```suggestion
      github.repository == 'nf-core/modules' &&
        contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association)
```
did something similar for tools, adds an extra barrier

mashehu (approved):
looks okay, but so much repetition. Can't we have a reusable workflow for all of this that takes runners, tag, profile, etc. as arguments and then does the rest?

.github/workflows/update-gpu-snapshot.yml
ewels: Agreed - restored.

ewels:
> Can't we have a reusable workflow for all of this that takes runners, tag, profile, etc. as arguments and then does the rest?

Yup, I would like that as well. I was thinking that this was a short-term quick fix to get stuff secure now, and that we could then work on a refactor (which may take longer to do). See Slack discussion for details.

SPPearce:
Merging this to get it in, additional changes can be made in a later PR

PR #12745, opened 2026-08-18
fix(fastq_qc_trim_filter_setstrandedness): stop genome fasta channel leaking sample meta into Salmon index

pinin4fjords:
When no genome fasta is supplied (e.g. a pseudoalignment-only run with kallisto), `ch_fasta` carries the sentinel value `[]`. The subworkflow derives the fasta to use for auto-strandedness Salmon indexing with:

```groovy
ch_fasta
    .combine(ch_strand_fastq.auto_strand)
    .map { items -> items.first() }
    .first()
    .set { ch_genome_fasta }
```

`combine()` splices list-valued items into the result. An empty list contributes zero elements, so `items.first()` resolves to the sample meta map from `ch_strand_fastq.auto_strand` instead of the fasta. That meta map then reaches `SALMON_INDEX` as the `genome_fasta` path input, which fails with `Not a valid path value type: java.util.LinkedHashMap`. Reproduces for any sample with `strandedness: auto` when no genome fasta and no pre-built Salmon index are supplied.

Fix wraps `ch_fasta`'s value in a singleton list before combining, so `[]` survives as its own tuple element instead of being flattened away. This also gives the right behaviour downstream: `SALMON_INDEX` already treats a bare `[]` `genome_fasta` as "build a decoy-free, transcript-only index", which is what should happen when there's no genome fasta.

Added a regression test for auto strandedness with no genome fasta supplied.

Found via an nf-core/rnaseq user report.

## PR checklist

- [x] `nf-test test subworkflows/nf-core/fastq_qc_trim_filter_setstrandedness/tests/main.nf.test --profile docker`

### DIFF
--- a/subworkflows/nf-core/fastq_qc_trim_filter_setstrandedness/main.nf
+++ b/subworkflows/nf-core/fastq_qc_trim_filter_setstrandedness/main.nf
@@ -379,10 +379,6 @@ workflow FASTQ_QC_TRIM_FILTER_SETSTRANDEDNESS {
     // SUBWORKFLOW: Sub-sample FastQ files and pseudoalign with Salmon to auto-infer strandedness
     //
     // Return empty channel if ch_strand_fastq.auto_strand is empty so salmon index isn't created
-    //
-    // Wrapped in a singleton list because combine() splices list-valued items into the result;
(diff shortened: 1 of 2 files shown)


SPPearce (approved):

pinin4fjords:
Thanks @SPPearce

PR #12771, opened 2026-08-21
Migrate custom/pcaclustering, custom/clustermetrics and custom/clustervisualization

dbaku42:
## Description

Migrates the three custom modules used by the `snpclustering` subworkflow to topic channels:

- `custom/pcaclustering`
- `custom/clustermetrics`
- `custom/clustervisualization`

These modules use the `template` directive, so they keep generating `versions.yml` and emit it with `topic: versions` (the recommended transitional style).

This allows the `snpclustering` subworkflow PR (#11879) to remain clean and avoid needing a second update after these modules are migrated.

## Related PR

Related to #11879

## Checklist

- [x] Modules emit versions with `topic: versions`
- [x] `meta.yml` files updated with correct `topics` section
- [x] Tests updated and passing
- [x] No changes to the `snpclustering` subworkflow in this PR


modules/nf-core/custom/clustermetrics/meta.yml
```diff
@@ -88,8 +88,8 @@ output:
 topics:
   versions:
     - versions.yml:
-        type: string
-        description: The name of the process
+        type: file
```
LouisLeNezet, line 91: Could you please copy paste the same description as just above ?

modules/nf-core/custom/clustervisualization/meta.yml
```diff
@@ -93,8 +93,8 @@ output:
 topics:
   versions:
     - versions.yml:
-        type: string
-        description: The name of the process
+        type: file
```
LouisLeNezet, line 96: Same as previous comment

modules/nf-core/custom/pcaclustering/meta.yml
```diff
@@ -73,8 +73,8 @@ output:
 topics:
   versions:
     - versions.yml:
-        type: string
-        description: The name of the process
+        type: file
```
LouisLeNezet, line 76: Same

modules/nf-core/custom/clustervisualization/tests/main.nf.test
```diff
@@ -18,20 +22,26 @@ nextflow_process {
...
-                    process.out.umap_tsv,
-                    process.out.tsne_tsv,
+                    process.out.umap_tsv.collect { it[0] },
+                    process.out.tsne_tsv.collect { it[0] },
                     process.out.versions,
                     path(process.out.versions[0]).yaml
-                ).match() }
+                ).match() },
+                { assert process.out.umap_tsv },
+                { assert process.out.tsne_tsv }
```
LouisLeNezet: Could you justify these changes ?
It would be better to use `sanitizeOutput(process.out)` + the path().yaml line for the snapshot.

LouisLeNezet (changes requested):
Why is it needed ?

modules/nf-core/custom/clustermetrics/meta.yml
dbaku42: The type: string / description: "The name of the process" under topics.versions in master was leftover boilerplate — it doesn't describe versions.yml correctly. Since I was already touching this block to migrate it to topic channels, I fixed the type/description to match what the file actually is (type: file, "File containing software versions"), consistent with the output.versions entry and with other nf-core modules.

modules/nf-core/custom/clustervisualization/meta.yml
dbaku42: The type: string / description: "The name of the process" under topics.versions in master was leftover boilerplate — it doesn't describe versions.yml correctly. Since I was already touching this block to migrate it to topic channels, I fixed the type/description to match what the file actually is (type: file, "File containing software versions"), consistent with the output.versions entry and with other nf-core modules.

modules/nf-core/custom/pcaclustering/meta.yml
dbaku42: The type: string / description: "The name of the process" under topics.versions in master was leftover boilerplate — it doesn't describe versions.yml correctly. Since I was already touching this block to migrate it to topic channels, I fixed the type/description to match what the file actually is (type: file, "File containing software versions"), consistent with the output.versions entry and with other nf-core modules.

modules/nf-core/custom/clustervisualization/tests/main.nf.test
dbaku42: thanks @LouisLeNezet all test pass now

modules/nf-core/custom/clustervisualization/tests/main.nf.test
```diff
@@ -22,9 +22,7 @@ nextflow_process {
             assertAll(
                 { assert process.success },
                 { assert snapshot(
-                    process.out.umap_tsv,
-                    process.out.tsne_tsv,
-                    process.out.versions,
+                    sanitizeOutput(process.out, unstableKeys: ["umap_tsv", "tsne_tsv", "umap_png", "tsne_png"]),
```
LouisLeNezet: You don't need here to remove the md5sum of the tsv file as they are stable.
```suggestion
                    sanitizeOutput(process.out, unstableKeys: ["umap_png", "tsne_png"]),
```

modules/nf-core/custom/clustermetrics/meta.yml
```diff
@@ -88,8 +88,8 @@ output:
...
+        description: File containing software versions
```
LouisLeNezet, line 92: Please use all the information above:

```yaml
    - "versions.yml":
        type: file
        pattern: "versions.yml"
        ontologies:
          - edam: http://edamontology.org/format_3750
```

LouisLeNezet: Same for the other meta.yml

LouisLeNezet (changes requested):
Just a few changes needed

dbaku42:
Hi @LouisLeNezet,
Is there anything else that need to be fixed?

modules/nf-core/custom/clustervisualization/tests/main.nf.test
LouisLeNezet: This isn't resolved ?

LouisLeNezet:
Hi @dbaku42,
There is just the issue with "umap_tsv", "tsne_tsv" stability md5 to solve.
Is there a reason that you removed the md5sum ?

dbaku42:
Hi @LouisLeNezet,

Sorry for the back and forth on this — turns out my earlier commit didn't actually apply your suggestion (the `unstableKeys` still included `umap_tsv`/`tsne_tsv`), which is why the md5 issue seemed to persist. That's fixed now: the test no longer uses `sanitizeOutput`/`unstableKeys` at all and matches the structure already used on `master` — `umap_tsv`/`tsne_tsv` are checked normally via `snapshot().match()`, and only the pngs are excluded from the real-run snapshot (kept only in the stub test, where they're deterministic placeholders).

However, while regenerating the snapshot I found something that changes the picture a bit: the tsv md5 is **not** stable across execution profiles. With the `conda` profile I get:

test.umap.tsv:md5,d338c1ef0e979dbf653e21c3417e975e
test.tsne.tsv:md5,eb00f27d82530f665552b158e8e3c8ff


With `docker` I get:

test.umap.tsv:md5,7796626109d85349a09afd10e46b1639
test.tsne.tsv:md5,982f84681db9c804c70e91a9b366fee4


Same inputs, different output content depending on the profile — likely a difference in the underlying numerical libraries (numpy/scikit-learn/umap-learn) between the conda env and the docker image, causing tiny floating-point differences in the tsv values.

This means a single committed snapshot can't pass both the conda and docker CI jobs at once. Do you know if `environment.yml` and the container image are expected to pin the exact same library versions here? If not, I think we're back to needing `umap_tsv`/`tsne_tsv` in `unstableKeys` after all — not because the process itself is non-deterministic, but because it isn't reproducible bit-for-bit across profiles. Happy to go either way, just want to confirm before I touch the test again.

modules/nf-core/custom/clustervisualization/tests/main.nf.test
dbaku42: Hi @LouisLeNezet,

Just to give a status update on this: I've made several attempts to fix the tsv md5 stability issue, and thought I had it (the root cause was UMAP and TSNE both using multi-threaded computation by default, causing non-deterministic output even with a fixed `random_state`, so I added `n_jobs=1` to both). Locally the fix confirmed stable md5s across repeated runs and across conda/docker profiles.

However, 3 CI jobs are still failing. I'm still investigating — will update here once I've confirmed the actual cause. Apologies for the delay in getting this fully green.

dbaku42:
Closing this. The three processes already emit topic: versions on master (path "versions.yml", emit: versions, topic: versions), so there is nothing left to “swap” in main.nf.This PR ended up as meta.yml wording + a n_jobs=1 / snapshot regen for clustervisualization. The snapshot hashes are not stable across conda/docker/singularity (or even local Docker vs GHA), which is why only shard 3 is red. That is unrelated to topics.I’ll:
leave the modules as they are on master
drop the .mix(CUSTOM_*.out.versions) lines in #11879 so the subworkflow is born on topics
if we still want the topics.versions meta.yml cleanup, that can be a tiny follow-up without touching snapshots

PR #12773, opened 2026-08-22
Custom eigenvectotsv

dbaku42:
## Description

This PR extracts `EIGENVEC_TO_TSV` — currently an inline process inside the `snpclustering`
subworkflow (#11879) — into a standalone, reusable nf-core module: `custom/eigenvectotsv`.

The module converts the `.eigenvec` file produced by `plink2/pca` into a tab-separated file,
renaming the `IID` column to `sample_id` for downstream compatibility with the
`custom/pcaclustering`, `custom/clustermetrics`, and `custom/clustervisualization` modules
used later in the `snpclustering` pipeline.

Splitting this out as its own module:
- gives it dedicated nf-test coverage (content-level snapshot, not just filename checks),
  addressing the concern raised by @famosab in #11879 that subworkflow snapshots were only
  checking file names rather than content
- uses the current topic-channel convention for version reporting (`topic: versions`),
  consistent with the topic-based migration already underway in #11879
- allows it to be reused independently of the `snpclustering` subworkflow

### DIFF
--- a/modules/nf-core/custom/eigenvectotsv/environment.yml
+++ b/modules/nf-core/custom/eigenvectotsv/environment.yml
@@ -1,0 +1,6 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/environment-schema.json
+channels:
+  - conda-forge
+  - bioconda
+dependencies:
+  - "conda-forge::gawk=5.3.0"
--- a/modules/nf-core/custom/eigenvectotsv/main.nf
+++ b/modules/nf-core/custom/eigenvectotsv/main.nf
@@ -1,0 +1,43 @@
+process CUSTOM_EIGENVECTOTSV {
+    tag "$meta.id"
+    label 'process_single'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/gawk:5.3.0' :
+        'quay.io/biocontainers/gawk:5.3.0' }"
+
+    input:
+    tuple val(meta), path(eigenvec)
+
+    output:
+    tuple val(meta), path("*.tsv"), emit: tsv
+    tuple val("${task.process}"), val('gawk'), eval("awk --version | head -1 | sed 's/GNU Awk //;s/,.*//'"), emit: versions_gawk, topic: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    awk '
+    NR==1 {
+        sub(/^#/, "")
+        if (\$1 == "IID") {
+            \$1="sample_id"
+        }
+        print
+        next
+    }
+    {
+        print
+    }' OFS='\\t' ${args} ${eigenvec} > ${prefix}.tsv
+    """
+
+    stub:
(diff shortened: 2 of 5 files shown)


mahesh-panchal:
There's already a `gawk` module. We shouldn't be accepting custom modules that can be replicated with existing modules.

dbaku42:
Hi @mahesh-panchal,

Thanks for pointing this out. I agree that the transformation could technically be implemented using the existing `gawk` module.

The reason I extracted `EIGENVEC_TO_TSV` as a dedicated custom module is that it is part of the specific workflow logic of the `custom/snpclustering` module. It performs the conversion of the PLINK2 `.eigenvec` output to TSV and renames the `IID` column to `sample_id`, which is the format expected by the downstream PCA/clustering steps within `snpclustering`.

Keeping this as a separate process also makes the `snpclustering` workflow easier to read and allows this specific transformation to be tested independently.

However, if the preferred approach is to use the existing `gawk` module for this step, I can replace `EIGENVEC_TO_TSV` accordingly.

mahesh-panchal:
I don't think a new module is the way to go here, even for the sake of getting dedicated test coverage. Renaming a header column and tab-separating a file is exactly what the existing `GAWK` module already does, so wrapping it in a bespoke `custom/eigenvectotsv` module duplicates that tool, adds another container to maintain, and isn't reusable outside `snpclustering`.

I'd propose fixing this in #11879 instead of here: replace the inline `EIGENVEC_TO_TSV` process with a call to the existing `GAWK` module. `subworkflows/nf-core/gtf_hybridmerge_gffcompare` (`GAWK_FILTER`/`GAWK_CONCAT`) is an example you can follow for the same situation — a small text transform stitched between two nf-core modules.

1. Move the awk program out of the inline process into its own file next to the subworkflow, e.g. `subworkflows/nf-core/snpclustering/awk/eigenvec_to_tsv.awk`:
   ```awk
   NR==1 {
       sub(/^#/, "")
       if ($1 == "IID") {
           $1 = "sample_id"
       }
       print
       next
   }
   { print }
   ```

2. In `subworkflows/nf-core/snpclustering/main.nf`, drop the `EIGENVEC_TO_TSV` process and call `GAWK` instead:
   ```groovy
   include { GAWK as GAWK_EIGENVEC_TO_TSV } from '../../../modules/nf-core/gawk'
   ...
   ch_eigenvec_awk = file("${moduleDir}/awk/eigenvec_to_tsv.awk", checkIfExists: true)

   GAWK_EIGENVEC_TO_TSV(PLINK2_PCA.out.evecfile, ch_eigenvec_awk, false)
   ch_versions = ch_versions.mix(GAWK_EIGENVEC_TO_TSV.out.versions_gawk)
   ```
   Note the output emit becomes `GAWK_EIGENVEC_TO_TSV.out.output`, not `.out.tsv` — the downstream references (`CUSTOM_PCACLUSTERING` input, `ch_cluster_analysis_input`, and the subworkflow's `tsv` emit) need updating to match. The output keeps its default naming (`<meta.id>.eigenvec`); none of the downstream modules require a `.tsv` extension, only tab-separated content.

3. For the content-level snapshot coverage raised, I'd extend the `SNPCLUSTERING` nf-test to snapshot the tsv content (header rename + tab-separation) rather than add a new module's test suite — `GAWK` already has its own module-level tests, so there's no need to re-test the tool itself, only your specific awk program's output.

dbaku42:
Makes sense, thanks for the detailed pointer — I'll close this PR and apply the GAWK_EIGENVEC_TO_TSV approach directly in #11879 instead, following the gtf_hybridmerge_gffcompare pattern.

PR #12781, opened 2026-08-24
hmmer/hmmrank: add per-domain alignment coordinates

erikrikarddaniel:
## PR checklist

Relates to https://github.com/nf-core/phyloplace/issues/70 (different repo, so this won't auto-close it)

- [ ] Add a resource `label`
  - For modules:
    - [x] `nf-core modules test hmmer/hmmrank --profile docker`
    - [ ] `nf-core modules test hmmer/hmmrank --profile singularity`
    - [ ] `nf-core modules test hmmer/hmmrank --profile conda`
  - For subworkflows:
    - [x] `nf-core subworkflows test fasta_hmmsearch_rank_fastas --profile docker`
    - [ ] `nf-core subworkflows test fasta_hmmsearch_rank_fastas --profile singularity`
    - [ ] `nf-core subworkflows test fasta_hmmsearch_rank_fastas --profile conda`

## Description

`hmmrank` summarises hmmsearch hits per sequence, which is the right level for "which profile does this protein belong to", but it carries no alignment coordinates. That makes two things impossible from the ranked table alone: working out how much of a profile a hit actually covers, and telling one continuous match apart from several pieces scattered along the target. Those coordinates exist only in hmmsearch's per-domain table.

The motivating case is a class II ribonucleotide reductase split over two to four adjacent ORFs, where no single ORF covers enough of the profile to be classified, but together they cover essentially all of it.

### `hmmer/hmmrank`

The domtblout files become an **optional** third element of the input tuple. When they are supplied, each `(accno, profile)` pair gains `tlen` and `qlen`, plus `from`, `to`, `len` and `n_islands` for each of the `hmm`, `ali` and `env` coordinate sets. Ranking still comes from the tblout numbers; the coordinates are joined on `(accno, profile)`.

`from` and `to` are the outer bounds over the pair's domains. `len` is the size of the **union** of those domains, which is the only part needing care: `to - from + 1` counts the gaps between separate matches, and summing over domains double counts the overlaps. `n_islands` is how many disjoint pieces the pair falls into, which comes out of the same calculation for free.

The union is computed in one sorted pass rather than the repeated self-join this kind of code usually gets:

```r
mutate(island = cumsum(f > lag(cummax(t), default = 0L)))
```

Sorting by start means anything overlapping a row lies before it, so `cummax` of the ends, lagged, is how far everything before the current row reached. A start beyond that opens a new island, and since starts are sorted no later domain can fill that gap either. Using `cummax` rather than the previous end is what makes fully contained and duplicated domains need no special handling. The `0L` default is deliberate: `lag()` casts it to the column type and errors on a lossy double-to-integer cast.

**Omitting the domtblouts leaves the output exactly as it was**, which matters because nf-core/ampliseq vendors this module too. The pre-existing `hmmrank - 16S` snapshot passes unchanged.

### Drive-by fix

The patterns deriving the profile name from the filename were regexes with an unescaped dot and no anchor, so `'.tbl.gz'` matched any character followed by `tbl` then any character then `gz`. On `NrdJ.domtbl.gz` that stripped `mtbl.gz` and gave the profile as `NrdJ.do`. Now anchored and escaped, which also hardens the existing tblout path.

### `fasta_hmmsearch_rank_fastas`

Passes `HMMER_HMMSEARCH.out.domain_summary` on to `HMMER_HMMRANK` when `save_domtblout` asked for those tables. Since `domain_summary` is an optional output it leaves an empty channel when the flag is off, which would stall the `combine` and stop the ranking running at all, so it goes through `ifEmpty`. The non-`save_domtblout` snapshot is unchanged; only the `save_domtblout` one moves.

### Testing

Beyond the existing tests, there is a new one asserting the full expected table for curated input rather than snapshotting whatever comes out, so it cannot drift into blessing wrong output. The data (nf-core/test-datasets#2232, merged) is excerpted from a real search of the `NrdJ` and `NrdJm1` profiles against GTDB r226 representatives: five targets, nine pairs, chosen to cover a single domain, overlapping domains, disjoint domains, and domains that overlap in profile coordinates while lying far apart on the target.

That last case is the point. `AACBMDAH_04658` against `NrdJm1` spans positions 14 to 803 on the sequence but covers only 400 of them, and that gap between span and merged length is exactly the signal the ranked table could not previously show.


piplus2 (commented):
I've asked Claude to help me with the review:

### 1. `n_islands` counts abutting domains as separate islands

  `modules/nf-core/hmmer/hmmrank/main.nf`:

  ```r
  mutate(island = cumsum(f > lag(cummax(t), default = 0L)))
  ```

  `f > previous_max_end` opens a new island, so domains that abut exactly (one ends at 96, the next starts at 97) are counted 
  as two pieces even though the coverage is continuous. `len` is unaffected; only `n_islands` is.

  This isn't hypothetical — it fires on two of the nine rows being blessed as ground truth:

  | pair | hmm domains | asserted `hmm_n_islands` | actual disjoint pieces |
  |---|---|---|---|
  | `AAGAFPGM_00302`/`NrdJ` | (12,47) (60,96) **(97,186)** (191,348) (340,406) | 4 | 3 |
  | `DPHPGHLN_02900`/`NrdJ` | (62,96) **(97,187)** (205,337) (336,384) | 3 | 2 |

  The meta.yml calls it "the number of disjoint pieces it falls into", and telling one continuous match apart from several
  scattered pieces is the stated motivation for the PR — so this is the column that matters most. The fix is one token:

  ```r
  mutate(island = cumsum(f > lag(cummax(t), default = 0L) + 1L))
  ```

 If counting abutting domains separately *is* deliberate, could you say so in the meta.yml? As written, "disjoint" doesn't
  mean disjoint.

### 2. Coverage includes domains hmmer doesn't consider significant

  The parse captures c-Evalue and i-Evalue as `d9`/`d10` and discards them, so every *reported* domain feeds the union. On this
  test data that is not a small effect:

  | pair | `hmm_len` (all domains) | `hmm_len` (i-Evalue ≤ 0.01) | non-significant domains |
  |---|---|---|---|
  | `AAGAFPGM_00302`/`NrdJ` | 379 | 158 | 4/5 |
  | `MNMMFBCN_00658`/`NrdJ` | 353 | 148 | 4/5 |
  | `DPHPGHLN_02900`/`NrdJ` | 306 | 133 | 3/4 |
  | `AACBMDAH_04658`/`NrdJ` | 68 | 0 | 2/2 |

  The first data row of `NrdJ.domtbl.gz` has i-Evalue `1.1e+03` and score `12.3` — and it is what sets `hmm_from = 12`.

  I suspect this is intentional: `AACBMDAH_04658`/`NrdJ` is a significant *full-sequence* hit whose individual domains are all
  weak, which is exactly the fragmented case you're chasing, and any threshold would erase it. But someone reading "how much of
  the profile does this hit cover" won't expect weak envelopes in the number. Could you document the behaviour in meta.yml,
  and maybe expose an i-Evalue cutoff via `task.ext.args`? The columns are already parsed, so it's cheap.

erikrikarddaniel:
Thanks @piplus2 ! Interesting with Claude checking its own PRs. Now, I'll let it have a look at the conversation.

erikrikarddaniel:
### 1. Abutting domains

Real bug, and the argument that settles it is internal: `len` is computed as `sum(e - s + 1)`, i.e. inclusive coordinates. Under that same convention there is no gap between 96 and 97, so `len` was already treating those domains as one continuous stretch while `n_islands` called them two. The two columns disagreed about the same union, two lines apart. Applied your one-token fix and said why in the comment.

Scope, checked cell by cell rather than assumed: across the 27 (pair, coordinate-set) cells of the known-table test, `n_islands` moves on exactly the two you identified and `from`/`to`/`len` move nowhere — as expected, since the union never contained the phantom gap. The subworkflow's 16S data turned out to have one abutting case too (`euk18s` env, 2 → 1), which is why its snapshot md5 changes; I confirmed all 273 cells of that table against an independent union implementation before updating it. The non-`domtblout` snapshots are untouched, so backwards compatibility for ampliseq and phyloplace is unaffected. 

### 2. Weak domains in the coverage

Your reading is right, and it is deliberate — `AACBMDAH_04658`/`NrdJ` is the case that makes it so. It clears the full-sequence threshold on the strength of its domains taken together while every individual domain is weak (i-Evalue 1e+04 and 6.6e+03), and any per-domain cutoff blanks it to `hmm_len = 0` with no coordinates at all. That is exactly the fragmented hit these columns exist to expose, so filtering by default would hide the motivating case.

You are right that it wasn't written down anywhere, though, and "how much of the profile does this cover" does invite the other reading. The `domtblouts` description now states it outright.

On exposing a cutoff through `task.ext.args`: I'd rather not in this PR. The module has no `args` handling today and nothing resembling a command line to append to, so it would mean inventing an option-parsing convention inside the R script — a config surface on a module two pipelines vendor, added on speculation. And the output is one row per (accno, profile) while the threshold is per-domain, so a filtered-out pair yields NA coordinates rather than a smaller number, which needs more explaining than it saves. Easy to add later behind a concrete use case. Happy to be argued out of it if you have one in mind.

*(Reviewed and revised with help from Claude Code.)*

piplus2 (approved):
LGTM

PR #12786, opened 2026-08-25
fix(comebin/runcomebin): include database inside the container

dialvarezs:
Replaces the earlier `checkm_db` parameter approach in this PR, per the discussion below.

## Problem

COMEBin runs `checkm analyze` and `checkm qa` to pick its final binning result, so it needs the CheckM reference data. The container doesn't carry it, so any real run fails:

```
ERROR: Input file does not exists: /opt/conda/checkm_data/pfam/Pfam-A.hmm.dat
ERROR: Models must be parsed before identifying HMM hits.
```

The `checkm-genome` post-link script *does* download the database. It dies unpacking it, because Wave's `mambaorg/micromamba:2-amazon2023` build image has no `tar` ([build log](https://wave.seqera.io/view/builds/bd-92fd7581742206a2_1), step `#10 [build 3/3]`):

```
checkm_data_2015_01_16.tar.gz saved [288590617/288590617]
/opt/conda/staging/checkm_data_2015_01_16.tar.gz: OK
/opt/conda/bin/.checkm-genome-post-link.sh: line 50: tar: command not found
```

It still runs `checkm data setRoot` afterwards, which is why the image ends up with a data root pointing at an empty directory instead of an obvious failure:

```
$ apptainer exec --cleanenv --no-home <img> python -c \
    "import os; d='/opt/conda/checkm_data'; print(os.listdir(d), os.path.getsize(d+'/.dmanifest'))"
['.dmanifest'] 68
```

Only a stub test exists for this module, which is why this went unnoticed.

## Change

Pin `tar` and `gzip` so the post-link script can finish, and rebuild the containers.

## Testing

The rebuilt container carries the database:

```
checkm_data: 1.39 GB
HMM_MODELS exists: True
PFAM exists: True
```

## Notes for reviewers

The containers here were built with Wave's `conda/micromamba:v2` template. `nf-core modules containers create` hardcodes `conda/pixi:v1`, and pixi does not run conda post-link scripts at all, so re-running that command on this module would silently drop the database again. That's worth an nf-core/tools issue. It also explains the conda lock file switching format in this diff, from pixi's YAML to micromamba's explicit list.

Image size goes up by the database, ~1.4 GB.

I'm working both on GPU support and a test that can run on CI, will continue on follow-up PRs.

## PR checklist

- [ ] If you've fixed a bug or added code that should be tested, add tests! — no real test yet, there is still no test dataset that produces bins; see note above
- Ensure that the test works with either Docker / Singularity:
  - [x] `nf-core modules test comebin/runcomebin --profile singularity`
  - [ ] `nf-core modules test comebin/runcomebin --profile docker` — not run
  - [ ] `nf-core modules test comebin/runcomebin --profile conda` — not run


prototaxites:
I'm working on a fix for this also, fixing the container build to include the CheckM database: https://nfcore.slack.com/archives/C09V4RAHV0X/p1787646230855859

prototaxites:
amd64 builds: 

Docker: `community.wave.seqera.io/library/comebin:1.1.0--92fd7581742206a2`
Singularity ORAS: `oras://community.wave.seqera.io/library/comebin:1.1.0--7de47d4800e371c1`
Singularity HTTPS: `https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/c6/c664441f34d1de5cd187fe7704b865b6ac570f80e9a222fd4f3a8912d8179d75/data`

The problem is that nf-core tools forces container builds using Pixi, which doesn't run by default the post-install script for CheckM that initialises the database. The Seqera containers website uses Conda, which will run these scripts - so the above were built that way and the scripts have run according to the log.

Though allowing a user-specifiable DB is not a bad extension to the module!

modules/nf-core/comebin/runcomebin/main.nf
prototaxites:
```suggestion
       ? 'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/c6/c664441f34d1de5cd187fe7704b865b6ac570f80e9a222fd4f3a8912d8179d75/data'
       : 'community.wave.seqera.io/library/comebin:1.1.0--92fd7581742206a2'}"
```

modules/nf-core/comebin/runcomebin/meta.yml
prototaxites:
```suggestion
```

dialvarezs:
> I'm working on a fix for this also, fixing the container build to include the CheckM database: [nfcore.slack.com/archives/C09V4RAHV0X/p1787646230855859](https://nfcore.slack.com/archives/C09V4RAHV0X/p1787646230855859)

Oh, I didn't see that thread, sorry!

> Though allowing a user-specifiable DB is not a bad extension to the module!

But if we get the database included in the container with your fix, then adding the parameter won't be useful after all 🤔. CheckM shouldn't change at this point, so I don't see why you would want to supply a db different from the included one.

prototaxites:
> But if we get the database included in the container with your fix, then adding the parameter won't be useful after all 🤔. CheckM shouldn't change at this point, so I don't see why you would want to supply a db different from the included one.

I don't know how clever CheckM is - is it possible to run with a custom DB?

dialvarezs:
> I don't know how clever CheckM is - is it possible to run with a custom DB?

In theory it seems possible, it doesn't have a checksum validation like CheckM2, but the db uses a specific structure (https://github.com/Ecogenomics/CheckM/blob/master/checkm/defaultValues.py) and the scripts to build it are not  available in the repo. So in practice it would be tricky.

So I would prefer dropping the parameter instead and keep it as it is. Suplying a custom db seems a very niche feature anyway.

prototaxites:
OK, sounds like a plan. Let's bump the containers and check that they work for now, we can always revisit this later.

modules/nf-core/comebin/runcomebin/main.nf
dialvarezs: These containers don't have the db either, `/opt/conda/checkm_data/` is empty.

dialvarezs: Well, it seems that it's because tar is not in the base image
```
#10 62.15 2026-08-25 08:32:49 (31.2 MB/s) - ‘/opt/conda/staging/checkm_data_2015_01_16.tar.gz’ saved [288590617/288590617]
#10 62.15 
#10 62.15 /opt/conda/staging/checkm_data_2015_01_16.tar.gz: OK
#10 62.15 /opt/conda/bin/.checkm-genome-post-link.sh: line 50: tar: command not found
```

prototaxites (approved):

PR #12788, opened 2026-08-25
hmmer/hmmrank: key domain coordinates on the query model

erikrikarddaniel:
## PR checklist

Closes #XXX 

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

## Description

Fixes a bug in the coordinate columns added to `hmmer/hmmrank` in #12781, found while reviewing the pipeline PR that vendors them (nf-core/phyloplace#72).

### The bug

`hmmrank` takes its `profile` column from the **file name** of each table. But an HMM file may hold several models, and the whole file is then searched in one go, so a single `*.tbl.gz` / `*.domtbl.gz` pair can carry rows for several query names under one file-derived `p
rofile`. The new coordinate code keyed on `(accno, profile)` alone, which breaks in two ways:
   
- `distinct(accno, profile, tlen, qlen)` returns one row per model, since `qlen` differs between models. The `left_join` against a tblout side that likewise has one row per model is then a cartesian product.
- `islands()` grouped by `(accno, profile)` only, so it unioned domains belonging to *different* models. `hmm_from`/`hmm_to` are positions in the profile, so a union across two profiles is not meaningful.

Concatenating the two existing `NrdJ` fixtures into one file reproduces it — **17 rows instead of 9, and every coordinate corrupted**:
  
| accno | model | `hmm_from` `hmm_to` `hmm_len` | correct |
| --- | --- | --- | --- | 
| `AACBMDAH_04658` | NrdJ | 1 397 397 | 11 95 68 | 
| `AAGAFPGM_00302` | NrdJ | 6 711 676 | 12 406 379 |
| `AAGAFPGM_00302` | NrdJm1 | 6 711 676 | 6 711 588 |

`AACBMDAH_04658`/NrdJ simply reports NrdJm1's coordinates. `AAGAFPGM_00302` gets `hmm_len` 676 for both models — a cross-model union matching neither.

Nothing fires without `--domtblout`, so output from before #12781 is unaffected.

### The fix

The query name is already present on both sides and distinguishes the models: column 3 of the tblout, parsed as `profile_desc`, and column 4 of the domtblout, until now discarded as a dummy. Keeping it and carrying it through `islands()` and the joins closes both halves.

**Output columns are unchanged.** The final join maps `profile_desc` onto the domtblout query (`by = c('accno', 'profile', 'profile_desc' = 'query')`), so no new column is emitted and single-model input produces byte-identical output. All existing `hmmer/hmmrank` and `fasta_hmmsearch_rank_fastas` snapshots pass untouched — worth stating explicitly, since nf-core/ampliseq vendors both.

### Tests

A regression test builds the multi-model case by gunzipping, concatenating and re-gzipping the two existing fixtures inline, so no near-duplicate test data is needed. It fails on the previous code (`17 != 9`) and passes here, asserting the same per-model coordinate table as the single-file test, keyed on the query name.

### Two smaller things from the same review

- `tests/main.nf.test` read the process output in the `then` block, before `assert process.success` ran inside `assertAll`. On a failure `process.out.hmmrank` is empty and that line threw `IndexOutOfBoundsException`, hiding the real failure and its log. The reads now happen inside the assertions.
- The `hmm_len` assertion required an integer on every row, which rules out the `NA` a hit legitimately gets when it has no domain records — `hmmsearch` reports a sequence on the per-sequence threshold, while a domain also has to clear the per-domain one. `NA` is the intended behaviour there, being the clearest signal that no domains were reported for the hit, so the assertion now allows it and `meta.yml` says so.

<sub>Prepared with help from Claude Code.</sub>

### DIFF
--- a/modules/nf-core/hmmer/hmmrank/tests/main.nf.test
+++ b/modules/nf-core/hmmer/hmmrank/tests/main.nf.test
@@ -89,12 +89,12 @@ nextflow_process {
                 { assert path(process.out.hmmrank[0][1]).linesGzip.size() > 1 },
                 { def rows = path(process.out.hmmrank[0][1]).linesGzip
                   def header = rows[0].split('\t') as List
+                  def hmmLens = rows[1..-1].collect { it.split('\t')[header.indexOf('hmm_len')] }
                   // NA is a legitimate value: hmmsearch reports a sequence on the per-sequence
-                  // threshold, so a hit can be ranked with no domain records behind it at all
-                  assert rows[1..-1].every {
-                      def v = it.split('\t')[header.indexOf('hmm_len')]
-                      v.isInteger() || v == 'NA'
-                  } },
+                  // threshold, so a hit can be ranked with no domain records behind it at all.
+                  // any() as well as every(), since all-NA is what a broken join looks like
+                  assert hmmLens.every { it.isInteger() || it == 'NA' }
+                  assert hmmLens.any { it.isInteger() } },
                 { assert snapshot(
                     sanitizeOutput(process.out),
                     path(process.out.versions[0]).yaml
@@ -174,11 +174,13 @@ nextflow_process {
                 def combine = { name, srcs ->
                     def out = file("\${workDir}/multimodel/\${name}")
                     out.parent.mkdirs()
-                    def os = new java.util.zip.GZIPOutputStream(out.newOutputStream())
-                    srcs.each { src ->
-                        os.write(new java.util.zip.GZIPInputStream(file(src, checkIfExists: true).newInputStream()).bytes)
+                    new java.util.zip.GZIPOutputStream(out.newOutputStream()).withStream { os ->
+                        srcs.each { src ->
+                            new java.util.zip.GZIPInputStream(file(src, checkIfExists: true).newInputStream()).withStream { gz ->
+                                os.write(gz.bytes)
+                            }
+                        }
                     }
-                    os.close()
                     out
                 }
                 def base = params.phyloplace_testdata_base_path + 'hmmrank/'
--- a/modules/nf-core/hmmer/hmmrank/tests/main.nf.test
+++ b/modules/nf-core/hmmer/hmmrank/tests/main.nf.test
@@ -75,20 +75,26 @@ nextflow_process {
         }
 
         then {
-            def rows = path(process.out.hmmrank[0][1]).linesGzip
-            def header = rows[0].split('\t') as List
-
+            // read the output only inside the assertions, so a failed process is reported as
+            // such instead of throwing out of this block and hiding its log
             assertAll(
                 { assert process.success },
                 // the coordinate columns are present and populated
-                { assert header.containsAll([
+                { assert path(process.out.hmmrank[0][1]).linesGzip[0].split('\t').toList().containsAll([
                     'tlen', 'qlen',
                     'hmm_from', 'hmm_to', 'hmm_len', 'hmm_n_islands',
                     'ali_from', 'ali_to', 'ali_len', 'ali_n_islands',
                     'env_from', 'env_to', 'env_len', 'env_n_islands'
                 ]) },
-                { assert rows.size() > 1 },
-                { assert rows[1..-1].every { it.split('\t')[header.indexOf('hmm_len')].isInteger() } },
+                { assert path(process.out.hmmrank[0][1]).linesGzip.size() > 1 },
+                { def rows = path(process.out.hmmrank[0][1]).linesGzip
+                  def header = rows[0].split('\t') as List
+                  // NA is a legitimate value: hmmsearch reports a sequence on the per-sequence
+                  // threshold, so a hit can be ranked with no domain records behind it at all
+                  assert rows[1..-1].every {
+                      def v = it.split('\t')[header.indexOf('hmm_len')]
+                      v.isInteger() || v == 'NA'
+                  } },
                 { assert snapshot(
                     sanitizeOutput(process.out),
                     path(process.out.versions[0]).yaml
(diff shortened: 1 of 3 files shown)


piplus2 (commented):
The fix looks correct to me. I checked the keying against the HMMER3 tabular formats rather than taking the description on trust:

- `--domtblout` is `target, tacc, tlen, query, qacc, qlen, [E-value, score, bias, #, of, c-Evalue, i-Evalue, score, bias], hmm_from, hmm_to, ali_from, ali_to, env_from, env_to, acc, desc` — 23 fields. Renaming `d1` to `query` and shifting `d2..d10` over the nine middle fields keeps the count and still lands `hmm_from` on column 16.
- `--tblout` column 3 is the query name, which the module already parses as `profile_desc`, so `by = c('accno', 'profile', 'profile_desc' = 'query')` is model-name to model-name.
- Output columns really are unchanged: `left_join` with a renamed key keeps `profile_desc` from the left side and drops `query`. The untouched snapshots confirm it.

Both failure modes are real and the extra key closes both. The regression test is a good one — the expected table is identical to the two-file test's, so it proves concatenation is a no-op, and `linesGzip.size() == expected.size() + 1` pins the cartesian product directly.

Four comments, none blocking. Two of them are on lines outside the diff, so they are here rather than inline:

**`main.nf:97` — the rank tie-break has the same root cause.**

```r
# ...let ties be resolved by profile in alphabetical order
arrange(desc(score), evalue, profile)
```

For exactly the multi-model file this PR is about, `profile` is constant across models, so the documented tie-break silently degrades to input order. Same reasoning as the rest of the change: the file name is not the model identity. `arrange(desc(score), evalue, profile, profile_desc)` would restore it.

**`meta.yml:70` — the output description is now stale.** It still says "one row per (accno, profile) pair"; after this PR it is one row per (accno, profile, profile_desc). Worth fixing in the same breath, since `profile_desc` becomes load-bearing here and its name does not advertise that it holds the query model name — downstream consumers now have to key on it.

**One limitation, non-blocking and possibly not worth acting on:** two models sharing a `NAME` within one HMM file still collide on `(accno, profile, query)` and would reproduce both bugs. The query accession is the only remaining discriminator and it is usually `-`, so this may just be unresolvable from the tables alone. Noting it for the record rather than asking for a change.

<sub>Review prepared with Claude Code.</sub>

modules/nf-core/hmmer/hmmrank/tests/main.nf.test
```diff
@@ -75,20 +75,26 @@ nextflow_process {
...
-                { assert rows[1..-1].every { it.split('\t')[header.indexOf('hmm_len')].isInteger() } },
+                { assert path(process.out.hmmrank[0][1]).linesGzip.size() > 1 },
+                { def rows = path(process.out.hmmrank[0][1]).linesGzip
+                  def header = rows[0].split('\t') as List
+                  // NA is a legitimate value: hmmsearch reports a sequence on the per-sequence
+                  // threshold, so a hit can be ranked with no domain records behind it at all
+                  assert rows[1..-1].every {
+                      def v = it.split('\t')[header.indexOf('hmm_len')]
+                      v.isInteger() || v == 'NA'
+                  } },
```
piplus2: Allowing `NA` is right, but `every { isInteger() || == 'NA' }` is also satisfied when *every* row is `NA` — which is precisely what a broken join looks like. The assertion no longer fails on the regression it is closest to.

The snapshot in the same test still guards the actual values, so this is not a hole so much as an assertion that stopped carrying its own weight. Pairing it with an `any` would restore that without losing the legitimate `NA`:

```groovy
assert rows[1..-1].any { it.split('\t')[header.indexOf('hmm_len')].isInteger() }
```

modules/nf-core/hmmer/hmmrank/tests/main.nf.test
```diff
@@ -142,16 +148,79 @@ nextflow_process {
...
+        when {
+            process {
+                """
+                // gunzip, concatenate and re-gzip, rather than carrying a near-duplicate fixture
+                def combine = { name, srcs ->
+                    def out = file("\${workDir}/multimodel/\${name}")
+                    out.parent.mkdirs()
+                    def os = new java.util.zip.GZIPOutputStream(out.newOutputStream())
+                    srcs.each { src ->
+                        os.write(new java.util.zip.GZIPInputStream(file(src, checkIfExists: true).newInputStream()).bytes)
```
piplus2: Nit: the `GZIPInputStream`s are never closed — only `os` is. Four handles in a test, so harmless, but `withStream`/`withCloseable` would tidy it:

```groovy
srcs.each { src ->
    file(src, checkIfExists: true).newInputStream().withStream { is ->
        os.write(new java.util.zip.GZIPInputStream(is).bytes)
    }
}
```

erikrikarddaniel:
Thanks — all four taken, in 17c03248f.

**Rank tie-break.** You're right, and it's the same root cause: `profile` is the file name, so for a multi-model file it's constant across the models and the documented alphabetical tie-break degrades to input order. Worth stressing that this is a determinism problem with teeth — `rank == 1` is what `fasta_hmmsearch_rank_fastas` selects on for `SEQTK_SUBSEQ`, so arbitrary rank ordering means arbitrary sequence selection downstream. Now `arrange(desc(score), evalue, profile, profile_desc)`. It's a strict refinement: it only takes effect where score, evalue *and* profile are all equal, which cannot happen across separate files, so single-model input is untouched and no snapshot moved.

**`meta.yml` output description.** Fixed — one row per `(accno, profile, profile_desc)`, and it now says outright that `profile_desc` holds the query name, i.e. the model. Agreed the name doesn't advertise it. I've left the name alone deliberately: it's part of the output contract and nf-core/ampliseq vendors this module, so renaming it would break consumers for a cosmetic gain. Documenting it seemed the better trade.

**The `every { isInteger() || == 'NA' }` hole.** Good catch, and a fair characterisation — it stopped carrying its own weight rather than being a hole as such. It now asserts `every` *and* `any { isInteger() }`, so all-NA fails.

**Unclosed `GZIPInputStream`s.** Fixed, with `withStream` on both directions so the output stream is closed on an exception path too.

**Two models sharing a `NAME` in one file** — agreed, and agreed it's not actionable here. The query accession is the only other discriminator and is `-` in practice, so the tables genuinely don't carry enough to tell them apart. HMMER doesn't enforce unique names, so it's reachable in principle, but a file like that is ambiguous to anything reading the output too, not just to this module. Left as-is; your note is the record.

<sub>Prepared with help from Claude Code.</sub>

erikrikarddaniel:
Good points @piplus2 , I think we've addressed them if you want to have another look, thanks!

piplus2 (approved):
LGTM

erikrikarddaniel:
Thanks @piplus2 , I owe you some now since these were a bit larger than initially planned. Ping me if needed.

PR #12796, opened 2026-08-25
feat(comebin/runcomebin): add GPU support, fix outputs and add a real test

dialvarezs:
## GPU support

Container and conda environment are selected from `task.accelerator`, and the resolved device is passed to `run_comebin.sh` with `-d`. The CUDA runtime version goes to the versions topic.

The GPU environment pins `pytorch-gpu=2.13.0` with `cuda-version=12.9`, the lowest CUDA that pytorch 2.13 ships on linux-64.

## Fixes

- The declared `embeddings.tsv` and `covembeddings.tsv` outputs do not exist. COMEBin 1.1.0 writes `embeddings.npy` and `covembeddings.npy`, plus `embedding_ids.txt` giving the row order, which is now emitted too.

- Passing `-d` is also needed on the CPU path since 1.1.0. `run_comebin.sh` defaults to `device=cuda`, and 1.1.0 turned that into a hard error rather than the silent CPU fallback 1.0.4 did, so a real run on the current container raises before training:

	```
	RuntimeError: CUDA training was requested, but the installed PyTorch is CPU-only
	```

- `find ${prefix}/comebin_res_bins/*.fa -exec gzip {} \;` fails under Singularity (no `find` in the image).
	The glob already does the work, so this is now a plain `gzip`.

## Tests

The stub-only test is replaced by a real one over a two-genome community, plus a GPU test tagged `gpu`.

Bin contents are not reproducible across machines. Pinning `-s 1` and `cpus` fixes COMEBin's own threading, but not the BLAS and OpenMP pools underneath, which follow the host's core count. So the snapshot keeps the stable outputs and the bin count is checked by assertion, as `semibin/multieasybin` already does for the same reason.

## PR checklist

- [x] Add a resource `label`
- Ensure that the test works with either Docker / Singularity:
  - [x] `nf-core modules test comebin/runcomebin --profile singularity`
  - [ ] `nf-core modules test comebin/runcomebin --profile docker`
  - [ ] `nf-core modules test comebin/runcomebin --profile conda`


modules/nf-core/comebin/runcomebin/main.nf
```diff
@@ -32,14 +39,15 @@ process COMEBIN_RUNCOMEBIN {
...
+        -d ${device} \\
         -a local_assembly.fasta \\
         -p bam/ \\
         -o . \\
         ${args}
 
     mv comebin_res ${prefix}
 
-    find ${prefix}/comebin_res_bins/*.fa -exec gzip {} \\;
+    gzip ${prefix}/comebin_res_bins/*.fa
```
prototaxites, line 51: This fails if no bins are produced - that's why I wrote the find. Probably not a realistic happening in practice but I think I ran into it while developing the module originally.

prototaxites (commented):
One small comment, but looks really good - thanks for doing this!

modules/nf-core/comebin/runcomebin/main.nf
dialvarezs: It also fails with that expression:
image

But even if we change the expression to one not failing when there are no files, output channel is not optional, so it will fail anyway.

modules/nf-core/comebin/runcomebin/tests/main.gpu.nf.test
```diff
@@ -0,0 +1,45 @@
...
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert process.out.bins.get(0).get(1).size() >= 1 },
+                { assert snapshot(
+                    sanitizeOutput(process.out,
+                        // bin contents vary between machines, see the BLAS and OpenMP note in the PR
```
dialvarezs:
```suggestion
                        // bin contents vary between machines, due to BLAS and OpenMP
```

modules/nf-core/comebin/runcomebin/tests/main.nf.test
```diff
@@ -9,12 +9,43 @@ nextflow_process {
...
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert process.out.bins.get(0).get(1).size() >= 1 },
+                { assert snapshot(
+                    sanitizeOutput(process.out,
+                        // bin contents vary between machines, see the BLAS and OpenMP note in the PR
```
dialvarezs:
```suggestion
                        // bin contents vary between machines due to BLAS and OpenMP
```

prototaxites (commented):
OK, it looks good to me but I will leave approval to someone else with some knowledge of GPUs and their nf-core implementation

modules/nf-core/comebin/runcomebin/main.nf
```diff
@@ -2,10 +2,14 @@ process COMEBIN_RUNCOMEBIN {
     tag "${meta.id}"
     label 'process_high'
```
mahesh-panchal, line 3: Does this need `label 'process_gpu'`? That would enable the gpu when someone uses the `gpu` profile. As it currently is, it would need to be enabled by param in the pipeline.

mahesh-panchal (approved):
Aside from the label comment 👍🏽

modules/nf-core/comebin/runcomebin/main.nf
dialvarezs: Good catch. I'll fix it.

dialvarezs: Done, thanks!

PR #12798, opened 2026-08-26
fix(seqkit/grep): stub emits the same filename as a real run

erikrikarddaniel:
## PR checklist

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

## Description

`seqkit/grep`'s stub does not write the filename the process declares.

It builds `suffix = output_type + compression_suffix` and then appends another `.gz`:

```groovy
def compression_suffix = sequence.getExtension() == "gz" ? ".gz" : ""
def suffix = output_type + compression_suffix
"""
echo "" | gzip > ${prefix}.${suffix}.gz
"""
```

so the two cases come out as:

| input | real run writes | stub writes |
| --- | --- | --- |
| `genome.fasta.gz` | `test.fa.gz` | `test.fa.gz.gz` |
| `genome.fasta` | `test.fa` | `test.fa.gz` |

`test.fa.gz.gz` does not match the declared `path("*.{fa,fq,fa.gz,fq.gz}")`, so any `-stub` run with gzipped input fails with a missing output while the task exits 0 -- which makes it look like a pipeline problem rather than a stub problem.

The ungzipped case happened to match the glob, and the existing `with_file - stub` test uses `genome.fasta`, so the gzipped path was never covered.

### Changes

- The stub writes `${prefix}.${suffix}`, gzipped only when the input is gzipped.
- New `with_file - gzipped input - stub` test. On master it fails with `cmd: echo echo "" | gzip > test.fa.gz.gz`; with this change it passes.
- The existing stub snapshot changes from `test.fa.gz` to `test.fa`, which is what a real run produces for that input.

Ran `nf-test test modules/nf-core/seqkit/grep/tests/main.nf.test --profile docker` -- 5 tests pass. Docker only; I have not run the singularity or conda profiles.

Found in nf-core/metatdenovo, where `seqkit/grep` selects protein-cluster representatives from a gzipped FASTA and this blocked a stub run of the whole pipeline.

Fix drafted with help from Claude Code.

### DIFF
--- a/modules/nf-core/seqkit/grep/main.nf
+++ b/modules/nf-core/seqkit/grep/main.nf
@@ -48,6 +48,10 @@ process SEQKIT_GREP {
     """
     echo ${args}
 
-    echo "" | gzip > ${prefix}.${suffix}.gz
+    if [ "${compression_suffix}" = ".gz" ]; then
+        echo "" | gzip > ${prefix}.${suffix}
+    else
+        touch ${prefix}.${suffix}
+    fi
     """
 }
--- a/modules/nf-core/seqkit/grep/tests/main.nf.test
+++ b/modules/nf-core/seqkit/grep/tests/main.nf.test
@@ -111,4 +111,31 @@ nextflow_process {
             )
         }
     }
+
+    test("with_file - gzipped input - stub") {
+        options '-stub'
+
+        when {
+            params {
+                modules_args = ''
+            }
+            process {
+                """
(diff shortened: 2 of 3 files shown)


mahesh-panchal (approved):

Kevin-Brockers (approved):
Looks good!

erikrikarddaniel:
Thanks @mahesh-panchal & @Kevin-Brockers

PR #12800, opened 2026-08-26
fix(seqtk/subseq): let ext.prefix name the output file

### LINKED ISSUE #12779: seqtk/subseq: output filename concatenates input name and ext.prefix with no separator

### Description of the bug

`seqtk/subseq` builds its output filename by concatenating the input filename and `ext.prefix` with no separator between them:

https://github.com/nf-core/modules/blob/master/modules/nf-core/seqtk/subseq/main.nf#L32

```groovy
gzip --no-name > ${sequences}${prefix}.${ext}.gz
```

`$sequences` is the input file's full name, extension included, and `prefix` defaults to `${meta.id}`, so with the default prefix the module produces names like:

```
domain_16s.fnaarc.fa.gz
domain_16s.fnabac16s.fa.gz
```

for `meta.id` values `arc` and `bac16s` searched against `domain_16s.fna`. The input extension is kept in the middle of the name and the id is glued straight onto it.

The module's own test hides this by setting a prefix that carries its own leading dot ([`tests/standard.config`](https://github.com/nf-core/modules/blob/master/modules/nf-core/seqtk/subseq/tests/standard.config)):

```groovy
ext.prefix = { ".filtered" }
```

which yields `genome.fasta.filtered.fa.gz`. So the implicit contract is "`ext.prefix` is a suffix appended to the input filename and must supply its own separator". That is the opposite of the usual nf-core meaning of `ext.prefix` (the output basename), it is not documented in `meta.yml`, and it makes the *default* value of `prefix` unusable: `${meta.id}` can never produce a well-formed name.

Found while working on nf-core/phyloplace, where `seqtk/subseq` is called from `fasta_hmmsearch_rank_fastas` with the default prefix.

### Suggested fix

Follow the normal convention and let the prefix determine the basename:

```groovy
gzip --no-name > ${prefix}.${ext}.gz
```

with the same change in the `stub` block, and drop the leading dot from `tests/standard.config`.

This changes output filenames for existing callers, so it needs a snapshot update here and in any pipeline that consumes the module, hence an issue rather than a straight PR. If keeping the input filename in the output is considered worth preserving, `${sequences.baseName}.${prefix}.${ext}.gz` would be the compatible-in-spirit alternative, still a rename but at least a well-formed one.

Happy to open the PR once there is agreement on which of the two the module should produce.

### System information

- Nextflow 26.04.6
- Desktop, local executor
- Docker
- Ubuntu, Linux 6.8
- `quay.io/biocontainers/seqtk:1.4--he4a0461_1`

erikrikarddaniel:
## PR checklist

Closes #12779

- [ ] Add a resource `label`
  - For modules:
  - For subworkflows:

## Description

Fixes #12779: `seqtk/subseq` built its output filename by concatenating the input filename and `ext.prefix` with no separator.

```groovy
gzip --no-name > ${sequences}${prefix}.${ext}.gz
```

`$sequences` is the input file's full name, extension included, and `prefix` defaults to `${meta.id}`, so the default produced names like `domain_16s.fnaarc16s.fa.gz` -- the input extension stranded in the middle, the id glued straight onto it.
The module's own test hid this by setting a prefix that carried its own leading dot (`ext.prefix = { ".filtered" }`), which made the implicit contract "`ext.prefix` is a suffix and must supply its own separator": the opposite of what `ext.prefix` means everywhere else, undocumented in `meta.yml`, and it left the default value unable to produce a well-formed name at all.

This takes the first of the two options in the issue, the one that follows the usual convention.

### Changes

- `${prefix}.${ext}.gz` in both the `script` and the `stub` block.
- `tests/standard.config` sets `ext.prefix = { "${meta.id}.filtered" }`, so the test exercises a prefix shaped the way callers actually write one, rather than one compensating for the bug.

### Snapshots

Output files are renamed, so both consumers in this repository move:

| | before | after |
| --- | --- | --- |
| `seqtk/subseq` | `genome.fasta.filtered.fa.gz` | `test.filtered.fa.gz` |
| `fasta_hmmsearch_rank_fastas` | `domain_16s.fnaarc16s.fa.gz` | `arc16s.fa.gz` |

**Every md5 is unchanged**, so this is a pure rename with no change in content.
`fasta_hmmsearch_rank_fastas` calls the module with the default prefix, so its snapshot is also the coverage for the default path this fixes.

Downstream pipelines that consume `seqtk/subseq` will need a snapshot update for the same reason.

Ran `nf-test test modules/nf-core/seqtk/subseq --profile docker` (2 tests) and `nf-test test subworkflows/nf-core/fasta_hmmsearch_rank_fastas --profile docker` (3 tests), all passing; `nf-core modules lint seqtk/subseq` and `nf-core subworkflows lint fasta_hmmsearch_rank_fastas` clean apart from a pre-existing `has_meta_containers` warning. Docker only -- I have not run the singularity or conda profiles.

Fix drafted with help from Claude Code.

### DIFF
--- a/modules/nf-core/seqtk/subseq/main.nf
+++ b/modules/nf-core/seqtk/subseq/main.nf
@@ -31,7 +31,7 @@ process SEQTK_SUBSEQ {
         $args \\
         $sequences \\
         $filter_list | \\
-        gzip --no-name > ${sequences}${prefix}.${ext}.gz
+        gzip --no-name > ${prefix}.${ext}.gz
     """
 
     stub:
@@ -41,6 +41,6 @@ process SEQTK_SUBSEQ {
         ext = "fq"
     }
     """
-    echo "" | gzip > ${sequences}${prefix}.${ext}.gz
+    echo "" | gzip > ${prefix}.${ext}.gz
     """
 }
--- a/modules/nf-core/seqtk/subseq/tests/main.nf.test.snap
+++ b/modules/nf-core/seqtk/subseq/tests/main.nf.test.snap
@@ -7,7 +7,7 @@
                         {
                             "id": "test"
                         },
-                        "genome.fasta.filtered.fa.gz:md5,31c95c4d686526cf002f6119bc55b2b2"
+                        "test.filtered.fa.gz:md5,31c95c4d686526cf002f6119bc55b2b2"
                     ]
                 ],
                 "1": [
@@ -22,7 +22,7 @@
                         {
                             "id": "test"
                         },
-                        "genome.fasta.filtered.fa.gz:md5,31c95c4d686526cf002f6119bc55b2b2"
+                        "test.filtered.fa.gz:md5,31c95c4d686526cf002f6119bc55b2b2"
                     ]
                 ],
                 "versions_seqtk": [
@@ -34,11 +34,11 @@
                 ]
             }
         ],
+        "timestamp": "2026-08-26T09:12:10.98619525",
         "meta": {
-            "nf-test": "0.9.2",
-            "nextflow": "25.10.2"
-        },
-        "timestamp": "2026-01-14T10:29:32.149681094"
+            "nf-test": "0.9.5",
+            "nextflow": "26.04.6"
+        }
     },
     "sarscov2_subseq_fa_stub": {
         "content": [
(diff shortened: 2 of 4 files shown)


piplus2 (approved):
LGTM

erikrikarddaniel:
Thanks @piplus2

PR #12817, opened 2026-08-26
Add custom/resolvetaxonomy module

erikrikarddaniel:
## PR checklist

- [x] Add a resource `label`
  - For modules:
    - [x] `nf-core modules test custom/resolvetaxonomy --profile docker`
    - [ ] `nf-core modules test custom/resolvetaxonomy --profile singularity`
    - [ ] `nf-core modules test custom/resolvetaxonomy --profile conda`

## Description

New module resolving a set of sequences' taxonomy from either an explicit tab-separated file, or, when omitted, embedded GTDB-style taxonomy text in each FASTA header (`>id taxonomy;string`), matching GTDB's own single-file reference convention. Headers are always stripped down to a bare id in the output sequences, since some downstream tools keep the whole header line as the leaf/sequence name.

This generalises nf-core/sativa's existing local `resolvetaxonomy` module into a shared component. Two pipelines will consume it:

- **nf-core/sativa** already uses the local version; this PR's module is functionally identical for that use case (a required-taxonomy, single-alignment-file invocation).
- **nf-core/phyloplace** wants the same feature for its `--taxonomy`/`--refseqfile` pair, but its `--taxonomy` is fully optional -- no taxonomy at all is a legitimate, common case, so it must not fail the run.

To serve both, the module gains a `taxonomy_required` boolean input that wasn't in sativa's local version:

- `true` reproduces sativa's existing behaviour unchanged -- a record with neither an explicit-file entry nor embedded header text fails the run.
- `false` (phyloplace's case) drops such records from the resolved taxonomy output instead, with a warning, rather than failing.

Sequence format (FASTA, Clustal or PHYLIP) is sniffed from content, same as sativa's original module. Detecting whether a given input format has "room" for embedded taxonomy at all (FASTA does, PHYLIP/Clustal effectively don't) and gating whether to invoke this module is left to the calling pipeline, not handled inside the module.

### Test data

Fixtures added in nf-core/test-datasets (`modules` branch): erikrikarddaniel/test-datasets#TBD (branch `add-resolvetaxonomy-fixtures`, PR pending) -- five small hand-authored files under `data/generic/fasta/` and `data/generic/tax/` covering all three taxonomy sources plus the `taxonomy_required` true/false split. This PR's own test file references them via `params.modules_testdata_base_path`; tests were run locally against the fork branch before those fixtures were merged and passed twice independently (not just once with `--update-snapshot`).

### Notes

- `nf-core modules lint custom/resolvetaxonomy` is clean apart from one `bioconda_version` false positive on `biopython=1.84` -- the Anaconda API query the linter uses currently returns a stale/truncated version list for the `bioconda::biopython` package (tops out at `1.70`), unrelated to this module; the same pin is already in production use in nf-core/sativa's local module.
- Docker profile only -- I have not run the singularity or conda profiles.

Drafted with help from Claude Code.


modules/nf-core/custom/resolvetaxonomy/main.nf
```diff
@@ -0,0 +1,102 @@
...
+    script:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    // Nextflow stages an absent optional path(taxonomy) as an empty list -- falsy in
+    // Groovy -- rather than as a file, so this correctly distinguishes "no taxonomy
+    // file given" from a real one, without ever interpolating the literal text "[]"
+    // into the command line.
+    def taxonomy_in = taxonomy ? "${taxonomy}" : ''
+    """
+    python3 - "${taxonomy_in}" "${sequences}" "${taxonomy_required}" "${prefix}.resolved.tax" "${prefix}.resolved.${sequences.extension}" "${prefix}.warnings.txt" << 'PYEOF'
+import sys
```
leapicard, line 31: If I have one suggestion, it would be to actually put the python script in a resources/usr/bin folder to make main.nf cleaner

leapicard:
As far as I see, everything looks good, but it's my first time reviewing so maybe someone else should take a look?

leapicard (approved):

modules/nf-core/custom/resolvetaxonomy/main.nf
erikrikarddaniel: It is quite long, which would argue for that. Personally, I find it good to have the code here as it then ends up directly in `.command.sh` which I think makes it easier to debug by stepping through. But quite a personal opinion.

Thanks for the review!

PR #12824, opened 2026-08-26
Fix custom/resolvetaxonomy: malformed output filename for extensionless input

erikrikarddaniel:
## PR checklist

  - For modules:
    - [x] `nf-core modules test custom/resolvetaxonomy --profile docker`
    - [ ] `nf-core modules test custom/resolvetaxonomy --profile singularity`
    - [ ] `nf-core modules test custom/resolvetaxonomy --profile conda`

## Description

Found while wiring `custom/resolvetaxonomy` (merged in #12817) into nf-core/phyloplace: its `sequences` output is named `"*.resolved.${sequences.extension}"`, but `Path.extension` is `''` (not null) when the input file has no extension at all -- e.g. a bare download URL whose path is just a numeric file id (`.../files/47244067`, no filename suffix). That produced a malformed `"${prefix}.resolved."` output name, matching nothing, so the process silently failed to produce a usable `sequences` output.

Fix: fall back to `'fasta'` via `sequences.extension ?: 'fasta'`.

One thing worth flagging for other module authors: I first tried factoring this into a single `def ext = sequences.extension ?: 'fasta'` in the `script:` block and reusing `${ext}` in the `output:` declaration (the same pattern `def prefix = task.ext.prefix ?: ...` uses everywhere) -- that's a compile-time error, `` `ext` is not defined ``. `output:` glob patterns can only reference input-bound variables (`sequences`, `meta`, etc.) directly, not `script:`-scoped `def` locals. Confirmed this empirically rather than assuming from the `prefix` precedent, since `prefix` itself is only ever used in `script:`/`stub:`, never in an `output:` pattern, in this module or apparently elsewhere. So the fallback is duplicated inline in `script:`, `stub:` and the `output:` glob itself.

Added a test case with a synthesized extension-less input file (no new fixture needed -- built inline in the test from the existing `resolvetaxonomy_embedded.fasta` fixture) covering the regression.

Ran `nf-test test modules/nf-core/custom/resolvetaxonomy --profile docker` (8/8 passing, including the new case) twice to confirm stability, and `nf-core modules lint custom/resolvetaxonomy` (clean). Docker only -- I have not run the singularity or conda profiles.

Generated by Claude

### DIFF
--- a/modules/nf-core/custom/resolvetaxonomy/main.nf
+++ b/modules/nf-core/custom/resolvetaxonomy/main.nf
@@ -11,9 +11,12 @@ process CUSTOM_RESOLVETAXONOMY {
     tuple val(meta), path(taxonomy), path(sequences), val(taxonomy_required)
 
     output:
-    tuple val(meta), path("*.resolved.tax"),                    emit: taxonomy
-    tuple val(meta), path("*.resolved.${sequences.extension}"), emit: sequences
-    tuple val(meta), path("*.warnings.txt"),                    emit: warnings
+    // The fallback to 'fasta' matters when sequences has no extension at all (e.g. a
+    // download URL with no filename suffix) -- Path.extension is '' there, which
+    // would otherwise produce a malformed "*.resolved." glob matching nothing.
+    tuple val(meta), path("*.resolved.tax"),                                emit: taxonomy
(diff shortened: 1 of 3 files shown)


piplus2 (approved):
LGTM

PR #12829, opened 2026-08-27
transdecoder/predict: fix module never resuming and failing on re-execution

### LINKED ISSUE #12799: transdecoder/predict cannot be resumed: it writes checkpoints into its staged input, then fails on re-execution
## Summary

`TransDecoder.Predict` writes its checkpoint directory into the LongOrfs directory it receives as a staged input.
That directory belongs to the `TRANSDECODER_LONGORF` task, so two things follow, and both are reproducible from a clean run:

1. **The task never resumes.** Because it modifies its own input, a following `-resume` re-executes it even when nothing has changed.
2. **The re-execution then fails.** Nextflow wipes the task directory before re-running, but not the LongOrfs directory the symlink points at, so the checkpoints outlive the outputs. `TransDecoder.Predict` sees a complete checkpoint set, skips every step including the ones that create the outputs, and exits **0** having written nothing.

The failure surfaces as `Missing output file(s) '*.transdecoder.pep'` with `Command exit status: 0`, which points at the module rather than at the checkpoints, and it is permanent: every retry repeats it.

## Reproduction

With nf-core/metatdenovo's smallest test profile, about two minutes per run:

```bash
nextflow run nf-core/metatdenovo -profile test,docker --outdir out --orf_caller transdecoder
# [SUCCESS] completed=63 failed=0 cached=0

nextflow run nf-core/metatdenovo -profile test,docker --outdir out --orf_caller transdecoder -resume
# TRANSDECODER_LONGORF   CACHED
# TRANSDECODER_PREDICT   FAILED
# [FAILED] completed=1 failed=1 cached=39
```

Nothing changed between the two runs.
`TRANSDECODER_LONGORF` is cached as expected; `TRANSDECODER_PREDICT` is not, and its re-execution fails.

From the failing task's `.command.err`, every step is skipped, including the final copies:

```
-- Skipping CMD: cp .../longest_orfs.cds.best_candidates.gff3.revised_starts.gff3
                    .../megahit_assembly.contigs.fa.transdecoder.gff3,
   checkpoint [.../megahit_assembly.contigs.fa.transdecoder_dir/__checkpoints_TDpredict/copy_to_final_outputfile.974da7....ok] exists.
transdecoder is finished.
```

and in the task directory the checkpoints resolve outside it:

```
megahit_assembly.contigs.fa.transdecoder_dir ->
  .../work/6a/36d125.../megahit.transdecoder/megahit_assembly.contigs.fa.transdecoder_dir
```

Versions: `transdecoder 5.7.1--pl5321hdfd78af_0`, Nextflow 26.04.6, local executor with Docker.

## Why the module cannot just pass this through

This is how the tool is meant to work -- `TransDecoder.Predict`'s own help says `--output_dir | -O <string>  same output directory from the TransDecoder.LongOrfs`, so writing into the LongOrfs directory is by design.
What is specific to the module is that the directory is a *staged input* rather than a working copy, so the tool's writes land in another task's directory and Nextflow's caching sees an input that changes after use.

## Possible fixes

In rough order of preference:

1. **Give Predict a private, writable view of the LongOrfs directory.**
   Building a directory of symlinks to each file inside the staged one costs no extra disk: reads follow the symlinks, and everything `Predict` creates -- the checkpoint directory and its intermediates -- is local to the task.
   Worth verifying first that `Predict` only ever *adds* files there and never overwrites a LongOrfs output.

2. **Copy the directory** (`cp -rL`) into the task directory before running.
   Simplest and certainly correct, but it duplicates the LongOrfs output, which is large for metagenome assemblies -- `longest_orfs.cds` alone can reach several GB.

3. **At minimum, remove a stale `__checkpoints_TDpredict` before running**, so a re-executed task redoes the work instead of failing.
   One line, and it fixes the hard failure, but it writes into another task's directory and leaves the self-invalidation in place, so it is a fallback rather than a fix.

Independently of which is chosen, the module could check that its declared outputs exist and fail with a message naming the checkpoints, instead of exiting 0 and leaving Nextflow to report a missing file.

## Impact

Any interrupted run, any `errorStrategy 'retry'`, and any `-resume` can hit this.
With a retry policy it converts a transient failure into a permanent one, and on large assemblies the wasted work is substantial: LongOrfs plus Predict over a multi-sample metagenome assembly is hours.

---

Reported with help from Claude Code.

erikrikarddaniel:
## PR checklist

Closes #12799

- [x] Add a resource `label`
  - For modules:

## Description

`TransDecoder.Predict` writes its checkpoint directory into the LongOrfs directory it receives via `-O`.
That directory is staged as a symlink into `TRANSDECODER_LONGORF`'s own work directory, so writing through it there mutates another task's output after the fact.

Two consequences, both reproducible from a clean run and both explained in #12799 with a full trace:

1. **Predict itself never resumes.** Because it modifies its own staged input, a following `-resume` sees a changed input and re-executes it even when nothing upstream changed.
2. **The re-execution then fails.** Nextflow wipes the Predict task directory before re-running, but not the LongOrfs directory the symlink points at, so the stale `__checkpoints_TDpredict` from the first run survives. `TransDecoder.Predict` sees a complete checkpoint set, skips every step -- including the ones that write `*.transdecoder.pep`/`.gff3`/`.cds`/`.bed` -- and exits **0** with nothing produced. The failure then surfaces as a generic "missing output file" error that points at the module rather than at the real cause.

### Fix

Give Predict a private, writable directory of symlinks to the same files, instead of writing through the staged symlink directly:

```bash
real_fold=$(readlink -f $fold)
rm $fold
mkdir $fold
ln -s "$real_fold"/* $fold/
```

Reads still resolve through the per-file symlinks to the original LongOrfs outputs. Every write Predict makes -- the checkpoint directory and its intermediates -- lands in this new, task-local directory instead, so the LongOrfs task's own output is never touched and Predict's declared input stays byte-identical across runs (making it correctly cacheable), while a genuine re-execution always starts from a fresh directory with no stale checkpoints to short-circuit it.

### Testing

- All three existing `transdecoder/predict` module tests (`sarscov2 - uncompressed`, `sarscov2 - compressed`, `sarscov2 - uncompressed - stub`) pass unmodified against the pre-existing committed snapshot -- confirming output content is unchanged, not just plausibly so.
- Reproduced the original bug end-to-end and verified the fix against a full pipeline, not just the module in isolation: ran nf-core/metatdenovo's `test,docker` profile with `--orf_caller transdecoder`, then `-resume` with no changes.
  - **Before this fix:** `TRANSDECODER_PREDICT` fails on resume with `Missing output file(s) '*.transdecoder.pep'`, `Command exit status: 0` -- exactly as described in #12799.
  - **After this fix:** the trace report shows `TRANSDECODER:TRANSDECODER_PREDICT (megahit.transdecoder)` as `CACHED`, and the whole pipeline resumes cleanly (`completed=1 failed=0 cached=76`, only MULTIQC re-running as expected).
- `nf-core modules lint transdecoder/predict`: clean except one pre-existing, unrelated warning (bioconda `transdecoder` 5.7.1 -> 6.0.0 available) that this PR doesn't touch.
- `prek run` on the changed file: clean.

Reported and fixed with help from Claude Code.

### DIFF
--- a/modules/nf-core/transdecoder/predict/main.nf
+++ b/modules/nf-core/transdecoder/predict/main.nf
@@ -24,6 +24,15 @@ process TRANSDECODER_PREDICT {
     script:
     def args = task.ext.args ?: ''
     """
+    # \$fold is staged as a symlink into TRANSDECODER_LONGORF's own work directory.
+    # TransDecoder.Predict writes its checkpoints and intermediates there, which mutates
+    # that other task's output and makes this task uncacheable across -resume
+    # (nf-core/modules#12799). Give it a private, writable directory of symlinks instead.
+    real_fold=\$(readlink -f $fold)
+    rm $fold
+    mkdir $fold
+    ln -s "\$real_fold"/* $fold/
+
     TransDecoder.Predict \\
         $args \\
         -O . \\


Joon-Klaps (approved):
Im going to trust your explanation on this. 

Would think that `rm` would actually interfere with the caching.

erikrikarddaniel:
@Joon-Klaps Reasonable worry, but it goes the other way around — the `rm` is what *fixes* caching rather than interfering with it.

Nextflow computes a task's cache key from its declared inputs before the script ever runs. For `$fold`, that key is derived from the *source* directory still living in `TRANSDECODER_LONGORF`'s own work directory — not from anything inside Predict's task directory.

`rm $fold` only removes the **local staged symlink** sitting in Predict's own ephemeral task directory; it never touches the LongOrfs directory itself (removing a symlink never touches its target). That's the whole point: previously, writes went *through* that symlink into the shared source, mutating it and changing its hash between runs, which is what broke caching in the first place. Now nothing ever writes to the source again, so its hash stays stable across runs, and on an actual resumed (`CACHED`) execution the script body — including the `rm`/`mkdir`/`ln -s` — never runs at all; Nextflow just reuses the prior outputs.

Verified directly: with this fix, `TRANSDECODER_PREDICT` shows `CACHED` in the trace report on `-resume`, and `TRANSDECODER_LONGORF` (whose directory we now leave untouched) stays cached too.

PR #12831, opened 2026-08-27
transdecoder/predict: fix cloud/copy-staging failure (follow-up to #12829)

erikrikarddaniel:
## PR checklist

- [x] Add a resource `label`
  - For modules:

## Description

Follow-up to #12829, flagged in review on the downstream nf-core/metatdenovo consumer PR ([nf-core/metatdenovo#480](https://github.com/nf-core/metatdenovo/pull/480), review by @piplus2 -- thanks for catching this).

#12829 fixed `TRANSDECODER_PREDICT`'s `-resume` failure by giving `Predict` a private, writable directory of symlinks instead of writing through the staged LongOrfs directory. That fix assumed the staged input (`$fold`) is always a symlink, which only holds for local/shared-filesystem staging.

Under `stageInMode 'copy'` -- the default for cloud storage (AWS/GCS/Azure) without Fusion, or whenever a user/institutional config sets `process.stageInMode` explicitly -- Nextflow stages directory inputs as real copies, not symlinks. `rm` on a real directory fails outright, and any pipeline with strict bash flags (`set -e`, which nf-core/metatdenovo among others sets) aborts the task right there. That turns a `-resume` fix into a hard failure on every cloud-executed run using this module.

### Fix

Replace `rm`+`mkdir` with `mv`, which handles both staging modes identically: it moves a symlink as a symlink, and renames a real directory in place -- so nothing is deleted before the symlink farm is built from it, regardless of staging mode.

Also add `shopt -s dotglob nullglob` before the glob-based `ln -s`: without it, a dot-prefixed file in the LongOrfs directory would be silently skipped, and an empty match would create a literal dangling `*` symlink rather than failing loudly. Latent today since TransDecoder 5.7.1 writes no dotfiles there, but worth guarding against a future version that does (the module's own bioconda-update warning already flags 5.7.1 -> 6.0.0 as available).

### Testing

- The module's own 3 tests (`sarscov2 - uncompressed`, `sarscov2 - compressed`, `sarscov2 - uncompressed - stub`) pass unmodified against the existing committed snapshot.
- Reproduced the actual failure mode the review flagged, not just reasoned about it: ran a full nf-core/metatdenovo pipeline (`-profile test,docker --orf_caller transdecoder`) with `-process.stageInMode copy` to simulate cloud staging, under that pipeline's own strict shell settings (`process.shell = ["bash", "-C", "-e", "-u", "-o", "pipefail"]`). The previous `rm`-based version aborts the task under copy staging; this version completes the full pipeline cleanly, including every task downstream of `Predict`.
- `nf-core modules lint transdecoder/predict` and `prek run`: clean (one pre-existing, unrelated bioconda-version warning, out of scope here).

Generated by Claude

### DIFF
--- a/modules/nf-core/transdecoder/predict/main.nf
+++ b/modules/nf-core/transdecoder/predict/main.nf
@@ -30,13 +30,13 @@ process TRANSDECODER_PREDICT {
     # task's output and makes this task uncacheable across -resume (nf-core/modules#12799). Give it a
     # private, writable directory of symlinks instead. mv handles both staging modes: it moves a
     # symlink as a symlink, and renames a real directory in place, so nothing is deleted either way.
-    shopt -s dotglob nullglob
+    # find avoids the dotglob/nullglob shell-option juggling that caused three earlier bugs here.
     mv "$fold" "${fold}.staged"
     real_fold=\$(readlink -f "${fold}.staged")
+    [ -d "\$real_fold" ] || { echo "TRANSDECODER_LONGORF's directory is missing: \$real_fold" >&2; exit 1; }
     [ -n "\$(ls -A "\$real_fold")" ] || { echo "TRANSDECODER_LONGORF produced an empty directory: \$real_fold" >&2; exit 1; }
     mkdir "$fold"
-    ln -s "\$real_fold"/* "$fold"/
-    shopt -u dotglob nullglob
+    find "\$real_fold" -mindepth 1 -maxdepth 1 -exec ln -s {} "$fold"/ \\;
 
     TransDecoder.Predict \\
         $args \\
--- a/modules/nf-core/transdecoder/predict/main.nf
+++ b/modules/nf-core/transdecoder/predict/main.nf
@@ -31,10 +31,12 @@ process TRANSDECODER_PREDICT {
     # private, writable directory of symlinks instead. mv handles both staging modes: it moves a
     # symlink as a symlink, and renames a real directory in place, so nothing is deleted either way.
     shopt -s dotglob nullglob
-    mv $fold ${fold}.staged
-    real_fold=\$(readlink -f ${fold}.staged)
-    mkdir $fold
-    ln -s "\$real_fold"/* $fold/
+    mv "$fold" "${fold}.staged"
+    real_fold=\$(readlink -f "${fold}.staged")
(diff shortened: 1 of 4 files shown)


piplus2:
Three minor findings on the new lines:

**1. `shopt -s dotglob nullglob` is never unset (line 33)**

It stays in effect for the rest of the script, including the `TransDecoder.Predict $args` expansion. Since `$args` is interpolated unquoted, any glob-like token in `ext.args` that matches nothing is silently deleted from the command line. With `ext.args = '-T nosuchfile.*'` the command expands to:

```
TransDecoder.Predict -T -O . -t genome.fasta
```

so `-T` swallows `-O` and TransDecoder runs with a corrupted argument list instead of failing on the missing file. Suggest scoping it:

```bash
( shopt -s dotglob nullglob; ln -s "$real_fold"/* $fold/ )
```

or `shopt -u dotglob nullglob` immediately after the `ln`.

**2. The empty-directory case fails with a misleading error (line 37)**

If `$real_fold` is empty, nullglob makes the glob vanish and `ln -s "$real_fold"/* $fold/` collapses to a single-operand `ln`, which aborts with:

```
ln: failed to create symbolic link './genome.fasta.transdecoder_dir': File exists
```

That points at a name collision rather than an empty upstream output. It also only fails at all because `mkdir $fold` happens to run first — reorder those two lines and the same case would silently create a stray symlink in the CWD. If a real guard is intended:

```bash
[ -n "$(ls -A "$real_fold")" ] || { echo "empty LongOrfs dir: $real_fold" >&2; exit 1; }
```

**3. Unquoted `$fold` operands (line 34 onward)**

`mv $fold ${fold}.staged` leaves both operands unquoted, as did the `rm $fold` it replaces — so inherited style rather than a regression. But with `nullglob` now enabled there's a new failure mode: a FASTA basename containing `*` or `?` that matches nothing drops the operand entirely, and one containing whitespace makes `mv` receive three-plus operands. Quoting `"$fold"` / `"${fold}.staged"` on the `mv`, `mkdir`, `readlink` and `ln` lines closes the whole class.

---

Checked and cleared: `mv` on a relative symlink under `stageInMode 'rellink'`; the `link` → copy fallback for directories; `${fold}.staged` colliding with the `*.transdecoder.{pep,gff3,cds,bed}` output globs; and `dotglob` pulling in `.`/`..`.

piplus2 (approved):
LGTM

erikrikarddaniel:
Thanks @piplus2

PR #12836, opened 2026-08-27
correct process label

leapicard:
small process label update to better fit the resources needed for the average analysis


erikrikarddaniel (approved):
Reviewed with some help from Claude Code. Label bump looks reasonable — BUSTED does multiple branch-site optimizations per run, checked the other hyphy siblings and the label range there already spans low/medium/high (e.g. `gard` is `process_high`), so this isn't out of line.

PR #12841, opened 2026-08-28
Add pcaone module

dbaku42:
Adds the pcaone module for [[PCAone](https://github.com/Zilong-Li/PCAone)](https://github.com/Zilong-Li/PCAone), a fast PCA tool for large-scale genotype data. Takes PLINK bed/bim/fam, outputs .eigvecs/.eigvals (plus .loadings/.mbim when --printv is set via ext.args).
Test note: the small plink_simulated test dataset crashes PCAone's default winSVD algorithm (floating point exception), so tests/nextflow.config pins --svd 0 (IRAM) and cpus=1 for the test only — production default (--svd 2) is untouched. The log output is excluded from the snapshot since it embeds a non-reproducible container hostname/timestamp.


modules/nf-core/pcaone/meta.yml
```diff
@@ -0,0 +1,148 @@
...
+          ontologies: []
+  eigvecs2:
+    - - meta:
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. `[ id:'sample1' ]`
+      - "*.eigvecs2":
+          type: file
+          description: Eigenvectors with a header line, produced when requested via ext.args
```
erikrikarddaniel: PCAone's own README lists `.eigvecs2` ("eigenvectors with header line") among the files produced by a plain example run with no `--printv`/other extra flags -- it looks like it's written unconditionally alongside `.eigvecs`, not gated behind anything passed via `ext.args`. `.loadings` and `.mbim` do look correctly gated behind `--printv` per the docs, so this seems specific to `eigvecs2`.

If that's right, two small things worth double-checking (not blocking, since `optional: true` doesn't break anything even if the file is always present):
- this description (and the equivalent one in `main.nf`'s output line) might want to drop "produced when requested via ext.args"
- none of `eigvecs2`/`loadings`/`mbim` are exercised by the current tests (only `eigvecs`/`eigvals`/versions are snapshotted) -- might be worth a quick assertion that `eigvecs2` shows up on the default-args test, and/or a variant test with `--printv` for the other two.

erikrikarddaniel (approved):
See comment regarding optionality of certain files, otherwise great!

modules/nf-core/pcaone/meta.yml
dbaku42: Good catch, confirmed — the README's Quick Start example (./PCAone -b example/plink, no extra flags) lists eigvecs2 among the produced files, and the "Output files" section only gates .loadings (and .mbim, indirectly) behind --printv, with no equivalent note for eigvecs2. Removed optional: true and the misleading description, and added test coverage: the default test now also checks eigvecs2, and I added a --printv variant test asserting loadings/mbim. Thanks for flagging this

dbaku42: All tests pass. Ready to merge if there are no further objections, @erikrikarddaniel.

erikrikarddaniel: > All tests pass. Ready to merge if there are no further objections, @erikrikarddaniel.

I already approved, go ahead! :-)

PR #484, opened 2021-05-06
Add Nextclade module

drpatelh:
* Bumped Pangolin version to latest

PR #12846, opened 2026-08-29
New module: custom/collectfeaturecounts

erikrikarddaniel:
## PR checklist

- [x] Module has been tested locally with real Docker runs (not just lint), see below.
- [x] `nf-core modules lint` passes clean.
- [x] Test data is available (nf-core/test-datasets#2247, not yet merged, so CI here won't be green until it lands).

## Description

Collects one or more `subread/featurecounts` raw output files (typically one per sample, for the same feature type/ORF caller) into a single long-format table (`orf`, `chr`, `start`, `end`, `strand`, `length`, `sample`, `count`, `tpm`), computing per-sample TPM and dropping zero-count rows.

Ported from two independently-hardened, now-functionally-identical local modules — nf-core/magmap's and nf-core/metatdenovo's own `COLLECT_FEATURECOUNTS` — as part of a cross-repo effort to share their summary-table-generation code (tracked in [nf-core/magmap#237](https://github.com/nf-core/magmap/issues/237)). Both pipelines' local copies were first made caller/pipeline-agnostic by moving their one pipeline-specific ID-reconciliation special-case (magmap's genome-accession join, metatdenovo's Transdecoder `cds.`-prefix stripping) into small local post-processing steps — see nf-core/magmap#238 and nf-core/metatdenovo#483.

`tpm` is rounded to 6 decimal places. This carries over a fix from nf-core/metatdenovo#483/[#484](https://github.com/nf-core/metatdenovo/issues/484): when the same underlying counts get independently re-computed elsewhere (e.g. after a caller-consolidation step using a different R backend — `dtplyr`/`data.table` here vs plain `dplyr` there), the unrounded double can round its last significant digit differently for the same mathematical value, which made an exact-equality test downstream intermittently fail. Rounding here keeps output stable regardless of which backend a consumer uses to recompute the same value.

### Testing

Verified with real Docker runs against two new fixtures (two samples, mixed zero/non-zero counts, one shared multi-exon gene) — nf-core/test-datasets#2247. `tpm` correctly sums to exactly `1e6` per sample after rounding, zero-count rows are dropped, sample names are correctly stripped of their `.sorted.bam` suffix.

Generated by Claude

### DIFF
--- a/modules/nf-core/custom/collectfeaturecounts/main.nf
+++ b/modules/nf-core/custom/collectfeaturecounts/main.nf
@@ -1,0 +1,93 @@
+process CUSTOM_COLLECTFEATURECOUNTS {
+    tag "$meta.id"
+    label 'process_high'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine in ['singularity', 'apptainer'] && !task.ext.singularity_pull_docker_container ?
+        'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/14/143121bf17b3f4e9539d8cd17aaaa428aae54ce827471c4fc6399465263efa2e/data' :
+        'community.wave.seqera.io/library/r-base_r-dplyr_r-readr_r-purrr_pruned:0f879b99d6a89834' }"
+
+    input:
+    tuple val(meta), path(inputfiles)
+
+    output:
+    tuple val(meta), path("${prefix}.counts.tsv.gz"), emit: counts
+    path "versions.yml"                             , emit: versions, topic: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    #!/usr/bin/env Rscript
+
+    library(data.table)
+    library(dtplyr)
+    library(readr)
+    library(dplyr)
+    library(tidyr)
+    library(stringr)
+
+    setDTthreads($task.cpus)
+
+    tibble(f = Sys.glob('*.featureCounts.tsv')) %>%
+        mutate(
+            d = purrr::map(
+                f,
+                function(file) {
+                    fread(file, sep = '\\t', skip = 1) %>%
+                        melt(measure.vars = c(ncol(.)), variable.name = 'sample', value.name = 'count') %>%
+                        lazy_dt() %>%
+                        filter(count > 0) %>%
+                        mutate(
+                            sample = str_remove(sample, '.sorted.bam'),
+                            r = count/Length
+                        ) %>%
+                        rename(orf = Geneid, chr = Chr, start = Start, end = End, strand = Strand, length = Length) %>%
+                        group_by(sample) %>%
+                        # Rounded so tables independently computed elsewhere from the same underlying
+                        # counts (e.g. after a caller-consolidation step using a different R backend)
+                        # stay byte-comparable -- data.table/dtplyr and plain dplyr can otherwise round
+                        # the last significant digit of the same mathematical value differently.
+                        mutate(tpm = round(r/sum(r) * 1e6, 6)) %>% ungroup() %>%
+                        select(-r) %>%
+                        as_tibble()
+                }
+            )
+        ) %>%
+        tidyr::unnest(d) %>%
+        select(-f) %>%
+        write_tsv("${prefix}.counts.tsv.gz")
+
+    writeLines(
+        c(
+            "\\"${task.process}\\":",
+            paste0("    R: ", paste0(R.Version()[c("major","minor")], collapse = ".")),
+            paste0("    dplyr: ", packageVersion('dplyr')),
+            paste0("    readr: ", packageVersion('readr')),
+            paste0("    stringr: ", packageVersion('stringr')),
+            paste0("    dtplyr: ", packageVersion('dtplyr')),
+            paste0("    data.table: ", packageVersion('data.table'))
+        ),
+        "versions.yml"
+    )
+    """
+
+    stub:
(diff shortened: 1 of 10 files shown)


modules/nf-core/custom/collectfeaturecounts/main.nf
```diff
@@ -0,0 +1,93 @@
...
+    library(data.table)
+    library(dtplyr)
+    library(readr)
+    library(dplyr)
+    library(tidyr)
+    library(stringr)
+
+    setDTthreads($task.cpus)
+
+    tibble(f = Sys.glob('*.featureCounts.tsv')) %>%
```
Joon-Klaps: Don't think this should be the way to go. 

Should be defined as "$inputfiles" 

Also thinking if in general it wouldn't be better to have a Rscript executable stored in `templates/collectfeaturecounts.R`

Might also be a good example to use https://github.com/nf-core/r-nf-core-utils to facilitate maintanability.

modules/nf-core/custom/collectfeaturecounts/meta.yml
```diff
@@ -0,0 +1,93 @@
+name: "custom_collectfeaturecounts"
+description: |
+  Collects one or more subread/featureCounts raw output files (typically one per sample,
+  for the same feature type/ORF caller) into a single long-format table with per-sample
+  TPM computed.
```
Joon-Klaps:
```suggestion
  TPM (Transcripts Per Million)  computed.
```

modules/nf-core/custom/collectfeaturecounts/meta.yml
```diff
@@ -0,0 +1,93 @@
...
+keywords:
+  - featurecounts
+  - collect
+  - counts
+  - tpm
+tools:
+  - "collectfeaturecounts":
+      description: |
+        In-house R script that reshapes one or more raw featureCounts output files into a
+        single long-format table (one row per feature/sample), computing TPM per sample.
```
Joon-Klaps:
```suggestion
        single long-format table (one row per feature/sample), computing transcripts per million (TPMs) per sample.
```

Joon-Klaps (changes requested):
The R script should not use a globs search but a direct pointer to the input file.

modules/nf-core/custom/collectfeaturecounts/main.nf
erikrikarddaniel: Thanks for the push on this, @Joon-Klaps — this comment led to a bigger cleanup than the one-line fix it sounds like, and a special thanks is due since it surfaced how suboptimal the state of this code we inherited from our own local pipeline modules actually was.

What changed, in order:
1. Read files from the declared `path(inputfiles)` input instead of `Sys.glob('*.featureCounts.tsv')`.
2. Moved the R body into `templates/collectfeaturecounts.R` per your suggestion. Turned out a Groovy variable computed in the `script:` block isn't visible inside a `template` file (confirmed against `custom/matrixfilter`'s own comment about this), so ended up staging every input file into a fixed `input/` dir via `stageAs: "input/*"` (same pattern `nacho/qc` uses) and globbing that from R — no Groovy→template variable passing needed at all.
3. While re-reading the R itself: the per-file `fread → melt → lazy_dt() → dplyr chain → as_tibble()`, stitched back together with `tidyr::unnest()`, ran N independent small dtplyr query plans instead of one. Consolidated into `rbindlist()` + a single `lazy_dt()` pipeline over the combined data.
4. That consolidation removed the only uses of `tidyr`/`purrr`, so dropped both from `environment.yml` and regenerated the container + conda-lock files to match.

Each step verified with a real (non-stub) Docker `nf-test` run against the existing snapshot, plus a clean `nf-core modules lint`.

Still open: your `nf-core/r-nf-core-utils` suggestion — haven't looked into what it offers yet, will take a look separately rather than block this on it. Let me know if you think it should hold this PR up.

Joon-Klaps: Ok fair point, won't force you to use that R package.

modules/nf-core/custom/collectfeaturecounts/main.nf
```diff
@@ -0,0 +1,40 @@
...
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        R: 4.1.0
+        dplyr: 1.0.7
+        readr: 2.0.0
+        stringr: 1.4.0
+        dtplyr: 1.1.0
+        data.table: 1.14.0
+    END_VERSIONS
```
Joon-Klaps, line 38: Think this should be resolved and not hard coded.

Joon-Klaps (changes requested):
The stub versions, should be resolved not harcoded. 

The R nf-core package does allow you to do this easily

modules/nf-core/custom/collectfeaturecounts/main.nf
erikrikarddaniel: Fair, fixed — the stub's `versions.yml` was hardcoding stale literal numbers left over from the original local module (`R: 4.1.0` etc.), unrelated to what's actually pinned in `environment.yml`. Now resolves each value via `Rscript -e ...` at run time, same pattern `custom/matrixfilter`'s stub already uses for its `r-base` line.

Verified with a real Docker `nf-test` run — resolved values now match `environment.yml`'s pins exactly (R 4.5.3, dplyr 1.2.1, readr 2.2.0, stringr 1.6.0, dtplyr 1.3.3, data.table 1.17.8) — and updated the stub test's snapshot accordingly.

Joon-Klaps (approved):
LGTM!

erikrikarddaniel:
Thanks @Joon-Klaps

PR #12851, opened 2026-08-31
fix: snapaligner module lint and test improvements

maxulysse:
## Changes

- **snapaligner/align**: Use `${prefix}` in stub section instead of hardcoded `test.bam`, ensuring stub output respects `task.ext.prefix`
- **snapaligner/index**: Fix `meta.yml` output definition — changed `meta` type from `file` to `map` and `snap/*` to `snap` to match the emit name in `main.nf`
- **snapaligner/index**: Filter `OverflowTable` from non-stub test snapshot to avoid empty file md5 lint failure (`d41d8cd98f00b204e9800998ecf8427e`)
- Update test snapshots with `sanitizeOutput` and nf-test 0.9.5

## Lint results

- `snapaligner/align`: 62/62 passed
- `snapaligner/index`: 59/59 passed

## Tests

All 5 tests pass (3 align + 2 index).

Generated by opencode
Verified by @maxulysse

### DIFF
--- a/modules/nf-core/snapaligner/index/main.nf
+++ b/modules/nf-core/snapaligner/index/main.nf
@@ -1,17 +1,17 @@
 process SNAPALIGNER_INDEX {
-    tag "$fasta"
+    tag "${fasta}"
     label 'process_high'
 
     conda "${moduleDir}/environment.yml"
-    container "${ workflow.containerEngine in ['singularity', 'apptainer'] && !task.ext.singularity_pull_docker_container ?
-        'https://depot.galaxyproject.org/singularity/snap-aligner:2.0.5--h077b44d_2':
-        'quay.io/biocontainers/snap-aligner:2.0.5--h077b44d_2' }"
+    container "${workflow.containerEngine in ['singularity', 'apptainer'] && !task.ext.singularity_pull_docker_container
+        ? 'https://depot.galaxyproject.org/singularity/snap-aligner:2.0.5--h077b44d_2'
+        : 'quay.io/biocontainers/snap-aligner:2.0.5--h077b44d_2'}"
 
     input:
     tuple val(meta), path(fasta), path(altcontigfile), path(nonaltcontigfile), path(altliftoverfile)
 
     output:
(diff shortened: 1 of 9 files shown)


modules/nf-core/snapaligner/index/main.nf
```diff
@@ -1,17 +1,17 @@
 process SNAPALIGNER_INDEX {
-    tag "$fasta"
+    tag "${fasta}"
```
LouisLeNezet: Why not meta.id ?

modules/nf-core/snapaligner/index/main.nf
```diff
@@ -1,17 +1,17 @@
...
+        ? 'https://depot.galaxyproject.org/singularity/snap-aligner:2.0.5--h077b44d_2'
+        : 'quay.io/biocontainers/snap-aligner:2.0.5--h077b44d_2'}"
 
     input:
     tuple val(meta), path(fasta), path(altcontigfile), path(nonaltcontigfile), path(altliftoverfile)
 
     output:
-    tuple val(meta), path("snap/*") ,emit: index
+    tuple val(meta), path("snap"), emit: index
     tuple val("${task.process}"), val('snap-aligner'), eval("snap-aligner 2>&1 | sed 's/^.*version //;s/.\$//;q'"), topic: versions, emit: versions_snapaligner
```
LouisLeNezet, line 15: What does ';q' mean, just curious ?

LouisLeNezet (approved):
Just small comments, but good to go 👍

modules/nf-core/snapaligner/index/main.nf
maxulysse: ;q ?

maxulysse: oooh, in the versions, no idea, I did not noticed this to be honest

maxulysse: so whatever is before remove all trailing things, and q quit the command.
Might be redundant, but harmless according to opencode

PR #12863, opened 2026-09-01
feat(semibin): add GPU support to single_easy_bin and multi_easy_bin

dialvarezs:
Same treatment COMEBin got in #12796, now for SemiBin2.

Both `semibin/singleeasybin` and `semibin/multieasybin` pick their container and conda environment from `task.accelerator`, add the `process_gpu` label, and pass the resolved device to SemiBin2 with `--engine`. Both also emit the CUDA runtime version to the versions topic.

The `singleeasybin` test moves to the shared metagenome binning dataset, with all three BAMs. That meant dropping `--environment global`: the pretrained model only supports single-sample binning, and it skips the training the GPU is for. Training makes bin names and contents vary across machines, so the snapshot now keeps the stable outputs plus a bin count assertion, as `multieasybin` already does.

`multieasybin` moves off `delete_me/semibin2/` to `binning/multisample/`, which has the five samples SemiBin2 requires before it accepts `--abundance` instead of BAMs. Its contigs use `C` as the sample separator, Vamb's default, so the tests pass `-s C`.

## PR checklist

- [x] Add a resource `label`
- [x] `nf-core modules test semibin/singleeasybin --profile singularity`
- [x] `nf-core modules test semibin/multieasybin --profile singularity`

### DIFF
--- a/modules/nf-core/semibin/multieasybin/environment.gpu.yml
+++ b/modules/nf-core/semibin/multieasybin/environment.gpu.yml
@@ -1,0 +1,10 @@
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/environment-schema.json
+channels:
+  - conda-forge
+  - bioconda
+dependencies:
+  - bioconda::semibin=2.4.1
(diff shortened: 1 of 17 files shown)


Joon-Klaps (approved):
Nicely implemented.

PR #12871, opened 2026-09-02
Add isoseq/cluster2 module

mahesh-panchal:
isoseq cluster is phased out upstream in favour of cluster2 (faster, no
memory limit on read count). New module rather than a change to
isoseq/cluster: CLI and outputs differ (no HQ/LQ split, no .cluster/
.transcriptset.xml). isoseq/cluster is deprecated in favour of this in a
follow-up PR.

## PR checklist

- [x] This comment contains a description of changes (with reason).
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [x] If you've added a new tool - have you followed the module conventions in the [contribution docs](https://github.com/nf-core/modules/tree/master/.github/CONTRIBUTING.md)
- [x] Remove all TODO statements.
- [x] Broadcast software version numbers to `topic: versions` - [See version_topics](https://nf-co.re/blog/2025/version_topics)
- [x] Follow the naming conventions.
- [x] Follow the parameters requirements.
- [x] Follow the input/output options guidelines.
- [x] Add a resource `label`
- [x] Use BioConda and BioContainers if possible to fulfil software requirements.
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - For modules:
    - [x] `nf-core modules test isoseq/cluster2 --profile docker`
    - [ ] `nf-core modules test isoseq/cluster2 --profile singularity`
    - [ ] `nf-core modules test isoseq/cluster2 --profile conda`

Generated by Claude Sonnet 5

PR #12869, opened 2026-09-02
Deprecate isoseq3/tag in favour of isoseq/tag

mahesh-panchal:
Depends on #12868.
Closes #11823

Deprecates isoseq3/tag per nf-core guideline (kept, not deleted). script/stub
now assert false with a message pointing at isoseq/tag; meta.yml flagged
deprecated: true; test collapsed to one case asserting the failure. Also
migrates versions output to topic channel, since the file is touched anyway.

## PR checklist

- [x] This comment contains a description of changes (with reason).
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [x] Broadcast software version numbers to `topic: versions` - [See version_topics](https://nf-co.re/blog/2025/version_topics)
- Ensure that the test works with either Docker / Singularity. Conda CI tests can be quite flaky:
  - For modules:
    - [x] `nf-core modules test isoseq3/tag --profile docker`
    - [ ] `nf-core modules test isoseq3/tag --profile singularity`
    - [ ] `nf-core modules test isoseq3/tag --profile conda`

Generated by Claude Sonnet 5

PR #12872, opened 2026-09-02
Deprecate isoseq/cluster in favour of isoseq/cluster2

mahesh-panchal:
Depends on #12871.

Deprecates isoseq/cluster per nf-core guideline (kept, not deleted).
script/stub now assert false pointing at isoseq/cluster2; meta.yml flagged
deprecated: true. Test collapsed to one case asserting the failure, which
also sidesteps a pre-existing sanitizeOutput bug in the old test
(unstableKeys referenced "transcripts" and "xml", neither a real emit
name).

## PR checklist

  - For modules:
    - [x] `nf-core modules test isoseq/cluster --profile docker`
    - [ ] `nf-core modules test isoseq/cluster --profile singularity`
    - [ ] `nf-core modules test isoseq/cluster --profile conda`

Generated by Claude Sonnet 5


bhargava-morampalli (approved):
Same deprecation pattern/implementation as #12869 and looks correct to me.

PR #12880, opened 2026-09-03
New module for delly. Using version 2.x.x with sr instead of call

gustav-001:
New module for delly, with latest version (v2.6.0) with new `sr` instead of `call` command. 

## PR checklist
  - For modules:
 
## Changes
 - Changed all mentions of `call` to `sr` and updated to latest version (v2.6.0)
 - Added path `modules/delly/sr`
 - Added meta.yml with new tool description
 - main.nf command uses `DELLY_SR`
Reason
`call` was changed to `sr` from v2.3.0 onwards, thus changing `call` to `sr` is necessary to support anything from v2.3.0 or later.


modules/nf-core/delly/sr/tests/nextflow.config
```diff
@@ -0,0 +1,5 @@
+process {
+    withName: "DELLY_SR" {
+        cpus = 2
+    }
+}
```
nvnieuwk: This is not necessary since the tests already have a cap on their CPU usage

modules/nf-core/delly/sr/meta.yml
```diff
@@ -0,0 +1,131 @@
...
+      - "delly --version |& sed -n '1s/Delly version: *v//p'":
+          type: eval
+          description: The expression to obtain the version of the tool
+authors:
+  - "@projectoriented"
+  - "@nvnieuwk"
+  - "@gustav-001"
+maintainers:
+  - "@projectoriented"
+  - "@nvnieuwk"
```
nvnieuwk:
```suggestion
  - "@gustav-001"
maintainers:
  - "@nvnieuwk"
```

You can leave me in as maintainer but no need to add other people that are not involved :p

nvnieuwk (commented):
Two small comments

modules/nf-core/delly/sr/tests/nextflow.config
gustav-001: If i remove this file as it's not necessary, do i remove the `test("homo_sapiens - bam, bai, [], [], [], fasta, fai, bcf - config") {` block from `main.nf.test` too?

nvnieuwk:
You should update the test results with `nf-core modules test delly/sr --update`

gustav-001:
> You should update the test results with `nf-core modules test delly/sr --update`

Done, snapshot has been regenerated and old one removed. Removed `tests/nextflow.config` together with the `- config` test in the same commit. Retested on docker and passed 7/7 tests.

nvnieuwk (approved):

PR #12890, opened 2026-09-04
Add easypqp/convert module

jonasscheid:
New module wrapping `easypqp convert` ([EasyPQP](https://github.com/grosenberger/easypqp)): peptide identifications (pepXML or idXML) plus spectra (mzML, mzXML, MGF) -> PSM and peak pickles for spectral library generation. Ported from nf-core/mhcquant.

- Output basenames follow the spectra file, since `easypqp library` pairs pickles to runs by basename.
- Optional UniMod XML input.
- Tests depend on nf-core/test-datasets#2258. Companion module: easypqp/library.

## PR checklist

- [x] If you have added a new module, follow the [nf-core modules guidelines](https://nf-co.re/docs/guidelines/components/modules).
- [x] Add a resource `label`
- [x] `nf-core modules lint <MODULE>`


jonasscheid:
Companion PR: https://github.com/nf-core/modules/pull/12891

modules/nf-core/easypqp/convert/environment.yml
```diff
@@ -0,0 +1,8 @@
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/environment-schema.json
+channels:
+  - conda-forge
+  - bioconda
+dependencies:
+  - "bioconda::easypqp=0.1.59"
+  - "bioconda::pyopenms=3.4.1"
```
erikrikarddaniel, line 8: Any particular reason to pin `pyopenms` to `3.4.1` rather than the current latest, `3.5.0`? easypqp's own bioconda recipe for `0.1.59` doesn't pin a `pyopenms` version itself, so this looks like it could just be bumped to latest -- but flagging rather than suggesting the change directly since I haven't verified `easypqp convert` actually works against 3.5.0.

erikrikarddaniel (commented):
Reviewed with some help from Claude Code. Module conventions look solid overall (bioconda pin for `easypqp` itself is already latest, versions via topic channel, stub test, meta.yml structure, `unimod` correctly handled as an optional broadcast path). One question inline on the `pyopenms` pin, nothing else jumped out.

erikrikarddaniel (approved):
One tiny comment.

modules/nf-core/easypqp/convert/environment.yml
jonasscheid: Yes, deliberate: with pyopenms 3.5.0 `easypqp convert` fails on idXML input because `IdXMLFile.load` no longer accepts the plain Python lists easypqp passes (`Exception: can not handle type of ('...idXML', [], [])`). The conda job on this PR hit exactly that before the pin: https://github.com/nf-core/modules/actions/runs/33905443957/job/101129507645. The biocontainer ships pyopenms 3.4.1, so the pin keeps conda and container behaviour identical. The proper fix is an upper bound in the bioconda easypqp recipe; until then the pin has to stay.

PR #12891, opened 2026-09-04
Add easypqp/library module

jonasscheid:
New module wrapping `easypqp library` ([EasyPQP](https://github.com/grosenberger/easypqp)): PSM and peak pickles from `easypqp convert` -> spectral library TSV. Ported from nf-core/mhcquant.

- Library name controlled via `ext.prefix`; the per-run `*_run_peaks.tsv` is emitted as a separate optional channel.
- The test chains `easypqp/convert` via nf-test `setup`, so CI passes once #12890 is merged.

## PR checklist

- [x] If you have added a new module, follow the [nf-core modules guidelines](https://nf-co.re/docs/guidelines/components/modules).
- [x] Add a resource `label`
- [x] `nf-core modules lint <MODULE>`


modules/nf-core/easypqp/library/environment.yml
```diff
@@ -0,0 +1,8 @@
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/environment-schema.json
+channels:
+  - conda-forge
+  - bioconda
+dependencies:
+  - "bioconda::easypqp=0.1.59"
+  - "bioconda::pyopenms=3.4.1"
```
erikrikarddaniel, line 8: Same `pyopenms=3.4.1` pin as on #12890's `easypqp/convert` -- already confirmed there as deliberate (pyopenms 3.5.0 breaks `IdXMLFile.load` on idXML input, so the pin keeps conda/container behaviour identical). No question here, just noting it's consistent with that PR.

erikrikarddaniel (commented):
Reviewed with some help from Claude Code. Module conventions look solid (bioconda pin for `easypqp` itself is already latest, container tag exists, versions topic, stub test, meta.yml structure). Checked the underlying `easypqp library`/`generate()` source: it filters `infiles` by filename substring (`psmpkl`/`peakpkl`) rather than positional order, so passing all PSM files then all peak files (as this module's `${psmpkl} ${peakpkl}` interpolation does for a multi-run list) is fine -- no pairing bug there. Two small non-blocking notes inline.

modules/nf-core/easypqp/library/tests/main.nf.test
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
+            )
+        }
+    }
+}
```
erikrikarddaniel, line 71: The test only exercises a single run (one psmpkl/peakpkl pair). The meta.yml describes the module as taking "one or more" PSM/peak files -- i.e. the multi-run library-aggregation case, which is presumably the main real-world use. Might be worth a second test case chaining two `EASYPQP_CONVERT` runs and joining/grouping their outputs into one list, just to lock in that the multi-file path actually works end to end. Non-blocking.

erikrikarddaniel (approved):
One comment on test coverage below.

modules/nf-core/easypqp/library/environment.yml
jonasscheid: Yeah, latest easypqp doesn't have a breaking change of pyopenms 3.5.0 implemented, hence the pin. Will be unpinned in future releases once the breaking change is accounted for

modules/nf-core/easypqp/library/tests/main.nf.test
jonasscheid: Good point, wold require a second mzML, which is more of an edge case of this module and I would rather not do it to not pollute the test data more

jonasscheid:
Thanks!

PR #12895, opened 2026-09-06
fix(fastq_qc_trim_filter_setstrandedness): expose fastp's merged reads

pinin4fjords:
## Summary

`FASTQ_FASTQC_UMITOOLS_FASTP` already emits fastp's merged overlapping read pairs (`FASTP.out.reads_merged`), but `FASTQ_QC_TRIM_FILTER_SETSTRANDEDNESS` never captured or re-emitted that channel. Any caller enabling `fastp_merge` gets its merged reads silently dropped, with no output, no error, and no way to consume them downstream.

## What changed

- Added `ch_trim_reads_merged`, captured from `FASTQ_FASTQC_UMITOOLS_FASTP.out.trim_reads_merged` in the fastp branch, initialised to `channel.empty()` elsewhere.
- Emitted it as a new `trim_reads_merged` subworkflow output.
- Documented the new output in `meta.yml`.
- Added an assertion to the existing `homo_sapiens paired-end [fastq] fastp sortmerna` test (which already runs with `fastp_merge = true`) confirming the merged-reads file is non-empty with the expected read count.

This is purely additive — no existing output changes shape, so downstream pipelines that don't consume the new output are unaffected.

Found while reviewing nf-core/riboseq's v2.0.0 release PR, which vendors this subworkflow and hits the bug via `--fastp_merge`.


pinin4fjords:
The failing \`nf-test-changes\` check is an unrelated infra issue, not a problem with this PR: the job builds a Docker image from [adamrtalbot/detect-nf-test-changes](https://github.com/adamrtalbot/detect-nf-test-changes) (pinned at its latest commit, `de3c3c8e`, last updated Nov 2024), whose `Dockerfile` installs from a `python:3.12-slim-bullseye` base. `apt-get install -y git` currently 404s fetching `libperl5.32_5.32.1-4+deb11u5` from `deb.debian.org`'s bullseye-security archive — a package the mirror appears to have dropped.

This is repo-wide right now, not specific to this PR — the identical failure is happening on #12894. Retried 4 times over several hours with no change. `nf-core lint` on the subworkflow content passes, and the fix is verified with a full local nf-test run.

jonasscheid (approved):

pinin4fjords:
Thanks @jonasscheid !

PR #12896, opened 2026-09-06
fix(orftable_fasta_gtf_buildorfcatalogue): keep meta.id as the true sample id

pinin4fjords:
## Summary

`ORFTABLE_FASTA_GTF_BUILDORFCATALOGUE` disambiguates `CUSTOM_ORFNORMALISE`'s output filenames per caller by appending the caller name to `meta.id` (`"${meta.id}.${caller}"`) before calling the module. But `CUSTOM_ORFNORMALISE` writes `meta.id` straight through into each row's `sample_id` column, and `CUSTOM_ORFMERGE`'s consensus view counts distinct `sample_id`s for `n_samples`/`--min-samples` filtering. So an ORF called by two different callers on the *same* sample gets counted as two samples, not one — corrupting the published `n_samples`/`samples` catalogue columns, and the `--min-samples` consensus filter for anyone raising it above the default of 1.

The subworkflow's own doc comment already states the intended contract ("caller id carried as a per-record val (not in meta)") — the code just didn't follow it.

## What changed

- `ch_normalise_in` now sets `meta.caller` instead of mutating `meta.id`.
- Added an `ext.prefix` override in this subworkflow's own `nextflow.config` (`"${meta.id}.${meta.caller}.normalised"`) to keep output filenames byte-identical to before, following the exact convention already established in `custom/orfmerge`'s own test fixture (`tests/setup_prefix.config`).

No test changes needed — the existing `ribotish + ribocode` test already exercises two callers on the same sample id (`sample1`), and its snapshot now correctly shows `sample_id = sample1` (was `sample1.ribotish`/`sample1.ribocode`).

Found while reviewing nf-core/riboseq's v2.0.0 release PR, which vendors this subworkflow.


pinin4fjords:
The failing \`nf-test-changes\` check is an unrelated infra issue, not a problem with this PR: the job builds a Docker image from [adamrtalbot/detect-nf-test-changes](https://github.com/adamrtalbot/detect-nf-test-changes) (pinned at its latest commit, `de3c3c8e`, last updated Nov 2024), whose `Dockerfile` installs from a `python:3.12-slim-bullseye` base. `apt-get install -y git` currently 404s fetching `libperl5.32_5.32.1-4+deb11u5` from `deb.debian.org`'s bullseye-security archive — a package the mirror appears to have dropped.

This is repo-wide right now, not specific to this PR — the identical failure is happening on #12894. Retried 4 times over several hours with no change. `nf-core lint` on the subworkflow content passes, and the fix is verified with a full local nf-test run.

subworkflows/nf-core/orftable_fasta_gtf_buildorfcatalogue/main.nf
```diff
@@ -32,12 +32,13 @@ workflow ORFTABLE_FASTA_GTF_BUILDORFCATALOGUE {
...
-    //    default `${meta.id}` prefix yields caller-disambiguated filenames
+    //    `caller` val. Stash it in `meta.caller` too so this subworkflow's own
+    //    `ext.prefix` config can disambiguate output filenames per caller
     //    (the merger stages multiple BED12s into `beds/*` and needs unique
-    //    names per caller-sample combination).
+    //    names per caller-sample combination) while meta.id keeps carrying
+    //    the true sample id through to CUSTOM_ORFNORMALISE's sample_id column.
     ch_normalise_in = ch_orf_tables.map { meta, table, caller ->
-        [ meta + [ id: "${meta.id}.${caller}" ], table, caller ]
+        [ meta + [ caller: caller ], table, caller ]
```
luisas: Will the meta not have any id: field anymore? Could this break assumptions if this subworkflow was reused somehwere else? 

And now the caller info appears twice in the channel - I guess it makes it easier for the downstream logic to extract, it is just repetitive.

luisas (approved):

subworkflows/nf-core/orftable_fasta_gtf_buildorfcatalogue/main.nf
pinin4fjords: I'd seen the duplication too and wondered. Actually it wasn't necessary-removed.

pinin4fjords: Thanks!

PR #12897, opened 2026-09-07
CI: detect changed components with nf-test instead of detect-nf-test-changes

### LINKED ISSUE #12893: CI: nf-test-changes job fails on all PRs (detect-nf-test-changes Docker build, Debian bullseye EOL)
Since 2026-09-05 every PR fails the `nf-test-changes` job before any test runs. The pinned `adamrtalbot/detect-nf-test-changes@de3c3c8` (v0.0.6) builds from `python:3.12-slim-bullseye`, and after the Debian 11 EOL its `apt-get install git` 404s on the bullseye-security archive.

Upstream issue: https://github.com/adamrtalbot/detect-nf-test-changes/issues/18

Once a fixed release is tagged, the pin in `.github/workflows/nf-test.yml` and `nf-test-gpu.yml` needs bumping.

Examples: #12890, #12891

jonasscheid:
Fixes #12893. The pinned `detect-nf-test-changes` action builds a Debian bullseye image that no longer installs, breaking `nf-test-changes` on every PR.

Replaces it with a local composite action that runs `nf-test test --dry-run --changed-since <merge-base> --related-tests` and maps the related test files to component directories. Same `components` output, so `get-shards` and the test matrix are unchanged.

- No Docker build in CI any more.
- Diff starts at the merge base with `origin/master`, so changes merged to master since the branch diverged are no longer tested.
- Tests that directly include a changed module (e.g. subworkflows) are now picked up too.
- Tag filtering kept via `tags` / `exclude_tags`; the CPU workflow now excludes `gpu_highmem` as well as `gpu`.
- Checked locally against a module-only branch, a GPU-tagged module, and a no-change case.

### DIFF
--- a/.github/actions/detect-nf-test-changes/detect.sh
+++ b/.github/actions/detect-nf-test-changes/detect.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+# Usage: detect.sh <base-ref> [include-tags] [exclude-tags]
+# Prints a JSON array of nf-core component directories whose nf-tests are related
+# to the changes between merge-base(<base-ref>, HEAD) and HEAD.
+set -euo pipefail
+
+base_ref="$1"
+include_tags="${2:-}"
+exclude_tags="${3:-}"
+
+base=$(git merge-base "$base_ref" HEAD)
+
+tests=$(nf-test test --dry-run --changed-since "$base" --related-tests --filter process,workflow 2>&1 \
+    | awk '/related test\(s\)/ {f=1; next} /Dry run mode/ {f=0} f && /\.nf\.test$/ {sub(/^[[:space:]]*•[[:space:]]*/, ""); print}' \
+    | sed "s#^$PWD/##")
+
+has_tag() { grep -qE "^[[:space:]]*tag[[:space:]]+[\"']$2[\"']" "$1"; }
+
+components=()
+for t in $tests; do
+    keep=1
+    if [ -n "$include_tags" ]; then
+        keep=0
+        for tag in ${include_tags//,/ }; do has_tag "$t" "$tag" && keep=1; done
+    fi
+    for tag in ${exclude_tags//,/ }; do has_tag "$t" "$tag" && keep=0; done
+    [ "$keep" = 1 ] || continue
+    comp=$(dirname "$(dirname "$t")")
+    [[ "$comp" =~ ^(modules|subworkflows)/nf-core/ ]] && components+=("$comp")
+done
+
+printf '%s\n' "${components[@]+"${components[@]}"}" | { grep . || true; } | sort -u | jq -R . | jq -cs .
--- a/.github/actions/detect-nf-test-changes/detect.sh
+++ b/.github/actions/detect-nf-test-changes/detect.sh
@@ -8,6 +8,7 @@ include_tags="${2:-}"
 exclude_tags="${3:-}"
 
 base=$(git merge-base "$base_ref" HEAD)
+echo "TEMP: simulated detector failure for the fail-closed CI check" >&2 && exit 3
 csv=$(mktemp)
 rm -f "$csv"
 
--- a/.github/actions/detect-nf-test-changes/detect.sh
+++ b/.github/actions/detect-nf-test-changes/detect.sh
@@ -8,7 +8,6 @@ include_tags="${2:-}"
 exclude_tags="${3:-}"
 
 base=$(git merge-base "$base_ref" HEAD)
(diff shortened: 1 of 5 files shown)


.github/actions/detect-nf-test-changes/detect.sh
```diff
@@ -0,0 +1,32 @@
...
+components=()
+for t in $tests; do
+    keep=1
+    if [ -n "$include_tags" ]; then
+        keep=0
+        for tag in ${include_tags//,/ }; do has_tag "$t" "$tag" && keep=1; done
+    fi
+    for tag in ${exclude_tags//,/ }; do has_tag "$t" "$tag" && keep=0; done
+    [ "$keep" = 1 ] || continue
+    comp=$(dirname "$(dirname "$t")")
```
adamrtalbot: The old action used `n_parents: 0`, so its output contained exact `.nf.test` paths. Converting the filtered paths to component directories undoes the CPU tag exclusion: when a component has both `main.nf.test` and `gpu.nf.test`, passing the directory to the untagged CPU invocation rediscovers and runs both. I reproduced this with `amulety/antiberta2`: the directory dry run selected four tests, including the GPU suite, while the exact CPU test file selected two. Please preserve the exact test-file paths.

Comment added by Codex

.github/actions/detect-nf-test-changes/detect.sh
```diff
@@ -0,0 +1,32 @@
...
+# to the changes between merge-base(<base-ref>, HEAD) and HEAD.
+set -euo pipefail
+
+base_ref="$1"
+include_tags="${2:-}"
+exclude_tags="${3:-}"
+
+base=$(git merge-base "$base_ref" HEAD)
+
+tests=$(nf-test test --dry-run --changed-since "$base" --related-tests --filter process,workflow 2>&1 \
```
adamrtalbot: This parses nf-test's human-readable headings and bullet formatting. If that presentation changes, `awk` exits successfully with no matches and the action emits `[]`, silently skipping the matrix. Please consume a machine-readable report such as the dry-run CSV `filename` column, or validate that a recognised tests-found or no-tests result was observed before emitting an empty selection.

Comment added by Codex

adamrtalbot (changes requested):
Thanks for moving change detection onto native nf-test; removing the external Docker action is the right direction. I found two blocking coverage problems in the current implementation, described inline.

There is also an existing fail-open path that this change should close: `confirm-pass-nf-test` and `confirm-pass-gpu` depend only on the test matrices, not on `nf-test-changes`. A selector or setup failure therefore skips the matrices while the required confirmation checks can still succeed. Please include the selector job in both confirmation jobs' dependencies and explicitly require it to succeed.

The current PR checks exercise only the no-related-tests case because the PR changes CI files; all CPU and GPU matrices were skipped. Before merging, please exercise at least a mixed CPU/GPU component change and a detector-failure case.

Comment added by Codex

.github/actions/detect-nf-test-changes/detect.sh
jonasscheid: Fixed in 21c8f49: the script now emits the exact `.nf.test` paths from nf-test's CSV report. Verified on CI with a temporary touch of `amulety/antiberta2`: CPU selected only `tests/main.nf.test`, the gpu run only `tests/gpu.nf.test`, gpu_highmem nothing (https://github.com/nf-core/modules/actions/runs/34109307779, https://github.com/nf-core/modules/actions/runs/34109307726).

.github/actions/detect-nf-test-changes/detect.sh
jonasscheid: Fixed in 21c8f49: the selection comes from the `--csv` dry-run report (`filename` column). The script exits non-zero if nf-test fails or if there is neither a CSV nor a `Nothing to do` message, so an unrecognised output can no longer yield an empty selection.

jonasscheid:
Addressed all three points:

- Exact test-file output and CSV-based parsing (see inline replies).
- `confirm-pass-nf-test` and `confirm-pass-gpu` now depend on `nf-test-changes` and fail unless it succeeded.

CI exercised via temporary commits, both reverted:
- Mixed CPU/GPU component change: https://github.com/nf-core/modules/actions/runs/34109307779, https://github.com/nf-core/modules/actions/runs/34109307726
- Simulated detector failure, both confirm jobs fail at "Change detection did not succeed" with matrices skipped: https://github.com/nf-core/modules/actions/runs/34110403218, https://github.com/nf-core/modules/actions/runs/34110403245

adamrtalbot (approved):
A little messy, but removes a whole dependency so worth it.

Issue #12903, opened 2026-09-08
`QUAST`: test fails with `conda`


nschan:
### Have you checked the docs?

- [x] [nf-core website: troubleshooting](https://nf-co.re/usage/troubleshooting)
- [x] [nf-core modules documentation](https://nf-co.re/docs/contributing/modules)

### Description of the bug

It appears that `environment.yml` is missing some dependencies. I noticed this during a `conda` test for `genomeassembler`: https://github.com/nf-core/genomeassembler/actions/runs/34197001948/job/101966929399?pr=218#step:3:1077

### Command used and terminal output

```console

```

nschan:
This appears to be caused by python version >=3.12 no longer bundling `distutils`.

nschan:
Closed via #12905

PR #12905, opened 2026-09-08
pin python 3.11 in quast env

nschan:
QUAST does not work with python >3.11 because `distutils` was deprecated in python 3.12. See https://github.com/ablab/quast/issues/298 and #12903


erikrikarddaniel (approved):
No patch?

nschan:
This is modules :)

PR #12904, opened 2026-09-08
Update custom/pcaclustering

dbaku42:
## PR title
feat(custom/pcaclustering): accept PLINK2 `.eigenvec` input

## Description
`custom/pcaclustering` currently expects a generic sample-by-feature TSV
(first column = sample IDs). Downstream use in `snpclustering` therefore
needed a separate `gawk` job to turn `plink2/pca` `.eigenvec` into that TSV
(`#` header, `FID`/`IID`).

This PR teaches the module to read PLINK2 `.eigenvec` directly, so the extra
conversion process is no longer required.

Supported inputs:

- TSV/TXT: first column is sample IDs (any name), remaining columns are
  numeric features (unchanged)
- PLINK2 `.eigenvec`: `#FID IID PC1 ...` — `FID`/`SID` are dropped and `IID`
  is used as `sample_id`

Column detection is based on header names, not file extension.

After this is merged, `#11879` can pass `PLINK2_PCA.out.evecfile` straight
into `CUSTOM_PCACLUSTERING` and drop `GAWK_EIGENVEC_TO_TSV`.

## Changes
- `templates/clustering.py`: parse TSV or PLINK2 `.eigenvec` in `load_features`
- `meta.yml`: document both formats; pattern `*.{tsv,txt,eigenvec}`
- tests: keep existing TSV tests; add `clustering - plink2 eigenvec` using
  `genomics/homo_sapiens/popgen/clustering/test.eigenvec`

`data/genomics/homo_sapiens/popgen/clustering/test.eigenvec`

## Checklist
- [x] This comment contains a description of changes (with reason)
- [x] If you've printed output from the tool, please upload this file
- [x] `nf-core modules lint custom/pcaclustering` — local
- [x] `nf-test` for `custom/pcaclustering` — local, docker

PR #12906, opened 2026-09-08
feat(custom/clustermetrics, custom/clustervisualization): accept PLIN…

dbaku42:
This PR updates `load_features` in both modules so they accept the same
inputs as `custom/pcaclustering`:

- TSV/TXT with a `sample_id` column (unchanged)
- PLINK2 `.eigenvec` (`#FID IID PC1 ...` or `#IID PC1 ...`); `FID`/`SID`
  are dropped and `IID` is used as `sample_id`

Also:

- `custom/clustermetrics`: k-distance plot (`*.k_distance.png`) to help
  choose DBSCAN `eps`. Optional `ext.args`: `--k-neighbors`, `--eps`.
- `custom/clustervisualization`: `n_jobs=1` on UMAP and t-SNE (with the
  existing `random_state=42`) to stabilize embeddings across runners.

`main.nf` input signatures are unchanged (`path(features)`). Docs in
`meta.yml` now list `*.{tsv,txt,eigenvec}`.

### Related
- nf-core/modules#12904
- nf-core/modules#11879 (can drop `GAWK_EIGENVEC_TO_TSV` after this)

### DIFF
--- a/modules/nf-core/custom/clustermetrics/templates/cluster_metrics.py
+++ b/modules/nf-core/custom/clustermetrics/templates/cluster_metrics.py
@@ -13,6 +13,7 @@ import matplotlib
 
 matplotlib.use("Agg")
 import matplotlib.pyplot as plt
+import numpy as np
 import pandas as pd
 import sklearn
 import yaml
@@ -22,16 +23,32 @@ from sklearn.metrics import (
     davies_bouldin_score,
     silhouette_score,
 )
+from sklearn.neighbors import NearestNeighbors
 
 
 def load_features(path):
-    """Read a TSV of `sample_id` + numeric feature columns, indexed by sample_id."""
-    df = pd.read_csv(path, sep="\\t")
+    """Read sample-by-feature matrix, indexed by sample_id.
+
+    Accepts a TSV with a sample_id column, or a PLINK2 .eigenvec file
+    (`#FID IID PC1 ...` or `#IID PC1 ...`). FID/SID are dropped.
+    """
+    df = pd.read_csv(path, sep="\\t", dtype=str)
+    df.columns = [str(col).lstrip("#") for col in df.columns]
+
+    if "IID" in df.columns:
+        df = df.rename(columns={"IID": "sample_id"})
+    elif "sample_id" not in df.columns:
+        df = df.rename(columns={df.columns[0]: "sample_id"})
+
+    drop = [col for col in ("FID", "SID") if col in df.columns]
+    if drop:
+        df = df.drop(columns=drop)
+
     if "sample_id" not in df.columns:
         raise ValueError(f"features file must have a 'sample_id' column. Found: {list(df.columns)}")
+
     df["sample_id"] = df["sample_id"].astype(str)
     return df.set_index("sample_id").apply(pd.to_numeric, errors="coerce").fillna(0.0)
-
 
 def load_clusters(path):
     """Read a CSV of `sample_id` + `cluster`, returning a Series of int labels."""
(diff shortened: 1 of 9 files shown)


modules/nf-core/custom/clustermetrics/templates/cluster_metrics.py
```diff
@@ -22,13 +23,30 @@
...
+    Accepts a TSV with a sample_id column, or a PLINK2 .eigenvec file
+    (`#FID IID PC1 ...` or `#IID PC1 ...`). FID/SID are dropped.
+    """
+    df = pd.read_csv(path, sep="\\t", dtype=str)
+    df.columns = [str(col).lstrip("#") for col in df.columns]
+
+    if "IID" in df.columns:
+        df = df.rename(columns={"IID": "sample_id"})
+    elif "sample_id" not in df.columns:
+        df = df.rename(columns={df.columns[0]: "sample_id"})
```
erikrikarddaniel: Worth double-checking: when neither `IID` nor `sample_id` is present, this now silently renames whichever column happens to be first to `sample_id`, instead of the previous behavior of raising the `ValueError` below (line 48) immediately. A genuinely malformed/mislabeled input (columns in the wrong order, or an ID column named something other than `sample_id`/`IID`) would now be used as-is rather than failing with the old, clear "features file must have a 'sample_id' column" message — any mismatch would instead surface later as a confusing empty join or a garbage plot. Is the silent fallback intentional (e.g. to support a headerless first-column-is-ID convention), or should this `elif` just keep raising, with only the two known PLINK2 header shapes special-cased above it?

modules/nf-core/custom/clustervisualization/templates/cluster_viz.py
```diff
@@ -29,10 +29,26 @@
...
+    Accepts a TSV with a sample_id column, or a PLINK2 .eigenvec file
+    (`#FID IID PC1 ...` or `#IID PC1 ...`). FID/SID are dropped.
+    """
+    df = pd.read_csv(path, sep="\\t", dtype=str)
+    df.columns = [str(col).lstrip("#") for col in df.columns]
+
+    if "IID" in df.columns:
+        df = df.rename(columns={"IID": "sample_id"})
+    elif "sample_id" not in df.columns:
+        df = df.rename(columns={df.columns[0]: "sample_id"})
```
erikrikarddaniel: Same fallback question as in `clustermetrics/templates/cluster_metrics.py` — see that comment.

erikrikarddaniel (approved):
Claude didn't find complete test coverage: "Missing test coverage — the new IID/FID/SID PLINK2-header parsing branch in load_features isn't exercised by any test in either module; both still only feed the plain-TSV fixture, and even custom/pcaclustering's existing "eigenvec" test turns out to reuse that same TSV rather than a real .eigenvec-shaped file."

Two more comments in the code.

Not blocking, but worth thinking about.

dbaku42:
Good catch — that silent fallback wasn't intentional. Removed the elif branch in both cluster_metrics.py and cluster_viz.py; now only IID is special-cased, and anything without sample_id/IID raises the original ValueError again instead of silently renaming the first column.

Also added a test in each module (clustermetrics - features from eigenvec  and clustervisualization - features from eigenvec) using test.eigenvec from test-datasets, which exercises the actual PLINK2 header-parsing branch — turns out it was already sitting unused in the popgen/clustering folder. Snapshots updated accordingly.

dbaku42:
thanks @erikrikarddaniel,
All tests pass now, let me know if you have any other obsevations.

erikrikarddaniel:
> thanks @erikrikarddaniel, All tests pass now, let me know if you have any other obsevations.

You have my approval, just go ahead and merge!

